PR #1364, opened 2015-08-21
gzip compression for tcp connections

gbu-censhare:
enable gzip compression on tcp connection (incoming traffic)

### DIFF
--- a/graylog2-inputs/src/main/java/org/graylog2/inputs/transports/TcpTransport.java
+++ b/graylog2-inputs/src/main/java/org/graylog2/inputs/transports/TcpTransport.java
@@ -28,7 +28,6 @@ import org.graylog2.plugin.configuration.ConfigurationRequest;
 import org.graylog2.plugin.configuration.fields.BooleanField;
 import org.graylog2.plugin.configuration.fields.ConfigurationField;
 import org.graylog2.plugin.configuration.fields.NumberField;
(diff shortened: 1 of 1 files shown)


joschi:
What's the use case for this specifically? Are there any clients which do support GZIP (or any other) compression in this context already?

gbu-censhare:
we have a wide use of  Mapped Diagnostic Context (http://logback.qos.ch/manual/mdc.html) in our application logging (slf4j/logback), tough we have more data per log entry (but, we don't need to have parse rules to get to this data).

but as gelf/json can be quite verbose with the keys, the amount of data is ofcourse  increased  vs stand syslog-logging. thus we want to compress it on transfer. 

i'm not sure if any existing library supports this, we support it in our plain tcp connector, and it should be pretty ease to add e.g. to your log4j connector.

joschi:
I'm still not convinced of the usefulness of this particular change. The GELF codec (or the [GELF specification](https://www.graylog.org/resources/gelf/) in general) already supports compression with ZLib and GZIP and in general it's more of a topic you'd handle in a specific codec rather than on a transport-level.

gbu-censhare:
hm, okay, i must have missed this feature of GELF ...

but a transport stream compression would be able to handle compression more efficently, i guess, as a dictonary for the stream could be build up, not only for one message?

gbu-censhare:
i'll check with single message compression, try to figure out the difference ..

PR #1366, opened 2015-08-24
Retry bulk indexing messages on timeout

### LINKED ISSUE #1270: Global cluster timeout should retry bulk indexing when timeout occurs
In https://github.com/Graylog2/graylog2-server/blob/56e53093b21458d1ca4f9990c89258ecc0c12ac7/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java#L109 we don't catch the TimeoutException thrown by GlobalTimeoutClient (because the code did not retry before anyway).

We should keep on retrying to flush the batch again until it succeeds.

joschi:
Closes #1270

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -150,22 +150,18 @@ public class Messages {
     }
 
     private void propagateFailure(BulkItemResponse[] items, List<Message> messages, String errorMessage) {
-        LOG.error(
-                "Failed to index [{}] messages. Please check the index error log in your web interface for the reason. Error: {}",
-                items.length,
-                errorMessage);
-
         // Get all failed messages.
-        List<DeadLetter> deadLetters = Lists.newArrayList();
+        final List<DeadLetter> deadLetters = Lists.newArrayList();
         for (BulkItemResponse item : items) {
             if (item.isFailed()) {
                 deadLetters.add(new DeadLetter(item, messages.get(item.getItemId())));
             }
         }
 
-        boolean r = deadLetterQueue.offer(deadLetters);
+        LOG.error("Failed to index [{}] messages. Please check the index error log in your web interface for the reason. Error: {}",
+                deadLetters.size(), errorMessage);
 
-        if (!r) {
+        if (!deadLetterQueue.offer(deadLetters)) {
             LOG.debug("Could not propagate failure to failure queue. Queue is full.");
         }
     }
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -18,9 +18,11 @@ package org.graylog2.indexer.messages;
 
 import com.github.joschi.jadconfig.util.Duration;
 import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.WaitStrategies;
 import com.google.common.base.Predicate;
+import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 import org.elasticsearch.ElasticsearchTimeoutException;
 import org.elasticsearch.action.WriteConsistencyLevel;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -67,48 +81,59 @@ public Messages(Client client, ElasticsearchConfiguration configuration) {
...
+                    .call(new Callable<BulkResponse>() {
+                        @Override
+                        public BulkResponse call() throws Exception {
+                            return c.bulk(requestBuilder.request()).actionGet();
+                        }
+                    });
+        } catch (ExecutionException | RetryException e) {
+            LOG.error("Couldn't bulk index " + messages.size() + " messages.", e);
+            return false;
+        }
```
bernd: I think it would be good to use the retryer only in the exceptional case like I did in 6623e80. As it is now, we create lots of garbage objects for every bulk request.

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -67,49 +90,44 @@ public Messages(Client client, ElasticsearchConfiguration configuration) {
...
+        List<String> tokens = Lists.newArrayList();
+        AnalyzeRequestBuilder arb = new AnalyzeRequestBuilder(c.admin().indices(), index, string);
+        AnalyzeResponse r = c.admin().indices().analyze(arb.request()).actionGet();
+
+        for (AnalyzeToken token : r.getTokens()) {
+            tokens.add(token.getTerm());
+        }
+
+        return tokens;
+    }
```
kroepke: whitespace?

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -117,38 +135,60 @@ public boolean bulkIndex(final List<Message> messages) {
...
-                "Failed to index [{}] messages. Please check the index error log in your web interface for the reason. Error: {}",
-                items.length,
-                errorMessage);
+    private BulkResponse runBulkRequest(final BulkRequest request) {
+        try {
+            return c.bulk(request).actionGet();
+        } catch (ElasticsearchTimeoutException timeoutException) {
+            LOG.debug("Bulk indexing request timed out. Retrying.", timeoutException);
+            try {
+                return BULK_REQUEST_RETRYER.call(new BulkRequestCallable(c, request));
```
kroepke: Does the BulkRequest resolve index aliases at this point or later?
If it does so early we might run into the problem that the deflector has already been cycled and the old resolved index is already read_only, so this will fail to write the bulk forever.
Can we instead create the bulk request on every retry by passing the message list?

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
joschi: Yes, now corrected. The code was one level "too deep" before. :wink:

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
joschi: AFAIK the `BulkRequest` itself (the structure being passed to the Elasticsearch client which then performs the actual remote call) doesn't cache or resolve anything on construction. So this should be fine™.

kroepke: ok, cool

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -67,49 +90,44 @@ public Messages(Client client, ElasticsearchConfiguration configuration) {
...
+        return tokens;
+    }
 
     public boolean bulkIndex(final List<Message> messages) {
         if (messages.isEmpty()) {
             return true;
         }
 
-        final BulkRequestBuilder request = c.prepareBulk();
+        final BulkRequestBuilder requestBuilder = c.prepareBulk().setConsistencyLevel(WriteConsistencyLevel.ONE);
```
bernd: Is this consistency level change intentional? How does this change the behaviour?

bernd: Sorry, it's not been changed. Was there before...

PR #1367, opened 2015-08-24
Remove output from OutputRegistry when destroying it

### LINKED ISSUE #1325: stop() method of Output plugin not called after deleting the output globally
Steps to Reproduce:

1. Create an output plugin globally
2. Associate the global instance to a stream
3. See the plugin getting the events 
4. Now delete the plugin globally from System->Outputs
5. The plugin instance is deleted and disassociated from stream, but the stop() method is not called.

joschi:
Fixes #1325

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.outputs;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
```diff
@@ -137,33 +140,35 @@ public MessageOutput call() throws Exception {
...
     public Set<MessageOutput> getMessageOutputs() {
-        final ImmutableSet.Builder<MessageOutput> builder = ImmutableSet.builder();
-
-        builder.addAll(this.runningMessageOutputs.asMap().values());
-        builder.add(defaultMessageOutput);
-        return ImmutableSet.copyOf(builder.build());
+        return ImmutableSet.<MessageOutput>builder()
+                .addAll(runningMessageOutputs.asMap().values())
+                .add(defaultMessageOutput)
+                .build();
```
kroepke: :+1:

kroepke:
lgtm

PR #1369, opened 2015-08-25
Prevent excessive rotation during startup when using time-based retention

kroepke:
Instead of forcing a deflector rotation, use the creation time of the deflector index to determine whether we need to rotate it.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -436,7 +436,7 @@ public class Indices implements IndexManagement {
                 .request();
         try {
             final GetIndexResponse response = c.admin().indices()
-                    .getIndex(indexRequest).actionGet(1,TimeUnit.SECONDS);
+                    .getIndex(indexRequest).actionGet();
             final Settings settings = response.settings().get(index);
             if (settings == null) {
                 return null;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -420,4 +427,24 @@ public ClusterHealthStatus waitForStatus(String index, ClusterHealthStatus clust
...
+
+    @Nullable
+    public DateTime indexCreationDate(String index) {
+        final GetIndexRequest indexRequest = c.admin().indices().prepareGetIndex()
+                .addFeatures(GetIndexRequest.Feature.SETTINGS)
+                .addIndices(index)
+                .request();
+        try {
+            final GetIndexResponse response = c.admin().indices()
+                    .getIndex(indexRequest).actionGet(1,TimeUnit.SECONDS);
```
joschi: Why use an explicit timeout here?

graylog2-server/src/test/java/org/graylog2/indexer/rotation/TimeBasedRotationStrategyTest.java
```diff
@@ -88,13 +101,16 @@ public void shouldRotateHourly() {
...
+        Indices indices = mock(Indices.class);
+        when(indices.indexCreationDate(anyString())).thenReturn(initialTime.minus(Period.minutes(5)));
+
         final Period period = Period.hours(1);
-        final TimeBasedRotationStrategy hourlyRotation = new TimeBasedRotationStrategy(period);
+        final TimeBasedRotationStrategy hourlyRotation = new TimeBasedRotationStrategy(period, indices);
 
         RotationStrategy.Result result;
 
         result = hourlyRotation.shouldRotate("ignored");
```
joschi: As a matter of fact the index name is not ignored anymore. :wink:
But let's leave it that way to keep the change set as small as possible.

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
kroepke: no good reason, i'll remove it.

PR #1371, opened 2015-08-25
Consolidate Graylog Collector resources

### LINKED ISSUE #1273: Collector resource does not respond to GET for collector id.
The collector resource in the server API defines a [GET method](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/rest/resources/system/collector/CollectorResource.java#L96) to retrieve details for a specific collector id. The resource method used to PUT a collector registration is defined in a [separate resource class](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/rest/resources/system/collector/CollectorRegistrationResource.java#L62). This was done because the registration method does not require authentication while all other collector resource methods do and defining authentication to be required class-wide helps to prevent accidental opening of resource methods to unauthenticated users.

Unfortunately Jersey seems not to be able to handle these overlapping URL specs. Although the internal jersey resource configuration looks good, only the PUT method is accessible. This could be solvable either by finding a workaround for Jersey, restructuring the resource classes or adding means to authenticate collectors.

joschi:
Fixes #1273


kroepke:
looks good to me, I'd rather have the security annotations repeated than a resource spread across multiple files, too.

PR #1375, opened 2015-08-26
allow to retrieve field stats for non-numeric fields

kroepke:
All fields, except cardinality, will be 0

 - this works by repeating the query but only asking for the cardinality


graylog2-server/src/main/java/org/graylog2/rest/resources/search/SearchResource.java
```diff
@@ -109,8 +109,12 @@ protected void validateInterval(String interval) {
         try {
             return searches.fieldStats(field, query, filter, timeRange);
         } catch (Searches.FieldTypeException e) {
-            LOG.error("Stats query failed. Make sure that field [{}] is a numeric type.", field);
-            throw new BadRequestException();
+            try {
+                return searches.fieldStats(field, query, filter, timeRange, true, true);
+            } catch (Searches.FieldTypeException e1) {
+                LOG.error("Stats query failed. Make sure that field [{}] is a numeric type.", field);
+                throw new BadRequestException();
```
joschi: Adding a short descriptive message and maybe even the root cause would make sense here. Otherwise the user won't know what went wrong without looking into the Graylog server logs.

kroepke: :+1:

PR #1377, opened 2015-08-28
Use filtered queries instead of post-filter

bernd:
Using post-filter filters only after the search has been done. This will lead to bad performance. When using filtered queries, the filtering is done first.

This reduced the ES query time from 240ms to 150ms on one of our machines.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -184,7 +184,7 @@ public class Searches {
         request.searchType(SearchType.COUNT);
 
         SearchResponse r = c.search(request).actionGet();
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -624,24 +628,29 @@ private SearchHit oneOfIndex(String index, QueryBuilder q, SortOrder sort) {
         }
     }
 
-    private BoolFilterBuilder standardFilters(TimeRange range, String filter) {
-        BoolFilterBuilder bfb = FilterBuilders.boolFilter();
-
-        boolean set = false;
+    private void recordEsMetrics(SearchResponse r, @Nullable TimeRange range) {
```
joschi: :+1:

bernd:
Update after some feedback.

joschi:
LGTM. :+1:

PR #1380, opened 2015-08-31
Fix standard deviation statistical widget

edmundoa:
We accidentally renamed the standard deviation statistical function in 1.1, these changes make both old and new widgets work as intended.

Fixes Graylog2/graylog2-web-interface#1565.


bernd:
LGTM :+1:

PR #1386, opened 2015-09-01
Enable fieldstats for non-numeric fields

kroepke:
Returns NaN for invalid field types

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
@@ -84,7 +84,7 @@ public class FieldStatsResult extends IndexQueryResult {
 
 
     public long getCardinality() {
-        return cardinality == null ? Long.MIN_VALUE : cardinality.getValue();
+        return cardinality != null ? cardinality.getValue() : Long.MIN_VALUE;
     }
 
     public List<ResultMessage> getSearchHits() {
--- a/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
@@ -19,17 +19,17 @@ package org.graylog2.indexer.results;
 import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.search.SearchHits;
-import org.elasticsearch.search.aggregations.bucket.filter.Filter;
 import org.elasticsearch.search.aggregations.metrics.cardinality.Cardinality;
 import org.elasticsearch.search.aggregations.metrics.stats.extended.ExtendedStats;
+import org.elasticsearch.search.aggregations.metrics.valuecount.ValueCount;
 
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.util.List;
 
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
```diff
@@ -28,60 +29,67 @@
...
     }
 
     public long getCount() {
-        return extendedStats != null ? extendedStats.getCount() : 0;
+        if (valueCount != null) {
+            return valueCount.getValue();
+        } else if (extendedStats != null) {
+            return extendedStats.getCount();
+        }
+        return Long.MIN_VALUE;
```
edmundoa: Shouldn't we return 0 here instead of `Long.MIN_VALUE` when the count is not available?

graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
```diff
@@ -28,60 +29,67 @@
...
 
     public double getStdDeviation() {
-        return extendedStats != null ? extendedStats.getStdDeviation() : 0;
+        return extendedStats != null ? extendedStats.getStdDeviation() : Double.NaN;
     }
 
 
     public long getCardinality() {
-        return cardinality == null ? 0 : cardinality.getValue();
+        return cardinality != null ? cardinality.getValue() : Long.MIN_VALUE;
```
edmundoa: Also wondering about `Long.MIN_VALUE` in here :)

kroepke: I originally wanted some kind of `NaN` here, but it's a long, so I wasn't sure how to signal that it isn't available as opposed to being actually 0.

Both cardinality and count cannot be negative by definition, but I agree that the value seems strange.

graylog2-server/src/main/java/org/graylog2/rest/resources/search/SearchResource.java
```diff
@@ -112,7 +112,7 @@ protected void validateInterval(String interval) {
         } catch (Searches.FieldTypeException e) {
             try {
                 LOG.debug("Stats query failed, make sure that field [{}] is a numeric type. Retrying without numeric statistics to calculate the field's cardinality.", field);
-                return searches.fieldStats(field, query, filter, timeRange, true, true);
+                return searches.fieldStats(field, query, filter, timeRange, true, false, false);
```
edmundoa: I think we should be calling this with a `true` in the include count argument:
`return searches.fieldStats(field, query, filter, timeRange, true, false, true);`

graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
edmundoa: Maybe we can check in the web interface if the value for count or cardinality is negative and show N/A in that case. What do you think about that?

edmundoa: I added the code to support that in here: https://github.com/Graylog2/graylog2-web-interface/commit/c89cd319f9d60582ef4297460db007de4e554c93

edmundoa:
As commented to @kroepke, I flipped the count flag when getting statistics for non-numeric fields and this is ready to merge now :shipit:

Issue #951, opened 2015-02-06
Support for LDAP Groups


rikatz:
Not an issue, but a feature request for next versions:

It would be great if support to LDAP groups is added to Graylog.

Example: I have my user, it's part of the group 'group1'.

On the stream, I add that everybody on the group 'group1' have the view of that stream, so I don't need to add manually a user to a stream when it joins a new group (a development group, as an example).

And thanks for the hard work guys, the software is amazing!!

zolech:
+1

 LDAP Group synchronization is what we are waiting for

rikatz:
@kroepke great job!!! Thanks guys for this awesome software, Graylog is transforming on a game changer!!

PR #1387, opened 2015-09-01
in LDAP group lookup use the user specified filter instead of limitin…

kroepke:
…g it to objectClass

 previously we only allowed to filter for objectClass, which could be problematic if LDAP contained thousands of groups.
 this lets the user specify an arbitrary filter for looking up the relevant groups

 #951

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
@@ -48,6 +48,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
+
 public class LdapConnector {
     private static final Logger LOG = LoggerFactory.getLogger(LdapConnector.class);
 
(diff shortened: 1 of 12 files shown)


graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
```diff
@@ -190,21 +192,28 @@ public LdapEntry search(LdapNetworkConnection connection,
...
-                            groups.add(groupId);
+                    // test if the given dn parameter is actually member of any of the found groups
+                    String memberAttribute;
+                    if (e.hasObjectClass("groupOfUniqueNames")) {
+                        memberAttribute = "uniqueMember";
+                    } else if (e.hasObjectClass("groupOfNames") || e.hasObjectClass("group")) {
+                        memberAttribute = "member";
+                    } else {
+                        LOG.warn(
+                                "Unable to auto-detect the LDAP group object class, using assuming 'member' is the correct attribute.");
```
joschi: "using assuming"? Should probably only be "assuming".

graylog2-server/src/main/java/org/graylog2/security/ldap/LdapSettingsImpl.java
```diff
@@ -353,14 +353,14 @@ public void setGroupIdAttribute(String groupIdAttribute) {
     }
 
     @Override
-    public String getGroupObjectClass() {
-        final Object o = fields.get(GROUP_OBJECT_CLASS);
+    public String getGroupSearchPattern() {
+        final Object o = fields.get(GROUP_SEARCH_PATTERN);
```
joschi: Could be simply `return Strings.nullToEmpty((String) fields.get(GROUP_SEARCH_PATTERN));`

joschi:
LGTM. :+1:

PR #1388, opened 2015-09-02
Do not recalculate index ranges on index rotation

joschi:
Since the index range information is stored directly inside the Elasticsearch indices themselves, recalculating them after index rotation is unnecessary. Before the change, this was necessary for housekeeping in the MongoDB "index_ranges" collection.


kroepke:
:+1:

PR #1389, opened 2015-09-02
Migrate user permissions to roles

### LINKED ISSUE #1382: Migrate user permissions
There are some cases where the old permissions management clashes with roles. To avoid those situations we should automatically migrate all users with `*` permissions into the `admin` role. We could also do something similar for other users, migrating users with reader permissions into the `reader` role.

kroepke:
add periodical to migrate user's permissions to admin or reader roles during startup of the master

 - existing stream and dashboard permissions are preserved
 - existing other permissions set via the API are preserved
 - RestPermissions.readerPermissions are migrated to reader role
 - '*' is migrated to admin role

fixes #1382

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
@@ -1,0 +1,167 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.periodical;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import org.graylog2.plugin.database.ValidationException;
+import org.graylog2.plugin.database.users.User;
+import org.graylog2.plugin.periodical.Periodical;
+import org.graylog2.shared.security.RestPermissions;
+import org.graylog2.shared.users.UserService;
+import org.graylog2.users.RoleService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * This task migrates users' permission sets from pre-1.2 style to the built-in roles: admin and reader.
+ *
+ * It does so by preserving the existing stream and dashboard permissions, which can still be assigned to individual users.
+ * It is recommend to migrate to using roles exclusively, but for now both styles will work.
+ */
+public class UserPermissionMigrationPeriodical extends Periodical {
+    private static final Logger log = LoggerFactory.getLogger(UserPermissionMigrationPeriodical.class);
+
+    private final UserService userService;
+    private final RoleService roleService;
+
+    @Inject
+    public UserPermissionMigrationPeriodical(final UserService userService, final RoleService roleService) {
+        this.userService = userService;
+        this.roleService = roleService;
+    }
+
+    @Override
+    public void doRun() {
+        final List<User> users = userService.loadAll();
+        final String adminRoleId = roleService.getAdminRoleObjectId();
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
```diff
@@ -0,0 +1,167 @@
...
+        for (User user : users) {
+            if (user.isLocalAdmin()) {
+                log.debug("Skipping local admin user.");
+                continue;
+            }
+
+            final Set<String> fixedPermissions = Sets.newHashSet();
+            final Set<String> fixedRoleIds = Sets.newHashSet(user.getRoleIds());
+
+            final HashSet<String> permissionSet = Sets.newHashSet(user.getPermissions());
```
joschi: Minor nitpick: Why declare the variable as `HashSet` (implementation) instead of `Set` (interface)?

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
```diff
@@ -0,0 +1,167 @@
...
+        this.roleService = roleService;
+    }
+
+    @Override
+    public void doRun() {
+        final List<User> users = userService.loadAll();
+        final String adminRoleId = roleService.getAdminRoleObjectId();
+        final String readerRoleId = roleService.getReaderRoleObjectId();
+
+        for (User user : users) {
```
joschi: So the migration process will always run on start. Is there some easy way to just run it only once?

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
```diff
@@ -0,0 +1,167 @@
...
+            if (!user.getRoleIds().isEmpty() && hasCompleteReaderSet && hasWildcardPermission) {
+                log.debug("Not migrating user {}, it has already been migrated.", user.getName());
+                continue;
+            }
+            if (hasCompleteReaderSet && !user.getRoleIds().contains(readerRoleId)) {
+                // need to add the reader role to this user
+                fixedRoleIds.add(readerRoleId);
+            }
+            // filter out the individual permissions to dashboards and streams
+            final ArrayList<String> dashboardStreamPermissions = Lists.newArrayList(
```
joschi: Minor nitpick: Why declare the variable as `ArrayList` (implementation) instead of `List` (interface)?

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
```diff
@@ -0,0 +1,167 @@
...
+                    Sets.filter(permissionSet,
+                                new Predicate<String>() {
+                                    @Override
+                                    public boolean apply(
+                                            String permission) {
+                                        return !basePermissions.contains(
+                                                permission) && !permission.equals(
```
joschi: This formatting looks weird.

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
joschi: For example you could store a simple boolean in MongoDB using `ClusterConfigService`.

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
kroepke: because cmd-v did it.
i'll change it (how can i make intellij make a better choice!?)

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
kroepke: will do, even though i think it will not make much of a difference, because it won't touch users after they are migrated.
still might be confusing people on debug level

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
```diff
@@ -0,0 +1,180 @@
...
+        this.clusterConfigService = clusterConfigService;
+    }
+
+    @Override
+    public void doRun() {
+
+        final UserPermissionMigrationState migrationState =
+                clusterConfigService.getOrDefault(UserPermissionMigrationState.class,
+                                                  UserPermissionMigrationState.create(false));
+        if (migrationState.migrationDone()) {
```
joschi: We could probably just move this into the `startOnThisNode()` method and return `migrationState.migrationDone()`, which would just not run the periodical in clusters which have already been migrated.

kroepke: JA :+1:

graylog2-server/src/main/java/org/graylog2/periodical/UserPermissionMigrationPeriodical.java
joschi: :white_check_mark:

joschi:
LGTM. :+1:

PR #1390, opened 2015-09-02
Initialise persisted widgets with wrong timeranges

edmundoa:
When a persisted widget contains an invalid timerange, initialise the widget nonetheless, so it can be edited or deleted from the web interface. Unfortunately I couldn't find a better way of handling this error than initialising the affected widgets with a `null` timerange.

In that case, an error will be logged into the graylog server logs, and an exception will be raised when computing the value for the affected widgets. That exception will be visible in the web interface once the changes in https://github.com/Graylog2/graylog2-web-interface/pull/1590 are merged.

Fixes Graylog2/graylog2-web-interface#1505.


graylog2-server/src/main/java/org/graylog2/dashboards/widgets/StreamSearchResultCountWidget.java
```diff
@@ -47,6 +47,10 @@ public StreamSearchResultCountWidget(MetricRegistry metricRegistry, Searches sea
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/StatisticalCountWidget.java
```diff
@@ -140,6 +140,10 @@ private Number getStatisticalValue(FieldStatsResult fieldStatsResult) {
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/StackedChartWidget.java
```diff
@@ -89,6 +89,10 @@ public StackedChartWidget(MetricRegistry metricRegistry, Searches searches, Stri
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/SearchResultCountWidget.java
```diff
@@ -75,6 +75,10 @@ protected ComputationResult compute() {
     }
 
     protected ComputationResult computeInternal(String filter) {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/SearchResultChartWidget.java
```diff
@@ -69,6 +69,10 @@ public TimeRange getTimeRange() {
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/QuickvaluesWidget.java
```diff
@@ -89,6 +89,10 @@ public TimeRange getTimeRange() {
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/FieldChartWidget.java
```diff
@@ -77,6 +77,10 @@ public FieldChartWidget(MetricRegistry metricRegistry, Searches searches, String
 
     @Override
     protected ComputationResult compute() {
+        if (timeRange == null) {
+            throw new RuntimeException("Invalid time range provided");
```
joschi: `Preconditions.checkArgument(timeRange != null);`

joschi:
There's a lot of repetition when checking the `timeRange` field. Can we extract that into a common parent class or better check it in the constructor of those widgets and throw an `IllegalArgumentException` there?

edmundoa:
The best I can think of to avoid so much repetition is to move that logic into the TimeRange classes, but those classes also have no common parent (other than an interface), so we would need to create an abstract class or add a `checkTimeRange` method to all the hierarchy. I don't think that would be much better though.

We can't raise an exception in the widget constructors, because otherwise we are in the same situation we were at the beginning: the widgets are not [initialised with the dashboard](https://github.com/Graylog2/graylog2-server/blob/web-issue-1505/graylog2-server/src/main/java/org/graylog2/dashboards/DashboardServiceImpl.java#L96) and can't be found when they need to be edited or deleted. Also that wouldn't help with code repetition, as `timeRange` is not in any common parent of the widgets.

edmundoa:
I'm using `Preconditions.checkArgument` as suggested. Unfortunately I don't know how to fix the code duplication without doing lots of yak shaving.

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/DashboardWidgetCreator.java
```diff
@@ -20,15 +20,13 @@
 import com.codahale.metrics.MetricRegistry;
 import com.mongodb.BasicDBObject;
 import org.graylog2.indexer.searches.Searches;
-import org.graylog2.indexer.searches.timeranges.AbsoluteRange;
-import org.graylog2.indexer.searches.timeranges.InvalidRangeParametersException;
-import org.graylog2.indexer.searches.timeranges.KeywordRange;
-import org.graylog2.indexer.searches.timeranges.RelativeRange;
-import org.graylog2.indexer.searches.timeranges.TimeRange;
+import org.graylog2.indexer.searches.timeranges.*;
```
joschi: Please don't use wildcard imports.

Instructions for IntelliJ IDEA: http://myshittycode.com/2013/12/30/intellij-preventing-wildcard-imports/

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/DashboardWidgetCreator.java
```diff
@@ -104,28 +105,35 @@ public DashboardWidget fromPersisted(Searches searches, BasicDBObject fields) th
...
+                    String to = new DateTime(timerangeConfig.get("to"), DateTimeZone.UTC).toString(Tools.ES_DATE_FORMAT);
+
+                    timeRange = new AbsoluteRange(from, to);
+                    break;
+                default:
+                    throw new InvalidRangeParametersException("range_type not recognized");
+            }
+        } catch (InvalidRangeParametersException e) {
+            LOG.error("Invalid time range provided on widget " + widgetId, e);
+            timeRange = null;
```
joschi: I think this will lead to problems in `getPersistedConfig()` of the concrete widget classes.

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/DashboardWidgetCreator.java
edmundoa: It looks like the setting changed when updating to the latest version, will do it again :cry:

edmundoa:
I moved the `timeRange` field to `DashboardWidget` as we spoke, and also moved any checks on the field to that class. To access the `timeRange` down in the hierarchy we can only use a getter that will check the `timeRange` and raise an exception if `null`.

I think this way is a bit cleaner and nicer, hope you find it too :)

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/DashboardWidget.java
```diff
@@ -91,6 +95,15 @@ public int getCacheTime() {
         return cacheTime;
     }
 
+    public TimeRange getTimeRange() {
+        this.checkTimeRange();
+        return timeRange;
+    }
+
+    private void checkTimeRange() {
```
joschi: FWIW, I don't think it's necessary to extract that single check into its own method.

joschi:
LGTM. :+1:

PR #1393, opened 2015-09-03
Allow disabling TLS certificate check in REST client

joschi:
AsyncHttpClient is verifying SSL server certificates since version 1.9.x. Previous versions did also accept untrusted/unknown certificates.

This PR adds a configuration option to configure the behavior, with the pre-1.9.x behavior being default.


bernd:
Tested without REST TLS, with self-signed REST TLS and with an "official" certificate.

LGTM :+1:

PR #1397, opened 2015-09-07
Add missing short-circuit to stream router engine

### LINKED ISSUE #1396: 1.2.0-rc.2 stream rules problem with facility
We currently have two graylog servers, one with the new 1.2.0-rc.2 version and the old one with v1.1.6.

Both of them have the same streams, because the old one forwards through an output to the new one, and one of the streams gets more messages than it should on the new RC version only.

The rules configured for this particular stream are:
- ctxt_stream must match exactly feed
- facility must match exactly shop.live

The wrong behaviour is that the stream receives messages with only facility condition true, but without "ctxt_stream" attribute to be even present. If I remove the facility rule it seems to be fine.

Can you replicate this bug? Otherwise tell me how we can debug this, because on 1.1.6 are routed only the correct messages.

Thank you!

joschi:
Fixes #1396


bernd:
LGTM :+1:

PR #1401, opened 2015-09-08
Convert configuration values in Alarmcallback Config update

### LINKED ISSUE #1399: Alarm callbacks and inputs need type conversion for configuration fields
These API endpoints can accidentally accept string values for number fields when creating or editing the configuration.

This leads to errors (mostly NPEs) when using the configuration. Use `org.graylog2.utilities.ConfigurationMapConverter#convertValues` to convert to correct values wherever possible.

kroepke:
ensure that alarm callback configurations conform to the declared types in the requested plugin config

 - the web interface had a bug where integers were treated as strings on updates causing the wrong config to be stored
 - also fix an NPE that could happen with exceptions having a null message
 - remove AlarmCallbackConfigurationService#update in favor of load/update object/save to allow for sanitizing the configuration
 - removed various author javadocs which we don't use in our codebase

fixes #1399

### DIFF
--- a/graylog2-rest-models/src/main/java/org/graylog2/rest/models/alarmcallbacks/AlarmCallbackError.java
+++ b/graylog2-rest-models/src/main/java/org/graylog2/rest/models/alarmcallbacks/AlarmCallbackError.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
+import com.google.common.base.Strings;
 
 @AutoValue
 @JsonAutoDetect
@@ -32,6 +33,6 @@ public abstract class AlarmCallbackError extends AlarmCallbackResult {
 
     @JsonCreator
     public static AlarmCallbackError create(@JsonProperty("error") String error) {
-        return new AutoValue_AlarmCallbackError(error);
+        return new AutoValue_AlarmCallbackError(Strings.nullToEmpty(error));
     }
 }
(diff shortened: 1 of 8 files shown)


graylog2-rest-models/src/main/java/org/graylog2/rest/models/alarmcallbacks/AlarmCallbackError.java
```diff
@@ -32,6 +33,6 @@
 
     @JsonCreator
     public static AlarmCallbackError create(@JsonProperty("error") String error) {
-        return new AutoValue_AlarmCallbackError(error);
+        return new AutoValue_AlarmCallbackError(Strings.nullToEmpty(error));
```
joschi: Why is this necessary? The error message shouldn't be `null` (or empty for that matter) but mention the reason for the error.

graylog2-server/src/main/java/org/graylog2/rest/resources/alarmcallbacks/AlarmCallbackResource.java
```diff
@@ -207,17 +215,55 @@ public void delete(@ApiParam(name = "streamid", value = "The stream id this alar
...
+                aCC.getCreatorUserId());
+
+        try {
+             alarmCallbackConfigurationService.save(newConfig);
+        } catch (ValidationException e) {
+            throw new BadRequestException("Unable to save alarm callback configuration", e);
+        }
+    }
+
+    private Map<String, Object> convertConfigurationValues(@ApiParam(name = "JSON body", required = true) CreateAlarmCallbackRequest alarmCallbackRequest) {
```
joschi: Unnecessary `@ApiParam` annotation.

graylog2-rest-models/src/main/java/org/graylog2/rest/models/alarmcallbacks/AlarmCallbackError.java
bernd: We ran into this where `error` was null when we had a null pointer exception.

kroepke: This caused an exception when rendering the alarm callback history item, because error is not nullable. Since we didn't want to chase all the usages of this field in the web interface code, we decided to make it "" instead.

edmundoa:
LGTM :+1:

PR #1403, opened 2015-09-08
allow Message timestamp to be a java.util.Date and convert it properl…

### LINKED ISSUE #1394: Grok date converter creates bad timestamp field
For the following example message:

    [03/Sep/2015:06:32:07 +0000]

Using the following grok pattern:

    \[%{HTTPDATE:timestamp;date;dd/MMM/yyyy:HH:mm:ss Z}\]

In the extractor interface, the date is successfully parsed out, and shows:

    timestamp
    Thu Sep 03 06:32:07 UTC 2015

However, when a message actually comes in and is routed through the extractor, it fails when trying to write to the index with the following exception:

    WriteFailureException; nested: MapperParsingException[failed to parse [timestamp]]; nested: 
    MapperParsingException[failed to parse date field [2015-09-03T06:32:07.000Z], tried both date 
    format [yyyy-MM-dd HH:mm:ss.SSS], and timestamp number with locale []]; nested: 
    IllegalArgumentException[Invalid format: "2015-09-03T06:32:07.000Z" is malformed at
    "T06:32:07.000Z"];

It seems like the grok converter is turning it into an ISO 8601 date string, and graylog is failing when trying to parse that into a timestamp field.

The only workaround I've found is to first use a RegEx extractor with a date converter to pull out the timestamp, and then use the grok pattern on the rest of the message, adding an UNWANTED tag to the HTTPDATE pattern.

kroepke:
…y to DateTime

 - this fixes output from java-grok, where we currently do not allow converters to run (because of UI limitations)
 - also applies to any plugin that accidentally sets timestamp to a java.util.Date

 fixes #1394

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
@@ -40,6 +40,7 @@ import org.slf4j.LoggerFactory;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


joschi:
LGTM. :+1:
Paw bump

PR #1406, opened 2015-09-08
on syncing a user account from LDAP allow existing role assignments t…

### LINKED ISSUE #1405: Retain user roles for LDAP users when group mapping or config is missing
In order to make roles usable for LDAP users, without having a LDAP group mapping, Graylog should take into account whether the optional group search settings are set or not and whether any group mapping is configured.

Also allow editing roles of users even if they are created from a LDAP account.

This can have weird behavior when the default role or additional default roles are changed in the LDAP configuration, which was the reason this wasn't possible in the first place.
On each new login the entire account is synced from LDAP, including the roles. However, Graylog cannot differentiate between setting roles manually or them having been created via a (previous) mapping.
We'll have to live with this corner case for now.

kroepke:
…o survive if no group mapping has been configured

fixes #1405


edmundoa:
LGTM :+1:

PR #1412, opened 2015-09-10
Optimization: Only save changed index ranges

joschi:
In order to prevent potentially long-running cluster-level operations (like setting a write block on an index), the EsIndexRangeService#save() method now checks, whether an identical index range has already been stored for the index and skips saving the new but identical index range in this case.


kroepke:
:santa:

PR #1418, opened 2015-09-14
Fix isMatched results for OR streams

### LINKED ISSUE #1417: Test OR stream rules against message returns incorrect value
On 1.2, when testing a message against a stream having more than one stream rule, and using an OR operator, the result of the test is incorrect when the result of all rules is the same.

These are the responses the web interface forwards from the server:
`{"matches":true,"rules":{"55f6b28bbee8968146a18918":false,"55f6b285bee8968146a18911":false}}`
`{"matches":false,"rules":{"55f6b285bee8968146a18911":true,"55f6b28bbee8968146a18918":true}}`

edmundoa:
When testing a stream using an OR operator, check if any of the values for the rules is true instead of false. I also added some tests for this case.

Fixes #1417

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -319,7 +319,7 @@ public class StreamRouterEngine {
         public boolean isMatched() {
             switch (matchingType) {
                 case OR:
-                    return matches.values().contains(false);
+                    return matches.values().contains(true);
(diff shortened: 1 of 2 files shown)


dennisoelkers:
Could you maybe add an integration test similar to the test case but testing the REST endpoint too?

edmundoa:
Added integration tests for it here: https://github.com/Graylog2/graylog2-server/commit/5579eaa19c72c209b75fa827d59680a3d41f6e6e

PR #1425, opened 2015-09-16
Do not enforce grok patterns field in content packs

edmundoa:
Content packs created with older Graylog version will not contain that field, but we still need to be able to import them.

Fixes Graylog2/graylog2-web-interface#1605.


edmundoa:
Forgot to mention, that it needs to be cherry-picked into 1.2 after being merged :+1:

joschi:
Merged into `1.2` in 2307f974accb775029ac91678dd6c9ac048bd147.

Issue #1428, opened 2015-09-17
Cannot load alert callbacks created on 1.0


edmundoa:
Graylog 1.2.0 can't load alert callbacks created with Graylog 1.0 or earlier. The problem is that the MongoDB documents for those old alerts contain two ID fields: `_id`, and `id`. The first one is an `ObjectId` whereas the second one is a `string`, and that is confusing `MongoJack` when loading the objects.

A temporary solution while we try to fix it in a better way, is to run this MongoDB command:
`db.alarmcallbackconfigurations.update({"id": {$exists: true}}, {$unset: {"id": ""}}, { multi: true })`

**Note**: Don't forget to [backup your MongoDB](http://docs.mongodb.org/manual/tutorial/backup-and-restore-tools/) if you run the previous command. Please stop Graylog server before doing it, and only restart it after the backup and the update command are completed.

edmundoa:
Thank you for the review!
![](http://www.reactiongifs.com/r/crwdc.gif)

PR #1429, opened 2015-09-18
Remove duplicated ID field from alarm callbacks

edmundoa:
This fixes issue #1428. In order to migrate old alarm callbacks and be able to load them, I added a periodical that will remove the duplicated ID.

Here is an example of alarm callback with the problem described in the issue:
```
{
    "_id": ObjectId("5464af3a24acdd8389e0f577"),
    "id": "5464af3a24acdd8389e0f577",
    "created_at": ISODate("2014-11-13T13:16:42.690Z"),
    "stream_id": "55fabc4fbee88fa9e9b2350c",
    "creator_user_id": "admin",
    "configuration": {
        "sender": "foo@bar",
        "body": "##########\r\nDate: ${check_result.triggeredAt}\r\nStream ID: ${stream.id}\r\nStream title: ${stream.title}\r\nStream URL: ${stream_url}\r\n\r\nTriggered condition: ${check_result.triggeredCondition}\r\n##########\r\n\r\nLast messages accounting for this alert:\r\n${if backlog_size > 0}${foreach backlog message}\r\n${message}\r\n${end}\r\n${else}<No backlog.>${end}\r\n\r\n",
        "subject": "Graylog2 alert for stream: ${stream.title}"
    },
    "type": "org.graylog2.alarmcallbacks.EmailAlarmCallback"
}
```

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
@@ -17,10 +17,7 @@
 package org.graylog2.alarmcallbacks;
 
 import com.google.common.collect.Lists;
-import com.mongodb.BasicDBObject;
 import com.mongodb.DBCollection;
-import com.mongodb.DBObject;
-import com.mongodb.QueryBuilder;
 import org.bson.types.ObjectId;
 import org.graylog2.bindings.providers.MongoJackObjectMapperProvider;
 import org.graylog2.database.CollectionName;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
```diff
@@ -94,4 +100,13 @@ private AlarmCallbackConfigurationAVImpl implOrFail(AlarmCallbackConfiguration c
             throw new IllegalArgumentException("Supplied output must be of implementation type AlarmCallbackConfigurationAVImpl, not " + callback.getClass());
         }
     }
+
+    // Remove duplicated ID stored in `id` field to avoid MongoJack fetching it instead of `_id`.
+    // See https://github.com/Graylog2/graylog2-server/issues/1428 for more details.
+    public void migrate() {
```
joschi: I'd rather move this method into `AlarmCallbacksMigrationPeriodical` than "pollute" the public interface of `AlarmCallbackConfigurationServiceMJImpl` and having to cast the injected `AlarmCallbackConfigurationService` in that periodical.

edmundoa: I saw we did something like this on `MongoIndexRangeService` and thought it was a bit cleaner than having all the logic in the periodical, but you are probably right. Will take care of it.

graylog2-server/src/main/java/org/graylog2/periodical/AlarmCallbacksMigrationPeriodical.java
```diff
@@ -0,0 +1,103 @@
...
+    public AlarmCallbacksMigrationPeriodical(MongoConnection mongoConnection) {
+        final String collectionName = AlarmCallbackConfigurationAVImpl.class.getAnnotation(CollectionName.class).value();
+        this.dbCollection = mongoConnection.getDatabase().getCollection(collectionName);
+    }
+
+    @Override
+    public void doRun() {
+        LOG.debug("Starting alarm callbacks migration");
+
+        final DBObject selection = QueryBuilder.start("id").exists(true).get();
```
joschi: Either replace `"id"` with the `DUPLICATED_ID` constant or the other way round in the next line. IMHO, using the string literal is completely sufficient here. :wink:

edmundoa: Makes sense :smile:

joschi:
LGTM. :+1:

PR #1437, opened 2015-09-21
Fix Stream#MatchingType handling when applying content packs

### LINKED ISSUE #1434: Serialization error while applying content pack
After importing and applying the content pack, getting following error: can't serialize class org.graylog2.plugin.streams.Stream$MatchingType

MatchingType enum in Stream class needs to implement Serializable interface.

joschi:
Fixes #1434

### DIFF
--- a/integration-tests/src/test/java/integration/system/bundles/ContentPackStreamsTest.java
+++ b/integration-tests/src/test/java/integration/system/bundles/ContentPackStreamsTest.java
@@ -1,0 +1,78 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/integration/system/bundles/ContentPackStreamsTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void createContentPackWithStreams() {
+        final ValidatableResponse validatableResponse = given().when()
+                .body(jsonResourceForMethod()).post("/system/bundles")
+                .then()
+                .statusCode(201)
+                .statusLine(notNullValue());
+
+        uri = URI.create(validatableResponse.extract().header(HttpHeaders.LOCATION));
```
dennisoelkers: Could you please return this instead of passing it around in a field?
And: as this is also used as a test case, could you please add assertions testing the returned location?

joschi: JUnit requires test methods to be of type `void` (i. e. have no return value).

The only assertion I could think of would be quite short (e. g. testing the URI's path started with `/system/bundles`). There are implicit assertions here which are already tested, like the response status code (201) and that the status line ("HTTP/1.1 200 **OK**") wasn't blank.

PR #1447, opened 2015-09-24
Pluggable Password Hashing Support

dennisoelkers:
This PR adds functionality to allow users to change the password algorithm used for hashing user passwords. It also chooses the algorithm used for verifying a supplied user password according to the format of the stored password of the user. Plugins are also able to supply password hashing algorithms.

Two algorithms are supplied:
- SHA-1 for legacy purposes (as passwords have been stored in that format before)
- BCrypt

Two configuration options have been added:

``user_password_default_algorithm``: Valid options are "bcrypt" and "sha-1", defaulting to "bcrypt"
``user_password_bcrypt_salt_size``: If bcrypt is used, this defines the complexity. Defaulting to 10. Higher numbers increase computing cost for each hashing.

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
@@ -1,17 +1,17 @@
 /**
  * The MIT License
  * Copyright (c) 2012 Graylog, Inc.
- *
+ * <p>
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
- *
+ * <p>
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
- *
+ * <p>
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -24,6 +24,8 @@ package org.graylog2.plugin.security;
 
 public interface PasswordAlgorithm {
     boolean supports(String hashedPassword);
+
     String hash(String password);
+
     boolean matches(String hashedPassword, String otherPassword);
 }
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
@@ -1,0 +1,7 @@
+package org.graylog2.plugin.security;
+
+public interface PasswordAlgorithm {
+    boolean supports(String hashedPassword);
+    String hash(String password);
+    boolean matches(String hashedPassword, String otherPassword);
+}
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
@@ -1,17 +1,17 @@
 /**
  * The MIT License
  * Copyright (c) 2012 Graylog, Inc.
- * <p>
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
- * <p>
+ *
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
- * <p>
+ *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
@@ -1,3 +1,25 @@
+/**
+ * The MIT License
+ * Copyright (c) 2012 Graylog, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
 package org.graylog2.plugin.security;
 
 public interface PasswordAlgorithm {
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -134,6 +134,12 @@ public class Configuration extends BaseConfiguration {
     @Parameter(value = "dashboard_widget_default_cache_time", validator = PositiveDurationValidator.class)
     private Duration dashboardWidgetDefaultCacheTime = Duration.seconds(10L);
 
+    @Parameter(value = "user_password_default_algorithm")
+    private String userPasswordDefaultAlgorithm = "bcrypt";
+
+    @Parameter(value = "user_password_bcrypt_salt_size")
+    private Integer userPasswordBCryptSaltSize = 10;
+
     public boolean isMaster() {
         return isMaster;
     }
@@ -269,5 +275,13 @@ public class Configuration extends BaseConfiguration {
 
     public Duration getDashboardWidgetDefaultCacheTime() {
         return dashboardWidgetDefaultCacheTime;
+    }
+
+    public String getUserPasswordDefaultAlgorithm() {
+        return userPasswordDefaultAlgorithm;
+    }
+
+    public Integer getUserPasswordBCryptSaltSize() {
+        return userPasswordBCryptSaltSize;
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -281,7 +281,7 @@ public class Configuration extends BaseConfiguration {
         return userPasswordDefaultAlgorithm;
     }
 
-    public Integer getUserPasswordBCryptSaltSize() {
+    public int getUserPasswordBCryptSaltSize() {
         return userPasswordBCryptSaltSize;
     }
 }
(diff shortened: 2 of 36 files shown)


graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
```diff
@@ -0,0 +1,29 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.graylog2.plugin.security;
+
+public interface PasswordAlgorithm {
```
joschi: Please run "Reformat Code" once (prevents unnecessary code formatting changes in the future).

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -134,6 +134,12 @@
     @Parameter(value = "dashboard_widget_default_cache_time", validator = PositiveDurationValidator.class)
     private Duration dashboardWidgetDefaultCacheTime = Duration.seconds(10L);
 
+    @Parameter(value = "user_password_default_algorithm")
+    private String userPasswordDefaultAlgorithm = "bcrypt";
+
+    @Parameter(value = "user_password_bcrypt_salt_size")
+    private Integer userPasswordBCryptSaltSize = 10;
```
joschi: This could be a primitive `int` and should be validated with the `PositiveIntegerValidator`.

graylog2-server/src/main/java/org/graylog2/bindings/providers/DefaultPasswordAlgorithmProvider.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.inject.Provider;
+import java.util.Map;
+
+public class DefaultPasswordAlgorithmProvider implements Provider<PasswordAlgorithm> {
+    private final PasswordAlgorithm defaultPasswordAlgorithm;
+
+    @Inject
+    public DefaultPasswordAlgorithmProvider(@Named("user_password_default_algorithm") String defaultPasswordAlgorithmName,
+                                            Map<String, PasswordAlgorithm> passwordAlgorithms) {
+        this.defaultPasswordAlgorithm = passwordAlgorithms.get(defaultPasswordAlgorithmName);
```
joschi: This should throw a `IllegalArgumentException` if the given algorithm couldn't be found. Ideally include the list of valid algorithms (`passwordAlgorithms.keySet()`) in the exception's message.

graylog2-server/src/main/java/org/graylog2/security/hashing/BCryptPasswordAlgorithm.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public String hash(String password) {
+        return hash(password, BCrypt.gensalt(this.saltSize));
+    }
+
+    @Override
+    public boolean matches(String hashedPasswordAndSalt, String otherPassword) {
+        final Splitter splitter = Splitter.on(SALTPREFIX);
```
joschi: Instead of using `Splitter` here, I'd use standard string operations here, for example:

```java
    @Override
    public boolean matches(String hashedPasswordAndSalt, String otherPassword) {
        final int saltIndex = hashedPasswordAndSalt.lastIndexOf(SALTPREFIX);
        checkArgument(saltIndex > 0, "Couldn't find index of salt prefix");

        final String salt = hashedPasswordAndSalt.substring(saltIndex + SALTPREFIX.length());

        return hash(otherPassword, salt).equals(hashedPasswordAndSalt);
    }
```

joschi: This method should also check if `hashedPasswordAndSalt` actually starts with `PREFIX`.

graylog2-server/src/main/java/org/graylog2/security/hashing/BCryptPasswordAlgorithm.java
```diff
@@ -0,0 +1,60 @@
...
+import org.graylog2.plugin.security.PasswordAlgorithm;
+import org.mindrot.jbcrypt.BCrypt;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.List;
+
+public class BCryptPasswordAlgorithm implements PasswordAlgorithm {
+    private static final String PREFIX = "{bcrypt}";
+    private static final String SALTPREFIX = "{salt}";
```
joschi: Minor nitpick: `SALT_PREFIX`.

graylog2-server/src/main/java/org/graylog2/security/hashing/SimpleHashPasswordAlgorithm.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.shiro.crypto.hash.SimpleHash;
+import org.graylog2.Configuration;
+import org.graylog2.plugin.security.PasswordAlgorithm;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.util.regex.Pattern;
+
+public class SimpleHashPasswordAlgorithm implements PasswordAlgorithm {
```
joschi: Could we rename this class to `Sha1PasswordAlgorithm` or `SHA1PasswordAlgorithm`? In the end it's hard coded to SHA-1 and couldn't use any other algorithm.

graylog2-server/src/test/java/org/graylog2/security/PasswordAlgorithmFactoryTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @Mock
+    private PasswordAlgorithm passwordAlgorithm1;
+    @Mock
+    private PasswordAlgorithm passwordAlgorithm2;
+
+    private Map<String, PasswordAlgorithm> passwordAlgorithms;
+
+    @Before
+    public void setUp() throws Exception {
+        this.passwordAlgorithms = ImmutableMap.<String, PasswordAlgorithm>builder()
```
joschi: Since the map is immutable, this could be done once (and maybe even change `passwordAlgorithms` to a constant, i. e. `static final`).

graylog2-server/src/test/java/org/graylog2/security/hashing/BCryptPasswordAlgorithmTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BCryptPasswordAlgorithmTest {
+    private BCryptPasswordAlgorithm bCryptPasswordAlgorithm;
+
+    @Before
+    public void setUp() throws Exception {
+        final Configuration configuration = new Configuration();
+        this.bCryptPasswordAlgorithm = new BCryptPasswordAlgorithm(configuration.getUserPasswordBCryptSaltSize());
```
joschi: I'm not sure that the `Configuration` object is really required here…

graylog2-server/src/test/java/org/graylog2/security/PasswordAlgorithmFactoryTest.java
joschi: :white_check_mark: Ah, no it can't. The `passwordAlgorithm1` and `passwordAlgorithm2` mocks have to be inserted in the `@Before` method.

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
```diff
@@ -198,20 +216,24 @@ public void setHashedPassword(final String hashedPassword) {
...
         }
     }
 
     @Override
-    public boolean isUserPassword(final String password, final String passwordSecret) {
-        final String oldPasswordHash = new SimpleHash(HASH_ALGORITHM, password, passwordSecret).toString();
-        return getHashedPassword().equals(oldPasswordHash);
+    public boolean isUserPassword(final String password) {
+        final PasswordAlgorithm passwordAlgorithm = passwordAlgorithmFactory.forPassword(getHashedPassword());
+        if (passwordAlgorithm == null) {
```
joschi: This could be collapsed to `return passwordAlgorithm != null && passwordAlgorithm.matches(getHashedPassword(), password);`

graylog2-server/src/main/java/org/graylog2/security/hashing/SimpleHashPasswordAlgorithm.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String passwordSecret;
+
+    @Inject
+    public SimpleHashPasswordAlgorithm(@Named("password_secret") String passwordSecret) {
+        this.passwordSecret = passwordSecret;
+    }
+
+    @Override
+    public boolean supports(String hashedPassword) {
+        return !prefixPattern.matcher(hashedPassword).find();
```
joschi: This method should check the length of `hashedPassword` (20 characters for a SHA-1 hash) and that it only consists of `[0-9a-f]` (maybe using `CharMatcher` or a regular expression).

joschi: While I think about it, the regular expression `[0-9a-f]{20}` should pretty much hit it. :wink:

graylog2-server/src/test/java/org/graylog2/security/hashing/BCryptPasswordAlgorithmTest.java
dennisoelkers: I am reusing the current default for the salt size here.

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
dennisoelkers: I prefer the explicit style as it helps to understand what's going on at first glance.

graylog2-server/src/test/java/org/graylog2/security/hashing/BCryptPasswordAlgorithmTest.java
joschi: This will unnecessarily break the test if we change the default hash size. Using `Configuration` here doesn't add anything to the test at all (please correct me, if I overlooked something).

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
joschi: :white_check_mark:

graylog2-server/src/test/java/org/graylog2/security/hashing/BCryptPasswordAlgorithmTest.java
dennisoelkers: No, you are actually 100% right. I overlooked the fact that changing the default salt size would break this test, thanks!

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/security/PasswordAlgorithm.java
dennisoelkers: ✓

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -270,4 +276,12 @@ public Duration getCollectorInactiveThreshold() {
     public Duration getDashboardWidgetDefaultCacheTime() {
         return dashboardWidgetDefaultCacheTime;
     }
+
+    public String getUserPasswordDefaultAlgorithm() {
+        return userPasswordDefaultAlgorithm;
+    }
+
+    public Integer getUserPasswordBCryptSaltSize() {
```
joschi: This getter should also return an `int` to prevent unnecessary boxing.

graylog2-server/src/main/java/org/graylog2/security/hashing/BCryptPasswordAlgorithm.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public String hash(String password) {
+        return hash(password, BCrypt.gensalt(this.saltSize));
+    }
+
+    @Override
+    public boolean matches(String hashedPasswordAndSalt, String otherPassword) {
+        if (supports(hashedPasswordAndSalt)) {
+            final Splitter splitter = Splitter.on(SALT_PREFIX);
```
joschi: Instead of using `Splitter` here, I'd use standard string operations here, for example:

```java
        final int saltIndex = hashedPasswordAndSalt.lastIndexOf(SALTPREFIX);
        checkArgument(saltIndex > 0, "Couldn't find index of salt prefix");

        final String salt = hashedPasswordAndSalt.substring(saltIndex + SALTPREFIX.length());

        return hash(otherPassword, salt).equals(hashedPasswordAndSalt);
```
Additionally, the current implementation will fail with an `IndexOutOfBoundsException` if the `SALT_PREFIX` is missing (which should rather be an `IllegalArgumentException`.

graylog2-server/src/main/java/org/graylog2/security/hashing/BCryptPasswordAlgorithm.java
```diff
@@ -0,0 +1,64 @@
...
+    private final Integer saltSize;
+
+    @Inject
+    public BCryptPasswordAlgorithm(@Named("user_password_bcrypt_salt_size") Integer saltSize) {
+        this.saltSize = saltSize;
+    }
+
+    @Override
+    public boolean supports(String hashedPassword) {
+        return hashedPassword.startsWith(PREFIX);
```
joschi: Maybe check for `SALT_PREFIX` here as well?

graylog2-server/src/main/java/org/graylog2/security/hashing/BCryptPasswordAlgorithm.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public boolean matches(String hashedPasswordAndSalt, String otherPassword) {
+        if (supports(hashedPasswordAndSalt)) {
+            final Splitter splitter = Splitter.on(SALT_PREFIX);
+            final List<String> splitted = splitter.splitToList(hashedPasswordAndSalt);
+            final String salt = splitted.get(1);
+
+            return hash(otherPassword, salt).equals(hashedPasswordAndSalt);
+        } else {
+            throw new IllegalArgumentException("Supplied hashed password is not supported, it does not start with "+ this.PREFIX);
```
joschi: static final variable referenced via `this`.

PR #1452, opened 2015-09-28
Make internal logs accessible via REST API call

lennartkoopmann:
* Log4J appender writing logs to a fixed size in memory destination
* REST API call to access the logs

This allows Apollo to collect the most recent logs and also the graylog-web-interface to display server logs. This will make maintenance easier and our support faster.

### DIFF
--- a/graylog2-bootstrap/src/main/resources/log4j.xml
+++ b/graylog2-bootstrap/src/main/resources/log4j.xml
@@ -10,7 +10,8 @@
         </layout>
     </appender>
 
-    <appender name="memory" class="org.graylog2.MemoryAppender">
+    <!-- Internal Graylog log appender. Please do not disable. This makes internal log messages available via REST calls. -->
+    <appender name="graylog-internal-logs" class="org.graylog2.log4j.MemoryAppender">
         <param name="BufferSize" value="500"/>
     </appender>
 
@@ -52,6 +53,6 @@
     <root>
         <priority value="warn"/>
         <appender-ref ref="console"/>
-        <appender-ref ref="memory" />
+        <appender-ref ref="graylog-internal-logs" />
     </root>
 </log4j:configuration>
--- a/graylog2-bootstrap/src/main/resources/log4j.xml
+++ b/graylog2-bootstrap/src/main/resources/log4j.xml
@@ -10,6 +10,10 @@
         </layout>
     </appender>
 
+    <appender name="memory" class="org.graylog2.MemoryAppender">
+        <param name="BufferSize" value="500"/>
+    </appender>
+
     <!-- Application Loggers -->
     <logger name="org.graylog2">
         <level value="info"/>
@@ -48,5 +52,6 @@
     <root>
         <priority value="warn"/>
         <appender-ref ref="console"/>
+        <appender-ref ref="memory" />
     </root>
 </log4j:configuration>
--- a/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
+++ b/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
@@ -1,0 +1,39 @@
+package org.graylog2.rest.models.system.loggers.responses;
+
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+import java.util.List;
+
+@AutoValue
+@JsonAutoDetect
+public abstract class LogMessage {
+
+    @JsonProperty
+    public abstract String message();
+
+    @JsonProperty("class_name")
+    public abstract String className();
+
+    @JsonProperty
+    public abstract String level();
+
+    @JsonProperty
+    public abstract String timestamp();
+
+    @JsonProperty
+    public abstract List<String> throwable();
+
+    @JsonCreator
+    public static LogMessage create(@JsonProperty("message") String message,
+                                    @JsonProperty("class_name") String className,
+                                    @JsonProperty("level") String level,
+                                    @JsonProperty("timestamp") String timestamp,
+                                    @JsonProperty("throwable") List<String> throwable) {
+        return new AutoValue_LogMessage(message, className, level, timestamp, throwable);
+    }
+
+}
--- a/graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
+++ b/graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
@@ -1,0 +1,87 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.graylog2;
+
+import com.google.common.collect.Lists;
+import edu.emory.mathcs.backport.java.util.Arrays;
+import org.apache.commons.collections.buffer.CircularFifoBuffer;
+import org.apache.commons.collections.iterators.ReverseListIterator;
+import org.apache.log4j.AppenderSkeleton;
+import org.apache.log4j.spi.LoggingEvent;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A Log4J appender that keeps a configurable number of messages in memory. Used to make recent internal log messages
+ * available via the REST API.
+ */
+public class MemoryAppender extends AppenderSkeleton {
+
+    private CircularFifoBuffer buffer;
+    private int bufferSize;
+
+    @Override
+    protected void append(LoggingEvent e) {
+        buffer.add(e);
+    }
+
+    @Override
+    public void close() {
+        buffer.clear();
+    }
+
+    @Override
+    public void activateOptions() {
+        this.buffer = new CircularFifoBuffer(bufferSize);
+    }
+
+    @Override
+    public boolean requiresLayout() {
+        return false;
+    }
+
+    public List<LoggingEvent> getLogMessages(int max) {
+        if (buffer == null) {
+            throw new IllegalStateException("Cannot return log messages: Appender is not initialized.");
+        }
+
+        List<LoggingEvent> messages = Lists.newArrayList();
+        ReverseListIterator iter = new ReverseListIterator(new ArrayList<LoggingEvent>(Arrays.asList(buffer.toArray())));
+        int i = 0;
+        while(iter.hasNext()) {
+            messages.add((LoggingEvent) iter.next());
+            if (i >= max-1) {
+                break;
+            }
+
+            i++;
+        }
+
+        return messages;
+    }
+
+    public int getBufferSize() {
+        return bufferSize;
+    }
+
+    public void setBufferSize(int bufferSize) {
+        this.bufferSize = bufferSize;
+    }
+
+}
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
@@ -26,24 +26,19 @@ import com.wordnik.swagger.annotations.ApiResponse;
 import com.wordnik.swagger.annotations.ApiResponses;
 import org.apache.log4j.Level;
 import org.apache.log4j.Logger;
+import org.apache.log4j.spi.LoggingEvent;
 import org.apache.shiro.authz.annotation.RequiresAuthentication;
-import org.graylog2.rest.models.system.loggers.responses.LoggersSummary;
-import org.graylog2.rest.models.system.loggers.responses.SingleLoggerSummary;
-import org.graylog2.rest.models.system.loggers.responses.SingleSubsystemSummary;
-import org.graylog2.rest.models.system.loggers.responses.SubsystemSummary;
+import org.graylog2.MemoryAppender;
+import org.graylog2.rest.models.system.loggers.responses.*;
 import org.graylog2.shared.rest.resources.RestResource;
 import org.graylog2.shared.security.RestPermissions;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.slf4j.LoggerFactory;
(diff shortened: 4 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.graylog2;
```
lennartkoopmann: I was not sure about the location but this like felt the best.

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+import java.util.List;
+
+@AutoValue
+@JsonAutoDetect
+public abstract class LogMessage {
```
dennisoelkers: "LogMessage" is too generic and too easy to mix up with the "Message" class. Please use something like "InternalLogMessage" at least if you can't come up with a more precise name.

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.google.common.collect.Lists;
+import edu.emory.mathcs.backport.java.util.Arrays;
```
dennisoelkers: Wrong import, please use java.util.Arrays

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -26,24 +26,19 @@
...
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.PUT;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
+import javax.ws.rs.*;
 import javax.ws.rs.core.MediaType;
-import java.util.Enumeration;
-import java.util.Map;
+import java.util.*;
```
dennisoelkers: Please don't use wildcard imports.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +146,43 @@ public void setSingleLoggerLevel(
...
 
+    @GET
+    @Timed
+    @ApiOperation(value = "Get recent log messages")
+    @ApiResponses(value = {
+            @ApiResponse(code = 404, message = "Memory appender not configured.")
+    })
+    @Path("/messages/recent")
+    @Produces(MediaType.APPLICATION_JSON)
+    public LogMessagesSummary messages(@ApiParam(name = "max", required = true) @QueryParam("max") int max) {
```
dennisoelkers: The commonly used parameter for a maximum number of elements to return in our REST resources is "limit", please use that instead of "max".

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +146,43 @@ public void setSingleLoggerLevel(
...
+        List<LogMessage> messages = new ArrayList<>();
+
+        Logger logger = Logger.getRootLogger();
+        if (logger.getAppender("memory") == null) {
+            throw new WebApplicationException(404);
+        }
+
+        MemoryAppender appender = (MemoryAppender) logger.getAppender("memory");
+        for (LoggingEvent loggingEvent : appender.getLogMessages(max)) {
+            List<String> throwable = new ArrayList<>();
```
dennisoelkers: Why are you initializing a new ArrayList here just to overwrite it soon after? If it's for the purpose of using an empty list in case of ``getThrowableStrRep()`` returning null, then doing that in an else-clause and using ``Collections.emptyList()`` would be better.

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
dennisoelkers: It really feels misplaced there. Maybe put it in an own package?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
joschi: Alternatively, we use "size" in different places as well.

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
joschi: `org.graylog.log4j` would work, IMHO. It's not really bound to Graylog in any way, so we could even create a separate artifact for it. :trollface:

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @JsonProperty
+    public abstract String message();
+
+    @JsonProperty("class_name")
+    public abstract String className();
+
+    @JsonProperty
+    public abstract String level();
```
joschi: Using an enum here would be better.

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @JsonProperty
+    public abstract String timestamp();
```
joschi: Why not `DateTime`?

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+        return false;
+    }
+
+    public List<LoggingEvent> getLogMessages(int max) {
+        if (buffer == null) {
+            throw new IllegalStateException("Cannot return log messages: Appender is not initialized.");
+        }
+
+        List<LoggingEvent> messages = Lists.newArrayList();
+        ReverseListIterator iter = new ReverseListIterator(new ArrayList<LoggingEvent>(Arrays.asList(buffer.toArray())));
```
joschi: TooManyCollectionsException…
```java
    public List<LoggingEvent> getLogMessages(int max) {
        if (buffer == null) {
            throw new IllegalStateException("Cannot return log messages: Appender is not initialized.");
        }

        final List<LoggingEvent> result = new ArrayList<>(max);
        final Object[] messages = buffer.toArray();
        for(int i = messages.length - 1; i >= 0 && i >= messages.length - max; i--) {
            result.add((LoggingEvent) messages[i]);
        }

        return result;
    }
```

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+        int i = 0;
+        while(iter.hasNext()) {
+            messages.add((LoggingEvent) iter.next());
```
joschi: This will cause an off-by-1 error if `max == 0`.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +146,43 @@ public void setSingleLoggerLevel(
...
+            @ApiResponse(code = 404, message = "Memory appender not configured.")
+    })
+    @Path("/messages/recent")
+    @Produces(MediaType.APPLICATION_JSON)
+    public LogMessagesSummary messages(@ApiParam(name = "max", required = true) @QueryParam("max") int max) {
+        List<LogMessage> messages = new ArrayList<>();
+
+        Logger logger = Logger.getRootLogger();
+        if (logger.getAppender("memory") == null) {
+            throw new WebApplicationException(404);
```
joschi: There's a `NotFoundException` which automatically sets the correct response status. Additionally, a short explaining message would be nice (e. g. "In-Memory appender not found. Please check the log4j configuration of this Graylog instance.")

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +146,43 @@ public void setSingleLoggerLevel(
...
+        }
+
+        MemoryAppender appender = (MemoryAppender) logger.getAppender("memory");
```
joschi: This should check the actual type of the appender with `instanceof` first and throw an internal server error if it doesn't fit.

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+
+        return messages;
+    }
+
+    public int getBufferSize() {
+        return bufferSize;
+    }
+
+    public void setBufferSize(int bufferSize) {
+        this.bufferSize = bufferSize;
```
joschi: This should check for `bufferSize >= 0`, e. g. with `checkArgument`.

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.log4j.spi.LoggingEvent;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A Log4J appender that keeps a configurable number of messages in memory. Used to make recent internal log messages
+ * available via the REST API.
+ */
+public class MemoryAppender extends AppenderSkeleton {
```
joschi: Maybe rename to `InMemoryAppender`.

graylog2-bootstrap/src/main/resources/log4j.xml
```diff
@@ -10,6 +10,10 @@
         </layout>
     </appender>
 
+    <appender name="memory" class="org.graylog2.MemoryAppender">
```
joschi: The name should be more obviously point to the internal usage of this appender, e. g. "graylog-internal-logs", and have a short comment that this shouldn't be removed or changed.

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
lennartkoopmann: an enum for what? level?

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
lennartkoopmann: What would be the benefit of `InMemoryAppender` vs `MemoryAppender`?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
lennartkoopmann: I think it makes sense to have a wildcard import here given all the `java.util` imports that happen. Actually IntelliJ keeps overwriting it to a wildcard. Namespace cluttering/confusion from `java.util` is not expected IMO.

lennartkoopmann:
Thanks for the review! :) I have addressed/fixed all issues and if not then left a comment at your respective note.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
joschi: We're not using wildcard imports at all in the Graylog codebase (as part of our informal coding standards). See http://stackoverflow.com/a/3348855/49505 for how to disable that in IntelliJ IDEA.

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/loggers/responses/LogMessage.java
joschi: Yes, exactly. Either use [`Level`](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Level.html) or write your own enum. This way you don't have to juggle with that stringly typed API. :wink:

graylog2-server/src/main/java/org/graylog2/MemoryAppender.java
joschi: "Naming things is hard". In my opinion, *In*MemoryAppender describes better what this class is doing. But I'm not too opinionated about this.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +148,49 @@ public void setSingleLoggerLevel(
...
+    @GET
+    @Timed
+    @ApiOperation(value = "Get recent log messages")
+    @ApiResponses(value = {
+            @ApiResponse(code = 404, message = "Memory appender is disabled or of wrong instance.")
+    })
+    @Path("/messages/recent")
+    @Produces(MediaType.APPLICATION_JSON)
+    public LogMessagesSummary messages(@ApiParam(name = "limit", required = true) @QueryParam("limit") int limit) {
+        List<InternalLogMessage> messages = new ArrayList<>();
```
joschi: This list could be initialized with the correct size (`appender.getLogMessages(limit).size()`).

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
lennartkoopmann: Roger!

lennartkoopmann:
Thank you, sir! Just pushed a commit that addresses your proposed changes.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -151,6 +159,48 @@ public void setSingleLoggerLevel(
...
 
+    @GET
+    @Timed
+    @ApiOperation(value = "Get recent log messages")
+    @ApiResponses(value = {
+            @ApiResponse(code = 404, message = "Memory appender is disabled or of wrong instance.")
+    })
+    @Path("/messages/recent")
+    @Produces(MediaType.APPLICATION_JSON)
+    public LogMessagesSummary messages(@ApiParam(name = "limit", required = true, defaultValue = "500") @QueryParam("limit") int limit) {
```
joschi: Unfortunately that's not enough, as `@ApiParam` is just an annotation used by Swagger to generate the documentation/Swagger specification. `@DefaultParam` is the correct annotation. :trollface:

PR #1464, opened 2015-10-05
Forbidden API checker

joschi:
This PR adds the [Forbidden API checker](https://github.com/policeman-tools/forbidden-apis) to the "pedantic" build profile.

It's currently running in "soft-mode" so that builds don't actually fail but the violations are printed on stdout.

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/journal/RawMessage.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/journal/RawMessage.java
@@ -32,6 +32,7 @@ import org.graylog2.plugin.Tools;
 import org.graylog2.plugin.configuration.Configuration;
 import org.graylog2.plugin.system.NodeId;
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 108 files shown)


lennartkoopmann:
:+1:

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/journal/RawMessage.java
```diff
@@ -166,7 +167,7 @@ public int getVersion() {
     }
 
     public DateTime getTimestamp() {
-        return new DateTime(msgBuilder.getTimestamp()); // TODO PERFORMANCE object creation
+        return new DateTime(msgBuilder.getTimestamp(), DateTimeZone.UTC); // TODO PERFORMANCE object creation
```
lennartkoopmann: Can this cause problems on the reading side of the journal? Just changing it from the local TZ to UTC might have consequences.

lennartkoopmann:
If we make sure that switching the `DateTime` constructor to always use `DateTimeZone.UTC` has no negative side effects this looks like a great addition to me.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/journal/RawMessage.java
joschi: As far as I can see this is fine. This method is only used in `RawMessage#toString()` and in some codecs to retrieve the message timestamp. Since we pass around a complete `DateTime` object (which explicitly contains a timestamp), this won't change anything (famous last words).

bernd:
LGTM :+1:

PR #1466, opened 2015-10-06
Add regex replace extractor

joschi:
This PR adds an extractor which supports replacing a message field with a constant value or the matches of a regular expression.

Similar to the already existing Regex extractor, this extractor tries to match a regular expression against a message field (see [`Pattern`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)) and replaces its value (or the value of the target field) with the string configured in the "replacement" setting (see [`Matcher#replaceAll()`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#replaceAll-java.lang.String-)). The extractor also allows to only replace the first match or all matches (default).

### DIFF
--- a/graylog2-bootstrap/pom.xml
+++ b/graylog2-bootstrap/pom.xml
@@ -29,7 +29,7 @@
     <parent>
         <groupId>org.graylog2</groupId>
         <artifactId>graylog2-parent</artifactId>
-        <version>1.2.2-SNAPSHOT</version>
+        <version>1.3.0-SNAPSHOT</version>
     </parent>
 
     <name>graylog2-bootstrap</name>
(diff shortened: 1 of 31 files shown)


kroepke:
Shouldn't this go in to master, as it is a new feature?

joschi:
@kroepke Since we don't have a fixed roadmap for the next version of Graylog and we will most likely release a patch version for Graylog 1.2.x anyway, I thought it might as well go into the next Graylog 1.2.x version.

This being said, the patch set would also apply to `master`.

kroepke:
Ok, let's discuss during the next days :)

dennisoelkers:
I am confused. This seems to be the internal log message REST API PR, but it's not named differently and contains different stuff. What happened here?

joschi:
Closed due to being based on the wrong branch (`1.2` instead of `1.3`).

PR #1476, opened 2015-10-10
This fixes #1475

### LINKED ISSUE #1475: JSON Extractor fails to convert to fields when there are fields with null value
Working with the latest 1.2.1.

Even though when testing with "Try" I can see all the fields, even the ones with null value, the extractor doesn't convert the string to fields.

I believe the fields with null value should be extracted with string "null" instead of null value

gvagenas:
Example and description of the problem https://github.com/Graylog2/graylog2-server/issues/1475


joschi:
As mentioned in #1475, this would change the semantics of the JSON extractor in an undesired way. We'll find another solution to the problem.

PR #1478, opened 2015-10-11
Report selected configuration variables via REST API

lennartkoopmann:
This pull request adds a REST resource that is reporting a selected list of configuration variables. This is important for Apollo but can on the longer term also be integrated into the web interface.

The `ConfigurationVariable` class might look a bit convoluted but I tried to make sure that only numbers, strings and booleans can be added to avoid weird `toString` behavior if somebody is not careful when adding a new configuration variable that is reported.

Please take a close look at the JSON structure in the response. I struggled with that one and could not really find one that satisfied me.

### DIFF
--- a/graylog2-bootstrap/src/main/resources/log4j.xml
+++ b/graylog2-bootstrap/src/main/resources/log4j.xml
@@ -10,6 +10,11 @@
         </layout>
     </appender>
 
+    <!-- Internal Graylog log appender. Please do not disable. This makes internal log messages available via REST calls. -->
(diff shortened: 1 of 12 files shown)


joschi:
@lennartkoopmann Please rebase this change set on the `1.3` branch and create a new PR for it.

Too bad GitHub doesn't support changing the target branch of a pull request. :disappointed:

lennartkoopmann:
Got it! Will do. Thanks.

PR #1479, opened 2015-10-12
Switching language level to Java 8.

dennisoelkers:
This small changeset switches the compiler's source and target level to Java 8 for maven. You might need to switch your project's language level and SDK to Java 8 in IntelliJ as well.


joschi:
LGTM. :+1:

PR #1482, opened 2015-10-12
Remove radio and references to it.

dennisoelkers:
As radio is deprecated, this PR removes all of it from master for 2.0


joschi:
We should probably also remove the radio-related inputs in this PR.

lennartkoopmann:
bring it back

PR #1484, opened 2015-10-13
Report selected configuration variables via REST API

lennartkoopmann:
(Adopted to be ready to be merged into the 1.3 branch as requested in #1478)

This pull request adds a REST resource that is reporting a selected list of configuration variables. This is important for Apollo but can on the longer term also be integrated into the web interface.

The `ConfigurationVariable` class might look a bit convoluted but I tried to make sure that only numbers, strings and booleans can be added to avoid weird `toString` behavior if somebody is not careful when adding a new configuration variable that is reported.

Please take a close look at the JSON structure in the response. I struggled with that one and could not really find one that satisfied me.


graylog2-server/src/main/java/org/graylog2/rest/resources/system/ConfigurationResource.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /*
+     * This call is manually building a map of variables to return because we need to guarantee never to return any
+     * sensitive variables like passwords etc. - See this as a whitelist approach.
+     */
+    @GET
+    @Timed
+    @ApiOperation(value = "Get all reported configuration variables and their values")
+    public ConfigurationList getAll() {
```
joschi: This looks awfully manual and error-prone if we introduce or change configuration settings.

JadConfig supports [dumping the provided configuration as `Map<String, String>`](https://joschi.github.io/JadConfig/0.11.0/apidocs/com/github/joschi/jadconfig/JadConfig.html#dump--), so maybe we should use this instead of listing the configuration settings manually.

lennartkoopmann: IMO it is more error prone to dump the whole configuration and blacklisting parts of it because somebody may add a sensitive parameter that can easily be forgotten to be excluded here. Also user defined/plugin beans would be always reported no matter what is in there. This is why I made it so explicit.

The goal is not to dump all configuration but selected values that are useful.

dennisoelkers: What about adding jackson annotations (JsonProperty/JsonIgnore, whichever way around we prefer) to the Configuration class and return it in the REST resource instead of using these convoluted value objects and the very verbose way of creating them?

lennartkoopmann: That could work if it is a whitelist approach but I'd use an own annotation like `@ApiReported` or something to keep it cleaner.

I don't know if having to scan the class for annotations is less convoluted though.

bernd:
Any progress on this? 1.3 is due soon! :smiley:

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/configuration/ConfigurationVariable.java
```diff
@@ -0,0 +1,50 @@
...
+@AutoValue
+public abstract class ConfigurationVariable {
+
+    @JsonProperty
+    public abstract String name();
+
+    @JsonProperty
+    public abstract Object value();
+
+    @JsonCreator
```
joschi: The `@JsonCreator` annotation cannot be used on multiple factory methods or constructors. I've added a failing unit test to demonstrate this and to add some test coverage…

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ConfigurationResource.java
joschi: The method and the description of the HTTP resource should at least not say that it dumped *all* configuration variables. Maybe "selected" or "relevant" would be better?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ConfigurationResource.java
```diff
@@ -0,0 +1,97 @@
...
+        List<ConfigurationVariable> config = new ArrayList<>();
```
joschi: Minor nitpick: I'd use `ImmutableList.Builder<ConfigurationVariable>` which at least supports a chained builder interface.

lennartkoopmann: Done! :)

graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/configuration/ConfigurationVariable.java
lennartkoopmann: Done in 4337fb5

graylog2-server/src/main/java/org/graylog2/configuration/ExposedConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+    private final int streamProcessingMaxFaults;
+    private final long outputModuleTimeout;
+    private final int staleMasterTimeout;
+
+    private final boolean disableIndexOptimization;
+    private final int indexOptimizationMaxSegments;
+
+    private final String gcWarningThreshold;
+
+    public ExposedConfiguration(Configuration configuration, ElasticsearchConfiguration esConfiguration) {
```
joschi: What do we need `ConfigurationVariable` and `ConfigurationList` for if we could simply expose this class as JSON?

lennartkoopmann: I thought that we might want to dump this into other places in the future, too so I kept it not bound to JSON.

joschi: I'd rather get rid of this (currently) unnecessary layer of indirection.

lennartkoopmann: `org.graylog2.rest` cannot access the classes in `org.graylog2.configuration` so the `ExposedConfiguration` class is required to map configuration values to a `List`. We'd have to do this in the JAX-RS logic and that smells.

Given how close the code freeze is and how much this small change is improving our support process I'd ask you to merge it the way it is if you don't have strong opinions against it except a preference for not using this class and moving the logic.

You can of course also perform the change yourself and then merge if you see an easier way that makes sense.

Thank you! :)

PR #1485, opened 2015-10-13
Add regex replace extractor

joschi:
This PR adds an extractor which supports replacing a message field with a constant value or the matches of a regular expression.

Similar to the already existing Regex extractor, this extractor tries to match a regular expression against a message field (see [`Pattern`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)) and replaces its value (or the value of the target field) with the string configured in the "replacement" setting (see [`Matcher#replaceAll()`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#replaceAll-java.lang.String-)). The extractor also allows to only replace the first match or all matches (default).

Supersedes #1466.


edmundoa:
LGTM :+1:

edmundoa:
Also cherry-picked it to master.

PR #1488, opened 2015-10-15
Fix memory leak in GelfChunkAggregator

### LINKED ISSUE #1462: Possible memory leak in GelfChunkAggregator
TL;DR: GelfChunkAggregator.ChunkEntry has a compareTo override that can be misleading to sortedEvictionSet (ConcurrentSkipListSet). In high volume, it can cause a memory leak. See patch at the bottom as a suggestion to fix this.

---

Using 1.2.1 with GELF UDP, we were seeing memory consumption (graylog-server) goes up until the process spends more time doing GC than actual work. Investigating a heap dump showed that `..inputs.codecs.GelfChunkAggregator` class was accumulating `ChunkEntry`s.

I did some digging in `GelfChunkAggregator`, trying to find out why this happens. I noticed that at some point, `sortedEvictionSet` and `chunks` are going out of sync, so the `ChunkEvictionTask` is getting stuck on the `sortedEvictionSet.first()` entry and always tries to evict it.

What I saw was this:

1. ChunkEvictionTask called.
1. First entry in sortedEvictionSet is selected and the ID is passed to expireEntry.
1. expireEntry calls getAndCleanupEntry with that ID.
1. getAndCleanupEntry calls chunks.remove(ID), which returns null.
1. sortedEvictionSet.remove(null) throws a NPE, leaving the sortedEvictionSet as it is.
1. ChunkEvictionTask catches/ignores the exception, and the process begins all over again.

Investigating further revealed that the ChunkEvictionTask, for the given ID, does succeed at `chunks.remove(ID)` on the first time, returning an entry, but then `sortedEvictionSet.remove(entry)` fails (just returns false because no element was removed). This was interesting, since the eviction initiates from the first element of the sortedEvictionSet, so how is it possible that it does not find it?

Some info about our usage. Our application servers are pushing UDP packets at the rate of ~7K/s per graylog-server. Now, at this rate, some entries are bound to have the same timestamp (in millis). These are sorted into the sortedEvictionSet (by the ChunkEntry.compareTo).

What I found was that, sometimes, if two different entries share the same timestamp, the ConcurrentSkipListSet implementation will miss it when trying to call remove on it. This is probably because of how the skip list is being looked into, narrowing down the list of elements, causing it to miss the element that should be found.

For example, if the eviction set looks like this (id:timestamp): `[ a:1 | b:1 | .. | c:5 | d:7 ]`, trying to lookup a:1 will "lead" the algorithm to the range between b:1 and c:5. Essentially making the call `sortedEvictionSet.remove(sortedEvictionSet.first())` fail, when it shouldn't. This is reproducible with a small data set as well, although not consistent.

I wasn't sure how to address that without changing or breaking other stuff, so I added another level of sorting to compareTo. If the timestamps are equal, it'll compare the message IDs. After this addition, the problem went away.

```diff
diff --git a/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java b/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
index 085e4ff..0c9a29d 100644
--- a/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
+++ b/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
@@ -238,6 +238,9 @@ public class GelfChunkAggregator implements CodecAggregator {
             if (equals(o)) {
                 return 0;
             }
+            if (firstTimestamp == o.firstTimestamp) {
+                return id.compareTo(o.id);
+            }
             return firstTimestamp < o.firstTimestamp ? -1 : 1;
         }
     }
```

bernd:
The memory leak showed up because of a problem in the ChunkEntry#compareTo() method.

If two ChunkEntry objects had the same firstTimestamp value, the compareTo() method returned a wrong result because only the timestamp has been taken into account. This made sortedEvictionSet.remove(entry) not remove the entry from the set in some cases.

The GelfChunkAggregator class maintains a chunks ConcurrentMap and a sortedEvictionSet ConcurrentSkipListSet with ChunkEntry objects. The ChunkEvictionTask gets the first entry from the sortedEvictionSet and basically removes it from the chunks map and then tries to remove it from the sortedEvictionSet as well.

The removal from the chunks map works, the removal from the sortedEvictionSet does not work. (due to the compareTo() bug)

During the next scheduled execution of ChunkEvictionTask#run(), the same entry is returned from the sortedEvictionSet. This entry is not in the chunks map anymore so chunks.remove(entry) returns null. The return value is used to remove the entry from the sortedEvictionSet. This throws a NullPointerException because the return value was null. This NPE was ignored and not logged.

This change fixes the ChunkEntry#compareTo() method to compare the ID if both timestamps are equal. We also now log any error that happens during a ChunkEvictionTask#run() execution.

This issue was found, debugged and reported in great detail with a patch to fix it by @lightpriest and @onyxmaster. Thank you!

See https://github.com/Graylog2/graylog2-server/issues/1462 for details.

Fixes #1462

### DIFF
--- a/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
+++ b/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
@@ -198,13 +198,14 @@ public class GelfChunkAggregator implements CodecAggregator {
         return sb.toString();
     }
 
-    private static class ChunkEntry implements Comparable<ChunkEntry> {
+    @VisibleForTesting
+    static class ChunkEntry implements Comparable<ChunkEntry> {
         private final AtomicInteger chunkSlotsWritten = new AtomicInteger(0);
         private final long firstTimestamp;
         private final AtomicReferenceArray<GELFMessageChunk> payloadArray;
         private final String id;
 
-        private ChunkEntry(int chunkCount, long firstTimestamp, String id) {
+        public ChunkEntry(int chunkCount, long firstTimestamp, String id) {
             this.payloadArray = new AtomicReferenceArray<>(chunkCount);
             this.firstTimestamp = firstTimestamp;
             this.id = id;
@@ -238,6 +239,12 @@ public class GelfChunkAggregator implements CodecAggregator {
             if (equals(o)) {
                 return 0;
             }
+            // If two chunk entries have the same timestamp, we have to compare the IDs. Otherwise the removal from
+            // the eviction set might not work and leak memory.
+            // See: https://github.com/Graylog2/graylog2-server/issues/1462
+            if (firstTimestamp == o.firstTimestamp) {
+                return id.compareTo(o.id);
+            }
             return firstTimestamp < o.firstTimestamp ? -1 : 1;
         }
     }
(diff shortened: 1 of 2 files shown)


kroepke:
lgtm!

PR #1490, opened 2015-10-18
Use a better default for ring sizes

lennartkoopmann:
I just realized that our ring sizes are still having a default size that comes from the pre-journal age when they acted as buffers in case of backing up messages. When messages queue up, the rings will fill up before the journal starts backing up. This can lead to the objects in there being promoted down the memory pool chain and really funky GC behavior with stop the worlds that make the existing problems (why the journal is backing up) even worse.

Also all that memory is pre-allocated at the start and just eats up tons of heap for nothing. I have not yet seen a setup processing so many messages to reasonably use so many slots in the rings.

4096 seems to be a reasonable default for me. 

Do our packages use this config file or would we have to update those, too? (cc @mariussturm @bernd)

### DIFF
--- a/misc/graylog2.conf
+++ b/misc/graylog2.conf
@@ -242,9 +242,9 @@ processor_wait_strategy = blocking
 # For optimum performance your LogMessage objects in the ring buffer should fit in your CPU L3 cache.
 # Start server with --statistics flag to see buffer utilization.
 # Must be a power of 2. (512, 1024, 2048, ...)
-ring_size = 65536
+ring_size = 4096
 
-inputbuffer_ring_size = 65536
+inputbuffer_ring_size = 4096
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 


bernd:
We actually increased the ring sizes after the journal has been introduced. (See: f43bc8a2c49b07d8921269daf05d14cf0e0ba808) The idea was to get better throughput into/from the journal by having bigger batches to write/read into/from the journal.

> When messages queue up, the rings will fill up before the journal starts backing up.

Sorry, I don't understand what you want to say here.

Regarding pre-allocation, we only allocate the holder objects for the RawMessage and Message objects in the disruptor. As soon as the RawMessage/Message objects are not needed anymore, they are set to null in the holder objects to avoid promotion to the next generation.

Did you run into actual problems because of the current default ring sizes? Our operating system packages are shipped with a 1GB heap limit by default. We are not getting much reports about this being a problem in smaller setups.

Regading your question about the packages: We have to adjust the OS packages, appliances and config management modules if we change this. (Ansible/Puppet/Chef)

kroepke:
The ring sizes should not impact promotion to old gen, unless the elasticsearch connection is completely down, in which case the ring buffer content might be promoted in any case.
With 4k sizes the throughput starts to suffer on more powerful systems, usually by reading too few messages from the journal in one batch.
Even promoting 65k average sized messages will usually not take more than 200MB of heap, including object overhead.

Unless we have clear evidence that the ring sizes are inappropriate, I'd vote against changing the defaults.

lennartkoopmann:
Right, the oldgen promotion was a misconception on my side. The objects in the ring buffer are never actually replaced when writing new messages into it so they should end up in oldgen pretty fast no matter if messages are queuing up or not.

Let me quickly investigate something else. I'll close this if I don't find anything new. Thanks!

lennartkoopmann:
In a setup with all standard parameters once all buffers are filled we end up with 235 of 512MB max OldGen space usage.

This looks indeed fine. I got on the wrong path by not thinking about the ring containers pre-allocation correctly. Thanks for correcting me! :)

lennartkoopmann:
The test setup also fully recovered very fast. It had to put it some major GCs but struggled at no moment in time.

PR #1492, opened 2015-10-19
Handle user email values with multiple addresses

### LINKED ISSUE #1439: Error sending messages with multiple LDAP mail values
# Situation
I have *inetOrgPerson* LDAP users mapped into Graylog.

# Behaviour
When a user has multiple values within the *mail* field, messages are not sent and the following error message is shown:

```
The Graylog server encountered an error while trying to send an email. This is the detailed error message: org.apache.commons.mail.EmailException: javax.mail.internet.AddressException: Illegal address in string ``tillo@tillo.ch, martino@dellambrogio.ch'' (javax.mail.internet.AddressException: Illegal address in string ``tillo@tillo.ch, martino@dellambrogio.ch'')
```

# Expected behaviour
I guess it should use an array instead of a comma to set multiple recipients, seems to be compatible with both org.apache.commons.mail and javax.mail.internet. I expect to have the message delivered to both recipients.

# Debug information
I'm running  latest **graylog-server** **1.2.0-5** on latest **Debian stable** **8.2** with latest **Oracle JDK** **1.7.0_80**.

# Additional comment
As org.apache.commons.mail is bundled into graylog.jar I don't know what version it is.
It really shouldn't be bundled, neither should 99% of the code in there, but I understand it's difficult to keep up with various distributions dependencies.

bernd:
LDAP users can have multiple addresses in the email field. Split the email field on the "," character and add all addresses as recipient.

Convert recipients to a Set to avoid sending duplicate emails.

Fixes #1439

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/StaticEmailAlertSender.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/StaticEmailAlertSender.java
@@ -16,8 +16,9 @@
  */
 package org.graylog2.alerts;
 
+import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
-import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
 import org.apache.commons.mail.DefaultAuthenticator;
 import org.apache.commons.mail.Email;
 import org.apache.commons.mail.EmailConstants;
@@ -44,6 +45,7 @@ import org.slf4j.LoggerFactory;
 import javax.inject.Inject;
 import java.net.URI;
 import java.util.List;
+import java.util.Set;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/StaticEmailAlertSender.java
```diff
@@ -222,7 +224,11 @@ public void sendEmails(Stream stream, AlertCondition.CheckResult checkResult, Li
                 final User user = userService.load(username);
 
                 if(user != null && user.getEmail() != null && !user.getEmail().isEmpty()) {
-                    recipients.add(user.getEmail());
+                    // LDAP users might have multiple email addresses defined.
+                    // See: https://github.com/Graylog2/graylog2-server/issues/1439
+                    final List<String> addresses = Splitter.on(",").omitEmptyStrings().trimResults().splitToList(user.getEmail());
```
joschi: Using [`Splitter#split()`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Splitter.html#split(java.lang.CharSequence)) here would safe some intermediate object instances.

```java
recipients = new HashSet<>(Splitter.on(",").omitEmptyStrings().trimResults().split(user.getEmail()));
```

joschi:
LGTM. :+1:

Issue #1433, opened 2015-09-20
Allow forcing "group object class" / "memberAttribute" for LDAP auth


RustyDust:
I'm trying to run the LDAP authentication against a Univention Corporate Server. While this basically works I'm unable to successfully use the LDAP group to role mapping. Looking at the logs I see a couple lines saying:

`2015-09-20T12:55:13.523+02:00 WARN  [LdapConnector] Unable to auto-detect the LDAP group object class, assuming 'member' is the correct attribute.`

The reason for this is that UCS doesn't use the `objectClass "groupOfUniqueNames"` as expected by the LDAP connector (/org/graylog2/security/ldap/LdapConnector.java:236) and while the LDAP entry actually has the `uniqueMember` attribute Graylog makes the wrong choice here. Here's what I get when querying the LDAP directory for a group (shortened for brevity):

```
uniqueMember: cn=user-01,cn=users,dc=example,dc=muc
uniqueMember: cn=user-02,cn=users,dc= example,dc=com
uniqueMember: cn=user-03,cn=users,dc= example,dc=com
uniqueMember: cn=user-04,cn=users,dc= example,dc=com
uniqueMember: cn=team-01,cn=groups,dc= example,dc=com
uniqueMember: cn=team-02,cn=groups,dc= example,dc=com
uniqueMember: cn=team-02,cn=groups,dc= example,dc=com
(...)
objectClass: top
objectClass: posixGroup
objectClass: univentionGroup
objectClass: sambaGroupMapping
objectClass: univentionObject
```

Now if I could tell Graylog what objectClass filter to use (univentionGroup) or what the memberAttribute is (uniqueMember) the problem would be easily solved I guess. An alternative would be to add another check for "univentionGroup" to the LdapConnector.java. But I think in the long run a user definable option would make more sense than possibly adding tons of different and maybe conflicting tests there.

andham:
Having memberAttribute configurable would also solve #1436.

olivierdes:
I'm running into the same issue as my ldap groups are configured like this : 
dn: cn=Domain Users,ou=groups,dc=example,dc=com
objectClass: top
objectClass: posixGroup
objectClass: sambaGroupMapping
gidNumber: 513
cn: Domain Users
(....)
memberUid: user1
memberUid: user2
....
So it would be nice if there was way to specify the memberUid attribute :)

bernd:
[...]

PR #1494, opened 2015-10-19
Improve compatibility with different LDAP schemas

bernd:
Improve the auto detection of the member attribute to support more LDAP setups out there. The member attribute should be configurable, but this will be done as a feature addition in the future because it's too much code for the stable release.

Also support `posixGroup` groups by checking for the `memberUid` attribute. The `memberUid` references a UID instead of a DN so we have to compare it against the UID of the LDAP entry.

Refs #1433

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/security/ldap/LdapConnectorTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/security/ldap/LdapConnectorTest.java
@@ -21,7 +21,7 @@ import org.apache.directory.ldap.client.api.LdapConnectionConfig;
 import org.apache.directory.ldap.client.api.LdapNetworkConnection;
 import org.apache.directory.server.annotations.CreateLdapServer;
 import org.apache.directory.server.annotations.CreateTransport;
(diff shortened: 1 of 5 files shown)


bernd:
I am currently working on some LDAP tests.

graylog2-server/src/test/java/org/graylog2/security/ldap/LdapConnectorTest.java
```diff
@@ -0,0 +1,256 @@
...
+                                @CreateIndex(attribute = "ou")
+                        }
+
+                )
+        },
+        loadedSchemas = {
+                @LoadSchema(name = "nis", enabled = true)
+        }
+)
+@ApplyLdifs(
```
joschi: Could we replace this with `@ApplyLdifFiles` so that the LDIF could also be easily applied to another LDAP server?

bernd: Done.

joschi:
LGTM. :+1:

PR #1498, opened 2015-10-21
Remove dead letters feature

joschi:
This PR removes the rarely used dead letters feature from Graylog.


bernd:
LGTM :+1:

runningman84:
please delete the old data during the first start of this version...

joschi:
@runningman84 This feature was disabled by default and I'm sure that users will cry out if we simply remove the collection and its contents from MongoDB if they have been using that feature. We will add a note to the release notes of Graylog 2.x about this, though.

dannykopping:
Given that this feature was removed, is there any way to view the full message that could not be indexed?

dannykopping:
@joschi any advice for fixing the indexing issues then? Is there a log that the indexing errors will go to or...?

joschi:
@dannykopping Indexing failures will be logged by Graylog and written into the `index_failures` collection in MongoDB.

dannykopping:
Thanks @joschi

dannykopping:
image
i guess not... it doesn't show the full message. Unless I'm being a moron. I don't know Mongo at all

joschi:
@dannykopping No, as I wrote before, the indexing failures (i. e. a description of the failure) will be written into the `index_failures` collection in MongoDB.

joschi:
@dannykopping I'm sorry, you were completely correct the whole time. Storing of indexing failures in MongoDB (`index_failures`) was accidentally removed in this PR.

It will be added back (see #2633).

Sorry

dannykopping:
Lol, thanks @joschi 👍

bdentino:
I'm a little confused by this. Correct me if I'm wrong, but it sounds like @dannykopping was asking if it's possible to view the actual message that failed to index (not just a description of the failure). It seems that this is still not possible, with 2.1, nor is it intended. Is that correct? If so, then to the original question, is there any way to recover the actual messages that failed to index or are they lost for good?

PR #1500, opened 2015-10-21
Fix SessionsResource / Add method to return stream rule types

dennisoelkers:
This is a small collection of changes to support the current state of the react-only web interface. It includes:

* Add CORS flags required by browsers for cross domain requests to server's REST interface
* Remove code that results in returning 404 after successfull session removal
* Add resource method to return available stream rule types

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
@@ -72,7 +72,6 @@ public enum StreamRuleType {
         throw new IllegalArgumentException("Invalid Stream Rule Type specified: " + name);
     }
 
-    @JsonValue
     public Map<String, Object> toMap() {
         final Map<String, Object> result = ImmutableMap.<String, Object>of(
                 "id", value,
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
@@ -24,11 +24,6 @@ package org.graylog2.plugin.streams;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.common.collect.ImmutableMap;
-
-import java.util.HashMap;
-import java.util.Map;
 
 public enum StreamRuleType {
     EXACT(1, "match exactly", "match exactly"),
@@ -51,7 +46,7 @@ public enum StreamRuleType {
         return value;
     }
 
-    public static StreamRuleType fromInteger(@JsonProperty("value") final int numeric) {
+    public static StreamRuleType fromInteger(final int numeric) {
         for (final StreamRuleType streamRuleType : StreamRuleType.values()) {
             if (streamRuleType.value == numeric) {
                 return streamRuleType;
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
@@ -22,24 +22,36 @@
  */
 package org.graylog2.plugin.streams;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.HashMap;
+import java.util.Map;
+
 public enum StreamRuleType {
-    EXACT(1),
-    GREATER(3),
-    SMALLER(4),
-    REGEX(2),
-    PRESENCE(5);
+    EXACT(1, "match exactly", "match exactly"),
+    REGEX(2, "match regular expression", "match regular expression"),
+    GREATER(3, "greater than", "be greater than"),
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/streams/rules/StreamRuleResource.java
```diff
@@ -192,4 +195,19 @@ public void delete(@ApiParam(name = "streamid", value = "The stream id this new
             throw new NotFoundException();
         }
     }
+
+    @GET
+    @Path("/types")
+    @Timed
+    @ApiOperation(value = "Get all available stream types")
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<Map<String, Object>> types(@ApiParam(name = "streamid", value = "The stream id this new rule belongs to.", required = true)
```
joschi: Please don't use `Map<String, Object>` as return type for JAX-RS resource methods. This doesn't play well with Swagger and isn't type-safe in any way.

dennisoelkers: Agreed. Was a bit hackish. :)

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
```diff
@@ -22,24 +22,36 @@
...
+        this.shortDesc = shortDesc;
+        this.longDesc = longDesc;
     }
 
     public int toInteger() {
         return value;
     }
 
-    public static StreamRuleType fromInteger(final int numeric) {
+    public static StreamRuleType fromInteger(@JsonProperty("value") final int numeric) {
```
joschi: What purpose does the `@JsonProperty` annotation serve here? It would only be relevant in a `@JsonCreator` method.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/responses/StreamRuleTypeResponse.java
```diff
@@ -0,0 +1,46 @@
...
+@AutoValue
+@JsonAutoDetect
+public abstract class StreamRuleTypeResponse {
+    @JsonProperty
+    public abstract int id();
+
+    @JsonProperty
+    public abstract String name();
+
+    @JsonProperty("short_desc")
```
joschi: What's the difference between the "short" and the "long" description anyway? At least in `StreamRuleType` it's identical.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/rules/StreamRuleResource.java
```diff
@@ -192,4 +194,19 @@ public void delete(@ApiParam(name = "streamid", value = "The stream id this new
             throw new NotFoundException();
         }
     }
+
+    @GET
+    @Path("/types")
+    @Timed
+    @ApiOperation(value = "Get all available stream types")
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<StreamRuleTypeResponse> types(@ApiParam(name = "streamid", value = "The stream id this new rule belongs to.", required = true)
```
joschi: This method should probably check whether the user is actually allowed to access the stream.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -126,10 +127,5 @@ public SessionResponse newSession(@Context ContainerRequestContext requestContex
     public void terminateSession(@ApiParam(name = "sessionId", required = true) @PathParam("sessionId") String sessionId) {
         final Subject subject = getSubject();
         securityManager.logout(subject);
-
-        final org.apache.shiro.session.Session session = subject.getSession(false);
```
joschi: Why was this check removed? If at all, I'd move it before the call to `securityManager.logout(subject)` (because right now it effectively removes any active session the subject might have.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/streams/StreamRuleType.java
```diff
@@ -22,24 +22,36 @@
  */
 package org.graylog2.plugin.streams;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.HashMap;
```
joschi: Unused import.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/rules/StreamRuleResource.java
joschi: Looking at the signature and the implementation of this method, I'm not sure that this JAX-RS resource is the best place for it.

dennisoelkers: I had the same thought. Unfortunately there is not really a better alternative right now.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
dennisoelkers: At this place the check does not make sense and returned a 404 every time, that's why it was removed.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/responses/StreamRuleTypeResponse.java
dennisoelkers: I am keeping the naming from the web interface. Refactoring this is a future step.

PR #1507, opened 2015-10-26
Add missing role assignment when updating user

### LINKED ISSUE #1456: Roles field not updated after PUT on REST API
Creating users, including setting roles works. 
Updating users also works, but setting roles doesn't seem to have any effect. The online REST docs do show "roles" as a field for PUT on /users/{username}.

There are no errors in the logs. When setting other fields besides roles, the other fields get updated successfully.

bernd:
Fixes #1456


joschi:
LGTM. :+1:

PR #1510, opened 2015-10-27
Add Timezone resource returning list of available timezones

dennisoelkers:
This is supporting the web interface which renders this list for user creation/editing.


dennisoelkers:
Not required anymore.

PR #1515, opened 2015-10-28
added shell script to help manage ES

bmanojlovic:
Few script i use in my daily work with graylog and elasticsearch that i find users could find usable


bernd:
Thank you for the contribution! :smiley:  It would be great if you could submit this to the [Graylog Marketplace](https://marketplace.graylog.org/). I think it would greatly fit into "Other Solutions". Having it in there lets other users find it more easily.

Issue #1495, opened 2015-10-21
Add default grok patterns


kroepke:
Many users are surprised to see their grok patterns not working which for first time users often is because they haven't added any patterns in the first place.

Take a well known set of patterns and populate the database with them by default.

bernd:
Closed via #1552

PR #1521, opened 2015-10-29
Support loading and applying content packs on startup

joschi:
This PR extends the original `ContentPackLoaderPeriodical` to be able to load and apply content packs from a configurable directory on startup.

The content pack loader scans `content_packs_dir` (default: `data/contentpacks`) for JSON files (`*.json`) and loads those into Graylog (existing content packs with the same category and name will not be overwritten). After that, all content packs from files listed in the `content_packs_auto_load` configuration setting (comma-separated list of file names) will be applied, i. e. the described entities (inputs, dashboards, streams, outputs) will be created in Graylog.

The whole mechanism can be enabled/disabled using the `content_packs_loader_enabled` configuration setting.

As a proof of concept (and to solve refs #1495) it includes a content pack which contains a core set of grok patterns (taken from https://github.com/logstash-plugins/logstash-patterns-core/tree/master/patterns/grok-patterns).


bernd:
The OS packages and the appliances need to incorporate the new config values and directories.

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -134,6 +140,15 @@
...
     private Duration dashboardWidgetDefaultCacheTime = Duration.seconds(10L);
 
+    @Parameter(value = "content_packs_loader_enabled")
+    private boolean contentPacksLoaderEnabled = true;
+
+    @Parameter(value = "content_packs_dir", validators = DirectoryPathReadableValidator.class)
+    private Path contentPacksDir = Paths.get("data", "contentpacks");
+
+    @Parameter(value = "content_packs_auto_load", converter = TrimmedStringListConverter.class)
+    private List<String> contentPacksAutoLoad = Collections.emptyList();
```
bernd: I think it might be better to make this a `Set<String>` to avoid potential problems with duplicates.

graylog2-server/src/main/java/org/graylog2/periodical/ContentPackLoaderPeriodical.java
```diff
@@ -89,37 +112,112 @@ protected Logger getLogger() {
...
             try {
-                contentPack = objectMapper.readValue(resourceBytes, ConfigurationBundle.class);
+                contentPack = objectMapper.readValue(bytes, ConfigurationBundle.class);
             } catch (IOException e) {
-                LOG.warn("Couldn't parse content pack " + resource + ". Skipping", e);
+                LOG.warn("Couldn't parse content pack in file " + file + ". Skipping", e);
+                continue;
+            }
+
+            final ConfigurationBundle existingContentPack = contentPackExists(contentPack);
```
bernd: The `contentPackExists()` method is a bit weirdly named. When I read that it sounds like it returns a boolean. Maybe something like `lookupContentPack()` reads a bit better?

graylog2-server/src/main/java/org/graylog2/periodical/ContentPackLoaderPeriodical.java
```diff
@@ -89,37 +112,112 @@ protected Logger getLogger() {
...
+        if (!contentPackLoaderConfig.equals(changedContentPackLoaderConfig)) {
+            clusterConfigService.write(changedContentPackLoaderConfig);
+        }
+    }
+
+    private ConfigurationBundle contentPackExists(final ConfigurationBundle contentPack) {
+        final ConfigurationBundle bundle = bundleService.findByNameAndCategory(contentPack.getName(), contentPack.getCategory());
+
+        if (bundle != null) {
+            LOG.debug("Content pack {}/{} already exists in database. Skipping.", contentPack.getCategory(), contentPack.getName());
```
bernd: I would move the log statement out of this method. This method just does a lookup for the given content pack. It does not decide anything.

Maybe we can just get rid of the method?

graylog2-server/src/main/java/org/graylog2/periodical/ContentPackLoaderPeriodical.java
joschi: As a matter of fact the method returned a boolean before but was refactored to return the actual content pack. :wink:

bernd:
If the user updates an existing content pack in the content pack directory it will be skipped because it already exists.

Do we have a way to update to the latest content pack?

joschi:
> If the user updates an existing content pack in the content pack directory it will be skipped because it already exists.
> 
> Do we have a way to update to the latest content pack?

We could store a checksum of the files but the question is how complicated we actually want to have this functionality. There are also some edge cases regarding applying changed content packs (e. g. duplicate/conflicting inputs).

bernd:
> We could store a checksum of the files but the question is how complicated we actually want to have this functionality. There are also some edge cases regarding applying changed content packs (e. g. duplicate/conflicting inputs).

I agree, it should not become super complicated. We should just think about how this works and document it. I bet this question will come up. :smiley:

bernd:
LGTM :+1:

PR #1526, opened 2015-11-02
Create separate shaded artifact of graylog2-server

joschi:
This PR changes the POM of graylog2-server so that the Maven Shade plugin will create a separate shaded artifact instead of replacing the original (unshaded) artifact.

This enables plugin authors to depend on the unshaded version of the graylog2-server artifact during development (1.3 MiB vs 73 MiB) and thus save some bandwidth (chances are good that other transitive dependencies are already in their local Maven caches).


bernd:
LGTM :+1:

PR #1528, opened 2015-11-03
Add rest-client parts for usage stats opt-out

bernd:
Depends on Graylog2/graylog-plugin-anonymous-usage-statistics#10.

### DIFF
--- a/graylog2-rest-client/src/main/java/org/graylog2/restclient/models/UsageStatsService.java
+++ b/graylog2-rest-client/src/main/java/org/graylog2/restclient/models/UsageStatsService.java
@@ -28,7 +28,8 @@ import java.io.IOException;
 
 public class UsageStatsService {
     private static final Logger LOG = LoggerFactory.getLogger(UsageStatsService.class);
-    private static final String PATH = "/plugins/org.graylog.plugins.usagestatistics/opt-out";
+    private static final String CONFIG_PATH = "/plugins/org.graylog.plugins.usagestatistics/config";
+    private static final String OPT_OUT_PATH = "/plugins/org.graylog.plugins.usagestatistics/opt-out";
 
     private final ApiClient api;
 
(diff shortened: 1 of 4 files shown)


graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/usagestats/UsageStatsOptOutState.java
```diff
@@ -0,0 +1,34 @@
...
+package org.graylog2.rest.models.system.usagestats;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class UsageStatsOptOutState {
```
joschi: Since this class is only used in the REST client and not shared with graylog2-server, it should be moved into graylog2-rest-client.

bernd:
Updated after comment.

graylog2-rest-client/src/main/java/org/graylog2/restclient/models/UsageStatsService.java
```diff
@@ -0,0 +1,66 @@
...
+import org.graylog2.restclient.lib.ApiClient;
+import org.graylog2.restroutes.PathMethod;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import java.io.IOException;
+
+public class UsageStatsService {
```
joschi: Maybe it would be useful to add a check if the Anonymous Usage Statistics plugin is installed at all.

bernd: I will do that in the web interface by using the config endpoint. (Graylog2/graylog-plugin-anonymous-usage-statistics#11)

joschi: Unfortunately that particular resource won't exist, if the plugin is not installed. :wink:

joschi:
LGTM. :+1:

graylog2-rest-client/src/main/java/org/graylog2/restclient/models/UsageStatsService.java
bernd: Yes, that's why I handle 404s.

PR #1455, opened 2015-09-29
Add resources to show/recalculate index range for single index

joschi:
Closes #1451

Issue #1451, opened 2015-09-28
/system/indices/ranges/{indice_id}/rebuild


pdepaepe:
Hello,

Current POST method /system/indices/ranges/rebuild ask to recheck all indices (in a such random way). This can take a while and is not necessary useful when we know which ones have to be check:

2015-09-28T20:43:14.721Z WARN  [EsIndexRangeService] Couldn't load index range for index graylog2_486
2015-09-28T20:43:14.722Z WARN  [EsIndexRangeService] Couldn't load index range for index graylog2_519

So may you please add a POST /system/indices/ranges/{indice_id}/rebuild, or patch /system/indices/ranges/rebuild to start by the indices already identified as "without index range" ?

Thanks.

joschi:
Closed by merging #1455.

PR #1532, opened 2015-11-04
Recalculate single index range

joschi:
This PR adds support for recalculating a single index range to graylog2-rest-client.

Refs #1451, refs #1455


edmundoa:
LGTM :+1:

PR #1535, opened 2015-11-04
Add periodical for index ranges cleanup

### LINKED ISSUE #1511: Index ranges can get out of sync when manually deleting ES indices
When the Graylog server is not running and an index gets deleted from Elasticsearch, the index ranges get out of sync. Even manually recalculating index ranges does not remove the stale entries from MongoDB.

This can break searches when Graylog thinks an index exists but in reality it does not. As of 1.2.2, there is no way to fix this besides manually deleting the stale documents from MongoDB.

joschi:
This PR adds a periodical which checks for stale index range information in the database and removes all index range information for non-existing indices in a configurable interval (`index_ranges_cleanup_interval`).

Fixes #1511

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
@@ -84,7 +84,7 @@ public class IndexRangesCleanupPeriodical extends Periodical {
         }
 
         if (!removedIndices.isEmpty()) {
-            LOG.info("Removing index range information for non-existing indices: {}", removedIndices);
+            LOG.info("Removing index range information for unavailable indices: {}", removedIndices);
             eventBus.post(IndicesDeletedEvent.create(removedIndices));
         }
     }
--- a/graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
@@ -1,0 +1,131 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


joschi:
This should probably be merged into `1.2` too.

graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
```diff
@@ -0,0 +1,131 @@
...
+    private final Deflector deflector;
+    private final IndexRangeService indexRangeService;
+    private final EventBus eventBus;
+    private final int periodSeconds;
+
+    @Inject
+    public IndexRangesCleanupPeriodical(final Cluster cluster,
+                                        final Deflector deflector,
+                                        final IndexRangeService indexRangeService,
+                                        final EventBus eventBus,
```
bernd: Shouldn't this be annotated with `@ClusterEventBus`?

joschi: Only the Graylog master node manages the storage backend (removes indices etc.), so emitting this event into the cluster is unnecessary.

bernd: Ah, I see now. The MongoIndexRangeService subscribes to both event buses. Thanks for the clarification.

graylog2-server/src/main/java/org/graylog2/periodical/IndexRangesCleanupPeriodical.java
```diff
@@ -0,0 +1,131 @@
...
+
+        final List<String> removedIndices = new ArrayList<>();
+        for (IndexRange indexRange : indexRanges) {
+            if (!indexNames.contains(indexRange.indexName())) {
+                removedIndices.add(indexRange.indexName());
+            }
+        }
+
+        if (!removedIndices.isEmpty()) {
+            LOG.info("Removing index range information for non-existing indices: {}", removedIndices);
```
bernd: Should we change the wording a bit here? This also gets logged for closed indices.

```
2015-11-11 17:17:43,701 INFO : org.graylog2.periodical.IndexRangesCleanupPeriodical - Removing index range information for non-existing indices: [graylog2_2]
```

kroepke:
lgtm

PR #1537, opened 2015-11-05
additional statistics for alarm callbacks and alert counts

kroepke:
- uses mongodb aggregation for group by, should be available starting with mongodb v2.2 but we need to verify

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
@@ -114,9 +114,4 @@ public class AlarmCallbackConfigurationServiceMJImpl implements AlarmCallbackCon
             throw new IllegalArgumentException("Supplied output must be of implementation type AlarmCallbackConfigurationAVImpl, not " + callback.getClass());
         }
     }
-
-    private static class CountAggrType {
-        public String _id;
-        public long total;
-    }
 }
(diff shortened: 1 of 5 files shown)


joschi:
> should be available starting with mongodb v2.2 but we need to verify

We currently mention MongoDB 2.0 or higher in our [system requirements](http://docs.graylog.org/en/1.2/pages/installation/manual_setup.html#prerequisites) because of Debian Wheezy, which only contains [MongoDB 2.0.6](https://packages.debian.org/wheezy/mongodb). If we decide to raise this requirement, we need to clearly communicate this in the release notes.

graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackConfigurationServiceMJImpl.java
```diff
@@ -94,4 +112,9 @@ private AlarmCallbackConfigurationAVImpl implOrFail(AlarmCallbackConfiguration c
             throw new IllegalArgumentException("Supplied output must be of implementation type AlarmCallbackConfigurationAVImpl, not " + callback.getClass());
         }
     }
+
+    private static class CountAggrType {
```
joschi: Please create a proper class (e. g. with `@AutoValue`) with `hashCode()` and `equals()`, especially since instances of this class are stored in a `HashMap`.

joschi: Ah, wait. They aren't…

kroepke: :)

joschi:
LGTM. :+1:

PR #1539, opened 2015-11-05
Allow fetching index statistics for all indices in bulk

joschi:
Instead of retrieving index statistics only for a single index (and even that in a very inefficient way) the Graylog REST API now provides a resource at `/system/indexer/indices` which returns the index statistics of all Elasticsearch indices managed by Graylog.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -16,7 +16,9 @@
  */
 package org.graylog2.indexer.indices;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -375,25 +385,56 @@ protected Boolean checkForReopened(IndexMetaData metaData) {
         return reopenedIndices;
     }
 
+    @Nullable
     public IndexStatistics getIndexStats(String index) {
-        final IndexStatistics stats = new IndexStatistics();
+        if (index.startsWith(configuration.getIndexPrefix())) {
```
bernd: Shouldn't this be `if (! index.startsWith...`?

joschi: Argh, yes of course.

bernd:
LGTM :+1:

PR #1540, opened 2015-11-05
Remove unnecessary complete index range recalculation

joschi:
Closing or deleting an index via the Graylog REST API doesn't require a recalculation of index range information for all indices any more.


bernd:
LGTM :+1:

PR #1542, opened 2015-11-09
Remove DashboardRegistry

dennisoelkers:
The DashboardRegistry came from a time, when there was no
DashboardService and therefore it was a good idea to have a collection
keeping available dashboards. This is useless now and also creates an
implicit cache which produces synchronization issues when mutations are
called on non-master nodes. Therefore DashboardResource operations
mutating the existing state where restricted to the master node.

This is neither useful, desirable nor necessary, so this PR is removing
the dashboard registry and its usages.


joschi:
LGTM. :+1:

PR #1543, opened 2015-11-09
Prevent conversion of `null` to "null" in ConfigurationMapConverter

joschi:
`ConfigurationMapConverter` used to convert `null` values in strings into a string with content "null" instead of an empty string which caused unwanted effects when saving a configuration with optional values (that could as well be `null` or empty).

Fixes Graylog2/graylog2-web-interface#1677.


bernd:
The code throws exceptions now. Might this break existing setups out there?

joschi:
> The code throws exceptions now. Might this break existing setups out there?

The code threw exceptions before as well: https://github.com/Graylog2/graylog2-server/blob/1.2.2/graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java#L45

This being said, we can change the `IllegalArgumentException` into a `ValidationException `. This will affect setups with totally broken configurations for alarm callbacks or outputs, e. g. using non-existing field types but I think it's preferable to show people that their configuration is broken instead of quietly swallowing the error and potentially return an invalid configuration map.

bernd:
LGTM :+1:

PR #1548, opened 2015-11-09
Add installation_source configuration setting and log it on startup

### LINKED ISSUE #1530: Graylog should log which packaging was used to install it
We want to be able to more easily see from log files which packaging option (rpm, deb, OVA, tarball, docker etc) was used to install Graylog to assist in common tuning and config issues.

Proposal is that the packages and other pre-made deployment options write an extra file next to the config file which will contain the string identifying the method used to install Graylog.

joschi:
The `installation_source` configuration setting (default: "unknown") can be overridden by an environment variable (`GRAYLOG2_INSTALLATION_SOURCE`) or a Java system property (`graylog2.installation_source`).

Example output on startup:
```
2015-11-09 17:24:50,503 INFO : org.graylog2.bootstrap.ServerBootstrap - Graylog server 1.3.0-SNAPSHOT (4489a5e) starting up (JRE: Oracle Corporation 1.8.0_66 on Mac OS X 10.10.5, installed from: unknown)
```
The `installation_source` setting could also be collected and transmitted by the Anonymous Usage Statistics plugin to get a better overview about our installation base and which type of distribution channel users have chosen.

Closes #1530

### DIFF
--- a/graylog2-bootstrap/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
+++ b/graylog2-bootstrap/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
@@ -74,7 +74,8 @@ public abstract class ServerBootstrap extends CmdLineTool {
 
     @Override
     protected void startCommand() {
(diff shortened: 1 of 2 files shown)


bernd:
LGTM :+1:

PR #1554, opened 2015-11-11
Use the root user's timezone for LDAP users by default

### LINKED ISSUE #1000: Set default timezone on new LDAP users
New LDAP users haven't their timezone set, using UTC as default. It would be better to use something more meaningful as a default value, probably the admin user's timezone.

bernd:
Ideally we would use the timezone configured in the web interface, but we do not have access to that from the server.

Fixes #1000


joschi:
LGTM. :+1:

PR #1562, opened 2015-11-13
Do not allow overwriting the Message timestamp with a non Date object

### LINKED ISSUE #1556: Extractors can overwrite timestamp field with wrong type
A user had huge issues because a extractor wrote a String into the timestamp field. From the mapping:

    "timestamp" : {
        "type" : "string"
    }

It should be:

    "timestamp" : {
        "type" : "date",
         "doc_values" : true,
         "format" : "yyyy-MM-dd HH:mm:ss.SSS"
    }

This causes different problems, including a now completely failing index time range calculation. (See #1517 for a similar case.)

While the best way is for sure to get rid of the `RESERVED_SETTABLE_FIELDS`/`RESERVED_FIELDS` in combation with the `Map<String, Object>` in the `Message` class I am voting for a quick and good-enough(tm) fix that can go into 1.3.0. This is an issue that can easily happen and leave a common user in a helpless situation.

Something like this might do the trick until we have an overall improvement in future versions:

    if(key.equals("timestamp") and !(value instanceof DateTime)) {
        LOG.warn("Field 'timestamp' must be of type DateTime. Make sure to apply a date converter when using extractors. Not overwriting field.");
    }

bernd:
This is a sign of bad extractors/converters.

Fixes #1556

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
@@ -256,6 +256,12 @@ public class Message {
             return;
         }
 
+        if (FIELD_TIMESTAMP.equals(key) && !(value instanceof DateTime) && !(value instanceof Date)) {
+            LOG.warn("Field [{}] must be of value Date or DateTime. Make sure to apply a date converter when using extractors. Not overwriting field.",
+                    FIELD_TIMESTAMP);
+            return;
+        }
+
         if(value instanceof String) {
             final String str = ((String) value).trim();
 
--- a/graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/MessageTest.java
+++ b/graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/MessageTest.java
@@ -22,6 +22,7 @@
  */
 package org.graylog2.plugin;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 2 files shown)


graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -256,6 +256,12 @@ public void addField(final String key, final Object value) {
             return;
         }
 
+        if (FIELD_TIMESTAMP.equals(key) && !(value instanceof DateTime) && !(value instanceof Date)) {
+            LOG.warn("Field \"{}\" must be of value DateTime. Make sure to apply a date converter when using extractors. Not overwriting field.",
```
lennartkoopmann: One very small thing: I'd always enclose field names etc in square brackets. This way a field name cannot mess up or confuse the meaning of the message.

lennartkoopmann:
This is the way of solving it I thought about. The performance hit of `instanceof` should be insignificant but we might want to run it through a profiler to make sure.

bernd:
I updated the log message.

lennartkoopmann:
Thank you! For a change in a core part like this I'd like to get a second opinion from for example @kroepke or @joschi or anybody else who is still in the daily dev business. :)

bernd:
As @joschi pointed out, the index mapping Graylog installs should prevent storing messages with a bad timestamp in Elasticsearch. So this PR does not really solve the issue. There seems to be something else that is broken when bad timestamps can enter Elasticsearch.

joschi:
FWIW, I think this change is a good safeguard against accidentally overwriting the message timestamp with an incompatible type. The additional check with `instanceof` is cheap enough and doesn't add too much overhead given all the other checks that are being run in the `Message#addField()` method.

LGTM. :+1:

lennartkoopmann:
Thanks guys! :)

Issue #1502, opened 2015-10-22
[put-mapping] can timeout


pdepaepe:
Hello,

On a big ES cluster, while rotating indice, put-mapping operation can timeout and block following index range calculation.

```
2015-10-21T15:08:48.622Z ERROR [IndexRotationThread] Couldn't point deflector to a new index
org.elasticsearch.cluster.metadata.ProcessClusterEventTimeoutException: failed to process cluster event (put-mapping [index_range]) within 30s
	at org.elasticsearch.cluster.service.InternalClusterService$2$1.run(InternalClusterService.java:278)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
2015-10-21T15:08:52.310Z WARN  [IndexRotationThread] Deflector is pointing to [graylog2_993], not the newest one: [graylog2_994]. Re-pointing.
2015-10-21T15:09:33.380Z WARN  [EsIndexRangeService] Couldn't load index range for index graylog2_994
2015-10-21T15:09:33.381Z WARN  [EsIndexRangeService] Couldn't load index range for index graylog2_993
```

I know, index range is rollbacked to mongo, but instead of set field mapping dynamically, why not inject a "graylog2 template" on graylog startup and so avoid completely this operation?

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html

bernd:
In #1563 we changed the index creation operation to include the mapping instead of having it in two separate operations.

Creating index templates on Graylog boot is something that we might consider doing later. Marking this as feature.

Thank you for the report!

bernd:
Graylog is creating a Elasticsearch index template during boot since 2.0.

PR #1563, opened 2015-11-16
Merge ES index and mapping creation in Indices#create()

joschi:
When creating an Elasticsearch index and putting the respective mapping in separate operations, the mapping creation might fail (e. g. due to a timeout) and lead to invalid messages being indexed into Elasticsearch (e. g. with a wrong type for "timestamp" or other important message fields). Merging these operations should at least solve some of those error situations.

Refs #1502


bernd:
LGTM :+1:

Issue #1561, opened 2015-11-13
Probable concurrency bug in GelfChunkAggregator.checkForCompletion ()


uttg:
Hello,

I investigated on #1544 a little bit further.

For a message with 6 chunks I have the following behaviour:
* *reception of the first chunks...*
* reception of chunk number 3
  * checkForCompletion detects that we have every chunks
  * it fails to construct a full message. Chunks 4, 5 and 6 are not yet in entry.payloadArray
* reception of chunk number 6
  * checkForCompletion detects that we have every chunks
  * it fails to construct a full message. Chunks 1, 2 and 3 are not yet in entry.payloadArray

I noticed that the test [if (chunkWatermark == sequenceCount)](https://github.com/Graylog2/graylog2-server/blob/99d3e248090547114cb8d5b0ad4f07cf9abbde27/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java#L144) may return true multiple times for the same message.

Somehow, [incrementAndGet()](https://github.com/Graylog2/graylog2-server/blob/99d3e248090547114cb8d5b0ad4f07cf9abbde27/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java#L141) returns the same value multiple times event though the operation should be atomic.

In graylog.conf we have:
```
processor_wait_strategy = blocking
inputbuffer_processors = 2
inputbuffer_wait_strategy = blocking
```

tomtomgod:
This is probably not the cause, but the method equals() of ChunkEntry will always return false.
Line 255 tries to compare two AtomicInteger:
```c#
 if (!chunkSlotsWritten.equals(that.chunkSlotsWritten)) return false;
```
This comparison is always false. But I don't see how that would affect the rest of the code.

tomtomgod:
I found the problem: it's in the GELF Client we are using :confounded:
Gelf4Net's UDP appender generates consecutive messages with the same messageIds when logging at high frequency.

I don't know what you want to do with my comments on GelfChunkAggregator. Maybe add some defensive code when receiving the same chunk number for the same messageId. Something like:

```c#
if (!entry.payloadArray.compareAndSet(chunk.getSequenceNumber(), null, chunk)) {
    log.error("Duplicated messageId {}", chunk.getId());
    return null;
}
```

kroepke:
[...]

PR #1566, opened 2015-11-16
Fix GelfChunkAggregator.ChunkEntry#equals() and #hashCode()

joschi:
This PR fixes the implementations of `GelfChunkAggregator.ChunkEntry#equals()` and `GelfChunkAggregator.ChunkEntry#hashCode()`.

The previous implementations didn't take into account that `AtomicInteger` and `AtomicReferenceArray<T>` do not override `Object#equals()` and `Object#hashCode()` and thus are never equal and the hash code is never the same.

While the new implementations are slower (citation needed…), they are at least *correct*.

Refs #1561, specifically https://github.com/Graylog2/graylog2-server/issues/1561#issuecomment-156433533

### DIFF
--- a/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
+++ b/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
@@ -19,6 +19,7 @@ package org.graylog2.inputs.codecs;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.Maps;
 import org.graylog2.inputs.codecs.gelf.GELFMessage;
 import org.graylog2.inputs.codecs.gelf.GELFMessageChunk;
@@ -30,9 +31,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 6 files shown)


graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java
```diff
@@ -221,19 +224,25 @@ public boolean equals(Object o) {
 
             final ChunkEntry that = (ChunkEntry) o;
 
+            if (!id.equals(that.id)) return false;
             if (firstTimestamp != that.firstTimestamp) return false;
-            if (!chunkSlotsWritten.equals(that.chunkSlotsWritten)) return false;
-            //noinspection RedundantIfStatement
-            if (!payloadArray.equals(that.payloadArray)) return false;
+            if (chunkSlotsWritten.get() != that.chunkSlotsWritten.get()) return false;
+            if (!equalsAtomicReferenceArray(this.payloadArray, that.payloadArray)) return false;
```
kroepke: I'm actually not sure if we should include those fields in the comparison at all.

Looking at the usage of the sorted eviction set, the lookup sequence is: timestamp, messageid. The other fields aren't actually ever used for the purpose of ordering the `ChunkEntry` instances in the set.

The atomic fields are merely there to avoid a mutex around both field mutations.

Can we safely get rid of these two comparisons here and their calculation of `hashCode` and document the fact why they aren't included?

kroepke: Specifically https://github.com/Graylog2/graylog2-server/blob/a0f396e330f45f1d78376796329eaad9a93cbc04/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java#L134-L138 and https://github.com/Graylog2/graylog2-server/blob/a0f396e330f45f1d78376796329eaad9a93cbc04/graylog2-inputs/src/main/java/org/graylog2/inputs/codecs/GelfChunkAggregator.java#L182 show the access pattern for eviction and do not care about the atomic fields at all.

joschi: :white_check_mark: Agreed on removing the `Atomic*` instance variables from `equals()` and `hashCode()`.

PR #1569, opened 2015-11-17
Fix Html serialization

edmundoa:
The JSON serialization of the formatted `full_message` field is not working properly, returning an empty object (`{}`). As we render those fields in Javascript, we can safely return the string representation of that
Html.

I don't feel really strong about the change, so please let me know if you know a better way of fixing the serialization of `play.twirl.api.Html` objects. 

Refs Graylog2/graylog2-web-interface#1686


joschi:
LGTM. :+1:

But I have to cringe when I look at this class. :cry:

edmundoa:
So do I, hopefully will be gone soon™

PR #1590, opened 2015-11-26
Proxied cluster metrics

dennisoelkers:
Due to the web interface now talking to a single node only (in contrast to the play web interface component connecting to every node in the cluster) we need a mechanism to retrieve metrics (and other resources) from other nodes than the one we are talking to. Therefore a ClusterMetricsResource was created that proxies requests to other nodes for metrics retrieval.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
@@ -1,0 +1,108 @@
+package org.graylog2.rest.resources.cluster;
+
+import com.codahale.metrics.annotation.Timed;
+import com.wordnik.swagger.annotations.Api;
+import com.wordnik.swagger.annotations.ApiOperation;
+import com.wordnik.swagger.annotations.ApiParam;
+import com.wordnik.swagger.annotations.ApiResponse;
+import com.wordnik.swagger.annotations.ApiResponses;
+import org.apache.shiro.authz.UnauthorizedException;
+import org.apache.shiro.authz.annotation.RequiresAuthentication;
+import org.apache.shiro.authz.annotation.RequiresPermissions;
+import org.graylog2.cluster.Node;
+import org.graylog2.cluster.NodeNotFoundException;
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+
+@RequiresAuthentication
+@Api(value = "Cluster/Metrics", description = "Cluster-wide Internal Graylog2 metrics")
+@Path("/cluster/{nodeId}/metrics")
+public class ClusterMetricsResource extends RestResource {
+    private final RemoteMetricsResource remoteMetricsResource;
+
+    @Inject
+    public ClusterMetricsResource(@PathParam("nodeId") String nodeId,
```
edmundoa: There is no way of setting the `nodeId` from the swagger UI. It looks like swagger only generates an input when the parameter is in the resource actions.
The parameter is also missing the `@ApiParam` annotation.

pom.xml
```diff
@@ -690,6 +690,16 @@
                 <artifactId>javax.el-api</artifactId>
                 <version>3.0.0</version>
             </dependency>
+            <dependency>
+                <groupId>com.squareup.retrofit</groupId>
+                <artifactId>retrofit</artifactId>
+                <version>2.0.0-beta2</version>
```
edmundoa: Do we need to use a beta version of retrofit? Probably there is a good reason for it, but I just want to check :smile:

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
dennisoelkers: :+1: - will change.

pom.xml
dennisoelkers: The syntax has changed very much between 1.x and 2.0 and from what I've seen the 2.0 release is being prepared and supposed to be released soon, hopefully before our 2.0 code freeze. :)

edmundoa:
Looking good :+1:

PR #1591, opened 2015-11-26
Getting started dismiss state

kroepke:
Support REST APIs for determining whether or not to show the getting started guide for this server version.

### DIFF
--- a/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/DisplayGettingStarted.java
+++ b/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/DisplayGettingStarted.java
@@ -1,0 +1,17 @@
+package org.graylog2.rest.models.system;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class DisplayGettingStarted {
+
+    @JsonProperty("show")
+    public abstract boolean show();
+
+    @JsonCreator
+    public static DisplayGettingStarted create(@JsonProperty("show") boolean show) {
+        return new AutoValue_DisplayGettingStarted(show);
+    }
+}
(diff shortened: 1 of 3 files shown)


graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/DisplayGettingStarted.java
```diff
@@ -0,0 +1,33 @@
...
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.rest.models.system;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class DisplayGettingStarted {
```
joschi: Missing `@JsonAutoDetect` annotation (which is only used so that our Swagger generator knows which classes to scan).

graylog2-server/src/main/java/org/graylog2/gettingstarted/GettingStartedState.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * A map of Version.getMajor + Version.getMinor -> true/false.
+     * <br/>
+     * A true value means the automatic display of the getting started page has been turned of for that minor version (we don't show it again for patch versions).
+     * Missing values are treated as false.
+     *
+     * @return dismissal state of getting started pages across all stored versions
+     */
+    @JsonProperty
+    public abstract Map<String, Boolean> dismissedInVersions();
```
joschi: Why not simply use a `Set<String>` here?

kroepke: Good point!

joschi:
LGTM. :+1:

joschi:
Ported into `master` in commit c99c327bd2b56e0e23e42782a67cae6b6606c233.

PR #1595, opened 2015-12-02
improve CORS performance

kroepke:
- answer OPTIONS requests early
 - include Access-Control-Max-Age header to allow caching the CORS preflight in browser that support this

this greatly reduces the number of requests necessary to request data for the UI

### DIFF
--- a/graylog2-shared/src/main/java/org/graylog2/shared/rest/CORSFilter.java
+++ b/graylog2-shared/src/main/java/org/graylog2/shared/rest/CORSFilter.java
@@ -17,22 +17,46 @@
 package org.graylog2.shared.rest;
 
 import javax.ws.rs.container.ContainerRequestContext;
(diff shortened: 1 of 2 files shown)


graylog2-shared/src/main/java/org/graylog2/shared/rest/CORSFilter.java
```diff
@@ -35,4 +40,23 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
             responseContext.getHeaders().add("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS");
```
dennisoelkers: Does it make sense to include the `Access-Control-Max-Age`-header in each response so that caching time for suitable requests is enlengthened?

kroepke: good point!

PR #1596, opened 2015-12-02
Integrated web interface

dennisoelkers:
These are prerequisite works to be able to serve the web interface from the server. This includes:

 * Adding another HTTP endpoint which is serving static files, returning the index.html boilerplate code per default instead of a 404 (to make the react-router work)
 * Generate `config.js` on the fly using the server configuration
 * Generate `index.html` on the fly which loads plugin files and web interface bootstrapping code


runningman84:
I am afraid that existing problems like https://github.com/Graylog2/graylog2-web-interface/issues/1694 are goging to crash the whole server. :/

joschi:
@runningman84 Those kind of request are already processed by the server and haven't been cached by the web interface, so there's no change for this kind of workload.

runningman84:
@joschi the problem is that the web interface does not respond anymore if users export too many result sets. This means we restart the web interface daily and if the interface does not respond anymore. If we have to restart the "full server" everytime we have problems with the web interface we would even loose messages.

joschi:
@runningman84 The new web interface is an SPA, so only the user's browser would hang. Load on the server is the same as before.

runningman84:
What means SPA here?

joschi:
@runningman84 Single-page Application: https://en.wikipedia.org/wiki/Single-page_application

graylog2-server/pom.xml
```diff
@@ -43,6 +43,7 @@
 
     <properties>
         <mainClass>org.graylog2.bootstrap.Main</mainClass>
+        <webInterface.path>../graylog2-web-interface</webInterface.path>
```
kroepke: Ideally I'd like to see the entire web-interface related code to live in a separate module, not being tied into the server build process.
But I guess we can refactor this afterwards in the interest of limiting the branch lifetime.

graylog2-server/pom.xml
```diff
@@ -359,8 +360,76 @@
...
+                <directory>${webInterface.path}/build</directory>
+                <targetPath>web-interface/assets</targetPath>
+            </resource>
+            <resource>
+                <directory>src/main/resources</directory>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
+                <groupId>com.github.eirslett</groupId>
```
kroepke: Question: This is only used for the production/package builds, right? Development would still use npm directly?

graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/BaseConfigurationTest.java
```diff
@@ -44,6 +44,9 @@
         @Parameter(value = "rest_listen_uri", required = true)
         private URI restListenUri = URI.create("http://127.0.0.1:12900/");
 
+        @Parameter(value = "web_listen_uri", required = true)
+        private URI webListenUri = URI.create("http://127.0.0.1:9000/");
```
kroepke: Follow up task: I would like to proxy the REST API on port 9000 as well, so that deployments which have restricted Graylog traffic to port 9000 continue to work. This is out of scope for this branch, but for reference we should be able to use https://jersey.java.net/documentation/latest/filters-and-interceptors.html Section 10.2.1.1. Pre-matching and post-matching filters

graylog2-server/src/main/java/org/graylog2/initializers/WebInterfaceService.java
```diff
@@ -0,0 +1,121 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.initializers;
+
+import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.glassfish.jersey.server.ContainerFactory;
+import org.glassfish.jersey.server.ResourceConfig;
+import org.graylog2.jersey.container.netty.NettyContainer;
```
kroepke: This currently clashes with https://github.com/Graylog2/graylog2-server/pull/1613, which one do we first?

graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public class IndexHtmlGenerator {
+    private final Engine engine = new Engine();
+    private final String content;
+
+    @Inject
+    public IndexHtmlGenerator(final String title, final Collection<String> cssFiles, final Collection<String> jsFiles) throws IOException {
+        final String template = IOUtils.toString(ClassLoader.getSystemResourceAsStream("web-interface/index.html.template"));
```
kroepke: Can we get the equivalent Guava `Resources` methods to achieve this (to eventually get rid of the commons-io dependency) and use a class loader which loaded this class instead of the system classloader? The latter could be problematic once we move this into a plugin.

graylog2-server/src/main/java/org/graylog2/initializers/WebInterfaceService.java
dennisoelkers: #1613 is not ready for review yet (according to its labels), so this one should go first. @joschi knew upfront about the conflicts, so I guess he considered them already when building it.

graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java
kroepke: Also should we throw an explicit exception here to have a better error message in case something goes really wrong? An exception would be part of the injector errors which tend to be messy for default exceptions.

graylog2-server/src/main/java/org/graylog2/initializers/WebInterfaceService.java
kroepke: Ok, makes sense!

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@Path("/config.js")
+public class AppConfigResource {
+    private static final Engine engine = new Engine();
+    private final String content;
+
+    @Inject
+    public AppConfigResource(Configuration configuration) throws IOException {
```
kroepke: same as graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java regarding IOException

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -0,0 +1,53 @@
...
+        final String template = IOUtils.toString(ClassLoader.getSystemResourceAsStream("web-interface/config.js.template"));
```
kroepke: same as https://github.com/Graylog2/graylog2-server/pull/1596/files#r46925061

graylog2-server/src/main/java/org/graylog2/web/resources/WebInterfaceAssetsResource.java
```diff
@@ -0,0 +1,125 @@
...
+    private final ObjectMapper objectMapper;
+
+    @Inject
+    public WebInterfaceAssetsResource(ObjectMapper objectMapper,
+                                      Set<Plugin> plugins) throws IOException {
+        this.objectMapper = objectMapper;
+        final List<String> jsFiles = new ArrayList<>();
+        final List<String> cssFiles = new ArrayList<>();
+
+        plugins.stream().forEach(plugin -> {
```
kroepke: This is being done on every request because the jersey resource should be constructed on every request. Can we move loading files from plugin's resources into a common plugin class somehow?

graylog2-server/src/main/java/org/graylog2/web/resources/WebInterfaceAssetsResource.java
```diff
@@ -0,0 +1,125 @@
...
+            final ModuleManifest pluginManifest = manifestForPlugin(plugin);
+            final String pathPrefix = pluginPathPrefix + plugin.metadata().getUniqueId() + "/";
+            if (pluginManifest != null) {
+                jsFiles.addAll(pluginManifest.files().jsFiles().stream().map(file -> file.startsWith(pathPrefix) ? file : pathPrefix + file).collect(Collectors.toList()));
+                cssFiles.addAll(pluginManifest.files().cssFiles().stream().map(file -> file.startsWith(pathPrefix) ? file : pathPrefix + file).collect(Collectors.toList()));
+            }
+        });
+        final InputStream packageManifest = ClassLoader.getSystemResourceAsStream(pathPrefix + "/" + manifestFilename);
```
kroepke: System classloader vs the class' classloader (ETOOMANYCLASS)

graylog2-shared/src/main/java/org/graylog2/shared/initializers/AbstractJerseyService.java
```diff
@@ -0,0 +1,226 @@
...
+
+        for (Class additionalComponent : additionalComponents)
+            rc.registerClasses(additionalComponent);
+
+        if (enableGzip) {
+            EncodingFilter.enableFor(rc, GZipEncoder.class);
+        }
+
+        if (enableCors) {
+            LOG.info("Enabling CORS for REST API");
```
kroepke: minor nitpick: probably wrong wording when used in web interface context

graylog2-server/pom.xml
dennisoelkers: Yes, exactly. Especially because we want to facilitate hot reloading.

graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/BaseConfigurationTest.java
dennisoelkers: This is non-trivial, because the URL schemes of the REST API and the web interface clash. So this would mean that we would either have them a) live on two separate namespaces (like it is now) or b) invalidate all bookmarks that people might have created to the web interface.

graylog2-server/pom.xml
kroepke: Perfect!

graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/BaseConfigurationTest.java
kroepke: ah right, I forgot that one aspect: I would vote for eventually moving the REST API into /rest/* to avoid the clashes.
We can discuss this later though :+1:

dennisoelkers: I would be happy about that! :+1:

kroepke:
lgtm!

PR #1597, opened 2015-12-02
Bulk indices calls

dennisoelkers:
Add bulk call which returns all indices (open, closed, reopened), also includes the deflector config in the deflector index call. This helps reducing the number of round trips for the indices page in the web interface.

### DIFF
--- a/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/indexer/responses/BulkIndices.java
+++ b/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/indexer/responses/BulkIndices.java
@@ -1,3 +1,19 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
 package org.graylog2.rest.models.system.indexer.responses;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
--- a/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/indexer/responses/BulkIndices.java
+++ b/graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/indexer/responses/BulkIndices.java
@@ -1,0 +1,26 @@
+package org.graylog2.rest.models.system.indexer.responses;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+@JsonAutoDetect
+public abstract class BulkIndices {
+    @JsonProperty("closed")
+    public abstract ClosedIndices closed();
+
+    @JsonProperty("reopened")
+    public abstract ClosedIndices reopened();
+
+    @JsonProperty("all")
+    public abstract AllIndicesInfo all();
+
+    @JsonCreator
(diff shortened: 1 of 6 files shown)


graylog2-rest-models/src/main/java/org/graylog2/rest/models/system/indexer/responses/BulkIndices.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @JsonProperty("reopened")
+    public abstract ClosedIndices reopened();
+
+    @JsonProperty("all")
+    public abstract AllIndicesInfo all();
+
+    @JsonCreator
+    public static BulkIndices create(@JsonProperty("closed") ClosedIndices closed,
+                                         @JsonProperty("reopened") ClosedIndices reopened,
```
joschi: Nit-pick: indentation.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndicesResource.java
```diff
@@ -173,6 +174,16 @@ public boolean apply(String indexName) {
         return ClosedIndices.create(reopenedIndices, reopenedIndices.size());
     }
 
+    @GET
+    @Timed
+    @Path("/all")
+    @ApiOperation(value = "Get a list of all indices, open, closed and reopened.")
```
joschi: This reads strange. Why not "List all open, closed, and reopened indices"?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndicesResource.java
```diff
@@ -173,6 +174,16 @@ public boolean apply(String indexName) {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public BulkIndices bulk() {
```
joschi: Wouldn't it make sense to rename `all()` and the `AllIndicesInfo` class to `openIndices()` and `OpenIndicesInfo`? Currently it's kind of irritating that "all" and "bulk" do different things.

dennisoelkers: I think the whole API needs a bigger overhaul. The naming is inconsistent all over the place. I would prefer to decouple this from the creation of a bulk call though. Maybe you can spend some time doing that outside of the scope of this PR?

joschi: A simple rename should be possible within the scope of this PR. :wink:

edmundoa:
Please merge #1606 once the remaining issues are resolved. Thank you!

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndicesResource.java
dennisoelkers: Done.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndicesResource.java
```diff
@@ -104,10 +105,10 @@ public IndexInfo single(@ApiParam(name = "index") @PathParam("index") String ind
 
     @GET
     @Timed
-    @ApiOperation(value = "Get information of all indices managed by Graylog and their shards.")
+    @ApiOperation(value = "Get information of all open indices managed by Graylog and their shards.")
     @RequiresPermissions(RestPermissions.INDICES_READ)
     @Produces(MediaType.APPLICATION_JSON)
-    public AllIndicesInfo all() {
+    public OpenIndicesInfo open() {
```
joschi: I think this resource should be named similar to the other resources in this class, i. e. "/open" &rarr; "/system/indexer/indices/open".

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndicesResource.java
```diff
@@ -173,6 +174,16 @@ public boolean apply(String indexName) {
         return ClosedIndices.create(reopenedIndices, reopenedIndices.size());
     }
 
+    @GET
+    @Timed
+    @Path("/all")
```
joschi: I'm not too opinionated about it, but would it make sense to host this resource at "/system/indexer/indices" (i. e. where the `open()` method is currently placed)?

dennisoelkers: As I said before, this PR is scoped for a small optimization supporting the web interface. Any further redesign of the API should be done outside of it.

PR #1598, opened 2015-12-02
Support Elasticsearch 2.x

joschi:
This PR adds support for Elasticsearch 2.x (internal Elasticsearch node version: 2.1.0) to Graylog.

Since Elasticsearch 1.x and 2.x aren't compatible on the transport protocol level, Graylog won't be able to speak to an Elasticsearch 1.x cluster. Users with existing setups will have to migrate to Elasticsearch 2.x (requiring a [full cluster restart](https://www.elastic.co/guide/en/elasticsearch/reference/2.1/restart-upgrade.html)).


rikatz:
@joschi I've faced a problem with another setup of ElasticSearch (not with Graylog, but it's also valid): It doesn't accept anymore fields with the character '.' on it. (https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#_field_names_may_not_contain_dots)

As this may not be a problem to Graylog by itself, some extractors (mainly the JSON extractor) transform nested objects on fields named object.subobject.

This can break stuff, so I'm reporting this to you before the merge :)

bernd:
Regarding @rikatz note with the `.` character in message field names, we should probably ensure that the serialized `Message` object does not contain those.

rikatz:
@bernd As there are some cases of use with nested JSON, I don't know if that's better to replace that character with another one (like '_')

As I'm not a developer, I don't know if that's possible, but I do have some use cases that a JSON contains another JSON. Today, Graylog is able to parse that (generating fields like payload.bla), but will try to also create another extractor to parse that field also (flatten that, than parse).

Thanks

graylog2-bootstrap/pom.xml
```diff
@@ -80,21 +80,6 @@
...
-                <filtering>true</filtering>
-                <includes>
-                    <include>**/*.properties</include>
-                </includes>
-            </resource>
-            <resource>
-                <directory>src/main/resources</directory>
-            </resource>
-        </resources>
-
```
bernd: Why did you remove this?

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
```diff
@@ -37,8 +37,7 @@
         return ImmutableMap.of(
                 "properties", partFieldProperties(analyzer),
                 "dynamic_templates", partDefaultAllInDynamicTemplate(),
-                // Compress source field
-                "_source", enabledAndCompressed(),
+                "_source", enabled(),
```
bernd: Is the `_source` field compressed by default now? Or what is the reason to remove the compressions part?

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -127,22 +124,22 @@ public Period getPeriod() {
             return period;
         }
 
-        public DateHistogram.Interval toESInterval() {
+        public org.elasticsearch.search.aggregations.bucket.histogram.DateHistogramInterval toESInterval() {
```
bernd: Why are we using the fqcn here?

graylog2-server/src/main/java/org/graylog2/initializers/IndexerSetupService.java
```diff
@@ -61,8 +61,8 @@
 @Singleton
 public class IndexerSetupService extends AbstractIdleService {
     private static final Logger LOG = LoggerFactory.getLogger(IndexerSetupService.class);
-    private static final Version MINIMUM_ES_VERSION = Version.V_1_3_4;
-    private static final Version MAXIMUM_ES_VERSION = Version.fromString("1.7.99");
+    private static final Version MINIMUM_ES_VERSION = Version.V_2_0_0;
+    private static final Version MAXIMUM_ES_VERSION = Version.fromString("2.0.99");
```
bernd: Does this need an update to `2.1.99`?

graylog2-bootstrap/pom.xml
bernd: Ah, it's added to the main pom. Can you please move the comment as well?! Thanks!

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
joschi: This field is compressed by default now: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_compress_and_compress_threshold

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
joschi: Because of the conflicting nested enum of the same name in this very class.

graylog2-server/src/main/java/org/graylog2/initializers/IndexerSetupService.java
joschi: Yes. :wink:

joschi: Although we could discuss whether this check (for the maximum version) is still required or if it should simply be set to "2.99.99".

graylog2-bootstrap/pom.xml
joschi: It's not only the git-commit-id-plugin that requires resource filtering, so the comment was wrong or at least incomplete in the first place.

graylog2-server/src/main/java/org/graylog2/initializers/IndexerSetupService.java
bernd: :+1: for `2.99.99`

graylog2-bootstrap/pom.xml
bernd: :+1:

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
bernd: Aye.

bernd:
I am getting this every 30s.

```
2015-12-04 14:24:51,987 ERROR: org.graylog2.periodical.IndexerClusterCheckerThread - Uncaught exception in periodical
java.lang.NullPointerException
	at org.graylog2.periodical.IndexerClusterCheckerThread.doRun(IndexerClusterCheckerThread.java:68) ~[classes/:?]
	at org.graylog2.plugin.periodical.Periodical.run(Periodical.java:83) [classes/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_66]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_66]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_66]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_66]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_66]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_66]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_66]
```

bernd:
Looks like the `DATA_NODE_PREDICATE` in `Cluster.java` does not work as intended. An Elasticsearch data node does not have the "node.data" field...

bernd:
@joschi Any idea why the integrations build fails now? I can't really see the changes since you force pushed the amended commit.

graylog2-server/src/main/java/org/graylog2/bindings/providers/EsNodeProvider.java
```diff
@@ -48,43 +47,45 @@ public EsNodeProvider(ElasticsearchConfiguration configuration) {
     }
 
     @Override
-    @Singleton
```
bernd: Is moving the `@Singleton` in front of the class actually the same as having it on the `get()` method?

joschi: AFAIK the annotation is being ignored at the `Provider#get()` method. Only methods annotated with `@Provides` and `@Singleton` inside a Guice module will have the desired effect.

Additionally `EsNodeProvider` is explicitly being bound as eager singleton in `ServerBindings`, so the `@Singleton` annotation in `EsNodeProvider` is more for documentation purposes.

bernd: Aye, I just tested it.

- `@Singleton` on the `get()` method: does nothing
- `@Singleton` on the class: caches the provider instance, not the object returned from `get()`
- `.in(Scopes.SINGLETON)` or `.asEagerSingleton()` in bindinds: caches both, `get()` method return *and* the provider instance

joschi: :heart:

bernd:
LGTM :+1: Thank you!

s-schweer:
And why don't you just switch to the rest-client instead of using this shitty transport client which causes issues with every f***ing es update, is completely broken by design and subverts every serious security concept

joschi:
@s-schweer First of all, swearing won't get you anywhere, so please watch your language. Second, Graylog is not using the Elasticsearch transport client, but is spinning up an Elasticsearch client node instead. This makes a huge difference and is the reason for the quite tight coupling with Elasticsearch major versions.

With Graylog being part of the Elasticsearch cluster (via the embedded ES client node), it's possible to register cluster state listeners which send notifications if the ES cluster state changed (e. g. new nodes joined the cluster or indices were created/closed/deleted). Additionally, the embedded ES client node is aware of the cluster topology (and changes in that topology) so that routing messages to their destination (i. e. wherever that shard or replica shard is stored) are more efficient. All of that would not be possible if Graylog was using a simple ES transport client or even the ES HTTP interface.

s-schweer:
@joschi First of all, excuse my language, but I lots of java/scala developers like using the transport/node client for accessing es andthis is just ugly and causing many unnecessary Problems. 

> Graylog is not using the Elasticsearch transport client, but is spinning up an Elasticsearch client node instead

This is not better, this is even worse because this can lead to Situations where the embedded es node causes cluster inconsistencies and even cluster breakdowns.

 > With Graylog being part of the Elasticsearch cluster (via the embedded ES client node), it's possible to register cluster state listeners which send notifications if the ES cluster state changed (e. g. new nodes joined the cluster or indices were created/closed/deleted)

And for what do you need this? You can just query for this...

> Additionally, the embedded ES client node is aware of the cluster topology (and changes in that topology) so that routing messages to their destination (i. e. wherever that shard or replica shard is stored) are more efficient. All of that would not be possible if Graylog was using a simple ES transport client or even the ES HTTP interface.

You normally use dedicated indexer nodes for this task, ideallly behind a load-balancer.

joschi:
@s-schweer 
> This is not better, this is even worse because this can lead to Situations where the embedded es node causes cluster inconsistencies and even cluster breakdowns.

No, being a client node (i. e. doesn't store any data and cannot be elected master) it doesn't have any of those mentioned effects on an Elasticsearch cluster.

> And for what do you need this? You can just query for this...

This would require polling the (complete) cluster state every few seconds (or even every second) which imposes an unnecessary load on the ES cluster. Being a part of the ES cluster, Graylog  benefits from the incremental cluster state updates (since ES 2.0.0) and gets notified about changes instantly (instead of having to wait for the poll interval).

> You normally use dedicated indexer nodes for this task, ideallly behind a load-balancer.

General-purpose load-balancers don't know anything about the Elasticsearch cluster topology (e. g. on which nodes shards of an index exist).

If you're interested in continuing this discussion, please post to our public mailing list or join #graylog on freenode IRC.

s-schweer:
> No, being a client node (i. e. doesn't store any data and cannot be elected master) it doesn't have any of those mentioned effects on an Elasticsearch cluster.

Yes, in theory, in reality i saw exactly this caused by logstash's node client.

> General-purpose load-balancers don't know anything about the Elasticsearch cluster topology (e. g. on which nodes shards of an index exist).

Yeah, this is what the indexer nodes do...

Issue #737, opened 2014-10-10
Create tool for fixing wrong timestamps in Elasticsearch


kroepke:
This is to clean up the malformed messages of #728

Issue #728, opened 2014-09-26
Timestamp parse failed - v0.90.0


mcsage:
I found several logs like these and messages are displayed in a strange way.
See attached files

```
java.lang.IllegalArgumentException: Invalid format: "1411738868705" is malformed at "8705"
    at org.joda.time.format.DateTimeFormatter.parseDateTime(DateTimeFormatter.java:899)
    at org.graylog2.indexer.results.ResultMessage.setMessage(ResultMessage.java:80)
    at org.graylog2.indexer.results.ResultMessage.parseFromSource(ResultMessage.java:53)
    at org.graylog2.indexer.results.IndexQueryResult.buildResults(IndexQueryResult.java:64)
    at org.graylog2.indexer.results.SearchResult.<init>(SearchResult.java:43)
    at org.graylog2.indexer.searches.Searches.search(Searches.java:164)
    at org.graylog2.rest.resources.search.RelativeSearchResource.searchRelative(RelativeSearchResource.java:93)
    at sun.reflect.GeneratedMethodAccessor68.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
    at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:151)
    at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:171)
    at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$TypeOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:195)
    at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:104)
    at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:387)
    at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:331)
[...]

PR #1610, opened 2015-12-07
Remove ESTimestampFixup command

joschi:
The ESTimestampFixup was only used to fix a mapping problem when upgrading from Graylog 0.20.x or 0.90.x to Graylog 1.x.

Refs #728, refs #737.


bernd:
The OS packages need to be adjusted as well since they ship the script.

bernd:
LGTM :+1:

PR #1612, opened 2015-12-07
properly remove roles from users when the role is deleted

### LINKED ISSUE #1608: [1.3 beta] Regression: no users list
Hello,

Currently testing 1.3 beta, i'm not able to get users list page.
On log side, i see:

2015-12-04T15:35:13.134Z ERROR [RestResource] Unable to load role names for role IDs [563c874ee4b001c07c5f02a9, 563c87fae4b001c07c5f0364] for user UserImpl{fields={_id=5640d260e4b02855eda908f6, username=pdepaepe, email=pdepaepe@localhost.com, permissions=[ "users:edit:pdepaepe"], password=xxxxxxxxxx, full_name=pdepaepe's account, roles=[ { "$oid" : "563c874ee4b001c07c5f02a9"} , { "$oid" : "563c87fae4b001c07c5f0364"}]}, id=5640d260e4b02855eda908f6}

When i call GET roles, i don't see 563c874ee4b001c07c5f02a9 and 563c87fae4b001c07c5f0364, so i guess, theses references are stuck in user object ?

PS: roll-backed to 1.2, users are back again

kroepke:
- local admin users are not touched
 - when resolving roles of a user do not throw an exception for missing roles, it prevented the user from being editable
 - integration tests for role removal

fixes #1608

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/roles/RolesResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/roles/RolesResource.java
@@ -155,10 +155,30 @@ public class RolesResource extends RestResource {
 
     @DELETE
     @Path("{rolename}")
-    @ApiOperation(value = "Remove the named role")
+    @ApiOperation(value = "Remove the named role and unassign any users from it")
     public void delete(@ApiParam(name = "rolename", required = true) @PathParam("rolename") String name) throws NotFoundException {
         checkPermission(RestPermissions.ROLES_DELETE, name);
-        // TODO unassign roles from current role members
+
+        final Role role = roleService.load(name);
+        if (role.isReadOnly()) {
+            throw new BadRequestException("Cannot delete read only system role " + name);
+        }
+        final Collection<User> usersInRole = userService.loadAllForRole(role);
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/roles/RolesResource.java
```diff
@@ -155,10 +155,30 @@ public RoleResponse update(
...
         checkPermission(RestPermissions.ROLES_DELETE, name);
-        // TODO unassign roles from current role members
+
+        final Role role = roleService.load(name);
+        if (role.isReadOnly()) {
+            throw new BadRequestException("Cannot delete read only system role " + name);
+        }
+        final Collection<User> usersInRole = userService.loadAllForRole(role);
+        // remove role from any user still assigned
+        for (User user : usersInRole) {
```
bernd: How about moving the user related stuff into the `UserService`? That cleans up the resource and the code might be useful in other places as well.

kroepke: good point

bernd:
LGTM :+1:

PR #1613, opened 2015-12-07
Replace jersey-netty with Grizzly

joschi:
Instead of maintaining a custom Jersey adapter for Netty, Graylog is now using the Grizzly HTTP server which is tightly coupled with the Jersey (and Glassfish) project.

There are some backward-incompatible changes, though:

* `rest_worker_threads_max_pool_size` and `web_worker_threads_max_pool_size ` has been removed. Only `rest_thread_pool_size` and `web_thread_pool_size` are being used.


bernd:
> Support for private keys and X.509 certificates in PEM format has been removed 

Is this because it's plain impossible to do it with PEM formatted certificates or it's a pain to implement it?

I think it would be really bad if we lose PEM format support.

garybot2:

joschi:
> Is this because it's plain impossible to do it with PEM formatted certificates or it's a pain to implement it?

I've re-added support for PKCS#8 private keys and X.509 certificates in PEM format.

bernd:
> I've re-added support for PKCS#8 private keys and X.509 certificates in PEM format.

Awesome, thank you! :+1:

bernd:
I am currently unable to review this line by line because it's a pretty large diff. That said, I believe that we should merge this ASAP to make sure we have lots of time to run into issues.

I glanced over the changes and tested the following:

* Web interface
* API browser
* TLS without configuring any certificates and keys (it runs in with a generated certificate)
* TLS with custom certificates and key. (works as it did before :+1:)

LGTM

PR #1614, opened 2015-12-08
Support isMemberOf attribute of Sun Directory Server

### LINKED ISSUE #1583: make "memberOf" Searchfilter configurable
Hi,
we cannot use Ldap group Mapping because we are using a Sun Directory Server, 
the Ldapsearch `ldapsearch -x -LLL -h ldap.server  -b"dc=company,dc=tld" '(uid=user.name)' memberOf`
returns an empty Result.
Searching for `ismemberOf`instead, returns all the Groups.
The memberOf Filter seems to be hardcoded in
`graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java `
Line 135 and 154
Since this will be quite rare, maybe a configuration Setting in server.conf will suffice?

best regards
Serge

bernd:
Fixes #1583


kroepke:
lgtm!

PR #1615, opened 2015-12-08
Add message ID

edmundoa:
On the current master we are not returning IDs along with messages, breaking the API we had. These changes add them back into the message documents.


kroepke:
lgtm

PR #1616, opened 2015-12-08
Improve version.properties handling

bernd:
Just use the project.version and parse it in the Version class to avoid requiring "mvn process-resources".


kroepke:
lgtm

PR #1619, opened 2015-12-09
remove git-commit-id-plugin in favor of the simpler buildnumber helper

kroepke:
- properly allows filtering resources
 - remove most unused git info which we didn't need in the first place
 - manually short sha1 for display purposes

this also enables using the bundle intellij maven3 because git-commit-id-plugin requires a more recent version without any direct benefit to us

### DIFF
--- a/graylog2-bootstrap/pom.xml
+++ b/graylog2-bootstrap/pom.xml
@@ -80,6 +80,12 @@
     </dependencies>
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+            </resource>
+        </resources>
         <plugins>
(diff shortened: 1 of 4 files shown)


graylog2-bootstrap/pom.xml
```diff
@@ -80,6 +80,12 @@
     </dependencies>
 
     <build>
+        <resources>
```
joschi: Is this necessary? A similar section (filtering properties files) is already in the parent POM.

kroepke: Ah, will check. Actually the resources section in that pom looks odd, too, imho

joschi:
LGTM. :+1:

PR #1622, opened 2015-12-09
Replace SLF4JBridgeHandler with Log4j2 JDK Logging Adapter

joschi:
The SLF4JBridgeHandler didn't work as intended as the `java.util.logging.LogManager` only reads its configuration from regular files ("logging.properties") but not from the class path.

Since both variants, the SLF4J Bridge Handler and the Log4j2 LogManager has to be installed programmatically, this PR chooses the more performant variant.

* https://logging.apache.org/log4j/log4j-2.4/log4j-jul/index.html
* http://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html

### DIFF
--- a/graylog2-bootstrap/pom.xml
+++ b/graylog2-bootstrap/pom.xml
@@ -74,6 +74,10 @@
             <artifactId>log4j-core</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


bernd:
LGTM, tested with the Grizzly branch. :+1:

PR #1628, opened 2015-12-11
Install a Graylog index template instead of set mappings on index creation

### LINKED ISSUE #1624: ES dynamic template functionality is broken in Graylog 1.3
Hello, 

Since Graylog 1.3, the use of clever dynamic template to specify values type and doc_values is not possible anymore. Since you create the index with your own index template, you prevent anyone to specify any special mapping with a dynamic template. 
Here is the mapping we have now with graylog 1.3

```
{
  "graylog2_321" : {
    "mappings" : {
      "message" : {
        "dynamic_templates" : [ {
          "internal_fields" : {
            "mapping" : {
              "index" : "not_analyzed",
              "doc_values" : true
            },
            "match" : "gl2_*"
          }
        }, {
          "store_generic" : {
            "mapping" : {
              "index" : "not_analyzed"
            },
            "match" : "*"
          }
        },  {
          "num_suffix" : {
            "mapping" : {
              "type" : "double",
              "doc_values" : true
            },
            "match" : "*_num"
          }
        }, {
          "others" : {
            "mapping" : {
              "index" : "not_analyzed",
              "type" : "string",
              "doc_values" : true
            },
            "match" : "*"
          }
...
"properties" : {
          "thread" : {
            "type" : "long"
          },
...
```
"thread" should here a string, but it is now a long.  Note that also doc_values can not be specified anymore for non analyzed fields.
This is a problem since it was useful to prevent collisions to happen by enforcing conventions with ES.
I have also seen the dynamic_templates features as a solution proposed in some issues. I changed the order of the template but it doesn't help. It is working perfectly fine in 1.2 since the dynamic template mapping of Graylog is applied after the dynamic template present in ES.

bernd:
In commit edf0d8b we changed the index creation process from "create index" + "install index mapping" to one operation to avoid race conditions.

This broke setups where users had installed custom Elasticsearch index templates because the user templates where applied after our mapping was installed into the index. Before edf0d8b the user templates have been applied before we installed our mapping because the index creation was split into two steps.

We now install a "graylog-internal" index template into Elasticsearch with a "order" or -99 to make sure user templates are able to override our settings.

Fixes #1624

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -46,6 +46,7 @@ import org.elasticsearch.action.admin.indices.stats.IndexStats;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsRequest;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsResponse;
 import org.elasticsearch.action.admin.indices.stats.ShardStats;
+import org.elasticsearch.action.admin.indices.template.get.GetIndexTemplatesResponse;
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequest;
 import org.elasticsearch.action.bulk.BulkRequestBuilder;
 import org.elasticsearch.action.bulk.BulkResponse;
@@ -58,6 +59,7 @@ import org.elasticsearch.client.Client;
 import org.elasticsearch.client.IndicesAdminClient;
 import org.elasticsearch.cluster.ClusterState;
 import org.elasticsearch.cluster.metadata.IndexMetaData;
+import org.elasticsearch.cluster.metadata.IndexTemplateMetaData;
 import org.elasticsearch.cluster.metadata.MappingMetaData;
 import org.elasticsearch.cluster.routing.ShardRouting;
 import org.elasticsearch.common.collect.UnmodifiableIterator;
@@ -89,6 +91,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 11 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -211,6 +214,26 @@ public String aliasTarget(String alias) {
         return response.getAliases().isEmpty() ? null : response.getAliases().keysIt().next();
     }
 
+    public boolean createIndexTemplate() {
+        final Map<String, Object> template = indexMapping.messageTemplate(allIndicesAlias(), configuration.getAnalyzer());
+        final PutIndexTemplateRequest itr = c.admin().indices().preparePutTemplate(GRAYLOG_INTERNAL_TEMPLATE_NAME)
+                .setOrder(-99) // Make sure templates with "order: 0" are applied after our template!
```
joschi: Why -99 and not -1 or `Integer.MIN_VALUE`?

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -99,6 +100,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(Indices.class);
 
+    private static final String GRAYLOG_INTERNAL_TEMPLATE_NAME = "graylog-internal";
```
joschi: Why "internal"? Is there another Graylog index template?

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -211,6 +214,26 @@ public String aliasTarget(String alias) {
         return response.getAliases().isEmpty() ? null : response.getAliases().keysIt().next();
     }
 
+    public boolean createIndexTemplate() {
+        final Map<String, Object> template = indexMapping.messageTemplate(allIndicesAlias(), configuration.getAnalyzer());
+        final PutIndexTemplateRequest itr = c.admin().indices().preparePutTemplate(GRAYLOG_INTERNAL_TEMPLATE_NAME)
```
joschi: Could this cause problems if the template already exists and/or has been modified?

bernd: It will just overwrite any changes with our configured values, which is what we want I guess.

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
bernd: I chose -99 because I didn't want to use -1 because we might need some space in the future. :wink: 

I will change it to `Integer.MIN_VALUE`.

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
bernd: I used that name to make it clear that this is our Graylog internal mapping and because users might have a template named "graylog" already.

bernd:
Updated after some discussions.

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
joschi: :white_check_mark: Fair enough.

joschi:
LGTM. :+1:

joschi:
Merged into `master` in be54753ed5979af44c53ac05748619be58e231aa.

PR #1629, opened 2015-12-14
Add hostname to periodic cluster node registration record.

dennisoelkers:
This is done to avoid additional roundtrips to the server for displaying
a node's hostname, which was available only when querying the
SystemResource of that specific node directly.
The hostname is used when linking to nodes to show it along with the
short node id, trying to help the user identifying specific nodes.

Right now the hostname is determined only when the initial node
registration is created. If this turns out to be a problem, we can do it
everytime the node records is touched.


joschi:
LGTM. :+1:

PR #1631, opened 2015-12-14
Simplify Inputs handling in server

dennisoelkers:
This PR does:

* Change InputsResource into plain CRUD resource
* Extract inputs launching into InputStateListener triggered by ClusterEventBus events
* Add missing functionality of Inputs page back

The inputs page is not complete yet, this PR should be reviewed earlier because it has already grown way too big.


graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterInputStatesResource.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.rest.resources.cluster;
+
+import com.codahale.metrics.annotation.Timed;
+import com.wordnik.swagger.annotations.Api;
+import com.wordnik.swagger.annotations.ApiOperation;
+import edu.emory.mathcs.backport.java.util.Collections;
```
joschi: Wrong import. :wink:

dennisoelkers: Yikes :)

graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/InputsResource.java
```diff
@@ -0,0 +1,208 @@
...
+                .path("{inputId}")
+                .build(input.getId());
+
+        return Response.created(inputUri).entity(InputCreated.create(input.getId())).build();
+    }
+
+    private InputSummary getInputSummary(Input input) {
+        return InputSummary.create(input.getTitle(),
+                input.isGlobal(),
+                input.getType(),
```
edmundoa: This is changing the values we return for input names from a human readable name to be the same as the input type (which is something like `org.graylog2.inputs.random.FakeHttpMessageInput`). IMO we should replicate the old behaviour or, if there is a good reason to not do that, remove the `name` field from the response.

graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
```diff
@@ -174,6 +224,16 @@ public void addStaticField(Input input, final String key, final String value) th
     }
 
     @Override
+    public Input disable(Input input) {
+        return null;
+    }
+
+    @Override
+    public Input enable(Input input) {
```
edmundoa: What are the `enable` and `disable` methods for? I guess you added them to handle the input state, but I ask anyway.

dennisoelkers: This is a residue of the postponed plan to add those. I will remove them for this PR.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/InputsResource.java
dennisoelkers: Totally right. Gonna change the value of these back to the one we had. Good find, thanks! :+1:

edmundoa:
LGTM :+1:

PR #1633, opened 2015-12-15
Add option to AMQP transports to bind the queue to the exchange

### LINKED ISSUE #1599: Raw AMQP not creating exchange binding
Starting some tests with RabbitMQ + Graylog, after creating a raw/plaintext AMQP input I've noticed that Graylog doesn't create the exchange => queue binding.

It creates the queue succesfully, but the queue has empty bindings, shown on the attached images.

glogos1
glogos2
glogos3

EDIT: This way, I do have to manually bind the exchange with the log queue.
Forgot to mention, I'm using Graylog Server 1.2.2.

bernd:
This was missing before and the exchange and routing-key options were not used.

The binding option defaults to "false" to avoid existing setups to fail when the given exchange (there is a default) does not exist.

Fixes #1599

### DIFF
--- a/graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpConsumer.java
+++ b/graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpConsumer.java
@@ -54,6 +54,7 @@ public class AmqpConsumer {
 
     private final String queue;
     private final String exchange;
+    private final boolean exchangeBind;
     private final String routingKey;
     private final boolean requeueInvalid;
 
@@ -71,7 +72,7 @@ public class AmqpConsumer {
     private AtomicLong lastSecBytesReadTmp = new AtomicLong(0);
 
     public AmqpConsumer(String hostname, int port, String virtualHost, String username, String password,
-                        int prefetchCount, String queue, String exchange, String routingKey, int parallelQueues,
+                        int prefetchCount, String queue, String exchange, boolean exchangeBind, String routingKey,int parallelQueues,
                         boolean tls, boolean requeueInvalid, int heartbeatTimeout, MessageInput sourceInput,
                         ScheduledExecutorService scheduler, AmqpTransport amqpTransport) {
         this.hostname = hostname;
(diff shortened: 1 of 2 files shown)


graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpTransport.java
```diff
@@ -292,6 +294,15 @@ public ConfigurationRequest getRequestedConfiguration() {
             );
 
             cr.addField(
+                    new BooleanField(
+                            CK_EXCHANGE_BIND,
+                            "Bind to exchange?",
```
joschi: Is this a question? :wink:

graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpConsumer.java
```diff
@@ -108,6 +110,9 @@ public void run() throws IOException {
         for (int i = 0; i < parallelQueues; i++) {
             final String queueName = String.format(queue, i);
             channel.queueDeclare(queueName, true, false, false, null);
+            if (exchangeBind) {
+                channel.queueBind(queueName, exchange, routingKey);
```
joschi: Should we check whether the exchange exists and issue a warning if it doesn't?

graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpTransport.java
```diff
@@ -292,6 +294,15 @@ public ConfigurationRequest getRequestedConfiguration() {
...
+                            false,
+                            "Binds the queue to the configured exchange. Make sure the exchange exists!"
```
joschi: Instead of "Make sure the exchange exists!" maybe write "The exchange must already exist."

graylog2-inputs/src/main/java/org/graylog2/inputs/transports/AmqpConsumer.java
bernd: If the input cannot be started due to a missing exchange, there is a warning in the web interface that is enough to debug the issue. It actually says that the exchange does not exist.

bernd:
I updated the wording for the config options.

joschi:
LGTM. :+1:

PR #1635, opened 2015-12-15
Make index retention strategies pluggable

bernd:
- Convert RetentionStrategy to an interface
- Add addRetentionStrategy() helper to PluginModule
- Move retention strategy configuration into the cluster configuration
- Move index count based strategy logic to abstract class


joschi:
LGTM. :+1:

PR #1636, opened 2015-12-15
Merge IndexRotationThread and IndexRetentionThread into one

bernd:
Move "perform_retention" config option into IndexManagementConfig.


bernd:
Merging both threads might create issues which long running retention strategies.

A long retention job might block the rotation for a long time. Rotation is running every 10 seconds.

This needs more thinking. Maybe it just does not make sense to merge them.

bernd:
We will keep them separated for now.

PR #1645, opened 2015-12-21
Add input state controls

edmundoa:
This PR adds back the input state controls on the web interface, and also adds `start` and `stop` resources on `ClusterInputStatesResource`, allowing to start and stop global inputs in all cluster nodes.

Input states are not persisted at the moment, so restarting a node will also start all stopped inputs.


graylog2-web-interface/src/stores/inputs/InputStatesStore.js
```diff
@@ -28,13 +28,53 @@ const InputStatesStore = Reflux.createStore({
...
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.ClusterInputStatesController.start(input.id).url);
+    } else {
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.InputStatesController.start(input.id).url);
+    }
+
+    return fetch('PUT', url)
+      .then(
+        () => {
+          this.list();
+          UserNotification.success(`Input '${input.title}' started successfully`);
```
dennisoelkers: This does not really take the response into consideration, it just checks if the server response is a 200, which is the case even if starting the input failed on all nodes concerned.

Maybe we should add a simple check for the presence of any null elements in the hash response. If there are successful (!= null) and failed (== null) entries, we could notify a warning, if there are only null entries we should indicate an error.

graylog2-web-interface/src/stores/inputs/InputStatesStore.js
```diff
@@ -28,13 +28,53 @@ const InputStatesStore = Reflux.createStore({
...
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.ClusterInputStatesController.stop(input.id).url);
+    } else {
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.InputStatesController.stop(input.id).url);
+    }
+
+    return fetch('DELETE', url)
+      .then(
+        () => {
+          this.list();
+          UserNotification.success(`Input '${input.title}' stopped successfully`);
```
dennisoelkers: See comment for `start(input)`.

graylog2-web-interface/src/stores/inputs/InputStatesStore.js
edmundoa: Nice catch. The problem is that the response from the server is not returning anything different when the input could start or not, so I think that's the first thing to change. Will improve that and update the PR with the changes :)

edmundoa: It turns out, it is difficult to know when an input is started or stopped in the server side. Maybe I'm missing something, but it looks like Guava's EventBus does not provide any way of checking the status of the listeners, and we are using an async event bus, which makes everything a bit more complicated, as the execution of listeners may be delayed.

In the end I did something similar to what you suggested, checking the result of the input states list request. I don't like that we don't provide feedback on the start/stop actions when using the REST API, but I guess we'll have to live with that for now.

graylog2-web-interface/src/stores/inputs/InputStatesStore.js
```diff
@@ -28,13 +28,109 @@ const InputStatesStore = Reflux.createStore({
...
+        });
+  },
+
+  stop(input) {
+    let url;
+
+    if (input.global) {
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.ClusterInputStatesController.stop(input.id).url);
+    } else {
+      url = URLUtils.qualifyUrl(jsRoutes.controllers.api.InputStatesController.stop(input.id).url);
```
dennisoelkers: What happens when the input is not global, but not running on the node we're connected too?

I think we should not differentiate the input type (global/non-global) on the frontend side, but in the ClusterInputStatesController, which has far more knowledge to determine the termination strategy.

graylog2-web-interface/src/stores/inputs/InputStatesStore.js
```diff
@@ -28,13 +28,109 @@ const InputStatesStore = Reflux.createStore({
...
+    }
+
+    const handleSuccess = () => UserNotification.success(`Input '${input.title}' stopped successfully`);
+    const handleError = (message) => UserNotification.error(message, `Input '${input.title}' could not be stopped`);
+
+    return fetch('DELETE', url)
+      .then(
+        () => {
+          this.list().then(inputStatesList => this._checkStoppedState(input, inputStatesList, handleSuccess, handleError));
```
dennisoelkers: I am still not happy about this. My comment was targeted towards the results of the individual start/stop calls. As input launching/stopping is done asynchronously, we cannot check for the success of the whole operation, but we can (and should) check if _triggering_ them is successful. 

The way it is done now implies that the user is notified only when input startup/termination succeeds/fails during the short time span of sending the request which triggers those and finishing the `list` operation. This might be too short for some inputs. So my suggestion is:

1) Just check for the individual success/failure of the operations triggering async start/stop and indicate the result (all succeeded, mixed, all failed) to the user.
2) Signal status of input states separately using badges in input list (what we do already)

graylog2-web-interface/src/stores/inputs/InputStatesStore.js
edmundoa: Completely missed that :+1:

edmundoa:
I saw some problems with metrics while testing the scenario where a node is not reachable, I'll add an issue for that so we don't forget.

graylog2-web-interface/src/components/inputs/InputStateControl.jsx
```diff
@@ -0,0 +1,48 @@
...
+  mixins: [Reflux.connectFilter(InputStatesStore, 'inputState', inputStateFilter)],
+  _isInputRunning() {
+    if (!this.state.inputState) {
+      return false;
+    }
+
+    const nodeIDs = Object.keys(this.state.inputState);
+    let isInputRunning = true;
+
+    for (let i = 0; i < nodeIDs.length && isInputRunning; i++) {
```
dennisoelkers: What happens when `nodeIDs.length` is 0? In this case the input is running nowhere, but `isInputRunning` is still true. Therefore there will be no way to start the input as the controls will only show a stop button.

Unfortunately this is not a corner case, because I think it will happen every time an input is stopped, as the input state is deleted when stopping an input.

edmundoa: I haven't seen any case where `nodeIDs.length` was 0, it certainly didn't happen when stopping inputs on local and global inputs on my tests. The response from the server includes information for each node, even if the information is `null` or an empty array.

Anyway, you are right that it may lead to some weird state if that ever happen, so I'll take a quick look into it.

PR #1665, opened 2016-01-05
Replace mentions of "Graylog2" with "Graylog"

joschi:
There are still several places where the code base mentions "Graylog2" instead of "Graylog". This PR replaces those occurrences in the configuration file (and the corresponding code) and renames the prefixes to override configuration settings to `GRAYLOG_` (environment variables) and `graylog.` (JVM system properties) respectively.


bernd:
What is the upgrade path for the changed ES cluster name, index prefix and the changed MongoDB database?

joschi:
> What is the upgrade path for the changed ES cluster name, index prefix and the changed MongoDB database?

Just the same as when users change those settings manually: They'll have to adjust the other systems accordingly. For existing users, they can simply set the configuration values to their old defaults.

bernd:
@joschi Please add a `UPGRADE-NOTES.md` or `CHANGELOG.md` and document the upgrade steps for this.

I think we should track this information in the repository so we can compile the changelogs from this. Otherwise it's hard to track if the needed documentation is done already.

bernd:
LGTM :+1:

PR #1666, opened 2016-01-05
Add "remove" method to ClusterConfigService

joschi:
Add "remove" method to `ClusterConfigService` so that existing cluster configuration entries can not only be written and retrieved, but also be removed.


bernd:
LGTM :+1:

PR #1667, opened 2016-01-06
Handling "." characters in message field names with ES 2.0

bernd:
Elasticsearch 2.0 does not allow "." characters in keys anymore. We have to decide what to do.

I see the following options:

1. Ignore fields with `.` in the key name
1. Convert the `.` character to something else (problem if the converted key already exists)
1. Encode the `.` to something that can be stored in ES before writing it and decode it when reading from ES

I implemented the first option in this PR as a start for the discussion.

There are some more ES changes that we might have to take into account.

Example: https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#_type_names_may_not_be_longer_than_255_characters

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
@@ -65,7 +65,7 @@ public class Message implements Messages {
     public static final String FIELD_LEVEL = "level";
     public static final String FIELD_STREAMS = "streams";
 
-    private static final Pattern VALID_KEY_CHARS = Pattern.compile("^[\\w\\.\\-@]*$");
+    private static final Pattern VALID_KEY_CHARS = Pattern.compile("^[\\w\\-@]*$");
 
     public static final ImmutableSet<String> RESERVED_FIELDS = ImmutableSet.of(
             // ElasticSearch fields.
@@ -254,6 +254,9 @@ public class Message implements Messages {
         if (RESERVED_FIELDS.contains(key) && !RESERVED_SETTABLE_FIELDS.contains(key) || !validKey(key)) {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Ignoring invalid or reserved key {} for message {}", key, getId());
+            }
+            if (key != null && key.contains(".")) {
+                LOG.warn("Keys must not contain a \".\" character! Ignoring field \"{}\"=\"{}\" in message [{}].", key, value, getId());
             }
             return;
         }
--- a/graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/MessageTest.java
+++ b/graylog2-plugin-interfaces/src/test/java/org/graylog2/plugin/MessageTest.java
@@ -259,13 +259,14 @@ public class MessageTest {
         assertTrue(Message.validKey("foo123"));
         assertTrue(Message.validKey("foo-bar123"));
         assertTrue(Message.validKey("foo_bar123"));
-        assertTrue(Message.validKey("foo.bar123"));
+        assertTrue(Message.validKey("foo@bar"));
         assertTrue(Message.validKey("123"));
         assertTrue(Message.validKey(""));
 
         assertFalse(Message.validKey("foo bar"));
         assertFalse(Message.validKey("foo+bar"));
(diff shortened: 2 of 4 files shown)


bernd:
No comments on this? @joschi @dennisoelkers @kroepke @lennartkoopmann @edmundoa @mariussturm

joschi:
Fourth option: support nested structures in indexed messages, although that will lead to a lot of yaks yearning to be shaved.

I think option 1 (dropping message fields with a dot in their name) is the easiest solution but will hit some users by surprise (even if we state that clearly in our formidable upgrade notes).

The other two options (basically mapping dots to another character) have their share of problems, first and foremost the handling of duplicate keys (e. g. if we replace `.` with `_`). This being said, I think finding a suitable replacement character might just work good enough. I'm proposing to use a single forward slash as a replacement for the dot, e. g. `foo.bar` &rarr; `foo/bar`. We're currently not allowing `/` as part of a valid key in GELF 1.1, so that should keep the collisions with existing setups to a minimum.

joschi:
elastic/elasticsearch#9059 might also be interesting in this context, but it won't result in code very soon (currently scheduled for ES 2.3.0).

bernd:
I agree, supporting nested structures is a huge undertaking which is definitely out of scope at the moment.

@joschi If we use a replacement character like `/`, would you just convert the char on ingest or also during queries? Like parsing the query, replacing `.` with `/`, converting field names from `/` to `.` when showing them in the UI. Or would you just replace it and then work with the `/` for the rest?

joschi:
@bernd The latter, convert the dots in field names to forward slashes during ingest and then only handle the modified field name. Showing the dotted field names in the UI when they're really not there might just cause more confusion and we'd have to change a lot of functionality for this "translation".

bernd:
> The latter, convert the dots in field names to forward slashes during ingest and then only handle the modified field name. Showing the dotted field names in the UI when they're really not there might just cause more confusion and we'd have to change a lot of functionality for this "translation".

Sounds good to me. It's better than ignoring fields.

kroepke:
As I've said on the PR before, I think we should simply not allow the `.` character in field names.

IMHO there are two options:
* allowing `.` everywhere but make it the output's (and the corresponding search classes') responsibility to translate the field names to whatever is acceptable in its domain
* not allowing `.` anywhere, this includes setting it from extractors, drools or custom code

All the intermediate solutions are brittle because they open up the opportunity to silently and worse unexpectedly break.

In an ideal world I would like to implement this in both output and search classes, allowing field names to contain arbitrary characters. Of course this creates all kinds of other problems including having field name restrictions based on whatever query language is supported and necessitates escaping all over the place. Worse, part of the logic lives in the front end code right now.

Here's how "the other guys" handle it:
```
Splunk only accepts field names that contain alpha-numeric characters or an underscore:

* Valid characters for field names are a-z, A-Z, 0-9, or _ .
* Field names cannot begin with 0-9 or _ . Leading underscores are reserved for Splunk's internal variables.
* International characters are not allowed.
```

I think we can force the issue this way, `_` is an acceptable character replacement for `.` imho. The case where a message contains fields that would collide after transforming the field name should be the absolute corner case.

joschi:
> Here's how "the other guys" handle it:

And the other "other guys": https://www.elastic.co/blog/introducing-the-de_dot-filter

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -253,6 +253,9 @@ public void addField(final String key, final Object value) {
             if (LOG.isDebugEnabled()) {
                 LOG.debug("Ignoring invalid or reserved key {} for message {}", key, getId());
             }
+            if (key != null && key.contains(".")) {
+                LOG.warn("Ignoring field \"{}\"=\"{}\" - Keys with a \".\" character are not supported anymore since Graylog 2.0! (Message[{}])", key, value, toString());
```
joschi: I think this shouldn't print the complete message but only the message Id for reference. Otherwise the log statements might become quite big and unwieldy.

joschi: I would also the "since Graylog 2.0" part in the warning message. In the end users don't care since when a feature didn't work anymore. :wink:

PR #1679, opened 2016-01-08
Always include index range of deflector target index

### LINKED ISSUE #1672: Search with absolute-time shows no message, relative shows them
Hi,

When i perform a search with absolute time-stamps it shows a found messages value but didn't display anything. Only with relative time-stamps i get the messages.

It looks like i can only get the messages out of the last closed index.
http://i.imgur.com/6AnvTF6.png http://i.imgur.com/v1EXGhT.png

System:
Graylog 1.3.2 (e7c49b6) (Hansa)

joschi:
The index range of the deflector target index can't be calculated until the index has been rotated, so a dummy range is being used. For this reason, the latest index *always* has to be included in searches and not only in the case of searches with relative time range.

Fixes #1672

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
@@ -16,6 +16,8 @@
  */
 package org.graylog2.indexer;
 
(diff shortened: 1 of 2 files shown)


bernd:
LGTM :+1:

PR #1686, opened 2016-01-11
Add support for plugins providing routes and (system) navigation elements.

dennisoelkers:
This PR adds support for routes to complete pages (embedded in the general layout of the web interface) and navigational elements. Usage of this is described in the default [`index.jsx`](https://github.com/Graylog2/graylog-plugin-archetype/blob/master/src/main/resources/archetype-resources/src/web/index.jsx) file generated by the maven plugin archetype.


edmundoa:
LGTM :+1:

PR #1688, opened 2016-01-12
Discard Message Output for Graylog2 1.3

henrikjohansen:
I have butchered the Benchmark Output into something that simply discards messages - the rationale behind that is documented here : 

https://medium.com/@henrikjohansen/graylog2-forwarder-nodes-5ec5e4f97e21#.yptnbrtv0

Please consider adding this to the 1.3 branch :)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/outputs/DiscardMessageOutput.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/DiscardMessageOutput.java
@@ -1,0 +1,102 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/outputs/DiscardMessageOutput.java
```diff
@@ -0,0 +1,102 @@
...
+        long maxOffset = Long.MIN_VALUE;
+
+        for (final Message message : messages) {
+            maxOffset = Math.max(message.getJournalOffset(), maxOffset);
+        }
+
+        journal.markJournalOffsetCommitted(maxOffset);
+    }
+
+    public interface Factory extends MessageOutput.Factory<GelfOutput> {
```
joschi: This should rather be typed to `DiscardMessageOutput` instead of `GelfOutput` (same for the `create` method).

joschi:
I'll fix the return types in `DiscardMessageOutput.Factory` after merging and add a simple metric of the number of discarded messages after merging.

Thanks!

razvanphp:
Just to be clear, can we use this in a graylog server without any ES connected to it?

Would it be something like a forwarder to another graylog server (that does the storage instead) ?

kroepke:
Yes, that is one use case for it. Another one could be to run benchmarks having to have storage attached, although we do ship with a benchmark output that writes timings and counts to a CSV file. However if you don't want that you could use this output.

razvanphp:
Well how do we launch it? It's not present in graylog-web 1.3.3, I can only choose GELF Output and STDOUT Output.

joschi:
@razvanphp This message output is supposed to replace the default output (see https://github.com/Graylog2/graylog2-server/blob/1.3.3/graylog2-server/src/main/java/org/graylog2/Configuration.java#L131-L132) and not to be used as a stream output.

Issue #725, opened 2014-09-25
Remember last rotation time for an index


kroepke:
To avoid having to rotate on startup with time based retention we need to know the time when an index was created and rotated.
Create a utility type and document in each elasticsearch index to store metadata in.
This can also be used for index ranges.

Follow up to #137

Ubay:
After upgrading to version 1.2.0 i'm still facing this issue:

2015-09-16T12:30:21.097+01:00 INFO  [IndexRotationThread] Deflector index <graylog2_142> should be rotated, Pointing deflector to new index now!
2015-09-16T12:30:21.097+01:00 INFO  [Deflector] Cycling deflector to next index now.
2015-09-16T12:30:21.131+01:00 INFO  [Deflector] Cycling from <graylog2_142> to <graylog2_143>
2015-09-16T12:30:21.132+01:00 INFO  [Deflector] Creating index target <graylog2_143>...
2015-09-16T12:30:21.644+01:00 INFO  [Deflector] Waiting for index allocation of <graylog2_143>
2015-09-16T12:30:21.646+01:00 INFO  [Deflector] Done!
2015-09-16T12:30:21.647+01:00 INFO  [Deflector] Pointing deflector to new target index....

After restarting server the deflector is rotated again:

2015-09-16T12:59:18.830+01:00 INFO  [IndexRotationThread] Deflector index <graylog2_143> should be rotated, Pointing deflector to new index now!
2015-09-16T12:59:18.830+01:00 INFO  [Deflector] Cycling deflector to next index now.
2015-09-16T12:59:18.869+01:00 INFO  [Deflector] Cycling from <graylog2_143> to <graylog2_144>
2015-09-16T12:59:18.870+01:00 INFO  [Deflector] Creating index target <graylog2_144>...
2015-09-16T12:59:19.257+01:00 INFO  [Deflector] Waiting for index allocation of <graylog2_144>
2015-09-16T12:59:19.260+01:00 INFO  [Deflector] Done!
2015-09-16T12:59:19.260+01:00 INFO  [Deflector] Pointing deflector to new target index....

Thank you.

kroepke:
Could you please provide a debug log for the logger: org.graylog2.periodical.IndexRotationThread ?

Also, I notice that the log statements you pasted are edited, this makes it really difficult to help.

joschi:
> Also, I notice that the log statements you pasted are edited, this makes it really difficult to help.

[...]

PR #1693, opened 2016-01-14
Fix creation_date key.

macfergus:
I think this will fix https://github.com/Graylog2/graylog2-server/issues/725, in which the index is always rotated at startup when you're using TimeBasedRotationStrategy.

The settings key was wrong, so the creation_date always came up as 0L, which indicated that the index was overdue for rotation. Tested this patch in our setup and it works like a champ.


joschi:
Thanks! I've corrected the used index settings key and added some tests to avoid regressions.

PR #1703, opened 2016-01-19
Avoid duplicate node registration in NodeService

bernd:
This fixes an issue where `NodeService#registerServer()` created a duplicate entry for a node instead of updating the existing one.


joschi:
LGTM. :+1:

PR #1706, opened 2016-01-20
Fix REST API browser with plugins

bernd:
Without this, REST resources from plugins will not show up in the API browser.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
@@ -25,6 +25,7 @@ import com.fasterxml.jackson.module.jsonSchema.JsonSchema;
 import com.fasterxml.jackson.module.jsonSchema.factories.SchemaFactoryWrapper;
 import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -39,6 +40,7 @@ import org.reflections.Reflections;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.FormParam;
 import javax.ws.rs.GET;
@@ -62,6 +64,9 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.google.common.base.Strings.nullToEmpty;
 
 /**
  * This is generating API information in Swagger format.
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
```diff
@@ -81,18 +85,37 @@
...
+            reflections = new Reflections((Object[]) packageNames.keySet().toArray());
         }
     }
 
     public Generator(String packageName, ObjectMapper mapper) {
-        this(new String[]{packageName}, mapper);
+        this(ImmutableMap.of(packageName, ""), mapper);
+    }
+
+    public String prefixedPath(Class<?> resourceClass, @Nullable String annotationPath) {
```
joschi: This method should be `private` or at least `protected` (if we had a test for it).

graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
```diff
@@ -81,18 +85,37 @@
...
+        final String path = (annotationPath == null ? "" : annotationPath);
+        final String resourcePackage = resourceClass.getPackage().getName();
+        final Optional<String> key = packageNames.keySet().stream().filter(resourcePackage::startsWith).findFirst();
```
joschi: It was not immediately clear to me, why this was needed or what the contents of this variable will be. Maybe `parentPackage` would be a better name?
Additionally, `packageNames.keySet()` isn't necessarily ordered or correctly ordered. Does this matter at all?

bernd: I will rename the variable.

The order would only matter if two plugins would use the same package.

bernd:
Updated after comments.

graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
joschi: > The order would only matter if two plugins would use the same package.

Or the same package prefix, e. g. it might lead to unexpected behavior if one plugin was using the package `com.example.plugins` and another one `com.example.plugins.foobar`.

bernd:
> Or the same package prefix, e. g. it might lead to unexpected behavior if one plugin was using the package com.example.plugins and another one com.example.plugins.foobar.

I pushed an update with a different approach. This should work. :smile:

graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
```diff
@@ -81,18 +83,42 @@
...
     }
 
     public Generator(String packageName, ObjectMapper mapper) {
-        this(new String[]{packageName}, mapper);
+        this(ImmutableSet.of(packageName), ImmutableMap.of(), "", mapper);
+    }
+
+    private String prefixedPath(Class<?> resourceClass, @Nullable String resourceAnnotationPath) {
+        final String resourceClassName = resourceClass.getCanonicalName();
+        final String resourcePath = (resourceAnnotationPath == null ? "" : resourceAnnotationPath);
```
joschi: Minor nitpick: This could be replaced by [`Strings#nullToEmpty(String)`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Strings.html#nullToEmpty(java.lang.String)).

bernd: Done

joschi:
LGTM. :+1:

PR #1708, opened 2016-01-20
Log error when a SystemJob throws an exception

bernd:
Before this there was no error indication if a SystemJob failed.


joschi:
LGTM. :+1:

PR #1713, opened 2016-01-21
Messageprocessor refactor

kroepke:
Refactor ProcessBufferProcessors to use MessageProcessor instead of MessageFilter instances

 - Message processors transform Messages into Messages, iterating over each registered processor passing single messages.
 - Existing MessageFilters are being handled by the MessageFilterChainProcessor to preserve backwards compatiblity with existing filter plugins.
 - MessageProcessor is pluggable, and their order can be controlled at runtime through a MessageProcessorOrder cluster configuration.
 - ResultMessage now directly contains Message objects, as a first step to unify the two views onto a Message.

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Messages.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Messages.java
@@ -1,0 +1,26 @@
+/**
+ * The MIT License
+ * Copyright (c) 2012 Graylog, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.graylog2.plugin;
+
+public interface Messages extends Iterable<Message> {
+}
--- a/graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java
@@ -25,7 +25,6 @@ import org.graylog2.plugin.Message;
 
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 /**
@@ -65,10 +64,10 @@ public class SearchResult extends IndexQueryResult {
 
         Iterator<ResultMessage> i = hits.iterator();
         while(i.hasNext()) {
-            final Map<String, Object> message = i.next().getMessage();
-            allFields.addAll(message.keySet());
+            final Message message = i.next().getMessage();
+            allFields.addAll(message.getFieldNames());
 
-            for (String field : message.keySet()) {
+            for (String field : message.getFieldNames()) {
                 if (!Message.RESERVED_FIELDS.contains(field)) {
                     filteredFields.add(field);
                 }
--- a/graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
@@ -1,0 +1,113 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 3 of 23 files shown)


graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Messages.java
```diff
@@ -0,0 +1,38 @@
...
+package org.graylog2.plugin;
+
+import java.util.Collections;
+import java.util.Iterator;
+
+public interface Messages extends Iterable<Message> {
+
+    EmptyMessages EMPTY_MESSAGES = new EmptyMessages();
+
+    class EmptyMessages implements Messages {
```
joschi: I would move this class next to `SingletonMessages` (for consistency and so it doesn't "pollute" the marker interface).

graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java
```diff
@@ -65,10 +64,10 @@ public long getTotalResults() {
 
         Iterator<ResultMessage> i = hits.iterator();
         while(i.hasNext()) {
```
joschi: I'm wondering why we're not using foreach here?
```java
for(ResultMessage message : hits) {
    // ...
}
```

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+
+import static com.codahale.metrics.MetricRegistry.name;
+
+public class MessageFilterChainProcessor implements MessageProcessor {
+    private static final Logger LOG = LoggerFactory.getLogger(MessageFilterChainProcessor.class);
+
+    private final List<MessageFilter> filterRegistry;
```
joschi: Shouldn't this be a `SortedSet` to match the constructor argument `filterRegistry`?

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
```diff
@@ -0,0 +1,112 @@
...
+            public int compare(MessageFilter filter1, MessageFilter filter2) {
+                return ComparisonChain.start()
+                        .compare(filter1.getPriority(), filter2.getPriority())
+                        .compare(filter1.getName(), filter2.getName())
+                        .result();
+            }
+        }).immutableSortedCopy(filterRegistry);
+
+        if (filterRegistry.size() == 0)
+            throw new RuntimeException("Empty filter registry!");
```
joschi: This should be an `IllegalArgumentException` (could also use `Preconditions#checkArgument()`).

graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java
kroepke: because i didn't touch any pre-existing code :smile:

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Messages.java
kroepke: makes sense!

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
```diff
@@ -0,0 +1,53 @@
...
+    public abstract long changeVersion();
+
+    @JsonProperty
+    public abstract List<String> classOrder();
+
+    @JsonProperty
+    public abstract Set<String> disabledMessageProcessors();
+
+    @JsonCreator
+    public static MessageProcessorOrder create(@JsonProperty("changeVersion") long changeVersion,
```
joschi: We're using lower case identifiers with underscores everywhere else (and that's also the default naming strategy for `ObjectMapper` in Graylog), so this might not work as expected (e. g. `changeVersion` will become `change_version` on serialization but for deserialization it expects `changeVersion`).

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
```diff
@@ -0,0 +1,53 @@
...
+                                               @JsonProperty("classOrder") List<String> classOrder,
+                                               @JsonProperty("disabledMessageProcessors") Set<String> disabledMessageProcessors) {
+        return new AutoValue_MessageProcessorOrder(changeVersion, classOrder, disabledMessageProcessors);
+    }
+
+    @JsonCreator
```
joschi: I'm not sure about the behavior of Jackson with multiple `@JsonCreator` annotations in a class. I think it'll only ever try to use the first one.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
```diff
@@ -0,0 +1,53 @@
...
+    public static MessageProcessorOrder create(@JsonProperty("changeVersion") long changeVersion,
+                                               @JsonProperty("classOrder") List<String> classOrder) {
+        return new AutoValue_MessageProcessorOrder(changeVersion, classOrder, Collections.emptySet());
```
joschi: This should call the other `create` method instead.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
kroepke: Well, this is actually a List, because we sort it from the set. I think this was also pre-existing code, which I haven't touched. The class was just moved.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
kroepke: Also pre-existing code

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
kroepke: I'll take a look, at some point the objectmappers were different, not sure if that ever changed.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
kroepke: This is probably a copy'n'waste error, I'll have a look.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorOrder.java
kroepke: :+1:

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
joschi: The real question is if it should be possible to add identical message filters to this collection. If yes, a list is fine, otherwise I'd use a sorted set.

garybot2:

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
kroepke: In theory yes, but I'd really like to keep the previous code the same, because this is what existing filter plugins use. This code was merely moved, I haven't actually touched it.

PR #1715, opened 2016-01-22
Delay authenticated fetches until session has been established.

dennisoelkers:
This is a very small but significant change to the `FetchProvider` which fetches resources via http and returns promises to its callers. The default fetch strategy is to do an authenticated call to the server using the previously established session token. When this is done too early or too late at a point where no session is established, they will fail. To prevent this, this change checks if a session is existing before doing call, returning a new promise that is waiting for a session to be properly established before resolving it.


edmundoa:
LGTM

PR #1716, opened 2016-01-22
Node journal

edmundoa:
Working on a proxied resource for the journal information, I saw that `JournalSummaryResponse` and `KafkaJournalConfiguration` were not playing nice when deserialising their JSON responses. Here I fix the issue and add back the journal configuration to the node overview page, but I'm not entirely sure I did it in the _right way_. I would appreciate if someone take a look. Thank you!

### DIFF
--- a/graylog2-plugin-interfaces/pom.xml
+++ b/graylog2-plugin-interfaces/pom.xml
@@ -110,7 +110,6 @@
         <dependency>
             <groupId>javax.validation</groupId>
             <artifactId>validation-api</artifactId>
-            <version>1.1.0.Final</version>
         </dependency>
     </dependencies>
 
--- a/graylog2-plugin-interfaces/pom.xml
+++ b/graylog2-plugin-interfaces/pom.xml
@@ -106,6 +106,12 @@
             <groupId>com.github.joschi</groupId>
             <artifactId>jadconfig</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <version>1.1.0.Final</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
@@ -27,9 +27,9 @@ import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.github.joschi.jadconfig.Parameter;
 import com.github.joschi.jadconfig.util.Size;
-import com.sun.istack.internal.NotNull;
 import org.joda.time.Duration;
 
+import javax.validation.constraints.NotNull;
 import java.io.File;
 import java.util.Objects;
 
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
@@ -22,6 +22,7 @@
  */
 package org.graylog2.plugin;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.github.joschi.jadconfig.Parameter;
@@ -31,6 +32,25 @@ import org.joda.time.Duration;
 import java.io.File;
 
 public class KafkaJournalConfiguration {
+
+    public KafkaJournalConfiguration() { }
+
+    @JsonCreator
+    public KafkaJournalConfiguration(@JsonProperty("directory") File messageJournalDir,
+                                     @JsonProperty("segment_size") long messageJournalSegmentSize,
+                                     @JsonProperty("segment_age") Duration messageJournalSegmentAge,
+                                     @JsonProperty("max_size") long messageJournalMaxSize,
+                                     @JsonProperty("max_age") Duration messageJournalMaxAge,
+                                     @JsonProperty("flush_interval") long messageJournalFlushInterval,
+                                     @JsonProperty("flush_age") Duration messageJournalFlushAge) {
+        this.messageJournalDir = messageJournalDir;
+        this.messageJournalSegmentSize = Size.bytes(messageJournalSegmentSize);
+        this.messageJournalSegmentAge = messageJournalSegmentAge;
+        this.messageJournalMaxSize = Size.bytes(messageJournalMaxSize);
+        this.messageJournalMaxAge = messageJournalMaxAge;
+        this.messageJournalFlushInterval = messageJournalFlushInterval;
+        this.messageJournalFlushAge = messageJournalFlushAge;
+    }
 
     @Parameter(value = "message_journal_dir", required = true)
     @JsonProperty("directory")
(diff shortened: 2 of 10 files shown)


graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
```diff
@@ -32,6 +33,25 @@
...
+
+    @JsonCreator
+    public KafkaJournalConfiguration(@JsonProperty("directory") File messageJournalDir,
+                                     @JsonProperty("segment_size") long messageJournalSegmentSize,
+                                     @JsonProperty("segment_age") Duration messageJournalSegmentAge,
+                                     @JsonProperty("max_size") long messageJournalMaxSize,
+                                     @JsonProperty("max_age") Duration messageJournalMaxAge,
+                                     @JsonProperty("flush_interval") long messageJournalFlushInterval,
+                                     @JsonProperty("flush_age") Duration messageJournalFlushAge) {
+        this.messageJournalDir = messageJournalDir;
```
joschi: This should probably be guarded by [`Objects#requireNonNull()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-) as it's a required value.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/KafkaJournalConfiguration.java
```diff
@@ -32,6 +33,25 @@
 
 public class KafkaJournalConfiguration {
 
+    public KafkaJournalConfiguration() { }
+
+    @JsonCreator
+    public KafkaJournalConfiguration(@JsonProperty("directory") File messageJournalDir,
```
joschi: Maybe add `@NotNull` here so that Jersey/JAX-RS can check the validity of the incoming data when deserializing it.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
```diff
@@ -0,0 +1,89 @@
...
+    @RequiresPermissions(RestPermissions.JOURNAL_READ)
+    @Path("{nodeId}")
+    public JournalSummaryResponse get(@ApiParam(name = "nodeId", value = "The id of the node to get message journal information.", required = true)
+                                      @PathParam("nodeId") String nodeId) throws IOException, NodeNotFoundException {
+        final Node targetNode = nodeService.byNodeId(nodeId);
+
+        final RemoteJournalResource remoteJournalResource = remoteInterfaceProvider.get(targetNode,
+                this.authenticationToken,
+                RemoteJournalResource.class);
+        final Response<JournalSummaryResponse> response = remoteJournalResource.get().execute();
```
joschi: This will always execute a remote call, even if it's the local node. Can we short-circuit it so that only remote nodes will be called over the network? (Would be nice to have this as a general functionality of `RemoteInterfaceProvider`.

Otherwise we're just doubling the HTTP requests for the local node for no good reason.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
```diff
@@ -0,0 +1,89 @@
...
+        if (response.isSuccess()) {
+            return response.body();
+        } else {
+            LOG.warn("Unable to get message journal information on node " + nodeId + ": " + response.message());
```
joschi: Could be using a [parameterized log statement](http://www.slf4j.org/faq.html#logging_performance) here

joschi: Should this throw a `WebApplicationException` in the else-branch in case of a failure?

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import java.io.IOException;
+
+@RequiresAuthentication
+@Api(value = "Cluster", description = "Journal information of any nodes in the cluster")
```
joschi: This leads to a naming conflict in the Swagger API browser because there's already another resource claiming the "Cluster" category. Maybe use "Cluster/Journal" instead?

edmundoa: Didn't see that, will rename it as you suggested :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
edmundoa: This will affect many other proxied resources. I think it would make sense taking care of that separately, and also use the occasion to make `ProxiedResource` extend `RestResource`. If it sounds good, I can take care of that once this is merged.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
edmundoa: Good catch! Will do so.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
joschi: :white_check_mark:

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+        final RemoteJournalResource remoteJournalResource = remoteInterfaceProvider.get(targetNode,
+                this.authenticationToken,
+                RemoteJournalResource.class);
+        final Response<JournalSummaryResponse> response = remoteJournalResource.get().execute();
+        if (response.isSuccess()) {
+            return response.body();
+        } else {
+            LOG.warn("Unable to get message journal information on node {}: {}", nodeId, response.message());
+            throw new WebApplicationException(response.message(), response.code());
```
joschi: While it might be useful to know the HTTP response code of the proxied request, I think this method should return either HTTP status 500 (Internal Server Error) or 502 (Bad Gateway) in case of an error.

joschi: Bad Gateway – Internetkurort

graylog2-plugin-interfaces/pom.xml
```diff
@@ -106,6 +106,12 @@
             <groupId>com.github.joschi</groupId>
             <artifactId>jadconfig</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <version>1.1.0.Final</version>
```
joschi: Please add a corresponding entry into the `dependencyManagement` section of `graylog2-parent`, so that the dependency is being pulled in using the same version everywhere.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterJournalResource.java
edmundoa: 502 it is!

joschi:
LGTM. :+1:

PR #1719, opened 2016-01-25
Fix message iteration in MessageFilterChainProcessor

joschi:
The implementation of `MessageFilterChainProcessor` has only been running messages through message filters until the first filter succeeded and didn't filter out the message (i. e. the default case) due to the early `return` statement inside the inner loop (iterating over the message filters).


kroepke:
lgtm, thanks for the quick fix!

PR #1721, opened 2016-01-25
Sanitization of message field names

joschi:
Beginning with Elasticsearch 2.0.0, field names must not contain dots ('.') anymore.

In order to make migration to Elasticsearch 2.x simpler for Graylog users, invalid characters like dots will be automatically replaced in field names with a configurable string (default: `_`).

Refs #1667.

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
@@ -38,6 +38,7 @@ import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import java.net.InetAddress;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


bernd:
Why did you implement this as a message filter instead of doing it in `Message#addField()`? Even though the filter has priority `Integer.MAX_VALUE`, this might become a problem in some cases.

joschi:
@bernd The main reason to implement it as a filter is that the replacement string for dots should be configurable. If it was part of the `Message` class (as is without a major refactoring), that wouldn't be possible.

I understand the disadvantages of using a message filter for this but it seemed worth the tradeoff.

joschi:
@bernd Moved the sanitization of message field names into `ServerProcessBufferProcessor`, as discussed.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -65,7 +66,7 @@
     public static final String FIELD_LEVEL = "level";
     public static final String FIELD_STREAMS = "streams";
 
-    private static final Pattern VALID_KEY_CHARS = Pattern.compile("^[\\w\\.\\-@]*$");
+    private static final Pattern VALID_KEY_CHARS = Pattern.compile("^[\\w\\.\\-@/]*$");
```
bernd: I think we can remove the `/` as allowed character again.

joschi: The question is rather which characters we want to allow additionally. In the end, this pattern controls which replacement characters can be used in the first place.

bernd: I think we should just remove the `/` right know and not allow more characters at the moment. Because we do not know yet how other characters might behave when searching, filtering etc.

I also still see this as a temporary solution that might not be in the final 2.0 because there are way to many ways to still introduce `.` character in field names. In the end this needs to be done when adding fields to the message object.
Until we are there, this approach is okay in my opinion.

joschi: :+1:

kroepke:
I'm not entirely clear whether the current approach is the best to take.
Preferably I would make it impossible to add fields with `.` in them, because that would prevent from saving them anyhow.

While a MessageProcessor for this purpose has the benefit of being able to turn it off (by I'd prefer to expose the functionality in the UI/API rather than another config variable), I think we should prevent almost silent errors that would occur by allowing to set fields with this character.

So before merging this I'd like to discuss the various options again.

joschi:
@kroepke We've had a discussion about the various options in https://github.com/Graylog2/graylog2-server/pull/1667#issuecomment-173179025. Feel free to join in.

joschi:
Closed in favor of #1667.

PR #1724, opened 2016-01-26
Make it possible to get a cluster system job by job ID

bernd:
- Add new /cluster/jobs/{jobId} REST endpoint.
- Add getJob() function to SystemJobsStore.
- Add new getJob() action to SystemJobsActions.


edmundoa:
LGTM :+1:

PR #1726, opened 2016-01-26
convert legacy TimeRange interface into auto value class

kroepke:
in order to properly serialize time range objects in the API using Jackson we want to have proper subtypes
also unifies some of the handling and removes the enum which served no actual purpose apart from being a String


graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesConfigBuilder.java
```diff
@@ -74,6 +74,6 @@ public SearchesConfigBuilder setSorting(Sorting sorting) {
     }
 
     public SearchesConfig build() {
-        return new SearchesConfig(query, filter, fields, range, limit, offset, sorting);
+        return null;
```
joschi: This will lead to strange errors if the `SearchesConfigBuilder` class is being used somewhere. Can't we just remove the class?

kroepke: Yes, I simply forgot to remove it.

PR #1729, opened 2016-01-27
Updated Random Message Generator with time fields

zexxon:
The purpose of the code change is to add time fields for event deduplication, first seen, last seen and time bucket association.


joschi:
Thanks!

I've made some minor changes to your first implementation (e. g. using Joda-Time instead of the old Java Date/Time API) and extended the `FakeHttpRawMessageGenerator` class to fill additional fields for all requests and not just for `GET` requests.

zexxon:
Awesome! Thanks.

PR #1733, opened 2016-01-28
Add IndicesStore#listOpen() and the corresponding action

bernd:


dennisoelkers:
Why not just use the `list` action and filter the returned indices based on their attributes? The data is already there and this would prevent an additional round trip when the IndicesStore is already initialized.

bernd:
Hmm, I think I misunderstood the `all` field. I thought I had to do some massive filtering in the client, but it seems like I don't have to. Thanks for the heads-up.

PR #1734, opened 2016-01-28
Use chaining/delegating class loader for plugin classes

joschi:
In order to use [polymorphic deserialization with Jackson](http://wiki.fasterxml.com/JacksonPolymorphicDeserialization), the `ObjectMapper` must be able to access the classes loaded from plugins which live in their own class loaders.

The `ChainingClassLoader` class loader implementation queries all registered class loaders sequentially until the class/resource has been found in one of the child class loaders.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
@@ -1,0 +1,124 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.shared.plugins;
+
+import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ChainingClassLoader extends ClassLoader {
+    public static final Logger LOG = LoggerFactory.getLogger(ChainingClassLoader.class);
+
+    private final List<ClassLoader> classLoaders = new CopyOnWriteArrayList<>();
+
+    public ChainingClassLoader(ClassLoader parent) {
+        classLoaders.add(parent);
+    }
+
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+        Class<?> cls = null;
+        for (ClassLoader classLoader : classLoaders) {
+            try {
+                cls = classLoader.loadClass(name);
+            } catch (ClassNotFoundException e) {
+                LOG.trace("Class " + name + " not found", e);
+            }
+
+            if (cls != null) {
+                break;
+            }
+        }
+
+        if (cls == null) {
+            throw new ClassNotFoundException("Class " + name + " not found.");
+        }
+
+        return cls;
+    }
+
+    @Override
+    public URL getResource(String name) {
+        URL url = null;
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ChainingClassLoader extends ClassLoader {
+    public static final Logger LOG = LoggerFactory.getLogger(ChainingClassLoader.class);
+
+    private final List<ClassLoader> classLoaders = new CopyOnWriteArrayList<>();
+
+    public ChainingClassLoader(ClassLoader parent) {
+        classLoaders.add(parent);
```
kroepke: Could we use `this(parent);` here and call the super methods in the other methods below to get the default behavior when no plugins are there? Then the parent would not have to live in the COW array list.

graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+        Class<?> cls = null;
+        for (ClassLoader classLoader : classLoaders) {
+            try {
+                cls = classLoader.loadClass(name);
```
kroepke: Should this include the `resolve` parameter, too?

graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        if (stream == null && LOG.isTraceEnabled()) {
+            LOG.trace("Resource " + name + " not found.");
+        }
+
+        return stream;
+    }
+
+    public List<ClassLoader> getClassLoaders() {
```
kroepke: this is only visible for testing, right?
Can we make it package local in that case?

graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
```diff
@@ -87,8 +89,9 @@ public PluginLoader(File pluginDir) {
             } else {
                 try {
                     LOG.debug("Loading <" + jar.getAbsolutePath() + ">");
-                    final ClassLoader pluginClassLoader = new URLClassLoader(new URL[]{jar.toURI().toURL()}, classLoader);
-                    final ServiceLoader<Plugin> pluginServiceLoader = ServiceLoader.load(Plugin.class, pluginClassLoader);
+                    final ClassLoader pluginClassLoader = URLClassLoader.newInstance(new URL[]{jar.toURI().toURL()});
+                    classLoader.addClassLoader(pluginClassLoader);
+                    final ServiceLoader<Plugin> pluginServiceLoader = ServiceLoader.load(Plugin.class, classLoader);
```
kroepke: Can this create problems if we ever try to unload plugins?
The ChainingClassLoader owns all plugin classloaders, which would prevent them from being garbage collected, as opposed to the other way around when plugin classloaders get a parent.

Since we need the ability to see classes from other plugins in a plugin as we create more extension points, making the chaining loader accessible is right, but I would prefer to pass it as a parent loader to the URLClassLoader and making the ChainingLoader smarter in not delegating to the calling loader (i.e. to prevent infinite recursion).

That should allow clean unloading if we ever implement it by simply removing a plugin's classloader from the ChainingClassLoader (and additionally cleaning the class usages, of course, but that's out of scope here).

graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
joschi: It should but unfortunately the [`ClassLoader#loadClass(String, boolean)`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#loadClass-java.lang.String-boolean-) method is `protected` and can't be called here.

graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
joschi: I've opted to use this variant (with the parent class loader being the first item in the class loaders list) to simplify the overridden methods. Otherwise they'd need two code paths: One calling the corresponding method in `super` and one iterating over the list of class loaders.

graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
kroepke: We'll address this issue later once it arises.

graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
kroepke: ok, fair enough

kroepke:
lgtm, we'll iterate over the review points later if we need to.

PR #1735, opened 2016-01-28
Improvements in proxied resources

edmundoa:
Fix some of the issues seen on proxied resources:

- `ProxiedResource` class extends now `RestResource` to provide common functionality available in other REST resources
- Ensure all proxied resources inherit from `ProxiedResource`
- Throw `WebApplicationException` when accessing a resource in another node, and its response is not successful

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterInputStatesResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterInputStatesResource.java
@@ -96,7 +96,7 @@ public class ClusterInputStatesResource extends ProxiedResource {
                             LOG.warn("Unable to fetch input states from node {}: {}", entry.getKey(), response.message());
                         }
                     } catch (IOException e) {
-                        LOG.warn("Unable to fetch input states from node {}: {}", entry.getKey(), e);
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterInputStatesResource.java
```diff
@@ -100,10 +93,10 @@ public ClusterInputStatesResource(NodeService nodeService,
                         if (response.isSuccess()) {
                             return response.body().states();
                         } else {
-                            LOG.warn("Unable to fetch input states from node " + entry.getKey() + ": " + response.message());
+                            LOG.warn("Unable to fetch input states from node {}: {}", entry.getKey(), response.message());
                         }
                     } catch (IOException e) {
-                        LOG.warn("Unable to fetch input states from node " + entry.getKey() + ": ", e);
+                        LOG.warn("Unable to fetch input states from node {}: {}", entry.getKey(), e);
```
joschi: Leaving out the last placeholder (`{}`) in the message would print the exception's stack trace instead of just calling the `toString()` method: http://www.slf4j.org/faq.html#paramException

edmundoa: :heavy_check_mark:

PR #1738, opened 2016-01-29
[WEB] Removing shared module, passing Store/ActionsProvider in context.

dennisoelkers:
This allows us to share singleton instances of stores and actions
throughout all components of the web interface and all plugins.

The shared bundle has been removed, because it prevents hot reloading of
shared components. Components which are reused by plugins are bundled
into plugins instead of referenced from the shared bundle. This has the
downside that plugins need rebundling when components of the web
interface are modified (because of bugfixing or enhancements) to benefit from their improvements, but it
also makes them more resilient to API changes.


graylog2-web-interface/src/injection/StoreProvider.jsx
```diff
@@ -0,0 +1,50 @@
...
+      UniversalSearch: () => require('stores/search/UniversalSearchStore'),
+      Session: () => require('stores/sessions/SessionStore'),
+      SystemProcessing: () => require('stores/system-processing/SystemProcessingStore'),
+      SystemShutdown: () => require('stores/system-shutdown/SystemShutdownStore'),
+      System: () => require('stores/system/SystemStore'),
+      SystemJobs: () => require('stores/systemjobs/SystemJobsStore'),
+      SystemMessages: () => require('stores/systemmessages/SystemMessagesStore'),
+      Focus: () => require('stores/tools/FocusStore'),
+      CurrentUser: () => require('stores/users/CurrentUserStore'),
+      Startpage: () => require('stores/users/StartpageStore'),
```
edmundoa: Could you please also add the `JournalStore`?

edmundoa:
Other than that missing store, I guess we could also remove `graylog2-web-interface/karma.shared.js` and `graylog2-web-interface/webpack.shared.js`, now that the shared bundle is gone.

graylog2-web-interface/src/injection/StoreProvider.jsx
dennisoelkers: I am wondering why I missed it. Was it created just recently?

dennisoelkers:
Yes, good point! :+1:

dennisoelkers:
Removed `webpack.shared.js` in 599d7c1

edmundoa:
Oh, I see. Then it is good to go I think! :+1:

PR #1741, opened 2016-01-31
Convert LDAP settings pages to React

### LINKED ISSUE #1621: Convert LDAP settings page to react

edmundoa:
This PR contains all necessary changes to make the LDAP settings and LDAP group mappings pages work again.

Sorry about the long list of commits, somehow got out of hand :confused: 

Fixes #1621.


graylog2-web-interface/public/stylesheets/graylog2.less
```diff
@@ -1312,6 +1312,21 @@ td.limited {
   padding-top: 1px;
 }
 
+.ldap-test-login-result {
```
dennisoelkers: Maybe it makes sense to defines these classes in a separate style and load/bundle it using the style-loader? This way we would start to bundle styles together with the components that use them and we could decouple/decompose the big global less file.

graylog2-web-interface/src/components/users/LdapComponent.jsx
```diff
@@ -0,0 +1,426 @@
...
+      ldapUri: undefined,
+      roles: undefined,
+    };
+  },
+
+  componentDidMount() {
+    RolesStore.loadRoles().then(roles => {
+      this.setState({
+        roles: roles
+          .sort((r1, r2) => r1.name.localeCompare(r2.name))
```
dennisoelkers: Could you please extract the logic processing the roles into a separate function? This way we would have to change less when we change the RolesStore to reflux.

graylog2-web-interface/src/pages/LdapPage.jsx
```diff
@@ -0,0 +1,34 @@
...
+import DocsHelper from 'util/DocsHelper';
+
+import CurrentUserStore from 'stores/users/CurrentUserStore';
+
+import { PageHeader } from 'components/common';
+import DocumentationLink from 'components/support/DocumentationLink';
+import LdapComponent from 'components/users/LdapComponent';
+
+const LdapPage = React.createClass({
+  mixins: [Reflux.connect(CurrentUserStore)],
```
dennisoelkers: Why do we need this store in the page?

graylog2-web-interface/src/stores/users/LdapGroupsStore.ts
```diff
@@ -1,49 +1,47 @@
...
+import jsRoutes = require('routing/jsRoutes');
+const fetch = require('logic/rest/FetchProvider').default;
 const UserNotification = require('util/UserNotification');
 const URLUtils = require('util/URLUtils');
 
+const LdapGroupsStore = {
+  loadGroups(): Promise<string[]> {
+    const promise = fetch('GET', URLUtils.qualifyUrl(jsRoutes.controllers.LdapController.apiGroups().url));
+    promise.catch(error => {
+      if (error.additional.status !== 404) {
```
dennisoelkers: What happens when the result is 404?

graylog2-web-interface/src/stores/users/LdapGroupsStore.ts
```diff
@@ -1,49 +1,47 @@
...
+          'Could not load LDAP group list');
+      }
+    });
+
+    return promise;
+  },
+  loadMapping(): Promise<Object> {
+    const promise = fetch('GET', URLUtils.qualifyUrl(jsRoutes.controllers.LdapController.apiGroupsMapping().url));
+    promise.catch(error => {
+      if (error.additional.status !== 404) {
```
dennisoelkers: Same here, what happens when the result code is 404?

graylog2-web-interface/src/components/users/LdapComponent.jsx
```diff
@@ -0,0 +1,426 @@
...
+
+  _setAdditionalDefaultGroups(rolesString) {
+    // only keep non-empty entries
+    const roles = rolesString.split(',').filter((v) => v !== '');
+    this._setSetting('additional_default_groups', roles);
+  },
+
+  _saveSettings(event) {
+    event.preventDefault();
+    LdapStore.update(this.state.ldapSettings);
```
dennisoelkers: Why is the store called directly and not decoupled via an action?

graylog2-web-interface/src/components/users/TestLdapConnection.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    // Reset connection status if ldapSettings changed
+    if (JSON.stringify(this.props.ldapSettings) !== JSON.stringify(nextProps.ldapSettings)) {
+      this.setState({serverConnectionStatus: {}});
+    }
+  },
+
+  _testServerConnection() {
+    LdapStore.testServerConnection(this.props.ldapSettings)
```
dennisoelkers: Why is the store called directly and not decoupled via an action?

dennisoelkers:
I know that this was a huge change and very tedious to implement and I am very glad you did that, but wouldn't it make sense to rewrite the Stores that are touched in this PR to Reflux stores? Or is it too much work for this already very big changeset?

graylog2-web-interface/src/components/users/LdapComponent.jsx
edmundoa: Makes sense :ballot_box_with_check:

graylog2-web-interface/src/pages/LdapPage.jsx
edmundoa: As far as I know, we don't. Cleaning it up!

graylog2-web-interface/src/stores/users/LdapGroupsStore.ts
edmundoa: Good question, the check was already in the store and I simply adapted it. I don't see anything special in the backend code, so I think we can remove it.

graylog2-web-interface/src/stores/users/LdapGroupsStore.ts
edmundoa: Same as the other one.

graylog2-web-interface/src/components/users/LdapComponent.jsx
edmundoa: No good reason, fixed it :+1:

graylog2-web-interface/public/stylesheets/graylog2.less
edmundoa: Done! For now I put it in the same level as the component, as I think it's easier to see when files move or disappear, but I'm open for feedback on that.

edmundoa:
As I mentioned in Slack, the reason for not migrating the stores is that they didn't need much work to fix them (basically changing the promises and routes), and the list of changes was indeed large. I will migrate the `LdapGroupsStore`, as it's only used in the files already touched in this PR, but will leave the `RolesStore` as is for now.

PR #1744, opened 2016-02-01
Rename "server_id" attribute to "node_id" in /system

joschi:
For consistency reasons, the `server_id` attribute in the JSON response of the `/system` resource shall be named `node_id`.


bernd:
LGTM :+1:

PR #1745, opened 2016-02-01
Remove LocalMetricsInput and LocalMetricsTransport

joschi:
The `LocalMetricsInput` (and subsequently `LocalMetricsTransport`) hasn't been working in a long time and should be removed to put them out of their misery instead of being dragged along.


bernd:
LGTM :+1:

PR #1746, opened 2016-02-01
Do not index GELF "version" field into Elasticsearch

joschi:
The `version` field of GELF messages doesn't carry any information that needs to be indexed into Elasticsearch, so remove this field when building a proper message out of the raw messages from the message journal.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -167,7 +167,10 @@ public class GelfCodec extends AbstractCodec {
             final Map.Entry<String, JsonNode> entry = fields.next();
 
             String key = entry.getKey();
-            final JsonNode value = entry.getValue();
+            // Do not index useless GELF "version" field.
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
```diff
@@ -179,6 +179,11 @@ public Message decode(@Nonnull final RawMessage rawMessage) {
                 continue;
             }
 
+            // Do not index useless GELF "version" field.
+            if ("version".equals(key)) {
+                continue;
+            }
+
```
bernd: This needs to be done before removing the `_` character from additional fields. ([here](https://github.com/Graylog2/graylog2-server/blob/do-not-index-gelf-version-field/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java#L171))

Otherwise a `version` cannot be set at all.

joschi: Good catch, thanks!

bernd:
LGTM :+1:

PR #1754, opened 2016-02-02
Show overlay when server connection is not working

dennisoelkers:
Before this change, a non-working server connection was indicated only
by javascript errors in the browser console and neverending spinners.
With this change, whenever requests are unsuccessfull in a fundamental
way, or the occasional polling for the server does not succeed, an
overlay will be rendered indicating the error.

When the server connection is restored, no browser refresh is necessary, the web interface will restore to the previous state automatically.

Right now the design is very basic and questionable, additional styling
and displaying of the error can be added later.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/CORSFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/CORSFilter.java
@@ -29,7 +29,7 @@ public class CORSFilter implements ContainerRequestFilter, ContainerResponseFilt
     public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
         // we have already added the necessary headers for OPTIONS requests below
         if (requestContext.getRequest().getMethod().equalsIgnoreCase("options")) {
-            if(Response.Status.Family.familyOf(responseContext.getStatus()) == Response.Status.Family.INFORMATIONAL) {
+            if(Response.Status.Family.familyOf(responseContext.getStatus()) == Response.Status.Family.SUCCESSFUL) {
                 return;
             }
             responseContext.setStatus(Response.Status.NO_CONTENT.getStatusCode());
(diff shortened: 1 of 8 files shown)


edmundoa:
Will make it use the `web` label, I think it's nicer than adding tags in the title :)

graylog2-server/src/main/java/org/graylog2/shared/rest/CORSFilter.java
```diff
@@ -29,7 +29,11 @@
     public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
         // we have already added the necessary headers for OPTIONS requests below
         if (requestContext.getRequest().getMethod().equalsIgnoreCase("options")) {
-            return;
+            if(Response.Status.Family.familyOf(responseContext.getStatus()) == Response.Status.Family.INFORMATIONAL) {
+                return;
+            }
+            responseContext.setStatus(Response.Status.NO_CONTENT.getStatusCode());
+            responseContext.setEntity("");
         }
```
edmundoa: This is not working as expected, as there are requests that will get the CORS HTTP headers set twice. Chrome was complaining about that, and the result was that the web interface was not loading even when the server was running. I think we should change it into this:

```
if (Response.Status.Family.familyOf(responseContext.getStatus()) != Response.Status.Family.INFORMATIONAL) {
    responseContext.setStatus(Response.Status.NO_CONTENT.getStatusCode());
    responseContext.setEntity("");
}
return;
```

dennisoelkers: The `return` is there to prevent double response headers for 20x responses. From what I've seen, Jersey is not adding the headers for non-20x responses. Which responses did you encounter that had double headers?

UPDATE: I just noticed that it's the wrong Response family I am using to compare. Fixing.

edmundoa: Ah, that makes sense. I was wondering why checking informational responses :) 
The requests I was talking about were 20xs, can't remember if it also happened with other responses but will double check now.

edmundoa:
All changes are looking good to me. @dennisoelkers I did some style changes and followed your advise regarding showing more details. If you like the changes feel free to merge :smile:

PR #1755, opened 2016-02-02
Upgrade to Elasticsearch 2.2.0

joschi:
https://www.elastic.co/guide/en/elasticsearch/reference/2.2/release-notes-2.2.0.html


graylog2-server/src/main/java/org/graylog2/indexer/elasticsearch/GlobalTimeoutClient.java
```diff
@@ -84,7 +84,15 @@ public GlobalTimeoutAction(final Action<Request, Response, RequestBuilder> actio
         @Override
         public TransportRequestOptions transportOptions(Settings settings) {
             final TransportRequestOptions result = super.transportOptions(settings);
-            return result.timeout() == null ? result.withTimeout(timeout) : result;
+            if (result.timeout() == null) {
```
bernd: Why this change?

joschi: `TransportRequestOptions#withTimeout()` has been removed in Elasticsearch 2.2.0 and we now need to use the corresponding builder which looks quite ugly using a conditional expression.

bernd:
I am getting the following error when restarting Elasticsearch when Graylog is running.

https://gist.github.com/bernd/5107d3bf3c4b3d423df1

graylog2-server/src/main/java/org/graylog2/indexer/elasticsearch/GlobalTimeoutClient.java
bernd: :+1:

joschi:
@bernd The startup process of Elasticsearch 2.2.x seems to have changed in a way that the cluster health state is flapping which leads to `IndexChangeMonitor` firing `IndicesClosedEvent` and `IndicesReopenedEvent` although no indices were actually closed or reopened. This also happens when using an older Elasticsearch 2.x embedded node in Graylog and is not specific to upgrading the internal ES node to version 2.2.0.

This being said, I've added specific handling of `IndexClosedException` in the `MongoIndexRangeService#handleIndexReopening()` method which only outputs the fact on DEBUG level and won't pollute the logs during normal operations.

PR #1757, opened 2016-02-03
Turn streams List into Set

kroepke:
To support iteratively building up the set of assigned streams for a message we don't want filters to overwrite the entire collection all the time, thus `setStreams()` is deprecated.

Instead expose `addStream()`, `addStreams()`, `removeStream()` and turn the internal field into a `Set<Stream>`.

`getStreamIds()` would cause NPEs if the streams field wasn't set, which is counter-intuitive. In fact, its behavior is plain wrong, but I haven't changed it in this PR yet.

To allow other `MessageProcessor` instances to change the stream assignments, don't let `StreamMatcherFilter` blindly overwrite all streams, but add to the set.

### DIFF
--- a/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
@@ -41,7 +41,6 @@ import org.slf4j.LoggerFactory;
 
 import java.net.InetAddress;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -347,16 +349,32 @@ public boolean hasField(String field) {
...
-        this.streams = Lists.newArrayList(streams);
+        this.streams = Sets.newHashSet(streams);
     }
 
     public List<Stream> getStreams() {
-        return this.streams;
+        return Lists.newArrayList(this.streams);
+    }
+
+    public void addStream(Stream stream) {
```
joschi: Please add some unit tests for those new methods.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/MessageCollection.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.common.collect.Iterators;
+
+import java.util.Iterator;
+
+public class MessageCollection implements Messages  {
+
+    private final ImmutableList<Message> messages;
+
+    public MessageCollection(Iterable<Message> other) {
+        messages = ImmutableList.copyOf(other);
```
joschi: This might cause a performance hit. Can't we just store the `Iterable<Message>` (or the `Iterator<Message>`) directly? [`Iterators#filter(java.util.Iterator, com.google.common.base.Predicate)`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterators.html#filter(java.util.Iterator,%20com.google.common.base.Predicate)) returns an [`UnmodifiableIterator`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/UnmodifiableIterator.html) anyway.

kroepke: Yeah, I expect this class to change completely over the next few iterations. I simply needed a placeholder for now.

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/Message.java
kroepke: done

graylog2-plugin-interfaces/src/main/java/org/graylog2/plugin/MessageCollection.java
kroepke: In fact, probably everything around the `Message` class likely needs to change to support making it immutable.

joschi: :white_check_mark:

PR #1767, opened 2016-02-04
Returning NotFoundException in InputServiceImpl when id is invalid.

### LINKED ISSUE #1718: Passing invalid input id to /system/inputs/{id} should not return 500
`GET http://localhost:12900/system/inputs/undefined` returns `500 (Internal Server Error)` because it doesn't catch the ObjectId conversion error.

This should return 404 instead to not cause callers to assume a hard error.

dennisoelkers:
Closes #1718.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
@@ -142,6 +142,9 @@ public class InputServiceImpl extends PersistedServiceImpl implements InputServi
 
     @Override
     public Input find(String id) throws NotFoundException {
+        if (!ObjectId.isValid(id)) {
+            throw new NotFoundException("Input id <" + id + "> is invalid!");
+        }
         final DBObject o = get(org.graylog2.inputs.InputImpl.class, id);
         if (o == null) {
             throw new NotFoundException("Input <" + id + "> not found!");


graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
```diff
@@ -142,6 +142,9 @@ public Input create(Map<String, Object> fields) {
 
     @Override
     public Input find(String id) throws NotFoundException {
+        if (!ObjectId.isValid(id)) {
+            throw new NotFoundException("Input id <" + id + "> is invalid!");
```
joschi: IMHO this should be an `IllegalArgumentException`.

dennisoelkers: I don't think so. This should be a pretty high-level abstraction of the Input persistence. From what the consumer of it needs to know is that there is no entity with that id, not that the id is misconstructed, which does not help the consumer of this method in any way. The id which is invalid for this implementation could be perfectly valid for a different implementation anyhow.

joschi: `InputService` and its implementation are actually quite technical. The high-level access is channeled through the classes calling the methods of `InputService`, e. g. `InputsResource`.

It would even be better to have a proper JAX-RS and Jackson serializer and deserializer for `ObjectId` which would respond with an appropriate HTTP status code (e. g. 400 Bad Request) if no valid `ObjectId` was provided. This way it wouldn't only be "point repair" in `InputService` but could be used everywhere.

dennisoelkers: I agree to the second point, but not to the first. From a separation of responsibility, `InputsResource` shouldn't bother with the id format, but if the entity is present or not. Therefore the way the code is currently structured throwing this exception is perfectly fine. Having a proper deserializer for `ObjectId` later on - sure. But not in the scope of this fix.

joschi: But even then I don't like the mix of different exception semantics here. An invalid ID is definitively not "not found" but simply an invalid ID for which there is the `IllegalArgumentException` in Java. Even the message of the `NotFoundException` in your changeset says that.

I still think this should be an IAE (which then could be automatically mapped to an HTTP 400 response using a JAX-RS exception mapper).

PR #1768, opened 2016-02-04
Fix Accept header for non-json endpoints in swagger ui

### LINKED ISSUE #723: API browser error for load balancer status
Trying the `/system/lbstatus` results in an 406 error because of a wrong content type in the request.

g2-api-browser-error

`text/plain` is needed, not `application/json`.

bernd:
Fixes #723

### DIFF
--- a/graylog2-server/src/main/resources/swagger/lib/swagger.js
+++ b/graylog2-server/src/main/resources/swagger/lib/swagger.js
@@ -1004,13 +1004,16 @@
       }
       responseContentType = null;
       if (this.type === "POST" || this.type === "GET" || this.type === "PATCH") {
+        responseContentType = "application/json";
+
         if (this.opts.responseContentType) {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/resources/swagger/lib/swagger.js
```diff
@@ -1004,10 +1004,15 @@
       }
       responseContentType = null;
       if (this.type === "POST" || this.type === "GET" || this.type === "PATCH") {
+        responseContentType = "application/json";
```
joschi: Couldn't this be pulled into line 1005? The else branch overwrites it with `null` anyway, so this could simply be the initial value of `responseContentType`.

bernd: I removed the else branch instead.

joschi:
LGTM.

PR #1774, opened 2016-02-05
Handle invalid IDs in InputsResource

joschi:
This is the second take on solving #1718 after the discussion in #1767.

This PR introduces an exception mapper which translates `InvalidArgumentException`s into HTTP responses with status code 400. Additionally it introduces quick validity checks to the methods in `InputServiceImpl` which are handling MongoDB `ObjectId`s as a precondition for those methods.

Contrary to https://github.com/Graylog2/graylog2-server/pull/1767#discussion_r52009588, this change set doesn't add serialization/deserialization for `ObjectId` in Jackson or Jersey because at least the JAX-RS resource classes shouldn't be "polluted" with details of the underlying database (and e. g. use `ObjectId` directly as `@PathParam`).

Fixes #1767.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java
@@ -17,7 +17,6 @@
 package org.graylog2.inputs;
 
 import com.google.common.base.Optional;
-import com.google.common.base.Predicate;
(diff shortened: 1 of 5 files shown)


bernd:
I am closing this in favor of #1767 because I think returning a 404 is more correct and also because it is the smaller diff for 2.0.

We might want to revisit this after 2.0.

PR #1778, opened 2016-02-08
Upgrade to Mongo Java Driver 3.2.1 and MongoJack 2.5.1

joschi:
This PR upgrades Graylog to Mongo Java Driver 3.2.1 and MongoJack 2.5.1.

This raises the system requirements for Graylog to MongoDB 2.4.x or newer but opens up the opportunity to use "advanced" features like aggregations, that have existed for quite some time in MongoDB.

Driver compatibility chart: https://docs.mongodb.org/ecosystem/drivers/driver-compatibility-reference/#java-driver-compatibility

* Ubuntu 14.04 LTS (trusty): http://packages.ubuntu.com/trusty/mongodb
* Debian Linux (jessie): https://packages.debian.org/jessie/mongodb
* Debian Linux (wheezy-backports): https://packages.debian.org/wheezy-backports/mongodb
* CentOS 6 and 7: No MongoDB packages (needed to use the official MongoDB repositories anyway)
* Any other major Linux distribution: https://docs.mongodb.org/manual/administration/install-on-linux/#recommended

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
@@ -86,16 +86,18 @@ public class ClusterEventPeriodical extends Periodical {
 
         DBCollection coll = db.getCollection(COLLECTION_NAME);
 
+        /* FIXME: Not supported by Fongo
         if (coll.isCapped()) {
             LOG.warn("The \"{}\" collection in MongoDB is capped which will cause problems. Please drop the collection.", COLLECTION_NAME);
         }
+        */
 
         coll.createIndex(DBSort
                 .asc("timestamp")
                 .asc("producer")
                 .asc("consumers"));
 
-        coll.setWriteConcern(WriteConcern.FSYNCED);
+        coll.setWriteConcern(WriteConcern.JOURNALED);
 
         return coll;
     }
@@ -180,7 +182,7 @@ public class ClusterEventPeriodical extends Periodical {
         final ClusterEvent clusterEvent = ClusterEvent.create(nodeId.toString(), className, event);
 
         try {
-            final String id = dbCollection.save(clusterEvent, WriteConcern.FSYNCED).getSavedId();
+            final String id = dbCollection.save(clusterEvent, WriteConcern.JOURNALED).getSavedId();
             LOG.debug("Published cluster event with ID <{}> and type <{}>", id, className);
         } catch (MongoException e) {
             LOG.error("Couldn't publish cluster event of type <" + className + ">", e);
--- a/graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
@@ -86,12 +86,6 @@ public class ClusterEventPeriodical extends Periodical {
 
         DBCollection coll = db.getCollection(COLLECTION_NAME);
 
-        /* FIXME: Not supported by Fongo
-        if (coll.isCapped()) {
-            LOG.warn("The \"{}\" collection in MongoDB is capped which will cause problems. Please drop the collection.", COLLECTION_NAME);
-        }
-        */
-
         coll.createIndex(DBSort
                 .asc("timestamp")
                 .asc("producer")
--- a/UPGRADING.asciidoc
+++ b/UPGRADING.asciidoc
@@ -32,11 +32,20 @@ One of the most important breaking changes in Elasticsearch 2.x is that https://
 
 Using the https://github.com/elastic/elasticsearch-migration[Elasticsearch Migration Plugin] might help to highlight some potential pitfalls if an existing Elasticsearch 1.x cluster should be upgraded to Elasticsearch 2.x.
 
+
+### MongoDB
+
+Graylog 2.x requires MongoDB 2.4 or newer. We recommend using MongoDB 3.x and the https://docs.mongodb.org/v3.2/core/wiredtiger/[WiredTiger storage engine].
+
+When upgrading from MongoDB 2.0 or 2.2 to a supported version, make sure to read the https://docs.mongodb.org/manual/release-notes/[Release Notes] for the particular version.
+
+
 ### Log4j 2 migration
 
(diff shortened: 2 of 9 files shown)


bernd:
Please add this to [UPGRADING.asciidoc](https://github.com/Graylog2/graylog2-server/blob/master/UPGRADING.asciidoc).

graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
```diff
@@ -86,16 +86,18 @@ static DBCollection prepareCollection(final MongoConnection mongoConnection) {
 
         DBCollection coll = db.getCollection(COLLECTION_NAME);
 
+        /* FIXME: Not supported by Fongo
```
bernd: What's the plan with this? Can we just drop the warning? I don't know why it's a problem when the collection is capped.. :wink:

joschi: This is actually an artifact of the first implementation of the Cluster Event feature which had been using a capped collection to be able to use a tailable cursor and automatically get rid of "old" events. The problem was, that a capped collection only allows fixed size documents, so updating the events document (e. g. the list of nodes that have already processed the event) wasn't possible.

This never made it into a release, so I think we can safely remove the check (I'd do this in this PR if we reach consensus about it). The alternatives would be to either disable the related tests or extend Fongo to support capped collections again (they were supported in the past).

See 8d4765a3c831260e18f8a5d4935b4025846cc573 for details.

graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
```diff
@@ -86,16 +86,18 @@ static DBCollection prepareCollection(final MongoConnection mongoConnection) {
...
         }
+        */
 
         coll.createIndex(DBSort
                 .asc("timestamp")
                 .asc("producer")
                 .asc("consumers"));
 
-        coll.setWriteConcern(WriteConcern.FSYNCED);
+        coll.setWriteConcern(WriteConcern.JOURNALED);
```
bernd: We are using `FSYNCED` in some other places as well. Should we switch those as well? 
```
$ git grep -F WriteConcern.
graylog2-server/src/main/java/org/graylog2/cluster/ClusterConfigServiceImpl.java:        coll.setWriteConcern(WriteConcern.FSYNCED);
graylog2-server/src/main/java/org/graylog2/cluster/ClusterConfigServiceImpl.java:        dbCollection.update(DBQuery.is("type", canonicalClassName), clusterConfig, true, false, WriteConcern.FSYNCED);
graylog2-server/src/main/java/org/graylog2/database/MongoConnectionImpl.java:            db.setWriteConcern(WriteConcern.SAFE);
graylog2-server/src/main/java/org/graylog2/events/ClusterEventCleanupPeriodical.java:            final WriteResult<ClusterEvent, String> writeResult = dbCollection.remove(query, WriteConcern.FSYNCED);
graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java:        coll.setWriteConcern(WriteConcern.JOURNALED);
graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java:            final String id = dbCollection.save(clusterEvent, WriteConcern.JOURNALED).getSavedId();
graylog2-server/src/main/java/org/graylog2/metrics/MongoDbMetricsReporter.java:            collection.insert(docs, WriteConcern.UNACKNOWLEDGED);
graylog2-server/src/test/java/org/graylog2/cluster/ClusterConfigServiceImplTest.java:        assertThat(collection.getWriteConcern()).isEqualTo(WriteConcern.FSYNCED);
graylog2-server/src/test/java/org/graylog2/cluster/ClusterConfigServiceImplTest.java:        assertThat(collection.getWriteConcern()).isEqualTo(WriteConcern.FSYNCED);
graylog2-server/src/test/java/org/graylog2/events/ClusterEventPeriodicalTest.java:        assertThat(collection.getWriteConcern()).isEqualTo(WriteConcern.JOURNALED);
graylog2-server/src/test/java/org/graylog2/events/ClusterEventPeriodicalTest.java:        assertThat(collection.getWriteConcern()).isEqualTo(WriteConcern.JOURNALED);
```

bernd:
I was testing this with MongoDB 2.0.4 and I didn't ran into any error or warning. I was able to create dashboards, streams, inputs. No problem. I the minimum requirement is actually 2.4, we warn the user or maybe even stop the server if it's connected to an older MongoDB version.

```
2016-02-15 10:31:05,827 INFO : org.mongodb.driver.cluster - Monitor thread successfully connected to server with description ServerDescription{address=127.0.0.1:27017, type=STANDALONE, state=CONNECTED, ok=true, version=ServerVersion{versionList=[2, 0, 4]}, minWireVersion=0, maxWireVersion=0, maxDocumentSize=16777216, roundTripTimeNanos=1896308}
```

bernd:
Can you rebase against master?! Thanks! :smiley:

graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
joschi: Yes, I think replacing `FSYNCED` with `JOURNALED` is a good idea as it doesn't block a single server to fsync data to disk but makes sure that the document has been written to MongoDB's on disk journal.

* https://docs.mongodb.org/v3.2/reference/command/fsync/
* http://api.mongodb.org/java/3.2/com/mongodb/WriteConcern.html#FSYNCED
* http://api.mongodb.org/java/3.2/com/mongodb/WriteConcern.html#JOURNALED

graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
bernd: I think we can remove the warning then.

graylog2-server/src/main/java/org/graylog2/events/ClusterEventPeriodical.java
bernd: :+1: Also, the documentation for `FSYNC` says: "prefer JOURNALED"

bernd:
LGTM :+1:

PR #1780, opened 2016-02-09
Upgrade Javascript dependencies

joschi:
This PR upgrades the primary Javascript dependencies to Node.js 4.3.0 LTS, npm 3.7.2, and frontend-maven-plugin 0.0.28.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: false
+sudo: required
+dist: trusty
 language: java
 jdk:
   - oraclejdk8
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -561,7 +561,7 @@
                                     <goal>install-node-and-npm</goal>
                                 </goals>
                                 <configuration>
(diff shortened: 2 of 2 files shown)


joschi:
The build currently fails on Travis CI but succeeds locally and on our Jenkins CI instance.

dennisoelkers:
I think as long as it fails on Travis because of this change, it is of no value for us.

joschi:
@dennisoelkers I agree, that's why I've removed the "ready-for-review" label.

It would be of value to use a still supported version of node.js but we'll need some spare time to find out why Travis CI won't build the native extensions in that version.

dennisoelkers:
:+1:

joschi:
Running the build on the Ubuntu 14.04 Trusty Tahr infrastructure seems to have fixed the issue with the ancient C++ compiler on the old Ubuntu 12.04 LTS containers. Unfortunately this is still in beta at Travis CI and takes a little bit longer until builds start.

PR #1783, opened 2016-02-09
Dynamic search results update

dennisoelkers:
* Updating search results, histogram and field charts periodically.
* Add controls to pause/resume updates or change interval.
* Pause updates when a message is expanded.


bernd:
Refs Graylog2/graylog2-web-interface#133

edmundoa:
I took the liberty of making some changes on the way of refreshing the search histogram data. I will take a look at the issues with field graphs tomorrow.

edmundoa:
@dennisoelkers I committed the changes we did together this morning, and continued making the auto-refresh only update the field graphs data. Feel free to merge if you see no issues with those changes :+1:

lennartkoopmann:
Thanks!! :)

What was the reason to only update field graphs and not the histogram etc?

edmundoa:
We update everything, on my last comment I meant that we refresh the field graphs data instead of re-rendering them completely. We also do that in the search histogram btw :)

lennartkoopmann:
Yippie! :)

PR #1786, opened 2016-02-09
Only load full namespace on demand, not when nodes collection updates.

### LINKED ISSUE #1782: MetricsStore.js is loading 130kb of JSON every 2 seconds
The JavaScript metrics store is currently loading 130kb of JSON every 2 seconds. It should only load the metrics it actually needs instead of everything all the time.

dennisoelkers:
Closes #1782


edmundoa:
LGTM :+1:

PR #1787, opened 2016-02-09
Configurable rotation/retention strategy (REST API)

joschi:
This PR adds the capabilities to dynamically set and configure the rotation and retention strategies through the Graylog REST API.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
@@ -1,0 +1,153 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.rest.resources.system.indices;
+
+import com.codahale.metrics.annotation.Timed;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.module.jsonSchema.JsonSchema;
+import com.fasterxml.jackson.module.jsonSchema.factories.SchemaFactoryWrapper;
+import com.wordnik.swagger.annotations.Api;
+import com.wordnik.swagger.annotations.ApiOperation;
+import com.wordnik.swagger.annotations.ApiParam;
+import org.apache.shiro.authz.annotation.RequiresAuthentication;
+import org.graylog2.indexer.management.IndexManagementConfig;
+import org.graylog2.indexer.retention.strategies.RetentionStrategyConfig;
+import org.graylog2.plugin.cluster.ClusterConfigService;
+import org.graylog2.plugin.indexer.retention.RetentionStrategy;
+import org.graylog2.rest.models.system.indices.RetentionStrategies;
+import org.graylog2.rest.models.system.indices.RetentionStrategySummary;
+import org.graylog2.shared.rest.resources.RestResource;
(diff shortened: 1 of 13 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/RetentionStrategyResource.java
```diff
@@ -0,0 +1,154 @@
...
+
+        return RetentionStrategies.create(strategies.size(), strategies);
+    }
+
+    @GET
+    @Path("strategies/{strategy}")
+    @Timed
+    @ApiOperation(value = "Show JSON schema for configuration of given retention strategies",
+            notes = "This resource returns a JSON schema for the configuration of the given retention strategy.")
+    public JsonSchema configSchema(@PathParam("strategy") @NotEmpty String strategyName) {
```
bernd: What's the reason for using json schema here? AFAIK we are not using this anywhere else (yet) and we should make sure it makes sense to introduce this. :smile:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/RetentionStrategyResource.java
```diff
@@ -0,0 +1,154 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.util.Map;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+@Api(value = "System/Retention", description = "Retention strategy settings")
+@Path("/system/retention")
```
bernd: I think this should live under `/system/index/...` or `/system/indices/...` or something similar. It's strongly related to indexing in my opinion.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/RetentionStrategyResource.java
joschi: Well, we have to start somewhere. :wink:

Generating the JSON schema from an Jackson-annotated class basically comes for free via the Jackson JSON Schema module and it's a more or less standardized format for which there is also support on the (Javascript) client-side to automatically generate forms and validate inputs, e. g. https://mozilla-services.github.io/react-jsonschema-form/, https://github.com/tdegrunt/jsonschema, and https://github.com/AppliedMathematicsANU/plexus-form.

BTW, we're already using JSON Schema implicitly in the API browser.

bernd: Okay, cool. I just wanted to make sure that using json schema is a deliberate decision, that we want to continue to use it for other parts as well and that everyone knows about it. Otherwise it would become a "oh yeah, in this corner we are using json schema, but here we are using xyz". :wink:

joschi: We just have to get rid of `ConfigurationRequest` at some point. :trollface:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
```diff
@@ -0,0 +1,150 @@
...
+
+        return RetentionStrategies.create(strategies.size(), strategies);
+    }
+
+    @GET
+    @Path("strategies/{strategy}")
+    @Timed
+    @ApiOperation(value = "Show JSON schema for configuration of given retention strategies",
+            notes = "This resource returns a JSON schema for the configuration of the given retention strategy.")
+    public JsonSchema configSchema(@PathParam("strategy") @NotEmpty String strategyName) {
```
bernd: This is missing `@ApiParam()` to make it show up in the API browser. (same for the `RotationStrategyResource`)

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
```diff
@@ -0,0 +1,150 @@
...
+        return RetentionStrategySummary.create(strategyName, config);
+    }
+
+    @PUT
+    @Path("config")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Timed
+    @ApiOperation(value = "Configuration of the current retention strategy",
+            notes = "This resource stores the configuration of the currently used retention strategy.")
+    public void config(@Valid @NotNull RetentionStrategySummary retentionStrategySummary) {
```
bernd: This also needs `@ApiParam`, no text box for the input in the API browser right now.

bernd:
LGTM :+1:

PR #1788, opened 2016-02-09
Pluggable index management UI

bernd:
Adds UI plugin hooks for index rotation and retention configuration components and UI plugins for all core retention and rotation strategies.

~~Includes two new pieces of infrastructure.~~

#### ~~GraylogWebState~~

~~This provides an API to store global state in a controlled way. All state is stored in a single field in the `document` as a ES6 `Map`. (currently `document.GRAYLOG_WEB_STATE`)~~

~~Trying to overwrite an existing k/v throws an error to avoid accidental state mutations. A k/v needs to be removed before it can be set again.~~

~~Using global state should be avoided, but is sometimes needed to share state between the core and plugins. We might also migrate our existing usage of global state in our application over to this to avoid global namespace pollution. (i.e. `window.appConfig`, `window.plugins`)~~

#### ~~WebEvents~~

~~In `WebEvents.js` I introduced an implementation independent event bus API for the web app which is also usable from plugins. For the pluggable retention/rotation UI this is needed to have child components listen to state changes in a parent component.~~

~~It currently uses the `eventemitter3` library, which is also used by Reflux.~~

### DIFF
--- a/graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
+++ b/graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
@@ -2,6 +2,8 @@ import React from 'react';
 
 import { Select } from 'components/common';
 
+import style from '!style!css!./IndicesConfiguration.css';
+
 const IndexMaintenanceStrategiesConfiguration = React.createClass({
   propTypes: {
     title: React.PropTypes.string.isRequired,
@@ -61,26 +63,26 @@ const IndexMaintenanceStrategiesConfiguration = React.createClass({
 
   _availableSelectOptions() {
     return this.props.pluginExports.map((config) => {
-      return {value: config.type, label: config.displayName}
+      return {value: config.type, label: config.displayName};
     });
   },
 
   _getConfigurationComponent(selectedStrategy) {
     if (!selectedStrategy || selectedStrategy.length < 1) {
-      return;
+      return null;
     }
 
-    const strategy = this.props.pluginExports.find((strategy) => strategy.type === selectedStrategy);
+    const strategy = this.props.pluginExports.find((exportedStrategy) => exportedStrategy.type === selectedStrategy);
 
     if (!strategy) {
-      return undefined;
+      return null;
     }
 
     const strategyConfig = this._getStrategyConfig(selectedStrategy);
     const element = React.createElement(strategy.configComponent, {
       config: strategyConfig,
       jsonSchema: this._getStrategyJsonSchema(selectedStrategy),
-      updateConfig: this._onConfigUpdate
+      updateConfig: this._onConfigUpdate,
     });
 
     return (<span key={strategy.type}>{element}</span>);
@@ -94,22 +96,22 @@ const IndexMaintenanceStrategiesConfiguration = React.createClass({
     return (
       <span>
         <h3>{this.props.title}</h3>
-        <div style={{marginTop: 10}}>
+        <div className={style.topMargin}>
           <p>{this.props.description}</p>
         </div>
-        <div style={{marginTop: 10}}>
+        <div className={style.topMargin}>
           <Select placeholder={this.props.selectPlaceholder}
                   options={this._availableSelectOptions()}
-                  matchProp='value'
+                  matchProp="value"
                   value={this._activeSelection()}
                   onValueChange={this._onSelect}/>
         </div>
-        <div style={{marginTop: 10}}>
+        <div className={style.topMargin}>
           {this._getConfigurationComponent(this._activeSelection())}
         </div>
       </span>
     );
-  }
+  },
 });
 
 export default IndexMaintenanceStrategiesConfiguration;
--- a/graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
+++ b/graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
@@ -1,0 +1,115 @@
+import React from 'react';
+
+import { Select } from 'components/common';
+
+const IndexMaintenanceStrategiesConfiguration = React.createClass({
+  propTypes: {
+    title: React.PropTypes.string.isRequired,
+    description: React.PropTypes.string.isRequired,
+    selectPlaceholder: React.PropTypes.string.isRequired,
+    pluginExports: React.PropTypes.array.isRequired,
+    strategies: React.PropTypes.array.isRequired,
+    activeConfig: React.PropTypes.object.isRequired,
+    updateState: React.PropTypes.func.isRequired,
+  },
+
+  getInitialState() {
+    return {
+      activeStrategy: this.props.activeConfig.strategy,
+      activeConfig: this.props.activeConfig.config,
+      newStrategy: this.props.activeConfig.strategy,
+      newConfig: this.props.activeConfig.config,
+    };
+  },
+
+  _getDefaultStrategyConfig(selectedStrategy) {
+    const result = this.props.strategies.find((strategy) => strategy.type === selectedStrategy);
+    return result ? result.default_config : undefined;
+  },
+
+  _getStrategyJsonSchema(selectedStrategy) {
+    const result = this.props.strategies.find((strategy) => strategy.type === selectedStrategy);
+    return result ? result.json_schema : undefined;
+  },
+
+  _getStrategyConfig(selectedStrategy) {
+    if (this.state.activeStrategy === selectedStrategy) {
+      // If the newly selected strategy is the current active strategy, we use the active configuration.
+      return this.state.activeConfig;
+    } else {
+      // If the newly selected strategy is not the current active strategy, we use the selected strategy's default config.
+      return this._getDefaultStrategyConfig(selectedStrategy);
+    }
+  },
+
+  _onSelect(newStrategy) {
+    if (!newStrategy || newStrategy.length < 1) {
+      this.setState({newStrategy: undefined});
+      return;
+    }
+
+    const newConfig = this._getStrategyConfig(newStrategy);
+
+    this.setState({newStrategy: newStrategy, newConfig: newConfig});
+    this.props.updateState(newStrategy, newConfig);
+  },
+
+  _onConfigUpdate(newConfig) {
+    this.setState({newConfig: newConfig});
+    this.props.updateState(this.state.newStrategy, newConfig);
+  },
+
+  _availableSelectOptions() {
+    return this.props.pluginExports.map((config) => {
+      return {value: config.type, label: config.displayName}
+    });
+  },
+
+  _getConfigurationComponent(selectedStrategy) {
+    if (!selectedStrategy || selectedStrategy.length < 1) {
+      return;
+    }
+
+    const strategy = this.props.pluginExports.find((strategy) => strategy.type === selectedStrategy);
+
+    if (!strategy) {
+      return undefined;
+    }
+
+    const strategyConfig = this._getStrategyConfig(selectedStrategy);
+    const element = React.createElement(strategy.configComponent, {
+      config: strategyConfig,
+      jsonSchema: this._getStrategyJsonSchema(selectedStrategy),
+      updateConfig: this._onConfigUpdate
+    });
+
+    return (<span key={strategy.type}>{element}</span>);
+  },
+
+  _activeSelection() {
+    return this.state.newStrategy;
+  },
+
+  render() {
+    return (
+      <span>
+        <h3>{this.props.title}</h3>
+        <div style={{marginTop: 10}}>
+          <p>{this.props.description}</p>
+        </div>
+        <div style={{marginTop: 10}}>
+          <Select placeholder={this.props.selectPlaceholder}
+                  options={this._availableSelectOptions()}
+                  matchProp='value'
+                  value={this._activeSelection()}
+                  onValueChange={this._onSelect}/>
+        </div>
+        <div style={{marginTop: 10}}>
+          {this._getConfigurationComponent(this._activeSelection())}
+        </div>
+      </span>
+    );
+  }
+});
+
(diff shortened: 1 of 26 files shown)


bernd:
Screenshot:

graylog

dennisoelkers:
I am already commenting on this although it is WIP, because those parts explicitly mentioned in the description are not subject to change, but I think we do not need them and should use common parts of our existing infrastructure instead.

 **WebEvents:**
This pattern is basically why React/Flux exists. If you want to share data/events between a parent component and its children, you use React's props to pass them down. If you want to listen to actions, you use Reflux. I think introducing a separate entity circumventing this does not generate any benefit for us, but introduces more complexity and confusion and also circumvents React's system of conditional rerendering based on state/props. The fact that we can do this more easily also makes the introduction of the `GraylogWebState` unnecessary.

 **GraylogWebState:**
I think we agree that global state is avoidable (and should be) in most cases. In general global state is only necessary, when two namespaces (which means two "files" in the javascript world) need to be crossed. In this case, where it is used to share a common instance of the event emitter I've already pointed out why it is not necessary (and I would strongly recommend against it) to implement this. From a general perspective I don't see the need for this right now either. Right now there are two scenarios where global state is used and both are very specific:

  1) `AppConfig`: The Graylog server/the user (for dev) wants to pass information to the web interface. This could actually also be implemented without global state by appending the file onto the rest of the bundle, but for the sake of simplicity it was done like this for now. It is wrapped in a separate class and immutable to make it replaceable easily in the future.

 2) `PluginStore`: This is used for registering plugins dynamically loaded. The only application (from my perspective) where we cannot get rid of global state right now, as we are crossing file boundaries without knowing in advance which files we are gonna load/which symbols they export. *This is the only bridge which should exist between plugins and the app.*

Implicitly utilizing the `PluginStore` we use React contexts, which allow us to pass down anything to everywhere if we really need to. We can even track which components are using something, by checking their `contextTypes` property and we are able to inject a different instance easily (for testing). Even if we would need the `WebEvents` implementation you suggested, we shouldn't implement a general solution for global state but use `context` instead.

**TL;DR:**
  * Replace `WebEvents` with Reflux actions/store
  * Remove `GraylogWebState`

bernd:
Thanks for the feedback!

#### WebEvents

> This pattern is basically why React/Flux exists. If you want to share data/events between a parent component and its children, you use React's props to pass them down.

What I need is the following:

I have a modal in `IndicesConfiguration`. In that modal I render rotation/retention configuration components from plugins. If I press "Save" in the modal, I have to notify the configuration components inside the modal about that ("save button pressed") so they can save their state to the server.

How do I do this with Reflux? :smiley:

#### GraylogWebState

This was indeed just built because I needed global state for the WebEvents and didn't want to randomly put it into the global namespace.

Regarding React contexts, when I read about it in the React documentation it read like "experimental, might get removed/changed, do not use" to me. (probably too pessimistic from my side) :wink: But since we are already using it anyway, we should also use it if we need to share things between core and plugins. Agreed.
If we are going to use this more often, I think it's important to expose an abstraction to plugins and not just telling plugin authors to use the React context directly. Otherwise it becomes really hard to change it later. But I guess this is future talk.

dennisoelkers:
`How do I do this with Reflux?`:

(Please take note that I don't know the data model which is stored in the backend. I would assume that it's an object containing two type string defining the strategies and a map for each holding the type-specific config.)
The way that I would recommend is to create two stores named something along the lines of `IndexRetentionConfigStore` & `indexRotationConfigStore` (assuming those two entities are indepent). Both stores should contain functionality to fetch and update the type/config. Then you would create the corresponding actions `IndexRetentionConfigActions` & `IndexRotationConfigActions`. In your components you would consume the stores and use the actions when saving an entity. Just make sure that you refresh the store's state after a completed update.

bernd:
@dennisoelkers I updated the pull request and removed the `WebEvents` and `GraylogWebState` objects by using regular React/Reflux infrastructure.

This is now ready for review.

Screenshots:

h tggeniws4daaaaabjru5erkjggg
---
9fwjcrzkatifnaaaaaelftksuqmcc

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -0,0 +1,115 @@
...
+  },
+
+  _onConfigUpdate(newConfig) {
+    this.setState({newConfig: newConfig});
+    this.props.updateState(this.state.newStrategy, newConfig);
+  },
+
+  _availableSelectOptions() {
+    return this.props.pluginExports.map((config) => {
+      return {value: config.type, label: config.displayName}
```
dennisoelkers: There is a missing semicolon at the end. You can also write it a bit more concise by doing

```
return this.props.pluginExports.map((config) => {value: config.type, label: config.displayName});
```

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  },
+
+  _getConfigurationComponent(selectedStrategy) {
+    if (!selectedStrategy || selectedStrategy.length < 1) {
+      return;
+    }
+
+    const strategy = this.props.pluginExports.find((strategy) => strategy.type === selectedStrategy);
```
dennisoelkers: You are declaring `strategy` twice here. Although it happens in different scopes, it is potentially misleaded and should be avoided.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+    if (!strategy) {
+      return undefined;
```
dennisoelkers: Please return `null` instead of `undefined here. `undefined` usually implies erroneous results while `null` means that there is no value for this function call.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -0,0 +1,115 @@
...
+    return (
+      <span>
+        <h3>{this.props.title}</h3>
+        <div style={{marginTop: 10}}>
+          <p>{this.props.description}</p>
+        </div>
+        <div style={{marginTop: 10}}>
+          <Select placeholder={this.props.selectPlaceholder}
+                  options={this._availableSelectOptions()}
+                  matchProp='value'
```
dennisoelkers: Please use double quotes for html props, instead of single quotes.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
bernd: :heavy_check_mark:

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
bernd: Aye true, I will change that.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesSummary.jsx
```diff
@@ -0,0 +1,28 @@
...
+    config: React.PropTypes.object.isRequired,
+  },
+
+  render() {
+    if (!this.props.config) {
+      return (<Spinner />);
+    }
+
+    const activeStrategy = this.props.config.strategy;
+    const strategy = this.props.pluginExports.find((strategy) => strategy.type === activeStrategy);
```
dennisoelkers: `pluginExports` is missing in `propTypes`, also `strategy` is declared twice, in different scopes.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
bernd: Sure will do.

I just noticed that we are mostly using `undefined` for this kind of stuff. Also for values in `getInitialState()`.

This is only semantics, right? In comparisons like `let a = null; if ("foo" === a)` it doesn't make any difference if a is `null` or `undefined`?

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
bernd: I can change that. What's the reason to use double quotes for that?

dennisoelkers: It's for consistency. We aim to enforce consistent use of single for js, double quotes for html. It also helps to visually distinguish both at first glance when you are used to it.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
dennisoelkers: It is okay (from my perspective) to do it in `getInitialState` because it signals that this variable is not initialized. I guess it is only done in older components, because it is in 99% of the cases omittable, actually. It helps when debugging/logging to see if an expression is `null` or `undefined`, because for the first case it is very probable that it was returned on purpose.

For the evaluation of an equality check it does not matter, yes.

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  _activeSelection() {
+    return this.state.newStrategy;
+  },
+
+  render() {
+    return (
+      <span>
+        <h3>{this.props.title}</h3>
+        <div style={{marginTop: 10}}>
```
dennisoelkers: Is there a need for inline style definitions? I know we are using them a lot in the existing codebase, but this is mainly because of legacy HTML that was ported over. For new code we should extract those into separate CSS/LESS fragments.

graylog2-web-interface/src/components/indices/IndicesConfiguration.jsx
```diff
@@ -0,0 +1,145 @@
+import React from 'react';
+import Reflux from 'reflux';
+
+import { Button, Input, Row, Col } from 'react-bootstrap';
```
dennisoelkers: `Input` is imported but not used.

graylog2-web-interface/src/components/indices/IndicesConfiguration.jsx
```diff
@@ -0,0 +1,145 @@
...
+import BootstrapModalForm from 'components/bootstrap/BootstrapModalForm';
+import { Select } from 'components/common';
```
dennisoelkers: Unused import.

dennisoelkers:
There is quite a number of minor issues that could have been caught by using our ESLint configuration and turning on ESLint support in your IDE, or run the changeset through ESLint manually.

bernd:
> There is quite a number of minor issues that could have been caught by using our ESLint configuration and turning on ESLint support in your IDE, or run the changeset through ESLint manually. Please do this once and fix the linter hints, at least the errors.

Ah, cool. I wasn't aware of that. Do we have the instructions on how to setup the environment for all of this written down somewhere?

bernd:
@dennisoelkers Addressed all your comments in the latest commit.

joschi:
After updating the rotation strategy to time-based with an 1 hour interval (PT1H), the System &rarr; Indices page doesn't show the configuration anymore.

The Chrome Developer console shows the following warning:
```
Unhandled rejection TypeError: duration.format is not a function
    at _humanizedPeriod (webpack:///./src/components/indices/rotation/TimeBasedRotationStrategySummary.jsx?:33:21)
    at render (webpack:///./src/components/indices/rotation/TimeBasedRotationStrategySummary.jsx?:63:16)
    at y._renderValidatedComponentWithoutOwnerOrContext (http://127.0.0.1:9000/vendor.js:407:10311)
    at y._renderValidatedComponent (http://127.0.0.1:9000/vendor.js:407:10487)
    at i [as _renderValidatedComponent] (http://127.0.0.1:9000/vendor.js:12:21248)
    at y.mountComponent (http://127.0.0.1:9000/vendor.js:407:5463)
    at i [as mountComponent] (http://127.0.0.1:9000/vendor.js:12:21248)
    at Object.a.mountComponent (http://127.0.0.1:9000/vendor.js:13:5381)
    at y.v.Mixin.mountChildren (http://127.0.0.1:9000/vendor.js:408:12300)
    at y.Mixin._createContentMarkup (http://127.0.0.1:9000/vendor.js:407:22049)
    at y.Mixin.mountComponent (http://127.0.0.1:9000/vendor.js:407:20949)
    at Object.a.mountComponent (http://127.0.0.1:9000/vendor.js:13:5381)
    at y.mountComponent (http://127.0.0.1:9000/vendor.js:407:5559)
    at i [as mountComponent] (http://127.0.0.1:9000/vendor.js:12:21248)
    at Object.a.mountComponent (http://127.0.0.1:9000/vendor.js:13:5381)
    at y.v.Mixin._mountChildByNameAtIndex (http://127.0.0.1:9000/vendor.js:408:13844)
```

joschi:
@bernd The latest commit fixes the issue.

LGTM. :+1:

shakalandy:
Hi Guys,

will it be possible to use multiple retention strategies with the pluggable architecture in the future?

Thanks!

bernd:
@shakalandy  The pluggable architecture allows you to write your own retention strategies, it does not allow different retention strategies for different kind of data, yet.

shakalandy:
@bernd is something like "allow different retention strategies" already on the roadmap? Didn't find anything related in the issues? Would file it if it could help :)

bernd:
@shakalandy Yes, we have that as a feature request already: https://graylog.ideas.aha.io/ideas/GL2E-I-356

Issue #1775, opened 2016-02-05
Default limit value is not applied on searches


edmundoa:
Using the current master, the default limit parameter for searches is not working. Here's the result of the query `http://hostname:12900/search/universal/relative?query=*&range=3600&pretty=true`:

```
{
  "query" : "*",
  "built_query" : "{\"from\":0,\"query\":{\"bool\":{\"must\":{\"match_all\":{}},\"filter\":{\"bool\":{\"must\":{\"range\":{\"timestamp\":{\"from\":\"2016-02-05 17:25:44.854\",\"to\":\"2016-02-05 18:25:44.854\",\"include_lower\":true,\"include_upper\":true}}}}}}},\"sort\":[{\"timestamp\":{\"order\":\"desc\"}}],\"highlight\":{\"require_field_match\":false,\"fields\":{\"*\":{\"fragment_size\":0,\"number_of_fragments\":0}}}}",
  "used_indices" : [ {
    "index_name" : "graylog2_19",
    "begin" : "1970-01-01T00:00:00.000Z",
    "end" : "1970-01-01T00:00:00.000Z",
    "calculated_at" : "2016-02-05T13:28:28.379Z",
    "took_ms" : 9
  }, {
    "index_name" : "graylog2_5",
    "begin" : "2015-12-17T10:34:30.794Z",
    "end" : "2016-12-17T17:26:41.000Z",
    "calculated_at" : "2016-01-26T14:11:56.019Z",
    "took_ms" : 13
  } ],
  "messages" : [ {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
    "index" : "graylog2_19",
    "highlight_ranges" : null
  }, {
    "message" : {},
[...]

PR #1793, opened 2016-02-10
fix wrong default of limit search parameter

kroepke:
the builder didn't properly default a 0 value to the actual default

fixes issue #1775

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesConfig.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesConfig.java
@@ -26,7 +26,7 @@ import java.util.List;
 
 @AutoValue
 public abstract class SearchesConfig {
-    private final static int LIMIT = 150;
+    public final static int DEFAULT_LIMIT = 150;
 
     @JsonProperty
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesConfig.java
```diff
@@ -87,10 +87,19 @@ public static Builder builder() {
...
+
         public abstract Builder offset(int offset);
 
         public abstract Builder sorting(Sorting sorting);
 
-        public abstract SearchesConfig build();
+        abstract SearchesConfig autoBuild();
+
+        public SearchesConfig build() {
+            if (limit() == 0) {
```
joschi: The condition before (line 68) was `limit <= 0`. Does that make any difference in practice?

kroepke: Good point, I guess only if someone supplies `< 0` via the API, which is nonsensical anyway.

I'll update this and the test to provide a sensible default.

joschi:
LGTM. :+1:

PR #1801, opened 2016-02-11
Refactor retention/rotation configuration

bernd:
- Move config interfaces to the plugin-interfaces module to make them
  accessible from the strategy interfaces.
- Add `defaultConfiguration()` method to the Retention/RotationStrategy
  interfaces.
- Add a static `createDefault()` method to each strategy config class
  which returns the default configuration for the strategy.
- Remove `toDeflectorConfigResponse()` from the rotation strategy
  config interface.
- Remove the `/system/deflector/config` REST API endpoint and remove the
  config data from the `/system/deflector` endpoint.
- Remove the DeflectorConfigSummary from the web UI. This will be
  replaced later.

All these changes are in preparation for the retention/rotation configuration UI.


joschi:
LGTM. :+1:

PR #1806, opened 2016-02-12
Simple query time range limit

joschi:
In order to guard the system against accidental queries covering a large time range, the `query_time_range_limit` setting now provides a limit for those queries.

If a search query exceeds the time range given in ´query_time_range_limit`, the query is modified to fit within that limit by adjusting the starting time.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
@@ -1,0 +1,58 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.indexer.searches;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import org.joda.time.Period;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class SearchesClusterConfig {
+    private static final Period DEFAULT_QUERY_TIME_RANGE_LIMIT = Period.ZERO;
+
+    @JsonProperty("query_time_range_limit")
+    public abstract Period queryTimeRangeLimit();
+
+    @JsonCreator
+    public static SearchesClusterConfig create(@JsonProperty("query_time_range_limit") Period queryTimeRangeLimit) {
+        return builder()
+                .queryTimeRangeLimit(queryTimeRangeLimit)
+                .build();
+    }
+
+    public static SearchesClusterConfig createDefault() {
+        return builder()
+                .queryTimeRangeLimit(DEFAULT_QUERY_TIME_RANGE_LIMIT)
+                .build();
+    }
+
+    public static Builder builder() {
+        return new AutoValue_SearchesClusterConfig.Builder();
+    }
+
+    public abstract Builder toBuilder();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+        public abstract Builder queryTimeRangeLimit(Period queryTimeRangeLimit);
+
+        public abstract SearchesClusterConfig build();
+    }
+}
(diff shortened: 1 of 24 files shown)


bernd:
I was wondering why we apply the limit at the "outer" layer in the REST resources instead of the `Searches` class? This way we might miss some places. (i.e. existing dashboards)

graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Map;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class SearchesClusterConfig {
+    private static final org.joda.time.format.PeriodFormatter ISO_PERIOD_FORMAT = ISOPeriodFormat.standard();
+
+    private static final Period DEFAULT_QUERY_TIME_RANGE_LIMIT = Period.ZERO;
+    private static final Map<Period, String> DEFAULT_RELATIVE_TIMERANGE_OPTIONS = ImmutableMap.<Period, String>builder()
+            .put(ISO_PERIOD_FORMAT.parsePeriod("PT5M"), "Search in the last 5 minutes")
```
joschi: Wouldn't it be easier to use the static factory methods of [`Period`](http://joda-time.sourceforge.net/apidocs/org/joda/time/Period.html), e. g. `Period.minutes(5)`? Or do they result in different objects?

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+        <dl className={style.deflist}>
+          {timerangeOptionsSummary}
+        </dl>
+
+        <IfPermitted permissions="clusterconfigentry:edit">
+          <Button bsStyle="info" bsSize="xs" onClick={this._openModal}>Update</Button>
+        </IfPermitted>
+
+        <BootstrapModalForm ref="searchesConfigModal"
```
edmundoa: The component feels too big and I would actually move this form outside, just to make things more clear.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -0,0 +1,288 @@
...
+          let value = e.target.value;
+
+          if (field === 'period') {
+            value = value.toUpperCase();
+            if (!value.startsWith('P')) {
+              value = `P${value}`;
+            }
+          }
+
+          options[idx][field] = value;
```
edmundoa: Here we are mutating the component's state directly, and this is highly discourage in the documentation: https://facebook.github.io/react/docs/component-api.html#setstate. You can use `ObjectUtils.clone` to clone the state object before modifying it.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -0,0 +1,288 @@
...
+  _onPeriodUpdate(field) {
+    return () => {
+      const update = this.state.config;
+      let period = this.refs[field].getValue().toUpperCase();
+
+      if (!period.startsWith('P')) {
+        period = `P${period}`;
+      }
+
+      update[field] = period;
```
edmundoa: Here we are modifying the state directly as well

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+  _updateTimerangeOptionsState(options) {
+    const config = this.state.config;
+    const newTimerangeOptions = {};
+
+    options.forEach((option) => {
+      newTimerangeOptions[option.period] = option.description;
+    });
+
+    config.relative_timerange_options = newTimerangeOptions;
```
edmundoa: Here we are modifying the state directly as well.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -0,0 +1,288 @@
...
+
+  render() {
+    const config = this.state.config;
+    const duration = moment.duration(config.query_time_range_limit);
+    const limit = this._isEnabled() ? `${config.query_time_range_limit} (${duration.format()})` : 'disabled';
+
+    let timerangeOptions = null;
+    let timerangeOptionsSummary = null;
+    if (this.state.timerangeOptionsList) {
+      timerangeOptions = this.state.timerangeOptionsList.map((option, idx) => {
```
edmundoa: I like to put these large blocks creating some jsx into a function for clarity, also keeps the render method short. That's something of personal taste, of course :)

graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
```diff
@@ -0,0 +1,61 @@
...
+      this.trigger({configuration: configuration});
+      UserNotification.success('Configuration updated successfully');
+    }, this._errorHandler('Search config update failed', `Could not update search config: ${configType}`));
+
+    ConfigurationActions.update.promise(promise);
+  },
+
+  _errorHandler(message, title) {
+    return (error) => {
+      let errorMessage;
```
edmundoa: You shouldn't need to do this here, promises are returning a `FetchError` object that should already have the right message set. If, for some reason, that is not working properly, we should adjust it here so all promises get the right message: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/logic/rest/FetchProvider.js#L7-L11

edmundoa:
Other than the mentioned issues, we should also try to keep the UI more or less consistent with the existing one. But we can fix that later on, of course :+1:

PR #1808, opened 2016-02-12
Allow overriding web if server uri from X-Graylog-Server-URL header.

dennisoelkers:
This allows http clients (including proxies) to define the server uri
that is used by the web interface code to contact the REST API.

This is especially useful for reverse proxies exposing internal Graylog
servers to the outside. In the proxy configuration you can just set the
X-Graylog-Server-URL that is passed to the backend to a URL that is
visible from outside enabling access to the server from the web if.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
@@ -18,26 +18,39 @@ package org.graylog2.web.resources;
 
 import com.floreysoft.jmte.Engine;
 import com.google.common.io.Resources;
+import org.apache.directory.api.util.Strings;
 import org.graylog2.Configuration;
 
 import javax.inject.Inject;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 @Path("/config.js")
 public class AppConfigResource {
+    private static final String CK_OVERRIDE_SERVER_URI = "X-Graylog-Server-URL";
     private static final Engine engine = new Engine();
-    private final String content;
+    private final Configuration configuration;
 
     @Inject
     public AppConfigResource(Configuration configuration) {
+        this.configuration = configuration;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    public String get(@Context HttpHeaders headers) {
         final URL templateUrl = this.getClass().getResource("/web-interface/config.js.template");
         final String template;
         try {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -18,26 +18,39 @@
 
 import com.floreysoft.jmte.Engine;
 import com.google.common.io.Resources;
+import org.apache.directory.api.util.Strings;
```
joschi: Wrong import? Should probably be `com.google.common.base.Strings`.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -18,26 +18,39 @@
...
-    private final String content;
+    private final Configuration configuration;
 
     @Inject
     public AppConfigResource(Configuration configuration) {
+        this.configuration = configuration;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
```
joschi: Should this produce `application/octet-stream` instead of `application/json`?

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -18,26 +18,39 @@
...
+    public String get(@Context HttpHeaders headers) {
         final URL templateUrl = this.getClass().getResource("/web-interface/config.js.template");
```
joschi: Not in scope for this PR but out of interest: Why is a template being used here instead of a Jackson-annotated POJO like most other places? Should the template contain JavaScript code that goes beyond the JSON specification at one point?

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -45,17 +58,30 @@ public AppConfigResource(Configuration configuration) {
...
-    @GET
-    @Produces(MediaType.APPLICATION_OCTET_STREAM)
-    public String get() {
-        return this.content;
+    private String serverUriFromHeadersOrDefault(HttpHeaders httpHeaders) {
+        final List<String> overrideServerUriHeaders = httpHeaders.getRequestHeader(CK_OVERRIDE_SERVER_URI);
+        if (overrideServerUriHeaders != null && !overrideServerUriHeaders.isEmpty()) {
+            Optional<String> firstHeader = overrideServerUriHeaders.stream().filter(s -> {
+                try {
+                    return s != null && Strings.isNotEmpty(s) && new URL(s) != null;
```
joschi: Unfortunately `URL` will execute a DNS lookup which (since this is called every time the `/config.js` resource is being requested) might slow down the system as a whole. Also see http://www.eishay.com/2008/04/javas-url-little-secret.html

Please use [`URI`](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#URI(java.lang.String)) instead.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
dennisoelkers: No, it's not json.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
dennisoelkers: It is only `URL.hashCode()` or (implicitly) equality checks which do DNS lookups, not the constructor or `toString()`.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
joschi: :white_check_mark: 
Using `text/javascript` or `application/x-javascript` would be an alternative but not necessarily in this PR.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
dennisoelkers: :+1:

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
joschi: :white_check_mark:

runningman84:
How does this feature work?

This is my configuration:
```
    location / {
        proxy_pass http://graylog_web;
        proxy_set_header X-Graylog-Server-URL $scheme://$http_host/api;
        add_header X-Graylog-Server-URL $scheme://$http_host/api;
    }
    location /api/ {
        proxy_pass http://graylog_api;
        rewrite ^/api/(.*)$ /$1 break;
    }
```
and I get this error:
```
Internal Server Error
java.net.URISyntaxException: Illegal character in hostname at index 14: http://graylog_web:80/

     1: org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpContainer.getBaseUri(GrizzlyHttpContainer.java:457)
     2: org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpContainer.service(GrizzlyHttpContainer.java:365)
     3: org.glassfish.grizzly.http.server.HttpHandler$1.run(HttpHandler.java:224)
     4: com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:176)
     5: java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
     6: java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
     7: java.lang.Thread.run(Thread.java:745)
```
My config was wrong during my initial configuration, It looks like the graylog server saved a wrong url?

dennisoelkers:
Your hostnames contains an underscore ('_') which is not valid, therefore parsing the url passed in the header fails.

SjonHortensius:
Is this functionality included in b1? Nginx lowercases the header-name, and my web-interface only says

> We are experiencing problems connecting to the Graylog server running on http://127.0.0.1:12900//. Please verify that the server is healthy and working correctly.

The `http://127.0.0.1:12900/` part is not what I see in the header, but it is what is in `gl2ServerUrl` in config.js

joschi:
@SjonHortensius Yes, this should work in Graylog 2.0.0-beta.1. How exactly did you set the HTTP request header and to which value?

SjonHortensius:
@joschi thanks for the reply, I actually updated my comment after I found the issue :)

PR #1820, opened 2016-02-15
Append Graylog node ID to Elasticsearch node name

joschi:
In order for users to quickly identify the Graylog server nodes running a specific embedded Elasticsearch instance, the Graylog node ID is now being used to generate the Elasticsearch node name.

The old `elasticsearch_node_name` setting has been replaced with the new `elasticsearch_node_name_prefix` setting which is `graylog-` by default, leading to Elasticsearch node names like `graylog-17052010-1234-5678-abcd-1337cafebabe`.


bernd:
LGTM :+1:

PR #1822, opened 2016-02-15
Simple JAX-RS resource for CRUD operations on cluster configuration

joschi:
This PR adds a relatively simple JAX-RS resource for performing CRUD operations on the Graylog cluster configuration.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
@@ -1,0 +1,152 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.rest.resources.system;
+
+import com.codahale.metrics.annotation.Timed;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.module.jsonSchema.JsonSchema;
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
```diff
@@ -0,0 +1,152 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+@Api(value = "System/ClusterConfig", description = "Graylog Cluster Configuration")
+@RequiresAuthentication
```
bernd: Please also use the existing `RestPermissions.CLUSTER_CONFIG_*` permissions for the different endpoints.

bernd:
To avoid excessive HTTP calls from the web interface, we need one endpoint that takes a list of cluster config classes and returns the values for them. If a given class does exist but there is no entry for it in the database, it should return the default configuration for that entry.

We might have to add an interface for this with a `default()` method to ensure that cluster config objects have a default config.

joschi:
@bernd I'd like to tackle those issues/improvements in a follow-up PR to keep things quickly reviewable.

Regarding the default configuration, what would be the returned value if there is no (sane) default?

bernd:
> I'd like to tackle those issues/improvements in a follow-up PR to keep things quickly reviewable.

:+1: 

> Regarding the default configuration, what would be the returned value if there is no (sane) default?

I think every object in the cluster config that is a "system config" object like retention/rotation or email config has some default values. Otherwise it is not really usable.

After looking at the existing configs, I think there is none that has no default config. Even the `ClusterId` object, its default can be a new UUID every time the default is requested.

joschi:
> Even the `ClusterId` object, its default can be a new UUID every time the default is requested.

But exactly that use case is a problem. If `ClusterId` returns a random UUID by default which is not persisted anywhere, it'll lead to funny effects because different parts of the application will use different Cluster IDs.

bernd:
> But exactly that use case is a problem. If ClusterId returns a random UUID by default which is not persisted anywhere, it'll lead to funny effects because different parts of the application will use different Cluster IDs.

Okay, I see. But is it possible that this is not set in the database? This should be written to the DB as early as possible, no?

If this is a problem, we might have to extend the interface to allow different types of configs. One with default values, one without.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
joschi: :white_check_mark:

joschi:
@bernd The possibility to set a default configuration object will be implemented in the scope of another PR.

bernd:
LGTM :+1:

PR #1836, opened 2016-02-18
Allow to define default instances for cluster configuration classes

joschi:
Cluster configuration classes are simple Jackson-annotated POJOs which are being handled by `ClusterConfigService`.
This change set introduces the possibility to define a default instance being used if there is no explicit instance in the database.

The default instance must be provided by a static method in the respective cluster configuration POJO with the following signature (with `T` being the specific cluster configuration class):

```java
public static T defaultConfig();
```


dennisoelkers:
What is the use case for this? This has been open for so long and does not seem to fill a current need, so maybe we should close it for now.

kroepke:
I would close this and start another attempt to implement this later.
I remember trying to improve a few things but it got too tricky to justify the amount of work now, especially since we have so many different patterns of using defaults right now.

PR #1846, opened 2016-02-22
Remove remains of Graylog Radio

joschi:
This PR removes some remaining artifacts of Graylog Radio.

The Protocol Buffers definition of [`JournalMessage`](https://github.com/Graylog2/graylog2-server/blob/1.3.3/graylog2-plugin-interfaces/src/main/resources/org/graylog2/plugin/journal/raw_message.proto) hasn't been touched for compatibility reasons.


bernd:
LGTM :+1:

PR #1847, opened 2016-02-22
Add documentation to rest_listen_uri

Spitfire1900:
rest_listen_uri is also applicable to collectors connecting to the server. The collectors send PUT hearbeat messages to the server's rest interface. (as of v1.3.3 and collector v0.4.2)

### DIFF
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -34,7 +34,7 @@ root_password_sha2 =
 plugin_dir = plugin
 
 # REST API listen URI. Must be reachable by other graylog2-server nodes if you run a cluster.
-# This must also be reachable by collectors so that hearbeat messages can be received.
+# When using Graylog Collectors, this address is used to receive heartbeat messages and must be reachable by all collectors.
 rest_listen_uri = http://127.0.0.1:12900/
 
 # REST API transport address. Defaults to the value of rest_listen_uri. Exception: If rest_listen_uri
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -33,7 +33,8 @@ root_password_sha2 =
 # Set plugin directory here (relative or absolute)
 plugin_dir = plugin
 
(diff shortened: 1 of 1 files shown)


lennartkoopmann:
Thank you!

I suggest changing it to: 

> If you are using Graylog Collectors to send logs, this address is used to receive heartbeat messages and must be reachable by all collectors.

@bernd can you confirm that this is correct?

joschi:
This is also mentioned in the Graylog Collector documentation at http://docs.graylog.org/en/1.3/pages/collector.html and there's even a troubleshooting item about this very issue: http://docs.graylog.org/en/1.3/pages/collector.html#unable-to-send-heartbeat

I'd prefer not to "pollute" the Graylog server config file with Graylog Collector specifics.

lennartkoopmann:
IMO it is only one line and if it can avoid having to search for the FAQ in the first place I'd vote for including it.

Spitfire1900:
The unable to send heartbeat information is there but best case the configuration script that comes with the package should contain the information necessary to setup services. (barring the documentation site goes down or offline)

@lennartkoopmann 's addition is more descriptive but wordy, how about just:

> This address is also used to receive heartbeat messages and must be reachable by all collectors.
?

lennartkoopmann:
@Spitfire1900 The reason I suggested something more wordy was to make sure that people don't have to go on reading if they are not using collectors. Imagine reading this as someone who is not familiar with Graylog collectors yet. What do you think?

Spitfire1900:
@lennartkoopmann I can trim 24 characters and still retain the appropriate level of context.

> When using Graylog Collectors, this address is used to receive heartbeat messages and must be reachable by all collectors.

lennartkoopmann:
I like it! @jochen?

lennartkoopmann:
I like it! @joschi?

joschi:
@Spitfire1900 Thanks! I've made the change with a slightly different wording and without changing the first line.

PR #1849, opened 2016-02-23
Pluggable field analyzer

bernd:
* Replace hard coded list of field analyzers with a pluggable implementation.
* Convert analyzer buttons to a dropdown menu. (see screenshot below)

aeoy2egruu9iaaaaaelftksuqmcc


lennartkoopmann:
Nice! Great stuff.

i think we need to change the button that triggers the dropdown though. The arrow is not indicating that it would open a dropdown. Any ideas? I also need to think about that. :)

lennartkoopmann:
My suggestion is to get rid of that arrow icon and make the field name itself a link that triggers the dropdown.

bernd:
> My suggestion is to get rid of that arrow icon and make the field name itself a link that triggers the dropdown.

I agree that the current icon for the dropdown is not optimal. But if we make the name itself the dropdown trigger, the behavior changes. That would mean the field can only be shown in the search result if the checkbox is clicked.

bernd:
That said, please let's create a separate issue for the button changes so we can merge this ASAP. Thanks! :smiley:

graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
```diff
@@ -14,7 +14,15 @@ import RefreshStore from 'stores/tools/RefreshStore';
...
-    permissions: PropTypes.arrayOf(React.PropTypes.string),
+    from: PropTypes.any.isRequired,
+    to: PropTypes.any.isRequired,
+    resolution: PropTypes.any.isRequired,
+    stream: PropTypes.object,
+    permissions: PropTypes.arrayOf(PropTypes.string).isRequired,
+    query: PropTypes.string.isRequired,
+    page: PropTypes.number.isRequired,
+    rangeType: PropTypes.string.isRequired,
+    rangeParams: PropTypes.object.isRequired,
```
edmundoa: I'd vote for only declare in the component's `propTypes` the props that are actually used, as it makes more clear what is actually used by the plugin. In my opinion we should leave the declaration of what is passed to each kind of plugin to the documentation that we need to write.

graylog2-web-interface/src/components/search/SearchResult.jsx
```diff
@@ -195,23 +215,14 @@ const SearchResult = React.createClass({
           />
         </Col>
         <Col md={9} sm={12} id="main-content-sidebar">
-          <FieldStatistics ref="fieldStatisticsComponent"
-                           permissions={this.props.permissions}/>
-
-          <FieldQuickValues ref="fieldQuickValuesComponent"
-                            permissions={this.props.permissions}/>
+          {this._fieldAnalyzerComponents((analyzer) => analyzer.displayPriority >= 100)}
```
edmundoa: Can we please remove the magic number from here? Just to gain some clarity.

graylog2-web-interface/src/components/search/SearchResult.jsx
```diff
@@ -195,23 +215,14 @@ const SearchResult = React.createClass({
...
                            permissions={this.props.permissions}
                            isStreamSearch={this.props.searchInStream !== null}/>
 
-          <FieldGraphs ref="fieldGraphsComponent"
-                       resolution={this.props.histogram.interval}
-                       from={this.props.histogram.histogram_boundaries.from}
-                       to={this.props.histogram.histogram_boundaries.to}
-                       permissions={this.props.permissions}
-                       searchInStream={this.props.searchInStream}/>
+          {this._fieldAnalyzerComponents((analyzer) => analyzer.displayPriority < 100)}
```
edmundoa: Same as above :)

graylog2-web-interface/src/components/search/SidebarMessageField.jsx
```diff
@@ -1,12 +1,11 @@
 import React, {PropTypes} from 'react';
-import {Input, Button, ButtonGroup} from 'react-bootstrap';
+import {Input, Button, ButtonGroup, DropdownButton, MenuItem} from 'react-bootstrap';
```
edmundoa: Button and ButtonGroup are not used any more, we can get rid of them and the lint error.

graylog2-web-interface/src/components/search/SearchResult.jsx
```diff
@@ -195,23 +215,14 @@ const SearchResult = React.createClass({
           />
         </Col>
         <Col md={9} sm={12} id="main-content-sidebar">
-          <FieldStatistics ref="fieldStatisticsComponent"
```
edmundoa: Can't add it to the import sentences (as they weren't modified by the changes in the PR), but we don't need to import `FieldGraphs`, `FieldQuickValues`, and `FieldStatistics` directly here any more.

graylog2-web-interface/src/components/search/SidebarMessageField.jsx
```diff
@@ -18,6 +17,27 @@ const SidebarMessageField = React.createClass({
...
+  },
+
+  _onFieldAnalyzer(refId, fieldName) {
+    return () => {
+      this.props.onFieldAnalyzer(refId, fieldName);
+    };
+  },
+
+  _fieldAnalyzerMenuItems() {
+    return this.props.fieldAnalyzers.map((analyzer, idx) => {
```
edmundoa: We need to sort the analysers by `displayName`.

edmundoa:
Other than those remarks, looks good to me. I'm seeing some issues with the dropdown when using it on fields close to the bottom of the screen (we already had that issue on the old-old web interface). I'll think a bit about how to display the arrow for the dropdown field.

graylog2-web-interface/src/components/search/SidebarMessageField.jsx
```diff
@@ -18,6 +17,27 @@ const SidebarMessageField = React.createClass({
   _toggleShowActions() {
     this.setState({showActions: !this.state.showActions});
   },
+  _setShowActions(isOpen) {
+    this.setState({showActions: isOpen});
+  },
```
edmundoa: Why do we need `_setShowActions` in addition to `_toggleShowActions`?

edmundoa:
I was playing a bit with some ideas I had for displaying the dropdown (actually, I want to avoid using a dropdown if possible), but they will require quite some time to make them look good. I think we can take care of it later.

graylog2-web-interface/src/components/search/SidebarMessageField.jsx
bernd: Because `_setShowOptions` receives the `isOpen` value and this is needed to have sane behaviour of the arrow icon when closing the dropdown by clicking somewhere else or pressing escape. :wink:

graylog2-web-interface/src/components/search/SearchResult.jsx
bernd: Imports removed. :+1:

bernd:
Rebased against master and updated after comments.

graylog2-web-interface/src/components/search/SidebarMessageField.jsx
edmundoa: In that case we can remove `_toggleShowActions`, as the toggle will be controller by the `DropdownButton`, and will update the state using `_setShowActions`.

bernd: Heh, true. Fixed! Thank you! :smile:

edmundoa:
LGTM :+1:

PR #1850, opened 2016-02-23
Allow disabling embedded web interface with setting "web_enable"

joschi:
This PR adds the configuration setting `web_enable` which allows users to disable the embedded web interface on a Graylog node.


bernd:
LGTM :+1:

PR #1854, opened 2016-02-24
Allow plugins to provide system configuration components

bernd:
The provided config components will show up at System/Configurations.


edmundoa:
LGTM :+1:

PR #1860, opened 2016-02-25
Add separator and header to system menu if plugins are present

bernd:
This adds a separator and header to the system menu to make it easy to see which menu entries are from plugins.

Do we want this? Please comment!

l5ruzoopraaaaaabjru5erkjggg


bernd:
We decided against doing this.

PR #1862, opened 2016-02-25
Fix potential memory leak in Tools#decompressGzip() and Tools#decompressZlib()

### LINKED ISSUE #1857: Graylog server does not properly close zlib/gzip streams.
As detailed [in this article](http://www.evanjones.ca/java-native-leak-bug.html), `InflaterInputStream` and `GZIPInputStream` allocate native memory that is not freed until `close()` or when the finalizer runs. 

It is therefore best to `close()` these streams explicitly. This is currently not the case in the Graylog code base for [decompressZlib](https://github.com/Graylog2/graylog2-server/blob/c20874ec9dae79ca71794f114df4170d3ecd8292/graylog2-server/src/main/java/org/graylog2/plugin/Tools.java#L187) and [decompressGzip](https://github.com/Graylog2/graylog2-server/blob/c20874ec9dae79ca71794f114df4170d3ecd8292/graylog2-server/src/main/java/org/graylog2/plugin/Tools.java#L202).

It can therefore be the case, that the graylog server uses more memory than the configured heap size. This is a potential risk for production installation as the server might be killed due to out of memory conditions.

joschi:
Properly close streams in `Tools#decompressGzip()` and `Tools#decompressZlib()` by using try-with-resources.

Fixes #1857


StephanErb:
Change looks good. Thanks for the quick reaction!

bernd:
LGTM :+1:

PR #1863, opened 2016-02-25
Add classForName to ChainableClassLoader and make it injectable

bernd:
This is needed to make `Class.forName()` with classes from external plugins work. For example cluster config classes from external jar plugins.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
@@ -121,8 +121,4 @@ public class ChainingClassLoader extends ClassLoader {
     public boolean addClassLoader(ClassLoader classLoader) {
         return classLoaders.add(classLoader);
     }
-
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/shared/plugins/ChainingClassLoader.java
```diff
@@ -121,4 +121,8 @@ public InputStream getResourceAsStream(String name) {
     public boolean addClassLoader(ClassLoader classLoader) {
         return classLoaders.add(classLoader);
     }
+
+    public Class<?> classForName(String className) throws ClassNotFoundException {
```
joschi: Is this method really necessary? Wouldn't a call to [`ClassLoader#loadClass(String)`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#loadClass-java.lang.String-) suffice?

bernd: Good point! Yes, that's enough. :+1:

joschi:
LGTM. :+1:

PR #1864, opened 2016-02-25
Do not cache widget plugins to avoid problems with external plugins

bernd:
The global "plugins" variable might not be completely initialized yet because plugin JavaScript files have not been loaded.

Fixes an issue where a field analyzer/widget from a plugin does not work because the plugin widget is not in `plugins` yet.


edmundoa:
LGTM :+1:

PR #1877, opened 2016-03-01
Add REST API endpoints and UI to configure message processors

bernd:
- Add MessageProcessor.Descriptor interface
- Refactor config object
- Add REST endpoints for message processors configuration
- Add UI to enable/disable processors and to re-order them
- Change active processors order when the config changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfig.java
+++ b/graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfig.java
@@ -1,0 +1,92 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.messageprocessors;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+import java.util.ArrayList;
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfigWithDescriptors.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public static Builder builder() {
+        return new AutoValue_MessageProcessorsConfigWithDescriptors.Builder();
+    }
+
+    public abstract Builder toBuilder();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
```
joschi: Is having a builder really necessary here? In the end this class just has 2 fields which both have to be populated anyway.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfig.java
```diff
@@ -0,0 +1,92 @@
...
+        availableProcessors.stream()
+                .filter(processor -> !newOrder.contains(processor))
+                .sorted(String.CASE_INSENSITIVE_ORDER)
+                .forEach(newOrder::add);
+
+        return toBuilder().processorOrder(newOrder).build();
+    }
+
+    @AutoValue.Builder
+    public static abstract class Builder {
```
joschi: Is having a builder really necessary here? In the end this class just has 2 fields which both have to be populated anyway.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfigWithDescriptors.java
bernd: They are not necessary, but I like to have AutoValue objects with builders because once you want to modify those objects it gets painful with only a `create` method. (see `withProcessors`)

In this case they are not really necessary, but I would like to keep them. :smiley:

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfig.java
bernd: See above.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfigWithDescriptors.java
joschi: :white_check_mark:

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorsConfig.java
joschi: :white_check_mark:

PR #1878, opened 2016-03-02
Make rest permissions extensible via plugins

bernd:
This allows a plugin to extend the rest permissions. The plugin permissions will show up in `/system/permissions`.

I only implemented what's needed right now. The core permissions have not been converted to `RestPermission` objects yet and there is no API endpoint that shows the full `RestPermission` objects.

Example:

```java
public class CollectorRestPermissions implements RestPermissionsPlugin {
    public static final String COLLECTORS_READ = "collectors:read";
    public static final String COLLECTORS_CREATE = "collectors:create";
    public static final String COLLECTORS_UPDATE = "collectors:update";
    public static final String COLLECTORS_DELETE = "collectors:delete";

    private final ImmutableSet<RestPermission> permissions = ImmutableSet.of(
            RestPermission.create(COLLECTORS_READ, "Read collectors"),
            RestPermission.create(COLLECTORS_CREATE, "Create collectors"),
            RestPermission.create(COLLECTORS_UPDATE, "Update collectors"),
            RestPermission.create(COLLECTORS_DELETE, "Delete collectors")
    );

    @Override
    public Set<RestPermission> permissions() {
        return permissions;
    }
}
```

```java
public class CollectorModule extends PluginModule {
    @Override
    protected void configure() {
        addRestPermissions(CollectorRestPermissions.class);
   }
}
```

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermission.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermission.java
@@ -1,0 +1,56 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.plugin.security;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class RestPermission {
+    @JsonProperty("value")
+    public abstract String value();
+
+    @JsonProperty("description")
+    public abstract String description();
+
+    @JsonCreator
+    public static RestPermission create(@JsonProperty("value") String value,
+                                        @JsonProperty("description") String description) {
+        return builder()
+                .value(value)
+                .description(description)
+                .build();
+    }
+
+    public static Builder builder() {
+        return new AutoValue_RestPermission.Builder();
+    }
+
+    public abstract Builder toBuilder();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+        public abstract Builder value(String value);
+
+        public abstract Builder description(String description);
+
+        public abstract RestPermission build();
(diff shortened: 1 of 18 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermission.java
```diff
@@ -0,0 +1,56 @@
...
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class RestPermission {
+    @JsonProperty("value")
+    public abstract String value();
```
joschi: `value` seems a bit generic. How about `permission`?

graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermission.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @JsonProperty("description")
+    public abstract String description();
+
+    @JsonCreator
+    public static RestPermission create(@JsonProperty("value") String value,
```
joschi: Maybe adding `@NotBlank` would make sense here to find invalid permissions early.

graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermissionsPlugin.java
```diff
@@ -0,0 +1,23 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.plugin.security;
+
+import java.util.Set;
+
+public interface RestPermissionsPlugin {
```
joschi: I think this class name is a little bit misleading. It's not a plugin, but it's a set of permissions provided by a plugin. So `PluginPermissions` or `PluginRestPermissions` would probably make more sense.

graylog2-server/src/main/java/org/graylog2/shared/security/RestPermissions.java
```diff
@@ -150,33 +155,23 @@
...
-                final Iterator<String> split = Splitter.on(':').limit(2).split(permission).iterator();
-                final String group = split.next();
-                final String action = split.next();
-                all.put(group, action);
-            } catch (IllegalAccessException ignored) {
-            }
-        }
-        ALL_PERMISSIONS = all.asMap();
+    public RestPermissions() {
+        this(new HashSet<>());
```
joschi: Why not `Collections.emptySet()`?

graylog2-server/src/main/java/org/graylog2/plugin/security/RestPermission.java
```diff
@@ -0,0 +1,56 @@
...
+                                        @JsonProperty("description") String description) {
+        return builder()
+                .value(value)
+                .description(description)
+                .build();
+    }
+
+    public static Builder builder() {
```
joschi: Do we really need the builder given that this class only has 2 private fields and that both are supposed to be set anyway?

bernd:
Updated after feedback.

joschi:
LGTM. :+1:

PR #1879, opened 2016-03-02
Add CSV export again.

dennisoelkers:
This change set adds the CSV export functionality in the web interface
again. Due to the obsoletion of the play web interface middleware, we
need some more support from the backend, which now offers a separate
REST call for each search type which defaults to returning CSV and adds
the necessary headers to the response which indicate a file download and
a proper filename to the browser. All previously existing methods still
work the exact same way as before.


edmundoa:
LGTM :+1:

PR #1886, opened 2016-03-03
Remove collector related code from server

bernd:
- Config file options
- REST API endpoints
- Database related code
- Background jobs
- REST permissions
- CSS code
- Web components and pages
    
All of this will be moved into a plugin.


bernd:
Merging this so we can move on. Tested by myself. :smile:

PR #1894, opened 2016-03-03
Simplify cluster calls in remote resources

dennisoelkers:
This change set adds some helper methods to simplify writing resource methods that call the same remote resource method on all nodes in the cluster.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
@@ -89,7 +89,7 @@ public abstract class ProxiedResource extends RestResource {
                 final Node targetNode = nodeService.byNodeId(nodeId);
                 return Optional.of(this.remoteInterfaceProvider.get(targetNode, this.authenticationToken, interfaceClass));
             } catch (NodeNotFoundException e) {
-                LOG.warn("Node <" + nodeId + "> not found while trying to call " + interfaceClass.getName() + " on it.")
+                LOG.warn("Node <" + nodeId + "> not found while trying to call " + interfaceClass.getName() + " on it.");
                 return Optional.empty();
             }
         };
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
@@ -17,19 +17,73 @@
 
 package org.graylog2.shared.rest.resources;
 
+import org.graylog2.cluster.Node;
+import org.graylog2.cluster.NodeNotFoundException;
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -32,4 +53,45 @@ public ProxiedResource(@Context HttpHeaders httpHeaders) {
...
     }
+
+    protected <T, K> Map<String, Optional<K>> getForAllNodes(Function<T, Call<K>> fn, Function<String, Optional<T>> interfaceProvider) {
+        return getForAllNodes(fn, interfaceProvider, Function.identity());
+    }
+
+    protected <T, K, L> Map<String, Optional<K>> getForAllNodes(Function<T, Call<L>> fn, Function<String, Optional<T>> interfaceProvider, Function<L, K> transformer) {
+        return this.nodeService.allActive()
+            .entrySet()
+            .stream()
```
bernd: The previous `.stream` operations in the resources used `.parallel`, this doesn't. Is that intentional?

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -32,4 +53,45 @@ public ProxiedResource(@Context HttpHeaders httpHeaders) {
             this.authenticationToken = null;
         }
     }
+
+    protected <T, K> Map<String, Optional<K>> getForAllNodes(Function<T, Call<K>> fn, Function<String, Optional<T>> interfaceProvider) {
+        return getForAllNodes(fn, interfaceProvider, Function.identity());
+    }
+
+    protected <T, K, L> Map<String, Optional<K>> getForAllNodes(Function<T, Call<L>> fn, Function<String, Optional<T>> interfaceProvider, Function<L, K> transformer) {
```
bernd: It would be nice if you could use longer names for the types, T K L is a bit hard to handle while reading the code. :smile:

dennisoelkers: True :)

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
dennisoelkers: Hm, somehow the comment I had already written was either not saved or got lost.

I left it out for now, because it is inconsistently used and I am not 100% sure that we can exclude negative implications. The benefit of this change though is that we have a central point to switch it on/off if we want to.

bernd: :+1:

bernd:
LGTM :+1:

PR #1913, opened 2016-03-08
Upgrade to Swagger 1.5.7

joschi:
This PR upgrades the Swagger Annotations in Graylog to the latest supported version.

By doing this, migrating from our custom Swagger generator to the officially supported Jersey Swagger support should be easier in the future and we might even be able to support the [OpenAPI specification](https://openapis.org/) at one point.

On the downside, this would also require plugins to switch from the old annotations (`com.wordnik.swagger.*`) to the new ones (`io.swagger.*`).


kroepke:
lgtm, should merge the swagger-update branches afterwards

PR #1916, opened 2016-03-09
addField allows Date objects for timestamp fields

kroepke:
this leads to all kind of sadness later on: we really depend on timestamp being a DateTime instance
Enforce a DateTime in case it was a Date


bernd:
LGTM :+1:

PR #1917, opened 2016-03-09
Search for surrounding messages

bernd:
There is a new button to search for surrounding messages. The time range options can be configured at "System/Configurations".

8frudyza qct0aaaaasuvork5cyii

By default, the search is executed with the `source` and `gl2_source_input` fields as a filter. The filter fields are also configurable at "System/Configurations".

weze lb7gqbdgaaaabjru5erkjggg

The original message will be highlighted with a green bar so it's easier to find it.

hwpfuvk8qhfjaaaaaelftksuqmcc


bernd:
The first commit contains some refactoring so it might be easier to review the individual commits.

graylog2-web-interface/src/components/common/ISODurationUtils.js
```diff
@@ -0,0 +1,34 @@
+import moment from 'moment';
```
edmundoa: I would move this file with the other utilities, into the `util` folder.

graylog2-web-interface/src/components/common/ISODurationInput.jsx
```diff
@@ -14,6 +14,8 @@ const ISODurationInput = React.createClass({
     required: React.PropTypes.bool,
   },
 
+  mixins: [ISODurationUtils],
```
edmundoa: Why are we using `ISODurationUtils` as a mixin? I think it makes less obvious where the methods are coming from, and in this case the mixin is not using any hooks or working with the component state.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -258,27 +154,32 @@ const SearchesConfig = React.createClass({
...
-                   value={config.query_time_range_limit}
-                   help={'The maximum time range for searches. (i.e. "P30D" for 30 days, "PT24H" for 24 hours)'}
-                   addonAfter={this._formatDuration(this.state.config.query_time_range_limit)}
-                   bsStyle={this._validationState()}
-                   required/>
+            <ISODurationInput duration={config.query_time_range_limit}
+                              update={this._onUpdate('query_time_range_limit')}
+                              label="Query time range limit (ISO8601 Duration)"
+                              help={'The maximum time range for searches. (i.e. "P30D" for 30 days, "PT24H" for 24 hours)'}
+                              validator={(milliseconds) => milliseconds >= 1}
```
edmundoa: Adding arrow functions as props is discouraged, and in this case we should not use it: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

I don't know what functions we use as validators at the moment, but maybe it would make sense to provide some common validators in the `IsoDurationUtils`.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -258,27 +154,32 @@ const SearchesConfig = React.createClass({
...
-                Configure the available options for the <strong>relative</strong> time range selector as <strong>ISO8601 duration</strong>.
-              </span>
-              <div className="wrapper">
-                {timerangeOptions}
-              </div>
-              <Button bsSize="xs" onClick={this._onTimerangeOptionAdd}>Add option</Button>
-            </div>
+            <TimeRangeOptionsForm options={this.state.config.relative_timerange_options}
+                                  update={this._onUpdate('relative_timerange_options')}
+                                  validator={(milliseconds, dur) => milliseconds >= 1 || dur === 'PT0S'}
```
edmundoa: Same as above.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -258,27 +154,32 @@ const SearchesConfig = React.createClass({
...
+                                  title="Relative Timerange Options"
+                                  help={<span>Configure the available options for the <strong>relative</strong> time range selector as <strong>ISO8601 duration</strong></span>} />
+
+            <TimeRangeOptionsForm options={this.state.config.surrounding_timerange_options}
+                                  update={this._onUpdate('surrounding_timerange_options')}
+                                  validator={(milliseconds) => milliseconds >= 1}
```
edmundoa: Same as above.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -258,27 +154,32 @@ const SearchesConfig = React.createClass({
...
-                   onChange={this._onPeriodUpdate('query_time_range_limit')}
-                   value={config.query_time_range_limit}
-                   help={'The maximum time range for searches. (i.e. "P30D" for 30 days, "PT24H" for 24 hours)'}
-                   addonAfter={this._formatDuration(this.state.config.query_time_range_limit)}
-                   bsStyle={this._validationState()}
-                   required/>
+            <ISODurationInput duration={config.query_time_range_limit}
+                              update={this._onUpdate('query_time_range_limit')}
+                              label="Query time range limit (ISO8601 Duration)"
+                              help={'The maximum time range for searches. (i.e. "P30D" for 30 days, "PT24H" for 24 hours)'}
```
edmundoa: Here we could just pass the string directly, as we don't do anything fancy with the text.

graylog2-web-interface/src/components/configurations/TimeRangeOptionsForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+
+  _buildTimeRangeOptions() {
+    return this.state.optionsList.map((option, idx) => {
+      const period = option.period;
+      const description = option.description;
+      const errorStyle = this.durationStyle(period, this.props.validator, 'has-error');
+
+      return (
+        <div key={'timerange-option-' + idx}>
```
edmundoa: This is a bit of nitpicking, but just to follow the latest eslint rules, we should use templates for string concatenation: http://eslint.org/docs/rules/prefer-template

graylog2-web-interface/src/components/configurations/TimeRangeOptionsForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+    this.setState({optionsList: this._buildOptionsState(newProps.options)});
+  },
+
+  // Converts the object to a list of objects to make form handling easier.
+  _buildOptionsState(options) {
+    let optionsList = null;
+
+    if (options) {
+      optionsList = Object.keys(options).map((key) => {
+        return {period: key, description: options[key]};
```
edmundoa: More nitpicking: http://eslint.org/docs/rules/object-curly-spacing

graylog2-web-interface/src/components/configurations/TimeRangeOptionsForm.jsx
```diff
@@ -0,0 +1,151 @@
...
+    });
+
+    this.props.update(update);
+  },
+
+  _onAdd() {
+    const options = ObjectUtils.clone(this.state.optionsList);
+
+    if (options) {
+      options.push({period: 'PT0S', description: ''});
```
edmundoa: Same as above.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
```diff
@@ -258,27 +154,32 @@ const SearchesConfig = React.createClass({
...
+                                  update={this._onUpdate('surrounding_timerange_options')}
+                                  validator={(milliseconds) => milliseconds >= 1}
+                                  title="Surrounding Timerange Options"
+                                  help={<span>Configure the available options for the <strong>surrounding</strong> time range selector as <strong>ISO8601 duration</strong></span>} />
+
+            <Input type="text"
+                   label="Surrounding search filter fields"
+                   onChange={this._onFilterFieldsUpdate}
+                   value={this.state.surroundingFilterFields || filterFieldsString}
+                   help="foo bar"
```
edmundoa: Foo bar :trollface:

graylog2-web-interface/src/components/search/SurroundingSearchButton.jsx
```diff
@@ -0,0 +1,65 @@
...
+    });
+
+    return options;
+  },
+
+  _buildFilterFields() {
+    const fields = {};
+
+    if (this.props.searchConfig) {
+      this.props.searchConfig.surrounding_filter_fields.forEach((field) => fields[field] = this.props.messageFields[field]);
```
edmundoa: Another eslint issue: http://eslint.org/docs/rules/no-return-assign

graylog2-web-interface/src/components/configurations/TimeRangeOptionsForm.jsx
edmundoa: On the second commit, the default period changed from 'PT1S' to 'PT0S', and that makes two things go wrong:

1. Adding an option to the relative range options ends in a no-result, as the `PT0S` option is already there
2. Adding an option to the time range options makes the default value to be invalid (and displaying the field in red)

I think we shouldn't use the period as key in this case, because it makes hard to pick default values, and also merges options that use the same period as another one had. Even if it makes no sense having two options with the same period, they should still have their own field where we should indicate this situation.

edmundoa:
Saw another issue that I could reproduce like this:
1. Refresh search page
2. Go to a page not using the search bar (i.e. Streams)
3. Go back to search page and have this in the console:
`Warning: Failed propType: Required prop 'searchConfig' was not specified in 'SearchPage'. Check the render method of 'ReactProxy'.`

graylog2-web-interface/src/components/search/SurroundingSearchButton.jsx
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    return fields;
+  },
+
+  _onClick(range) {
+    return (e) => {
+      e.preventDefault();
+
+      const fromTime = moment.unix(this.props.timestamp - parseInt(range, 10)).toISOString();
```
edmundoa: What do we want to achieve with `parseInt`? I think we only use to convert the string `range` into a number, but then we should probably use `Number()` because of this:

> “parseInt() is useful for parsing strings, but you have to be aware that it stops at the first illegal digit. Parsing strings via Number() is less forgiving, but may produce nonintegers.”

> Excerpt From: Axel Rauschmayer. “Speaking JavaScript.”

graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
```diff
@@ -44,6 +46,19 @@
             .put(Period.days(30), "Search in the last 30 days")
             .put(Period.ZERO, "Search in all messages")
             .build();
+    private static final Map<Period, String> DEFAULT_SURROUNDING_TIMERANGE_OPTIONS = ImmutableMap.<Period, String>builder()
+            .put(Period.seconds(1), "1 second")
+            .put(Period.seconds(5), "5 seconds")
+            .put(Period.seconds(10), "10 seconds")
+            .put(Period.minutes(1), "1 minute")
```
edmundoa: I think we need at least one option between 10 seconds and 1 minute, the gap seems to big to me. How about adding 30 seconds by default?

graylog2-web-interface/src/components/configurations/TimeRangeOptionsForm.jsx
edmundoa: The issue I mentioned with the period as keys is actually worse than I thought:

As this is updating the state on the `onChange` of the text input, if while typing your duration is the same as another one, it will merge and the input disappears. For instance, using the default 'PT0S', replacing the 0 with 10 will make this happen.

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
bernd: It's because the string contains `"`. What else can I do?

bernd:
@edmundoa Updated after comments.

Also fixed the `Required prop 'searchConfig' was not specified in 'SearchPage'` issue.

Any alternative here? https://github.com/Graylog2/graylog2-server/pull/1917#discussion_r55827270

graylog2-web-interface/src/components/configurations/SearchesConfig.jsx
edmundoa: Ah, didn't see that. I think it's fine then, otherwise eslint would complain about it. I guess the lesson to learn is that there's always someone complaining :smile:

edmundoa:
LGTM :+1:

PR #1921, opened 2016-03-11
Alert annotations

edmundoa:
Add alert annotations back to the search histogram with some changes:

- Global search will show alerts for all streams, just as before
- Stream search will only show alerts for the current stream
- Alerts will refresh along with the rest of the search page, when using the refresh search button


graylog2-web-interface/src/logic/alerts/AlertsAnnotator.js
```diff
@@ -0,0 +1,57 @@
...
+
+    alerts.forEach(alert => {
+      const epoch = DateTime.fromUTCDateTime(alert.triggered_at).toMoment().unix();
+      annotator.add(epoch, this._getAlertAnnotation(alert));
+      annotator.update();
+    });
+  },
+
+  _getAlertAnnotation(alert) {
+    const stream = this.streams.filter(s => s.id === alert.stream_id)[0];
```
dennisoelkers: A check if `this.streams` is defined already is required here.

PR #1930, opened 2016-03-15
When loading web interface assets also handle jar URIs

### LINKED ISSUE #1924: Web interface not loading: java.lang.IllegalArgumentException: URI is not hierarchical
### Problem description

Compiled a fresh `graylog-server` from current `master`. Opening the web interface in Chrome on OSX returns an empty/white page and the JS console shows that `GET` for `vendor.js` and `app.5eed8f45f4bee35f56f5.js ` failed with an `HTTP 500`.

Note that the comments in this commit show someone else having the same problem: 515c3a242a3fdff1dcb3a4a2fece8351d0060c02

See description.

### Environment

* Graylog Version: `master` / d45f72d1e3f21294a55adc54e2f2dbe8088c7949
* Elasticsearch Version: 2.2.0
* MongoDB Version: 3.0.0
* Operating System: OSX 10.11.2
* Browser version: Chrome 48.0.2564.116 (64-bit)

kroepke:
After assembly we cannot use File objects to access individual asset files. Instead we need to use a JarFileSystem to access the file metadata.

We could also cache the etag and modification dates in case of jars, but this change currently does not do that. We may want to revisit this in the future.
The patch uses a LoadingCache as a simple "lock" when creating the filesystem (uris will overlap and reference mostly the shaded jar, the cache makes sure only one thread is creating the filesystem).

fix #1924


lennartkoopmann:
Thanks! :+1: :)

PR #1937, opened 2016-03-16
Fix import of Graylog 1.x extractors and fix export format

### LINKED ISSUE #1831: Clarify extractor compatibility
From the mailing list, there is a report of a change in the extractors structure. Extractors exported from versions pre 2.0, had an `extractor_type` field, whereas in 2.0 that field is called `type`.

We should verify that we support old extractors or clarify that in the documentation, probably even notify plugin creators.

bernd:
The import/export format wrangling should be done on the server side instead of doing all of this in the browser.

This is a minimal fix because refactoring this is too much at the moment.

Fixes #1831


edmundoa:
LGTM :+1:

PR #1938, opened 2016-03-16
Let field charts use the stream and time range of the current search

### LINKED ISSUE #1785: Queries of field charts in streams view leave scope of stream
This is very misleading as a user would probably expect them to append the scope of the stream to their queries.

bernd:
Before this, stream and the time range settings have been taken from the
stored values in localStorage. This made the graphs stay the same even
if the user changed the time range options or searched in a different
stream.

Fixes #1785
Fixes Graylog2/graylog2-web-interface#1620
Fixes Graylog2/graylog2-web-interface#1618
Fixes Graylog2/graylog2-web-interface#1485


edmundoa:
LGTM :+1:

PR #1944, opened 2016-03-17
Improve user authorization

bernd:
* Fix user name in authorization error log messages and add HTTP verb and path to ease debugging.
* Do not log a stack trace on authorization error.
* Disable annoying AbstractValidatingSessionManager INFO warnings in log4j2 config.
* Fix HTTP status code for authorization errors. (return 403 instead of 401)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroAuthorizationFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroAuthorizationFilter.java
@@ -18,7 +18,7 @@ package org.graylog2.shared.security;
 
 import org.apache.shiro.authz.AuthorizationException;
 import org.apache.shiro.authz.annotation.RequiresPermissions;
-import org.apache.shiro.subject.Subject;
+import org.graylog2.rest.RestTools;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/shared/security/ShiroAuthorizationFilter.java
```diff
@@ -44,17 +45,19 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+            final String userName = RestTools.getUserNameFromRequest(requestContext);
             final ContextAwarePermissionAnnotationHandler annotationHandler = new ContextAwarePermissionAnnotationHandler(context);
             try {
-                LOG.debug("Checking authorization for user {}, needs permissions {}", subject, annotation.value());
+                LOG.debug("Checking authorization for user [{}], needs permissions: {}", userName, annotation.value());
                 annotationHandler.assertAuthorized(annotation);
             } catch (AuthorizationException e) {
-                LOG.info("User " + subject + "not authorized.", e);
-                throw new NotAuthorizedException(e, "Basic realm=\"Graylog Server\"");
+                final String msg = String.format(Locale.US, "User [%s] not authorized. (%s %s)", userName,
```
edmundoa: Just wondering: why do we need to use `String.format` for this?

bernd: I just wanted to avoid using `+` to concatenate lots of strings. :wink:

edmundoa:
LGTM :+1:

PR #1948, opened 2016-03-18
Improve cluster event bus API

bernd:
Previously you had to publish cluster events on the cluster-event-bus and the internal server-event-bus to make sure the producer also received the event. This was unintuitive and lead to wrong usage where objects subscribed to the cluster-event-bus where they should have subscribed to the server-event-bus instead and thus not getting cluster events when running on a different server node.

With this change, the cluster event bus should be used as follows:

- If an object needs to publish cluster events, it uses the `ClusterEventBus` implementation to publish the events. It does NOT have to publish the event on the internal server event bus!
- If an object needs to receive cluster events, it registers with the internal server event bus.
- The `ClusterEventBus#register()` method throws an exception and should not be used by regular event subscribers. There is a `#registerClusterEventSubscriber()` method to actually subscribe to the cluster event bus events. This should only be used by the cluster event system internals.

Previously the event producer got the event via its internal server event bus and not from the database. (that's why a cluster event had to be published on the cluster-event-bus AND the server-event-bus)
The event producer is now the events the same way as any other nodes, it gets the event via the database.

Other changes:

- The DeadEvent handler is now logging to WARN to make sure we see unhandled events.
- Do not publish ThrottleState events in ThrottleStateUpdater because no object is subscribed to them.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
@@ -16,7 +16,6 @@
  */
 package org.graylog2.bindings;
 
-import com.google.common.eventbus.EventBus;
 import com.google.inject.Scopes;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
 import com.google.inject.multibindings.Multibinder;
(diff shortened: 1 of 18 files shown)


bernd:
The map-widget and pipeline-processor plugins need an update once this is merged.

(I already have diffs prepared)

joschi:
:+1:

PR #1949, opened 2016-03-18
Add plugin navigation permissions

edmundoa:
Check if the plugin provides a `permissions` key on the `navigation` and `systemnavigation` objects it provides, and check that before showing the navigation entry.

Example:
```
systemnavigation: [
  { path: '/system/pipelines', description: 'Pipelines', permissions: 'INPUTS_CREATE' },
],
```
Will only show the pipelines menu if the user can create inputs.


bernd:
LGTM :+1:

Issue #1100, opened 2015-04-08
Yellow status when journal is almost full


pdepaepe:
Hello,

May you please consider to cast a yellow status on /system/lbstatus/ when journal is almost full (~90%)?

Thanks.

dennusb:
+1

eorochena:
+1

PR #1952, opened 2016-03-20
THROTTLED status when journal utilization is over threshold

mikkolehtisalo:
Implementation of #1100. 

Quick notes:
- The feature is disabled by default
- Adds THROTTLED load balancer status with 429 TOO MANY REQUESTS (meant for throttling purposes).
- Evaluates the need to throttle at journal cleanups (alternatively could be also at metrics)
- Threshold percentage configured via new configuration value lb_throttle_threshold_percentage
- Should not break existing load balancer configurations, because they commonly consider anything besides ALIVE response as DEAD
- Needs to be documented [here](https://github.com/Graylog2/documentation/blob/master/pages/load_balancers.rst), if gets accepted

Please review and comment :)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -105,6 +105,9 @@ public class Configuration extends BaseConfiguration {
     @Parameter(value = "lb_recognition_period_seconds", validator = PositiveIntegerValidator.class)
     private int loadBalancerRecognitionPeriodSeconds = 3;
 
+    @Parameter(value = "lb_throttle_threshold_percentage", validator = PositiveIntegerValidator.class)
+    private int loadBalancerThrottleThresholdPercentage = 100; // 100% or over is effectively disabled
+
     @Parameter(value = "stream_processing_timeout", validator = PositiveLongValidator.class)
     private long streamProcessingTimeout = 2000;
 
@@ -314,4 +317,7 @@ public class Configuration extends BaseConfiguration {
     }
 
     public Set<IpSubnet> getTrustedProxies() { return trustedProxies; }
+
+    public int getLoadBalancerRequestThrottleJournalUsage() { return loadBalancerThrottleThresholdPercentage; }
+
 }
--- a/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
@@ -50,7 +50,9 @@ import kafka.utils.KafkaScheduler;
 import kafka.utils.Time;
 import kafka.utils.Utils;
 import org.graylog2.plugin.GlobalMetricNames;
+import org.graylog2.plugin.ServerStatus;
 import org.graylog2.plugin.ThrottleState;
+import org.graylog2.plugin.lifecycles.LoadBalancerStatus;
 import org.graylog2.shared.metrics.HdrTimer;
 import org.joda.time.DateTimeUtils;
 import org.joda.time.Duration;
@@ -102,6 +104,7 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
 
     public static final long DEFAULT_COMMITTED_OFFSET = Long.MIN_VALUE;
     public static final int NOTIFY_ON_UTILIZATION_PERCENTAGE = 95;
+    private final ServerStatus serverStatus;
 
     // this exists so we can use JodaTime's millis provider in tests.
     // kafka really only cares about the milliseconds() method in here
@@ -148,6 +151,8 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
     private final AtomicReference<ThrottleState> throttleState = new AtomicReference<>();
     private final AtomicInteger purgedSegmentsInLastRetention = new AtomicInteger();
 
+    private final double throttleThresholdPercentage;
+
     @Inject
     public KafkaJournal(@Named("message_journal_dir") File journalDirectory,
                         @Named("scheduler") ScheduledExecutorService scheduler,
@@ -157,8 +162,12 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
                         @Named("message_journal_max_age") Duration retentionAge,
                         @Named("message_journal_flush_interval") long flushInterval,
                         @Named("message_journal_flush_age") Duration flushAge,
-                        MetricRegistry metricRegistry) {
+                        @Named("lb_throttle_threshold_percentage") int throttleThresholdPercentage,
+                        MetricRegistry metricRegistry,
+                        ServerStatus serverStatus) {
         this.scheduler = scheduler;
+        this.throttleThresholdPercentage = throttleThresholdPercentage;
+        this.serverStatus = serverStatus;
 
         this.writtenMessages = metricRegistry.meter(name(this.getClass(), "writtenMessages"));
         this.readMessages = metricRegistry.meter(name(this.getClass(), "readMessages"));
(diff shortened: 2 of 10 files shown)


lennartkoopmann:
I did not review the code but I like the idea! Thanks Mikko.

bernd:
@mikkolehtisalo Thank you for the PR! :+1:  It's a bit late to put this in 2.0 unfortunately but we will review this for 2.1.

mikkolehtisalo:
Okay. I'll be glad to change it based on the feedback. I am still trying to catch on some things...

joschi:
@mikkolehtisalo Could you please rebase your PR on the current `master` branch?

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -157,8 +162,12 @@ public KafkaJournal(@Named("message_journal_dir") File journalDirectory,
                         @Named("message_journal_max_age") Duration retentionAge,
                         @Named("message_journal_flush_interval") long flushInterval,
                         @Named("message_journal_flush_age") Duration flushAge,
-                        MetricRegistry metricRegistry) {
+                        MetricRegistry metricRegistry,
+                        ServerStatus serverStatus,
+                        Configuration configuration) {
```
joschi: Please also use the named property here:
```java
@Named("lb_throttle_threshold_percentage") int throttleThresholdPercentage
```

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/LoadBalancerStatusResource.java
```diff
@@ -55,16 +55,40 @@ public LoadBalancerStatusResource(ServerStatus serverStatus) {
         this.serverStatus = serverStatus;
     }
 
+    /**
+     * Implementing StatusType for 429 TOO MANY REQUEST (the most suitable HTTP status code for
+     * requesting throttling), because javax.ws.rs.core.Response.Status does not implement it.
+     */
+    private class TooManyRequestsStatus implements Response.StatusType {
```
joschi: Please move this as a self-contained class into the `org.graylog2.rest` package.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/LoadBalancerStatusResource.java
```diff
@@ -55,16 +55,40 @@ public LoadBalancerStatusResource(ServerStatus serverStatus) {
...
+        @Override
+        public int getStatusCode() { return 429; }
+
+        @Override
+        public Response.Status.Family getFamily() { return Response.Status.Family.CLIENT_ERROR; }
+
+        @Override
+        public String getReasonPhrase() { return "TOO MANY REQUESTS"; }
```
joschi: "Too Many Requests" should be fine. No need to yell. 😉

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -101,6 +101,9 @@
     @Parameter(value = "lb_recognition_period_seconds", validator = PositiveIntegerValidator.class)
     private int loadBalancerRecognitionPeriodSeconds = 3;
 
+    @Parameter(value = "lb_throttle_threshold_percentage", validator = PositiveIntegerValidator.class)
+    private int loadBalancerThrottleThresholdPercentage = 110; // Over 100% is effectively disabled
```
joschi: I think setting it to 100 or even 95 by default is fine.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -102,6 +105,8 @@
 
     public static final long DEFAULT_COMMITTED_OFFSET = Long.MIN_VALUE;
     public static final int NOTIFY_ON_UTILIZATION_PERCENTAGE = 95;
+    private static ServerStatus serverStatus;
+    private static Configuration configuration;
```
joschi: Why use `static` variables here?

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -794,6 +803,31 @@ public Object apply(LogSegment segment) {
             return deletedSegments;
         }
 
+        /** Change the load balancer status from ALIVE to THROTTLE, or vice versa depending on the
+         * journal utilization percentage. As the utilization ratio is reliable only after cleanup,
+         * that's where this is called from.
+         */
+        private void updateLoadBalancerStatus(double utilizationPercentage) {
```
joschi: I'm a little bit uncomfortable with changing the server/load balancer status directly from this class.
An alternative could be emitting an event on the `EventBus` and a subscriber to that event which will change the `serverStatus` accordingly.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
mikkolehtisalo: Fixed.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/LoadBalancerStatusResource.java
mikkolehtisalo: Fixed.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/LoadBalancerStatusResource.java
mikkolehtisalo: Copy & paste... Fixed.

graylog2-server/src/main/java/org/graylog2/Configuration.java
mikkolehtisalo: Ok. Changed accordingly.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
mikkolehtisalo: Not static anymore.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
mikkolehtisalo: I wondered about this. How about just using ServerStatus.publishLifecycle? Other parts of code seemed to use that.

With events, probably should check the status and send only actual changes. I would feel uneasy about EventBus unless cases of direct use would be removed. Mixing the two cases might cause unforeseen consequences.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -157,8 +162,12 @@ public KafkaJournal(@Named("message_journal_dir") File journalDirectory,
                         @Named("message_journal_max_age") Duration retentionAge,
                         @Named("message_journal_flush_interval") long flushInterval,
                         @Named("message_journal_flush_age") Duration flushAge,
-                        MetricRegistry metricRegistry) {
+                        @Named("lb_throttle_threshold_percentage") double throttleThresholdPercentage,
```
joschi: This has to have the same type as in the configuration class (i. e. `int`).

mikkolehtisalo: Changed to int.

joschi:
@mikkolehtisalo Any reason you've closed this PR and deleted your branch? I think it was very useful.

mikkolehtisalo:
I think the idea might not be worth pursuit, because it increases complexity. However, feel free to work on it if you think it is worth the time.

joschi:
@mikkolehtisalo I've continued to work on this in #2312, so I'll close this PR again. Comments in the new PR are welcome!

PR #1954, opened 2016-03-21
Use cluster event bus for grok pattern changes

kroepke:
Introduces new GrokPatternRegistry to hand out compile Grok patterns, instead of having consumers reload them themselves (which can cause stalls if not done properly).

This PR doesn't change the current extractors on purpose to minimize changes in legacy code.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
@@ -46,6 +46,7 @@ import org.graylog2.events.ClusterEventBus;
 import org.graylog2.bindings.providers.ClusterEventBusProvider;
 import org.graylog2.filters.FilterService;
 import org.graylog2.filters.FilterServiceImpl;
+import org.graylog2.grok.GrokPatternRegistry;
 import org.graylog2.indexer.SetIndexReadOnlyJob;
 import org.graylog2.indexer.healing.FixDeflectorByDeleteJob;
 import org.graylog2.indexer.healing.FixDeflectorByMoveJob;
@@ -155,6 +156,7 @@ public class ServerBindings extends Graylog2Module {
         bind(BundleExporterProvider.class).in(Scopes.SINGLETON);
         bind(ClusterStatsModule.class).asEagerSingleton();
         bind(ClusterConfigService.class).to(ClusterConfigServiceImpl.class).asEagerSingleton();
+        bind(GrokPatternRegistry.class).in(Scopes.SINGLETON);
 
         bind(String[].class).annotatedWith(named("RestControllerPackages")).toInstance(new String[]{
                 "org.graylog2.rest.resources",
--- a/graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
+++ b/graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
@@ -42,6 +42,18 @@ public class GrokPattern {
                 .toString();
     }
 
+    public static GrokPattern create(String name, String pattern) {
+        return create(null, name, pattern, null);
+    }
+    public static GrokPattern create(ObjectId id, String name, String pattern, String contentPack) {
+        final GrokPattern grokPattern = new GrokPattern();
+        grokPattern.id = id;
+        grokPattern.name = name;
+        grokPattern.pattern = pattern;
+        grokPattern.contentPack = contentPack;
+        return grokPattern;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
--- a/graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
+++ b/graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
@@ -32,6 +32,18 @@ public class GrokPattern {
     public String pattern;
     public String contentPack;
 
+    public String name() {
+        return name;
+    }
+
+    public String pattern() {
+        return pattern;
+    }
+
+    public String contentPack() {
+        return contentPack;
+    }
+
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
--- a/graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
@@ -1,0 +1,108 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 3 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
```diff
@@ -155,6 +156,7 @@ private void bindSingletons() {
         bind(BundleExporterProvider.class).in(Scopes.SINGLETON);
         bind(ClusterStatsModule.class).asEagerSingleton();
         bind(ClusterConfigService.class).to(ClusterConfigServiceImpl.class).asEagerSingleton();
+        bind(GrokPatternRegistry.class).in(Scopes.SINGLETON);
```
joschi: We should really start to consolidate on how to define singleton bindings (either with `Scopes.SINGLETON`, with the `@Singleton` annotation, or with `asEagerSingleton()`)…

Nothing to change here, though. :wink:

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
```diff
@@ -42,6 +54,18 @@ public String toString() {
                 .toString();
     }
 
+    public static GrokPattern create(String name, String pattern) {
+        return create(null, name, pattern, null);
+    }
+    public static GrokPattern create(ObjectId id, String name, String pattern, String contentPack) {
```
joschi: This looks as if it should mimic an Auto Value class. So why not make it one in the first place (except for having to adapt the current clients of this class as well)?

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
```diff
@@ -0,0 +1,107 @@
...
+        // for now we simply reload everything and don't care what exactly has changed
+        daemonExecutor.schedule(this::reload, 0, TimeUnit.SECONDS);
+    }
+
+    public Grok cachedGrokForPattern(String pattern) {
+        try {
+            return grokCache.get(pattern);
+        } catch (ExecutionException e) {
+            log.error("Unable to load grok pattern " + pattern + " into cache", e);
+            throw new RuntimeException(e);
```
joschi: Do we really want to rethrow the `ExecutionException` or maybe the cause of it?

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
```diff
@@ -0,0 +1,107 @@
...
+    public void grokPatternsChanged(GrokPatternsChangedEvent event) {
+        // for now we simply reload everything and don't care what exactly has changed
+        daemonExecutor.schedule(this::reload, 0, TimeUnit.SECONDS);
+    }
+
+    public Grok cachedGrokForPattern(String pattern) {
+        try {
+            return grokCache.get(pattern);
+        } catch (ExecutionException e) {
+            log.error("Unable to load grok pattern " + pattern + " into cache", e);
```
joschi: Optional nitpick: This *could* also use a parameterized logging statement nowadays: http://www.slf4j.org/faq.html#paramException

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
```diff
@@ -0,0 +1,107 @@
...
+        // trigger initial loading
+        reload();
+
+        serverEventBus.register(this);
+    }
+
+    @Subscribe
+    public void grokPatternsChanged(GrokPatternsChangedEvent event) {
+        // for now we simply reload everything and don't care what exactly has changed
+        daemonExecutor.schedule(this::reload, 0, TimeUnit.SECONDS);
```
joschi: Why do we have to offload the call to `reload()` to another thread in this case?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
```diff
@@ -111,10 +121,10 @@ public Response bulkUpdatePatterns(@ApiParam(name = "patterns", required = true)
         }
 
         grokPatternService.saveAll(GrokPatterns.fromSummarySet(patternList.patterns()), replace);
-
+        clusterBus.post(GrokPatternsChangedEvent.create(Collections.emptySet(), updatedPatternNames));
```
joschi: It looks like `updatedPatternNames` is never updated.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
```diff
@@ -124,21 +134,27 @@ public GrokPattern updatePattern(@ApiParam(name = "patternId", required = true)
                                      @ApiParam(name = "pattern", required = true)
                                      GrokPatternSummary pattern) throws NotFoundException, ValidationException {
         checkPermission(RestPermissions.INPUTS_EDIT);
-        
+
         final GrokPattern oldPattern = grokPatternService.load(patternId);
-        
+
+        final Set<String> updatedNames = Sets.newHashSet(pattern.name, oldPattern.name);
```
joschi: Is this really the semantics of the `GrokPatternsChangedEvent#updatedPatterns` field? Shouldn't a renamed pattern produce an event which contains the old name in `GrokPatternsChangedEvent#deletedPatterns` (because the pattern with that doesn't exist anymore)?

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
kroepke: No good reason, just wanted to keep the change minimal at this point.

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
kroepke: I just realize that the server event bus is async already, in that case we actually don't need the additional thread.
This was simply to prevent blocking too many threads. I'll update it.

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
kroepke: ah finally!

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
kroepke: I think it's fair to simply rethrow here, its cause will almost exclusively be that the Pattern did not compile, but at this point there's nothing to be done for the client anyway, but we might as well get the full trace for debugging.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
kroepke: correct, this is misleading as the reloader actually reloads everything. I'll put the new pattern names in to avoid confusion down the road

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
kroepke: also correct

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
joschi: :white_check_mark:

joschi:
LGTM. :+1:

PR #1976, opened 2016-03-24
Remove unused "local mode"

joschi:
Graylog supports the so called "local mode" which - in theory - should make it simpler to run local development setups of Graylog. In practice, this mode hasn't been used at all lately and the only consumer of the flag is `VersionCheckThread`.

Instead of dragging this along over the next few versions, this PR simply removes "local mode".


bernd:
I would prefer keeping this because this also supresses usage stats requests. I am using the `--local` flag for all my development and test setups to avoid sending versioncheck and usage stats requests.

joschi:
@bernd Fine with me. Closing this PR.

Issue #1816, opened 2016-02-13
v2.0.0-alpha.2 RestAccessLogFilter does not honor remote address provided by upstream proxies


mikkolehtisalo:
With [bigger production setups](http://docs.graylog.org/en/2.0/pages/architecture.html#bigger-production-setup) the load balancer is typically a reverse proxy, or some other application level network device. This causes the connection to the target service to originate from the IP of the load balancer. The implementation of [RestAccessLogFilter](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/rest/RestAccessLogFilter.java) does not honor X-Forwarded-For provided by upstream proxies. As result, the access log contains the remote address of the load balancer.

It is a common requirement for access logs to log the IP of the end user's workstation, especially in internal networks. Most commonly centralized log management systems are within that domain. At this moment the produced access log is useless in determining the original IP.

The issue can be demonstrated on the OVA by forcing the nginx to provide bogus X-Forwarded-For header with configuration like:

`proxy_set_header X-Forwarded-For "66.66.66.66";`

The logs still show the IP of the load balancer even after the configuration change:

`2016-02-13_21:37:13.39526 2016-02-13 21:37:13,395 DEBUG: org.graylog2.rest.accesslog - 192.168.122.1 6ef0f337-7073-4be9-80ca-bb641f506944 [-] "POST cluster/1f7effbb-e6f4-4d97-a00d-ce7fc20fcf01/metrics/multiple" Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.103 Safari/537.36 200 -1
`

The implementation should be enhanced by trusting the [X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For) header that came from trusted reverse proxies. It should not be trusted blindly from every address, because malicious client might also send the header, and bypassing the load balancer would bypass also rewriting the header.

I would recommend adding a new configuration string, containing CIDR of trusted proxies, check whether it is not null/empty in RestAccessLogFilter, and if the getRemoteAddr was in the range of CIDR trust the X-Forwarded-For. With the configuration string missing, present operation should be done.

mikkolehtisalo:
Basic documentation is at https://github.com/Graylog2/documentation/pull/88.

PR #1981, opened 2016-03-28
RestAccessLogFilter to use X-Forwarded-For set by trusted proxies

mikkolehtisalo:
Solves #1816 for the REST API audit logging.

- Introduces new configuration value trusted_proxies (comma separated list of IPs, and subnets)
- Adds RestTools method for RestAccessLogFilter, returning X-Forwarded-For value for logging, if the header was set by trusted proxy
-  As result the REST access log produces client addresses instead of load balancer's address

The same configuration value could be used also for the HTTP based input transports for similar logic, fixing faulty gl2_remote_ip, and similar. Also using the same configuration value for REMOTE_USER implementation could be possible.


runningman84:
It would be really great if you would show the last used IP address in the user list.

mikkolehtisalo:
@runningman84 Not a bad idea, but you can get that information by sending the rest access logs to this nice log management product called Graylog ;)

runningman84:
@mikkolehtisalo Graylog does not help here because we need this information in order to identify which user overloads the elasticsearch / graylog cluster. At this time the normal search is most likely not working anymore...

mikkolehtisalo:
Interesting use case. I am unsure where the state information should be stored...

joschi:
Thanks a lot!

PR #1986, opened 2016-03-29
Fix gelf chunk handler

bernd:
* Check for 128 chunk limit
* Fix parsing of message chunks

I started to move everything to `ChannelBuffer`s to avoid copying data all the time, but that change was a bit too large for the beta.

Refs https://github.com/Graylog2/gelf-rb/issues/40

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/gelf/GELFMessageChunk.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/gelf/GELFMessageChunk.java
@@ -18,6 +18,8 @@ package org.graylog2.inputs.codecs.gelf;
 
 import org.graylog2.plugin.Tools;
 import org.graylog2.plugin.inputs.MessageInput;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/codecs/gelf/GELFMessageChunk.java
```diff
@@ -144,7 +131,7 @@ private void extractSequenceNumber() {
     // lol duplication
     private void extractSequenceCount() {
         if (this.sequenceCount == -1) {
-            final int seqCnt = this.sliceInteger(HEADER_PART_SEQCNT_START, HEADER_PART_SEQCNT_LENGTH);
+            final int seqCnt = payload.getUnsignedByte(HEADER_PART_SEQCNT_START);;
```
joschi: Superfluous semicolon. :wink:

bernd:
Semicolon removed

joschi:
LGTM. :+1:

PR #1990, opened 2016-03-29
Implement cluster metrics endpoint to fetch from all nodes

### LINKED ISSUE #1974: Create cluster metrics endpoint that can return metrics for multiple nodes
### Problem description

In a multiple node cluster the web interface does one HTTP call for every node in the cluster to fetch the metrics every 2 seconds.

1. Setup multi node cluster
2. See multiple HTTP requests in the network tab of the developer tools every 2 seconds

### Proposal

Implement a new REST endpoint that takes a list of node ids and returns the metrics for them.

### Environment

* Graylog Version: Graylog 2.0.0-beta.2-SNAPSHOT (c0846cf)

bernd:
Adjust frontend MetricsStore to use the new endpoint. Saves `nodes - 1` requests in a multi node setup.

Fixes #1974

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
@@ -23,28 +23,20 @@ import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import org.apache.shiro.authz.annotation.RequiresAuthentication;
-import org.apache.shiro.authz.annotation.RequiresPermissions;
-import org.graylog2.cluster.Node;
 import org.graylog2.cluster.NodeNotFoundException;
 import org.graylog2.cluster.NodeService;
 import org.graylog2.rest.RemoteInterfaceProvider;
 import org.graylog2.rest.models.system.metrics.requests.MetricsReadRequest;
-import org.graylog2.rest.models.system.metrics.responses.MetricNamesResponse;
 import org.graylog2.rest.models.system.metrics.responses.MetricsSummaryResponse;
 import org.graylog2.shared.rest.resources.ProxiedResource;
 import org.graylog2.shared.rest.resources.system.RemoteMetricsResource;
-import org.graylog2.shared.security.RestPermissions;
-import retrofit2.Response;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
-import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
-import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
```diff
@@ -49,12 +49,14 @@
...
 import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
 
 import static javax.ws.rs.core.Response.Status.BAD_GATEWAY;
 
 @RequiresAuthentication
 @Api(value = "Cluster/Metrics", description = "Cluster-wide Internal Graylog metrics")
-@Path("/cluster/{nodeId}/metrics")
+@Path("/cluster")
```
edmundoa: This change merges this resource with `ClusterSystemResource` in swagger. Not sure if we want to fix it, but it feels a bit confusing at least.

graylog2-web-interface/src/stores/metrics/MetricsStore.js
```diff
@@ -43,46 +42,41 @@ const MetricsStore = Reflux.createStore({
...
-        });
+    // Then collect all global metric registrations
+    Object.keys(this.globalRegistrations)
+      .filter((metricName) => this.globalRegistrations[metricName] > 0)
+      .forEach((metricName) => {
+        metricsToFetch[metricName] = 1;
       });
 
-    const promise = this._allResults(promises).then((responses) => {
+    const url = URLUtils.qualifyUrl(ApiRoutes.ClusterMetricsApiController.multipleAllNodes().url);
```
edmundoa: Do we really want to get all metrics for all nodes, even when we only need metrics from one of them? Going to System -> Nodes -> Node fetches metrics about that node, and all other nodes in the cluster, even though we are not using them.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
bernd: Hmm, good catch. I guess the only way out here is to create another resource class.

graylog2-web-interface/src/stores/metrics/MetricsStore.js
bernd: But we still request the throughput counter for every node. So I think it's a trade off between fetching a bit more data from all nodes than needed and doing separate requests for each node.

For the throughput counter we contact every node anyway, so we won't even save any requests to any nodes. The only drawback I see is that we request a bit more data than needed.

edmundoa: I didn't think of that, sounds reasonable IMO :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterMetricsResource.java
bernd: Or use something like `/cluster/_all/metrics/multiple` to avoid the `/cluster/metrics/multiple endpoint.

edmundoa: I guess the renaming solution would be more consistent with the rest of the API, I don't think we use `_all` to indicate all nodes anywhere else.

bernd:
@edmundoa Please check the latest commits. I removed the class level `@Path` to make the UI browser work. (needed a tweak for the Generator code...)

bernd:
Damn, there is a problem. Sorry, hang on.

bernd:
@edmundoa I renamed the old cluster metrics endpoint to `ClusterNodeMetricsResource` and created a new `ClusterMetricsResource` class.

edmundoa:
LGTM :+1:

PR #1998, opened 2016-03-30
Add Expect: 100-Continue support to HTTP input

### LINKED ISSUE #1939: Inputs using HTTP transport do not support Expect: 100-continue
### Problem description

If a HTTP GELF client fragments the message using [Expect: 100-continue](https://tools.ietf.org/html/rfc7231#section-5.1.1) into separate header and body parts the server fails to handle message correctly. 

1. Use a client that is affected by default, for example the .NET [HttpWebRequest](https://msdn.microsoft.com/en-us/library/system.net.httpwebrequest(v=vs.110).aspx) 
2. POST HTTP GELF using http version 1.1
3. Client receives 202 Accepted, thinks that everything went ok
4. Server throws `IllegalStateException("GELF message is too short. Not even the type header would fit.")` at [GelfMessage.java](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/inputs/codecs/gelf/GELFMessage.java)
5. Message is lost

### Network traffic

Looking at the network traffic reveals (ordinary syn/ack stuff omitted):

Client (one packet):
```
POST /gelf HTTP/1.1
User-Agent: User agent x
Content-Type: text/json
Host: target:port
Content-Length: xxxx
Expect: 100-continue
Accept-Encoding: gzip
Connection: Close
```
Server (one packet):
```
HTTP/1.1 202 Accepted
Content-Length: 0
Connection: close
```
Client(One or more packets):
```
GELF/JSON BODY
```
Server:
```
Huh?!
```

### Options

1. Document that fragmentation must not be used. For example .NET HttpWebRequest requires setting `Expect100Continue=false` using [ServicePointManager](https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager(v=vs.110).aspx).
2. Implement missing feature to HTTP transports.
3. Do not return 202 Accepted if body size == 0, return 4xx or 5xx instead, with a "body was empty" error message in body

kroepke:
Previously the input blindly answered with 202 Accepted but instead it should properly support the Expect header.
Failing to respond correctly resulted in loss of messages because the client's expectation was broken.

Adding a HttpChunkAggregator fixes the issue while still allowing non-chunked requests to be sent.

fix #1939

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
@@ -45,6 +45,7 @@ import org.jboss.netty.channel.ChannelHandlerContext;
 import org.jboss.netty.channel.MessageEvent;
 import org.jboss.netty.channel.SimpleChannelHandler;
 import org.jboss.netty.handler.codec.http.DefaultHttpResponse;
+import org.jboss.netty.handler.codec.http.HttpChunkAggregator;
 import org.jboss.netty.handler.codec.http.HttpContentDecompressor;
 import org.jboss.netty.handler.codec.http.HttpMethod;
 import org.jboss.netty.handler.codec.http.HttpRequest;
@@ -117,6 +118,12 @@ public class HttpTransport extends AbstractTcpTransport {
             @Override
             public ChannelHandler call() throws Exception {
                 return new HttpRequestDecoder(DEFAULT_MAX_INITIAL_LINE_LENGTH, DEFAULT_MAX_HEADER_SIZE, maxChunkSize);
+            }
+        });
+        baseChannelHandlers.put("aggregator", new Callable<ChannelHandler>() {
+            @Override
+            public ChannelHandler call() throws Exception {
+                return new HttpChunkAggregator(maxChunkSize);
             }
         });
         baseChannelHandlers.put("encoder", new Callable<ChannelHandler>() {


kroepke:
Sample HTTP traffic:
```
02:40:35 server1$ curl -XPOST -vvv -HContent-Type:application/json -HExpect:100-Continue --http1.1 http://127.0.0.1:12999/gelf -d @/tmp/gelf.json
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 12999 (#0)
> POST /gelf HTTP/1.1
> Host: 127.0.0.1:12999
> User-Agent: curl/7.43.0
> Accept: */*
> Content-Type:application/json
> Expect:100-Continue
> Content-Length: 122
>
< HTTP/1.1 100 Continue
* We are completely uploaded and fine
< HTTP/1.1 202 Accepted
< Content-Length: 0
< Connection: keep-alive
<
* Connection #0 to host 127.0.0.1 left intact
02:40:42 server1$ curl -XPOST -vvv -HContent-Type:application/json --http1.1 http://127.0.0.1:12999/gelf -d @/tmp/gelf.json
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 12999 (#0)
> POST /gelf HTTP/1.1
> Host: 127.0.0.1:12999
> User-Agent: curl/7.43.0
> Accept: */*
> Content-Type:application/json
> Content-Length: 122
>
* upload completely sent off: 122 out of 122 bytes
< HTTP/1.1 202 Accepted
< Content-Length: 0
< Connection: keep-alive
<
* Connection #0 to host 127.0.0.1 left intact
```

PR #2007, opened 2016-03-31
Fix date time JSON parsing

bernd:
1. Disables the `ADJUST_DATES_TO_CONTEXT_TIME_ZONE` setting for the jackson `ObjectMapper`.
2. Fixes `AbsoluteRange#parseDateTime()` to not lose time zone information.

See commit messages for details.

PR #2008, opened 2016-03-31
PoC fix for the histogram bucketing issue

### LINKED ISSUE #1830: Fix histogram bucketing
When selecting a day resolution, graphs are not adjusted to the user's time zone, but to UTC. That means that a search from 2016-02-16 00:00:00 to 2016-02-16 23:59:59 on a time zone different to UTC, will return two data points instead of one.

This issue is the same as https://github.com/Graylog2/graylog2-web-interface/issues/1526, just moving it here to gain some visibility.

bernd:
This should not be merged as it is and needs careful testing and thinking about the consequences. It might produce wrong results.

---

The problem in #1830 is that the histogram query time range gets converted to UTC and thus creates wrong buckets when using i.e. "day" resolution. 

The time range `2016-03-24T00:00:00.000+09:00 -> 2016-03-24T23:59:59.000+09:00` becomes `2016-03-23T15:00:00.000Z -> 2016-03-24T14:59:59.000Z` and the query returns two buckets for a "day" resolution instead of one.

The query can be fixed by using the `time_zone` option for the Elasticsearch range filter and datetime histogram.

Example:

``` json
{
  "size": 0,
  "query": {
    "query_string": {
      "query": "*",
      "allow_leading_wildcard": false
    }
  },
  "aggregations": {
    "gl2_filter": {
      "filter": {
        "bool": {
          "must": {
            "range": {
              "timestamp": {
                "from": "2016-03-24 00:00:00.000",
                "to": "2016-03-24 23:59:59.000",
                "time_zone": "+09:00",
                "include_lower": true,
                "include_upper": true
              }
            }
          }
        }
      },
      "aggregations": {
        "gl2_histogram": {
          "date_histogram": {
            "field": "timestamp",
            "interval": "1d",
            "time_zone": "+09:00"
          }
        }
      }
    }
  }
}
```

Related Elasticsearch documentation:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#_time_zone_in_range_queries
- https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#_time_zone

The search code needs some adjustments in the backend and the frontend to cope with this. The commits in the PR make this work but they might break other things.

This needs some careful thinking and testing to make sure nothing breaks.

Includes commits from https://github.com/Graylog2/graylog2-server/pull/2007

Fixes #1830

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
@@ -22,6 +22,8 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
+import org.elasticsearch.index.query.RangeQueryBuilder;
+import org.elasticsearch.search.aggregations.bucket.histogram.DateHistogramBuilder;
 import org.graylog2.database.NotFoundException;
 import org.graylog2.indexer.ranges.IndexRange;
 import org.graylog2.indexer.ranges.IndexRangeService;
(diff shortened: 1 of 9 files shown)


brianshen1990:
According to this change, `Search Page` works OK while `Dashboard Page` would have problems when it comes to `Histogram` because it stored parameters `without TimeZone information`.

brianshen1990:
Hi @bernd , Using this solution, when choosing `absolute` time range, it works OK, but there are some other problems:
1. When choosing `Relative` range and aggregated with `day`, it still have the same problem. So `Relative` some be fixed as well.
2. When saving the histogram to Dashborad, it cannot display successfully. Because backend still use UTC when it comes to a saved one histgram widget, and UI wants to parse them with Admin Browser Timezone, so the histogram will be empty (As the expected x point has no corresponding data).

I fixed this with a simple workround: ignore all the datetime timezone, in backend, just use `Admin Timezone`, so it can be quite easy to handle. In my case, it is OK, however, when the `browser timezone` is not as same as the `Admin Timezone`, it will still have problems. 

My fixed commit:  https://github.com/brianshen1990/graylog2-server-CN/commit/a11f67e24c1d1aa6021343210ed7b99a3dc2df3a

BR
Brian

dennisoelkers:
This PR is now almost two years old and has conflicts. What are we going to do with it? I would suggest closing it for now. If someone is working on the issue, it is still accessible by the reference.

bernd:
@dennisoelkers Yes, I think we can close this one. :+1:

cla9rk:
I am still facing this issue - data tables in graylog dashboard are displaying the wrong day due to timezone offset.

bernd:
@cla9rk Please follow this issue for updates: https://github.com/Graylog2/graylog2-server/issues/1830 Thanks!

Issue #2000, opened 2016-03-30
Refesh user profile information when user changes


edmundoa:
### Problem description
When the user edit page is opened, clicking on a link directing to another user edit page, the user data in the page is not properly updated.

### Steps to reproduce the problem
1. Open the user list, and click on "Edit" for one of the users
2. Open the dropdown in the navbar with your name, and click on "Edit profile"
3. The page shows now inconsistent data, mixing data from both users

### Environment

* Graylog Version: Graylog 2.0.0-beta.2-SNAPSHOT (ad47def)

Issue #1989, opened 2016-03-29
Cannot edit single permissions for users


edmundoa:
### Problem description
Currently it is not possible to edit single streams/dashboards permissions for users, as changes in the input fields are not send to the server.

### Steps to reproduce the problem

1. Go to the edit user profile page
2. Change some single permissions
3. Save the changes
4. Reload and see how the changes were not applied

### Environment

* Graylog Version: Graylog 2.0.0-beta.2-SNAPSHOT (ad47def73ae32fbbdc291e702db9df2811f61800)

PR #2012, opened 2016-04-01
Update single permissions in user edit form

edmundoa:
This PR takes care of submitting and storing single stream/dashboard permissions for users (#1989). It also takes care of properly refreshing the user, fixing #2000.


kroepke:
lgtm

PR #2013, opened 2016-04-01
Update extractor example when message is loaded

### LINKED ISSUE #1957: Extractor - Load another message shows new message but it's not processed
### Problem description

I wanted to improve an existing extractor, used the "load another message" link and added the document id and the index.
When doing th "Load Message" the message is properly shown
Still, when I click on Try, I see the preview for the original message, not the one I just loaded

spectacle e11454
The original message was this one (I am just extracting the part after two spaces)
> [ warning] [guestinfo] Failed to get vmstats.

### Environment

* Graylog Version: alpha5

edmundoa:
When a new message is loaded in the extractor page, not only update the example preview, but also in other components using the example message.

Fixes #1957.


kroepke:
lgtm

PR #2023, opened 2016-04-04
Replace message field analysers dropdown

### LINKED ISSUE #2022: Improve message field analysers usability
### Problem description
Using a drop-down to select the field analyser is far from optimal. We already tried that in the past and doesn't work well.

### Environment

* Graylog Version: Graylog 2.0.0-beta.2

edmundoa:
Improve usability of selecting field analysers by replacing the drop-down with a collapsible panel. The main advantage of this change is that it avoids scrolling inside the search sidebar when the number of fields is small and there's space on the screen.

Fixes #2022.

### DIFF
--- a/graylog2-web-interface/src/components/search/SidebarMessageField.css
+++ b/graylog2-web-interface/src/components/search/SidebarMessageField.css
@@ -1,0 +1,15 @@
+.field-analyzer {
+    margin: 5px 5px 5px 12px;
+}
+
+.field-analyzer .panel-body {
+    padding: 5px;
+}
+
+.field-analyzer ul {
+    padding-left: 10px;
+}
+
+.field-selector {
+    margin-left: 25px;
+}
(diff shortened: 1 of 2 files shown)


bernd:
Not sure if this is supposed to be fixed, but I still have to scroll in the side bar.

screenshot from 2016-04-12 17 09 54

Maybe I got your description wrong and that cannot be avoided.

edmundoa:
You cannot avoid scrolling when there is no space at the bottom. But you can avoid it when the sidebar can expand. This is what shouldn't happen:

screen shot 2016-04-12 at 17 15 17

In that case, the sidebar does not get to the bottom of the page, but the dropdown is still hidden, making you scroll when there is no need for it. The panel version will actually increase the sidebar height, allowing you to see all options without any further action.

This is specially annoying when you have three or four fields, as you will have to scroll all the time when using dropdowns.

bernd:
Thanks for the explanation! :smiley:

bernd:
LGTM :+1:

PR #2053, opened 2016-04-12
Expose cancellation of system jobs in REST API and UI

bernd:
Exposes the existing cancellation of system jobs via the REST API and in the UI component for system jobs.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
@@ -82,8 +82,6 @@ public class ClusterSystemJobResource extends ProxiedResource {
                 if (response.isSuccessful()) {
                     // Return early because there can be only one job with the same ID in the cluster.
                     return response.body();
-                } else {
-                    LOG.warn("Unable to fetch system job {} from node {}: {}", jobId, entry.getKey(), response);
                 }
             } catch (IOException e) {
                 LOG.warn("Unable to fetch system jobs from node {}:", entry.getKey(), e);
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
@@ -28,11 +28,13 @@ import org.graylog2.rest.RemoteInterfaceProvider;
 import org.graylog2.rest.models.system.SystemJobSummary;
 import org.graylog2.rest.resources.system.jobs.RemoteSystemJobResource;
 import org.graylog2.shared.rest.resources.ProxiedResource;
+import org.hibernate.validator.constraints.NotEmpty;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import retrofit2.Response;
 
 import javax.inject.Inject;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.Path;
@@ -42,7 +44,6 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import java.io.IOException;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/jobs/SystemJobResource.java
```diff
@@ -170,5 +171,35 @@ public Response trigger(@ApiParam(name = "JSON body", required = true)
...
+        if (systemJob == null) {
+            throw new NotFoundException("No system job with ID <" + jobId + "> found");
+        }
+
+        checkPermission(RestPermissions.SYSTEMJOBS_DELETE, systemJob.getClassName());
+
+        if (systemJob.isCancelable()) {
+            systemJob.requestCancel();
+        } else {
+            throw new ClientErrorException("System job with ID <" + jobId + "> cannot be cancelled", 422);
```
joschi: Why respond with [HTTP status 422 (Unprocessable Entity)](https://httpstatuses.com/422) instead of 404 (Not Found) for example?

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
```diff
@@ -91,4 +90,31 @@ public SystemJobSummary getJob(@ApiParam(name = "jobId", required = true) @PathP
...
+                        return resource.delete(jobId).execute();
+                    } catch (IOException e) {
+                        LOG.warn("Unable to fetch system jobs from node {}:", entry.getKey(), e);
+                        return null;
+                    }
+                })
+                .filter(response -> response != null && response.isSuccessful())
+                .findFirst(); // There should be only one job with the given ID in the cluster. Just take the first one.
+
+        if (summaryResponse.isPresent()) {
```
joschi: Could be shorter with [`Optional#orElseThrow()`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseThrow-java.util.function.Supplier-):
```java
return summaryResponse.orElseThrow(() -> new NotFoundException("System job with ID <" + jobId + "> not found!")).body();
```

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
```diff
@@ -81,8 +82,6 @@ public SystemJobSummary getJob(@ApiParam(name = "jobId", required = true) @PathP
                 if (response.isSuccessful()) {
                     // Return early because there can be only one job with the same ID in the cluster.
                     return response.body();
-                } else {
```
joschi: Any reason you've removed this branch?

bernd: As described in the commit message. :smiley: 

> Do not log an error or warning if a system job cannot be found

I think we do not have to log that case as a warning or error.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/jobs/SystemJobResource.java
bernd: A 404 does not sound correct to me because the job exists. Not entirely happy with 422, either. I also thought about 403.

joschi: Makes sense. Let's go with 403 (Forbidden).

graylog2-web-interface/src/components/systemjobs/SystemJob.jsx
```diff
@@ -3,17 +3,31 @@ import { ProgressBar } from 'react-bootstrap';
...
+      if (window.confirm(`Are you sure you want to cancel system job "${job.info}"?`)) {
+        SystemJobsActions.cancelJob(job.id);
+      }
+    };
+  },
   render() {
     const job = this.props.job;
     const progress = job.percent_complete < 100 ?
       (<ProgressBar bsStyle="info" active now={job.percent_complete}/>) : <span className="label label-success finished">Finished!</span>;
+    const cancel = job.is_cancelable ? (<span> - <a href="" onClick={this._onCancel(job)}>cancel job</a></span>) : null;
```
joschi: Instead of only displaying the link, this should be a button (similarly colored to the "Stop Input" button on the System &rarr; Inputs page).

bernd: I experimented with that, but it looked really ugly. I think the whole system job component could use some styling :heart:. This is the minimal change I would do for 2.0. We might want to revisit this for 2.x.

/cc @edmundoa

joschi: While not ideal, I think it doesn't look too bad:
Styled button

```diff
diff --git a/graylog2-web-interface/src/components/systemjobs/SystemJob.jsx b/graylog2-web-interface/src/components/systemjobs/SystemJob.jsx
index 46345a1..7a7fd92 100644
--- a/graylog2-web-interface/src/components/systemjobs/SystemJob.jsx
+++ b/graylog2-web-interface/src/components/systemjobs/SystemJob.jsx
@@ -19,7 +19,7 @@ const SystemJob = React.createClass({
     const job = this.props.job;
     const progress = job.percent_complete < 100 ?
       (<ProgressBar bsStyle="info" active now={job.percent_complete}/>) : <span className="label label-success finished">Finished!</span>;
-    const cancel = job.is_cancelable ? (<span> - <a href="" onClick={this._onCancel(job)}>cancel job</a></span>) : null;
+    const cancel = job.is_cancelable ? (<button type="button" className="btn btn-primary btn-xs" onClick={this._onCancel(job)}>Cancel Job</button>) : null;

     return (
       <div>
```

bernd: Here the link as comparison:

welgi6phqnxwqaaaabjru5erkjggg

edmundoa: To be honest I don't have a strong opinion about it. I like the button because it's more consistent with other elements of the interface, but the spacing and placement looks odd imho.

I was looking a bit and I think we can style the button to go to the right, so it looks a bit less weird. I can take care of that in my task of styling things a bit :)

bernd:
All comments addressed and branch updated.

joschi:
LGTM. :+1:

PR #2062, opened 2016-04-12
Improve handling of Elasticsearch indices

### LINKED ISSUE #2058: Problem with index name handling
### Problem description

In https://github.com/Graylog2/graylog2-server/commit/1728772d0c9e6271e67a14eac266b26afdd11a7d we introduced a stricter pattern to check if an index is handled by Graylog to avoid touching anything else.

I see two problems with that:

1. The "System/Indices" page still shows other indices. It looks like the pattern is not used when requesting the list of indices.
2. In some cases we actually need to handle indices that do not match the pattern. For example the upcoming archive plugin needs to be able to restore indices with a different name to avoid any name clashes.

Trying to open the details for the `graylog_17_restored_archive` does not work.

wsrjtnkaaaaaelftksuqmcc

The details do not show up, the spinner keeps spinning.

axhqnlxtfxlqaaaaaelftksuqmcc

In the developer tools I see a failed request for the archive because the name does not match the strict pattern.

```json
{
  "type":"ApiError",
  "message":"Index [graylog_17_restored_archive] doesn't look like an index managed by Graylog."
}
```

1. Use archive plugin in 2.0.0-beta.2 to create an archive for an index.
2. Restore the archive index.
3. Navigate to "System/Indices"
4. See an index named like `*_restored_archive`
5. Try to click "Show Details / Actions"
6. Spinner does not go away and see errors in the developer console.

### Solution Discussion

To avoid showing the non-`^graylog_[0-9]+$` indices on the "System/Indices" page, we could just make sure to apply the strict pattern for the index list as well.

The problem is that the archive plugin cannot be used for restores at the moment because of the strict pattern handling. An archived index can be restored but then there will be no index range for it and we also cannot delete it.

One idea would be to make the index name patterns pluggable to allow plugins to inject a new name format.

Any other or better ideas?

### Environment

* Graylog Version: 2.0.0-beta.3-SNAPSHOT (882c097)
* Elasticsearch Version: 2.3.0

bernd:
- Make sure to only return Graylog indices in the indices and indices
  overview resources.
- Until we find a better solution, hard code the index suffix for restored
  archive indices to make it possible to search in restored indices and
  close/delete them.

Fixes #2058

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -21,7 +21,6 @@ import com.google.common.collect.Lists;
 import org.elasticsearch.action.admin.indices.stats.IndexStats;
 import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.indices.InvalidAliasNameException;
-import org.graylog2.configuration.ElasticsearchConfiguration;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.indexer.ranges.CreateNewSingleIndexRangeJob;
 import org.graylog2.shared.system.activities.Activity;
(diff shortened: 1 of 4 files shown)


joschi:
LGTM. :+1:

PR #2065, opened 2016-04-12
override timestamp to now for bad timestamp values

### LINKED ISSUE #2064: Do not allow arbitrary value for timestamp field
### Problem description

Message inputs that use part of an incoming message as timestamp value or extractors that overwrite timestamp can lead to malformed timestamp fields. This leads to losing messages, because the mapping for timestamp in elasticsearch is very strict.

1. Send a minimal gelf message such as `{"version":"1.1", "short_message":"foo","host":"hostname"}`.
2. Create an extractor `copy_input` on field `message`, overwriting the field `timestamp`.
3. The message is discarded during indexing.

### Environment

* Graylog Version: all
* Elasticsearch Version: all
* MongoDB Version: -
* Operating System: - 
* Browser version: -

kroepke:
If the timestamp field of a message is incompatible to the ES format we are using, or an unknown data type (non Date, DateTime or String), force the timestamp to be "now".

This prevents bad extractors or inputs from causing these malformed messages to be discarded.

fix #2064

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -16,6 +16,8 @@
  */
 package org.graylog2.indexer.messages;
 
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.MetricRegistry;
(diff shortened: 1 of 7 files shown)


bernd:
LGTM :+1:

Issue #2059, opened 2016-04-12
Web Interface shows error message when session is stale


joschi:
### Problem description

The web interface shows an error message when opening it for the first time in a browser and a stale session exists in local storage.

### Steps to reproduce the problem

1. Open web interface (e. g. http://localhost:9000)
2. Log in
3. Let session become stale (e. g. set session timeout to 10 seconds)
4. Open web interface (e. g. http://localhost:9000)
5. See error message about unauthorized access to the last resource which was open during the last session.

<img width="512" alt="Screenshot" src="https://cloud.githubusercontent.com/assets/43951/14458046/0180e4f2-00af-11e6-99da-1fa51285e447.png">

### Environment

* Graylog Version: Graylog 2.0.0-beta.3-SNAPSHOT

PR #2073, opened 2016-04-13
Delay requests for invalid session

### LINKED ISSUE #1891: Getting started guide does not show up
### Problem description

When login to Graylog for the first time the Getting-started guide does not show up. Reloading startpage shows Getting-started guide

1. I login to Graylog after building the QA box
2. When login in I expect the Getting-started guide but page is not loading
3. After new logout/ login I am getting directed to /search instead of /getting-started
4. After page reload the Getting-started guide shows up

### Environment

* QA Setup
* `graylog-2.0.0-alpha.5-SNAPSHOT-20160302193709-graylog-server-tarball.tar.gz`
* I am using Firefox 44.0.2

dennisoelkers:
This PR adds a simple check during initialisation of the `SessionStore` which validates that a persisted session id is still valid. If it is not, then all requests coming in will be delayed while listening for the login event, just like in the case where no session id exists at all.

This fixes a number of problems related to stores doing initialisation only once and fail when the session id is stale.

Fixes #1891, #2059.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
@@ -27,11 +27,11 @@ import org.apache.shiro.mgt.DefaultSecurityManager;
 import org.apache.shiro.session.UnknownSessionException;
 import org.apache.shiro.subject.Subject;
 import org.apache.shiro.util.ThreadContext;
+import org.graylog2.plugin.database.users.User;
 import org.graylog2.rest.models.system.sessions.requests.SessionCreateRequest;
 import org.graylog2.rest.models.system.sessions.responses.SessionResponse;
 import org.graylog2.shared.rest.resources.RestResource;
 import org.graylog2.shared.security.ShiroSecurityContext;
-import org.graylog2.plugin.database.users.User;
 import org.graylog2.shared.users.UserService;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -120,6 +119,11 @@ public SessionResponse newSession(@Context ContainerRequestContext requestContex
         throw new NotAuthorizedException("Invalid username or password", "Basic realm=\"Graylog Server session\"");
     }
 
+    @GET
+    @ApiOperation(value = "Validate an existing session", notes = "Checks the session with the given ID: return OK if session is valid.")
```
edmundoa: Just a bit of nit-picking: the description says it will return an OK if the session is valid, but it actually responds No Content. We could change the method to return an OK or just say it returns a success code.

edmundoa:
Other than the (tiny) remark with the REST documentation, looks good to me.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
dennisoelkers: :+1: this is actually Jersey being more intelligent than me :)

edmundoa:
LGTM :+1:

PR #2078, opened 2016-04-14
Replace "." in message field keys with a "_" for ES >=2.0 compatibility

bernd:
The replacement character is currently hard coded to "_". If the transformed key already exists, we drop the field.

This reverts the Message parts of d07f06f.


joschi:
LGTM. :+1:

vladimirfx:
GELF spec clearly states that dot is valid character. This fix is data loss because of conflicting keys is just removed. With this "fix" data completeness is depended on runtime field set. Consider using other delimiter (eg double "_") as replacement separator or make it configurable.

PR #2079, opened 2016-04-14
Replace Atlassian IPMatcher with Netty IpSubnet

joschi:
Since the license of the [`com.atlassian.ip:atlassian-ip`](https://www.versioneye.com/java/com.atlassian.ip:atlassian-ip/3.1) artifact is unclear, we'll take the safe route and use the IpSubnet from Netty.


edmundoa:
LGTM :+1:

PR #2092, opened 2016-04-15
Use Maven wrapper

joschi:
Using the Maven wrapper should help to enable stable build results (e. g. using the same Maven version everywhere) and don't rely on more than a JDK to be installed on a given machine to be able to build Graylog.

Blatantly stolen from the Gradle documentation about their Gradle wrapper (which works similar to the Maven wrapper):
> Most tools require installation on your computer before you can use them. If the installation is easy, you may think that’s fine. But it can be an unnecessary burden on the users of the build. Equally importantly, will the user install the right version of the tool for the build? What if they’re building an old version of the software?

See https://github.com/takari/takari-maven-plugin and https://github.com/takari/maven-wrapper for details about the Maven wrapper.


bernd:
Moving this to 2.1. I don't want to fiddle with our builds shortly before we are doing a release.

joschi:
No need to cling to a PR that's never going to get merged.

PR #2093, opened 2016-04-15
Dashboard usability changes

edmundoa:
These changes address some issues with dashboards:

- It is hard to move widgets to a new empty row when there is not enough height for a new widget
- Action buttons look odd on `md` screen sizes
- Resizing the window may hide some widgets, putting them out of reach

One of the biggest changes is that these changes are removing the broken responsive widget positioning we were trying to use in previous versions.

The reason for this is that Gridster doesn't really support that, so it was working in a random way, and changing the ordering as it pleased. Integrating Gridster into React made this even a worse problem, as now it is quite hard to know who is changing the DOM in which way.

So, instead of trying to unsuccessfully do that, I have changed the behaviour to always respect the positioning users set, regardless of the resolution. It is still possible to see all widgets in a small screen by scrolling, but we won't try to be smart about resizes and re-arrange all widgets in a random way.

That is not the ideal way I would like to have, but I think it's at least consistent, and fixes some issues while we use Gridster and React together.

### DIFF
--- a/graylog2-web-interface/src/components/common/GridsterContainer.jsx
+++ b/graylog2-web-interface/src/components/common/GridsterContainer.jsx
@@ -1,4 +1,4 @@
-import React, {PropTypes} from 'react';
+import React, { PropTypes } from 'react';
 import ReactDOM from 'react-dom';
 
 import { GridsterWidget } from 'components/common';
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/common/GridsterContainer.jsx
```diff
@@ -63,14 +82,14 @@ const GridsterContainer = React.createClass({
...
-      return {id: position.id, col: position.col, row: position.row, width: position.size_x, height: position.size_y};
+      return { id: position.id, col: position.col, row: position.row, width: position.size_x, height: position.size_y };
     });
     this.props.onPositionsChange(positions);
   },
 
   render() {
     const children = (this.state.grid && React.Children.map(this.props.children, (child) => {
-      const position = this.props.positions[child.props.id] || {row: 0, col: 0, width: 1, height: 1};
+      const position = this.props.positions[child.props.id];
```
dennisoelkers: Why are you doing this? What happens when there is no position for the widget (for some reason)?

edmundoa: They always get them from `ShowDashboardPage`, but still you are right, and we shouldn't remove it from here.

STRML:
By the way, I have a project that may handle this more gracefully and is built for responsive layouts: https://github.com/STRML/react-grid-layout

Its behavior is very similar to what you have now. In fact, at my first time using it, I thought you were using RGL. Would be happy to help support if you want to use it and have any integration issues.

edmundoa:
@STRML Thank you for the suggestion! I think I was looking into it some months ago, and there was something missing to completely replace Gridster. We will take a look again, as Gridster is quite hard to work with once you use React.

gruselglatz:
@STRML would love to see such a good widget placement in graylog. Since there is so much space lost on the right side, even on widely used resolutions like 1080...

PR #2097, opened 2016-04-18
Unconditionally overwrite Graylog index template

### LINKED ISSUE #2089: graylog index template not using configured index prefix
### Problem description

 Version 2.0-beta.3 does not take the configured index prefix into account.

1. configure server.conf to use `graylog_` as index prefix
2. start graylog
3. newly created index name is `graylog_0` (correct)
4. index template `graylog-internal` is using `graylog2_*` as template pattern:

```
{
  "order": -2147483648,
  "template": "graylog2_*",
  "settings": {},
  "mappings": {
    "message": {
      "_ttl": {
        "enabled": true
      },
      "_source": {
        "compress": true,
        "enabled": true
      },
      "dynamic_templates": [
        {
          "internal_fields": {
            "mapping": {
              "index": "not_analyzed",
              "doc_values": true
            },
            "match": "gl2_*"
          }
        },
        {
          "store_generic": {
            "mapping": {
              "index": "not_analyzed"
            },
            "match": "*"
          }
        }
      ],
      "properties": {
        "full_message": {
          "analyzer": "standard",
          "index": "analyzed",
          "type": "string"
        },
        "source": {
          "analyzer": "analyzer_keyword",
          "index": "analyzed",
          "type": "string"
        },
        "message": {
          "analyzer": "standard",
          "index": "analyzed",
          "type": "string"
        },
        "timestamp": {
          "format": "yyyy-MM-dd HH:mm:ss.SSS",
          "type": "date",
          "doc_values": true
        }
      }
    }
  },
  "aliases": {}
}
```

### Environment

* Graylog Version: 2.0.0-beta.3
* MongoDB Version: any
* Operating System: any
* Browser version: any

joschi:
Instead of checking for the existence of the internal Graylog index template in Elasticsearch and only writing the index template if none exists, this PR changes the behavior to unconditionally write (and overwrite) the index template.

Users should use their own index templates which are being merged with the existing Graylog index template prior to index creation.

Fixes #2089

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -45,7 +45,6 @@ import org.elasticsearch.action.admin.indices.stats.IndexStats;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsRequest;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsResponse;
 import org.elasticsearch.action.admin.indices.stats.ShardStats;
-import org.elasticsearch.action.admin.indices.template.get.GetIndexTemplatesResponse;
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequest;
 import org.elasticsearch.action.bulk.BulkRequestBuilder;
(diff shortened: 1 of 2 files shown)


kroepke:
👍

PR #2100, opened 2016-04-18
Only return deflector indices in Deflector#getAllDeflectorIndices()

bernd:
Unbreaks count based index retention (again).

Also rename `#getAllDeflectorIndexNames()` to `#getAllIndexNames()` for clarity.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -212,7 +212,7 @@ public class Deflector { // extends Ablenkblech
      *
      * @return list of managed indices
      */
-    public String[] getAllIndexNames() {
+    public String[] getAllGraylogIndexNames() {
         final Map<String, IndexStats> indices = this.indices.getAll();
         final List<String> result = Lists.newArrayListWithExpectedSize(indices.size());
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
```diff
@@ -207,7 +207,12 @@ public int getNewestTargetNumber() throws NoTargetIndexException {
         return Collections.max(indexNumbers);
     }
 
-    public String[] getAllDeflectorIndexNames() {
+    /**
+     * Returns a list of all Graylog managed indices.
+     *
+     * @return list of managed indices
+     */
+    public String[] getAllIndexNames() {
```
joschi: Maybe `getAllGraylogIndexNames` would be more specific (and would also match `isGraylogIndex`).

joschi:
LGTM. :+1:

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
bernd: Aye, good point. Renamed to `getAllGraylogIndexNames` and `getAllGraylogDeflectorIndices`.

PR #2101, opened 2016-04-18
Add config setting to override web interface endpoint URI

joschi:
This PR adds the configuration setting `web_endpoint_uri` which allows to set the URI of the Graylog REST API used by the Graylog web interface.

By default, the `rest_transport_uri` is being used which might not work in more complicated network setups (e. g. on AWS using ELB and multiple Graylog server instances). The URI of the Graylog REST API can still be overridden on a per-request basis using the `X-Graylog-Server-URL` HTTP request header (see #1808).

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
@@ -17,7 +17,6 @@
 package org.graylog2.web.resources;
 
 import com.floreysoft.jmte.Engine;
-import com.google.common.base.Strings;
 import com.google.common.io.Resources;
 import org.graylog2.Configuration;
 import org.graylog2.rest.MoreMediaTypes;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -59,29 +59,27 @@ public String get(@Context HttpHeaders headers) {
...
+    private String buildEndpointUri(HttpHeaders httpHeaders) {
+        Optional<String> endpointUri = Optional.empty();
+        final List<String> headers = httpHeaders.getRequestHeader(CK_OVERRIDE_SERVER_URI);
+        if (headers != null && !headers.isEmpty()) {
+            endpointUri = headers.stream().filter(s -> {
                 try {
-                    return !Strings.isNullOrEmpty(s) && new URL(s) != null;
-                } catch (MalformedURLException e) {
+                    return new URI(s).isAbsolute();
+                } catch (NullPointerException | URISyntaxException e) {
```
dennisoelkers: Why is this change required?

joschi: I assume you mean the change from `java.net.URL` to `java.net.URI` (and with it the change of the caught exceptions).

The `URL` class doesn't have an equivalent of [`URI#isAbsolute()`](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#isAbsolute--) which checks that the URI contains a scheme component (e. g. `http://` or `https://`). Additionally, this change simplifies the condition in the `try` block.

dennisoelkers: For the reasons you have mentioned, this is an unnecessary change. The URL class does not have `isAbsolute`, because for an URL the scheme is mandatory in the first place. This was checked before by catching the `MalformedURLException` exception. Also, the condition might be shorter, but I really dislike the implicit null check by catching the exception, it makes it much less explicit to read.

Also, if we declare in the config file that `web_endpoint_uri` is an URI, we also need to mention that the scheme is mandatory if we assume this.

joschi: > For the reasons you have mentioned, this is an unnecessary change.

✅ I'll revert that part.

> Also, if we declare in the config file that `web_endpoint_uri` is an URI, we also need to mention that the scheme is mandatory if we assume this.

A valid URI always has a scheme component: https://tools.ietf.org/html/rfc3986#section-3

dennisoelkers: Java has a different opinion on that ;)

```
groovy:000> new URI("www.google.com")
===> www.google.com
groovy:000> new URL("www.google.com")
ERROR java.net.MalformedURLException:
no protocol: www.google.com
```

PR #2104, opened 2016-04-18
Hide search refresh controls on sources page

### LINKED ISSUE #1821: Search live reload and sources tab
As the controls for the search live reload are part of the search bar, they also appear in the sources tab, although they don't refresh the data in there. We should either remove (or disable) the live reload controls on the sources tab, or also refresh the sources data when live reload is active.

edmundoa:
For now we hide the refresh controls on the sources page. We could add them back in the future if we decide to allow refreshing the sources information.

Fixes #1821.


bernd:
LGTM :+1:

PR #2106, opened 2016-04-19
Upgrade to Drools 6.4.0.Final

joschi:
Release notes: https://docs.jboss.org/drools/release/6.4.0.Final/drools-docs/html/ch02.html#drools.ReleaseNotesDrools.6.4.0

Optional for Graylog 2.0.0, could also be moved to Graylog 2.1.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
         <apacheds-server.version>2.0.0-M20</apacheds-server.version>
         <surefire.version>2.19.1</surefire.version>
         <guice.version>4.0</guice.version>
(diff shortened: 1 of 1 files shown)


bernd:
Moving to 2.1.

kroepke:
I'd also like us to consider moving the Drools support into a plugin to get rid of the core dependency eventually.

joschi:
@kroepke Agreed. But since this is breaking compatibility and involves more changes, I'd implement that in a different PR: #2309

PR #2112, opened 2016-04-20
Move TCP keepalive setting into AbstractTcpTransport

joschi:
All kinds of TCP transports might profit from using TCP keepalive, not just line-based TCP transports (which are implemented by `TcpTransport`).


kroepke:
lgtm 👍

PR #2114, opened 2016-04-20
Allow clearing optional values in plugins

### LINKED ISSUE #2108: AlarmCallback does not allow clearing of optional fields with default-value
### Problem description

I noticed this on Graylog v1 and Graylog v2 (2.0.0-beta.3). I am not able to delete the value of an optional field which has a default value.

In the example below the field "MYFIELD" will show the default value of "SOME_DEFAULT_VALUE" when editing the plugin properties. If I delete the value in the web-UI, the config-field will take the default-value of the field.

```
  public ConfigurationRequest getRequestedConfiguration () {
      final ConfigurationRequest configurationRequest = new ConfigurationRequest();
      
      configurationRequest.addField (new TextField (
          "MYFIELD", "MYFIELDLABEL", "SOME_DEFAULT_VALUE", "DESCRIPTION",
      
      
      return configurationRequest;
  }
```

How can I provide my AlarmCallback with a default-value during configuration, but then also allow the user to delete that value. The only way to make this work at the moment is to ask the user to enter some string which I then identify in the plugin as an empty value (in this case "null"):

```
    String myField = getConfigValue (configuration.getString("MYFIELD"));

    private final String getConfigValue(String x) {
        // Dirty work-around to allow empty content on optional fields with default value
        // We return NULL if the field-value is empty or the user has entered "null"
        return (isNullOrEmpty(x) || x.equals("null") ? null : x);
    }

```

1. Write a AlarmCallback plugin with an optional field and a default value
2. Edit the callback-plugin and then try to delete the pre-populated default value
3. Save
4. Notice how the default value has overwritten the deleted string

### Environment

* Graylog Version: Graylog 2.0.0-beta.3 (1d03037)

edmundoa:
These changes allow users to clear out optional values on pluggable entities forms.

Fixes #2108

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java
+++ b/graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java
@@ -41,7 +41,7 @@ public class ConfigurationMapConverter {
             final String type = (String) fieldDescription.get("type");
 
             // Decide what to cast to. (string, bool, number)
-            final Object value;
+            Object value;
             switch (type) {
                 case "text":
                 case "dropdown":
@@ -51,7 +51,12 @@ public class ConfigurationMapConverter {
                     try {
                         value = Integer.parseInt(String.valueOf(entry.getValue()));
                     } catch (NumberFormatException e) {
-                        throw new ValidationException(field, e.getMessage());
+                        // If a numeric field is optional and not provided, use null as value
+                        if ("true".equals(String.valueOf(fieldDescription.get("is_optional")))) {
+                            value = null;
+                        } else {
+                            throw new ValidationException(field, e.getMessage());
+                        }
                     }
                     break;
                 case "boolean":
--- a/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
+++ b/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
@@ -31,7 +31,7 @@ const ConfigurationForm = React.createClass({
     data.configuration = {};
 
     $.map(this.state.configFields, function(field, name) {
-      data.configuration[name] = (values[name] === undefined  || values === null || String(values[name]).trim() === "" ? field.default_value : values[name]);
+      data.configuration[name] = values[name] || '';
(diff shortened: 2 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java
```diff
@@ -51,7 +51,12 @@
                     try {
                         value = Integer.parseInt(String.valueOf(entry.getValue()));
                     } catch (NumberFormatException e) {
-                        throw new ValidationException(field, e.getMessage());
+                        // If a numeric field is optional and not provided, use null as value
+                        if ("true".equals(String.valueOf(fieldDescription.get("is_optional")))) {
```
joschi: This expression could be simplified using [`Boolean#valueOf(String)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#valueOf-java.lang.String-):
```java
if (Boolean.valueOf(fieldDescription.get("is_optional")))
```

edmundoa: Not really, since `fieldDescription` values are of type `Object`. In the end we could cast the value in `is_optional` to a `String`, but that won't work if the value is actually a `Boolean`.

joschi: ✅

joschi:
LGTM. :+1:

magicdude4eva:
Is it possible to determine the Graylog version within an Alarmcallback (preferably during getRequestedConfiguration)?

This would then allow me to set default values for Graylog 2.0.0 (as a user can now delete the defaults) and for older versions not populating the defaults.

joschi:
@magicdude4eva We are using GitHub issues for tracking bugs in Graylog itself, but this doesn't look like one. Please post your questions to our [public mailing list](https://groups.google.com/forum/#!forum/graylog2) or join the [#graylog channel on freenode IRC](https://webchat.freenode.net/?channels=%23graylog).

Thank you!

PR #2122, opened 2016-04-21
Include default values in plugable entities forms

### LINKED ISSUE #2108: AlarmCallback does not allow clearing of optional fields with default-value

edmundoa:
Changes in #2114 caused another issue with the pluggable entities forms: default values were shown in forms, but never submitted to the server.

These changes move the logic taking care of default values to the `ConfigurationForm` component, also removing incorrect uses of `defaultValue` in inputs.

In this way, the object containing the data will contain the default values at the beginning, and they will be overriden by any input the user type in them.

Fixes #2108, this time hopefully in the right way :smile:

### DIFF
--- a/graylog2-web-interface/src/components/configurationforms/BooleanField.jsx
+++ b/graylog2-web-interface/src/components/configurationforms/BooleanField.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import FieldHelpers from './FieldHelpers';
 
+import FormsUtils from 'util/FormsUtils';
+
 const BooleanField = React.createClass({
   propTypes: {
     autoFocus: React.PropTypes.bool,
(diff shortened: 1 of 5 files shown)


bernd:
Tested, fixes the issue. :+1:

PR #2123, opened 2016-04-21
ignore empty authentication tokens in LdapUserAuthenticator

### LINKED ISSUE #2111: CSV export fails with 404
### Problem description
Apparently there are some issues exporting search results as CSV, from #1898: 

> FYI, I see this too, but with different symptoms
> 
> Under Firefox I get a warning
> 
> You are about to log in to the site "graylog2.trimble.com" with the username "e482ce63-0f1b-4d62-add3-c9acba4472ae", but the website does not require authentication. This may be an attempt to trick you.
> 
> Under Chrome I get no warning, but in both cases it returns a 404 and fails. I noticed the query begins with "//search", which I assume should have been "/search" - so maybe that's related
> 
> Jason

kroepke:
unauthenticated requests and requests which use sessions in the URL rather than via HTTP headers can have a null UsernamePasswordToken associated with them
The Ldap authenticator didn't check for null values before looking up the account, which lead to a harmless, but noisy exception which was not logged in previous versions

add null checks before use the token

fixes #2111

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.security.realm;
 
+import com.google.common.base.Strings;
 import org.apache.directory.api.ldap.model.cursor.CursorException;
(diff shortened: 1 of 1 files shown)


bernd:
LGTM :+1:

PR #2128, opened 2016-04-22
the async csv export writer dropped the first chunk

### LINKED ISSUE #2126: CSV export contains not full data
### Problem description
After exporting search result to CSV I've found some "holes" in data.
Interesting thing, is that in search result window all data displayed correctly.
To better illustrate problem i've created sample image from our soft:
hole
The problem is in hole in the left part. In reality there is no such hole in data.
Also data in CSV seems to be unordered (but it is not a big problem for us)
If it matters, request is relative - last 5 minnutes and we have about 5K messages per minnute.

### Environment

* Graylog Version: 2.0 RC1
* Browser version: Yandex Browser 16.3.0.7843

kroepke:
the first scroll chunk was discarded when writing the response
remember the initial search response and return the initial hits to avoid dropping data

fixes #2126


bernd:
LGTM :+1:

PR #2133, opened 2016-04-25
Properly generating link to node's api browser.

### LINKED ISSUE #2131: RC.1 OVA API Browser button does not work
### Problem description

Using the 2.0 rc.1 ova, the System/Nodes page's api browser button does not work.
The URL is missing a `/` between port and path.

### Environment

* Graylog Version: 2.0.0-rc.1 beta OVA

dennisoelkers:
We are now using URI.js to make sure that there is a good number of slashes (more than 0 and less than 2 between host/port and directory) included in the generated link target.

Fixes #2131

### DIFF
--- a/graylog2-web-interface/src/components/nodes/NodesActions.jsx
+++ b/graylog2-web-interface/src/components/nodes/NodesActions.jsx
@@ -1,6 +1,7 @@
 import React, {PropTypes} from 'react';
 import { LinkContainer } from 'react-router-bootstrap';
 import { Button, DropdownButton, MenuItem } from 'react-bootstrap';
+import URI from 'urijs';
 
 import { IfPermitted } from 'components/common';
 
@@ -38,6 +39,7 @@ const NodesActions = React.createClass({
     }
   },
   render() {
+    const apiBrowserURI = new URI(this.props.node.transport_address).directory('api-browser');
     return (
       <div className="item-actions">
         <LinkContainer to={Routes.SYSTEM.NODES.SHOW(this.props.node.node_id)}>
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/nodes/NodesActions.jsx
```diff
@@ -38,6 +39,7 @@ const NodesActions = React.createClass({
     }
   },
   render() {
+    const apiBrowserURI = new URI(this.props.node.transport_address).directory('api-browser');
```
joschi: I'm not 100% sure, but this could cause a problem if the transport address contains a path segment with trailing slash:
```
> URI("http://example.com/foo").directory("bar").toString()
"http://example.com/bar/foo"
> URI("http://example.com/foo/").directory("bar").toString()
"http://example.com/bar/"
```

dennisoelkers: Good point! urijs does not help us at all then :/

joschi: While certainly not nice, this would work:
```
> URI("http://example.com/foo" + "/" + "bar").toString()
"http://example.com/foo/bar"
> URI("http://example.com/foo/" + "/" + "bar").toString()
"http://example.com/foo//bar"
```
Or with the given code snippet:
```javascript
    const apiBrowserURI = new URI(this.props.node.transport_address + '/api-browser');
```
So in the end, this isn't *much* better than simply concatenating those strings but at least the result URI is normalized.

dennisoelkers:
This is based on `master` branch, so it would need to be merged/cherry-picked into `2.0` after merging too.

joschi:
LGTM. :+1:

PR #2136, opened 2016-04-25
UI improvements

edmundoa:
This PR takes care of some small UI improvements:

- Hide indexer failures table when there are no failures, showing a message instead
- Replace rotation with retention in index retention strategies
- Move system job cancel button to the right

Changes should be merged in master and 2.0.


bernd:
LGTM :+1:

PR #2139, opened 2016-04-25
Display not found page

edmundoa:
This adds the 404 page back when the route doesn't match another component. In the end I'm sending the "funny" in the PR. The image is under a Creative Commons Zero license, so don't think there is any problem using it: http://gratisography.com/terms.html

We can also ship a more serious version if necessary :smile:


bernd:
I like the funny 404 page. Any other opinions? @dennisoelkers @joschi @kroepke @lennartkoopmann @mariussturm

joschi:
I like the funny image but can we make it a little bit smaller? I find ~500 KB and 1836x1224 pixels a little too much…

edmundoa:
@joschi I will try, in the end it depends on how it affects the image quality. The original is 5MB so getting it down to 500KB was already a big weight loss :smile:

bernd:
LGTM :+1:

PR #2152, opened 2016-04-27
Enable/Disable Input Extractor

kira8565:
Add API For Backend to Toggle The Extractor,Sometimes, users don't want
to delete an extractor, they just want to temporarily disable it and
wait … until they need to start The Extractor


joschi:
@kira8565 Could you please clean up your PR a little bit? For example squash your commits, remove pure whitespace and Javadoc changes (i. e. adding those `<p>` tags), and add a more detailed description of what you're trying to accomplish with those changes.

Additionally, why use an `int` to present the state of the extractor instead of either a simple `boolean` or a proper `enum`?

kira8565:
Thank you so much for reminding me that I'm going to make it better ヽ(•̀ω•́ )ゝ

kira8565:
Hello,I refact the code,and re-commit the feature,  this feature can let the extractor support suspend and enabled operations, when we temporarily do not need an extractor, we can pause it, when we need to, we can enable it.

Here's a screenshot of this new feature.


I added a new Boolean attribute for the Extractor class to determine whether the Extractor is enabled, and the name is *extractor_enabled*
Then, I add the condition for * filter*  method of the ExtractorFilter class.

` 
        
        for (final Extractor extractor : loadExtractors(msg.getSourceInputId())) {
            try {
                if (extractor.getExaractorEnabled() == true) {
                    extractor.runExtractor(msg);
                }
           `

In the end, I added a button for the Extractor interface to enable or disable the specified Extractor

kira8565:
I am Sorry becouse the GFW....the screenshot of these feature seems can not upload =。=||

dennisoelkers:
@kira8565: thanks for your efforts supplying a PR. Unfortunately it is very messy and changing a lot of stuff, that is not relevant for this functionality. Please open a new PR which includes only the changes that the functionality you are adding requires.

PR #2156, opened 2016-04-28
Allow URI for X-Graylog-Server-URL to be relative.

dennisoelkers:
This helps us to not rely on having a properly configured, publicly
visible, consistent absolute URL for the Graylog server. This makes
configuration a lot simpler for example for setups where both the web and
the REST port are exposed on a single port with a URL prefix for the
latter ("/api" for example), because in this case we can set
`X-Graylog-Server-URL` to a relative URI, regardless of the hostname.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
@@ -17,6 +17,7 @@
 package org.graylog2.web.resources;
 
 import com.floreysoft.jmte.Engine;
+import com.google.common.base.Strings;
 import com.google.common.io.Resources;
 import org.graylog2.Configuration;
 import org.graylog2.rest.MoreMediaTypes;
@@ -28,7 +29,8 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
 import java.io.IOException;
-import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


kroepke:
@dennisoelkers I either struggle to configure this correctly, or there's a missing piece in the setup:

I have configured nginx like:
```
  server
    {
      listen      6001;
      server_name 172.16.0.17;

      location /
      {
          proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
          proxy_set_header    Host $http_host;
          proxy_set_header    X-Graylog-Server-URL /api;
          proxy_pass          http://127.0.0.1:8080;
      }

      location /api/
      {
          proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
          proxy_set_header    Host $http_host;
          proxy_pass          http://127.0.0.1:12900/;
      }
    }
```

Using that configuration and visiting `http://172.16.0.17:6001/` works as expected, but the internal metrics calls fail with:
```
2016-05-04 14:45:41,409 WARN : org.graylog2.shared.rest.resources.ProxiedResource - Unable to call http://172.16.0.17:6001/system/metrics/multiple on node <3040048c-8935-4c24-8307-0218fa6fbeeb>, result: Not Found
```

It appears the resource doesn't take the correct url into account. This might be unrelated.

Relevant server config:
```
rest_listen_uri = http://127.0.0.1:12900/
rest_transport_uri = http://172.16.0.17:6001/api/
web_endpoint_uri = http://172.16.0.17:6001/
```
This is running the webpack dev server, but that should not make a difference.

kroepke:
In this case the metrics cannot be loaded, of course.

dennisoelkers:
For this scenario, you neither need to set `rest_transport_uri` or `web_endpoint_uri`. The Graylog server shouldn't use the proxy for calling itself (or other nodes), so it should use `rest_listen_uri`. This is actually the good thing about the `X-Graylog-Server-URL`-header, because it allow per-proxy configuration and  does not require special server configuration.

kroepke:
After correcting my server config, by leaving out `rest_transport_uri`, this works as intended.

PR #2157, opened 2016-04-28
Returning an explicit session validation response instead of http status

dennisoelkers:
This helps us to prevent basic auth popups showing up for the request,
because of browsers intercepting 401s for non-CORS requests. Helps a lot
when the web interface is served over a single port using a relative
server URL like proposed and enabled in #2156.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
@@ -1,0 +1,42 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
```diff
@@ -0,0 +1,42 @@
...
+package org.graylog2.rest.models.system.sessions.responses;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+@JsonAutoDetect
+public abstract class SessionValidationResponse {
```
joschi: Maybe it would be useful to add the user name of the user who created the session in the response.

dennisoelkers: Good idea. That would also help us because we don't need to keep it around in `localStorage` in the web interface. This would be done in a separate PR though.

joschi:
LGTM.

PR #2172, opened 2016-05-02
Matching inverted exact/regex stream rules when field is not present

### LINKED ISSUE #2160: Bug in inverted stream rule
### Problem description
Messages without the test field are also excluded

1. ...Make a rule:
we can select all messages with "Field facility must match exactly runit-service" but not the reverse,
because "Field facility must not match exactly runit-service" also excluded the messages without the field facility or with empty field.

### Environment

* Graylog Version: 2.0.0
* MongoDB Version: 3.2.1
* Operating System: REDHAT 6.6
* Browser version: Firefox, chrome

dennisoelkers:
Before this change, inverted exact/regex stream rules did not match when the field was not present. This was counter intuitive and perceived as erroneus, this PR Is changing this behavior.

Fixes #2160

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/matchers/ExactMatcher.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/matchers/ExactMatcher.java
@@ -26,17 +26,13 @@ public class ExactMatcher implements StreamRuleMatcher {
 
 	@Override
 	public boolean match(Message msg, StreamRule rule) {
-        if (!msg.hasField(rule.getField())) {
-            return false;
+        if (msg.getField(rule.getField()) == null) {
+            return rule.getInverted();
         }
 
-		Object value = msg.getField(rule.getField());
-		
-		if (value == null) {
-			return false;
-		}
-		
-		return rule.getInverted() ^ value.toString().trim().equals(rule.getValue());
+		final String value = msg.getField(rule.getField()).toString();
+
+		return rule.getInverted() ^ value.trim().equals(rule.getValue());
 	}
 
 }
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -36,6 +36,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/streams/matchers/ExactMatcher.java
```diff
@@ -26,17 +26,13 @@
...
-		Object value = msg.getField(rule.getField());
-		
-		if (value == null) {
-			return false;
-		}
-		
-		return rule.getInverted() ^ value.toString().trim().equals(rule.getValue());
+		final String value = msg.getField(rule.getField()).toString();
+
+		return rule.getInverted() ^ value.trim().equals(rule.getValue());
```
joschi: While this works, I'd prefer using the logical OR operator here.

dennisoelkers: This is not part of the scope of this PR.

joschi: I think we could do some [opportunistic refactoring](http://martinfowler.com/bliki/OpportunisticRefactoring.html) in this case (and actually fix an error that is accidentally working).

dennisoelkers: What is the error here?

joschi: Using the bitwise OR operator (`^`) instead of the logical OR operator (`||`) as it's a logical expression to be evaluated and not some bit-shifting.

dennisoelkers: This is perfectly fine for booleans, as specified in the JLS: http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.22.2

joschi: Hm, I think I'm getting confused now. The `^` bitwise operator and the `||` logical operator also have different characteristics:

> For `^`, the result value is `true` if the operand values are different; otherwise, the result is `false`.

Is this the intended behavior? The `||` operator would return `true` if at least one of the operands is `true` while the `^` operator would return `true` only if the operands are different (e. g. both are `false`).

dennisoelkers: Yeah, this is intended. The critical difference is that if the rule is inverted, then it should not match if the message and the rule fields are equal.

joschi: ✅ I interpreted the intention of that expression completely wrong. Everything's fine.

Everything's fine

joschi:
LGTM. 👍

PR #2174, opened 2016-05-02
Fix insecure write to PID file

juergenhoetzel:
Ensure PID file doesn't exist and don't follow symlinks.

See: https://en.wikipedia.org/wiki/Symlink_race

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
+++ b/graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
@@ -44,8 +44,10 @@ import org.slf4j.LoggerFactory;
 
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.LinkOption;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


joschi:
@juergenhoetzel Thank you for your contribution to Graylog!

Before we can accept any code from you, we need you to sign our [Contributor License Agreement](https://s3.amazonaws.com/graylog2public/Individual+Contributor+Assignment+Agreement+-+Graylog.pdf). Please send us a signed copy to hello@graylog.com so we can merge your pull request.

This is only required once, no matter how many contributions you make in the future. Thank you for helping us with this unfortunate administrative requirement!

graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
```diff
@@ -157,7 +159,7 @@ protected void savePidFile(final String pidFile) {
                 throw new Exception("Could not determine PID.");
             }
 
-            Files.write(pidFilePath, pid.getBytes(StandardCharsets.UTF_8));
+            Files.write(pidFilePath, pid.getBytes(StandardCharsets.UTF_8), StandardOpenOption.WRITE, StandardOpenOption.CREATE_NEW, LinkOption.NOFOLLOW_LINKS);
```
joschi: Wouldn't it make sense to use `CREATE` and `TRUNCATE_EXISTING` instead of `CREATE_NEW` in this place? If the file already exists and `CREATE_NEW` is being used, the call would fail instead of overwriting the specified file.

juergenhoetzel: > Wouldn't it make sense to use CREATE and TRUNCATE_EXISTING instead of CREATE_NEW in this place? If the file already exists and CREATE_NEW is being used, the call would fail instead of overwriting the specified file.

This is exactly the purpose for this change. 

An attacker will place a symlink like
```
ln -s /PATH/TO/OVERWRITE_USING_GRAYLOG_USER  /tmp/graylog.pid
```
in the public ```/tmp``` folder.

In this case the startup should fail.

joschi: ✅ Makes sense. Thanks!

joschi:
LGTM. :+1:

PR #2181, opened 2016-05-04
Add back rotation/retention config options in misc/graylog.conf

bernd:
Also add a warning that these settings should be kept when upgrading to
2.0.0 to ensure migration of the settings.

This reverts parts of 5820b4e.

(cherry picked from commit e0c810c4cdf539cb68cdc0b882386c9872aaedbd)


joschi:
LGTM. :+1:

PR #2182, opened 2016-05-04
Make alert conditions resilient against changing number formats.

### LINKED ISSUE #2169: Stream alerts no longer firing after upgrade to 2.0
### Problem description
Stream alerts are no longer firing after upgrading to 2.0 even though the stream receives messages.

The Message count condition : 
Alert is triggered when there is more than one message in the last 3 minutes. Grace period: 0 minutes. Including last message in alert notification.

The above worked just fine in 1.3 but apparently no longer works in 2.0 - no alerts have been generated since upgrading.

There is a discrepancy between editing the existing message count condition and adding a new one which I cannot explain : 

screen shot 2016-05-02 at 11 11 33

dennisoelkers:
In 1.3.x numerical alert condition parameters were stored by MongoDB
(resp. the mongo java driver) in a way, that they are returned as
Doubles to the server, while alert conditions persisted in 2.0.0 return
numerical parameters as Integer. Therefore, 2.0.0 cannot handle alert
conditions stored in 1.3.x, so legacy installations upgraded to 2.0.0
do not check any alert conditions anymore.

This change allows to handle both formats by generally assuming they are
Number/Double types and returning their `intValue()`.

Fixes #2169

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
@@ -26,6 +26,7 @@ import org.graylog2.plugin.database.EmbeddedPersistable;
 import org.graylog2.plugin.streams.Stream;
 import org.joda.time.DateTime;
 
+import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
@@ -63,12 +64,7 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
         this.creatorUserId = creatorUserId;
         this.parameters = parameters;
 
-        if (this.parameters.containsKey("grace")) {
-            this.grace = (Integer) this.parameters.get("grace");
-        } else {
-            this.grace = 0;
-        }
-
+        this.grace = getInt(this.parameters.get("grace"), 0);
     }
 
     protected abstract AlertCondition.CheckResult runCheck();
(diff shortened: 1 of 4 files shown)


dennisoelkers:
Adding some tests as well.

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
```diff
@@ -195,4 +187,16 @@ public NegativeCheckResult(AlertCondition alertCondition) {
         }
     }
 
+    @VisibleForTesting
```
joschi: Minor semantic nitpick: Child classes are also using this (protected) method, so it's not just visible for testing. :wink:

graylog2-server/src/test/java/org/graylog2/alerts/AbstractAlertConditionTest.java
```diff
@@ -16,16 +16,17 @@
...
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
 import java.util.Map;
+import java.util.Optional;
 
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
joschi: Please don't use wildcard imports.

Configuration for IntelliJ IDEA: http://stackoverflow.com/a/3348855

joschi:
LGTM (except for those 2 minor nitpicks). :+1:

graylog2-server/src/test/java/org/graylog2/alerts/AbstractAlertConditionTest.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
dennisoelkers: ✅

joschi:
LGTM. :+1:

PR #2204, opened 2016-05-09
Also include webpack.config.js in a submodule

bernd:
Needed for plugins where the `webpack.config.js` file is in a submodule. Not optimal, but better than hardcoding paths.

(Needs to be merged into master as well!)


joschi:
LGTM. :+1:

PR #2206, opened 2016-05-09
Improve robustness for unreachable nodes

dennisoelkers:
Before this change, some pages were not loading when at least one node in the cluster was unreachable. With this change, information from reachable nodes is shown. Additionally, an error similar to the 404 page was created, which shows failed requests for pages which do not make sense to show at all when a node is not reachable.


edmundoa:
I did some small word changes in the new error page. The rest looks good to me 👍

PR #2208, opened 2016-05-10
Adding check is path is unset (or "/") for `web_listen_uri` parameter.

### LINKED ISSUE #2191: Graylog 2.0 web server - Unable to specify application context
### Problem description
In GL 2.0, an application context specified in 'web_listen_uri' is not honored.

1. Set 'web_listen_uri' to something like http://0.0.0.0:9000/graylog and it doesn't work.. site does bind to that URI, but all artifact paths are not re-written and remain at the root level, rather than out of the web context, so the page will not load.  This causes an issue on a shared hostname I have, where I route requests out of nginx from based on web context.  In graylog 1.x, you could specify the setting 'application.context' for this, but that setting has been removed from 2.0.

### Environment

* Graylog Version:  2.0
* Elasticsearch Version: 2.0
* MongoDB Version: 2.6.4
* Operating System: OEL 6.7
* Browser version: chrome/ie 11

dennisoelkers:
Right now we cannot handle a path prefix for the web interface assets, because of the way they are bundled. The config parameter allows one, so we need to explicitly check for it and raise an error when it is set to prevent undefined/unexpected behavior.

Fixes #2191.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
@@ -1,0 +1,31 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.configuration;
+
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.Validator;
+
+import java.net.URI;
+
+public class WebListenUriValidator implements Validator<URI> {
+    @Override
+    public void validate(String name, URI value) throws ValidationException {
+        if (value != null && !value.getPath().equals("/")) {
+            throw new ValidationException("Parameter " + name + " must not contain a path prefix (found " + value + ")");
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.Validator;
+
+import java.net.URI;
+
+public class WebListenUriValidator implements Validator<URI> {
+    @Override
+    public void validate(String name, URI value) throws ValidationException {
+        if (value != null && !value.getPath().equals("/")) {
```
joschi: Unfortunately [`URI#getPath()`](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#getPath--) can return `null` which would lead to a `NullPointerException` here. Maybe just invert the second condition (`"/".equals(…)`).

graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
```diff
@@ -0,0 +1,31 @@
...
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.configuration;
+
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.Validator;
+
+import java.net.URI;
+
+public class WebListenUriValidator implements Validator<URI> {
```
joschi: Just FYI, this could also be done with a [`ValidatorMethod`](https://joschi.github.io/JadConfig/0.12.2/apidocs/com/github/joschi/jadconfig/ValidatorMethod.html) directly in `Configuration `, but having a dedicated validator is also fine.

graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
dennisoelkers: :+1: - thanks for the hint! :)

graylog2-server/src/main/java/org/graylog2/configuration/WebListenUriValidator.java
dennisoelkers: Thanks for the heads up!

joschi:
LGTM. 👍

PR #2210, opened 2016-05-10
Getting index statistics could contain failed shards

### LINKED ISSUE #2194: Every odd index rolled too early
### Problem description
Every second index stays unfilled to configured condition.
image
image

### Environment
* Graylog Version: 2.0 GA

kroepke:
During rotation and retention we requested all index statistics multiple times, which, in a overloaded cluster, could lead to shard failures due to timeouts.
This failure wasn't logged and could lead to using the wrong (older) index to base rotation decisions on, effectively rotating indices too early.

This change makes Graylog use a more lightweight API to determine all index names including their aliases, reducing the usage of the expensive Index Statistics to the indices page only.
The current rotation and retention strategies do not need to know all index statistics which require to touch every single shard in the cluster.

fixes #2194

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -16,9 +16,6 @@
  */
 package org.graylog2.indexer;
 
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Lists;
-import org.elasticsearch.action.admin.indices.stats.IndexStats;
 import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.indices.InvalidAliasNameException;
 import org.graylog2.indexer.indices.Indices;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
```diff
@@ -182,29 +180,31 @@ public void cycle() {
...
 
             try {
-                indexNumbers.add(extractIndexNumber(indexName));
+                final int indexNumber = extractIndexNumber(indexName);
+                highestIndexNumber = Math.max(indexNumber, highestIndexNumber);
             } catch (NumberFormatException ex) {
-                LOG.debug("Couldn't extract index number from index name " + indexName, ex);
+                LOG.warn("Couldn't extract index number from index name " + indexName, ex);
             }
         }
```
dennisoelkers: What about replacing this with a streams call like this:

```
        final Optional<Integer> highestIndexNumber = indexNames.stream()
            .filter(indexName -> !this.isGraylogDeflectorIndex(indexName))
            .map(Deflector::extractIndexNumber)
            .max(Integer::max);
```
Makes it easier to grasp what is done (at least for me).

kroepke: In theory I agree, but in this case we'd need to refactor extractIndexNumber as well and its callers, because it throws an exception, and it already felt I changed a lot of code already :(

How about making those changes on master?

bernd:
LGTM :+1:

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
dennisoelkers: 👍

Issue #2200, opened 2016-05-07
Problem with access to the Rest_Api


Myote:
### Problem description
Is there a problem in 2.0.0 with REST API browser ?  Numerous timeout and I can't export to csv : {"type":"ApiError","message":"HTTP 404 Not Found"}

For the timeout I found the pb, it's when rest_enable_gzip = true; I removed it, but pb with export csv is always there.

My test configuration:
rest_listen_uri = http://0.0.0.0:12900/
rest_transport_uri = http://192.168.1.88:12900/
rest_enable_cors = true
#rest_enable_gzip = true  (comment to prevent timeout ...)

### Steps to reproduce the problem

1. ...

### Environment

* Graylog Version: 2.0.0-5 or 6
* Elasticsearch Version: 2.3.2
* MongoDB Version: 3.2.1
* Operating System: RHEL 6.6
* Browser version: all

kroepke:
Which request are you executing? In general the API browser works in 2.0.

Could you provide a screenshot of the resources you see for the search API?

Myote:
Hi Kroepke,

- If one of this var 'rest_listen_uri = http://0.0.0.0:12900/' and '#rest_transport_uri = http:192.168.1.88:12900/' is filled with a reel ip and not 0.0.0.0, then 'Export as CSV' in main search page is broken because Graylog try to resolv an url with // (x.x.x.x//search/...).
![sans titre](https://cloud.githubusercontent.com/assets/12958868/15131982/0c7b5f3e-1655-11e6-9399-fd34c9f07f4a.jpg)

- If rest_enable_gzip = true is active then there is numerous timeout on x.x.x.x:12900/api-browser before the page is eventually displayed after many secondes.
![capture1](https://cloud.githubusercontent.com/assets/12958868/15132057/fab85576-1655-11e6-8267-d2dabc5a9560.JPG)

edmundoa:
I can confirm there is a double slash in the generated URL to export the search results as CSV in certain situations. Will work on a fix for it.

jalogisch:
@Myote could you please verify if this issue is still present in the present version?

please reopen if still present - until now i will close.

PR #2213, opened 2016-05-10
Normalize URIs

edmundoa:
In some situations, the `gl2ServerUrl` configuration option may end in a slash. As we can't predict when that will happen, this commit is normalizing the path of generated URIs, avoiding URIs with duplicated slashes in their path (you can see an example in #2200).

I was able to reproduce the double slash issue in URIs by setting `rest_listen_uri` and `rest_transport_uri` to `http://127.0.0.1:12900/`. The most visible place to see them is the export CSV URL. Other URIs are also affected, but they are normalized by one of the network libraries we use.


joschi:
LGTM. :+1:

PR #2231, opened 2016-05-13
MessageProcessors must be instantiated per processing thread

### LINKED ISSUE #2119: Drools applies not to all messages, if more than 1 processor enabled
### Problem description
If I set processbuffer_processors to more than 1, Drools will process not every message

1. Set processbuffer_processors in server.conf to 3 (for example)
2. Create simple Drools rule, which will just create new field for every message:
`rule "Test Rule"
    when
        m : Message()
    then
        m.addField("test", "test_value");
end
`
3. Push some messages
4. Not all messages will have new field

### Environment
* Graylog Version:2.0 beta3
* Elasticsearch Version:2.3.1

### LINKED ISSUE #2188: rules.drl not always updating source field
### Problem description
We have the following Drools code **A** and code **B** working in Graylog 1.3.x
##### Code A
```
// ==================================================================
// Rewrite source and remove .company.com
// ------------------------------------------------------------------
rule "Rewrite remove .company.com from source name"
        when
                m : Message( source matches "(.*)\\.company\\.com" )
        then
                Matcher matcher = Pattern.compile("(.*)\\.company.com",Pattern.CASE_INSENSITIVE).matcher(m.getField("source").toString());
                System.out.println("[.company.com]: Found " + m.getField("source"));
                if (matcher.find()) {
                        m.addField("source", matcher.group(1).toLowerCase());
                        System.out.println("[.company.com]: source updated " + matcher.group(1).toLowerCase());
                }
end
```

###### Code A Output:
```
[.company.com]: Found ABC.company.com
[.company.com]: source updated abc
```

##### Code B
```
// ==================================================================
// Rewrite ip-192-168-1-100 to vm17
// ------------------------------------------------------------------
rule "Rewrite ip-192-168-1-100 host"
        when
                m : Message( source == "ip-192-168-1-100" )
        then
                m.addField("source", "vm17" );
                System.out.println( "[Overwrite ip-192-168-1-100 rule fired] : " + m.toString() );
end
```

With Graylog 2.0, 
Code B is still working fine, however Code A even though the System.out is printing the same messages as 1.3.x, the source field in Graylog is not updated
and is kept as ABC.company.com

quite confused as to why it would work for code B but not code A in Graylog 2.0

* Enable Drools
* include the Code A and B  in /etc/graylog/server/rules.drl
* Send Windows Event logs from ABC.company.com through nxlog 

### Environment

* Graylog Version: 2.0
* Elasticsearch Version: 2.3.2
* MongoDB Version: 3.2.6
* Browser version: Chrome

kroepke:
When introducing the MessageProcessor interface, the processing threads accidentally shared the instances (and by induction the MessageFilter instances as well).
That posed no problem for most of the filters, because they do not rely on shared state, but the Drools filter does and could skip messages (because of Drools itself returning early)

This change uses a Provider to get the OrderedMessageProcessor instances explicitly and those do not get shared across threads.

fixes #2119
fixes #2188

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
@@ -28,7 +28,6 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
-import javax.inject.Provider;
 
 /**
  * @author Dennis Oelkers <dennis@torch.sh>
@@ -42,10 +41,10 @@ public class ServerProcessBufferProcessor extends ProcessBufferProcessor {
 
     @Inject
     public ServerProcessBufferProcessor(MetricRegistry metricRegistry,
-                                        Provider<OrderedMessageProcessors> orderedMessageProcessorsProvider,
+                                        OrderedMessageProcessors orderedMessageProcessors,
                                         OutputBuffer outputBuffer) {
         super(metricRegistry);
-        this.orderedMessageProcessors = orderedMessageProcessorsProvider.get();
+        this.orderedMessageProcessors = orderedMessageProcessors;
         this.outputBuffer = outputBuffer;
     }
 
--- a/graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
+import javax.inject.Provider;
 
 /**
  * @author Dennis Oelkers <dennis@torch.sh>
@@ -41,10 +42,10 @@ public class ServerProcessBufferProcessor extends ProcessBufferProcessor {
 
     @Inject
     public ServerProcessBufferProcessor(MetricRegistry metricRegistry,
-                                        OrderedMessageProcessors orderedMessageProcessors,
+                                        Provider<OrderedMessageProcessors> orderedMessageProcessorsProvider,
                                         OutputBuffer outputBuffer) {
         super(metricRegistry);
-        this.orderedMessageProcessors = orderedMessageProcessors;
+        this.orderedMessageProcessors = orderedMessageProcessorsProvider.get();
         this.outputBuffer = outputBuffer;
     }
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
```diff
@@ -41,10 +42,10 @@
 
     @Inject
     public ServerProcessBufferProcessor(MetricRegistry metricRegistry,
-                                        OrderedMessageProcessors orderedMessageProcessors,
+                                        Provider<OrderedMessageProcessors> orderedMessageProcessorsProvider,
                                         OutputBuffer outputBuffer) {
         super(metricRegistry);
-        this.orderedMessageProcessors = orderedMessageProcessors;
+        this.orderedMessageProcessors = orderedMessageProcessorsProvider.get();
```
joschi: Why use a `Provider` here? If `OrderedMessageProcessors.class` isn't bound as a singleton, Guice will create a new instance anyway.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorModule.java
```diff
@@ -22,8 +22,8 @@
     @Override
     protected void configure() {
         addMessageProcessor(MessageFilterChainProcessor.class, MessageFilterChainProcessor.Descriptor.class);
-
-        bind(OrderedMessageProcessors.class).asEagerSingleton();
+        // must not be a singleton, because each thread should get an isolated copy of the processors
+        bind(OrderedMessageProcessors.class);
```
joschi: Maybe make it very explicit here with `bind(OrderedMessageProcessors.class).in(Scopes.NO_SCOPE)`?

graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
kroepke: I was being extra-super-duper careful and explicit, but you are correct, it is not strictly necessary.

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageProcessorModule.java
kroepke: aye!

graylog2-server/src/main/java/org/graylog2/buffers/processors/ServerProcessBufferProcessor.java
joschi: The thing is that a (singleton) provider could also return always the same instance of the injectee and it doesn't help clarifying the contract (that we need a new instance every time).

joschi:
LGTM. 👍

PR #2244, opened 2016-05-17
Stream Rule Titles

dennisoelkers:
This adds the necessary functionality allowing users to define descriptions for individual stream rules.
It includes:

  * Extending the `StreamRule` persistence/service classes and DTOs with the description attribute
  * Extending the Stream Rules form and display components to include the description
  * Refactoring all Stream Rules React components on the fly, fixing linter hints

### DIFF
--- a/graylog2-web-interface/src/components/streamrules/StreamRuleForm.jsx
+++ b/graylog2-web-interface/src/components/streamrules/StreamRuleForm.jsx
@@ -1,99 +1,107 @@
-'use strict';
+import React from 'react';
+import { Col, Input } from 'react-bootstrap';
+import LinkedStateMixin from 'react-addons-linked-state-mixin';
 
-var React = require('react');
-var LinkedStateMixin = require('react-addons-linked-state-mixin');
-var BootstrapModalForm = require('../bootstrap/BootstrapModalForm');
-var Input = require('react-bootstrap').Input;
-var HumanReadableStreamRule = require('./HumanReadableStreamRule');
-var Col = require('react-bootstrap').Col;
-var TypeAheadFieldInput = require('../common/TypeAheadFieldInput');
+import BootstrapModalForm from 'components/bootstrap/BootstrapModalForm';
+import { TypeAheadFieldInput } from 'components/common';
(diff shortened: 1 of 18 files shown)


graylog2-web-interface/src/components/streamrules/StreamRuleForm.jsx
```diff
@@ -1,99 +1,107 @@
...
-                            {' '}
-                            Field <HumanReadableStreamRule streamRule={this.state}
-                                                           streamRuleTypes={this.props.streamRuleTypes}/>
-                        </p>
-                    </Col>
-                    <Col md={4}>
-                        <div className="well well-sm matcher-github">
-                            The server will try to convert to strings or numbers based on the matcher type as good as it
-                            can.
+            <Input type="textarea" label="Description" valueLink={this.linkState('description')} />
```
edmundoa: We use text inputs for most descriptions: do we expect to have such long descriptions here that we need a textarea?

graylog2-web-interface/src/components/streamrules/StreamRuleList.jsx
```diff
@@ -1,34 +1,43 @@
...
+    if (this.props.stream) {
+      const streamRules = this._formatStreamRules(this.props.stream.rules);
+      return (
+        <ul className="streamrules-list">
+          {streamRules}
+        </ul>
+      );
     }
+    return (
+      <div><i className="fa fa-spin fa-spinner"/> Loading</div>
```
edmundoa: We could use the `Spinner` component in here.

graylog2-web-interface/src/components/streamrules/StreamRuleForm.jsx
edmundoa: Nevermind the previous comment, looking at the modal itself, I think that was the right choice, as the text input would look too small. Also, we have plenty of space to add the description alongside the stream rule.

edmundoa:
LGTM 👍 

I will replace the loading text with the `Spinner` component on master, no need to have this PR around for longer.

PR #2246, opened 2016-05-17
Set Cache-Control on rest resources

### LINKED ISSUE #2243: Caching issues on IE and Edge
### Problem description
Extract from [this mailing list thread](https://groups.google.com/forum/#!topic/graylog2/EhHIX-jcXdM):

On the search tab, using the "relative" search options, if I select "search in the last 5 minutes" at 7:30Am, and then I select "search in the last 15 minutes" at 7:45Am, both will return the correct time range of data the first time they are used as a search parameter.  However, if I do the same searches 2 hours later, it will retrieve the same data (7:25-7:30 and 7:30-7:45).  If I view the JSON query for these searches, the time frame it uses does not change.  I am always able to retrieve the most recent data if I use a search time frame I have not used yet. 

This behavior is seen in both 2.0.0 and 2.0.1, and appears in both of my graylog servers. (...) I also don't see any JS errors in the console, I am running IE11 (if it matters), and there are no proxies between the browser and graylog.  This behavior continues until I close the browser window and reopen it.

1. Go to the search page in IE11 or MS Edge
2. Execute a relative search in the last 15 minutes
3. Check the `from` and `to` fields in the ES search query
4. Wait a minute
5. Execute the search again
6. See that the `from` and `to` fields in the ES search query for the new search are identical. Also check that the XHR request for the search is being delivered from cache

### Environment

* Graylog Version:  2.0.0 and 2.0.1
* Browser version: IE11 / MS Edge

edmundoa:
Some browsers cache Ajax requests (like IE/Edge), so we need to set the `Cache-Control` header to `no-cache` to avoid caching issues.

In order to only set cache headers into requests coming from browsers, we now set the `X-Requested-With` header in all our XHR requests coming from the web interface.

Fixes #2243.

This PR should also be merged into master.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/initializers/AbstractJerseyService.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/initializers/AbstractJerseyService.java
@@ -37,6 +37,7 @@ import org.graylog2.shared.rest.CORSFilter;
 import org.graylog2.shared.rest.NodeIdResponseFilter;
 import org.graylog2.shared.rest.PrintModelProcessor;
 import org.graylog2.shared.rest.RestAccessLogFilter;
+import org.graylog2.shared.rest.XHRFilter;
 import org.graylog2.shared.rest.exceptionmappers.AnyExceptionClassMapper;
 import org.graylog2.shared.rest.exceptionmappers.BadRequestExceptionMapper;
 import org.graylog2.shared.rest.exceptionmappers.JacksonPropertyExceptionMapper;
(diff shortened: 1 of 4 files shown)


joschi:
LGTM. 👍

edmundoa:
It seems that these changes also have the added benefit of removing that annoying login popup window triggered in the login page by an Ajax request.

edmundoa:
Forget my last comment, that was already fixed in https://github.com/Graylog2/graylog2-server/pull/2157

Issue #2250, opened 2016-05-18
OutputBuffer metrics displayed for ProcessBuffer


hc4:
### Problem description
When click metrics on ProcessBuffer, OutputBuffer metrics displayed:
![image](https://cloud.githubusercontent.com/assets/17475270/15356279/8a83c720-1d00-11e6-9d8f-b1844db18c88.png)
![image](https://cloud.githubusercontent.com/assets/17475270/15356314/afd106fa-1d00-11e6-8f46-8c076f2b59ec.png)

Maybe it is better to show all ProcessBuffer metrics?
![image](https://cloud.githubusercontent.com/assets/17475270/15356330/c26e704a-1d00-11e6-807a-c2bac527053a.png)

And same for OutputBuffer and InputBuffer metrics

### Environment

* Graylog Version: 2.0.1

PR #2255, opened 2016-05-18
Metrics improvements

edmundoa:
While looking into #2250, I found a couple of issues with the metrics page:

- Doing a full page refresh on the metrics page raised an exception, as MetricsStore.name() was being called before the store had initialised the node list
- Writing an invalid regex in the metrics filter would raise an exception, and also break the page

This PR fixes those two issues, and also updates the filters to node buffers.


edmundoa:
I based the changes on master, but they should also be included in 2.0.2. Please let me know if there's any conflict doing so. Thank you!

PR #2277, opened 2016-05-24
Fix changing log level via REST API

### LINKED ISSUE #1904: Changing subsystem on Logging page changes all subsystems
### Problem description

Changing any Subsystem on the Logging page (except "Subsystem: Graylog") changes the level for all subsystems.

1. Go to System/Logging
2. Select "debug" for "Subsystem: Indexer"
3. Log level for "Sockets", "Indexer" and "Authentication" changes to "debug"
4. Check API request "/system/loggers/subsystems", same issue (so it's not a UI problem)

#### Analysis

It looks like the [LoggersResource#setLoggerLevel()](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java#L145-L151) method changes the root logger instead of the requested class because there is no explicit config for that. The graylog subsystem does not change because our default `log4j2.xml` has an explicit configuration for that.

I guess this is a behaviour change in log4j2.

### Environment

* Graylog Version: 2.0.0-alpha.5
* Elasticsearch Version: 2.2.0
* MongoDB Version: 2.4.14

joschi:
Fixes #1904

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
@@ -17,6 +17,7 @@
 package org.graylog2.rest.resources.system.logs;
 
 import com.codahale.metrics.annotation.Timed;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


joschi:
This should also be ported to `master`.

Issue #2264, opened 2016-05-19
Wrong Alert generated


hc4:
### Problem description
Just got wrong Alert.

Alert definition:
![image](https://cloud.githubusercontent.com/assets/17475270/15400814/9b1e212a-1df5-11e6-9ac9-632aa25de589.png)

Alert query generated correctly and it finds item:
`messages?rangetype=absolute&from=2016-05-18T15:00:20.372Z&to=2016-05-19T15:00:20.372Z&q=*
`And it correctly finds message at 2016-05-19T03:08:09.780Z
And I assured, that this message was here before Alert generated.

I've checked logs, but there is no errors.
Only thing I've mentioned, is that just at time of Alert there was index cycling procedure.
And the message was exactly from cycled index (message from graylog2_100 and cycling was from graylog2_100 to graylog2_101)

### Environment

* Graylog Version: 2.0.1

dennisoelkers:
At what time did the index cycling take place?

hc4:
There is logs from graylog:
```
2016-05-19T17:59:49.711+03:00 INFO  [AbstractRotationStrategy] Deflector index <graylog2_100> should be rotated, Pointing deflector to new index now!
2016-05-19T17:59:49.711+03:00 INFO  [Deflector] Cycling deflector to next index now.
2016-05-19T17:59:49.712+03:00 INFO  [Deflector] Cycling from <graylog2_100> to <graylog2_101>
2016-05-19T17:59:49.712+03:00 INFO  [Deflector] Creating index target <graylog2_101>...
2016-05-19T17:59:49.996+03:00 INFO  [Indices] Created Graylog index template "graylog-internal" in Elasticsearch.
2016-05-19T17:59:50.152+03:00 INFO  [Deflector] Waiting for index allocation of <graylog2_101>
2016-05-19T17:59:50.312+03:00 INFO  [Deflector] Done!
2016-05-19T17:59:50.312+03:00 INFO  [Deflector] Pointing deflector to new target index....
2016-05-19T17:59:50.466+03:00 INFO  [CreateNewSingleIndexRangeJob] Calculating ranges for index graylog2_100.
2016-05-19T17:59:50.466+03:00 INFO  [SystemJobManager] Submitted SystemJob <56016b10-1dd2-11e6-a59f-005056b37533> [org.graylog2.indexer.ranges.CreateNewSingleIndexRangeJob]
[...]

PR #2278, opened 2016-05-24
Waiting for index range calculation before switching deflector alias.

dennisoelkers:
Before this change, whenever the deflector was cycled, a system job was submitted to calculate the index range of the current write index/deflector (index 1). Immediately after submitting the job (not necessarily after the index range calculation job has finished already), the deflector alias was cycled to a newly created index (index 2). When the index range calculation job takes longer than cycling the deflector alias, there is a time range, where searches do not take the index which was recently cycled (index 1) into account.

This happens because `IndexHelper#determineAffectedIndices` returns an index if:
  * it has a valid index range and the supplied time range is an intersection of it
  * it is the current deflector

Both conditions do not hold true for index 1 at this time. As searches are also used in alerts, there is also the potential of false positives/negatives, as described in #2264.

There are two ways to fix this:

1. Wait for the index range calculation to complete before cycling the deflector alias
2. Return all indices without valid time ranges in `IndexHelper#determineAffectedIndices` as well.

Solution 2 was not chosen, because it seems to be a too intrusive approach, as it has an effect on all searches being done, especially when indices are pretty large.

Instead, solution 1 was implemented, by enhancing the result type of the [`SystemJobManager#submit`](https://github.com/Graylog2/graylog2-server/blob/issue-2264/graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java#L82) to include a [`ScheduledFuture`](https://github.com/Graylog2/graylog2-server/blob/issue-2264/graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java#L42-L58) as well, which is then [used to wait for the completion of the index range calculation](https://github.com/Graylog2/graylog2-server/blob/issue-2264/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java#L168) before switching the deflector alias. As we do not want to wait an infinite amount of time for the unlikely event of noncompletion of this system job, a configuration property was [introduced](https://github.com/Graylog2/graylog2-server/blob/issue-2264/graylog2-server/src/main/java/org/graylog2/Configuration.java#L157-L158) defaulting to a duration of one hour, which may or may not be a good idea.

Solution 1 has the downside though, that there is a timeframe which is at maximum the time between the start of the index range calculation and the time of the actual index alias switch, which would not be included in the index range calculated and there might be the need for a second index range calculation after the index was switched to read only.

*UPDATE*: After tinkering a bit with the scenarios and possible implementations, a combination of both solutions was chosen. Now, when the deflector is cycled, a static index range with "empty" (== 0) begin/end parameters is saved. Index range calculation for the old deflector is postponed until after it was made read only, so we are sure to be able to calculate the correct time ranges. When searching, all indices with "empty" begin/end parameters are included in the search, because we cannot determine reliably which time range they cover.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.indexer;
 
+import com.github.joschi.jadconfig.util.Duration;
 import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.indices.InvalidAliasNameException;
 import org.graylog2.indexer.indices.Indices;
@@ -31,7 +32,6 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -165,7 +165,7 @@ public class Deflector { // extends Ablenkblech
             if (scheduleResult != null) {
                 try {
                     LOG.debug("Waiting for index calculation to complete");
-                    scheduleResult.getFuture().get(deflectorIndexRangeCalculationTimeout.toMillis(), TimeUnit.MILLISECONDS);
+                    scheduleResult.getFuture().get(deflectorIndexRangeCalculationTimeout.toMilliseconds(), TimeUnit.MILLISECONDS);
                     LOG.debug("Completed!");
                 } catch (InterruptedException | ExecutionException | TimeoutException e) {
                     LOG.warn("Unable to calculate index range of index which is to be cycled: ", e);
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -21,6 +21,8 @@ import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.indices.InvalidAliasNameException;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.indexer.ranges.CreateNewSingleIndexRangeJob;
+import org.graylog2.indexer.ranges.IndexRange;
+import org.graylog2.indexer.ranges.IndexRangeService;
 import org.graylog2.shared.system.activities.Activity;
 import org.graylog2.shared.system.activities.ActivityWriter;
 import org.graylog2.system.jobs.SystemJob;
@@ -61,7 +63,8 @@ public class Deflector { // extends Ablenkblech
     private final SystemJobManager systemJobManager;
     private final ActivityWriter activityWriter;
     private final CreateNewSingleIndexRangeJob.Factory createNewSingleIndexRangeJobFactory;
-    private final Duration deflectorIndexRangeCalculationTimeout;
+    private final IndexRangeService indexRangeService;
+    private final Duration deflectorIndexReadOnlyTimeout;
     private final String indexPrefix;
     private final String deflectorName;
     private final Indices indices;
@@ -76,14 +79,16 @@ public class Deflector { // extends Ablenkblech
                      final SetIndexReadOnlyJob.Factory indexReadOnlyJobFactory,
                      final CreateNewSingleIndexRangeJob.Factory createNewSingleIndexRangeJobFactory,
                      final Indices indices,
-                     @Named("deflector_index_range_calculation_timeout") final Duration deflectorIndexRangeCalculationTimeout) {
+                     final IndexRangeService indexRangeService,
+                     @Named("deflector_index_read_only_timeout")final Duration deflectorIndexReadOnlyTimeout) {
         this.indexPrefix = indexPrefix;
 
         this.systemJobManager = systemJobManager;
         this.activityWriter = activityWriter;
         this.indexReadOnlyJobFactory = indexReadOnlyJobFactory;
         this.createNewSingleIndexRangeJobFactory = createNewSingleIndexRangeJobFactory;
-        this.deflectorIndexRangeCalculationTimeout = deflectorIndexRangeCalculationTimeout;
+        this.indexRangeService = indexRangeService;
+        this.deflectorIndexReadOnlyTimeout = deflectorIndexReadOnlyTimeout;
 
         this.deflectorName = buildName(indexPrefix);
         this.indices = indices;
@@ -150,6 +155,7 @@ public class Deflector { // extends Ablenkblech
         ClusterHealthStatus healthStatus = indices.waitForRecovery(newTarget);
         LOG.debug("Health status of index <{}>: {}", newTarget, healthStatus);
 
+        addDeflectorIndexRange(newTarget);
         LOG.info("Done!");
 
         // Point deflector to new index.
@@ -161,19 +167,6 @@ public class Deflector { // extends Ablenkblech
             pointTo(newTarget);
             activity.setMessage("Cycled deflector from <none> to <" + newTarget + ">");
         } else {
-            final SystemJobManager.ScheduleResult scheduleResult = addSingleIndexRanges(oldTarget);
-            if (scheduleResult != null) {
-                try {
-                    LOG.debug("Waiting for index calculation to complete");
-                    scheduleResult.getFuture().get(deflectorIndexRangeCalculationTimeout.toMilliseconds(), TimeUnit.MILLISECONDS);
-                    LOG.debug("Completed!");
-                } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                    LOG.warn("Unable to calculate index range of index which is to be cycled: ", e);
-                }
-            } else {
-                LOG.warn("Unable to schedule job calculating index range of index which is to be cycled.");
-            }
-
             // Re-pointing from existing old index to the new one.
             LOG.debug("Now switching over deflector alias.");
             pointTo(newTarget, oldTarget);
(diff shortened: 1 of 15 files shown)


bernd:
Two comments after quickly looking at this. (no time to do a full review currently, sorry)

1. This is a breaking change to the SystemJobs API. So we have to discuss if we want to to this and if we do it, we have to adjust all affected plugins and document the upgrade path.
2. As far as I can see, there is no second index range calculation to ensure the "index 1" range is correct. Without this, we will also run into alert problems just like it is now. :smile:

joschi:
@dennisoelkers Would it solve to the problem to insert a fake index range for the current write-active index with `end` being now + 1 hour just before index rotation and keep the index range calculation as it is?

I share the fear of @bernd that without the second calculation (after the deflector alias has been pointed to the new index) the index range will be off by a few messages (and a few more messages in loaded ES clusters which take their time to actually return the min and max timestamps of an index).

dennisoelkers:
@bernd:
1.) Correct. That's why it's scheduled for 2.1.0., to keep the semver breakage minimal :)

@bernd / @joschi:
2.) I added the "ready-for-review" label a bit early. Nonetheless: We actually have the same issue at the moment too, because the index range is calculated 30 seconds before it is switched read-only, because we expect more messages to come in. I am still thinking about how to solve this issue, but it is also a separate one from the one causing #2264.

hc4:
Why not just set range for current index to infinity and schedule job for index recalculation right after index became readonly?
It will not break anything and will not change current behaviour (assuming, that current deflector may contain any timerange).
Also ES 2.3 a lot better (than 1.7) filters data by timestamp. So it shouldn't be a big performance drop.

hc4:
I think commit https://github.com/Graylog2/graylog2-server/pull/2278/commits/d523b74265e32a608bde922cb5fe8c228939bfcb breaks logic.
Just created index can receive messages from the past.
For example in case of long journal queue, or messages from some queueing engines (like AMQP of Beats)

dennisoelkers:
Thanks for the input, @hc4, I came to the same conclusion. Please note though, that the PR Is still WIP. When it is finished, it will be labeled accordingly and any comments are welcome.

dennisoelkers:
For the documentation: @bernd, I changed the method signatures back to their original form and added new ones for the `SystemJobManager` class, so it is backwards compatible.

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
```diff
@@ -156,29 +168,36 @@ public void cycle() {
...
             // it can happen that an index request still writes to the old deflector target, while we cycled it above.
             // setting the index to readOnly would result in ClusterBlockExceptions in the indexing request.
             // waiting 30 seconds to perform the background task should completely get rid of these errors.
             final SystemJob makeReadOnlyJob = indexReadOnlyJobFactory.create(oldTarget);
             try {
-                systemJobManager.submitWithDelay(makeReadOnlyJob, 30, TimeUnit.SECONDS);
-            } catch (SystemJobConcurrencyException e) {
+                final SystemJobManager.ScheduleResult scheduleResult = systemJobManager.submitWithDelayForResult(makeReadOnlyJob, 30, TimeUnit.SECONDS);
+                if (scheduleResult != null) {
+                    scheduleResult.getFuture().get(deflectorIndexReadOnlyTimeout.toMilliseconds(), TimeUnit.MILLISECONDS);
```
joschi: In the worst case this would block the thread for the duration of `deflectorIndexReadOnlyTimeout` (1 hour by default). What would happen if there is the need to do another cycle during that time?

graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java
```diff
@@ -26,17 +26,39 @@
...
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
 import static com.codahale.metrics.MetricRegistry.name;
 
 public class SystemJobManager {
+
+    public class ScheduleResult {
```
joschi: This should be a static nested class or no nested class at all (as it's also used by other classes). It would also be suitable to be converted to an Auto Value class.

graylog2-server/src/main/java/org/graylog2/indexer/ranges/IndexRangeService.java
```diff
@@ -31,4 +31,5 @@
     void save(IndexRange indexRange);
 
     IndexRange calculateRange(String index);
+    IndexRange createForDeflector(String index);
```
joschi: I think `createEmpty` would be a better name as it's not inherently bound to the deflector alias (it's not even being used in the method implementations).

Maybe adding it as a default method (https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html) would make sense here as its implementation doesn't rely on any internals of the specific implementation of `IndexRangeService`.

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
dennisoelkers: From what I have seen in the code, this would mean that another cycle would be done, which would be waiting at the same point.

graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java
dennisoelkers: Why is it necessary to make it static? It does not access any members of the enclosing class.

graylog2-server/src/main/java/org/graylog2/indexer/ranges/IndexRangeService.java
dennisoelkers: That's a good idea! :+1:

dennisoelkers: Turning out it's not as easy as it seems, as it currently relies on the choice of the value object in the service implementation. Using `MongoIndexRange` in a default implementation in the interface would smell broken to me. I will rename the method to a more generic name though.

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
bernd: I wouldn't feel good adding this "waiting on system job" logic in here.

Would it be enough to trigger an index range calculation job in the "make-read-only" job after it's done?

- Do not call `addSingleIndexRanges(oldTarget)` in line 160
- Modify the `SetIndexReadOnlyJob` to trigger a `CreateNewSingleIndexRangeJob` for the index once the read-only status is set

With this we wouldn't need the timeout logic and all the changes to the system jobs to return a future. We can also be sure that the index rage is correct because the index cannot be written anymore. Until the index is read-only, it will be included in the searches because it has the "deflector" index range.

graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java
joschi: ✅ You're right.

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
dennisoelkers: Coupling the job which sets indices read only to the index range calculation job sounds wrong to me, because they are not logically coupled and executing the first one should not need you to know (and be surprised about) that the second one is triggered implicitly too. The coupling should be done from the outside (as it is done here) via composition without any inherent coupling.

The logic you are describing is exactly what is happening in the PR, why it should be better to give up a separation of concerns to shift the waiting logic one level lower, while gaining nothing, is beyond me, tbh.

hc4: I think the problem is at line 182.
Why should we wait for job finished?
addListener could be used instead, to call addSingleIndexRanges after set_readonly request finished.

From [docs](http://netty.io/4.0/api/io/netty/channel/ChannelFuture.html):
```
 // BAD - NEVER DO THIS
 @Override
 public void channelRead(ChannelHandlerContext ctx, Object msg) {
     ChannelFuture future = ctx.channel().close();
     future.awaitUninterruptibly();
     // Perform post-closure operation
     // ...
 }

 // GOOD
 @Override
 public void channelRead(ChannelHandlerContext ctx, Object msg) {
     ChannelFuture future = ctx.channel().close();
     future.addListener(new ChannelFutureListener() {
         public void operationComplete(ChannelFuture future) {
             // Perform post-closure operation
             // ...
         }
     });
 }
```

dennisoelkers: We are waiting for the job to finish, so we can make sure that it is read only before calculating the index ranges. This is not java-specific, but related to the implemention of our system jobs. That's why we are now returning a Future to wait on.

hc4: Why not to make same logic like with StartOptimize?
The Job for optimize is submitted at the end of makeReadlonly job.
For example composite job could be created, which will consist of 3 steps:
1. MakeReadonly
2. Submit job for range calculation
3. Submit job for optimize

dennisoelkers: I still have not heard valid objections against the approach. But, there is one which caught noone's eyes yet: Changing it this way changes the `cycle()` method from being asynchronous to half-synchronous, which also bubbles up to the `DeflectorResource#cycle` method. This is not a problem per se, but means that restoring the original asynchronous behavior of `DeflectorResource#cycle` would involve another thread which is not an option for me.

hc4: The only problem I see - you will wait here at least for `deflectorIndexReadOnlyTimeout `and will block scheduler (single thread) for that time.
however, I'am not sure if it is real problem :)

joschi: That's why I've asked if the thread is blocked and what happened if a new deflector cycle was necessary while waiting for the last one. :wink:

In the end, this would in the worst case consume all (executor) threads in `SystemJobManager` and block all subsequent deflector cycles (until some threads are freed again).

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
```diff
@@ -194,7 +194,7 @@ public void cycle() {
     }
 
     private void addDeflectorIndexRange(String newTarget) {
-        final IndexRange deflectorRange = indexRangeService.createForDeflector(newTarget);
+        final IndexRange deflectorRange = indexRangeService.createEmptyRange(newTarget);
```
hc4: As I understand EmptyRange later treated as "Any time".
It looks a bit confusing to me :)

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
dennisoelkers: I agree, but from what I've seen this is already the case before this change. I am not sure if you are implying that there is a regression in this PR or that we should also change the current behavior (which I would perceive as scope creep).

hc4: `scheduleResult.getFuture().get(deflectorIndexReadOnlyTimeout.toMilliseconds(), TimeUnit.MILLISECONDS)` was introduced by this commit.
In matster there is no any waiting.

bernd: > Coupling the job which sets indices read only to the index range calculation job sounds wrong to me, because they are not logically coupled and executing the first one should not need you to know (and be surprised about) that the second one is triggered implicitly too. The coupling should be done from the outside (as it is done here) via composition without any inherent coupling.

I agree that the coupling introduced by this wouldn't be nice. We could introduce a separate job class that takes care of both. (i.e. `SetIndexReadOnlyAndCalculateRange`) That would make it clear what the job is supposed to do.
I also agree that it's nicer to be able to compose smaller jobs to avoid having lots of specialized jobs, but in this case I would vote for having one to avoid the introduction of timeout logic in this class and to avoid the async/sync issue you described later.

Another option would be to extend the system jobs subsystem to be able to run another job (or more) after one has finished. This would make it possible to compose a chain of system jobs without having to use timeouts or block the calling thread.

> The logic you are describing is exactly what is happening in the PR, why it should be better to give up a separation of concerns to shift the waiting logic one level lower, while gaining nothing, is beyond me, tbh.

I would happily give up a separation of concern when I can avoid using a timeout. :wink: The problem I see with using timeouts is that it's really hard to find a good default for them and that the user probably needs to tune them to make them work with their setup which makes using the product harder.

joschi:
LGTM. 👍

PR #2280, opened 2016-05-24
Fix plugin inter-dependencies by using one class loader for all plugins

bernd:
We used to create on URL class loader for each plugin. This broke
plugins which depend on classes from other plugins.

Now we are using one class loader for all plugins so they can see each
other. Plugins which need classes from other plugins are working now.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
@@ -21,7 +21,6 @@ import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
 import org.graylog2.plugin.Plugin;
 import org.graylog2.plugin.PluginMetaData;
 import org.graylog2.plugin.PluginModule;
@@ -32,6 +31,7 @@ import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -85,7 +85,7 @@ public class PluginLoader {
         }
 
         LOG.debug("Loading [{}] plugins", files.length);
-        final List<URL> urls = Lists.newArrayList(files).stream()
+        final List<URL> urls = Arrays.stream(files)
                 .filter(File::isFile)
                 .map(jar -> {
                     try {
@@ -105,7 +105,7 @@ public class PluginLoader {
 
         final ServiceLoader<Plugin> pluginServiceLoader = ServiceLoader.load(Plugin.class, classLoader);
 
-        return ImmutableSet.<Plugin>builder().addAll(pluginServiceLoader).build();
+        return ImmutableSet.copyOf(pluginServiceLoader);
     }
 
     public static class PluginComparator implements Comparator<Plugin> {
--- a/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import org.graylog2.plugin.Plugin;
 import org.graylog2.plugin.PluginMetaData;
 import org.graylog2.plugin.PluginModule;
@@ -34,9 +35,11 @@ import java.net.URLClassLoader;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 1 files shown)


joschi:
@bernd Wouldn't this also make `ChainingClassLoader` obsolete if all plugins (and Graylog itself) were using the same class loader?

graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
```diff
@@ -82,25 +85,27 @@ public PluginLoader(File pluginDir, ChainingClassLoader classLoader) {
...
-
-                    plugins.addAll(pluginServiceLoader);
-                } catch (MalformedURLException e) {
-                    LOG.error("Cannot open JAR file for discovering plugins", e);
-                }
-            }
-        }
-
-        return plugins.build();
+        final List<URL> urls = Lists.newArrayList(files).stream()
```
joschi: This could simply use [`Arrays.stream(T[] array)`](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#stream-T:A-) instead of instantiating a new array list and then calling `stream()` on it.

graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
```diff
@@ -82,25 +85,27 @@ public PluginLoader(File pluginDir, ChainingClassLoader classLoader) {
...
+                .filter(File::isFile)
```
joschi: We'll lose the debug logging that a file was being skipped this way, but I'm fine with it.

graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
```diff
@@ -82,25 +85,27 @@ public PluginLoader(File pluginDir, ChainingClassLoader classLoader) {
...
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+
+        // Create one URL class loader for all plugins so they can see each other.
+        // This makes plugin inter-dependencies work.
+        classLoader.addClassLoader(URLClassLoader.newInstance(urls.toArray(new URL[urls.size()])));
+
+        final ServiceLoader<Plugin> pluginServiceLoader = ServiceLoader.load(Plugin.class, classLoader);
+
+        return ImmutableSet.<Plugin>builder().addAll(pluginServiceLoader).build();
```
joschi: This could simply be written as `return ImmutableSet.copyOf(pluginServiceLoader);`.

bernd:
> @bernd Wouldn't this also make ChainingClassLoader obsolete if all plugins (and Graylog itself) were using the same class loader?

Right now there are still two class loaders. Not sure if there are any other implications if we use only one class loader for server **and** plugins. /cc @kroepke

kroepke:
I guess technically we don't need the second classloader and could instead set the parent (i.e. Graylog's) in the URLClassloader as the chaining classloader really delegates to the parent first anyway.

However, if we don't need to touch all of that now I'd prefer not to. But I guess it's simple to test whether it makes a different. I'm unsure what the consequences for any plugin resources are. So far I haven't seen any conflicts, but that doesn't mean there aren't any somewhere.

joschi:
Works for me™.

magicdude4eva:
This is problematic. The `graylog-plugin-map-widget-1.0.3.jar` ships with a version of httpclient which is missing for example URIBuilder:
```
[root@gls1 plugin]$ jar -tvf graylog-plugin-map-widget-1.0.3.jar | grep -i "http/client/utils"
     0 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/
  1957 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/JdkIdn.class
   938 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/Punycode.class
   177 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/Idn.class
  1913 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/CloneUtils.class
  3017 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/Rfc3492Idn.class
  4526 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/URIUtils.class
  5214 Mon Jun 20 15:27:24 SAST 2016 org/apache/http/client/utils/URLEncodedUtils.class
```

My plugin can now not ship with httpclient as it conflicts with some classes in the map-plugin, but can also not ship without httpclient as `org/apache/http/client/utils/URIBuilder` is missing from the map plugin.

I can now try to "hack" together my plugin to include classes which are missing from the map-plugin as I can not include httpclient or I can ask users to completely remove the map-plugin from the plugins folder and then ship my plugin as before with httpclient and httpmime.

In my case I am using jira-client which uses URIBuilder which is pulled from my bundled version of httpclient, but then references 
```
java.lang.NoSuchMethodError: org.apache.http.client.utils.URLEncodedUtils.encPath(Ljava/lang/String;Ljava/nio/charset/Charset;)Ljava/lang/String;
	at org.apache.http.client.utils.URIBuilder.encodePath(URIBuilder.java:191) ~[graylog-plugin-jira-1.0.7.jar:?]
	at org.apache.http.client.utils.URIBuilder.buildString(URIBuilder.java:152) ~[graylog-plugin-jira-1.0.7.jar:?]
	at org.apache.http.client.utils.URIBuilder.build(URIBuilder.java:120) ~[graylog-plugin-jira-1.0.7.jar:?]
	at net.rcarz.jiraclient.RestClient.buildURI(RestClient.java:117) ~[graylog-plugin-jira-1.0.7.jar:?]

```

Alternatively, could the map-widget at least include versions of httpclient which are more up-to-date than 4.0.1?

kroepke:
@magicdude4eva We can update it, but for various reasons we need the plugins to share a classloader.
Unfortunately, without going an OSGi route (shudder), there's not going to be an easy solution.

Can you shade/relocate that specific dependency in your plugin? That should provide a workaround.

magicdude4eva:
I can try shade/relocate and hope that with the various dependencies https://github.com/rcarz/jira-client will still work (which I need for my plugin https://github.com/magicdude4eva/graylog-jira-alarmcallback).

Just to confirm: Was the classloader change for plugins introduced with 2.0.3 (I am currently running 2.0.3 and can not recall to having the issue before).

I think the big problem with one classloader is that any plugin could break anyone else and then it becomes plugin authors responsibility to try and fix a workaround.

If anyone can find a clever workaround for my problem, feel free to assist with the pom (I am honestly struggling with this as I am not very familiar with maven builds - https://github.com/magicdude4eva/graylog-jira-alarmcallback).

For now I will just ask users to delete the map-plugin as I have no time to try and find a work-around for this in the next week or two.

magicdude4eva:
Thank you @kroepke. One more question before I try another work-around:

When looking at the server log it shows that my "JIRA" plugin is loaded first:

```
2016-06-29T15:19:14.530+02:00 INFO  [CmdLineTool] Loaded plugin: JIRA integration plugin 1.0.7 [com.bidorbuy.graylog.alarmcallbacks.jira.JiraAlarmCallback]
2016-06-29T15:19:14.532+02:00 INFO  [CmdLineTool] Loaded plugin: Collector 1.0.3 [org.graylog.plugins.collector.CollectorPlugin]
2016-06-29T15:19:14.532+02:00 INFO  [CmdLineTool] Loaded plugin: Enterprise Integration Plugin 1.0.3 [org.graylog.plugins.enterprise_integration.EnterpriseIntegrationPlugin]
2016-06-29T15:19:14.532+02:00 INFO  [CmdLineTool] Loaded plugin: Pipeline Processor Plugin 1.0.0-beta.5 [org.graylog.plugins.pipelineprocessor.ProcessorPlugin]
2016-06-29T15:19:14.533+02:00 INFO  [CmdLineTool] Loaded plugin: Anonymous Usage Statistics 2.0.3 [org.graylog.plugins.usagestatistics.UsageStatsPlugin]

```

Would my assumption be correct, that since I bundle a newer version of httpclient, that if I just do a "dummy" instantiation of HTTPClient classes in a static initialiser block of my AlarmCallBack-class that this would then trigger the loading of my httpclient classes and then breaking the map-plugin?

kroepke:
The load order is pretty much undefined, so I wouldn't rely on it.

Whether or not the map plugin breaks depends on the binary compatibility of the different versions, that's tricky to say. 
What I've done in the past to work around issues like these was to include the library's code as a git submodule (from a fork) and then update the library to use a specific version.
That's messy and requires tracking changes, but is often the only sane way to deal with jar hell.

magicdude4eva:
I tried for the last 4 hours to work around this issue, but there is no good way solving it without hacking the build and it will be just a matter of time that another user has similar issues. 

For me personally it is okay to delete the map-plugin as I don't use it, but I think other users might have similar issues when moving to Graylog 2.0.2/2.0.3.

kroepke:
That is true, if we can find a better way that's not too intrusive (and find the time to make it work) my hope is to have a solution in time for 2.1.

PR #2282, opened 2016-05-25
Alert condition titles

dennisoelkers:
This PR allows users to add titles to alert conditions and makes them available for alarm callbacks.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/streams/Stream.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/streams/Stream.java
@@ -50,8 +50,6 @@ public interface Stream extends Persisted {
 
     String getContentPack();
 
-    Collection<AlertCondition> getAlertConditions();
-
     void setTitle(String title);
 
     void setDescription(String description);
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
@@ -133,8 +133,8 @@ public class StreamResource extends RestResource {
 
         final Map<String, String> result = ImmutableMap.of("stream_id", id);
         final URI streamUri = getUriBuilderToSelf().path(StreamResource.class)
-                .path("{streamId}")
-                .build(id);
+            .path("{streamId}")
+            .build(id);
 
         return Response.created(streamUri).entity(result).build();
     }
@@ -178,11 +178,11 @@ public class StreamResource extends RestResource {
     @ApiOperation(value = "Get a single stream")
     @Produces(MediaType.APPLICATION_JSON)
     @ApiResponses(value = {
-            @ApiResponse(code = 404, message = "Stream not found."),
-            @ApiResponse(code = 400, message = "Invalid ObjectId.")
+        @ApiResponse(code = 404, message = "Stream not found."),
+        @ApiResponse(code = 400, message = "Invalid ObjectId.")
     })
     public StreamResponse get(@ApiParam(name = "streamId", required = true)
-                      @PathParam("streamId") @NotEmpty String streamId) throws NotFoundException {
+                              @PathParam("streamId") @NotEmpty String streamId) throws NotFoundException {
         checkPermission(RestPermissions.STREAMS_READ, streamId);
 
         return streamToResponse(streamService.load(streamId));
@@ -195,13 +195,13 @@ public class StreamResource extends RestResource {
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     @ApiResponses(value = {
-            @ApiResponse(code = 404, message = "Stream not found."),
-            @ApiResponse(code = 400, message = "Invalid ObjectId.")
+        @ApiResponse(code = 404, message = "Stream not found."),
+        @ApiResponse(code = 400, message = "Invalid ObjectId.")
     })
     public StreamResponse update(@ApiParam(name = "streamId", required = true)
-                         @PathParam("streamId") String streamId,
-                         @ApiParam(name = "JSON body", required = true)
-                         @Valid @NotNull UpdateStreamRequest cr) throws NotFoundException, ValidationException {
+                                 @PathParam("streamId") String streamId,
+                                 @ApiParam(name = "JSON body", required = true)
+                                 @Valid @NotNull UpdateStreamRequest cr) throws NotFoundException, ValidationException {
         checkPermission(RestPermissions.STREAMS_EDIT, streamId);
         final Stream stream = streamService.load(streamId);
 
@@ -218,7 +218,7 @@ public class StreamResource extends RestResource {
                 stream.setMatchingType(Stream.MatchingType.valueOf(cr.matchingType()));
             } catch (IllegalArgumentException e) {
                 throw new BadRequestException("Invalid matching type '" + cr.matchingType()
-                        + "' specified. Should be one of: " + Arrays.toString(Stream.MatchingType.values()));
+                    + "' specified. Should be one of: " + Arrays.toString(Stream.MatchingType.values()));
             }
         }
 
@@ -232,8 +232,8 @@ public class StreamResource extends RestResource {
     @Timed
     @ApiOperation(value = "Delete a stream")
     @ApiResponses(value = {
-            @ApiResponse(code = 404, message = "Stream not found."),
-            @ApiResponse(code = 400, message = "Invalid ObjectId.")
+        @ApiResponse(code = 404, message = "Stream not found."),
+        @ApiResponse(code = 400, message = "Invalid ObjectId.")
     })
     public void delete(@ApiParam(name = "streamId", required = true) @PathParam("streamId") String streamId) throws NotFoundException {
         checkPermission(RestPermissions.STREAMS_EDIT, streamId);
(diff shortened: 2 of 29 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/streams/Stream.java
```diff
@@ -50,8 +50,6 @@ public static MatchingType valueOfOrDefault(String name) {
 
     String getContentPack();
 
-    Collection<AlertCondition> getAlertConditions();
```
joschi: This would break existing plugins that are working with Graylog 2.0.x and which are using this method. Any reason to remove it and not simply mark it as deprecated (to remove it in Graylog 3.x)?

dennisoelkers: The method, as it was before, was plain broken. It worked only because of type erasure, as it was actually returning a `Collection<BasicDBObject>`. It was only used implicitly by Jackson when serializing a `Stream` instance as JSON, so it was also using a different representation that any other place that was serializing an `AlertCondition` instance and it was only identical by accident.

It was also a duplicate of `StreamService#getAlertConditions` which goes the hard path of instantiating an actual `AlertCondition`. Unfortunately we cannot just move that logic to the Stream implementation class (or pass the list of correct alert conditions to its constructor) as the `AlertCondition` instance itself [references the `Stream`](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/plugin/alarms/AlertCondition.java#L35) which holds it.

It's not the cleanroom technique to just remove it for a minor version, but it's actually balancing between api breakage and offering a method that is just plain broken.

joschi: ✅ Works for me. At least it'll be obvious for plugin authors and users when stuff breaks.

But please add this to the `UPGRADING.rst` file.

dennisoelkers: Check!

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -397,7 +397,7 @@ private StreamResponse streamToResponse(Stream stream) {
             stream.getFields().get(StreamImpl.FIELD_CREATED_AT).toString(),
             stream.getDisabled(),
             stream.getStreamRules(),
-            stream.getAlertConditions(),
+            alertConditions,
```
edmundoa: Not a big thing, but this may affect some code consuming the API: Alert conditions returned in here seem to have the type written in upper case letters, whereas before they used lower case. Alert conditions returned by `StreamAlertConditionResource` also use lower case for the type: https://github.com/Graylog2/graylog2-server/pull/2282/files#diff-51de2a31470f13c3ee6f29bc5129b989R157

I think the safest way to avoid any troubles is to simply convert the types in here to lower case, even if clients may already do so.

graylog2-web-interface/src/components/alertconditions/AlertCondition.jsx
```diff
@@ -28,15 +28,25 @@ const AlertCondition = React.createClass({
...
+    request.type = this.props.alertCondition.type;
     AlertConditionsActions.update.triggerPromise(this.props.alertCondition.stream_id, this.props.alertCondition.id, request)
     .then(() => {
       this.setState({edit: false});
     });
   },
+  _formatTitle() {
+    const alertCondition = this.props.alertCondition;
+    const alertConditionType = new AlertConditionsFactory().get(alertCondition.type);
+    const title = alertCondition.title ? alertCondition.title : `Untitled ${alertConditionType.title} condition`;
```
edmundoa: I would simply leave out the title if not present, I don't see the need of adding that "Untitled" prefix. I think specially at the beginning will look really weird, as none of the conditions will have title, and all will contain the untitled prefix.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Well } from 'react-bootstrap';
+import { Input, Well } from 'react-bootstrap';
```
edmundoa: Unused import

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -10,20 +10,37 @@ const AlertConditionForm = React.createClass({
...
 
     if (typeDefinition !== undefined) {
-      return <typeDefinition.configuration_form ref="conditionForm" alertCondition={this.props.alertCondition}/>;
+      return <typeDefinition.configuration_form ref="conditionForm" alertCondition={this.props.alertCondition.parameters}/>;
     }
 
     return undefined;
   },
   render() {
+    const alertCondition = this.props.alertCondition || {parameters: {}};
```
edmundoa: The default object is missing some spaces :nitpick:

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -10,20 +10,37 @@ const AlertConditionForm = React.createClass({
...
+    return {
+      title: this.refs.title.value,
+      parameters: this.refs.conditionForm.getValue(),
+    };
   },
+  getDefaultProps() {
+    return {
+      alertCondition: {
+        parameters: {},
+      }
```
edmundoa: Missing trailing comma.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -10,20 +10,37 @@ const AlertConditionForm = React.createClass({
...
 
     return undefined;
   },
   render() {
+    const alertCondition = this.props.alertCondition || {parameters: {}};
     return (
       <Well className="alert-type-form alert-type-form-message-count form-inline well-sm">
+        Title: <input ref="title" type="text" className="form-control" autoComplete="off" defaultValue={alertCondition.title}/>
+        {' '}
+        <small>This title can also be included in alert notifications.</small>
```
edmundoa: I find this text a little confusing inside this form, as we mostly describe there how the condition should apply. I think adding some colour would help with that, I'll commit that so you can take a look.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -37,8 +37,9 @@ const AlertConditionForm = React.createClass({
     return (
       <Well className="alert-type-form alert-type-form-message-count form-inline well-sm">
         Title: <input ref="title" type="text" className="form-control" autoComplete="off" defaultValue={alertCondition.title}/>
-        {' '}
-        <small>This title can also be included in alert notifications.</small>
+        <span style={{ color: '#939393', marginLeft: 10 }}>
```
dennisoelkers: I am kinda opposed to adding even more inline styling. We should at least put it in some locally used stylesheet file.

graylog2-web-interface/src/components/alertconditions/AlertCondition.jsx
dennisoelkers: I think it makes sense to add it, because it hints users migrating to this version that they can now give titles to their alert conditions.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
dennisoelkers: It is a big thing, actually. Thanks for spotting it!

dennisoelkers: ✅

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
edmundoa: I only do that sometimes when the styling is only used in one place. Anyway, you are right, I'll put it into some file 👍

graylog2-web-interface/src/components/alertconditions/AlertCondition.jsx
edmundoa: Then I suggest using the same styling as newly added titles. I think it achieves the purpose you mentioned, and people can see the difference between the title and the type of condition.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -8,26 +8,44 @@ const AlertConditionForm = React.createClass({
     type: React.PropTypes.string.isRequired,
     alertCondition: React.PropTypes.object,
   },
-  alertConditionsFactory: new AlertConditionsFactory(),
+  getDefaultProps() {
+    return {
+      alertCondition: {
+        parameters: {},
```
edmundoa: I just realised that in 2.0, creating a new alert condition provides default values for parameters, and after these changes we don't do that any more. To fix that, we should remove the empty object from here, as each specific alert condition form provides some default props if the parameters are not given.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
```diff
@@ -8,26 +8,44 @@ const AlertConditionForm = React.createClass({
...
 
     if (typeDefinition !== undefined) {
-      return <typeDefinition.configuration_form ref="conditionForm" alertCondition={this.props.alertCondition}/>;
+      return <typeDefinition.configuration_form ref="conditionForm" alertCondition={this.props.alertCondition.parameters}/>;
     }
 
     return undefined;
   },
   render() {
+    const alertCondition = this.props.alertCondition || { parameters: {} };
```
edmundoa: At the moment this default alert condition will never be used, as the `getDefaultProps` method will already give us a default alert condition.

graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx
edmundoa: ✅

edmundoa:
LGTM 👍

PR #2289, opened 2016-05-26
Improve search histogram resolution auto selection

### LINKED ISSUE #2148: Search in all messages graphs use minute resolutions
### Problem description
When searching in all messages, both the search histogram, and field histograms use minute as default resolution. As that search query may ask for a long time of data, we should avoid doing that by default, and use a lower resolution instead.

If I remember correctly, we were using day as resolution before, but apparently that doesn't work any more.

1. Search in all messages
2. See that the search histogram has a minute resolution
3. Create a field graph. It also uses minute resolution

### Environment
* Graylog Version: 2.0.0

edmundoa:
This PR improves the way we auto calculate the histogram resolution on searches over all messages, by:

- Calculate the search duration from the oldest index we have (before we ordered the indices the other way around)
- Take into account that sometimes the deflector index may not have proper index ranges, and they default to 1 Jan 1970. In that case we try to get the search duration from the second oldest index (which would be, in fact the oldest one). In case there aren't any more indices available, we assume the search duration is immediate, so we will effectively use the "minute" resolution on the histogram.

This also applies to field charts, so please also take a little time looking that they work as expected.

The changes should go into master as well.

Fixes #2148

### DIFF
--- a/graylog2-web-interface/src/pages/SearchPage.jsx
+++ b/graylog2-web-interface/src/pages/SearchPage.jsx
@@ -92,16 +92,31 @@ const SearchPage = React.createClass({
     const inputs = InputsStore.inputsAsMap(state.inputs);
     this.setState({ inputs: Immutable.Map(inputs) });
   },
-  _determineHistogramResolution(response) {
-    let queryRangeInMinutes;
+  _determineSearchDuration(response) {
+    const searchTo = response.to;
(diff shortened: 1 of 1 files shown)


kroepke:
lgtm, also tested with just one (deflector) index, works as expected. new field graphs use the same resolution

PR #2291, opened 2016-05-26
Add Graylog cluster overview

edmundoa:
This PR adds a Graylog cluster overview component to the System -> Overview page. It also takes care of unifying some small things among all components being used in that page.

The changes should also be merged into master.


graylog2-web-interface/src/stores/nodes/NodesStore.js
```diff
@@ -43,6 +43,14 @@ const NodesStore = Reflux.createStore({
     return this.nodes[nodeId];
   },
 
+  getClusterId() {
+    return Object.keys(this.nodes).map(id => this.nodes[id]).map(node => node.cluster_id)[0].toUpperCase();
```
dennisoelkers: Why not doing something like `.find(clusterId => clusterId)` instead of `[0]`, so we protect against possible undefined/null cluster ids?

graylog2-web-interface/src/components/cluster/GraylogClusterOverview.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+  getInitialState() {
+    return {
+      clusterId: undefined,
+      nodeCount: 0,
+    };
+  },
+
+  _onNodesChange() {
+    this.setState({ clusterId: NodesStore.getClusterId(), nodeCount: NodesStore.getNodeCount() });
```
dennisoelkers: Why are the cluster id and the node count not included in the triggered state of the store, so we can just `connect()` to the store? We would also save those getters in the `NodesStore` which are kinda against Reflux principles (at least if they're avoidable).

graylog2-web-interface/src/components/cluster/index.js
```diff
@@ -0,0 +1 @@
+export GraylogClusterOverview from './GraylogClusterOverview';
```
dennisoelkers: This way of reexporting the default export of an ES6 module is not working in some ES6 implementations. Therefore we changed this at some point to follow [a form which works in all implementations](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/components/common/index.jsx).

graylog2-web-interface/src/components/indexers/IndexerFailuresComponent.jsx
```diff
@@ -50,21 +50,24 @@ const IndexerFailuresComponent = React.createClass({
...
-      return <Spinner />;
+      content = <Spinner />;
+    } else {
+      content = this._formatFailuresSummary();
     }
+
     return (
       <Row className="content">
         <Col md={12}>
-          <h2><i className="fa fa-truck" /> Indexer Failures</h2>
```
dennisoelkers: Why have the glyphicons been removed for each section?

graylog2-web-interface/src/components/cluster/index.js
edmundoa: I can't remember that we decided anything about that. Which ES6 implementations don't support that? Anyway, I think we still have some of them around, basically I copied this one from `graylog2-web-interface/src/components/times/index.jsx`.

dennisoelkers: Babel 6 is the one that matters the most for us. I am remembering we talked about this briefly when I changed most of the exports to the format.

graylog2-web-interface/src/components/indexers/IndexerFailuresComponent.jsx
edmundoa: We decided long ago to remove all icons from headers (basically in the UI changes we started in 1.1), as in the end they are mostly random and don't provide any value. The system overview page is the only place that still got some of them, and they weren't even used consistently, so some sections had them, and some not.

Now that I was adding a new section into the page, I decided it was time to get rid of them.

graylog2-web-interface/src/stores/nodes/NodesStore.js
edmundoa: ✅

graylog2-web-interface/src/components/indexers/IndexerFailuresComponent.jsx
dennisoelkers: ✅

graylog2-web-interface/src/components/cluster/index.js
edmundoa: I remember IntelliJ didn't like some of them (was highlighting them in red), but don't remember about that. Anyway, I will update this one on the PR and change others I can find in master. In that way we don't copy them around like in this case 👍

graylog2-web-interface/src/components/cluster/GraylogClusterOverview.jsx
edmundoa: ✅

edmundoa: Don't know why I didn't do that from the beginning. Using `listenTo` was actually problematic, as it could try to get some value from the store before the state was initialised.

PR #2298, opened 2016-05-27
Check if props actually contain configuration fields before copying them

### LINKED ISSUE #2297: Using "Clear value" functionality in input type dropdown breaks input creation
### Problem description

After using the "Clear value"-functionality (symbolized as a small x in the right part of the dropdown) of the input type dropdown, the dropdown does not work anymore. Selecting an input type fails, also opening the configuration modal.

1. Go to system/inputs page
2. Select an input type from inputs dropdown
3. Press the x-button ("Clear value") to the right of the input type inside the dropdown
4. Try to select an input type again

### Environment

* Graylog Version: 2.0.2
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
This helps against the dropdown breaking because of an unhandled
exception due to a deselected input type and the following nonpresence
of configuration fields in the on change handler.

Should also be merged into master.

Fixes #2297


edmundoa:
LGTM 👍

PR #2304, opened 2016-05-30
Upgrade to Jackson 2.7.4

joschi:
Previous versions of MongoJack were prone to generating a `StackOverflowError` with Jackson 2.6.3 and higher.

Those problems were fixed in MongoJack 2.6.1 (see mongojack/mongojack#114), so that we finally can upgrade to an newer Jackson version.


joschi:
@dennisoelkers Rebase to current `master`.

PR #2305, opened 2016-05-30
Disabling a configured proxy for requests to localhost/127.0.0.1/::1.

dennisoelkers:
This PR adds a bit of code to disable the proxy in the provider if the target URI containcs `localhost`, `127.0.0.1` or `::1`, because we assume that these are meant to go against the Graylog server host itself and not the proxy host.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
@@ -80,7 +80,7 @@ public class OkHttpClientProvider implements Provider<OkHttpClient> {
                 public List<Proxy> select(URI uri) {
                     switch (uri.getHost().toLowerCase(Locale.ENGLISH)) {
                         case "localhost":
-                        case "127.0.1":
+                        case "127.0.0.1":
                         case "::1":
                             return ImmutableList.of(Proxy.NO_PROXY);
                         default:
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
@@ -17,16 +17,24 @@
 package org.graylog2.shared.bindings.providers;
 
 import com.github.joschi.jadconfig.util.Duration;
+import com.google.common.collect.ImmutableList;
 import okhttp3.OkHttpClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
+import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Proxy;
+import java.net.ProxySelector;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
```diff
@@ -66,7 +75,26 @@ public OkHttpClient get() {
 
         if (httpProxyUri != null) {
             final Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(httpProxyUri.getHost(), httpProxyUri.getPort()));
-            clientBuilder.proxy(proxy);
+            final ProxySelector proxySelector = new ProxySelector() {
+                @Override
+                public List<Proxy> select(URI uri) {
+                    switch (uri.getHost().toLowerCase(Locale.ENGLISH)) {
+                        case "localhost":
+                        case "127.0.1":
```
joschi: Typo?

graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
```diff
@@ -66,7 +75,26 @@ public OkHttpClient get() {
...
+                        case "::1":
```
joschi: This will miss non-normalized IPv6 addresses, e. g. "0:0:00::1".

graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
```diff
@@ -66,7 +75,26 @@ public OkHttpClient get() {
 
         if (httpProxyUri != null) {
             final Proxy proxy = new Proxy(Proxy.Type.HTTP, new InetSocketAddress(httpProxyUri.getHost(), httpProxyUri.getPort()));
-            clientBuilder.proxy(proxy);
+            final ProxySelector proxySelector = new ProxySelector() {
+                @Override
+                public List<Proxy> select(URI uri) {
+                    switch (uri.getHost().toLowerCase(Locale.ENGLISH)) {
```
joschi: Maybe use `InetAddress` here, e. g. https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#isAnyLocalAddress-- and/or https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#isLoopbackAddress--

dennisoelkers: Good idea. Probably solves all of the remarks :)

dennisoelkers: Unfortunately this would rely on `InetAddress`'s usual name resolving, which rules out this solution.

joschi: Why is this a problem?

dennisoelkers: Hm, due to the fact that we're doing the resolving soon after acquiring the http client provider anyway, it might not be a problem to do this.

joschi:
LGTM. 👍

Issue #2132, opened 2016-04-24
SelfSignedCertificate should migrate from sun.security.*


mikkolehtisalo:
Some newest javacs seems to fail to compile SelfSignedCertificate without applying workarounds. Probably should migrate to java.security.cert, or some 3rd party library.

Tested on master, and openJDK 1.8.0_91:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project graylog2-server: Compilation failure: Compilation failure:
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[21,25] sun.security.x509.AlgorithmId is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[22,25] sun.security.x509.CertificateAlgorithmId is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[23,25] sun.security.x509.CertificateIssuerName is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[24,25] sun.security.x509.CertificateSerialNumber is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[25,25] sun.security.x509.CertificateSubjectName is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[26,25] sun.security.x509.CertificateValidity is internal proprietary API and may be removed in a future release
[ERROR] /home/mlehtisalo/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java:[27,25] sun.security.x509.CertificateVersion is internal proprietary API and may be removed in a future release
[...]

PR #2316, opened 2016-05-31
Migrate from sun.security.* to Bouncycastle

mikkolehtisalo:
Fixes issue #2132 

`sun.security.*` will probably not be available by default on Java 9, and Oracle does not seem to be going to provide replacement for all functionality. `Java.security.cert` is not able to perform certificate creation and signing. It is meant just for parsing already existing certificates.

Quick research shows that most projects depending on `sun.security.x509.*` have switched, or are in progress of switching, to Bouncycastle. Here's a quick implementation of that. I don't think it has to be really fancy, because self signed certificates should be used only for light testing.

Btw, running [jdeps](https://wiki.openjdk.java.net/display/JDK8/Java+Dependency+Analysis+Tool) on Graylog shows that these were the only JDK internal classes used, excluding 3rd party libraries.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
@@ -16,23 +16,33 @@
  */
 package org.graylog2.shared.security.tls;
 
+import org.apache.commons.io.IOUtils;
+import org.bouncycastle.asn1.ASN1InputStream;
+import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.asn1.x509.BasicConstraints;
+import org.bouncycastle.asn1.x509.Extension;
+import org.bouncycastle.asn1.x509.SubjectKeyIdentifier;
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.cert.bc.BcX509ExtensionUtils;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.crypto.params.AsymmetricKeyParameter;
+import org.bouncycastle.crypto.util.PrivateKeyFactory;
+import org.bouncycastle.jce.X509KeyUsage;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.DefaultDigestAlgorithmIdentifierFinder;
+import org.bouncycastle.operator.DefaultSignatureAlgorithmIdentifierFinder;
+import org.bouncycastle.operator.bc.BcRSAContentSignerBuilder;
+import org.bouncycastle.cert.X509v3CertificateBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import sun.security.x509.AlgorithmId;
-import sun.security.x509.CertificateAlgorithmId;
-import sun.security.x509.CertificateIssuerName;
-import sun.security.x509.CertificateSerialNumber;
-import sun.security.x509.CertificateSubjectName;
-import sun.security.x509.CertificateValidity;
-import sun.security.x509.CertificateVersion;
-import sun.security.x509.CertificateX509Key;
-import sun.security.x509.X500Name;
-import sun.security.x509.X509CertImpl;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
```diff
@@ -16,29 +16,28 @@
  */
 package org.graylog2.shared.security.tls;
 
+import org.apache.commons.io.IOUtils;
+import org.bouncycastle.asn1.ASN1InputStream;
+import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.asn1.x509.*;
```
joschi: Please don't use wildcard imports.

If you're using IntelliJ IDEA, here are instructions how to (almost) disable wildcard imports: http://stackoverflow.com/a/3348855

graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
```diff
@@ -16,29 +16,28 @@
...
 import java.io.IOException;
 import java.math.BigInteger;
-import java.security.GeneralSecurityException;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
-import java.security.KeyStore;
-import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
-import java.security.SecureRandom;
+import java.security.*;
```
joschi: Same here, please don't use wildcard imports.

graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
```diff
@@ -57,6 +56,8 @@
  * </p>
  */
 public final class SelfSignedCertificate {
+
+
```
joschi: Useless whitespace change.

graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
```diff
@@ -165,38 +166,50 @@ private static KeyPair generateKeyPair(SecureRandom random, int bits) {
...
-
-        // Prepare the information required for generating an X.509 certificate.
-        final X509CertInfo info = new X509CertInfo();
-        final X500Name owner = new X500Name("CN=" + fqdn);
-
-        info.set(X509CertInfo.VERSION, new CertificateVersion(CertificateVersion.V3));
-        info.set(X509CertInfo.SERIAL_NUMBER, new CertificateSerialNumber(new BigInteger(64, random)));
+    private static SubjectKeyIdentifier createSubjectKeyIdentifier(Key publicKey) throws IOException {
+        ASN1InputStream is = null;
         try {
```
joschi: I think we can use try-with-resources here to avoid the `finally` block: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

joschi:
LGTM except for those minor nitpicks. 👍

graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
mikkolehtisalo: Yeah, IDEA wanted to use wildcards. Sorry about that.

graylog2-server/src/main/java/org/graylog2/shared/security/tls/SelfSignedCertificate.java
mikkolehtisalo: Ooh shiny.

joschi:
Bouncycastle's take on U. S. export restrictions: [What is Bouncy Castle's export classification in the United States of America?](https://www.bouncycastle.org/wiki/display/JA1/Frequently+Asked+Questions#FrequentlyAskedQuestions-11.WhatisBouncyCastle'sexportclassificationintheUnitedStatesofAmerica?)

> At the time of writing (16 May 2007) Bouncy Castle is approved classified under ECCN code 5D002 and approved for export under License Exception TSU. If you also need to list algorithms available in the provider and the strengths supported, you can find the information in the specifications.html file provided with the distribution you are using. See [The Bureau of Industry and Security website](http://www.bis.doc.gov/licensing/exportingbasics.htm) for further details.

joschi:
@mikkolehtisalo Thanks!

PR #2320, opened 2016-06-01
Support changes for pipeline processor simulator

edmundoa:
These are required changes in common components being used in the first version of the message pipeline processor simulator, in summary:

- Add prop to `LoaderTabs` to control whether the loaded message should be display or not
- Add prop to `MessageDetail` to disable all links and buttons when showing message simulations
- Changes in `MessageShow` to ensure the props it receives are instances of immutable.js objects, so other components using it don't need to care about it
- Extract transformation of messages from the REST API into the internal object we use to it's own function, making it available in other places of the code

Changes required for https://github.com/Graylog2/graylog-plugin-pipeline-processor/pull/34

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageShow.jsx
+++ b/graylog2-web-interface/src/components/search/MessageShow.jsx
@@ -1,13 +1,30 @@
-import React, {PropTypes} from 'react';
+import React from 'react';
+import Immutable from 'immutable';
 import MessageDetail from './MessageDetail';
 
 const MessageShow = React.createClass({
   propTypes: {
-    message: PropTypes.object,
-    inputs: PropTypes.object,
-    streams: PropTypes.object,
-    nodes: PropTypes.object,
+    message: React.PropTypes.object,
+    inputs: React.PropTypes.object,
+    streams: React.PropTypes.object,
+    nodes: React.PropTypes.object,
   },
+
+  getInitialState() {
+    return this._getImmutableProps(this.props);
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/components/search/MessageShow.jsx
```diff
@@ -1,25 +1,42 @@
...
+  },
+
   possiblyHighlight(fieldName) {
     // No highlighting for the message details view.
-    return this.props.message.fields[fieldName];
+    return String(this.props.message.fields[fieldName]);
   },
   render() {
     return (
       <div className="row content">
```
dennisoelkers: Could you replace this with a `Row`-component?

graylog2-web-interface/src/components/search/MessageShow.jsx
```diff
@@ -1,25 +1,42 @@
...
         <div className="col-md-12">
```
dennisoelkers: Could you replace this with a `Col`-component?

graylog2-web-interface/src/components/search/MessageShow.jsx
edmundoa: ✅

graylog2-web-interface/src/components/search/MessageShow.jsx
edmundoa: ✅

PR #2347, opened 2016-06-12
Add caching to reverse dns queries

### LINKED ISSUE #999: Slow syslog processing speed with force_rdns enabled
There seems to be a problem when the `force_rdns` option for Syslog inputs is enabled. Users report high CPU usage and reduced message throughput.

See: https://groups.google.com/forum/?hl=en#!topic/graylog2/bW2glCdBIUI

mikkolehtisalo:
Fixes #999 

It seems Guava offered tools for this. Expiration, and maximum size handled. Should improve performance without having to install local dns caching daemon. Less surprises for administrators.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Tools.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Tools.java
@@ -16,6 +16,9 @@
  */
 package org.graylog2.plugin;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


dennisoelkers:
Thank you for submitting this PR. The problem I see with it, is that it is replicating functionality of the local resolver in a way, that it's inferior to a full-fledged caching resolver, because it does not respect the returned caching conditions of the dns response (refresh/retry/expiration times) but uses a fixed caching interval instead, which might lead to unexpected/unwanted results. I would still advise to install a local dns caching daemon because that is what it's intended job is and it's doing it way better than we could do.

PR #2353, opened 2016-06-14
Do not init available alarm callback types, fetch them explicitly.

### LINKED ISSUE #2352: User in group/role with Edit granted unable to manage stream alerts
### Problem description
A user in an Active Directory group mapped to a role granting Edit to a stream is unable to manage alerts on that stream, and receives an error message like:

```
Could not load user list
Loading user list failed with status: Error: cannot GET https://graylog.mysite.com:12345/users (403)
```

with the REST API returning HTTP 403:
```
{"type":"ApiError","message":"User [username123] not authorized. (GET users)"}
```

We would expect a user with Edit role granted to a stream to be able to manage alerts on that stream.

1. Have a LDAP server with a user John as a member of group "Team Foo Editors"
2. Set up a Graylog 2.0.2 environment and connect it to the LDAP server
3. Create a stream named Bar
4. Create a role named "Team Foo - Editors" and map the LDAP group "Team Foo Editors" to it
5. Grant the "Team Foo - Editors" role Read+Edit to the Bar stream
6. Log in as user John and navigate to Streams ("/streams")
7. Note that you can see the Bar stream, and can define rules on that stream
8. Open your browser's developer tools to monitor network requests
9. Click "Manage Alerts" and note that you are redirected back to the streams list and the following error appears:
```
Could not load user list
Loading user list failed with status: Error: cannot GET https://graylog.mysite.com:12345/users (403)
```
10. Note the following requests occurred:

```
GET /streams/STREAM_ID (returns HTTP 200 with stream JSON)
GET /streams/STREAM_ID/alerts/conditions (returns HTTP 200 with more JSON),
GET /users (returns HTTP 403 with JSON {"type":"ApiError","message":"User [John] not authorized. (GET users)"})
GET /streams/STREAM_ID/alarmcallbacks (returns HTTP 200 with JSON)
GET /streams/STREAM_ID/alarmcallbacks/available (returns HTTP 200 with JSON)
GET /streams/STREAM_ID/alerts/paginated?skip=0&limit=10 (returns HTTP 200 with JSON)
GET /system/gettingstarted (returns HTTP 200 with JSON)
GET /streams (returns HTTP 200 with JSON)
...
```

### Environment

* Graylog Version:  2.0.2 (4da1379)
* Elasticsearch Version:  2.2.0
* MongoDB Version:  2.6.12
* Operating System:  Red Hat Enterprise Linux Server release 6.7
* Browser version:  Firefox 47

dennisoelkers:
When we are fetching the available alarm callback types during init of
AlarmCallbacksStore, we don't have a stream id to fetch for yet.
Therefore we are running into permission problems in the backend
resource for non-admin users. This change replaces the init with an
explicit fetch of available alarm callback types in the consumer.

Fixes #2352

### DIFF
--- a/graylog2-web-interface/src/stores/alarmcallbacks/AlarmCallbacksStore.js
+++ b/graylog2-web-interface/src/stores/alarmcallbacks/AlarmCallbacksStore.js
@@ -12,13 +12,6 @@ const AlarmCallbacksStore = Reflux.createStore({
   listenables: [AlarmCallbacksActions],
   types: undefined,
 
-  init() {
-    this.available(undefined).then((types) => {
-      this.types = types;
-      this.trigger({types: types});
-    });
-  },
-
   getInitialState() {
     return {
       types: this.types,
--- a/graylog2-web-interface/src/stores/alarmcallbacks/AlarmCallbacksStore.js
+++ b/graylog2-web-interface/src/stores/alarmcallbacks/AlarmCallbacksStore.js
@@ -10,13 +10,6 @@ import fetch from 'logic/rest/FetchProvider';
 
 const AlarmCallbacksStore = Reflux.createStore({
   listenables: [AlarmCallbacksActions],
-  types: undefined,
-
-  getInitialState() {
-    return {
-      types: this.types,
-    };
-  },
 
   available(streamId) {
     const failCallback = (error) =>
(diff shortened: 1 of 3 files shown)


dennisoelkers:
Could/should be merged in 2.0.3 if possible.

graylog2-web-interface/src/stores/alarmcallbacks/AlarmCallbacksStore.js
```diff
@@ -12,24 +12,16 @@ const AlarmCallbacksStore = Reflux.createStore({
...
-  init() {
-    this.available(undefined).then((types) => {
-      this.types = types;
-      this.trigger({types: types});
-    });
-  },
-
   getInitialState() {
     return {
       types: this.types,
```
edmundoa: With the changes the store state is not used any more, at least I couldn't find any usages.

edmundoa:
LGTM 👍

PR #2354, opened 2016-06-14
Move custom analyzer into index template

joschi:
The custom `analyzer_keyword` analyzer should be part of the index template so that newly created indices automatically know about it.

This PR doesn't change the behavior of Elasticsearch or Graylog but simply moves the index settings into one place for better code coherence.


edmundoa:
LGTM 👍

PR #2355, opened 2016-06-15
Remove automatic private key/certificate generation

joschi:
This PR removes the automatic generation of a private key/certificate pair if TLS (HTTPS) support was enabled but the private key and/or the certificate are missing or aren't readable.

Unfortunately that feature is broken beyond repair:
* The generated private keys and certificates are using a key size of 1024 bits (RSA), which is ridiculous nowadays (2048 bits is minimum, 4096 bits is recommended).
* The generated private keys and certificates are using the host names from `rest_listen_uri`/`web_listen_uri` which breaks down if those aren't the exposed names (e. g. if `rest_transport_uri` or `web_endpoint_uri` is being used).
* The internal communication between Graylog nodes via HTTPS through the Graylog REST API breaks down, because the generated certificates aren't trusted and the HTTP client refuses to speak to untrusted servers.

And most important: People relied on those generated weak certificates for their production systems.

I think it's best to remove this mis-feature altogether and force people to at least think about which level of security they want.


bernd:
Please document the upgrade path for users that currently use this feature in `UPGRADING.rst`. Thanks!

bernd:
LGTM :+1:

bernd:
@joschi Please rebase and push again, thanks. I just wanted to press the merge button and it turned grey. :smile:

Issue #2151, opened 2016-04-27
Pause/Start Stream button has no loading indicator


lennartkoopmann:
On a slower connection, when starting or stopping a stream in the stream overview list, it is a bit confusing because there is no loading indicator. I'd expect some kind of color change or loading indicator between the toggle and the state change.

bernd:
This got fixed in #2357

dennisoelkers:
It actually got fixed only partially. What we're tracking in #2357 is only the REST API request being sent off successfully, not the state change. Therefore I am reopening the issue.

PR #2357, opened 2016-06-15
Improved feedback

edmundoa:
This PR adds feedback to buttons triggering actions that may take some time to receive a response from the server. In that way, the user can see the action is being processed and doesn't have to wonder if the click worked or not.

In order to use [`finally`](http://bluebirdjs.com/docs/api/finally.html) in our promises to reset the loading status, I had to set bluebird as the default promise factory in Reflux. I wrote a bit more about it in  	1ae78c0's commit message.


edmundoa:
I just saw that this fixes #2151

Issue #2360, opened 2016-06-16
incorrect node information and api browser link incorrect


jalogisch:
### desription

if you have a chained Proxy Setup for HTTPS (like in the image) 

![903_design](https://cloud.githubusercontent.com/assets/404238/16112374/5ec6d050-33b7-11e6-9b55-2126d935c8e3.png)

- you did not get node based data in System / Nodes (all Nodes show the same data)
- the API Browser link is displayed wrong.

Not sure if this is related to #2359 and/or #1576

### configuration

Apache Loadbalancer Setup

```
<VirtualHost *:443>
    ServerName H1

    SSLEngine On
    SSLProxyEngine On
    SSLCertificateFile conf.d/tls/server.pem
    SSLCertificateChainFile conf.d/tls/server.pem

    <Proxy balancer://web>
        BalancerMember "https://H2"
        BalancerMember "https://H3"
    </Proxy>
    ProxyPass        / balancer://web/
    ProxyPassReverse / balancer://web/
</VirtualHost>

```

Apache Setup on Graylog System
```
<VirtualHost *:443>
    ServerName H2
    DefaultType None
    SSLEngine On
    SSLCertificateFile conf.d/tls/server.pem
    SSLCertificateChainFile conf.d/tls/server.pem
   <Location /api/>
        ProxyPass http://localhost:12900/
        ProxyPassReverse http://localhost:12900/
    </Location>
    <Location />
        RequestHeader set X-Graylog-Server-URL "https://H1/api/"
        ProxyPass http://localhost:9000/
        ProxyPassReverse http://localhost:9000/
    </Location>
</VirtualHost>
```

Graylog Server Conf for Web and API

```
rest_listen_uri = http://127.0.0.1:12900/
web_enable = true
web_listen_uri = http://127.0.0.1:9000/
```

### Environment
* Graylog Version: 2.0.X

joschi:
@jalogisch How (and on which page exactly) is the link to the API Browser wrong?

jalogisch:
if you request the page via **H1 LB** the Link (when you use the API Browser button) is http://127.0.0.1:12900/api-browser

joschi:
@jalogisch `rest_transport_uri` must point to the accessible (!) address of each Graylog node. Otherwise the proxying node will always request the node information from `rest_listen_uri` (which points to localhost in this specific case).

jalogisch:
[...]

PR #2361, opened 2016-06-16
Fix link to API Browser on Node pages

joschi:
Refs #2360

### DIFF
--- a/graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
+++ b/graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
@@ -29,10 +29,6 @@ const NodeMaintenanceDropdown = React.createClass({
               <MenuItem>Configure internal logging</MenuItem>
             </LinkContainer>
           </IfPermitted>
-
-          <MenuItem href={`${this.props.node.transport_address}api-browser`} target="_blank">
-            API Browser <i className="fa fa-external-link"/>
-          </MenuItem>
         </DropdownButton>
       </ButtonGroup>
     );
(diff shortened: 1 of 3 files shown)


joschi:
This PR should also be ported into the `master` branch.

graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
```diff
@@ -30,7 +31,7 @@ const NodeMaintenanceDropdown = React.createClass({
             </LinkContainer>
           </IfPermitted>
 
-          <MenuItem href={`${this.props.node.transport_address}api-browser`} target="_blank">
+          <MenuItem href={URLUtils.qualifyUrl('api-browser')} target="_blank">
```
dennisoelkers: The link to the API browser is node-specific, as it is visually associated with a node element in the nodes list/part of the node details page of a specific node. The way it is changed here, makes it point to the API browser of the API endpoint node of the web interface, which might not be the same and is therefore counterintuitive. We should therefore disconnect the API browser link from the node element in the UI if it is not possible to reliably generate a link to the node's API browser.

graylog2-web-interface/src/components/nodes/NodesActions.jsx
```diff
@@ -50,7 +49,7 @@ const NodesActions = React.createClass({
           <Button bsStyle="info">Metrics</Button>
         </LinkContainer>
 
-        <Button bsStyle="info" href={apiBrowserURI} target="_blank">
+        <Button bsStyle="info" href={URLUtils.qualifyUrl('api-browser')} target="_blank">
```
dennisoelkers: Same as above.

PR #2362, opened 2016-06-16
Fix base URI for API documentation

joschi:
Refs #2360


joschi:
This PR should also be ported into the `master` branch.

Issue #2358, opened 2016-06-16
Has Permissions changed in Graylog 2.0.2


123dev:
### Problem description
Previously with Graylog 1.3.x we needed more granularity into a role, that was not directly possible to achieve from the ui.
Let me explain
We needed a developer role whose members could see the Search feature without necessarily having to be on a stream (search menu that is available for admins) along with few other permissions.

So what we did was create the dev role from the UI, and then update the permissions using the API

Our dev role was set to this
```
{
  "name": "Dev",
  "description": "Developers Role",
  "permissions": [
     "dashboards:create",
     "dashboards:read",
     "sources:read",
     "streams:create",
     "streams:read",
     "messages:analyze",
     "searches:absolute",
     "searches:keyword",
     "searches:relative"
  ],
  "read_only": false
}

```
And this was working fine in 1.3.x
After the upgrade to 2.0.2, this is no longer working, even though when I check the user's permissions that is a member of Dev role, (using the API) I get to see the above permissions.

Checking the [RestPermissions.java](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/security/RestPermissions.java) I see those permissions are still valid.

Did something change that I need to adjust? or is there a regression in the way permissions work?

To make matters worse, I tried to create a test role from the UI to see if there was any change in schema.
I tried with Chrome and Firefox and no matter what the Save button is not clickable
Basically can't add roles
Not sure if this is local to my environment or a possible regression.

Dev console of the browser gives the following warnings (not sure if they are relevant)
```
Warning: Failed propType: Invalid prop `items` of type `List` supplied to `TableList`, expected instance of `List`. Check the render method of `PermissionSelector`.

Warning: Failed propType: Invalid prop `data` of type `object` supplied to `TypeAheadDataFilter`, expected `array`. Check the render method of `TableList`.
```

#### Problem 1
- Create a Dev role using API as described above
- add a member to that role
- test it in 1.3.x, the user has access to Search / Sources ... (whatever permissions allow)
- Repeat the same in 2.0.2
- Search is not available

#### Problem two
- Try to add a new role using the WEB UI of 2.0.2
- Fill in a name, description and select at least one permission
- Can't save (not clickable)

### Environment
AWS Image
* Graylog Version: 2.0.2

Thanks

edmundoa:
Hi @123dev,

[...]

PR #2366, opened 2016-06-16
Fix permission checks for non admin users

edmundoa:
Some components were using wrong permission values in checks, so non-admin users could never get to do those actions.

Original issue: #2358

The changes should also be merged into 2.0, I already checked they apply cleanly.


dennisoelkers:
Could you please replace the `{this.isPermitted(...) &&` pattern with usages of the `<IfPermitted />` component? Makes the code much more readable imo.

edmundoa:
Sure thing, I was looking so much into writing the right permissions that I didn't even look at that 👍

edmundoa:
I have replaced most of the `isPermitted` in touched files with `IfPermitted`. I found some issues doing so in the navigation bar dropdown, as `IfPermitted` is swallowing additional props that react-bootstrap is giving to it.

We could clone the children elements given to `IfPermitted` to pass the props down, but I don't feel confident on doing so if the changes are going to 2.0.3, as we don't know if they affect other components or have any performance impact. I'm willing to make that change in master afterwards, as we can see the impact in a longer term. How does that sound?

PR #2367, opened 2016-06-16
Pluggable auth

### LINKED ISSUE #2232: Make auth providers fully pluggable
Currently Graylog does not support pluggable authentication providers.
In order to make it easier to integrate with SSO systems or other external authentication/authorization providers, the hardcoded list of built-in providers should be made extensible, configurable from the API/UI and non-core providers should move into plugins themselves.

This includes the following tasks:

- [x] Introduce new binding helper for plugging in `AuthorizingRealm`/`AuthenticatingRealm`
- [x] Allow ordering realms from configuration and during runtime
- [x] Refine interfaces to create shadow user accounts based on external authentication realms
- [x] Define interface to configure the `SecurityManager`

kroepke:
This change refactors the way the authentication subsystem works:

* Rather than hardcoding `AuthenticationRealm` implementations, they are contributed by an orderable, injected configuration.
* The UI for users, roles and authentication is unified and refactored to allow plugins to contribute new implementations.
* LDAP UI is now separate from the users list.
* a first implementation of an authentication plugin can be found at https://github.com/Graylog2/graylog-plugin-auth-sso
* to be able to use request context information a new token implementation exposes the request HTTP headers to authenticators if they declare to use them (again see sso plugin for an example)
* validating a session during `SessionStore` initialization can now return a new session and a new user to allow "automatic" login, which is used by the SSO plugin

fixes #2232


garybot2:

garybot2:

graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
```diff
@@ -21,22 +21,39 @@
...
+
 @AutoValue
 @JsonAutoDetect
 public abstract class SessionValidationResponse {
     @JsonProperty("is_valid")
     public abstract boolean isValid();
 
+    @JsonProperty("new_session_id")
+    @Nullable
+    public abstract String newSessionId();
```
joschi: Why "new_session_id" and not just "session_id"?

graylog2-server/src/main/java/org/graylog2/security/AuthenticationConfig.java
```diff
@@ -0,0 +1,99 @@
...
+                        PasswordAuthenticator.NAME,
+                        RootAccountRealm.NAME))
+                .disabledRealms(Collections.emptySet())
+                .build();
+    }
+
+    public AuthenticationConfig withRealms(final Set<String> availableRealms) {
+        final List<String> newOrder = new ArrayList<>();
+
+        // Check if processor actually exists.
```
joschi: "processor" vs. "realm" in comments

joschi:
When trying to access the System &rarr; Authentication page (http://localhost:9000/system/authentication) after running `mvn clean install` (which processes the JavaScript files), I get a blank page (except for the menu bar) and the following error in my Developer Console:
```
Uncaught Error: Requested actions "Authentication" is not registered.
plugin.org.graylog.plugins.pipelineprocessor.PipelineProcessorPlugin.c3f2b6e….js:14207 
getActions @ plugin.org.graylog.plugins.pipelineprocessor.PipelineProcessorPlugin.c3f2b6e….js:14207
(anonymous function) @ AuthenticationComponent.jsx:47
(anonymous function) @ AuthenticationComponent.jsx:225
827 @ 16.16.dcc96b5….js:3
n @ app.dcc96b5….js:1
(anonymous function) @ AuthenticationPage.jsx?./~/react-hot-loader!./~/babel-loader:37
(anonymous function) @ AuthenticationPage.jsx?./~/react-hot-loader!./~/babel-loader:93
1011 @ 16.16.dcc96b5….js:3
n @ app.dcc96b5….js:1
(anonymous function) @ AuthenticationPage.jsx:7
window.webpackJsonp @ app.dcc96b5….js:1
(anonymous function) @ 16.16.dcc96b5….js:1
```

graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
```diff
@@ -39,17 +41,21 @@
...
     }
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
         final boolean secure = requestContext.getSecurityContext().isSecure();
         final MultivaluedMap<String, String> headers = requestContext.getHeaders();
-        final String host = headers.getFirst(HttpHeaders.HOST);
+        final Request grizzlyRequest = grizzlyRequestProvider.get();
+
+        final String host = grizzlyRequest.getRemoteAddr();
```
joschi: This won't work with reverse proxies in front of Graylog. We should at least support the "standard" headers like `X-Forwarded-For` (also see `RestTools#getRemoteAddrFromRequest()`).

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String host;
+
+    public HttpHeadersToken(MultivaluedMap<String, String> httpHeaders, String host) {
+        this.httpHeaders = httpHeaders;
+        this.host = host;
+    }
+
+    @Override
+    public Object getPrincipal() {
+        return httpHeaders.get("remote_user");
```
joschi: Is this a header name chosen by us? If so, I'd recommend using `X-Graylog-Remote-User` instead.

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public Object getCredentials() {
+        return httpHeaders.get("remote_credentials");
```
joschi: Is this a header name chosen by us? If so, I'd recommend using `X-Graylog-Remote-Credentials` instead.

joschi:
https://github.com/Graylog2/graylog2-server/pull/2367#issuecomment-233352967 has been fixed by rebuilding all related projects (I hate computers…).

Thanks @edmundoa for helping to figure this out!

graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
kroepke: The intention was that this field contains a different session id than the one which had been passed in. But I guess session_id is enough 👍

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
kroepke: see below.

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
kroepke: These are really only return values because the interface requires it.
The intention for a HttpHeadersToken is that a plugin (such as the SSO one) can pick any header(s) it likes to get the relevant information. The only user of this token is not even using `getPrincipal`/`getCredentials`.
However, I think this should be clearer, so I'll add documentation comments to that effect.
I think `remote_user` is a good choice as it is often used, `remote_credentials` is probably obscure. I'll check if we can simply return null here safely.

graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/SessionValidationResponse.java
joschi: "really_new_new_session_id2_bak" :trollface:

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
joschi: Isn't "REMOTE_USER" usually just a CGI environment variable rather than an HTTP request header?

kroepke: I guess this is largely a question of convention. We can also leave it as `null`.

graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
kroepke: Aye!

graylog2-server/src/main/java/org/graylog2/shared/security/HttpHeadersToken.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * A HttpHeadersToken does not have a natural principal associated with it, so this is always null.
+     *
+     * @return null
+     */
+    @Override
+    @Nullable
+    public Object getPrincipal() {
+        return null;
```
joschi: Would throwing an `UnsupportedOperationException` work here or are we afraid that stuff™ will break because of it?

kroepke: I'd simply return null, not sure if this is used somewhere in the framework's code.

joschi: ✅

joschi:
LGTM. 👍

PR #2378, opened 2016-06-17
Make Message#getStreamIds() more reliable

joschi:
The `Message#getStreamIds()` method only checked for the existence of the "streams" message field and returned it. This only works for messages which have already been indexed and have that field.

For messages in-flight, the method always returned an empty list, even if `StreamMatcherFilter` had run before.

This commit extends `Message#getStreamIds()` to also query the internal "streams" list which is being populated by the `StreamMatcherFilter` if the "streams" message field doesn't exist.


joschi:
This should also be ported into the `master` branch.

PR #2385, opened 2016-06-20
Update search query on auto refresh

### LINKED ISSUE #2379: Clicking Search nav link when on Search page does not reset search state
### Problem description

Clicking the Search navigation link when on the Search page does not reset the search state.  The query from the previous search continues to be used.

We would expect clicking the Search navigation link to reset the search state and trigger a new search or, put another way, work the same as clicking the link from a different page (e.g. Streams).

1. Log in to Graylog as an admin
2. If needed, create and enable an input (e.g. `Random HTTP message generator`) to ensure messages are in the system
3. Click the "Search" link in the top navigation bar
4. Enter `source:invalid_source_yay` (or anything that will return no results) as the query and perform the search
5. Note that, as expected, "Nothing found" is shown
6. Note that the URL is something like `https://mygraylog:8443/search?rangetype=relative&fields=message%2Csource&width=1920&highlightMessage=&relative=300&q=source%3Ainvalid_source_yay`
7. Click the "Search" link in the top navigation bar again
8. Note the query input is cleared and the URL is now something like `https://mygraylog:8443/search`, however "Nothing found" is still displayed
9. Click the "Play" button to trigger automatic updates
10. Note this makes no difference, and "Nothing found" is still displayed
11. Click the "Streams" link in the top navigation bar, and then click the "Search" link
12. Note messages received recently are displayed as expected

### Environment

* Graylog Version:  2.0.2 (4da1379)
* Elasticsearch Version:  2.2.0
* MongoDB Version:  2.6.12
* Operating System:  Red Hat Enterprise Linux Server release 6.7
* Browser version:  Firefox 47

edmundoa:
When a search did not return results, clicking on the "Search" navigation menu, does not reset the search query used for auto-refresh.

Removing the query from the `SearchPage` state fixes the problem, as the search query is properly updated in the `SearchStore`.

Fixes #2379.

These changes should also be merged into master.

### DIFF
--- a/graylog2-web-interface/src/pages/SearchPage.jsx
+++ b/graylog2-web-interface/src/pages/SearchPage.jsx
@@ -36,7 +36,6 @@ const SearchPage = React.createClass({
   getInitialState() {
     return {
       selectedFields: ['message', 'source'],
-      query: SearchStore.query.length > 0 ? SearchStore.query : '*',
       error: undefined,
     };
   },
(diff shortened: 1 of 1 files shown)


dennisoelkers:
This does not fix #2379 though, as it does not retrigger the search request, so when entering a search query and clicking on "Search" in the top navbar, the message list stays the same.

edmundoa:
@dennisoelkers you are right, I misread the original issue, and the fix is not actually fixing what it should. Will update the PR or open a new one if the changes are completely unrelated. Sorry about all the confusion!

PR #2388, opened 2016-06-20
longer retention interval for journal tests

kroepke:
to prevent a cleaner thread race condition on slower build machines, increase the background retention duration to something much longer than 1 second


joschi:
Seems to work. :+1:

PR #2394, opened 2016-06-21
Remove "elasticsearch_discovery_zen_ping_multicast_enabled" setting

joschi:
ES multicast discovery has been moved into a plugin in Elasticsearch 2.0.0, been deprecated in Elasticsearch 2.2.0, and will be removed in Elasticsearch 5.0.0.

Additionally, the plugin (https://www.elastic.co/guide/en/elasticsearch/plugins/2.3/discovery-multicast.html) cannot be used with Graylog, so the `elasticsearch_discovery_zen_ping_multicast_enabled` setting is useless and should be removed.


bernd:
@joschi The multicast settings need to be removed from the Chef cookbook (and maybe other projects) as well.

bernd:
Otherwise LGTM

PR #2395, opened 2016-06-21
Fix time zone names in time zone dropdown

### LINKED ISSUE #2393: user preferred timezone not saved
### Problem description
if you create a user and set the timezone from UTC to something else and update the user the message "user has updated" was shown in green bar, and it is displayed on the page. 

But after reload or leaving the side it is switched back to UTC

1. create user
2. set timezone to something other than etc
3. "update user"
4. reload the page

### Environment

* Graylog Version: 2.0.3 (OVA)

edmundoa:
Some time zones may not have an area, and we were putting those into the `Etc` area, which is incorrect and failed when trying to get that time zone back in the server.

Fixes #2393 

The changes should also be merged into master.

### DIFF
--- a/graylog2-web-interface/src/components/common/TimezoneSelect.jsx
+++ b/graylog2-web-interface/src/components/common/TimezoneSelect.jsx
@@ -12,11 +12,16 @@ const TimezoneSelect = React.createClass({
   getValue() {
     return this.refs.timezone.getValue();
   },
+
+  // Some time zones are not stored into any areas, this is the group we use to put them apart in the dropdown
+  // https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+  _UNCLASSIFIED_AREA: 'Unclassified',
+
   _formatTimezones() {
     const timezones = {};
     moment.tz.names().forEach((timezone) => {
       const splitted = timezone.split('/');
-      const area = (splitted.length > 1 ? splitted[0] : 'Etc');
+      const area = (splitted.length > 1 ? splitted[0] : this._UNCLASSIFIED_AREA);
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/common/TimezoneSelect.jsx
```diff
@@ -12,25 +12,33 @@ const TimezoneSelect = React.createClass({
...
     return this.refs.timezone.getValue();
   },
+
+  // Some time zones are not stored into any areas, this is the group we use to put them apart in the dropdown
+  // https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+  _UNCLASSIFIED_AREA: 'Unclassified',
+
   _formatTimezones() {
     const timezones = {};
     moment.tz.names().forEach((timezone) => {
```
joschi: Just a remark: Wouldn't it be better to export [the list of supported time zones in Joda-Time](http://joda-time.sourceforge.net/timezones.html) once and use that instead of what [Moment.js](http://momentjs.com/) returns?

edmundoa: @joschi I think that would only move the problem to another place: when we use the time zone to convert UTC times into local times, as that is mostly done in the frontend nowadays. So, in the end, moment.js and Joda-Time need to support the same time zones, or we need to have a layer between them that does some conversion for us. So far I didn't find any example that breaks one of the libraries, so I assume they do use the same names currently.

joschi: ✅

joschi:
LGTM. 👍

Issue #1929, opened 2016-03-15
2.0a4 'Edit input' - form reloads itself while editing


SjonHortensius:
I'm trying to edit a global input on */system/inputs*, but while editing the form reloads the previously saved fields every few seconds. This is very annoying.

Possibly related; I see these notices in my console:

```
unreachable code after return statement plugin.org.graylog.plugins.map-widget.MapWidget.ddd103b678992ca86b75.js:24142:5
mutating the [[Prototype]] of an object will cause your code to run very slowly; instead create the object with the correct initial [[Prototype]] value using Object.create plugin.org.graylog.plugins.map-widget.MapWidget.ddd103b678992ca86b75.js:9906:416
unreachable code after return statement plugin.org.graylog.plugins.map-widget.MapWidget.ddd103b678992ca86b75.js:24142:5
unreachable code after return statement app.3ae8b0a28a3bb071fe6d.js:4773:4
unreachable code after return statement
```

edmundoa:
Hi SjonHortensius,

Which browser are you using?

SjonHortensius:
This is **Firefox 45.0**

dennisoelkers:
This is a duplicate of #1870, which is closed already. Expect this to be fixed in the next release.

Issue #1870, opened 2016-02-29
Username connection to AMQP broker lost settings


fbalicchia:
### Problem description
When I edit GELF AMQP and change connection userName when I try to
change the password  in the Username textbox appears
the old username. I've see this problem in graylog-2.0.0-alpha.5 webapp.
### Steps to reproduce the problem
1. Graylog env upand running
2. Create an GELF AMQP input (with userName and password)
3. Edit input
4. Change username and click on password text box
5. Username change is lost

### Environment

* Graylog Version: 2.0.0-alpha.5-SNAPSHOT
* Elasticsearch Version: 2.2.0
* MongoDB Version: 2.6.11
* Operating System: Ubuntu 14.10

PR #2399, opened 2016-06-22
Fix unrequested refresh of configuration forms/Reset configuration forms on cancel.

dennisoelkers:
This PR fixes two issues with the configuration forms:

 * When they are refreshing their props (either because the form is reinitialized or because a plain re-render is triggered by a store refresh) the internal state is [reinitialized](https://github.com/Graylog2/graylog2-server/blob/b24d09d31227a7454d5eee115089bc7b878bc443/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx#L19) with a mixture of the default values and the passed props, but not the already existing state which included changes made by the user. The outcome is that configuration forms which are part of a DOM subtree which is refreshed by a store refresh, periodically snap back into their default values.

* The configuration form's internal state is set back to the default values when the form is closed. This prevents that when e.g. a new input form is opened and some changes are made, closing and reopening it will still show the old values.

Both of these are also proper fixes for #1870 and #1929, which were fixed by preventing their periodical refresh at that time.

### DIFF
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
@@ -18,12 +18,12 @@ class BootstrapModalForm extends Component {
     this.open = this.open.bind(this);
     this.close = this.close.bind(this);
     this._submit = this._submit.bind(this);
-    this._onModalClose = this._onModalClose.bind(this);
+    this._onModalCancel = this._onModalCancel.bind(this);
   }
 
-  _onModalClose() {
-    if (typeof this.props.onModalClose === 'function') {
-      this.props.onModalClose();
+  _onModalCancel() {
+    if (typeof this.props.onCancel === 'function') {
+      this.props.onCancel();
     }
 
     this.close();
@@ -76,7 +76,7 @@ class BootstrapModalForm extends Component {
             {body}
           </Modal.Body>
           <Modal.Footer>
-            <Button type="button" onClick={this._onModalClose}>{this.props.cancelButtonText}</Button>
+            <Button type="button" onClick={this._onModalCancel}>{this.props.cancelButtonText}</Button>
             <Button type="submit" bsStyle="primary">{this.props.submitButtonText}</Button>
           </Modal.Footer>
         </form>
@@ -96,6 +96,7 @@ BootstrapModalForm.propTypes = {
   onModalOpen: PropTypes.func,
   onModalClose: PropTypes.func,
   onSubmitForm: PropTypes.func,
+  onCancel: PropTypes.func,
   /* Object with additional props to pass to the form */
   formProps: PropTypes.object,
   /* Text to use in the cancel button. "Cancel" is the default */
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
@@ -67,7 +67,10 @@ class BootstrapModalForm extends Component {
     );
 
     return (
-      <BootstrapModalWrapper ref="modal" onOpen={this.props.onModalOpen} onClose={this.props.onModalClose}>
+      <BootstrapModalWrapper ref="modal"
+                             onOpen={this.props.onModalOpen}
+                             onClose={this.props.onModalClose}
+                             onCancel={this._onModalCancel}>
         <Modal.Header closeButton>
           <Modal.Title>{this.props.title}</Modal.Title>
         </Modal.Header>
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
@@ -70,7 +70,7 @@ class BootstrapModalForm extends Component {
       <BootstrapModalWrapper ref="modal"
                              onOpen={this.props.onModalOpen}
                              onClose={this.props.onModalClose}
-                             onCancel={this._onModalCancel}>
+                             onHide={this._onModalCancel}>
         <Modal.Header closeButton>
           <Modal.Title>{this.props.title}</Modal.Title>
         </Modal.Header>
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
@@ -1,6 +1,6 @@
-import React, {Component, PropTypes} from 'react';
+import React, { Component, PropTypes } from 'react';
 import ReactDOM from 'react-dom';
-import {Modal, Button} from 'react-bootstrap';
+import { Modal, Button } from 'react-bootstrap';
 import $ from 'jquery';
 
 import BootstrapModalWrapper from 'components/bootstrap/BootstrapModalWrapper';
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalWrapper.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalWrapper.jsx
@@ -33,7 +33,6 @@ class BootstrapModalWrapper extends Component {
     if (typeof this.props.onCancel === 'function') {
       this.props.onCancel();
     }
-    this.onClose();
   }
 
   open() {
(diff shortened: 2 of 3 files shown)


graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
```diff
@@ -18,12 +18,12 @@ class BootstrapModalForm extends Component {
...
     this._submit = this._submit.bind(this);
-    this._onModalClose = this._onModalClose.bind(this);
+    this._onModalCancel = this._onModalCancel.bind(this);
   }
 
-  _onModalClose() {
-    if (typeof this.props.onModalClose === 'function') {
-      this.props.onModalClose();
+  _onModalCancel() {
+    if (typeof this.props.onCancel === 'function') {
```
edmundoa: This change will break some of the other components using `BootstrapModalForm`, as before they relied on the `onModalClose` callback being called when the modal was cancelled or closed, and now it won't be called all the time.

Anyway, why do we need to differentiate between both situations? The way I see it, closing the modal and clicking cancel is the same. Most of the time I also feel that the software I use behaves in that way too (closing a modal is usually triggering the same action as cancelling). That doesn't apply to "reset" buttons, but usually you can't find those in modals.

edmundoa: Another point on this topic: what would you expect to happen when you are editing the modal and press "Escape"?

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -5,18 +5,33 @@ import BootstrapModalForm from 'components/bootstrap/BootstrapModalForm';
...
+    children: React.PropTypes.node,
+    helpBlock: React.PropTypes.node,
+    includeTitleField: React.PropTypes.bool,
+    submitAction: React.PropTypes.func.isRequired,
+    title: React.PropTypes.string,
+    titleValue: React.PropTypes.string,
+    typeName: React.PropTypes.string,
+    values: React.PropTypes.object,
+  },
+
```
edmundoa: ❤️

graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
dennisoelkers: `onModalClose` is [still called](https://github.com/Graylog2/graylog2-server/blob/9d237670ae0d90ecae1ab35b0f9180fc7b1f9a8d/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx#L70). We need to differentiate those two cases, because resetting the form is not required after submitting it, while it is necessary when cancelling it.

edmundoa: I still don't see why we need to differentiate both cases. I think submitting the form also needs to reset the inputs in some cases (after creating an entity, for instance). I also think that some people will be mad because they pressed cancel instead of the "x" or escape and their inputs are gone, specially in forms using textareas like an alert callback email template or a sidecar snippet. At least I would be mad if I needed to re-enter the input for that reason. At the moment we are fairly inconsistent with that (as with many other things), but I see this as a step backwards in usability.

Regarding `onModalClose` not being called, I meant on external components using `BootstrapModalForm`. See [this example](https://github.com/Graylog2/graylog2-server/blob/c8e26cb9602bc4ceb95c4b77469dabb8f71b3f09/graylog2-web-interface/src/components/configurations/MessageProcessorsConfig.jsx#L157-L157) for instance: The `MesageProcessorConfig` expects the `onModalClose` callback to be called on close or cancel, but now it will only be called on close, changing the previous behaviour of that component.

dennisoelkers: I think anything else than resetting the form when explicitly dismissing it does not make sense, and I have yet to see an app where explicitly closing a form does not reset it.

In this scenario `onModalClose` is called through the explicit call of `close()` on `BootstrapModalWrapper`.

edmundoa: There's one example that came into my mind as I read your comment: editing emails. Few apps nowadays will delete the text you wrote even if you close the window or the application. In most situations in our app, the user just need to fill in a name and description, and it's not a big issue, but in others it may need to add passwords, or even big chunks of text.

This change is specially tricky, because we are doing different things when closing or cancelling the submit, which is in my opinion a technicality, and will lead to error and frustration.

I'll also add a quote from Jef Raskin, author of The Humane Interface:

> The system should treat all user input as sacred.

Regarding the the `onModalCancel`, I missed that, so ✅

dennisoelkers: The MTU case implies that there is some drafting mechanism, which we do not have (and is probably overkill for our use case). Those applications which are much closer to our use case are all following the semantics implemented by this PR, which means that when the user is cancelling (the action of doing this is btw a user input as well) the "draft" is discarded.

Another argument for changing the behavior we have today: when editing an existing input, making changes in one of the fields, cancelling the form and editing the entity again shows the changes that were just done and cancelled. This is imo unexpected for a majority of the users and leads to an inconsistency between the persisted state of the entity, the state that is shown in the overview and the state as shown in the edit form.

edmundoa: I don't really see it the same way, and also don't consider that a user making the modal go away, is implying that their changes are gone. We are imprecise by nature, and will find the easiest way to make that go away, even if you just want to look at some detail or copy something from somewhere else.

Either way, I don't think we can convince each other on this, and I don't want to discuss this forever. As long as the changes work consistently, we can go with them.

graylog2-web-interface/src/components/bootstrap/BootstrapModalWrapper.jsx
```diff
@@ -53,6 +65,7 @@ BootstrapModalWrapper.propTypes = {
   ]).isRequired,
   onOpen: PropTypes.func,
   onClose: PropTypes.func,
+  onCancel: PropTypes.func,
```
edmundoa: I don't see why we need this prop here, `BootstrapModalWrapper` has nothing to do with cancelling, or forms. It is simply a wrapper to make handling modals a bit easier, or at least that is its motivation.

graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
```diff
@@ -67,7 +67,10 @@ class BootstrapModalForm extends Component {
     );
 
     return (
-      <BootstrapModalWrapper ref="modal" onOpen={this.props.onModalOpen} onClose={this.props.onModalClose}>
+      <BootstrapModalWrapper ref="modal"
+                             onOpen={this.props.onModalOpen}
+                             onClose={this.props.onModalClose}
+                             onCancel={this._onModalCancel}>
```
edmundoa: After 6ce4be2e4dfcec03333fbcd9ada9457e78de23e2, things are starting to get confusing again. Shouldn't we rename the original `onModalClose` to `onCancel`, and just keep a single callback?

The thing is, if I didn't miss anything, closing the modal and cancelling is doing the same thing right now: call the `onCancel` callback if available, and then call `close` on the `BootstrapModalWrapper`, that will eventually call the `onModalClose` prop. Well, when the modal is closed with escape, clicking outside, or pressing the "x", we call `onModalClose` twice.

graylog2-web-interface/src/components/bootstrap/BootstrapModalWrapper.jsx
dennisoelkers: That is correct. I will rename it to `onHide` just as the `Modal` component itself names it.

graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
dennisoelkers: You are correct, I fixed that.

edmundoa:
These changes will require a consolidation on how we treat data when modals are closed, cancelled. Looking through the web interface, this is another point of inconsistency on our side. Will create a follow up ticket for that.

Other than that, looks good to me 👍

PR #2407, opened 2016-06-23
Web If: Updating a few dependencies which are safe to update.

dennisoelkers:
This PR is updating a few web if dependencies which seem to be safe for updating.

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -67,15 +67,17 @@
     "babel-plugin-react-transform": "^1.1.1",
     "clean-webpack-plugin": "^0.1.3",
     "css-loader": "^0.23.0",
-    "enzyme": "^1.4.1",
+    "enzyme": "^2.0.0",
     "eslint": "^2.2.0",
-    "eslint-config-airbnb": "6.0.2",
+    "eslint-config-airbnb": "^9.0.0",
     "eslint-loader": "^1.0.0",
-    "eslint-plugin-react": "^4.1.0",
+    "eslint-plugin-import": "^1.9.2",
+    "eslint-plugin-jsx-a11y": "^1.5.3",
+    "eslint-plugin-react": "^5.0.0",
     "estraverse-fb": "^1.3.1",
     "extract-text-webpack-plugin": "^0.9.x",
     "file-loader": "^0.8.4",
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/package.json
```diff
@@ -99,6 +101,6 @@
     "url-loader": "^0.5.6",
     "webpack": "^1.12.6",
     "webpack-dev-server": "^1.12.0",
-    "webpack-merge": "^0.7.x"
+    "webpack-merge": "^0.x.x"
```
joschi: Isn't this a bit too lenient?

https://docs.npmjs.com/files/package.json#dependencies

dennisoelkers: Due to its very limited usage and small api (we are consuming a single function) it should be okay.

graylog2-web-interface/package.json
```diff
@@ -67,15 +67,17 @@
     "babel-plugin-react-transform": "^1.1.1",
     "clean-webpack-plugin": "^0.1.3",
     "css-loader": "^0.23.0",
-    "enzyme": "^1.4.1",
+    "enzyme": "^2.0.0",
     "eslint": "^2.2.0",
-    "eslint-config-airbnb": "6.0.2",
+    "eslint-config-airbnb": "^9.0.0",
```
edmundoa: I can see why we want to keep airbnb's style guide more or less up-to-date, but honestly at the moment I feel against updating it.

First, every time this changes there are a bunch of rules which are annoying and we need to disable. For instance, after installing this new version all our internal imports are marked as red by eslint.

Additionally, I think it makes harder for us to achieve a decent styling across our project, because even the newly formatted code is again outdated. Before doing this update we still have over 500 styling issues (discarding the legacy folder) only in the web interface project.

My proposal would be to hold updates on the eslint config until we are in a good shape style-wise, at least in the web-interface project, and then we should enforce the linting check. Once we are there, I think it would be easy to update the configuration every now and then, and get a reasonable compromise between new rules, period of stability for the styling, and time wasted adapting code to it.

dennisoelkers: What about spending more amount of work updating it and disable the rules that do not make sense?

edmundoa:
LGTM 👍

PR #2408, opened 2016-06-23
Decorate search results

dennisoelkers:
First version of message decorators. Not completed yet, but already open for comments/remarks.

You need Graylog2/graylog-plugin-pipeline-processor@63430c1ee90371f988e9144736d9421225685dac for it to work for now, as this is the only message decorator that is provided at the moment.

The general idea of the implementation is to provide a `MessageDecorator` interface, which is utilized in the `SearchResource` to process `ResultMessage` lists before they are encapsulated in a `SearchResult` and returned to the user for a search request. In general, a `MessageDecorator` can do anything when processing a message list, even adding or removing whole messages.

Those decorators are added to a search result view from the sidebar. You can add one or more decorators to a search result. For now these are persisted globally (so every user sees it), but separated by stream (or no stream at all).

REST-wise, decorators can be turned on/off using the boolean `decorate`-query parameter, defaulting to `true`.


bernd:
It would be cool if you could elaborate on the implementation a bit. That would make it a bit easier to review. :smiley: Thanks!

kroepke:
I noticed that added fields from a decorator do not show up in the fields section in the sidebar.
They probably need some special handling because you cannot search for them because they might not exist (or contain the right data) in the search index.

dennisoelkers:
I elaborated on the implementation a bit in the description. :)

graylog2-web-interface/src/components/search/Decorator.jsx
```diff
@@ -0,0 +1,77 @@
...
+        </Col>
+        <Col md={12}>
+          <ConfigurationWell key={`configuration-well-decorator-${decorator._id}`}
+                             id={decorator._id}
+                             configuration={decorator.config}
+                             typeDefinition={this.props.typeDefinition}/>
+        </Col>
+        <ConfigurationForm ref="editForm"
+                           key="configuration-form-decorator"
+                           configFields={this.props.typeDefinition.requested_configuration}
```
kroepke: I would prefer if plugins could (optionally maybe?) supply their custom configuration component instead of having to fit everything into the configuration forms.
From a UX perspective I think especially the pipeline decorators would benefit from this.

dennisoelkers: 👍 I had the same in mind, but postponed it. I also think that we should just render a configuration form per default if no configuration component is supplied.

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorBindings.java
```diff
@@ -0,0 +1,32 @@
...
+import com.google.inject.multibindings.Multibinder;
+import org.graylog2.plugin.decorators.MessageDecorator;
+import org.graylog2.plugin.decorators.SearchResponseDecorator;
+import org.graylog2.plugin.inject.Graylog2Module;
+
+public class DecoratorBindings extends Graylog2Module {
+    @Override
+    protected void configure() {
+        Multibinder<MessageDecorator> messageDecoratorMultibinder = Multibinder.newSetBinder(binder(), MessageDecorator.class);
+        //messageDecoratorMultibinder.addBinding().to(UpperCaseDecorator.class);
```
kroepke: I guess we can remove that line.

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+import org.graylog2.indexer.results.ResultMessage;
+import org.graylog2.rest.resources.search.responses.SearchResponse;
+
+import java.util.List;
+import java.util.Optional;
+
+@ImplementedBy(DecoratorProcessorImpl.class)
+public interface DecoratorProcessor {
+    List<ResultMessage> decorate(List<ResultMessage> messages, Optional<String> stream);
+    SearchResponse decorate(SearchResponse searchResponse, Optional<String> stream);
```
kroepke: I'm not sure whether we really want both of these methods.
Currently, if a decorator creates a new field or additional messages, the metadata in `SearchResponse` is incorrect because of how it is assembled.
Additionally it seems like `decorate(SearchResponse)` includes what `decorate(List<ResultMessage>)` does. Both have access to the list of messages.

I'd vote for simplifying the interface to only process `SearchResponse`. I assume the first method is there to make it easier to implement a decorator, but I think then it becomes much harder to automatically keep the search metadata in sync (because we need to track the field names, result counts etc).

Additionally I think the result needs to be somehow annotated to allow the UI to hide the "search for field content" button: for synthetic fields there won't be anything to search for.

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorTypeInfo.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final String FIELD_LINK_TO_DOCS = "link_to_docs";
+
+    @JsonProperty(FIELD_TYPE)
+    public abstract String type();
+    @JsonProperty(FIELD_NAME)
+    public abstract String name();
+    @JsonProperty(FIELD_HUMAN_NAME)
+    public abstract String humanName();
+    @JsonProperty(FIELD_IS_EXCLUSIVE)
+    public abstract boolean isExclusive();
```
kroepke: This seems like it is unused/erroneously copied from inputs :)

graylog2-server/src/main/java/org/graylog2/decorators/UpperCaseDecorator.java
```diff
@@ -0,0 +1,31 @@
...
+package org.graylog2.decorators;
+
+import org.graylog2.indexer.results.ResultMessage;
+import org.graylog2.plugin.Message;
+import org.graylog2.plugin.decorators.MessageDecorator;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class UpperCaseDecorator implements MessageDecorator {
```
kroepke: Unused? Seems like a no-op, too.

graylog2-server/src/main/java/org/graylog2/rest/resources/search/AbsoluteSearchResource.java
```diff
@@ -62,8 +65,10 @@
     private static final Logger LOG = LoggerFactory.getLogger(AbsoluteSearchResource.class);
 
     @Inject
-    public AbsoluteSearchResource(Searches searches, ClusterConfigService clusterConfigService) {
-        super(searches, clusterConfigService);
+    public AbsoluteSearchResource(Searches searches,
+                                  ClusterConfigService clusterConfigService,
+                                  DecoratorProcessor decoratorProcessor) {
```
kroepke: Once merged this will require changes in https://github.com/Graylog2/graylog-plugin-map-widget/blob/03b07372fc5f8f097bf0327adfc92df2920ee39b/src/main/java/org/graylog/plugins/map/rest/MapDataResource.java#L56-L56 as well

graylog2-server/src/main/java/org/graylog2/rest/resources/search/SearchResource.java
```diff
@@ -58,22 +61,32 @@
...
 
     protected final Searches searches;
     private final ClusterConfigService clusterConfigService;
+    private final DecoratorProcessor decoratorProcessor;
 
     @Inject
-    public SearchResource(Searches searches, ClusterConfigService clusterConfigService) {
+    public SearchResource(Searches searches,
+                          ClusterConfigService clusterConfigService,
+                          DecoratorProcessor decoratorProcessor) {
```
kroepke: This requires changing https://github.com/Graylog2/graylog-plugin-map-widget/blob/03b07372fc5f8f097bf0327adfc92df2920ee39b/src/main/java/org/graylog/plugins/map/rest/MapDataResource.java#L56-L56 too

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -33,6 +33,10 @@ const ConfigurationForm = React.createClass({
     newState.values = $.extend(newState.values, values);
     this.setState(newState);
   },
+  /*shouldComponentUpdate(nextProps, nextState) {
```
kroepke: left over from experiments?

graylog2-web-interface/src/components/search/Decorator.jsx
kroepke: Makes sense!

graylog2-server/src/main/java/org/graylog2/shared/security/RestPermissions.java
```diff
@@ -34,216 +34,223 @@
      *These should all be in the form of "group:action", because {@link Permissions#allPermissionsMap()} below depends on it.
      * Should this ever change, you need to adapt the code below, too.
      */
-    public static final String USERS_CREATE = "users:create";
```
kroepke: I would prefer not to make formatting changes in this PR.

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessor.java
dennisoelkers: That is exactly the rationale behind this, ease of implementing a decorator vs. completeness. I also share your conclusion of using only the latter, but in the next iteration.

kroepke:
With the exception of the formatting changes in `RestPermissions` the patch looks good to me for a first alpha. The rest we can do with additional PRs to keep them a bit smaller and focused.

@bernd any comments from your side?

bernd:
@kroepke what about the following?

* https://github.com/Graylog2/graylog2-server/pull/2408/files#r69540832

Apart from that and the formatting changes: LGTM :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/search/SearchResource.java
dennisoelkers: Did that and opened a PR: https://github.com/Graylog2/graylog-plugin-map-widget/pull/14

kroepke:
lgtm we'll do more work in follow up PRs

PR #2435, opened 2016-06-29
Remove "_ttl" in index mapping

joschi:
The [`_ttl` field](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-ttl-field.html) isn't being used by Graylog, has been deprecated in Elasticsearch 2.0.0 and will be removed in Elasticsearch 5.0.0.

Refs elastic/elasticsearch#19031

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
@@ -52,9 +52,7 @@ public class IndexMapping {
         return ImmutableMap.of(
                 "properties", partFieldProperties(analyzer),
                 "dynamic_templates", partDefaultAllInDynamicTemplate(),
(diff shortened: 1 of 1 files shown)


edmundoa:
Should we also remove or mark to be removed the `_ttl` reserved field in the [`Message`](https://github.com/Graylog2/graylog2-server/blob/4fa032b235180139c637cfabdbe00f8b17bc028e/graylog2-server/src/main/java/org/graylog2/plugin/Message.java#L73-L73) class?

joschi:
> Should we also remove or mark to be removed the `_ttl` reserved field in the `Message` class?

I thought about that but would wait until Elasticsearch 5.x with that because the `_ttl` and `_timestamp` fields are deprecated in Elasticsearch 2.x but have not been removed.

edmundoa:
LGTM 👍

PR #2438, opened 2016-06-29
Add raw message loader

edmundoa:
This PR adds a raw message loader that can be used standalone or integrated in our [`LoaderTabs` component](https://github.com/Graylog2/graylog2-server/blob/35e16872bcce65d62441495244166b65786c4c6d/graylog2-web-interface/src/components/messageloaders/LoaderTabs.jsx). It is required as part of the pipeline simulator, so the component is still not in use.

Summary of the changes done in here:
- Add rest resource to parse raw message
- Add rest resource returning codec types and their configuration
- Convert old `MessagesStore` into a Reflux store
- Add stores and actions to be able to use the previous resources
- Add component capable of loading a raw message
- Adapt `LoaderTabs` to offer the possibility of using the raw message loader

As this PR is already too big, I have avoided refactoring and reformatting `LoaderTabs` and the old message loaders. I will work on that once this PR is merged, so it is a bit more sane reviewing that afterwards.

I mentioned that the `RawMessageLoader` component is still not used, so in order to test it you can modify one of the components using `LoaderTabs`. This is an example for `AddExtractorWizard`:
```diff
       const extractorFieldActions = <MessageFieldExtractorActions fieldName="" message={{}}/>;
       extractorForm = (
         <div className="stream-loader">
-          <LoaderTabs selectedInputId={this.props.inputId} customFieldActions={extractorFieldActions} />
+          <LoaderTabs tabs={['recent', 'messageId', 'raw']} selectedInputId={this.props.inputId} customFieldActions={extractorFieldActions} />
         </div>
       );
     }
```


bernd:
I created a simple raw message and switched to the "Recent Message" tab. The message from the "Raw Message" tab was sill visible. It only disappears when I pressed "Load Message" for the "Recent Message". This was a bit confusing. Not a big issue though.

Might be more confusing if there is no recent message. What happens then?

acccaaaiiiiaaaggggibfaorbi apiwaccccaaaiiiiaaaghykuaytfkf2ggggaaccccaaaiiiicarqkeqyvgkysaaggggaaccccaaaiiwcnwfx6avjl4bflvaaaaaelftksuqmcc

bernd:
I am getting the following error when trying to analyze a message with the netflow coded from the netflow plugin. Looks like the message parser cannot handle multi-message codecs yet.

```
2016-07-01 10:43:26,421 ERROR: org.graylog2.shared.rest.exceptionmappers.AnyExceptionClassMapper - Unhandled exception in REST resource
java.lang.UnsupportedOperationException: MultiMessageCodec class org.graylog.plugins.netflow.codecs.NetFlowCodec does not support decode()
	at org.graylog.plugins.netflow.codecs.NetFlowCodec.decode(NetFlowCodec.java:56) ~[?:?]
	at org.graylog2.rest.resources.messages.MessageResource.decodeMessage(MessageResource.java:155) ~[classes/:?]
	at org.graylog2.rest.resources.messages.MessageResource.parse(MessageResource.java:149) ~[classes/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_91]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_91]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_91]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_91]
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:161) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$TypeOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:205) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:99) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:389) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:347) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:102) ~[jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.ServerRuntime$2.run(ServerRuntime.java:326) [jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:271) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:267) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:315) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:297) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.internal.Errors.process(Errors.java:267) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:317) [jersey-common-2.22.1.jar:?]
	at org.glassfish.jersey.server.ServerRuntime.process(ServerRuntime.java:305) [jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.server.ApplicationHandler.handle(ApplicationHandler.java:1154) [jersey-server-2.22.1.jar:?]
	at org.glassfish.jersey.grizzly2.httpserver.GrizzlyHttpContainer.service(GrizzlyHttpContainer.java:384) [jersey-container-grizzly2-http-2.22.1.jar:?]
	at org.glassfish.grizzly.http.server.HttpHandler$1.run(HttpHandler.java:224) [grizzly-http-server-2.3.23.jar:2.3.23]
	at com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:176) [metrics-core-3.1.2.jar:3.1.2]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_91]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_91]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_91]
```

bernd:
I was also wondering if it makes sense to configure the codec when we want to create an extractor for an existing input. Maybe the component needs a flag to disable the codec config?

graylog2-web-interface/src/logic/message/MessageFormatter.js
```diff
@@ -4,18 +4,26 @@ import MessageFieldsFilter from 'logic/message/MessageFieldsFilter';
...
-      fields: message,
-      index: messageSummary.index,
-      source_node_id: message.gl2_source_node,
-      source_input_id: message.gl2_source_input,
+      fields: fields,
+      index: index,
+      source_node_id: fields.gl2_source_node,
+      source_input_id: fields.gl2_source_input,
       stream_ids: message.streams,
-      highlight_ranges: messageSummary.highlight_ranges,
```
bernd: The `highlight_ranges` isn't needed anymore?

edmundoa:
It's hard to reply in here making sense, as messages are not following any order, so I'll just quote from your comments:

> I created a simple raw message and switched to the "Recent Message" tab. The message from the "Raw Message" tab was sill visible. It only disappears when I pressed "Load Message" for the "Recent Message". This was a bit confusing. Not a big issue though.

I think we should remove the preview when the tabs change. The inputs should still have the data, so it's a matter of click "load message" again. Will take care of that.

> I am getting the following error when trying to analyze a message with the netflow coded from the netflow plugin. Looks like the message parser cannot handle multi-message codecs yet.

Multi-message codecs are not supported yet, and I think we only have one of those. Not sure how important it is, but I can take a look, probably it's not too hard to do.

> I was also wondering if it makes sense to configure the codec when we want to create an extractor for an existing input. Maybe the component needs a flag to disable the codec config?

For now we are not using the loader in there, and there will be some changes to do if we want to do so. For instance, the REST resource for that only supports loading messages with an ID and index. I would rather wait with any changes for that until we do it and see what we actually need for it.

graylog2-web-interface/src/logic/message/MessageFormatter.js
edmundoa: Probably I removed it by mistake, will investigate. Nice catch anyway 👍

bernd:
> Multi-message codecs are not supported yet, and I think we only have one of those. Not sure how important it is, but I can take a look, probably it's not too hard to do.

I think it's not important right know. A better error message would be good though. The user currently gets a 500 error message that doesn't say much.

graylog2-web-interface/src/logic/message/MessageFormatter.js
edmundoa: It was indeed needed, so I added it back ✅

bernd:
LGTM :+1:

PR #2440, opened 2016-06-30
Allowing path prefixes in web_listen_uri so web interface is accessible via path != "/"

### LINKED ISSUE #2271: Feature request: Specify application.context
As noted in the docs,

> Please take note that the application.context configuration parameter present in Graylog 1.x (and earlier) is not existing anymore. The web interface can currently only be served without a path prefix.

Can this feature be added into V2? We depend heavily on this functionality, and are unlikely to upgrade until it's back.

dennisoelkers:
We are now utilizing webpack's runtime path resolution for chunks which are loaded on demand to allow configuring a `web_listen_uri` in the config file which includes a path.

This is useful where setups where the web interface is supposed to be accessible via something like `http://example.org/graylog`.

Fixes #2271 

**Caution:** The way it works right now, it seems that the changes made to [URLUtils.js](https://github.com/Graylog2/graylog2-server/blob/c72f3984f121586c8ecba7dd1c4b7cf8d2a1d40b/graylog2-web-interface/src/util/URLUtils.js) require plugins to be rebuilt to work.


kroepke:
Extensively rebuilt everything at least 5 times: works fine now 👍

PR #2441, opened 2016-06-30
Extracting our customized ESLint config into separate module.

dennisoelkers:
This module is consumable by plugins too, so we are one step closed to
establishing consistent code quality.

All changes to the ESLint config should be done in the module from now
on, so they propagate to plugins consuming it as well.


edmundoa:
LGTM 👍

PR #2444, opened 2016-07-01
Allow access to MongoDatabase in MongoConnection

joschi:
This PR adds a method to `MongoConnection` which allows using the `MongoDatabase` interface to access MongoDB instead of the deprecated `DB` interface.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
@@ -23,7 +23,19 @@ import com.mongodb.client.MongoDatabase;
 public interface MongoConnection {
     Mongo connect();
 
+    /**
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
```diff
@@ -18,10 +18,12 @@
 
 import com.mongodb.DB;
 import com.mongodb.Mongo;
-import com.mongodb.MongoClient;
+import com.mongodb.client.MongoDatabase;
 
 public interface MongoConnection {
     Mongo connect();
 
     DB getDatabase();
```
bernd: I think we should mark this as `@Deprecated` so we can remove it for the next major release.

bernd:
LGTM :+1:

PR #2448, opened 2016-07-04
Proxies deflector cycle call to make it available on every node.

dennisoelkers:
This PR adds a new cluster resource that finds the master node and
triggers a deflector cycle. This is the last request that needs to be
directed against a master node and after this change, all nodes in a
cluster can be interchangeable called for every request.

### DIFF
--- a/graylog2-web-interface/src/routing/ApiRoutes.js
+++ b/graylog2-web-interface/src/routing/ApiRoutes.js
@@ -43,7 +43,7 @@ const ApiRoutes = {
     updatePositions: (dashboardId) => { return { url: `/dashboards/${dashboardId}/positions` }; },
   },
   DeflectorApiController: {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/routing/ApiRoutes.js
```diff
@@ -43,7 +43,7 @@ const ApiRoutes = {
     updatePositions: (dashboardId) => { return { url: `/dashboards/${dashboardId}/positions` }; },
   },
   DeflectorApiController: {
-    cycle: () => { return { url: '/system/deflector/cycle' }; },
+    cycle: () => { return { url: '/cluster/system/deflector/cycle' }; },
```
bernd: The URL is wrong, this should be `/cluster/deflector/cycle`.

bernd:
LGTM :+1:

PR #2450, opened 2016-07-04
Report syntax error when search query contains unescaped slash

### LINKED ISSUE #2372: 500 error on slash in query
### Problem description
When searching something with a forward-slash `/` but that is not escaped, the web ui keeps hanging in a `loading...` state. I would expect malformed syntax feedback.

1. Access the following url on a graylog instance: `/search?rangetype=relative&fields=message%2Csource&width=1879&highlightMessage=&relative=300&q=bla%3A%2Ffoo`

### Environment

* Graylog Version: 2.0.2
* Elasticsearch Version: 2.3.3
* MongoDB Version: 3.0.10
* Browser version: Chrome 51

edmundoa:
When a search query contains an unescaped slash, calling to `queryParser.generateParseException()` throws a `NullPointerException`, so the REST API returns a 500 error instead of 400.

Fixes #2372


joschi:
LGTM. 👍

PR #2453, opened 2016-07-04
Better logging in case of failed ClusterConfig updates

lennartkoopmann:
Before this change it was really hard to figure out why a ClusterConfig update would fail. Especially in case of a key mismatch between the web config and the ClusterConfig class, there was no way that out.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
@@ -32,6 +32,8 @@ import org.graylog2.shared.plugins.ChainingClassLoader;
 import org.graylog2.shared.rest.resources.RestResource;
 import org.graylog2.shared.security.RestPermissions;
 import org.hibernate.validator.constraints.NotBlank;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
@@ -58,6 +60,9 @@ import static java.util.Objects.requireNonNull;
 @Path("/system/cluster_config")
 @Produces(MediaType.APPLICATION_JSON)
 public class ClusterConfigResource extends RestResource {
+    private static final Logger LOG = LoggerFactory.getLogger(ClusterConfigResource.class);
+
+
     private final ClusterConfigService clusterConfigService;
     private final ChainingClassLoader chainingClassLoader;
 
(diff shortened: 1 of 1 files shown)


kroepke:
If this is targeted for 2.1, I think the PR should be against master. I'm trying to keep the released branches as lean as possible.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
```diff
@@ -107,10 +112,14 @@ public Response update(@ApiParam(name = "configClass", value = "The name of the
...
-        clusterConfigService.write(o);
-
-        return Response.accepted(o).build();
+        try {
+            final Object o = objectMapper.readValue(body, cls);
+            clusterConfigService.write(o);
+            return Response.accepted(o).build();
+        }catch(Exception e) {
+            LOG.error("Could not write new cluster config.", e);
+            return Response.status(400).build();
```
joschi: The response should contain some text about the nature of the problem.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
```diff
@@ -107,10 +112,14 @@ public Response update(@ApiParam(name = "configClass", value = "The name of the
...
 
-        final Object o = objectMapper.readValue(body, cls);
-        clusterConfigService.write(o);
-
-        return Response.accepted(o).build();
+        try {
+            final Object o = objectMapper.readValue(body, cls);
+            clusterConfigService.write(o);
+            return Response.accepted(o).build();
+        }catch(Exception e) {
```
joschi: Missing whitespace. :wink:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
joschi: Maybe simply throw a `WebApplicationException` here and add the original exception for its stack trace.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterConfigResource.java
```diff
@@ -58,6 +60,9 @@
 @Path("/system/cluster_config")
 @Produces(MediaType.APPLICATION_JSON)
 public class ClusterConfigResource extends RestResource {
+    private static final Logger LOG = LoggerFactory.getLogger(ClusterConfigResource.class);
+
+
```
joschi: Useless newline.

PR #2476, opened 2016-07-13
Handle situation where more than one index has the deflector alias

bernd:
This can happen when some ES nodes are down and index rotation/retention kicks in. Once the absent nodes are back, they might bring back an already deleted index including the deflector alias.

Once we detect more than one index for the deflector alias, we throw an exception. In that case the index rotation thread tries to cleanup the aliases by removing it from all indices but the latest one.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
@@ -274,7 +274,7 @@ public class Deflector { // extends Ablenkblech
     }
 
     @Nullable
-    public String getCurrentActualTargetIndex() {
+    public String getCurrentActualTargetIndex() throws Indices.ESAliasesException {
         return indices.aliasTarget(getName());
     }
 
@@ -296,5 +296,23 @@ public class Deflector { // extends Ablenkblech
 
     public boolean isGraylogIndex(final String indexName) {
         return !isNullOrEmpty(indexName) && !isDeflectorAlias(indexName) && indexPattern.matcher(indexName).matches();
+    }
+
+    /**
+     * Sorts the given set of index names and removes the deflector alias from all indices but the last one.
+     *
+     * This might be used when multiple indices have the same alias. This should be really rare!
+     *
+     * @param indexNames the set of index names that have the deflector alias
+     */
+    public void cleanupAliases(Set<String> indexNames) {
+        final Set<String> indicesToRemove = indexNames.stream()
+                .mapToInt(Deflector::extractIndexNumber)
+                .sorted()
+                .limit(indexNames.size() - 1) // Get all but the newest one
+                .mapToObj(num -> indexPrefix + SEPARATOR + num)
+                .collect(Collectors.toSet());
+
+        indices.removeAliases(getName(), indicesToRemove);
     }
 }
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -585,4 +596,21 @@ public TimestampStats timestampStatsOfIndex(String index) {
 
         return TimestampStats.create(min, max);
     }
+
+    public static class ESAliasesException extends Exception {
```
joschi: How about naming it `TooManyAliasTargetsException` (and also making it a standalone class)?

bernd: :+1:

graylog2-server/src/main/java/org/graylog2/indexer/indices/TooManyAliasesException.java
```diff
@@ -0,0 +1,36 @@
...
+public class TooManyAliasesException extends Exception {
+    private final Set<String> indices;
+
+    public TooManyAliasesException(final Set<String> indices) {
+        this.indices = indices;
+    }
+
+    @Override
+    public String getMessage() {
+        return "More than one index in deflector alias: " + indices.toString();
```
joschi: We can already set the message in the constructor (see [`Exception(String message)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html#Exception-java.lang.String-)) and get rid of this method:
```java
public TooManyAliasesException(final Set<String> indices) {
  super("More than one index in deflector alias: " + indices.toString());
  this.indices = indices;
}
```

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -494,6 +499,12 @@ public boolean cycleAlias(String aliasName, String targetIndex, String oldIndex)
                 .execute().actionGet().isAcknowledged();
     }
 
+    public boolean removeAliases(String alias, Set<String> indices) {
+        return c.admin().indices().prepareAliases()
+                .removeAlias(indices.toArray(new String[]{}), alias)
```
joschi: Aesthetic nit-pick: This could be `indices.toArray(new String[0])`

graylog2-server/src/main/java/org/graylog2/indexer/Deflector.java
```diff
@@ -297,4 +298,22 @@ public boolean isGraylogDeflectorIndex(final String indexName) {
...
+     * This might be used when multiple indices have the same alias. This should be really rare!
+     *
+     * @param indexNames the set of index names that have the deflector alias
+     */
+    public void cleanupAliases(Set<String> indexNames) {
+        final Set<String> indicesToRemove = indexNames.stream()
+                .mapToInt(Deflector::extractIndexNumber)
+                .sorted()
+                .limit(indexNames.size() - 1) // Get all but the newest one
+                .mapToObj(num -> indexPrefix + SEPARATOR + num)
```
joschi: This will break if the alias is pointing to indices that do not follow the Graylog index naming scheme. 😞

joschi: Hm, in the worst case it will already fail in line 311 with a `NumberFormatException`.

Issue #2429, opened 2016-06-28
Elasticsearch Red cluster state triggered by index rotation under some conditions. 


jehuty0shift:
### Problem description

When a node of the Elasticsearch cluster is down, A simple index rolling can put the cluster in red 

-  One Elasticsearch node is down (on scheduled maintenance for example), so the cluster turns Yellow. Shard reallocation is delayed (with  index.unassigned.node_left.delayed_timeout) since node is expected to return.  
-  The cluster cycle one index so delete a previous one.  
-  The node comes back and import the previous index that is now a dangling index. 
-  The cluster turns to red, unable to retrieve previous shards that have been deleted. 

### Steps to reproduce the problem

To reproduce this problem, so the following 

-   delay re-allocation of shards by using this index setting on the earliest index created : 

`{ "settings" : {"index.unassigned.node_left.delayed_timeout" :  "15m" } }`

-  stop one ES node with some shards of this index on it. 

-  Cycle the index. It will delete the earliest index if max number of index is reached 

-  Restart the ES node, it will import its cluster index, turning the cluster to Red.
 
### Environment

* Graylog Version: 2.0.3
* Elasticsearch Version: 2.3.3
* MongoDB Version: 3.2.7
* Operating System: Ubuntu 14.04
* Browser version: Chromium Version 50.0.2661.102 Ubuntu 15.10 (64-bit)

#### Additionnal Informations 

- The setting index.unassigned.node_left.delayed_timeout is very useful when you want to do maintenance on some nodes or in case of temporary failure (It will happen when you have a big ES cluster.), This prevent unneccesary reallocation of replicas, when you know that your node will come back. And If i recall correctly, in ES 2.x the default value is now 1 minute. 
- The real problem is that Elastic removed one setting from ES 2.0  : 
*gateway.local.auto_import_dangled*
Now all the indices are forcefullty imported (see this issue [https://github.com/elastic/elasticsearch/pull/10016](https://github.com/elastic/elasticsearch/pull/10016)). 

- The issue could be mitigated by implementing this feature:  #2371 and by alerting the administrators of the real state of the cluster.

kroepke:
This sounds very reasonable to me and the first glance looks like it should be relatively easy to implement.
Scheduling for 2.1.

bernd:
This has been implemented in #2477 and will be part of the upcoming Graylog 2.1. Thank you for the report!

PR #2477, opened 2016-07-13
Allow indexing when cluster state is red but write index state is healthy

### LINKED ISSUE #2371: Allow indexing even if ElasticSearch cluster health status is 'red', but current index (deflector) is 'green'
Today, Graylog 2.0.2 will pause indexing if the cluster health status is 'red'.  However (if I'm not mistaken), ElasticSearch will still permit indexing to occur if the index being written to isn't 'red'.

This is a feature request to add a configuration property to Graylog Server to enable indexing if the ElasticSearch index pointed-to by the deflector is 'green' (or even 'yellow') -- even if the cluster itself is reporting a health status of 'red' (which could mean an old index, or even an index not managed by Graylog is 'red').

See:

https://github.com/Graylog2/graylog2-server/blob/2.0.2/graylog2-server/src/main/java/org/graylog2/outputs/BlockingBatchedESOutput.java .

and:

https://github.com/Graylog2/graylog2-server/blob/2.0.2/graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java

Thanks!

bernd:
The cluster state can turn red even though the current write index is green. This might happen when running rotation/retention during an ES maintenance window where not all nodes are present.

For index related tasks we are now checking the state of the current write index (deflector) instead of the state for all Graylog managed indices.

The logs and UI are still showing the red cluster state to make sure the admin will be notified.

Refs #2429
Fixes #2371

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
@@ -71,8 +71,26 @@ public class Cluster {
         ClusterStateMonitor.setCluster(this);
     }
 
+    /**
+     * Requests the cluster health for all indices managed by Graylog. (default: graylog_*)
+     *
+     * @return the cluster health response
+     */
     public ClusterHealthResponse health() {
         ClusterHealthRequest request = new ClusterHealthRequest(deflector.getDeflectorWildcard());
+        return c.admin().cluster().health(request).actionGet();
+    }
+
+    /**
+     * Requests the cluster health for the current write index. (deflector)
+     *
+     * This can be used to decide if the current write index is healthy and writable even when older indices have
+     * problems.
+     *
+     * @return the cluster health response
+     */
+    public ClusterHealthResponse deflectorHealth() {
+        ClusterHealthRequest request = new ClusterHealthRequest(deflector.getName());
         return c.admin().cluster().health(request).actionGet();
     }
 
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -158,14 +176,37 @@ public boolean isHealthy() {
...
+    public void waitForConnectedAndDeflectorHealthy(long timeout, TimeUnit unit) throws InterruptedException, TimeoutException {
+        LOG.debug("Waiting until cluster connection comes back and the write index is healthy, checking once per second.");
 
         final CountDownLatch latch = new CountDownLatch(1);
         final ScheduledFuture<?> scheduledFuture = scheduler.scheduleAtFixedRate(() -> {
             try {
-                if (isConnected() && isHealthy()) {
-                    LOG.debug("Cluster is healthy again, unblocking waiting threads.");
+                if (isConnected() && isDeflectorHealthy()) {
+                    LOG.debug("Cluster and write index is healthy again, unblocking waiting threads.");
```
joschi: I think this should be "or".

I would remove the part about the cluster completely: "Write-active index is healthy again, unblocking waiting threads."

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -176,12 +217,18 @@ public void waitForConnectedAndHealthy(long timeout, TimeUnit unit) throws Inter
         scheduledFuture.cancel(true); // Make sure to cancel the task to avoid task leaks!
 
         if (!waitSuccess) {
-            throw new TimeoutException("Elasticsearch cluster didn't get healthy within timeout");
+            throw new TimeoutException("Elasticsearch cluster or write index didn't get healthy within timeout");
```
joschi: I would remove the part about the cluster completely: "Write-active index didn't get healthy within timeout."

joschi:
LGTM except for those 2 wordings. 👍

PR #2481, opened 2016-07-14
Add whitespace replacement and prefix options to JSON extractor

### LINKED ISSUE #1646: json extractor prefix
Right now you can not add a prefix to the json parser, this means if a json contains fields like timestamp or message the root field is overwritten.

In this case the message cannot be written to elasticsearch
```
[0]: index [graylog2_0], type [message], id [20e39820-a8bd-11e5-8ec4-00163e2b2cfe], message [MapperParsingException[failed to parse [timestamp]]; nested: MapperParsingException[failed to parse date field [2015-12-22T15:03:10.353722+0000], tried both date format [yyyy-MM-dd HH:mm:ss.SSS], and timestamp number with locale []]; nested: IllegalArgumentException[Invalid format: "2015-12-22T15:03:10.353722+0000" is malformed at "T15:03:10.353722+0000"]; ]
```

This is the a sample field sensu:
```
{"timestamp":"2015-12-22T15:08:26.058363+0000","level":"warn","message":"reconnecting to transport"}
```
With prefix graylog could create new fields like
sensu_level
sensu_message
...

### LINKED ISSUE #2434: json extractor should check for valid lucene keys
### Problem description
If fieldnames contain spaces, the extractor extracts those fields, but
later in the search those fields are not written/cannot be found, as `space` _is not allowed_ in lucene fieldnames.

#### Probable solution
One could just replace any invalid character in fieldnames produced by the json extractor (replace spaces)

1. send message `{"my key": "value"}`
2. Extract it with json extractor (in preview you will see the field extracted
3. Search for the message and there will be no key 'my key'

### Environment

* Graylog Version: 2.0.1-2
* Elasticsearch Version: 2
* MongoDB Version: 3
* Operating System: docker container
* Browser version: FF 45

bernd:
This adds two new options to the JSON extractor.

- Allow replacement of whitespace in the message keys (#2434)
- Add key prefix option (#1646)

Fixes #2434
Fixes #1646

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/extractors/JsonExtractor.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/extractors/JsonExtractor.java
@@ -47,6 +47,8 @@ public class JsonExtractor extends Extractor {
     private static final String CK_LIST_SEPARATOR = "list_separator";
     private static final String CK_KEY_SEPARATOR = "key_separator";
     private static final String CK_KV_SEPARATOR = "kv_separator";
+    private static final String CK_REPLACE_KEY_WHITESPACE = "replace_key_whitespace";
+    private static final String CK_KEY_WHITESPACE_REPLACEMENT = "key_whitespace_replacement";
     private static final RemoveNullPredicate REMOVE_NULL_PREDICATE = new RemoveNullPredicate();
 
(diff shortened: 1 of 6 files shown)


edmundoa:
LGTM 👍

PR #2482, opened 2016-07-14
Change decorators interface to operate on SearchResponse instead of ResultMessage

dennisoelkers:
This allows us to update the response with added fields etc. later on.

Also fixes the web interface so that any changes to the decorators config makes the search results refresh.

Requires merging Graylog2/graylog-plugin-pipeline-processor#43.


kroepke:
lgtm, waiting for CI runs to finish

PR #2483, opened 2016-07-14
Improve overflow behaviour on search results table

### LINKED ISSUE #2464: frontend: design breaks if choosing more fields (on search) than fits to browser
### Problem description

If you choose for example **show all fields** in the search and the fields did not fit into your current browser width the design is broken and no horizontal scrollbars are present. 

This should be handled better to improve the usability. Additional it does confuse users.

1. Go to the Search Page
2. resize your Browser Window
3. Choose "show all fields"

### Environment

* Graylog Version: 2.xx

edmundoa:
Let search results table scroll on overflow, and fix some styling issues while doing so.

Fixes #2464

### DIFF
--- a/graylog2-web-interface/public/stylesheets/graylog2.less
+++ b/graylog2-web-interface/public/stylesheets/graylog2.less
@@ -2169,6 +2169,36 @@ span.blob {
   border: 0;
 }
 
+table.messages {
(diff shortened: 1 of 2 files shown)


bernd:
LGTM and works for me. :+1:

PR #2484, opened 2016-07-14
Fix group membership lookup if DN contains whitespace

### LINKED ISSUE #1790: LDAP group mapping: stringwise comparison fails due to different DN formats
LDAP group mapping is a very appreciated feature for us. Thanks for implementing it!
However, the mapping seems broken if the RDNs are separated without whitespaces in the member attribute of a group (for example "uid=12345,ou=A,o=B").

This is what the trace shows:
`TRACE: org.graylog2.security.ldap.LdapConnector - DN uid=12345, ou=A, o=B == uid=12345,ou=A,o=B member?`

In LdapConnector.findGroups, the DN of ldapEntry is compared stringswise with each member attribute of a group:
```
if (dn.equalsIgnoreCase(member.getString())) {
    groups.add(groupId);
}
```
Since org.apache.directory.api.ldap.model.name.Dn.getName() returns a string where RDNs are separated by comma _and whitespace_, the strings do not match although the DNs are semantically equal.

According to RFC 2253, whitespaces are allowed between RDNs while RFC 4514 defines the string representation without whitespaces. So maybe you could remove whitespaces from the strings before comparing them by something like `replaceAll("\\s","")` or `replaceAll(",\\s",",")`. Perhaps parsing DNs would be another solution.

PS: We are using Graylog 1.3.3

bernd:
Normalize the DN before comparing the strings to avoid mapping problems if the DNs have different whitespace formatting.

Fixes #1790

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
@@ -27,10 +27,12 @@ import org.apache.directory.api.ldap.model.entry.Attribute;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.entry.Value;
 import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.exception.LdapInvalidDnException;
 import org.apache.directory.api.ldap.model.message.BindRequestImpl;
 import org.apache.directory.api.ldap.model.message.BindResponse;
 import org.apache.directory.api.ldap.model.message.ResultCodeEnum;
 import org.apache.directory.api.ldap.model.message.SearchScope;
+import org.apache.directory.api.ldap.model.name.Dn;
 import org.apache.directory.ldap.client.api.LdapConnectionConfig;
 import org.apache.directory.ldap.client.api.LdapNetworkConnection;
 import org.graylog2.plugin.DocsHelper;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
```diff
@@ -303,6 +305,32 @@ public LdapEntry search(LdapNetworkConnection connection,
...
+     */
+    private String normalizedDn(String dn) {
+        if (isNullOrEmpty(dn)) {
+            return dn;
+        } else {
+            try {
+                return new Dn(dn).getNormName();
+            } catch (LdapInvalidDnException e) {
+                LOG.warn("Invalid DN", e);
+                return dn;
```
joschi: Shouldn't this be throwing an exception instead of returning the arguably invalid DN back to the caller?

graylog2-server/src/test/resources/org/graylog2/security/ldap/base.ldif
```diff
@@ -35,6 +35,12 @@ objectClass: top
 cn: Engineers
 uniqueMember: cn=John Doe,ou=users,dc=example,dc=com
 
+dn: cn=Whitespace Engineers,ou=groups,dc=example,dc=com
```
joschi: Whitespace Engineers – Engineering something you can't see since 1942 (also non-breaking). 😆

graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
```diff
@@ -303,6 +305,32 @@ public LdapEntry search(LdapNetworkConnection connection,
...
+     *
+     *   input  = "cn=John Doe, ou=groups, ou=system"
+     *   output = "cn=John Doe,ou=groups,ou=system"
+     *
+     * See: https://github.com/Graylog2/graylog2-server/issues/1790
+     *
+     * @param dn denormalized DN string
+     * @return normalized DN string
+     */
+    private String normalizedDn(String dn) {
```
joschi: The method *could* be annotated with `@Nullable`.

graylog2-server/src/main/java/org/graylog2/security/ldap/LdapConnector.java
bernd: I didn't want to change the current behavior in case of an invalid DN. If we throw an exception here, we have to handle it somewhere or let it bubble up.

joschi: ✅

joschi:
LGTM. 👍

PR #2487, opened 2016-07-14
Make several refresh intervals configurable

### LINKED ISSUE #2391: StreamRouter interval should be configurable
Hello,

With hundreds of streams and dozens of graylog nodes, mongo load starts to be a (big) problem.
Obviously, this is mainly due to the hits done by StreamRouter every second for getting fresh rules.
As you can see on my Ganglia graph, switching this interval from 1 second to 1 minute saves me a lot of Mongo CPU.

mongo

I'm aware that this is introducing a minute of lag between stream creation and real routing and so probably need an explicit configuration directive in server.conf

bernd:
PLEASE DO NOT REVIEW, I WILL CREATE A NEW PR!

- Extractor filter
- Stream router
- Rules filter
- Static field filter
- In memory role permission resolver

Fixes #2391


bernd:
Replaced by https://github.com/Graylog2/graylog2-server/pull/2496

PR #2492, opened 2016-07-16
Change plugin REST resource injection to use Class instances

bernd:
Using the actual resource instances caused injection of all the dependencies, even when we don't need the actual instance at all. The registration simply needs to have the class. This also prevents early injection of objects which have a lifecycle but don't manage it. (this continues to be a problem, but cannot be solved on this level)


kroepke:
lgtm, good catch with missing plugins!

bernd:
> lgtm, good catch with missing plugins!

That was actually catched by the integration tests. :smile:

joschi:
LGTM. 👍

PR #2496, opened 2016-07-18
Switch message filters from polling to subscribing to change events

### LINKED ISSUE #2391: StreamRouter interval should be configurable

bernd:
Switch all `MessageFilter` implementations from polling for changes every second to subscribing to change events.

This avoids unneeded load on the Graylog and MongoDB servers.

Fixes #2391

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
@@ -16,39 +16,52 @@
  */
 package org.graylog2.filters;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
+import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
 import org.graylog2.database.NotFoundException;
 import org.graylog2.inputs.Input;
 import org.graylog2.inputs.InputService;
 import org.graylog2.plugin.Message;
 import org.graylog2.plugin.filters.MessageFilter;
 import org.graylog2.plugin.inputs.Extractor;
+import org.graylog2.rest.models.system.inputs.responses.InputCreated;
+import org.graylog2.rest.models.system.inputs.responses.InputDeleted;
+import org.graylog2.rest.models.system.inputs.responses.InputUpdated;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 public class ExtractorFilter implements MessageFilter {
     private static final Logger LOG = LoggerFactory.getLogger(ExtractorFilter.class);
     private static final String NAME = "Extractor";
+    private static final List<Extractor> EMPTY_LIST = ImmutableList.of();
 
-    private Cache<String, List<Extractor>> cache = CacheBuilder.newBuilder()
-            .expireAfterWrite(1, TimeUnit.SECONDS)
-            .build();
+    private final ConcurrentMap<String, List<Extractor>> extractors = new ConcurrentHashMap<>();
 
     private final InputService inputService;
+    private final ScheduledExecutorService scheduler;
 
     @Inject
-    public ExtractorFilter(InputService inputService) {
+    public ExtractorFilter(InputService inputService,
+                           EventBus serverEventBus,
+                           @Named("daemonScheduler") ScheduledExecutorService scheduler) {
         this.inputService = inputService;
+        this.scheduler = scheduler;
+
+        loadAllExtractors();
+
+        // TODO: This class needs lifecycle management to avoid leaking objects in the EventBus
+        serverEventBus.register(this);
     }
 
     @Override
@@ -57,7 +70,7 @@ public class ExtractorFilter implements MessageFilter {
             return false;
         }
 
-        for (final Extractor extractor : loadExtractors(msg.getSourceInputId())) {
+        for (final Extractor extractor : extractors.getOrDefault(msg.getSourceInputId(), EMPTY_LIST)) {
             try {
                 extractor.runExtractor(msg);
             } catch (Exception e) {
@@ -70,34 +83,43 @@ public class ExtractorFilter implements MessageFilter {
         return false;
     }
 
-    private List<Extractor> loadExtractors(final String inputId) {
+    @Subscribe
+    public void handleInputCreate(final InputCreated event) {
+        LOG.debug("Load extractors for input <{}>", event.id());
+        scheduler.schedule(() -> loadExtractors(event.id()), 0, TimeUnit.SECONDS);
+    }
+
+    @Subscribe
+    public void handleInputDelete(final InputDeleted event) {
+        LOG.debug("Removing input from extractors cache <{}>", event.id());
+        extractors.remove(event.id());
+    }
+
+    @Subscribe
+    public void handleInputUpdate(final InputUpdated event) {
+        scheduler.schedule(() -> loadExtractors(event.id()), 0, TimeUnit.SECONDS);
+    }
+
+    private void loadAllExtractors() {
         try {
-            return cache.get(inputId, new Callable<List<Extractor>>() {
-                @Override
-                public List<Extractor> call() throws Exception {
-                    LOG.debug("Re-loading extractors for input <{}> into cache.", inputId);
+            inputService.all().forEach(input -> loadExtractors(input.getId()));
+        } catch (Exception e) {
+            LOG.error("Unable to load extractors for all inputs", e);
+        }
+    }
 
-                    try {
-                        Input input = inputService.find(inputId);
+    private void loadExtractors(final String inputId) {
+        LOG.debug("Re-loading extractors for input <{}>", inputId);
 
-                        List<Extractor> sorted = Lists.newArrayList(inputService.getExtractors(input));
+        try {
+            final Input input = inputService.find(inputId);
+            final List<Extractor> sorted = Lists.newArrayList(inputService.getExtractors(input));
 
-                        Collections.sort(sorted, new Comparator<Extractor>() {
-                            public int compare(Extractor e1, Extractor e2) {
-                                return e1.getOrder().intValue() - e2.getOrder().intValue();
-                            }
-                        });
+            Collections.sort(sorted, (e1, e2) -> e1.getOrder().intValue() - e2.getOrder().intValue());
 
-                        return sorted;
-                    } catch (NotFoundException e) {
-                        LOG.warn("Unable to load input: {}", e.getMessage());
-                        return Collections.emptyList();
-                    }
-                }
-            });
-        } catch (ExecutionException e) {
-            LOG.error("Could not load extractors into cache. Returning empty list.", e);
-            return Collections.emptyList();
+            extractors.put(inputId, ImmutableList.copyOf(sorted));
+        } catch (NotFoundException e) {
+            LOG.warn("Unable to load input <{}>: {}", inputId, e.getMessage());
         }
     }
 
--- a/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
@@ -44,7 +44,6 @@ import java.util.concurrent.TimeUnit;
 public class ExtractorFilter implements MessageFilter {
     private static final Logger LOG = LoggerFactory.getLogger(ExtractorFilter.class);
     private static final String NAME = "Extractor";
-    private static final List<Extractor> EMPTY_LIST = ImmutableList.of();
 
     private final ConcurrentMap<String, List<Extractor>> extractors = new ConcurrentHashMap<>();
 
@@ -70,7 +69,7 @@ public class ExtractorFilter implements MessageFilter {
             return false;
         }
 
-        for (final Extractor extractor : extractors.getOrDefault(msg.getSourceInputId(), EMPTY_LIST)) {
+        for (final Extractor extractor : extractors.getOrDefault(msg.getSourceInputId(), Collections.emptyList())) {
             try {
                 extractor.runExtractor(msg);
             } catch (Exception e) {
--- a/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
@@ -38,7 +38,6 @@ import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 public class ExtractorFilter implements MessageFilter {
@@ -86,7 +85,7 @@ public class ExtractorFilter implements MessageFilter {
     @SuppressWarnings("unused")
     public void handleInputCreate(final InputCreated event) {
         LOG.debug("Load extractors for input <{}>", event.id());
-        scheduler.schedule(() -> loadExtractors(event.id()), 0, TimeUnit.SECONDS);
+        scheduler.submit(() -> loadExtractors(event.id()));
     }
 
     @Subscribe
@@ -99,7 +98,7 @@ public class ExtractorFilter implements MessageFilter {
     @Subscribe
     @SuppressWarnings("unused")
     public void handleInputUpdate(final InputUpdated event) {
-        scheduler.schedule(() -> loadExtractors(event.id()), 0, TimeUnit.SECONDS);
+        scheduler.submit(() -> loadExtractors(event.id()));
     }
 
     private void loadAllExtractors() {
--- a/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
@@ -17,7 +17,6 @@
 package org.graylog2.filters;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import org.graylog2.database.NotFoundException;
@@ -40,6 +39,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 public class ExtractorFilter implements MessageFilter {
     private static final Logger LOG = LoggerFactory.getLogger(ExtractorFilter.class);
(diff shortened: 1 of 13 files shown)


pdepaepe:
Event driven instead of arbitrary interval refresh is very smart, thanks!

May i suggest you to extend theses events to almost anything touching Mongo add/update/delete?
It will helps a lot to write any plugin based on @subscribe events.

Exemple: custom filter message chain with a cache flushed by StreamRuleAdd/Delete events

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static Builder builder() {
+        return new AutoValue_StreamsChangedEvent.Builder();
+    }
+
+    public abstract Builder toBuilder();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+        @JsonProperty(FIELD_STREAM_IDS)
```
joschi: I'm not sure that this will work as intended. Jackson will basically ignore the Builder class completely.

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
```diff
@@ -0,0 +1,63 @@
...
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
+
+@JsonAutoDetect
+@JsonDeserialize(builder = AutoValue_StreamsChangedEvent.Builder.class)
```
joschi: Any specific reason to do it this way and not with a "normal" static factory method annotated with `@JsonCreator`?

graylog2-server/src/main/java/org/graylog2/rest/models/system/inputs/responses/InputUpdated.java
```diff
@@ -0,0 +1,34 @@
...
+package org.graylog2.rest.models.system.inputs.responses;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@JsonAutoDetect
+@AutoValue
+public abstract class InputUpdated {
```
joschi: All other event classes have the "Event" suffix, why not this one?

bernd: For consistency with the existing `InputCreated` and `InputDeleted` classes.

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
bernd: I like the way you can build up immutable collections this way.

graylog2-server/src/main/java/org/graylog2/rest/models/system/inputs/responses/InputUpdated.java
joschi: Hm, okay.

I don't really like binding the internal events to the public response format of the Graylog REST API (as which `InputCreated` and `InputDeleted` are being used), but I guess I can live with it right now.

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
joschi: For the sake of consistency with the other classes, I'd prefer having a "stupid" static factory method with `@JsonCreator` in `StreamsChangedEvent`.

joschi:
@bernd What's the worst case if a node is missing one or more of those events?

`ClusterEventService` doesn't strictly guarantee that nodes eventually process the events (or process them twice).

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
joschi: ✅

graylog2-server/src/main/java/org/graylog2/rest/models/system/inputs/responses/InputUpdated.java
joschi: ✅

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -16,39 +16,52 @@
...
-import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 public class ExtractorFilter implements MessageFilter {
     private static final Logger LOG = LoggerFactory.getLogger(ExtractorFilter.class);
     private static final String NAME = "Extractor";
+    private static final List<Extractor> EMPTY_LIST = ImmutableList.of();
```
joschi: `ImmutableList.of()` already returns a constant. No need to define another one for redirection. 😉

graylog2-server/src/main/java/org/graylog2/rest/models/system/inputs/responses/InputUpdated.java
bernd: > I don't really like binding the internal events to the public response format of the Graylog REST API (as which InputCreated and InputDeleted are being used), but I guess I can live with it right now.

Me neither, but I didn't want to change all of this in this PR.

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -70,34 +83,43 @@ public boolean filter(Message msg) {
...
+    }
+
+    @Subscribe
+    public void handleInputDelete(final InputDeleted event) {
+        LOG.debug("Removing input from extractors cache <{}>", event.id());
+        extractors.remove(event.id());
+    }
+
+    @Subscribe
+    public void handleInputUpdate(final InputUpdated event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -70,34 +83,43 @@ public boolean filter(Message msg) {
...
 
-    private List<Extractor> loadExtractors(final String inputId) {
+    @Subscribe
+    public void handleInputCreate(final InputCreated event) {
+        LOG.debug("Load extractors for input <{}>", event.id());
+        scheduler.schedule(() -> loadExtractors(event.id()), 0, TimeUnit.SECONDS);
+    }
+
+    @Subscribe
+    public void handleInputDelete(final InputDeleted event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -70,34 +83,43 @@ public boolean filter(Message msg) {
         return false;
     }
 
-    private List<Extractor> loadExtractors(final String inputId) {
+    @Subscribe
+    public void handleInputCreate(final InputCreated event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -16,39 +16,52 @@
...
     private static final String NAME = "Extractor";
+    private static final List<Extractor> EMPTY_LIST = ImmutableList.of();
 
-    private Cache<String, List<Extractor>> cache = CacheBuilder.newBuilder()
-            .expireAfterWrite(1, TimeUnit.SECONDS)
-            .build();
+    private final ConcurrentMap<String, List<Extractor>> extractors = new ConcurrentHashMap<>();
 
     private final InputService inputService;
+    private final ScheduledExecutorService scheduler;
```
joschi: Why are we using a `ScheduledExecutorService` here instead of a normal `ExecutorService`? We're always starting the jobs right away (i. e. without any delay).

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -70,34 +83,43 @@ public boolean filter(Message msg) {
...
-        } catch (ExecutionException e) {
-            LOG.error("Could not load extractors into cache. Returning empty list.", e);
-            return Collections.emptyList();
+            inputService.all().forEach(input -> loadExtractors(input.getId()));
+        } catch (Exception e) {
+            LOG.error("Unable to load extractors for all inputs", e);
+        }
+    }
+
+    private void loadExtractors(final String inputId) {
```
joschi: Alternative suggestion (without having a intermediary list):
```java
    private void loadExtractors(final String inputId) {
        LOG.debug("Re-loading extractors for input <{}>", inputId);

        try {
            final Input input = inputService.find(inputId);
            final List<Extractor> sortedExtractors = inputService.getExtractors(input)
                    .stream()
                    .sorted((e1, e2) -> e1.getOrder().intValue() - e2.getOrder().intValue())

            extractors.put(inputId, sortedExtractors);
        } catch (NotFoundException e) {
            LOG.warn("Unable to load input <{}>: {}", inputId, e.getMessage());
        }
    }
```

joschi: We could also think about `InputServiceImpl#getExtractors()` returning a sorted list in the first place.

graylog2-server/src/main/java/org/graylog2/filters/RulesFilter.java
```diff
@@ -16,90 +16,79 @@
...
         // Always run the rules engine to make sure rules from the external rules file will be run.
-        privateSession.evaluate(msg, true);
+        privateSession.get().evaluate(msg, true);
 
         // false if not explicitly set to true in the rules.
         return msg.getFilterOut();
     }
 
+    @Subscribe
+    public void handleRulesUpdate(FilterDescriptionUpdateEvent ignored) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/RulesFilter.java
```diff
@@ -16,90 +16,79 @@
...
-        privateSession = rulesEngine.createPrivateSession();
+        loadRules();
+
+        // TODO: This class needs lifecycle management to avoid leaking objects in the EventBus
+        serverEventBus.register(this);
     }
 
     @Override
     public boolean filter(Message msg) {
-        final Set<FilterDescription> filters;
```
joschi: ❤️  the much simpler implementation!

graylog2-server/src/main/java/org/graylog2/filters/RulesFilter.java
```diff
@@ -16,90 +16,79 @@
...
  */
 public class RulesFilter implements MessageFilter {
     private static final Logger LOG = LoggerFactory.getLogger(RulesFilter.class);
 
+    private final RulesEngine rulesEngine;
     private final FilterService filterService;
-    private final RulesEngine.RulesSession privateSession;
-    private final Cache<String, Set<FilterDescription>> cache;
-    private Set<FilterDescription> currentFilterSet;
+    private final ScheduledExecutorService scheduler;
```
joschi: Same as before. This could (should?) be a normal `ExecutorService`.

graylog2-server/src/main/java/org/graylog2/filters/StaticFieldFilter.java
```diff
@@ -43,23 +48,32 @@
 
     private static final String NAME = "Static field appender";
 
-    private final Cache<String, List<Map.Entry<String, String>>> cache = CacheBuilder.newBuilder()
-            .expireAfterWrite(1, TimeUnit.SECONDS)
-            .build();
+    private static final List<Map.Entry<String, String>> EMPTY = ImmutableList.of();
```
joschi: `ImmutableList.of()` already returns a constant. No need to define another one for redirection. 😉 
Alternatively use `Collections.emptyList()` if you want to make the intention totally clear.

graylog2-server/src/main/java/org/graylog2/filters/StaticFieldFilter.java
```diff
@@ -43,23 +48,32 @@
...
+
+    private final ConcurrentMap<String, List<Map.Entry<String, String>>> staticFields = new ConcurrentHashMap<>();
 
     private final InputService inputService;
+    private final ScheduledExecutorService scheduler;
```
joschi: Same as before. This could (should?) be a normal `ExecutorService`.

graylog2-server/src/main/java/org/graylog2/filters/StaticFieldFilter.java
```diff
@@ -70,24 +84,38 @@ public boolean filter(Message msg) {
...
 
-    private List<Map.Entry<String, String>> loadStaticFields(final String inputId) {
+    @Subscribe
+    public void handleInputCreate(final InputCreated event) {
+        LOG.debug("Load static fields for input <{}>", event.id());
+        scheduler.schedule(() -> loadStaticFields(event.id()), 0, TimeUnit.SECONDS);
+    }
+
+    @Subscribe
+    public void handleInputDelete(final InputDeleted event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/StaticFieldFilter.java
```diff
@@ -70,24 +84,38 @@ public boolean filter(Message msg) {
         return false;
     }
 
-    private List<Map.Entry<String, String>> loadStaticFields(final String inputId) {
+    @Subscribe
+    public void handleInputCreate(final InputCreated event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/filters/StaticFieldFilter.java
```diff
@@ -70,24 +84,38 @@ public boolean filter(Message msg) {
...
+    }
+
+    @Subscribe
+    public void handleInputDelete(final InputDeleted event) {
+        LOG.debug("Removing input from static fields cache <{}>", event.id());
+        staticFields.remove(event.id());
+    }
+
+    @Subscribe
+    public void handleInputUpdate(final InputUpdated event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

graylog2-server/src/main/java/org/graylog2/streams/StreamRouter.java
```diff
@@ -39,24 +42,33 @@
...
+
+        // TODO: This class needs lifecycle management to avoid leaking objects in the EventBus
+        serverEventBus.register(this);
+    }
 
-        final StreamRouterEngineUpdater streamRouterEngineUpdater = new StreamRouterEngineUpdater(routerEngine, routerEngineFactory, streamService, executorService());
-        this.routerEngine.set(streamRouterEngineUpdater.getNewEngine());
-        scheduler.scheduleAtFixedRate(streamRouterEngineUpdater, 0, ENGINE_UPDATE_INTERVAL, TimeUnit.SECONDS);
+    @Subscribe
+    public void handleStreamsUpdate(StreamsChangedEvent event) {
```
joschi: Adding `@SuppressWarnings("unused")` would disable the compiler warnings (and make the IDE happy).

joschi:
Any chance for having tests for the event handlers (methods annotated with `@Subscribe`)?

graylog2-server/src/main/java/org/graylog2/streams/events/StreamsChangedEvent.java
bernd: I will change this instance because there is only one parameter, true. But in general I will keep using the builder pattern for this because the regular factory methods become really hard to work with once you have some more parameters.

And as said earlier, the builder pattern makes it easier to build up immutable collections. The factory pattern needs explicit conversion of collections to immutable collections to avoid having mutable collections in the value objects. I think we have some AutoValue instances in our code where this is broken.

bernd:
> @bernd What's the worst case if a node is missing one or more of those events?
> ClusterEventService doesn't strictly guarantee that nodes eventually process the events (or process them twice).

Worst case is that the node that didn't receive the message keeps running with the old configuration.

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
bernd: I don't have any warnings in my IDE and also didn't see any warnings during the build.

image

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
bernd: Because we do not have a regular `ExecutorService` pool. I wan to use the existing `daemonScheduler` pool.

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
joschi: Maybe some check in my IntelliJ IDEA configuration or in newer versions of IDEA?

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
joschi: Ah, got it.

In this case, we could simply use [`ExecutorService#submit(Runnable)`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html#submit-java.lang.Runnable-) instead of `ScheduledExecutorService#schedule(Runnable, long, TimeUnit)` in those places to get rid of the useless 0s delay.

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
bernd: I went with your implementation.

bernd:
>May i suggest you to extend theses events to almost anything touching Mongo add/update/delete?
It will helps a lot to write any plugin based on `@Subscribe` events.

@pdepaepe I actually tried to do that for this PR but the problem is that we use several different ways to access MongoDB so this is currently not really feasible.

bernd:
> Any chance for having tests for the event handlers (methods annotated with @Subscribe)?

I could do that, but I would rather spend my time on fixing more issues for 2.1 than writing tests to check if a job is submitted to the scheduler. :wink:

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
bernd: Done.

graylog2-server/src/main/java/org/graylog2/filters/RulesFilter.java
bernd: See above.

joschi:
> Worst case is that the node that didn't receive the message keeps running with the old configuration.

The problem I see with this is, that the configuration won't eventually converge to the correct one but would require a restart of the out-of-sync Graylog node or another change of the same type of entity.

Would it make sense to trigger the reloads periodically (e. g. every minute or so)? I'm well aware that we wanted to get rid of this, but I don't think that `ClusterEventService` is robust enough for this (without people actively checking for configuration drift in their Graylog cluster).

bernd:
> Would it make sense to trigger the reloads periodically (e. g. every minute or so)? I'm well aware that we wanted to get rid of this, but I don't think that ClusterEventService is robust enough for this (without people actively checking for configuration drift in their Graylog cluster).

Then we already have a problem because we use this pattern in several places. (at least pipeline, map widget and probably more)

joschi:
With the increased cleanup interval in #2509, this should be acceptable.

LGTM. 👍

PR #2497, opened 2016-07-18
Adds basic garbage collection statistics to node info

### LINKED ISSUE #154: Monitor server health and display on health page
Folks often have problems sizing the heap sizes of both ES and graylog2.
Help them make better decisions by monitoring heap size usage and GC times and add simply warnings on the health page.
Also expose this via the standard metrics endpoints.

Possibly extend this to ES cluster health monitoring as well, such as disk usage etc. Sigar is already part of ES so we can easily get to the data.

mikkolehtisalo:
Closes #154 by adding garbage collection counter and spent time to node information.

feat

The implementation aggregates information server side for the following reasons:

- The GC MXBean reports typically several garbage collectors that have to be queried separately
- The previous leads to Metrics like jvm.gc.ParNew.* and jvm.gc.ConcurrentMarkSweep.*
- Nodes are not guaranteed to use similar garbage collectors
- Aggregation server side is probably slightly less complex, and may be beneficial elsewhere

Lumped the client side implementation to JVM heap information, because it is closely related.

It should be noted that for detailed analysis enabling verbosegc is still probably the best tool available. Information obtainable from the GC MXBean may alert only to certain simple issues ("why is my GC running continuously?", or "why is my garbage collection time climbing up that fast?")

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/MetricRegistryProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/MetricRegistryProvider.java
@@ -22,6 +22,7 @@ import com.codahale.metrics.jvm.ClassLoadingGaugeSet;
 import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
+import org.graylog2.shared.utilities.AggregatedGarbageCollectorMetricSet;
 
 import javax.inject.Provider;
 import javax.inject.Singleton;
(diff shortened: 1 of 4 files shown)


joschi:
@mikkolehtisalo What information could you possibly retrieve from these GC stats?

The GC times are mostly interesting in a historic context, e. g. "this GC run took way longer than any before" or "the GC times are stable within the last hour". The simple summation of the GC times and number of GC runs is not very useful as you don't see why a GC run took long (i. e. which GC phase) and also not which GC run took long.

mikkolehtisalo:
@joschi User might be able to detect continually running GC (rare JVM related bug), but that's probably about it. 

The Metrics REST API is also missing some of the most important statistics, and the typical JVM tools (verbosegc, jstat and alike) may be required in any case, if one is investigating GC related issues. It is impossible to get all information directly from inside JVM, using any available API. 

I doodled the patch together, because if the feature is not needed or wanted, there should be no reason to keep #154 open either!

edmundoa:
Hi @mikkolehtisalo,

After some consideration, we think GC duration and runs are not useful enough, and will not merge this PR. As you mentioned, there is no reason to keep #154 open either, so I'll also close it.

In any case, thank you very much for taking the time to send us (yet another) PR!

PR #2499, opened 2016-07-19
Add ordering of decorators and decoration stats to UI

dennisoelkers:
* adds ordering of decorators (by using drag and drop) to the decorator sidebar.
* shows in message details which fields are decorated, allows to toggle displaying original content, disabled field actions for added fields.

Requires merging of #2482.


kroepke:
Overall this looks really good, but there are some rough edges:

* Synthetic fields in the sidebar are not marked as such, so the quickvalues etc dropdown is shown, but those functions don't work right now
* Drag'n'drop reordering sometimes applies too early, leading to jumbled results:
image
* creating/deleting decorators does not re-render the UI immediately right now

dennisoelkers:
> * creating/deleting decorators does not re-render the UI immediately right now

This seems to be an issue with a stale build. I just tested locally and it works fine.

kroepke:
Looks good to me, the only thing that is breaking right now is the following, but I think we can fix this in a subsequent bug fix. If you agree, I'd merge this.

1. Add a pipeline decorator that adds a simple static field to messages.
2. Select the field to show in the message table.
3. Remove the decorator from the result.
4. The field is still show as a table column.

kroepke:
Agreed to fix the remaining issue offline, I'll create a ticket.

PR #2500, opened 2016-07-19
Add option to only use named captures in Grok extractor

### LINKED ISSUE #1486: Grok extractor: Allow returning only named captures
The Grok extractor currently returns all matches for a grok pattern, e. g. also alternatives for the chosen pattern.
Graylog 1.2.x Grok extractor

It should be possible to only return named captures with the Grok extractor, similar to the [Logstash Grok filter](https://www.elastic.co/guide/en/logstash/1.5/plugins-filters-grok.html#plugins-filters-grok-named_captures_only) and the [Grok Debugger](https://grokdebug.herokuapp.com/).
Grok Debugger

bernd:
Update to the latest version of our java-grok port which adds support for this via upstream.

The option is disabled by default.

Fixes #1486 

Example screenshots:

screenshot from 2016-07-19 16 12 54
screenshot from 2016-07-19 16 12 36

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/extractors/GrokExtractor.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/extractors/GrokExtractor.java
@@ -33,8 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-
 public class GrokExtractor extends Extractor {
     private static final Logger log = LoggerFactory.getLogger(GrokExtractor.class);
 
@@ -70,7 +68,7 @@ public class GrokExtractor extends Extractor {
             throw new ConfigurationException("grok_pattern not set");
         }
 
-        final boolean namedCapturesOnly = firstNonNull((Boolean) extractorConfig.get("named_captures_only"), false);
+        final boolean namedCapturesOnly = (boolean) extractorConfig.getOrDefault("named_captures_only", false);
 
         try {
             // TODO we should really share this somehow, but unfortunately the extractors are reloaded every second.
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -641,7 +641,13 @@
             <dependency>
                 <groupId>org.graylog2.repackaged</groupId>
                 <artifactId>grok</artifactId>
-                <version>0.1.3-graylog</version>
+                <version>0.1.6-graylog</version>
+            </dependency>
+            <dependency>
+                <groupId>com.github.tony19</groupId>
+                <artifactId>named-regexp</artifactId>
```
joschi: What is this dependency used for? It's also already a transitive dependency of `org.graylog2.repackaged:grok:0.1.6-graylog`.

pom.xml
```diff
@@ -641,7 +641,13 @@
...
+                <version>0.2.3</version>
+                <scope>compile</scope>
```
joschi: `compile` is the default scope for dependencies.

graylog2-server/src/main/java/org/graylog2/inputs/extractors/GrokExtractor.java
```diff
@@ -68,13 +70,15 @@ public GrokExtractor(MetricRegistry metricRegistry,
             throw new ConfigurationException("grok_pattern not set");
         }
 
+        final boolean namedCapturesOnly = firstNonNull((Boolean) extractorConfig.get("named_captures_only"), false);
```
joschi: Aesthetic nitpick but not necessary to change: 
```java
final boolean namedCapturesOnly = (boolean) extractorConfig.getOrDefault("named_captures_only", false);
```

pom.xml
bernd: The grok extractor test didn't run at one point without this. But I guess that was some local issue. Works now.

joschi:
LGTM. 👍

Issue #2454, opened 2016-07-05
EC keys seem to be unsupported


adam-bishop:
I've defined a secure TCP input using the UI, and tried to use an ECDSA keypair. This fails due to the exception below.

I've had a quick look at the code (```transports/util/KeyUtil.java```).

From my reading of the code, it seems a keyspec is correctly created (```IllegalArgumentException``` isn't thrown), but because the keyfactory has 'RSA' hard coded (and it's an ECDSA key) ```generatePrivate()``` fails.

I guess the fix would be to to replace the hardcoded ```'RSA'``` with a section to detect the algorithm in use (```getAlgName()``` like in ```createKeySpec()```?)

### Steps to reproduce the problem

* Load an ECDSA keypair onto the system
* Create an input with a secure TCP transport using these keys
* Try to connect
* Exceptions logged to console

### Environment

* Graylog Version: 2.0.3-1

```
java.security.spec.InvalidKeySpecException: java.security.InvalidKeyException: Invalid RSA private key
	at sun.security.rsa.RSAKeyFactory.engineGeneratePrivate(RSAKeyFactory.java:217) ~[?:1.8.0_91]
	at java.security.KeyFactory.generatePrivate(KeyFactory.java:372) ~[?:1.8.0_91]
	at org.graylog2.plugin.inputs.transports.util.KeyUtil.loadPrivateKey(KeyUtil.java:153) ~[graylog.jar:?]
	at org.graylog2.plugin.inputs.transports.util.KeyUtil.initKeyStore(KeyUtil.java:111) ~[graylog.jar:?]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$1.createSslEngine(AbstractTcpTransport.java:205) ~[graylog.jar:?]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$1.call(AbstractTcpTransport.java:186) ~[graylog.jar:?]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$1.call(AbstractTcpTransport.java:182) ~[graylog.jar:?]
	at org.graylog2.plugin.inputs.transports.NettyTransport$1.getPipeline(NettyTransport.java:110) ~[graylog.jar:?]
	at org.jboss.netty.channel.socket.nio.NioServerBoss.registerAcceptedChannel(NioServerBoss.java:134) [graylog.jar:?]
[...]

PR #2503, opened 2016-07-19
Attempt reading DSA key if RSA failed

mikkolehtisalo:
Simple fix ~~sans tests~~ for #2454. Feel free to reject and close, or refine further! More explanation at #2454.


graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/util/KeyUtil.java
```diff
@@ -149,8 +149,14 @@ protected static PrivateKey loadPrivateKey(File file, String password) throws IO
                 throw new IllegalArgumentException("Unsupported key type");
             }
 
-            final KeyFactory kf = KeyFactory.getInstance("RSA");
-            return kf.generatePrivate(keySpec);
+            try {
+                final KeyFactory RSAkf = KeyFactory.getInstance("RSA");
+                return RSAkf.generatePrivate(keySpec);
+            } catch (InvalidKeySpecException e) {
+                LOG.debug("Generating RSA private key failed", e);
```
lennartkoopmann: IMO this should say that we are trying to generate a DSA key instead and not all hope is lost yet.

mikkolehtisalo: Good point. Rephrased. I will take a look at the tests next :P

joschi:
@mikkolehtisalo Thank you very much!

Merged in 97521bf2bd8170101173aa6c39a1cf025e6ab1e8 (after updating the log message).

PR #2508, opened 2016-07-20
Make it possible for plugins to request a shared class loader

### LINKED ISSUE #2436: Improve plugin classloader visiblity
Continuing the discussion from https://github.com/Graylog2/graylog2-server/pull/2280#issuecomment-229347328

Since 2.0.2 Graylog uses a single classloader to load all plugins. This is necessary so that plugins can extend/use other plugins.

The downside of this setup is that conflicting dependency versions now clash, even if a given plugin does not need access to other plugins.
One possible solution could be to let plugins declare their classloader isolation preference, with defaulting to "fully isolated".
Plugins which do need access to other plugins could then declare that preference and thus extend them.

This issue exists so we can discuss possible approaches and make a decision how to proceed (or to come up with workarounds for affected plugin authors)

bernd:
In #2280 we changed the plugin loader so that all plugins share one class loader and they can see each other. (needed for plugin inter-dependencies)

This is problematic when different plugins have conflicting dependencies. (see #2436)

With this change, every plugin gets its own class loader by default so we can avoid dependency clashes. Plugins which depend on other plugins can request a shared class loader via a graylog-plugin.properties file.

Fixes #2436

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
@@ -31,6 +31,7 @@ import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/shared/plugins/PluginLoader.java
```diff
@@ -99,9 +100,24 @@ public PluginLoader(File pluginDir, ChainingClassLoader classLoader) {
...
+                LOG.debug("Create isolated class loader for {}", url);
+                classLoader.addClassLoader(URLClassLoader.newInstance(new URL[]{url}));
+            } else {
+                LOG.debug("Use shared class loader for {}", url);
+                sharedClassLoaderUrls.add(url);
+            }
+        });
+
+        LOG.debug("Create shared class loader for {} plugins: {}", sharedClassLoaderUrls.size(), sharedClassLoaderUrls);
+        classLoader.addClassLoader(URLClassLoader.newInstance(sharedClassLoaderUrls.toArray(new URL[sharedClassLoaderUrls.size()])));
```
kroepke: Does this work properly when the array is empty?

bernd: Good point. If there are no plugins which request a shared class loader, we should not add an empty one.

kroepke:
lgtm

PR #2510, opened 2016-07-20
Always trim message field values on Message class

### LINKED ISSUE #1936: "match regular expression" stream rule type fails to properly validate message starting with a blank space
### Problem description

Whe you want to match the beginning of a syslog message that starts with a blank space against a "match regular expression" stream rule, you must not include the leading blank space in your regular expression if you want the UI to tell you the message will match. Though, you must include it if you want the StreamMatcherFilter to effectively route your message to your stream.

1. Have a system sending your graylog server a syslog message starting with an empty space (eg: " Login incorrect: \[root\] (from client BLABLABLA port 0 cli a.b.c.d)\n";
2. Set up the following stream rule to match the beginning of the message: "^Login incorrect.*";
3. Confirm that the UI tells you the message from step 1 will match;
4. Confirm that no messages get routed to the stream;
5. Set up the following stream rule to match the beginning of the message: "^.Login incorrect.*";
6. Confirm that the UI tells you the message from step 1 wont't match;
7. Confirm that messages get routed to the stream.

### Sample data 

Here are sample data to reproduce the problem:

#### Message

```json
{
  "message": {
    "message": " Login incorrect: [root] (from client BLABLABLA port 0 cli a.b.c.d)\n",
    "timestamp": "2016-03-16T09:27:44.249Z",
    "process_id": "22562",
    "level": 5,
    "application_name": "freeradius",
    "source": "louise",
    "facility": "local5",
    "_id": "50b39030-eb59-11e5-a255-08002703c127",
    "gl2_source_input": "56e13cd1e4b0deefe9f54cf8",
    "gl2_source_node": "3aee6e01-9fdb-43d4-83b7-324a58e265d6",
    "streams": [],
    "gl2_remote_port": 41948,
    "gl2_remote_ip": "w.x.y.z"
  },
  "index": "graylog2_0"
}
```

#### Rules

```json
 "stream_rules": [
    {
      "id": "56e91f0de4b0fcf9693e796a",
      "field": "application_name",
      "value": "freeradius",
      "stream_id": "56e91ebae4b0fcf9693e790f",
      "type": 1,
      "inverted": false
    },
    {
      "id": "56e91f2fe4b0fcf9693e798e",
      "field": "message",
      "value": "^Login incorrect.*",
      "stream_id": "56e91ebae4b0fcf9693e790f",
      "type": 2,
      "inverted": false
    }
  ]
```

### Environment

* Graylog Version: 1.3
* Elasticsearch Version: 1.7.5
* MongoDB Version: 3.2.4
* Browser version: Firefox 44.0.2

### IRC Logs

```
10:46:19 - StCyr: mmm, created a new stream, restarted the whole server, still not having my messages routed to the stream though the "Rules Editor UI" say they match the rules?!?
10:46:45 - StCyr: Here's a debug log after pausing-starting the stream => http://pastebin.com/40uCCexf
10:47:12 - StCyr: Message 50b39030-eb59-11e5-a255-08002703c127 matches the rule in the "Edit Rules" UI
10:54:37 - joschi: StCyr: could you post some example messages and the stream rules?
10:54:51 - joschi: StCyr: you don't need to restart Graylog when creating new streams or changing stream rules
10:54:58 - StCyr: pastebin?
10:58:35 - StCyr: message: http://pastebin.com/NsEwkzJG
10:59:01 - StCyr: rules: http://pastebin.com/UC0mGNg7
11:00:35 - StCyr: there's always this blank space at the start of all my messages, but "Edit Rules" UI doesn't mind
11:02:17 - StCyr: joschi: maybe I should try to add this blank space in my regular expression though "Edit Rules" UI doesn't mind?
11:05:53 - StCyr: the thing is: if I add this blank space in my regex the "Edit Rules" UI say message doesn't match O.o
12:09:27 - joschi: StCyr: the leading blank is definitely a problem. please create an issue for that problem (the web interface saying the rule matched while it didn't) at https://github.com/Graylog2/graylog2-server/issues
```

edmundoa:
Before this change, one of the public constructors of `Message` was using `addField()` to add fields in the `Message` object, while the other was adding the fields by hand to the internal `Map` used to store message fields.

That created an inconsistency, as the constructor using `addField()` trims all `String` field values by default, unlike the other constructor. It lead to the issue mentioned in: #1936, and it also affected to extractors and drool rules. Basically, every filter used in the `MessageFilterChainProcessor`.

This commit ensures both constructors trim `String` values, while still ensuring `Message(final String message, final String source, final DateTime timestamp)` will create a `Message` object containing the required fields given as arguments.

Making this change will break stream rules, extractors, and drool rules checking for one or more leading or trailing whitespaces in message fields.

I am still working on some documentation around this PR, that's why it is missing the `ready-to-review` tag, but feel free to add any concerns or comments 😄 

Fixes #1936.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -129,7 +129,6 @@ public class Message implements Messages {
     private ArrayList<Recording> recordings;
 
     public Message(final String message, final String source, final DateTime timestamp) {
-        // Adding the fields directly because they would not be accepted as a reserved fields.
         fields.put(FIELD_ID, new UUID().toString());
         addRequiredField(FIELD_MESSAGE, message);
         addRequiredField(FIELD_SOURCE, source);
@@ -304,24 +303,26 @@ public class Message implements Messages {
     }
 
     private void addField(final String key, final Object value, final boolean isRequiredField) {
+        final String trimmedKey = key.trim();
+
         // Don't accept protected keys. (some are allowed though lol)
-        if (RESERVED_FIELDS.contains(key) && !RESERVED_SETTABLE_FIELDS.contains(key) || !validKey(key)) {
+        if (RESERVED_FIELDS.contains(trimmedKey) && !RESERVED_SETTABLE_FIELDS.contains(trimmedKey) || !validKey(trimmedKey)) {
             if (LOG.isDebugEnabled()) {
-                LOG.debug("Ignoring invalid or reserved key {} for message {}", key, getId());
+                LOG.debug("Ignoring invalid or reserved key {} for message {}", trimmedKey, getId());
             }
             return;
         }
 
-        if (FIELD_TIMESTAMP.equals(key.trim()) && value != null && value instanceof Date) {
+        if (FIELD_TIMESTAMP.equals(trimmedKey) && value != null && value instanceof Date) {
             fields.put(FIELD_TIMESTAMP, new DateTime(value));
-        } else if(value instanceof String) {
+        } else if (value instanceof String) {
             final String str = ((String) value).trim();
 
             if (isRequiredField || !str.isEmpty()) {
-                fields.put(key.trim(), str);
+                fields.put(trimmedKey, str);
             }
-        } else if(value != null) {
-            fields.put(key.trim(), value);
+        } else if (value != null) {
+            fields.put(trimmedKey, value);
         }
     }
 
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -131,9 +131,9 @@ public class Message implements Messages {
     public Message(final String message, final String source, final DateTime timestamp) {
         // Adding the fields directly because they would not be accepted as a reserved fields.
         fields.put(FIELD_ID, new UUID().toString());
-        fields.put(FIELD_MESSAGE, message);
-        fields.put(FIELD_SOURCE, source);
-        fields.put(FIELD_TIMESTAMP, timestamp);
+        addRequiredField(FIELD_MESSAGE, message);
+        addRequiredField(FIELD_SOURCE, source);
+        addRequiredField(FIELD_TIMESTAMP, timestamp);
     }
 
     public Message(final Map<String, Object> fields) {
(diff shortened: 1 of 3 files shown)


bernd:
Please add the upgrade notes to `UPGRADING.rst` once you have them. :smiley:

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -309,7 +317,7 @@ public void addField(final String key, final Object value) {
         } else if(value instanceof String) {
             final String str = ((String) value).trim();
 
-            if(!str.isEmpty()) {
+            if (isRequiredField || !str.isEmpty()) {
```
joschi: Why would we handle "required" fields different from "normal" fields in this case? Wouldn't this introduce another inconsistency?

joschi: Also, why should we trim empty "required" fields?

edmundoa: Yes, you are right, it adds another consistency to keep the old behaviour, that is, messages created with the `Message(final String message, final String source, final DateTime timestamp)` constructor will contain at least  the `message`, and `source` fields, even if their values are empty (or spaces).

I added the special case to fix some tests, but maybe it doesn't make sense. How do you think we should handle the case when a message is created with empty `message` or `source` fields?

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -296,6 +296,14 @@ public void setSource(final String source) {
...
+        addField(key, value, false);
+    }
+
+    private void addRequiredField(final String key, final Object value) {
+        addField(key, value, true);
+    }
+
+    private void addField(final String key, final Object value, final boolean isRequiredField) {
         // Don't accept protected keys. (some are allowed though lol)
         if (RESERVED_FIELDS.contains(key) && !RESERVED_SETTABLE_FIELDS.contains(key) || !validKey(key)) {
```
dennisoelkers: While we are at it: why is the trimmed key used further down, while it is used non-trimmed here? Shouldn't we fix that to make it consistent?

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
dennisoelkers: I think we should still trim them to make the behavior consistent, we just should not let them be null (to prevent another breaking change and forcing all upstream producers to check for this, we could use `Strings.nullToEmpty` instead of rejecting the message construction by throwing an exception).

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -131,9 +131,9 @@
     public Message(final String message, final String source, final DateTime timestamp) {
         // Adding the fields directly because they would not be accepted as a reserved fields.
```
dennisoelkers: This comment is now out of sync with the following code.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
edmundoa: That's a good question, and I don't have the answer for it. Honestly, I didn't want to touch more than I had to, to fix this issue. I'll take a look.

joschi:
LGTM. 👍

PR #2515, opened 2016-07-22
Allowing to run REST API and web interface on same port.

dennisoelkers:
This loads the web interface resources in the REST API service if the `rest_listen_uri` and `web_listen_uri` configuration directives contain the same host and port. If this is the case, one of them also must have a different path prefix, the web interface path must be nonempty and different than `/`.

Additionally, requests going to `http://<graylog>:12900` expecting a `text/html` response are redirected to `web_listen_uri` if REST and web are running on the same port. The behavior stays the same for JSON requests.


graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
```diff
@@ -428,6 +443,32 @@ public void validateWebTlsConfig() throws ValidationException {
         }
     }
 
+    @ValidatorMethod
+    @SuppressWarnings("unused")
+    public void validateRestAndWebListenConfigConflict() throws ValidationException {
```
joschi: +1 for different validator methods. ❤️

dennisoelkers: +1 for implementing them in the first place <3

joschi:
LGTM. 👍

PR #2522, opened 2016-07-25
improve styling and wording a bit for authentication providers

kroepke:
some of the pages will probably get actual content for 2.1.0-final but for now a placeholder text that looks nicer is good enough


edmundoa:
LGTM 👍

PR #2523, opened 2016-07-25
Introduce CombinedProvider to sync actions and stores initialization.

dennisoelkers:
When using separate providers for actions and stores, there is the
possibility that only actions need to be imported in a component. Using
the actions then would have no effect though, if the corresponding store
listening to those actions was not initialized already. So there are
cases where the store is imported in components although it is unused.

This PR introduced the CombinedProvider which syncs actions and store
initializations. It can be used to import stores and actions separately
or combined in a component. Usage of it is shown in the SearchPage
component, which is shorted and more concise due to it. It can be used
as a blueprint for future conversions to using the CombinedProvider, but
usage of Actions/StoreProvider is still possible and benefits from
syncing the imports.

### DIFF
--- a/graylog2-web-interface/src/injection/ActionsProvider.jsx
+++ b/graylog2-web-interface/src/injection/ActionsProvider.jsx
@@ -1,54 +1,14 @@
+import CombinedProvider from 'injection/CombinedProvider';
+
 /* global actionsProvider */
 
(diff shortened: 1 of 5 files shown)


edmundoa:
LGTM 👍

PR #2525, opened 2016-07-25
Changing default to make REST API and web if to listen on same host/port

### LINKED ISSUE #2446: Allow serving web interface and REST API on the same port
Now that #2440 is merged, we can allow serving the web interface and REST API from the same port.

The suggestion is to allow setting the `web_listen_uri` to the same host/port combination as the `rest_listen_uri`.
If that's the case the `web_listen_uri` needs to verify that a path different than `/` is set.
The resources of the web interface are then added to the same API service that serves the REST endpoints.

Otherwise a second API service is started, which functions exactly as it does right now (i.e. separate port is bound, the two API services are different).

The current behavior should be the default to avoid surprises for current 2.0 users.

dennisoelkers:
This PR changes the defaults of the configuration to make the web interface use the same listener as the default REST API.

Needs to be merged after #2515 

Fixes #2446

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -22,15 +22,17 @@ Due to those shortcomings, the feature has been removed completely. Users need t
 Web Interface Listener
 ----------------------
 
-2.0.x was defaulting to separate listeners for the REST API and the web interface. The former defaulted to ``http://localhost:12900``, the latter to ``http://localhost:9000``.
-Beginning with 2.1.0 it is possible to run both the REST API and the web interface on the same host/port-combination and this is now the default. This means that the REST API is still running on ``http://localhost:12900`` per default, but the web interface is now running on ``http://localhost:12900/console``. Furthermore, all requests going to ``http://localhost:12900/`` requesting a content-type of ``text/html`` are redirected to the web interface, therefore making it even easier to set up Graylog and use it behind proxies, expose it externally, etc.
+Graylog 2.0.x has been using separate listeners for the REST API and the web interface by default. The Graylog REST API on ``http://127.0.0.1:12900``, the Graylog web interface on ``http://127.0.0.1:9000``.
+Beginning with Graylog 2.1.0 it is possible to run both the REST API and the web interface on the same host/port-combination and this is now the default. This means that the REST API is still running on ``http://127.0.0.1:12900`` per default, but the web interface is now running on ``http://127.0.0.1:12900/console``.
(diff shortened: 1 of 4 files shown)


bernd:
How about using `web` instead of `console`? The latter sounds weird to me.

joschi:
@bernd I don't have a strong opinion in this matter.
@dennisoelkers any comments?

kroepke:
I'd either go for `web` or simply `graylog` as the default.

dennisoelkers:
`/console` was chosen because it seemed to be the most appropriate semantically. `/web` seems to be like using `/foo` because it basically means nothing these days. I would still go with `/console` as it is only the default.

UPGRADING.rst
```diff
@@ -19,6 +19,21 @@ Previous versions of Graylog were automatically generating a private key/certifi
 Due to those shortcomings, the feature has been removed completely. Users need to use proper certificates or generate their own self-signed certificates and configure them with the appropriate settings, see `Using HTTPS <http://docs.graylog.org/en/2.0/pages/configuration/https.html>`_ for reference.
 
 
+Web Interface Listener
+----------------------
+
+Graylog 2.0.x has been using separate listeners for the REST API and the web interface by default. The Graylog REST API on ``http://127.0.0.1:12900``, the Graylog web interface on ``http://127.0.0.1:9000``.
+Beginning with Graylog 2.1.0 it is possible to run both the REST API and the web interface on the same host/port-combination and this is now the default. This means that the REST API is still running on ``http://127.0.0.1:12900`` per default, but the web interface is now running on ``http://127.0.0.1:12900/web``.
+Furthermore, all requests going to ``http://127.0.0.1:12900/`` requesting a content-type of ``text/html`` or ``application/xhtml+xml`` are redirected to the web interface, therefore making it even easier to set up Graylog and use it behind proxies, expose it externally etc.
```
shishkin: It's not the behavior I observe with docker 2.1.0-beta2-1. Web UI still starts on 9000 by default. Wen I overwrite `web_listen_uri` with http://0.0.0.0:12900/web the UI fails by redirecting to that literal 0.0.0.0 IP and `web_transport_uri` has no effect.

joschi: @shishkin Please create a new issue for this (if there is none yet covering the same problem).

Comments on PR commits can easily be overlooked and badly be tracked, especially if the PR is already closed.

shishkin: Sure, here it is: https://github.com/Graylog2/graylog2-server/issues/2606

PR #2527, opened 2016-07-25
Fix search results console warnings

edmundoa:
Expanding the message details from a message in the search result may lead to two console warnings:

- `Failed propType: Invalid prop 'value' of type 'number' supplied to 'MessageField', expected 'string'. Check the render method of 'MessageFields'`: This one is triggered when one of the field values of the message is not a string. The PR fixes it by converting it into a string.

- `Each child in an array or iterator should have a unique "key" prop. Check the render method of 'MessageFields'`: There were some arrays of React nodes that did not provide unique keys, so I added them.

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageFields.jsx
+++ b/graylog2-web-interface/src/components/search/MessageFields.jsx
@@ -19,7 +19,7 @@ const MessageFields = React.createClass({
 
   _formatFields(fields, showDecoration) {
     if (!showDecoration || !this.props.message.decoration_stats) {
-      return Object.keys(fields).sort().map(key => <MessageField {...this.props} fieldName={key} value={fields[key]} />);
+      return Object.keys(fields).sort().map(key => <MessageField {...this.props} fieldName={key} value={String(fields[key])} />);
     }
 
     const decorationStats = this.props.message.decoration_stats;
@@ -41,7 +41,7 @@ const MessageFields = React.createClass({
                                     originalValue={decorationStats.removed_fields[key]} />);
       }
 
-      return <MessageField {...this.props} fieldName={key} value={fields[key]} disableFieldActions={true} />;
+      return <MessageField {...this.props} fieldName={key} value={String(fields[key])} disableFieldActions={true} />;
     });
   },
   render() {
--- a/graylog2-web-interface/src/components/search/MessageFields.jsx
+++ b/graylog2-web-interface/src/components/search/MessageFields.jsx
@@ -19,7 +19,7 @@ const MessageFields = React.createClass({
 
   _formatFields(fields, showDecoration) {
     if (!showDecoration || !this.props.message.decoration_stats) {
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/search/MessageFields.jsx
```diff
@@ -19,7 +19,7 @@ const MessageFields = React.createClass({
 
   _formatFields(fields, showDecoration) {
     if (!showDecoration || !this.props.message.decoration_stats) {
-      return Object.keys(fields).sort().map(key => <MessageField {...this.props} fieldName={key} value={fields[key]} />);
+      return Object.keys(fields).sort().map(key => <MessageField key={key} {...this.props} fieldName={key} value={String(fields[key])} />);
```
dennisoelkers: Do we really want to convert it into a string? Wouldn't it make more sense to relax the prop type for the component? It might make sense to still have the original type in the component, so we could display numbers/whatever differently.

graylog2-web-interface/src/components/search/MessageFields.jsx
```diff
@@ -28,20 +28,22 @@ const MessageFields = React.createClass({
...
       if (decorationStats.removed_fields[key]) {
-        return (<ChangedMessageField fieldName={key}
-                                    originalValue={decorationStats.removed_fields[key]} />);
+        return (<ChangedMessageField key={key}
+                                     fieldName={key}
+                                     originalValue={decorationStats.removed_fields[key]} />);
       }
 
-      return <MessageField {...this.props} fieldName={key} value={fields[key]} disableFieldActions={true} />;
+      return <MessageField key={key} {...this.props} fieldName={key} value={String(fields[key])} disableFieldActions={true} />;
```
dennisoelkers: See above

graylog2-web-interface/src/components/search/MessageFields.jsx
edmundoa: The only problem I see with relaxing the prop, is that some types may not render properly, like booleans. As at the moment we don't do anything special with the value, I don't see any problem with converting it to a string to be honest.

dennisoelkers: We should still do the String conversion in the `MessageField` component, not externally.

edmundoa: I don't know why we should do it in `MessageField`, I guess the best place would be in `MessageFieldDescription`, which is the component rendering the field and should be the one knowing how to render the field value.

edmundoa: Looking at the code, `MessageFieldDescription` only uses the field value to add it to the search bar if that action is required. In the end `possiblyHighlight()` is converting and returning the value we want to display already, so I don't think we need to convert anything at this point. I will simply relax the prop as you suggested in the first place.

joschi:
LGTM. 👍

Issue #2045, opened 2016-04-07
LDAP Users are not created in 2.0 Beta


traviscosgrave:
### Problem description

After configuring LDAP authentication (using Active Directory) on 2.0.0-3.beta.2, All checks in the interface are successful, but users are not able to login. The important logs are at the end of the post.

After reviewing the trace and stepping through the code, an empty user is created in the `UserServiceImpl.syncFromLdapEntry` at [line 153](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/users/UserServiceImpl.java#L153). During `updateFromLdap` the user's permissions are checked at [line 195](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/users/UserServiceImpl.java#L195). The change to the `getPermissions` method from commit [7d087dc](https://github.com/Graylog2/graylog2-server/commit/7d087dc7152fe9060195657b3851c33bdd8f24eb#diff-4ffbbbd0f36b55d2314f253ba675e6ebL160) does not verify that `fields.get(PERMISSIONS)` exists and the `NullPointerException` is throw when `Sets.newHashSet` is called. The previous code returned `null` restulting in the `user.getPermissions() == null` condition being true and `user.setPermissions` being called on [line 196](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/users/UserServiceImpl.java#L196).

### Steps to reproduce the problem

1. Configure LDAP
1. Attempt to log in as a user which does not exist in MongoDB

### Environment

* Graylog Version: 2.0.0-3.beta.2
* Elasticsearch Version: 2.3.1
* MongoDB Version: 3.4
* Operating System: Ubuntu 14.04
* Browser version: Google Chrome 49.0.2623.110

### Log Messages

```
2016-04-07T10:08:11.684-04:00 DEBUG [ModularRealmAuthenticator] Realm [org.graylog2.security.realm.LdapUserAuthenticator@136982fb] threw an exception during a multi-realm authentication atte
java.lang.NullPointerException
        at com.google.common.collect.Sets.newHashSet(Sets.java:221) ~[graylog.jar:?]
        at org.graylog2.users.UserImpl.getPermissions(UserImpl.java:166) ~[graylog.jar:?]
        at org.graylog2.users.UserServiceImpl.updateFromLdap(UserServiceImpl.java:195) ~[graylog.jar:?]
[...]

PR #2529, opened 2016-07-25
do not blindly override permission set of ldap users

### LINKED ISSUE #2516: Missing setup streams and user dashboards after entering through the active directory
### Problem description

After obnovlenyai from 1.7 to 2.0 version, settings are lost after entering through Active Directory. So it should be, or am I doing something wrong? Previously, it was possible to adjust the stream and dashboards for each individual.

1. The user comes to the AD account
2. I add it streams and dashboards
3. After you log out and log in, setting the empty again

### Environment

* Graylog Version: 2.0
* Elasticsearch Version: 2.3.4
* MongoDB Version: 3.2.8
* Operating System: ubuntu 14.04
* Browser version: Chrome

kroepke:
when fixing #2045 we introduced a bug clearing the permission set for synced ldap accounts

this only shows up when not using roles but the permission sets directly

fixes #2516

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
@@ -48,7 +48,6 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
-import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
@@ -250,7 +249,8 @@ public class LdapUserAuthenticator extends AuthenticatingRealm {
             translatedRoleIds.addAll(user.getRoleIds());
         }
         user.setRoleIds(translatedRoleIds);
-        user.setPermissions(Collections.emptyList());
+        // preserve the raw permissions (the ones without the synthetic self-edit permissions or the "*" admin one)
+        user.setPermissions(user.getPermissions());
     }
 
 }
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
```diff
@@ -172,8 +173,9 @@ public void setEmail(final String email) {
 
     @Override
     public void setPermissions(final List<String> permissions) {
+        final List<String> perms = Lists.newArrayList(permissions);
         // Do not store the dynamic user self edit permissions
-        permissions.removeAll(this.permissions.userSelfEditPermissions(getName()));
+        perms.removeAll(this.permissions.userSelfEditPermissions(getName()));
```
joschi: I can has unit test for this?

kroepke: touché

graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
```diff
@@ -250,7 +249,8 @@ private void updateFromLdap(User user, LdapEntry userEntry, LdapSettings ldapSet
             translatedRoleIds.addAll(user.getRoleIds());
         }
         user.setRoleIds(translatedRoleIds);
-        user.setPermissions(Collections.emptyList());
+        // preserve the raw permissions (the ones without the synthetic self-edit permissions or the "*" admin one)
+        user.setPermissions(user.getPermissions());
```
joschi: I can has unit test for this?

joschi:
LGTM. 👍

angelsedoy:
I'm sorry, but this code does not work. Now permissions are not overwritten, but are overwritten role, in particular the role of flies admin.

joschi:
@angelsedoy Thanks for your feedback. Please create a new issue for this at https://github.com/Graylog2/graylog2-server/issues/new

PR #2535, opened 2016-07-26
Fix Version#fromClasspathProperties() when loading from JAR plugin

bernd:
Loading a `version.properties` file from a JAR plugin didn't work before
because `Resources.getResource()` didn't look into the plugin's class
loader.

This changes the main `#fromClasspathProperties()` method to accept a
Class parameter which will be used to lookup the resource in the correct
class loader.

It also adds `#fromPluginProperties()` a convenience method for plugins.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Version.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Version.java
@@ -152,7 +152,7 @@ public class Version implements Comparable<Version> {
     }
 
     /**
-     *  Try to read the version from the {@literal graylog-plugin.properties} file included in a plugin.
+     * Try to read the version from the {@literal graylog-plugin.properties} file included in a plugin.
      *
      * @param pluginClass     Class where the class loader should be obtained from.
      * @param path            Path of the properties file on the classpath which contains the version information.
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Version.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Version.java
@@ -16,7 +16,6 @@
  */
 package org.graylog2.plugin;
 
-import com.google.common.io.Resources;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -153,6 +152,18 @@ public class Version implements Comparable<Version> {
     }
 
     /**
+     *  Try to read the version from the {@literal graylog-plugin.properties} file included in a plugin.
+     *
+     * @param pluginClass     Class where the class loader should be obtained from.
+     * @param path            Path of the properties file on the classpath which contains the version information.
+     * @param propertyName    The name of the property to read as project version ("major.minor.patch-preReleaseVersion").
+     * @param defaultVersion  The {@link Version} to return if reading the information from the properties files failed.
+     */
+    public static Version fromPluginProperties(Class<?> pluginClass, String path, String propertyName, Version defaultVersion) {
+        return fromClasspathProperties(pluginClass, path, propertyName, null, null, defaultVersion);
+    }
+
+    /**
      * Try to read the version from {@literal version.properties} ({@code project.version} property)
      * and {@literal git.properties} ({@code git.commit.id} property) from the classpath.
      *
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/Version.java
```diff
@@ -217,6 +243,12 @@ public static Version fromClasspathProperties(String path, String propertyName,
         return defaultVersion;
     }
 
+    private static URL getResource(Class<?> clazz, String path) {
+        final URL url = requireNonNull(clazz, "Class argument is null!").getClassLoader().getResource(path);
```
joschi: I think the `getClassLoader()` call is not required here, as [`Class#getResource(String)`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getResource-java.lang.String-) already is using the object's class loader.

There's also a variant of [`Resources#getResource()`](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/io/Resources.html#getResource(java.lang.Class,%20java.lang.String)) for this.

bernd: > There's also a variant of Resources#getResource() for this.

I tried the `Resources#getResource()` variant but it didn't work... No idea why.

I will try again.

bernd: When I change the code like the following or use the `Resources#getResource(class, path)` method, I am getting the error below. It works fine with the current code in this PR.

```diff
diff --git graylog2-server/src/main/java/org/graylog2/plugin/Version.java graylog2-server/src/main/java/org/graylog2/plugin/Version.java
index 7f13e36..3a0be0a 100644
--- graylog2-server/src/main/java/org/graylog2/plugin/Version.java
+++ graylog2-server/src/main/java/org/graylog2/plugin/Version.java
@@ -244,7 +244,7 @@ public class Version implements Comparable<Version> {
     }
 
     private static URL getResource(Class<?> clazz, String path) {
-        final URL url = requireNonNull(clazz, "Class argument is null!").getClassLoader().getResource(path);
+        final URL url = requireNonNull(clazz, "Class argument is null!").getResource(path);
 
         return requireNonNull(url, "Resource <" + path + "> not found.");
     }
```

```
2016-07-26 17:03:11,877 ERROR: org.graylog2.plugin.Version - Unable to read version.properties, this build has no version number.
java.lang.NullPointerException: Resource <version.properties> not found.
	at java.util.Objects.requireNonNull(Objects.java:228) ~[?:1.8.0_101]
	at org.graylog2.plugin.Version.getResource(Version.java:249) ~[classes/:?]
	at org.graylog2.plugin.Version.fromClasspathProperties(Version.java:215) [classes/:?]
	at org.graylog2.plugin.Version.fromClasspathProperties(Version.java:199) [classes/:?]
	at org.graylog2.plugin.Version.fromClasspathProperties(Version.java:173) [classes/:?]
	at org.graylog2.plugin.Version.<clinit>(Version.java:66) [classes/:?]
	at org.graylog2.bootstrap.CmdLineTool.<clinit>(CmdLineTool.java:90) [classes/:?]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.8.0_101]
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) [?:1.8.0_101]
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) [?:1.8.0_101]
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423) [?:1.8.0_101]
	at io.airlift.airline.ParserUtil.createInstance(ParserUtil.java:19) [airline-0.7.jar:0.7]
	at io.airlift.airline.ParserUtil.createInstance(ParserUtil.java:37) [airline-0.7.jar:0.7]
	at io.airlift.airline.Cli.parse(Cli.java:120) [airline-0.7.jar:0.7]
	at io.airlift.airline.Cli.parse(Cli.java:97) [airline-0.7.jar:0.7]
	at org.graylog2.bootstrap.Main.main(Main.java:43) [classes/:?]
```

joschi: ✅ Alright, thanks for checking! Let's keep it the original way.

joschi:
LGTM. 👍

Issue #2536, opened 2016-07-26
graylog2 can not start!!!


nile:
graylog.conf
```
is_master = true
node_id_file = ~/node-id
password_secret = 0GfeFiddzix4IdYzTjT8PGSaYnTm5vYzaBk9QXyN12RhwSlIHHY1ewv3cIM4Cj5mgxRN9ImWeCqojc9hjvKDoS2ztrzY7igt
root_password_sha2 = 8c6976e5b5410415bde908bd4dee15dfb167a9c873fc4bb8a81f6f2ab448a918
root_timezone = Asia/Shanghai
plugin_dir = plugin
rest_listen_uri = http://192.168.99.100:12900/
web_listen_uri = http://192.168.99.100:9000/
rotation_strategy = count
elasticsearch_max_docs_per_index = 20000000
elasticsearch_max_number_of_indices = 20
retention_strategy = delete
elasticsearch_shards = 4
elasticsearch_replicas = 0
allow_leading_wildcard_searches = false
allow_highlighting = false
elasticsearch_cluster_name = elasticsearch
elasticsearch_discovery_zen_ping_unicast_hosts = 192.168.99.100:9300
elasticsearch_analyzer = standard
output_batch_size = 500
output_flush_interval = 1
output_fault_count_threshold = 5
output_fault_penalty_seconds = 30
processbuffer_processors = 5
outputbuffer_processors = 3
processor_wait_strategy = blocking
ring_size = 65536
inputbuffer_ring_size = 65536
inputbuffer_processors = 2
inputbuffer_wait_strategy = blocking
message_journal_enabled = true
message_journal_dir = data/journal
lb_recognition_period_seconds = 3
mongodb_uri = mongodb://localhost/graylog
mongodb_max_connections = 1000
mongodb_threads_allowed_to_block_multiplier = 5
content_packs_auto_load = grok-patterns.json
```

elasticsearch.yml
```
network.host: 192.168.99.100
discovery.zen.ping.unicast.hosts: ["192.168.99.100"]

```
```
> java  -Djava.net.preferIPv4Stack=true -jar graylog.jar server -f graylog.conf                                 ~/graylog-2.0.3@sam-VirtualBox
2016-07-26 18:15:43,351 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Collector 1.0.3 [org.graylog.plugins.collector.CollectorPlugin]
[...]

PR #2538, opened 2016-07-26
Validate that Elasticsearch home/data paths are readable

joschi:
These validations will ensure that cases like #2536 result in a better error message.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
@@ -17,8 +17,11 @@
 package org.graylog2.configuration;
 
 import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.ValidatorMethod;
 import com.github.joschi.jadconfig.converters.StringListConverter;
 import com.github.joschi.jadconfig.util.Duration;
+import com.github.joschi.jadconfig.validators.DirectoryWritableValidator;
 import com.github.joschi.jadconfig.validators.FilePathReadableValidator;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
```diff
@@ -268,4 +274,41 @@ public String getPathHome() {
...
+
+        if (Files.exists(parent) && !Files.isReadable(parent)) {
+            throw new ValidationException("Path " + parent.toAbsolutePath() + " cannot be read.");
+        }
+
+        if (Files.exists(path) && !Files.isDirectory(path)) {
+            throw new ValidationException("Path " + path.toAbsolutePath() + " is not a directory.");
+        }
+
+        if (Files.exists(path) && !Files.isReadable(path)) {
```
bernd: Is there a case where the directory also needs to be writable? If so, we should also test for that.

joschi: It has to be writable if data is being stored (`elasticsearch_node_data`). Other than that, I haven't found a case for which it had to be writable.

Funny enough, the directories don't even have to exist, it's sufficient if the parent directory is there.

bernd: the :-1: was wrong :smile:

bernd:
LGTM :+1:

graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
joschi: [So la la](https://www.youtube.com/watch?v=imBvAPQCD4E)…

PR #2539, opened 2016-07-26
Decorator UI Fixes

dennisoelkers:
Some minor fixes for decorator UI.
  
These include:
    
  - resetting the decorators form so old values are not kept on cancel
  - making add decorator component pure, to prevent flickering in select
  - checking if SearchPage is mounted before setting state in callback
  - adjusting the styles of some components to make layout consistent between Chrome and Safari


edmundoa:
LGTM 👍

PR #2542, opened 2016-07-26
Upgrade to Shiro 1.3.0

joschi:


bernd:
Tested by logging in with different users and backends. LGTM :+1:

Issue #1704, opened 2016-01-19
No Backlog even with RFC5424 rsyslog


navidf10:
Hi,

I'm new to graylog-server. I installed it on an ubuntu server and everything works fine except that when a message is routed to my test stream, the alert via email doesn't display any message (no backlog). The server sending the rsyslog messages is set to format its messages with RFC5424 format. the strange thing is that Graylog receives the messages correctly (with all the right fields and all) but still when routed loses data (??). 

I would appreciate a little help on this matter. 

Thank you in advance guys.
Great work. 

Regards,
N.

bernd:
I believe that this is related to https://github.com/Graylog2/graylog2-server/issues/2382 which will be fixed in the upcoming Graylog 2.1 via https://github.com/Graylog2/graylog2-server/pull/2546

I am closing this now, please re-open if you still have this issue with Graylog 2.1. Thank you for the report!

PR #2546, opened 2016-07-26
Fix timing issue in MessageCountAlertCondition

### LINKED ISSUE #2382: Alert triggered but message not included
### Problem description
We use [graylog-jira-alarmcallback](https://marketplace.graylog.org/addons/1e653721-0f51-43fd-9ab6-9159a39c0955) to log jira tickets when an alert on a stream is raised.

Every now and then we notice that an alert is raised without the message being passed to the plugin.
We have configured the alert condition as such.
More than **0** messages in the last **1** minute and wait at least **0** minutes until triggering a new alert.
When sending an alert, include the last **1** messages  ...

2016-06-18 10_41_21-graylog web interface

Checking the graylog logs around the problem time we see the following message.

`2016-06-17_16:52:04.15041 2016-06-17 16:52:04,149 WARN : com.bidorbuy.graylog.alarmcallbacks.jira.JiraAlarmCallback - Skipping JIRA-issue MD5 generation, alarmcallback did not provide a message`

We initailly logged an issue with the plugin author who after investigating the issue believed this is possibly a graylog bug.
You can see the discussion [here](https://github.com/magicdude4eva/graylog-jira-alarmcallback/issues/11#issuecomment-226945695)

It is worth to note that we are running more than one graylog servers in a clustered environment, in case that has any relevance.

Not easy, as it happens rarely

- Setup and configure the [graylog-jira-alarmcallback](https://marketplace.graylog.org/addons/1e653721-0f51-43fd-9ab6-9159a39c0955)
- Monitor the graylog logs for a message similar to this.
    `WARN : com.bidorbuy.graylog.alarmcallbacks.jira.JiraAlarmCallback - Skipping JIRA-issue MD5 generation, alarmcallback did not provide a message`

### Environment
AWS Image
* Graylog Version: 2.0.2 (but happens with 1.3.x as well)

Thanks

bernd:
This condition works by first running a count in Elasticsearch and then,
if the condition triggers, a search to fetch the messages that will be
included in the check result.

Both queries use a RelativeRange object which returns a new time for
every getFrom() and getTo() call that is made. This can result in
different messages being included in the check result or no messages at
all given the count query takes a while.

The RelativeRange is now converted to an AbsoluteRange object which is
then used to run the count and search query. This makes sure the exact
same time range is used no matter how much time is in between the calls.

Refs #1704
Fixes #2382

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/types/MessageCountAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/types/MessageCountAlertCondition.java
@@ -26,6 +26,7 @@ import org.graylog2.indexer.results.ResultMessage;
 import org.graylog2.indexer.results.SearchResult;
 import org.graylog2.indexer.searches.Searches;
 import org.graylog2.indexer.searches.Sorting;
+import org.graylog2.plugin.indexer.searches.timeranges.AbsoluteRange;
 import org.graylog2.plugin.indexer.searches.timeranges.InvalidRangeParametersException;
 import org.graylog2.plugin.indexer.searches.timeranges.RelativeRange;
 import org.graylog2.plugin.Message;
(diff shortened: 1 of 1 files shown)


edmundoa:
LGTM 👍

PR #2550, opened 2016-07-27
Fix REST API browser after changes to the PluginRestResource injection

bernd:
Create a wrapper class for the map of PluginRestResource classes to make
it injectable into Jersey resources.

Fallout from #2492.


joschi:
LGTM. 👍

Issue #2549, opened 2016-07-27
Indices Range-Start Tooltip gone after upgrade to 2.0


gruselglatz:
### Problem description
In Version 1.3 there was a useful tooltip when you hover over the "Contains messages from ... to ...." text on the "Indices" Page.

In Cases like "Contains messages from 5 years ago up to 14 days ago", you could fast get the detailed range when you hover over the text. In 2.0+ i have to make API-Calls to get to it which takes much longer...

### Steps to reproduce the problem

1. Go to Indices Page
2. Hover over the part which indicates the start and end of the index range
3. See that you see nothing :(

### Environment

* Graylog Version: 2.0.3
* Elasticsearch Version: -
* MongoDB Version: -
* Operating System: Windows 7
* Browser version: Chrome 50+

kroepke:
Good point, that should be trivial to add back in 2.1

kroepke:
@edmundoa this could probably just use the Timestamp component instead of DateTime directly, right?

edmundoa:
@kroepke I think so, will take a look.

coffee-squirrel:
Duplicate of #2486 ?

kroepke:
@coffee-squirrel true, sorry I missed it :(

kroepke:
I'll close the original one because this already has a PR attached.

coffee-squirrel:
@kroepke No worries, I'm glad to see there's a PR!

PR #2552, opened 2016-07-27
Display original date time of index ranges on hover

edmundoa:
Before it was possible to see the actual date time of index ranges by hovering over the relative time displayed in the UI, but on 2.0.3 that was not possible any more. This PR brings back that functionality, fixing #2549.


joschi:
Make index ranges great again! :trollface: 

LGTM. 👍

PR #2555, opened 2016-07-27
Fix load problems by setting __webpack_public_path__ early

bernd:
The __webpack_public_path__ needs to be set as early as possible to
ensure that all code injections into the web page know about it.

Use require() instead of import in src/index.jsx to avoid reordering of
the code.

**WARNING**: This fixes the issue in the core web interface, but plugins are still not getting the app path prefix because they are living in a separate webpack bundle. This needs to be fixed.


bernd:
There is a better fix coming up for this.

PR #2556, opened 2016-07-27
for HttpHeadersToken pass actual remote address

kroepke:
the "host" will be the actual client host in case trusted_proxies are set.
for the SSO plugin we need to know the actual remote address, which will be the proxy closest to Graylog, to verify that the proxy is ok to set the user header
otherwise we cannot verify anything

This is necessary for the SSO plugin to work in the current state on master.


joschi:
LGTM. 👍

PR #2560, opened 2016-07-28
Upgrade to moment 2.14.1

joschi:
Security vulnerability: https://nodesecurity.io/advisories/55


kroepke:
Does this apply to our plugins as well?

joschi:
> Does this apply to our plugins as well?

If they are bundling their own version of `moment`, they are affected as well.

edmundoa:
At least some of them do, not remember exactly which ones, though.

edmundoa:
LGTM 👍

PR #2563, opened 2016-07-29
Add explanation about the configuration file format

joschi:
The Graylog configuration file is a Java properties file which has to follow some rules.

This PR adds a short description of the format in the Graylog example configuration file.


joschi:
This could/should also be merged into the `2.0` and `1.3` branches (basically all branches, but those two should suffice for starters).

edmundoa:
LGTM 👍

PR #2564, opened 2016-07-29
Fix issues with app prefix

edmundoa:
The current implementation handling an app prefix has some issues:

1. Reloading a page in a subdirectory, e.g. `/system/overview`, is not able to fetch fonts and other assets. This is fixed by setting `__webpack_public_path__` earlier, and falling back to `/` when no app prefix is set
1. Routes used in plugins are not adding the app prefix set, making navigation within the plugin impossible if the application prefix is set. This is partially fixed, by adding a registry of plugin routes that are automatically prefixed if needed. Plugins need to use that registry to fix their navigation
1. Plugins do not know about the `__webpack_public_path__` setting, so they may use a wrong public path. This needs to be fixed in the plugins directly

### DIFF
--- a/graylog2-web-interface/src/routing/Routes.jsx
+++ b/graylog2-web-interface/src/routing/Routes.jsx
@@ -124,8 +124,8 @@ const qualifyUrls = (routes, appPrefix) => {
         qualifiedRoutes[routeName] = `${appPrefix}${routes[routeName]}`;
         break;
       case 'function':
-        qualifiedRoutes[routeName] = function () {
-          const result = routes[routeName](...arguments);
+        qualifiedRoutes[routeName] = (...params) => {
+          const result = routes[routeName](...params);
           return `${appPrefix}${result}`;
         };
         break;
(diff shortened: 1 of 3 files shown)


joschi:
The necessary changes to plugins need to be documented in the upgrade notes and in the documentation at http://docs.graylog.org/en/2.1/pages/plugins.html

graylog2-web-interface/src/routing/Routes.jsx
```diff
@@ -99,5 +141,9 @@ const qualifyUrls = (routes, appPrefix) => {
 };
 
 const defaultExport = AppConfig.gl2AppPathPrefix() ? qualifyUrls(Routes, AppConfig.gl2AppPathPrefix()) : Routes;
+window.pluginRoutes = AppConfig.gl2AppPathPrefix() ? qualifyUrls(pluginRoutes, AppConfig.gl2AppPathPrefix()) : pluginRoutes;
+
+// Plugin routes need to be prefixed separately, so we add them to the Routes object just at the end.
+defaultExport.pluginRoute = (key) => window.pluginRoutes[key];
```
bernd: Maybe we should throw an error here if the key does not exist to avoid throwing an error deep down in the stack?

edmundoa: Good point 👍

bernd:
LGTM :+1:

PR #2566, opened 2016-07-29
support "only named captures" for pipeline grok function

kroepke:
the server cache is necessary because the named captures support needs a separately compiled regex
so far the cache is only used by the grok function in the pipeline processor

fixes Graylog2/graylog-plugin-pipeline-processor#59


kroepke:
this is necessary to support https://github.com/Graylog2/graylog-plugin-pipeline-processor/pull/65

joschi:
LGTM. 👍

PR #2570, opened 2016-08-01
Update moment-timezone

edmundoa:
Includes new time zone information. Here's the list of changes if you are curious: https://github.com/moment/moment-timezone/blob/develop/changelog.md


joschi:
LGTM. 👍

Issue #2567, opened 2016-07-29
Make sure to include charset in getBytes and other relevant code sections


kroepke:
### Problem description

On systems where UTF-8 isn't the default character encoding loading messages with the default character set can fail if they contain non-ASCII characters.

Include the workaound mentioned in https://github.com/Graylog2/graylog-plugin-pipeline-processor/issues/61#issuecomment-236211200 to prevent the wrong encoding to be used.

### Steps to reproduce the problem

1. Use a Russian locale on your server
2. Use the pipeline simulator to send a cyrillic message
3. Press simulate
4. The non-ascii chars are replaced by `?` chars

### Environment

* Graylog Version: 2.1.0
* Elasticsearch Version: n/a
* MongoDB Version: n/a
* Operating System: n/a
* Browser version: n/a

joschi:
This particular error has been fixed in 2ccef59877da10a28f57416827219c828ee3c6fe.

IMHO, Graylog should always work, independent of the configured system locale (which is why we add an explicit encoding hint on all relevant methods, e. g. [`String.getBytes()`](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes-java.nio.charset.Charset-)).

bernd:
We are going to fix this in the code.

kroepke:
fixed by #2574

PR #2574, opened 2016-08-01
Check for Forbidden APIs in default build profile

joschi:
This PR moves the Forbidden APIs check into the default build profile (was in the `(P)edantic` profile before) and activates the plugin to fail builds on violation.

Additionally, this PR updates the Forbidden APIs signatures and fixes all violations in the current Graylog codebase.


joschi:
Refs #2567

kroepke:
lgtm!

PR #2575, opened 2016-08-01
Move Error Prone into default build profile 

joschi:
This PR moves the [Error Prone](http://errorprone.info/) compile checks into the default build profile (was in the `(P)edantic` profile before).

Additionally, this PR fixes many [Error Prone](http://errorprone.info/) warnings in the current Graylog codebase.


kroepke:
lgtm!

PR #2577, opened 2016-08-01
Fix GrokPatternRegistry.GrokReloader#load()

joschi:
`GrokPatternRegistry.GrokReloader#load()` used to run an unnecessary yet incorrect check whether the Grok pattern's name or the pattern itself was `null` or empty.

This check is already done by `Grok` itself and will throw a detailed exception which can be caught further up the stack.

Additionally this PR extracts the root cause of exceptions happening in the `cachedGrokForPattern()` method instead of simply propagating the wrapped `ExecutionException` and even adds a test for this.

Refs #2566


edmundoa:
LGTM 👍

PR #2582, opened 2016-08-02
Upgrade to Kafka 0.10.0.1

joschi:
This PR upgrades Kafka to version 0.10.0.0 and adapts `KafkaJournal` accordingly.

Unfortunately, the on-disk format of the Kafka Log changed so that this is not backward-compatible with old disk journal files.

On the positive side, the Kafka Log now supports adding an additional timestamp to entries and different compression algorithms can be configured (not supported by `KafkaJournal`).

On the long run, we should probably extract the Kafka Log (and rewrite it in Java, /cc @bernd) to get independent of the Apache Kafka update cycles.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
@@ -383,7 +383,7 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
                 final byte[] idBytes = entry.getIdBytes();
 
                 payloadSize += messageBytes.length;
-                messages.add(new Message(messageBytes, idBytes));
+                messages.add(new Message(messageBytes, idBytes, Message.NoTimestamp(), Message.MagicValue_V0()));
 
                 if (LOG.isTraceEnabled()) {
                     LOG.trace("Message {} contains bytes {}", bytesToHex(idBytes), bytesToHex(messageBytes));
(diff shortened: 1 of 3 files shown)


kroepke:
What does "not backwards-compatible" mean in this case?
Can it read old files, but older versions cannot read files created by 0.10?

joschi:
@kroepke In my tests it couldn't read existing journal files and journal files created by this version (Apache Kafka 0.10.0.0) couldn't be read by the old one (Apache Kafka 0.9.0.1).

kroepke:
@joschi ok then it sounds like we need a migration step to be able to update reliably 😞

mautematico:
¡Hola!

I am looking for a proper way to define a graylog2 steam output that sends logs out to kafka.
At first, I tought I'll need to write my own output plugin.
Bad lucky me, there is this #3010 issue blocking me on the plugin creation path, until new tool becomes available.

After finding this PR, now I think that maybe graylog2 in fact has a built in way (or a plugin) that can already achieve this.

Could you kindly point me on the right direction on this?

¡Gracias!

joschi:
@mautematico This PR doesn't include a Kafka output.

Feel free to create a feature request for this.

ckauf:
@joschi This doesn't look like that graylog2 will support Kafka 0.10 with SASL authentification as an input, is that correct? Is there already something planed or should I open an issue/feature request for that?

joschi:
@ckauf That's correct, Graylog currently doesn't support Kafka 0.10.x.

Since this PR is a bit old, we'll probably take another approach for it, but for now it's definitely not on our roadmap. Feel free to open another feature request for it.

joschi:
Happy anniversary! This PR isn't going to make it, so I'm closing it.

PR #2583, opened 2016-08-02
More fixes with prefixed URLs

edmundoa:
This PR includes fixes for these issues:

- Some components use routes from `ApiRoutes`, which is not the right place for them, as they are not properly prefixed
- Adding an app prefix breaks the selected system page that is displayed in the navigation bar


bernd:
LGTM :+1:

PR #2584, opened 2016-08-02
Fix extractor and static fields creation in multi-node setups

### LINKED ISSUE #2580: Extractor creation cannot load input in multi-node setup
### Problem description

When creating an extractor for a non-global input in a multi node setup, the extractor creation will fail if the input is not running on the node the web interface is being served from.

The REST resource checks its local input states for a running input, which it does not have and then fails the request.

1. Start at least two nodes (A and B)
2. Use the web interface from node A
3. Create an input on a node B
4. Send a message, search and create an extractor for a field
5. Configure arbitrary extractor
6. Save that extractor
7. Server returns 404 "no such input"

### Environment

* Graylog Version: 2.1.0-beta.2 SNAPSHOT
* Elasticsearch Version: n/a
* MongoDB Version: n/a
* Operating System: n/a
* Browser version: n/a

bernd:
The extractor and static fields resource was using the local input
registry to check if an input exists. This breaks in a multi-node setup
where the web interface runs on node A and the input on node B.

Fixes #2580

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
@@ -44,7 +44,6 @@ import org.graylog2.rest.models.system.inputs.extractors.responses.ExtractorCrea
 import org.graylog2.rest.models.system.inputs.extractors.responses.ExtractorMetrics;
 import org.graylog2.rest.models.system.inputs.extractors.responses.ExtractorSummary;
 import org.graylog2.rest.models.system.inputs.extractors.responses.ExtractorSummaryList;
-import org.graylog2.shared.inputs.InputRegistry;
 import org.graylog2.shared.inputs.PersistedInputs;
 import org.graylog2.shared.metrics.MetricUtils;
 import org.graylog2.shared.rest.resources.RestResource;
@@ -81,7 +80,6 @@ public class ExtractorsResource extends RestResource {
 
     private final InputService inputService;
     private final ActivityWriter activityWriter;
-    private final InputRegistry inputs;
     private final MetricRegistry metricRegistry;
     private final ExtractorFactory extractorFactory;
     private final PersistedInputs persistedInputs;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
```diff
@@ -120,14 +116,7 @@ public Response create(@ApiParam(name = "inputId", required = true)
...
 
-        final MessageInput input = inputs.getRunningInput(inputId);
-        if (input == null) {
-            final String msg = "Input <" + inputId + "> not found.";
-            LOG.error(msg);
-            throw new javax.ws.rs.NotFoundException(msg);
-        }
-
-        final Input mongoInput = inputService.find(input.getPersistId());
+        final Input mongoInput = inputService.find(inputId);
```
kroepke: I think we should test for `null` here. If we don't it's probably harmless, but the error wouldn't be nice.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
```diff
@@ -173,13 +162,7 @@ public ExtractorSummary update(@ApiParam(name = "inputId", required = true)
...
         checkPermission(RestPermissions.INPUTS_EDIT, inputId);
 
-        final MessageInput input = persistedInputs.get(inputId);
-        if (input == null) {
-            LOG.error("Input <{}> not found.", inputId);
-            throw new javax.ws.rs.NotFoundException("Couldn't find input " + inputId);
-        }
-
-        final Input mongoInput = inputService.find(input.getPersistId());
+        final Input mongoInput = inputService.find(inputId);
```
kroepke: Same `null` check here

kroepke:
other than the `null` check comments, lgtm 👍

graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
bernd: `InputService#find()` throws a `NotFoundException` when the input cannot be found.

See: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/inputs/InputServiceImpl.java#L148

graylog2-server/src/main/java/org/graylog2/rest/resources/system/inputs/ExtractorsResource.java
bernd: See above

joschi:
LGTM. 👍

PR #2587, opened 2016-08-02
Revert "Move link to API Browser into System menu"

joschi:
This reverts commit b6fe1a9604a59579b359cbfef94e8349a7132b7d.

### DIFF
--- a/graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
+++ b/graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
@@ -29,6 +29,10 @@ const NodeMaintenanceDropdown = React.createClass({
               <MenuItem>Configure internal logging</MenuItem>
             </LinkContainer>
           </IfPermitted>
+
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/nodes/NodeMaintenanceDropdown.jsx
```diff
@@ -29,6 +29,10 @@ const NodeMaintenanceDropdown = React.createClass({
               <MenuItem>Configure internal logging</MenuItem>
             </LinkContainer>
           </IfPermitted>
+
+          <MenuItem href={`${this.props.node.transport_address}api-browser`} target="_blank">
```
edmundoa: Could you please construct this link like we do with the one in `NodesActions`? Something like:
```javascript
new URI(`${this.props.node.transport_address}/api-browser`).normalizePathname();
```

I think we could add a component that unifies both usages, I leave it up to you.

bernd:
LGTM :+1:

PR #2597, opened 2016-08-03
updated issue template and added pr template

jalogisch:
i had the feeling that it would be a benefit if our issue template has more comments (like i had already done in the documentation repo)

additional maybe a pr template is useful for external code contribution, thats why i had added it.


joschi:
I think the pull request template is a bit overkill, but let's simply try it out.

kroepke:
tbh, this feels like Passierschein A38

Did we have an actual problem with submissions?

Issue #2596, opened 2016-08-03
Make possible to set rest_transport_uri relative


hc4:
### Problem description
If I configure `rest_transport_uri =` then graylog able to open main page and some other (e.g. streams, search), but some rest requests won't work.
For example:
![image](https://cloud.githubusercontent.com/assets/17475270/17359939/a2eb8af2-5971-11e6-96c8-7d41580bfc02.png)
return error: `{"type":"ApiError","message":"Illegal URL: "}`

Also error can be found in logs:
```
2016-08-03T11:56:34.538+03:00 ERROR [AnyExceptionClassMapper] Unhandled exception in REST resource
java.lang.IllegalArgumentException: Illegal URL:
        at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:437) ~[graylog.jar:?]
        at org.graylog2.rest.RemoteInterfaceProvider.get(RemoteInterfaceProvider.java:58) ~[graylog.jar:?]
        at org.graylog2.shared.rest.resources.ProxiedResource.lambda$createRemoteInterfaceProvider$1(ProxiedResource.java:90) ~[graylog.jar:?]
        at org.graylog2.shared.rest.resources.ProxiedResource.lambda$getForAllNodes$0(ProxiedResource.java:66) ~[graylog.jar:?]
        at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1321) ~[?:1.8.0_91]
        at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169) ~[?:1.8.0_91]
        at java.util.HashMap$EntrySpliterator.forEachRemaining(HashMap.java:1683) ~[?:1.8.0_91]
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481) ~[?:1.8.0_91]
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471) ~[?:1.8.0_91]
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708) ~[?:1.8.0_91]
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:1.8.0_91]
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499) ~[?:1.8.0_91]
[...]

PR #2589, opened 2016-08-02
Ensure that "web_endpoint_uri" is an absolute URI

joschi:
Refs https://groups.google.com/d/msg/graylog2/lEAl8DnGhMU/J49MbPn3BAAJ

PR #2600, opened 2016-08-03
Ensure that {rest,web}_{listen,transport,endpoint}_uri settings are absolute URIs

joschi:
This PR ensures that all URI configuration settings in the Graylog configuration file are indeed absolute URIs.

Refs #2596, supersedes #2589.


bernd:
LGTM :+1:

hc4:
web_endpoint_uri should be allowed to be relative
It allows me to use kerberos SSO auth with possibility to access graylog at different url's.

PR #2601, opened 2016-08-03
Print - instead of null when client did not provide user agent header

mikkolehtisalo:
HTTP clients not providing user agent header cause "null" to be printed in logs. Found this out when a braindead load balancer product started hammering /system/lbstatus... Just a cosmetic fix.

By the way, responseContext.getLength() can never work at RestAccessLogFilter, because Jersey has not produced the body yet - thus there is no Content-Length available. I think counting responseContext.getEntityStream() might work, that seemed like heavy to me.


joschi:
LGTM. 👍 

Thanks!

PR #2617, opened 2016-08-04
Reduce production .js files sizes by 51%

bernd:
- Do not use the react-hot loader for production builds
- Use "source-map" instead of "eval" for "devtool" setting in production
  builds
- Enable source maps in the UglifyJsPlugin for production builds

By not using "eval" for the "devtool" setting we are getting much
smaller .js files. Using "source-map" for "devtool" and enabling source
maps for the UglifyJsPlugin increases the production build time, but also
produces higher quality source maps for debugging.

Before:

- Build time: 0m39.788s
- All .js files size: 7.9MB

After:

- Build time: 1m4.028s
- All .js files size: 3.8MB

### DIFF
--- a/graylog2-web-interface/webpack.config.js
+++ b/graylog2-web-interface/webpack.config.js
@@ -28,7 +28,6 @@ const webpackConfig = {
     ],
     loaders: [
       { test: /\.json$/, loader: 'json-loader' },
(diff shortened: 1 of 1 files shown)


bernd:
This needs to be tested carefully that it doesn't break something.

edmundoa:
LGTM 👍

PR #2621, opened 2016-08-05
use default session attribute for principal

### LINKED ISSUE #2620: SSO authenticated users not shown as active
## Expected Behavior
Should be real remote IP of client

## Current Behavior
For me it shows 127.0.0.1
image

## Context
Maybe the problem caused by proxy.
Client actualy connects to proxy, and proxy to graylog.
Can graylog support `X-Forwarded-For` header?

## Your Environment
* Graylog Version: 2.1.0-beta2

kroepke:
## Description
the SessionResource set a custom session attribute to find the name of the user owning the session, but the if auth plugins forced a session creation that wasn't set.
Instead of trying to fix the auth plugins, rely on a shiro framework attribute to get the principal

## Motivation and Context
fixes #2620 

## How Has This Been Tested?
Use the SSO plugin and a proxy setting a auth header.
Before the change users that logged in via the SSO plugin aren't show as having an active session in the Users list.
After the change the users' sessions can be found and they are shown as logged in.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
@@ -129,7 +129,6 @@ public class SessionsResource extends RestResource {
             final User user = userService.load(createRequest.username());
             if (user != null) {
                 long timeoutInMillis = user.getSessionTimeoutMs();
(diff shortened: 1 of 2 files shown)


bernd:
LGTM :+1:

PR #2632, opened 2016-08-08
Add start-nohmr script to run webpack without HMR

bernd:
Makes it possible to run the development server without hot module reload.


edmundoa:
LGTM 👍

Issue #2603, opened 2016-08-03
Simplify usage of rest/web_listen_uri


mariussturm:
### Problem description
The desired state is to have the web interface on `<domain>/` and the api listener on `<domain>/api`.

1st try: `web_listen_uri: http://0.0.0.0:9000/` `rest_listen_uri: http://0.0.0.0:9000/api`
 Error message is: `com.github.joschi.jadconfig.ValidationException: If REST and Web Interface are served on the same host/port, the web interface must have a path prefix!`
 
This is confusing because the default setting is exactly the other way around: `web_listen_uri: http://0.0.0.0:12900/web` `rest_listen_uri: http://0.0.0.0:12900/` I would expect to swap both URIs without a problem.

2nd try: `web_listen_uri: http://0.0.0.0:9000/web` `rest_listen_uri: http://0.0.0.0:9000/api`
This works but nothing is on the root listener. At least a redirect to `web_listen_uri` or `web_endpoint_uri` should be send there.

3rd try `web_listen_uri: http://0.0.0.0:9000/web` `rest_listen_uri: http://0.0.0.0:9000/`
Opening `http://<IP>:9000` works after some seconds, a redirect to `http://<IP>:9000/web` happens but a stacktrace shows up in the server log:

```
graylog_1          | Caused by: java.io.IOException: Broken pipe
graylog_1          | 	at sun.nio.ch.FileDispatcherImpl.write0(Native Method) ~[?:1.8.0_91]
graylog_1          | 	at sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47) ~[?:1.8.0_91]
graylog_1          | 	at sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:93) ~[?:1.8.0_91]
graylog_1          | 	at sun.nio.ch.IOUtil.write(IOUtil.java:51) ~[?:1.8.0_91]
graylog_1          | 	at sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:471) ~[?:1.8.0_91]
graylog_1          | 	at org.glassfish.grizzly.nio.transport.TCPNIOUtils.flushByteBuffer(TCPNIOUtils.java:149) ~[graylog.jar:?]
graylog_1          | 	at org.glassfish.grizzly.nio.transport.TCPNIOUtils.writeCompositeBuffer(TCPNIOUtils.java:87) ~[graylog.jar:?]
[...]

PR #2634, opened 2016-08-08
Enable running Graylog REST API on different context path

joschi:
## Description

Refs #2603

## Motivation and Context

Users might (rightfully) want to run the Graylog REST API and the web interface on the same network listener with the web interface being served on the root context path (`/`) and the Graylog REST API being served from a path below that (`/api`).

This PR merges `RestApiService` and `WebInterfaceService` into a single service which handles setting up the required handlers accordingly.

## How Has This Been Tested?

Painfully.

Things that might still be broken and might only be working on my machine™:
* Swagger UI in general
* The generic 404-handler so that the web interface can have an arbitrary entry-point in the web browser
* Generated links in the Graylog REST API

## Types of changes

## Checklist:
- [x] I have updated the documentation accordingly
  - Well, the upgrade notes at least.


graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/RemoteDeflectorResource.java
```diff
@@ -20,6 +20,6 @@
 import retrofit2.http.POST;
 
 public interface RemoteDeflectorResource {
-    @POST("/system/deflector/cycle")
+    @POST("system/deflector/cycle")
```
kroepke: Are these changes strictly necessary, or rather can we catch this in a central place?
I'm worried that plugins that implement a ProxiedResource will forget to make the path relative and thus break at runtime.

joschi: Yes, these are necessary.

Otherwise the custom context path would be overridden (because the paths in these Retrofit resource classes are absolute and not relative as they should be). I haven't looked into it too deeply but I haven't found a suitable mechanism in [Retrofit's documentation](https://square.github.io/retrofit/) yet.

Also see [Retrofit 2 — Url Handling, Resolution and Parsing](https://futurestud.io/blog/retrofit-2-url-handling-resolution-and-parsing).

kroepke: Ok bummer. I was asking because at least the audit log plugin needs the corresponding change in this case.

kroepke: Could we potentially add a runtime check for the affected resources to log error messages?
Not sure where in the code this would have to go, the closest I could find is https://github.com/Graylog2/graylog2-server/blob/e638987fbf0e58e2211302a558a003fbfc18dff2/graylog2-server/src/main/java/org/graylog2/rest/RemoteInterfaceProvider.java#L41-L63

graylog2-server/src/main/resources/swagger/index.html.template
```diff
@@ -1,98 +1,98 @@
...
+    var apiTarget = window.location.protocol + "//" + window.location.hostname + ":" + port + "${apiPrefix}";
+    $(function () {
+      window.swaggerUi = new SwaggerUi({
+      url: apiTarget + "api-docs",
+      dom_id: "swagger-ui-container",
+      supportedSubmitMethods: ['get', 'post', 'put', 'delete'],
+      onComplete: function(swaggerApi, swaggerUi){
+
+        // We are taking the base path of the system API here. This will break if you should ever rename it lol.
+        if(swaggerApi.System && !apiTarget.startsWith(swaggerApi.System.basePath))   {
```
kroepke: I think `startsWith` is not supported in older browsers. Probably better to use
`apiTarget.lastIndexOf(swaggerApi.System.basePath, 0) === 0`

kroepke:
I need to test a couple of more scenarios tomorrow, but overall this looks good!

graylog2-server/src/main/resources/swagger/index.html.template
kroepke: 🎯

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/system/RemoteDeflectorResource.java
kroepke: We agreed to simply fix the affected resources in the relevant plugins.
Since the default is to run the API on a sub path, this will fail early on which is good enough.

kroepke:
I've tested the following combinations with 2 nodes:
- out of the box defaults (only adjusting ports for running on the same box)
- API on `/api` (node 1) and `/schmapi` (node 2), web default
- same as above, web on `/web`
- using all current plugins

The only thing that currently does not work is to put the web listener on a separate IP/Port. The assets then all end up in a 404.

kroepke:
Works now on the separate listener.

kroepke:
lgtm

PR #2637, opened 2016-08-08
Compile regex pattern for MetricFilter only once

mikkolehtisalo:
## Description
Noticed a static regex getting compiled 60 times per second on my system. Modified implementation so that it gets compiled only once. Also moved the anonymous class outside run() so that further tests are easier to implement.

## Motivation and Context

This tiny change frees whopping ~1 millisecond of CPU time every second for actual message processing duties.

## How Has This Been Tested?
Added a unit test for positive test case. 

hello-sir

## Types of changes

## Checklist:


joschi:
LGTM. 👍

Thanks, Mikko!

joschi:
We might even save some cycles by simply using `String#startsWith(String)` and `String#endsWith(String)`, but that's no concern right now.

PR #2647, opened 2016-08-09
The retention strategy defaults to "Do nothing" unless an other strategy was explicitly configured

mikkolehtisalo:
## Description

Index retention strategy has defaulted to "delete". That is not fail-safe default value, because with a certain type of issue with the configuration database, or a bug, might lead to the default value being used. That could lead to unwanted index deletion, and data loss.

This PR changes "noop" to be the new default unless if something else has been explicitly defined. Also reordered the selection list in the UI from least destructive to most destructive option so that it takes more conscious effort to select "delete" (it's not the first / default if nothing is set). 

## Motivation and Context

Fail-safe defaults should be used for potentially destructive configuration options. 

## How Has This Been Tested?

Added basic unit test for NoopRetentionStrategy. 

## Types of changes

People upgrading from 1.x to 2.x should probably check the setting after upgrading. There might be some edge case that would change the expected value for some setup? That's the primary potential breakage, added a note about that to `UPGRADING.rst`.

## Checklist:

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -101,6 +101,7 @@ The retention and rotation configuration settings can be retrieved using the fol
 * ``/system/indices/rotation/config``
 * ``/system/indices/retention/config``
 
+The retention strategy defaults to "Do nothing" unless an other strategy was explicitly configured. Please check the setting after upgrading.
 
 For Plugin Authors
 ------------------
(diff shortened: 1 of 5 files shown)


kroepke:
We have discussed this a couple of times, but have not reached a conclusion yet.

While we appreciate that destructive actions shouldn't be the default, many people simply never think about disk space until they run out of it, too, especially since most beginning users start with a single node.

At the moment I don't think there's a good middle ground, so my vote would be to keep the current behavior.

Any further input @Graylog2/engineers ?

mariussturm:
+1 - out of the box Graylog should be as stable as possible. Otherwise it would be a time bomb.

mikkolehtisalo:
:grimacing:

joschi:
@mikkolehtisalo Thanks for your contribution to Graylog!

After discussing this change, we've decided to keep the current default settings.

Issue #2610, opened 2016-08-04
Simulator with GELF codec not working, NullPointerException


tokred:
With beta.2, the pipeline simulator does not show simulations results but stays at "Loading" forever, while a NullPointerException is thrown according to server log. Initially occured https://github.com/Graylog2/graylog-plugin-pipeline-processor/issues/40

## Expected Behavior

Apply pipeline logic to simulation messages and show results.

## Current Behavior

Tried several raw messages in simulation with GELF codec, simulation result missing, exception thrown in log:
```
2016-08-04T10:57:30.479+02:00 ERROR [AnyExceptionClassMapper] Unhandled exception in REST resource
org.graylog.plugins.pipelineprocessor.ast.exceptions.FunctionEvaluationException: java.lang.NullPointerException
        at org.graylog.plugins.pipelineprocessor.ast.expressions.FunctionExpression.evaluateUnsafe(FunctionExpression.java:65) ~[?:?]
        at org.graylog.plugins.pipelineprocessor.ast.expressions.BooleanValuedFunctionWrapper.evaluateBool(BooleanValuedFunctionWrapper.java:35) ~[?:?]
        at org.graylog.plugins.pipelineprocessor.processors.PipelineInterpreter.processForResolvedPipelines(PipelineInterpreter.java:320) ~[?:?]
        at org.graylog.plugins.pipelineprocessor.processors.PipelineInterpreter.processForPipelines(PipelineInterpreter.java:281) ~[?:?]
        at org.graylog.plugins.pipelineprocessor.processors.PipelineInterpreter.process(PipelineInterpreter.java:238) ~[?:?]
        at org.graylog.plugins.pipelineprocessor.rest.SimulatorResource.simulate(SimulatorResource.java:76) ~[?:?]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_101]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_101]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_101]
        at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_101]
[...]

PR #2650, opened 2016-08-10
Allow selecting an input ID in RawMessageLoder to add to the parsed message

bernd:
In some cases it is necessary to get a message with an input ID.
(i.e. pipeline simulator)

Refs #2610


kroepke:
lgtm

PR #2652, opened 2016-08-10
Check for conflict of rest_listen_uri and web_listen_uri

joschi:
Refs #2634

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -319,10 +319,24 @@ public class Configuration extends BaseConfiguration {
     }
 
     @ValidatorMethod
+    @SuppressWarnings("unused")
     public void validatePasswordSecret() throws ValidationException {
         final String passwordSecret = getPasswordSecret();
         if (passwordSecret == null || passwordSecret.length() < 16) {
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -319,10 +319,24 @@ public int getLoadBalancerRequestThrottleJournalUsage() {
...
+    @ValidatorMethod
+    @SuppressWarnings("unused")
+    public void validateNetworkInterfaces() throws ValidationException {
+        final URI restListenUri = getRestListenUri();
+        final URI webListenUri = getWebListenUri();
+
+        if ((restListenUri.getPort() == webListenUri.getPort()) &&
+                !restListenUri.getHost().equals(webListenUri.getHost()) &&
+                ("0.0.0.0".equals(restListenUri.getHost()) || "0.0.0.0".equals(webListenUri.getHost()))) {
+            throw new ValidationException("Wildcard IP addresses cannot be used if the Graylog REST API and web interface listen on the same port.");
```
bernd: How about adding the current settings of `rest_listen_uri` and `web_listen_uri` to the message? I think that makes it easier to debug.

joschi: I was afraid the message would get too convoluted if those were added.

bernd:
LGTM :+1:

PR #2654, opened 2016-08-10
Decorator improvements

edmundoa:
This PR adds some improvements to decorators. I still have some open items in my list, but I think splitting them up will make the review process less annoying.

These are some of the changes included:
- Improve overall appearance, hiding some elements in the decorator sidebar to gain some space, changing other UI elements to make them more obvious to users, and doing some _margins and paddings_ work
- Enforce user to select a pipeline to create a decorator from a pipeline
- Only show "show original" button when message is decorated
- Search response includes a list of all fields added, modified or removed
- Hide field analysers and field value actions for all decorated fields, not only added field

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageField.jsx
+++ b/graylog2-web-interface/src/components/search/MessageField.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { DecoratedMessageFieldMarker, MessageFieldDescription } from 'components/search';
+import { MessageFieldDescription } from 'components/search';
 
 const MessageField = React.createClass({
   propTypes: {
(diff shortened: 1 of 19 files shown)


graylog2-web-interface/src/components/search/MessageField.jsx
```diff
@@ -40,14 +32,15 @@ const MessageField = React.createClass({
 
     return (
       <span>
-        <dt key={`${key}Title`}>{key} {this._decorationMarker(key)}</dt>
```
kroepke: minor nitpick, I'd leave the decorated marker here instead of underneath the value.
with the new marker it's actually quite readable (I guess this change was made before the icon was changed into a text)

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -50,31 +53,38 @@ const MessageFieldDescription = React.createClass({
...
+        {this._getFormattedFieldActions()}
         <div className="field-value">{this.props.possiblyHighlight(this.props.fieldName)}</div>
         {this._shouldShowTerms() &&
-        <Alert bsStyle="info" onDismiss={() => this.setState({messageTerms: Immutable.Map()})}>
+        <Alert bsStyle="info" onDismiss={() => this.setState({ messageTerms: Immutable.Map() })}>
           Field terms: &nbsp;{this._getFormattedTerms()}
         </Alert>
-          }
+        }
+        {this.props.isDecorated && <DecoratedMessageFieldMarker />}
```
kroepke: if we move the marker next to the field name, this should be removed

kroepke:
other than the marker placement this lgtm

graylog2-web-interface/src/components/search/MessageField.jsx
edmundoa: The creative director (aka @mariussturm) and I were looking into some alternatives and the current approach is the one we liked the most: The idea was to be non-obtrusive when normally looking around message fields, as mostly you don't care if fields are decorated or not, while still having the information somewhere if you really needed to know. Having the `(decorated)` marker right next to the field may be too distracting, specially when several fields are modified by decorators.

kroepke:
We agreed that we cannot agree on the placement comment above. So we choose to not agree and leave it like this.

good work!

PR #2660, opened 2016-08-11
add format string message decorator

kroepke:
this uses the JMTE template engine to create synthetic fields at search time
useful for scenarios where the messages are fully decomposed, but a nicer human readable name is required during presentation


dennisoelkers:
Would you mind doing a quick auto-indent? The indentation looks off and inconsistent.

kroepke:
@dennisoelkers which do you mean? This is my standard indentation I always use.
The line breaks are probably because of the relatively long lines due to the 120 char column.

I still have an open task for proposing a project-wide code style guide we all use, but until then I think we could postpone the decision on this?

PR #2662, opened 2016-08-11
Avoid adding unnecessary credentials on URI for CSV export

### LINKED ISSUE #2661: Bad url generated for csv export if web_endpoint_uri is relative
I have empty web_endpoint_uri in config and my export CSV link looks like:
`***:session@/search/universal/relative/export?query=%2A&range=300&fields=source%2Cmessage`

## Your Environment
* Graylog Version: 2.1.0-beta.3-SNAPSHOT+45cc198

joschi:
The web browser is already sending proper credentials via the `Authentication` request header and adding those will break the "qualified" URI if `web_endpoint_uri` was a relative path.

Fixes #2661

### DIFF
--- a/graylog2-web-interface/src/components/search/SearchSidebar.jsx
+++ b/graylog2-web-interface/src/components/search/SearchSidebar.jsx
@@ -126,11 +126,9 @@ const SearchSidebar = React.createClass({
     const fields = this.props.selectedFields;
     const timeRange = SearchStore.rangeType === 'relative' ? { range: SearchStore.rangeParams.get('relative') } : SearchStore.rangeParams.toJS();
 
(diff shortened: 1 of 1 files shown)


edmundoa:
Removing the authentication does not work at all for me. I am currently using these settings in my Graylog config:

```
rest_listen_uri = http://0.0.0.0:9000/api/
web_listen_uri = http://0.0.0.0:9000/
```

Every time I try to export the search as CSV I get asked for my credentials to the Graylog server. I have tried both in the latests Chrome and Firefox.

joschi:
@edmundoa Yes, I could reproduce that. Seems like those credentials were cached from another request to the Graylog REST API the first time.

bernd:
What about this?

PR #2666, opened 2016-08-11
Move decorator help link to bottom

dennisoelkers:
## Description

This PR moves the decorator help link below the select box used to create decorators and lets it float to the right. It also removes the glyph in the help link and aligns it with the right border of the other elements.

## Motivation and Context

Without the change the help link looked misplaced.

Before:
screen shot 2016-08-11 at 16 03 44

After:
screen shot 2016-08-16 at 13 54 23

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/search/DecoratorSidebar.jsx
+++ b/graylog2-web-interface/src/components/search/DecoratorSidebar.jsx
@@ -91,7 +91,7 @@ const DecoratorSidebar = React.createClass({
       <div>
         <AddDecoratorButton stream={this.props.stream} nextOrder={nextDecoratorOrder}/>
         <OverlayTrigger trigger="click" rootClose placement="right" overlay={popoverHelp}>
-          <Button bsStyle="link" className={DecoratorStyles.helpLink}>What are message decorators? <i className="fa fa-question-circle" /></Button>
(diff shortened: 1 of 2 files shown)


edmundoa:
I really dislike that new positioning. In first place there are three elements to click in a really small space, in the second place, the link breaks the flow of all the other elements. On top of that the link is not aligned with the other elements on the screen.

If the problem is that the link at the top looks odd with the tabs, or is just too prominent, I suggest moving it below the add decorator dropdown, but leaving it on the left. We could change the icon to be on the left, as we do in many other places for better consistency.

dennisoelkers:
This is why I like it, it creates a separation between the action (creating a new decorator) and the list of items, which works so well because of the white space on the left side between the select box and the list items. It makes the whole box appear more airy and less crammed than before.

I would be fine with moving the glyph to the left of the text though and remove a bit of padding on the right end of the link to align it with the right ending of the other elements.

edmundoa:
I still personally like the link on the left side better, but in the end it's a matter of taste and opinion, so let's agree to disagree.

I am basically fine with the change if we align the link to the right with the other elements in the sidebar. Regarding the font-awesome icon, my current thinking is that we don't really need it, and that it looks odd. I added it for consistency with other help links, but this one is already a question that looks like a link, so adding the icon doesn't provide much value in my opinion.

edmundoa:
Looks good to me 👍

PR #2667, opened 2016-08-11
do not load plugins for journal commands

kroepke:
previously, if plugins were present, they were loaded and then failed the injector because of missing config beans in the journal cmd tool


bernd:
LGTM :+1:

PR #2669, opened 2016-08-11
Improved journal resiliency

### LINKED ISSUE #2659: Message Journal retries forever in case of irrecoverable errors
Due to a never stopping retry-strategy in [`JournallingMessageHandler`](https://github.com/Graylog2/graylog2-server/blob/2.0.3/graylog2-server/src/main/java/org/graylog2/shared/buffers/JournallingMessageHandler.java#L49-L59), Graylog will forever retry to write messages into the message journal in the face of an error, even if it's irrecoverable.

Example:
```
2016-08-11T08:41:03.465+02:00 ERROR [JournallingMessageHandler] Unable to write to journal - retrying with exponential back-off
kafka.common.MessageSetSizeTooLargeException: Message set size is 112367208 bytes which exceeds the maximum configured segment size of 104857600. at kafka.log.Log.append(Log.scala:300) ~[graylog.jar:?]
```
The message set is simply too large to fit into one segment of the Kafka Log and thus will always fail. Unfortunately, Graylog will never stop to retry writing those messages and will require a restart to resolve this error.

bernd:
- Make sure the message set we want to write fits into a segment
- Max message size should not be larger than max segment size
- Do not try to write a message which is larger than max message size

Fixes #2659

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
@@ -437,9 +437,7 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
 
     private long flushMessages(List<Message> messages, long payloadSize) {
         if (messages.isEmpty()) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("No messages to flush, not trying to write an empty message set.");
-            }
+            LOG.debug("No messages to flush, not trying to write an empty message set.");
             return -1L;
         }
 
--- a/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
@@ -21,7 +21,6 @@ import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.github.joschi.jadconfig.util.Size;
-import com.google.common.base.Charsets;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
@@ -398,7 +397,7 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
 
                 if (newMessageSize > maxMessageSize) {
                     LOG.warn("Message with ID <{}> is too large to store in journal, skipping! (size: {} bytes / max: {} bytes)",
-                            new String(idBytes, Charsets.UTF_8), newMessageSize, maxMessageSize);
+                            new String(idBytes, StandardCharsets.UTF_8), newMessageSize, maxMessageSize);
                     payloadSize = 0;
                     continue;
                 }
--- a/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
@@ -195,8 +195,8 @@ public class KafkaJournal extends AbstractIdleService implements Journal {
                 .put(LogConfig.RetentionBytesProp(), retentionSize.toBytes())
                 // retentionMs: The age approximate maximum age of the last segment that is retained
                 .put(LogConfig.RetentionMsProp(), retentionAge.getMillis())
-                // maxMessageSize: The maximum size of a message in the log
-                .put(LogConfig.MaxMessageBytesProp(), Integer.MAX_VALUE)
+                // maxMessageSize: The maximum size of a message in the log (ensure that it's not larger than the max segment size)
+                .put(LogConfig.MaxMessageBytesProp(), Math.min(Ints.saturatedCast(segmentSize.toBytes()), Integer.MAX_VALUE))
                 // maxIndexSize: The maximum size of an index file
                 .put(LogConfig.SegmentIndexBytesProp(), Ints.saturatedCast(Size.megabytes(1L).toBytes()))
                 // indexInterval: The approximate number of bytes between index entries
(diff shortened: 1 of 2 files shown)


bernd:
I didn't put it into a milestone yet, we have to decide if we want to do that for 2.1 or not.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -168,6 +170,9 @@ public KafkaJournal(@Named("message_journal_dir") File journalDirectory,
         this.scheduler = scheduler;
         this.throttleThresholdPercentage = intRange(throttleThresholdPercentage, 0, 100);
         this.serverStatus = serverStatus;
+        this.maxSegmentSize = segmentSize.toBytes();
+        // Max message size should not be bigger than max segment size.
+        this.maxMessageSize = Math.min(Ints.saturatedCast(maxSegmentSize), Integer.MAX_VALUE);
```
joschi: `Ints#saturatedCast()` will always be less than or equal to `Integer.MAX_VALUE`.

bernd: True :smile:

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -376,31 +381,83 @@ public Entry createEntry(byte[] idBytes, byte[] messageBytes) {
...
-            long lastWriteOffset = appendInfo.lastOffset();
-            LOG.debug("Wrote {} messages to journal: {} bytes, log position {} to {}",
-                    entries.size(), payloadSize, appendInfo.firstOffset(), lastWriteOffset);
-            writtenMessages.mark(entries.size());
             return lastWriteOffset;
         }
     }
 
+    private long flushMessages(List<Message> messages, long payloadSize) {
+        if (messages.isEmpty()) {
```
kroepke: Can this ever happen?

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -376,31 +381,83 @@ public Entry createEntry(byte[] idBytes, byte[] messageBytes) {
...
 
                 payloadSize += messageBytes.length;
-                messages.add(new Message(messageBytes, idBytes));
+
+                final Message newMessage = new Message(messageBytes, idBytes);
+                // Calculate the size of the new message in the message set by including the overhead for the log entry.
+                final int newMessageSize = MessageSet.entrySize(newMessage);
+
+                if (newMessageSize > maxMessageSize) {
+                    LOG.warn("Message with ID <{}> is too large to store in journal, skipping! (size: {} bytes / max: {} bytes)",
```
kroepke: Do we want to track the number of discarded messages in a metric here?

bernd: Yes, good point.

bernd: I added a meter for that.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -168,9 +171,13 @@ public KafkaJournal(@Named("message_journal_dir") File journalDirectory,
         this.scheduler = scheduler;
         this.throttleThresholdPercentage = intRange(throttleThresholdPercentage, 0, 100);
         this.serverStatus = serverStatus;
+        this.maxSegmentSize = segmentSize.toBytes();
+        // Max message size should not be bigger than max segment size.
+        this.maxMessageSize = Ints.saturatedCast(maxSegmentSize);
 
         this.writtenMessages = metricRegistry.meter(name(this.getClass(), "writtenMessages"));
         this.readMessages = metricRegistry.meter(name(this.getClass(), "readMessages"));
+        this.writeDiscardedMessages = metricRegistry.meter(name(this.getClass(), "writeDiscardedMessages"));
```
joschi: Shouldn't this simply be named "discardedMessages"?

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -376,31 +383,84 @@ public Entry createEntry(byte[] idBytes, byte[] messageBytes) {
...
-            LOG.debug("Wrote {} messages to journal: {} bytes, log position {} to {}",
-                    entries.size(), payloadSize, appendInfo.firstOffset(), lastWriteOffset);
-            writtenMessages.mark(entries.size());
             return lastWriteOffset;
         }
     }
 
+    private long flushMessages(List<Message> messages, long payloadSize) {
+        if (messages.isEmpty()) {
+            if (LOG.isDebugEnabled()) {
```
joschi: Minor nitpick: This guard condition is unnecessary. Log4j2/SLF4J will only log on debug level if it's activated.

Usually those guard conditions are useful, if the generated log message contains "expensive" computations, e. g. converting a byte array or a map of things into a string.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
bernd: I named it `writeDiscardedMessages` because we have metrics for `writtenMessages` and `readMessages`. Naming it simply `discardedMessages` doesn't indicate if it's on read or write.

joschi: Is there a difference between discard on read or on write? In my opinion it's always the messages "in-flight" that are being discarded (i. e. after read and before writing).

bernd: It's the other way around. We first write to the journal, then read. So in-flight means they are written to disk. A message can be discarded when writing to the journal and could be discarded when reading from the journal.

joschi: ✅ Alright, got it! Thanks for the explanation.

PR #2674, opened 2016-08-12
More decorator improvements

### LINKED ISSUE #2519: Message decorators add internal fields to decorated messages
### Problem description

When a message decorator is used, the internal message id field `_id` is shown in the fields list.

1. Add a pipeline decorator to the search result.
2. The field `_id` is shown in the sidebar.

### Environment

* Graylog Version: 2.1.0-SNAPSHOT
* Elasticsearch Version: n/a
* MongoDB Version: n/a
* Operating System: n/a
* Browser version: n/a

edmundoa:
Here are some other improvements in message decorators:

- Do not return internal fields in the field list. Fixes #2519
- Maintain decorators sort after editing one of them, by submitting the `order` field to the server
- Make decorator sidebar scrollable, allowing users to access all decorators if there are many or the screen height is not large enough
- Attempt to resolve configuration IDs from the type definition. This allows us to show names in the configuration values instead of IDs

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessorImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessorImpl.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.decorators;
 
+import org.graylog2.plugin.Message;
 import org.graylog2.plugin.decorators.SearchResponseDecorator;
 import org.graylog2.rest.models.messages.responses.DecorationStats;
 import org.graylog2.rest.models.messages.responses.ResultMessageSummary;
@@ -87,8 +88,14 @@ public class DecoratorProcessorImpl implements DecoratorProcessor {
 
     private Set<String> extractFields(List<ResultMessageSummary> messages) {
         return messages.stream()
-            .map(message -> message.message().keySet())
-            .reduce(new HashSet<>(), (set1, set2) -> { set1.addAll(set2); return set1; });
+                .map(message -> message.message().keySet())
+                .reduce(new HashSet<>(), (set1, set2) -> {
+                    set1.addAll(set2);
+                    return set1;
+                })
+                .stream()
+                .filter(field -> !Message.RESERVED_FIELDS.contains(field))
+                .collect(Collectors.toSet());
     }
 
     private SearchDecorationStats getSearchDecoratorStats(List<ResultMessageSummary> decoratedMessages) {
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/search/Decorator.jsx
```diff
@@ -33,18 +33,40 @@ const Decorator = React.createClass({
...
+  // Attempts to resolve ID values in the decorator configuration against the type definition.
+  // This allows users to see actual names for entities in drop-downs, instead of their IDs.
+  _resolveConfigurationIds(config) {
+    const typeConfig = this.props.typeDefinition.requested_configuration;
+    const resolvedConfig = {};
+    const configKeys = Object.keys(config);
+
+    configKeys.forEach(key => {
+      const configValues = (typeConfig[key] ? typeConfig[key].additional_info.values : undefined);
+      if (typeof configValues === 'object' && !Array.isArray(configValues)) {
```
dennisoelkers: Why not just `if (configValues && configValues[config[key]]) {`?

edmundoa: I was being too cautious in case some other configuration field was using `values` for something else. I'll change the code 👍

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessorImpl.java
```diff
@@ -87,8 +88,14 @@ public SearchResponse decorate(SearchResponse searchResponse, Optional<String> s
...
-            .map(message -> message.message().keySet())
-            .reduce(new HashSet<>(), (set1, set2) -> { set1.addAll(set2); return set1; });
+                .map(message -> message.message().keySet())
+                .reduce(new HashSet<>(), (set1, set2) -> {
+                    set1.addAll(set2);
+                    return set1;
+                })
+                .stream()
+                .filter(field -> !Message.RESERVED_FIELDS.contains(field))
+                .collect(Collectors.toSet());
```
dennisoelkers: This could be simplified to:

`return messages.stream()
                .flatMap(message -> message.message().keySet().stream())
                .filter(field -> !Message.RESERVED_FIELDS.contains(field))
                .collect(Collectors.toSet());`

which is easier to read imo.

edmundoa: Looks much better indeed. ✅

dennisoelkers:
LGTM! 👍

PR #2684, opened 2016-08-15
Improvements in raw message loader

edmundoa:
These are the improvements done in the raw message loader:

- Use 127.0.0.1 as default remote address, making the field optional
- Do not use a fieldset for message input, as it only contains a single input
- Mark message input field as optional
- Mark required fields as required, to enable browser validation


bernd:
LGTM :+1:

PR #2685, opened 2016-08-15
Allow users of MessageFields to disable field actions

edmundoa:
Somewhere in the changes made for the message decorators, we forgot to check the `disableFieldActions` prop in `MessageFields` when deciding whether to show the field actions or not. This PR add that back.


bernd:
LGTM :+1:

PR #2687, opened 2016-08-15
Refactoring of audit events

bernd:
## Description
- Merge audit `object` and `action` into `action`
- Create constants for each audit action
- Create `PluginAuditActions` interface to make audit actions pluggable
- Rename `AuditLog` annotation to the more generic `AuditEvent`
- Enable `captureRequestEntity` and `captureResponseEntity` by default in `AuditEvent`
- Create `excludedFields` in `AuditEvent` to make it possible to skip fields like passwords in the audit trail
- Add test to check that all REST resources annotated with `POST`, `PUT` or `DELETE` also have an 
`AuditEvent` annotation or a `NoAuditEvent` annotation if they should not be included in the audit trail
- Add Jersey model processor to check that all loaded REST resources annotated with `POST`, `PUT` and `DELETE` have a `AuditEvent` or `NoAuditEvent` annotation
- Add missing `AuditEvent` annotations to REST resources
- Remove now unused `Actions` class
- Rename `AuditLogger` interface to the more generic `AuditEventSender`
- Rename `auditlog` package to `audit`
- Use URNs for audit actors

## Motivation and Context
- Make it possible to iterate over all existing audit actions
- Ensure that all REST resources which change state have an `@AuditEvent` annotation
- Unify audit actions
- More generic naming


bernd:
Sorry, long PR is long... :confused:

garybot2:

garybot2:

graylog2-server/src/main/java/org/graylog2/audit/jersey/AuditEventModelProcessor.java
```diff
@@ -0,0 +1,111 @@
...
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Checks all POST, PUT and DELETE resource methods for {@link AuditEvent} annotations and reports missing ones.
+ *
+ * It does not report methods which have a {@link NoAuditEvent} annotation.
```
kroepke: I would report `@NoAuditEvent` annotations that use an empty string here, but that's nitpicking.

kroepke: I would report `@NoAuditEvent` annotations that use an empty string here, but that's nitpicking.

graylog2-server/src/main/java/org/graylog2/audit/AuditEventSender.java
```diff
@@ -14,16 +14,16 @@
...
 
 import java.util.Map;
 
-public interface AuditLogger {
-    void success(String subject, String action, String object);
+public interface AuditEventSender {
+    void success(AuditActor actor, AuditEventType type);
 
-    void success(String subject, String action, String object, Map<String, Object> context);
+    void success(AuditActor actor, AuditEventType type, Map<String, Object> context);
```
kroepke: I'd add a couple of convenience methods here, to make the call sites a bit more concise:

With
```Java
default void success(String actor, String type, Map<String, Object> context) {
    success(AuditActor.user(actor), AuditEventType.create(type), context);
}
```
it would be much easier to read on the call sites.

There are a couple of more options, too, such as allowing to append a varargs of Object pairs to implicitly create the map (or just a single String, Object pair)

bernd: I can add a default method for `success(AuditActor, String, Map<String, Object>)`. I would not create one with a `String` argument for `AuditActor` because there are two different methods (`AuditActor.system(NodeId)` and `AuditActor.user(String)` that do different things. So this can easily break.

Creating `AuditEventType` from a string is fine because `AuditEventType.create()` will throw an error if the string is not in the correct format.

bernd:
Addressed review comments.

bernd:
Merging this on behalf of @kroepke who gave his "LGTM :+1:" offline.

Issue #2656, opened 2016-08-10
Client IP get refreshes only on logout


hc4:
I have SSO enabled and graylog behind reverse proxy.
The problem could be tested on Users page.

1. Open Users page - correct IP displayed for current user
2. Remove all trusted_proxies from graylog config and restart graylog
3. Users page still shows old user IP (but should be proxy ip)
4. Click "Log out"
5. User logs in automaticaly (thanks to SSO plugin)
6. Now proxy IP displayed as client ip for relogged user.
7. Return trsuted proxies to config and restart graylog.
8. Again correct client IP will be displayed only after relogging

## Your Environment
* Graylog Version: 2.1.0-beta3 snapshot

kroepke:
Weird, apparently not all session attributes are being updated when the session is validated. I'll investigate.

hc4:
Also strange thinks happens with last activity.
I have user, for which last activity was displayed "4 months ago".
Then he logged in using SSO, and last activity became "few seconds ago".
But after he closed browser, last activity was again "4 months ago"

kroepke:
@hc4 A user can have multiple sessions, for display purposes we get the latest one.
The user is probably configured so that her sessions never time out.
Also note that old sessions do not have their host stored with them, so previously existing sessions won't show the client IP.

hc4:
I've changed option session never expires, but this 4 months ago session still there :)
Also I think session should be validated against client ip.
i.e. You shouldn't be able to connect from another host with same session id.

And how can I clear all user's sessions?

hc4:
http://graylog/system/sessions shows only my session (despite I'm admin)
Can I see all sessions on server using REST API?

kroepke:
That is currently not implemented, no. It was meant to be in #2551 but that didn't make it into 2.1

kroepke:
Closer inspection of the framework code showed that this use case is not supported directly.
The session only saves the host where it was *originally initiated from*, not the current one. I suppose the intention was to be able to find sessions that were created from a different source than is currently being used.

We would need to save the current host differently, but that is out of scope for 2.1.

I'll leave it open so we can integrate it with #2551 and instead change the wording to make it clear that this is not the current host.

PR #2692, opened 2016-08-16
Add help text for session's client address

kroepke:
The auth framework does not support updating sessions beyond the last used timestamp.
Especially the host is not resettable after creation.

Users might expect the address to change, but if we want to expose that we need to save this information separately in the future.

This change merely informs the user about this circumstance and makes the headers available internally for some optimization regarding session extension (less DB requests).

Refs #2656

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
@@ -50,8 +50,6 @@ public class ShiroSecurityContextFilter implements ContainerRequestFilter {
     private Provider<Request> grizzlyRequestProvider;
     private final Set<IpSubnet> trustedProxies;
 
-    public static final String X_GRAYLOG_REMOTE_ADDRESS = "x-graylog-remote-address";
-
     @Inject
     public ShiroSecurityContextFilter(DefaultSecurityManager securityManager,
                                       Provider<Request> grizzlyRequestProvider,
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
@@ -20,6 +20,7 @@ import org.apache.shiro.authc.AuthenticationToken;
 import org.apache.shiro.authc.UsernamePasswordToken;
 import org.apache.shiro.mgt.DefaultSecurityManager;
 import org.apache.shiro.subject.Subject;
+import org.apache.shiro.util.ThreadContext;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
```diff
@@ -44,10 +45,13 @@
 
 @Priority(Priorities.AUTHENTICATION)
 public class ShiroSecurityContextFilter implements ContainerRequestFilter {
+    public static final String REQUEST_HEADERS = "REQUEST_HEADERS";
     private final DefaultSecurityManager securityManager;
     private Provider<Request> grizzlyRequestProvider;
     private final Set<IpSubnet> trustedProxies;
 
+    public static final String X_GRAYLOG_REMOTE_ADDRESS = "x-graylog-remote-address";
```
joschi: What is this constant being used for?

graylog2-web-interface/src/components/users/UserList.jsx
```diff
@@ -90,11 +105,13 @@ const UserList = React.createClass({
       </OverlayTrigger>);
     }
 
-    const roleBadges = user.roles.map((role) => <span key={role} className={`${UserListStyle.roleBadgeFixes} label label-${role === 'Admin' ? 'info' : 'default'}`} >{role}</span>);
+    const roleBadges = user.roles.map((role) => <span key={role}
+                                                      className={`${UserListStyle.roleBadgeFixes} label label-${role === 'Admin' ? 'info' : 'default'}`}>{role}</span>);
 
     let actions = null;
     if (user.read_only) {
-      const tooltip = <Tooltip id="system-user">System users can only be modified in the Graylog configuration file.</Tooltip>;
```
joschi: Unnecessary whitespace change.

graylog2-web-interface/src/components/users/UserList.jsx
```diff
@@ -90,11 +105,13 @@ const UserList = React.createClass({
       </OverlayTrigger>);
     }
 
-    const roleBadges = user.roles.map((role) => <span key={role} className={`${UserListStyle.roleBadgeFixes} label label-${role === 'Admin' ? 'info' : 'default'}`} >{role}</span>);
```
joschi: Unnecessary whitespace change.

graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
kroepke: Apparently left over. If unused I'll remove it.

joschi:
LGTM. 👍

PR #2698, opened 2016-08-16
Fix token creation via API browser

### LINKED ISSUE #2668: Unable to generate api token via API browser
Trying to create token using api browser.
But error returned:
```
{
  "type": "ApiError",
  "message": "HTTP 415 Unsupported Media Type"
}
```

Looks like the problem is in Content-Type header of request:
`Content-Type:text/plain;charset=UTF-8`

But according to [sources](https://github.com/Graylog2/graylog2-server/blob/920cd37045709d0e1b9b6fb1e8ea73e1914021c4/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java#L93) Users api consumes json:
```
@RequiresAuthentication
@Path("/users")
@Consumes(MediaType.APPLICATION_JSON)
@Api(value = "Users", description = "User accounts")
public class UsersResource extends RestResource {
```
and [generateNewToken](https://github.com/Graylog2/graylog2-server/blob/920cd37045709d0e1b9b6fb1e8ea73e1914021c4/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java#L443) api doesn't have `consumes` attribute at all.

## Your Environment
* Graylog Version: 2.1.0-beta.3-SNAPSHOT+8d14878

bernd:
Add a placholder API param to the token creation resource method.

Fixes #2668

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/roles/RolesResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/roles/RolesResource.java
@@ -215,7 +215,7 @@ public class RolesResource extends RestResource {
     @AuditLog(object = "role membership", captureResponseEntity = true)
     public Response addMember(@ApiParam(name = "rolename") @PathParam("rolename") String rolename,
                               @ApiParam(name = "username") @PathParam("username") String username,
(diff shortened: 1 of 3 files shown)


joschi:
LGTM. 👍

PR #2704, opened 2016-08-17
Ensure that rest_transport_uri can override the URI scheme of rest_listen_uri

### LINKED ISSUE #2680: Graylog2 provides wrong API URI after upgrade to 2.1 beta3
## Expected Behavior

Graylog 2 should provide the API URI with https, if rest_transport_uri is set with https URI.

## Current Behavior

Today I tried upgrading from Graylog 2.1 beta-2 to beta-3 without changing any settings. Currently our rest_transport_uri is set to https://graylog:443 and it worked fine for every version until beta-3.

We use SSL Offloading with a A10 loadbalancer and the API and Web endpoint of graylog do not use SSL themselves, so web_enable_tls is not set. They are available via different DNS names, and do not use the same IP address on the loadbalancer, but use the same servers.  When upgrading to beta-3, I get an error message on the web interface, that the graylog-server could not connect to the API endpoint "http://graylog:443". So it changes https to http, which is not working. Removing the 443 results in a HTTP code 301, since it automatically redirects to https, but this is also not working for me.

I am not quite sure if this is an issue with the new beta version and URI handling, or if this is simply a config error on my side. 

1. set rest_transport_uri to https://
2. do not have tls enabled for graylog, use SSL offloading
3. upgrade to beta-3
4. browse the webfrontend

## Context

Since I was testing on a preview instance, there was no real impact and reverting to beta-2 fixed the issue for me. We plan on upgrading our live environment to 2.1 once it has a stable release.

## Your Environment

* Graylog Version: Graylog 2.1.0-beta.2+ffa3355 and Graylog 2.1.0-beta.3+b817da4
* Elasticsearch Version: 2.3.2
* MongoDB Version: 3.0.7
* Browser version: Chrome 52.0.2743.116 (64-bit)

joschi:
Fixes #2680

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
@@ -183,7 +183,7 @@ public abstract class BaseConfiguration {
             LOG.warn("\"{}\" is not a valid setting for \"rest_transport_uri\". Using default [{}].", restTransportUri, getDefaultRestTransportUri());
             return getDefaultRestTransportUri();
         } else {
-            return Tools.normalizeURI(restTransportUri, getRestUriScheme(), GRAYLOG_DEFAULT_PORT, "/");
+            return Tools.normalizeURI(restTransportUri, restTransportUri.getScheme(), GRAYLOG_DEFAULT_PORT, "/");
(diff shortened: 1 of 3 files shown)


bernd:
LGTM :+1:

PR #2706, opened 2016-08-17
Reloading `CurrentUserStore` when updated user is the current user.

### LINKED ISSUE #2705: Editing the current logged in user does not update the data
## Expected behaviour
When the user logged in the web interface updates their data (i.e. name), the web interface should be updated to reflect the change.

## Current behaviour
Updating the current user does not refresh the information, so the old state is kept until the page is refreshed or there is an action that fetches the user again, like clicking on the Graylog logo

## Possible solution
Update the current logged in user after editing the user information, so the information seen on the screen is accurate. Probably a call to `CurrentUserStore.reload()` should suffice.

## Steps to reproduce (for bugs)
1. Log in to Graylog
2. Click on the user name in the navigation bar, and then edit profile
3. Update your username and press "Update user"
4. The name in the navigation bar is not updated

## Your environment
* Graylog Version: 2.1.0-beta.4-SNAPSHOT (f167fb00d3cd1e512507bfb36d22934e31fe4ac7)

dennisoelkers:
## Description
The `UserForm` component now triggers a reload of the `CurrentUserStore` when it is updating the user which is currently logged in.

## Motivation and Context
Before this change, the current user object was stale after updating it.

## Types of changes

## Checklist:

Fixes #2705.

### DIFF
--- a/graylog2-web-interface/src/components/users/UserForm.jsx
+++ b/graylog2-web-interface/src/components/users/UserForm.jsx
@@ -111,6 +111,9 @@ const UserForm = React.createClass({
       if (this.isPermitted(this.state.currentUser.permissions, ['users:list'])) {
         this.props.history.replaceState(null, Routes.SYSTEM.AUTHENTICATION.USERS.LIST);
       }
+      if (this.props.user.username === this.state.currentUser.username) {
+        CurrentUserStore.reload();
+      }
     }, () => {
       UserNotification.error('Could not update the user. Please check your logs for more information.', 'Updating user failed');
     });


graylog2-web-interface/src/components/users/UserForm.jsx
```diff
@@ -111,6 +111,9 @@ const UserForm = React.createClass({
       if (this.isPermitted(this.state.currentUser.permissions, ['users:list'])) {
         this.props.history.replaceState(null, Routes.SYSTEM.AUTHENTICATION.USERS.LIST);
       }
+      if (this.props.user.username === this.state.currentUser.username) {
+        CurrentUserStore.reload();
+      }
```
edmundoa: I was wondering, shouldn't we do the same when updating permissions or roles? If so, maybe we should trigger the reload from the (ancient) `UsersStore`, or even take the time to convert it to a Reflux store. We should probably wait for that until 2.1 is out, though.

dennisoelkers: Yes, I think we need to. I would vote for refactoring the `UsersStore` into a Reflux store. :+1:

edmundoa: Ok, I'll create a follow up task for that 👍

edmundoa:
LGTM 👍

PR #2711, opened 2016-08-18
Add a prop to display/hide the page selector

edmundoa:
In some circumstances, it may make sense to make use of `PaginatedList` but not display a page size select input.

This PR adds a prop to control whether it is hidden or shown (the default is shown).


bernd:
LGTM :+1:

Issue #2623, opened 2016-08-08
Nothing displayed on Indexer overview page if elastic cluster not available


hc4:
## Expected Behavior
Should be displayed information about cluster unavailability

## Current Behavior
It's just "Loading...".
And in console logged many errors:
`There was an error fetching a resource: cannot GET /system/indexer/overview (500). Additional information: all shards failed`

## Your Environment
* Graylog Version: 2.1.0-beta2

PR #2713, opened 2016-08-18
Handle indexer cluster down in web interface

edmundoa:
As described in #2623, when ES is not reachable, going to the indices page in the web interface only renders a grey page with a loading spinner.

This PR handles the situation when ES is not available and displays a warning dialogue with some information.

I also saw that the ES cluster health was not being refreshed in the system overview page, and now we update the information when the `IndexerClusterHealth` component is mounted.


joschi:
LGTM. 👍

PR #2736, opened 2016-08-24
Disable editing controls for decorator if user lacks permissions.

### LINKED ISSUE #2730: Decorators and non-admin users
We need to check the user permissions when displaying the decorators controls in the UI.

## Expected Behavior
Non-admin users can only see controls in the UI to apply, edit, or delete decorators when they have the right permissions for it.

## Current Behavior
Controls to apply, edit, and delete decorators are visible for all users. This is only a display problem, as the backend actions check the user permissions before making any changes in the system.

## Possible Solution
Check user permissions in the frontend, to only display the controls for the actions a user can perform.

1. Create a non-admin user and give them (read or write) access to a stream
2. Log-in with that user and go to the stream
3. Open the decorators tab and try to apply, edit or delete a decorator
4. You will see a 403 in the console, but no further indication of what happened

## Your Environment

* Graylog Version: 2.1.0-beta.5-SNAPSHOT (606e32947403d1630c040d83510b15a65062ac1d)

dennisoelkers:
## Description
This change does the following when a user lacks the
'decorators:edit:${streamId}' permission:

  * disables items in action menu, when user lacks
  * disables controls to create decorator
  * disables ability to reorder decorators by drag & drop

Fixes #2730.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/search/Decorator.jsx
+++ b/graylog2-web-interface/src/components/search/Decorator.jsx
@@ -8,9 +8,12 @@ import { ConfigurationForm, ConfigurationWell } from 'components/configurationfo
 
 import StoreProvider from 'injection/StoreProvider';
 const DecoratorsStore = StoreProvider.getStore('Decorators');
+const CurrentUserStore = StoreProvider.getStore('CurrentUser');
 
 import ActionsProvider from 'injection/ActionsProvider';
 const DecoratorsActions = ActionsProvider.getActions('Decorators');
+
+import PermissionsMixin from 'util/PermissionsMixin';
 
 import DecoratorStyles from '!style!css!components/search/decoratorStyles.css';
 
@@ -19,7 +22,7 @@ const Decorator = React.createClass({
     decorator: React.PropTypes.object.isRequired,
     typeDefinition: React.PropTypes.object.isRequired,
   },
-  mixins: [Reflux.connect(DecoratorsStore)],
+  mixins: [Reflux.connect(DecoratorsStore), Reflux.connect(CurrentUserStore), PermissionsMixin],
   componentDidMount() {
     DecoratorsActions.available();
   },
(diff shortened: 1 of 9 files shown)


graylog2-web-interface/src/components/search/Decorator.jsx
```diff
@@ -63,22 +66,32 @@ const Decorator = React.createClass({
...
     return (
       <span className={DecoratorStyles.fullWidth}>
         <div className={DecoratorStyles.decoratorBox}>
           <h6 className={DecoratorStyles.decoratorType}>{decoratorType.name}</h6>
-          <DropdownButton id={`decorator-${decorator._id}-actions`} bsStyle="default" bsSize="xsmall" title="Actions" pullRight>
-            <MenuItem onSelect={this._handleEditClick}>Edit</MenuItem>
-            <MenuItem divider/>
-            <MenuItem onSelect={this._handleDeleteClick}>Delete</MenuItem>
-          </DropdownButton>
+          {decoratorActionsMenu}
```
edmundoa: For better consistency with other parts of the UI, I would hide the whole drop-down menu, as both actions depend on the same permission to be set.

edmundoa:
Could you please rebase your changes with the current master? Thank you!

graylog2-web-interface/src/components/search/Decorator.jsx
dennisoelkers: My general rule of thumb for hiding functionality vs. disabling it is that:

  * hiding functionality which the user will never be able to use (i.e. requiring admin privileges)
  * disabling functionality which is not available in this context, but might be at some other point

This way it is transparent for the user (although in this case we could add some helpful tooltips) that the functionality exists, but is not usable (yet) and the layout stays the same.

We do not do that consistently, but that's not because the way we do it now is the right one in most places, but because we do not have a convention for this yet. :)

edmundoa: As far as I remember, we hide most buttons users have no permissions to use, but there's no convention, that is very true.

edmundoa: By the way, I didn't want to imply the "old" way was the right one, just more consistent with what I had in my head. So 👍 on this from my side.

dennisoelkers: Yeah, I agree with you on that one and your remark is totally correct. It's just a matter of balancing consistency and choosing the strategy which seems to make the most sense. I am also unsure which one's more important here. :)

edmundoa:
LGTM 👍

PR #2745, opened 2016-08-25
Invalidating widget result cache cluster wide when a widget changes.

### LINKED ISSUE #2732: dashboard / widget in load balanced setup
## Expected Behavior
If you run a widget on a dashboard and change the search relative time this should be propagated to all nodes. 

## Current Behavior
The search relative time is not set on all nodes instant. This leads to flipping widgets for different users that are not usable.

## Possible Solution
Propagate the changed relative time via cluster event to all nodes in the cluster.

## Steps to Reproduce
1. Loadbalanced Graylog (two nodes behind a load balancer)
2. Create a Dashboard Widget
3. Load the Widget Configuration and change time relative time

## Context
Having a Load balanced Environment of Graylog Nodes to shift the load of the Web Interface. No sticky connection let the people "jump" between the configured nodes. But every request is then served with the new (changed) relative time or the old that is not propagated to all nodes instant. 

## Your Environment

* Graylog Version: 2.X

dennisoelkers:
## Description
Before this, a change to a widget done on one node was not reflected
correctly on other nodes, due to the widget being cached in the
WidgetResultsCache.

This change now creates and tracks widget update events cluster-wide and
invalidates the local caches accordingly.

Fixes #2732.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/config.js
+++ b/graylog2-web-interface/config.js
@@ -1,5 +1,5 @@
 window.appConfig = {
-  gl2ServerUrl: 'http://localhost:12900/',
+  gl2ServerUrl: 'http://localhost:9000/api',
   gl2AppPathPrefix: '',
   rootTimeZone: 'Europe/Berlin',
 };
--- a/graylog2-web-interface/config.js
+++ b/graylog2-web-interface/config.js
@@ -1,5 +1,5 @@
 window.appConfig = {
-  gl2ServerUrl: 'http://localhost:9000/api',
+  gl2ServerUrl: 'http://localhost:12900/',
   gl2AppPathPrefix: '',
   rootTimeZone: 'Europe/Berlin',
 };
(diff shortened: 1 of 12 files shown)


graylog2-web-interface/config.js
```diff
@@ -1,5 +1,5 @@
 window.appConfig = {
-  gl2ServerUrl: 'http://localhost:9000/api',
+  gl2ServerUrl: 'http://localhost:12900/',
```
joschi: This shouldn't be committed.

joschi:
I don't think it is a good idea to publish the events in `DashboardsResource` and would prefer this happening in `DashboardServiceImpl`.

This being said, I'll merge this PR to get the functionality into Graylog 2.1.0 and moving the event publishing into `DashboardServiceImpl` would require a larger refactoring of said class.

PR #2746, opened 2016-08-25
Audit event cleanup

bernd:
Removes a few audit events when running index retention and rotation. Only emit events when indices are actually deleted/closed.


edmundoa:
LGTM 👍

PR #2748, opened 2016-08-26
Fixing linter hints in three of the top violating files.

dennisoelkers:


joschi:
Code looks sane and I couldn't find any regressions while manually testing the web interface. 👍

PR #2756, opened 2016-08-29
Allow child elements in the search form

bernd:
This is needed to be able to add more buttons and other elements.


edmundoa:
LGTM 👍

PR #2757, opened 2016-08-29
make key_prefix configuration optional

### LINKED ISSUE #2755: Key prefix field in JSON extractor configuration is mandatory but should not be
## Expected Behavior
When configuring a JSON extractor the `key prefix`field should only be used when desired by the user. 

## Current Behavior

A JSON extractor can not be created without entering a `key prefix`. This behavior is new in Graylog 2.1, previous versions do not have this field. When importing an extractor from an older version the extractor works without having `key prefix`filled out unless it is edited.

## Possible Solution

Make key prefix field not mandatory

1. Open a message and go to JSON extractor configuration 
2. Fill out configuration
3. Leave out `key prefix`
4. Click `create extractor`
5. You receive a message asking you to fill out the `key prefix` and can not create the extractor

## Your Environment

* Graylog Version: 2.1.0-rc
* Browser version: Firefox 48

kroepke:
The key_prefix defaults to an empty string so that the input field should not be required.

fixes #2755


bernd:
LGTM :+1:

PR #2762, opened 2016-08-31
Actually use the bluebird promise in FetchProvider

bernd:
Previously this was returning a native Promise when SessionStore.isLoggedIn() returned false.

Fixes NodesStore which failed when trying to call finally() on a native 

pasted image at 2016_08_31 16_21
promise.


edmundoa:
LGTM 👍

PR #2765, opened 2016-09-01
Upgrade to Elasticsearch 2.4.0

joschi:
## Description

Upgrade the internal ES node to Elasticsearch 2.4.0 and Jackson 2.8.2.

* https://www.elastic.co/blog/elasticsearch-2-4-0-released
* https://www.elastic.co/guide/en/elasticsearch/reference/2.4/release-notes-2.4.0.html

## Types of changes

- [X] Dependency update

## Checklist:

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/indexer/indices/IndicesTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/indexer/indices/IndicesTest.java
@@ -215,7 +215,7 @@ public class IndicesTest {
             "_source", ImmutableMap.of("enabled", false),
             "properties", ImmutableMap.of("message",
                 ImmutableMap.of(
-                    "type", "binary",
+                    "type", "string",
                     "index", "not_analyzed")));
         assertThat(client.preparePutTemplate(templateName)
                 .setTemplate(indices.allIndicesAlias())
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog2/indexer/indices/IndicesTest.java
```diff
@@ -215,7 +215,7 @@ public void testCreateOverwritesIndexTemplate() throws Exception {
             "_source", ImmutableMap.of("enabled", false),
             "properties", ImmutableMap.of("message",
                 ImmutableMap.of(
-                    "type", "binary",
+                    "type", "string",
```
bernd: Is there any explanation in the ES changelog why this is needed?

joschi: It looks like this is related to elastic/elasticsearch#17567 and elastic/elasticsearch#17568.

The test will fail with the following exception if the field type changed:
```
MapperParsingException[Failed to parse mapping [message]: mapper [message] has different [index] values from other types of the same index
]; nested: IllegalArgumentException[mapper [message] has different [index] values from other types of the same index];
```

bernd:
LGTM :+1:

PR #2782, opened 2016-09-06
Fix 404s on IE 11 using compatibility view

### LINKED ISSUE #2768: Error 404 search with IE 11
## Expected Behavior

When we search someone with IE 11, we have a 404 error. Before 2.1, we had not this problem.
image

1. Search someone
2. Click on search
3. Error 404

## Context

We use IE 11 on production... We know it's not the better browser but it's the most easier update in production.

## Your Environment

* Graylog Version: 2.1
* Elasticsearch Version: 2.3
* MongoDB Version: 3.2.7
* Operating System: RHEL 6.6
* Browser version: IE 11

edmundoa:
When Graylog is running on an intranet, IE 11 may be requesting the site in compatibility view, depending on the organization and user settings. That applies even when properly setting the X-UA-Compatible header and meta tags.

Making requests in compatibility view, IE does not include either `text/html` or `application/xhtml+xml` MIME types in the `Accept` header, which are the values we checked in order to deliver the web interface HTML document when requesting a resource that doesn't exist (i.e. when clicking on a page number on the search page).

Unfortunately, the way legacy IEs work is also odd in the way it includes `Accept` headers on requests, being `*/*` the only one that is always included. You can read more about it here:
https://blogs.msdn.microsoft.com/ieinternals/2009/07/01/ie-and-the-accept-header/

To solve this issue, these changes relax the media types comparison on `WebAppNotFoundResponseFilter`, by also serving the web interface HTML document if any of the request headers is compatible with `text/html` or `application/xhtml+xml`. In that way we also consider that requests with a wildcard MIME type will be able to handle the HTML document, but we still avoid serving HTML if the request specified another accepted MIME type.

Fixes #2768

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/filter/WebAppNotFoundResponseFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/filter/WebAppNotFoundResponseFilter.java
@@ -47,7 +47,8 @@ public class WebAppNotFoundResponseFilter implements ContainerResponseFilter {
         final Response.StatusType responseStatus = responseContext.getStatusInfo();
         final String requestPath = requestContext.getUriInfo().getAbsolutePath().getPath();
         final List<MediaType> acceptableMediaTypes = requestContext.getAcceptableMediaTypes();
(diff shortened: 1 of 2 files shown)


edmundoa:
The changes also need to be applied into master.

Issue #2795, opened 2016-09-07
CSV export does not use current search time range


edmundoa:
## Expected Behavior
When doing a search, I expect the CSV export values to contain the same time range all other search elements use, even if I modify the search bar time range but didn't execute the search.

## Current Behavior
Modifying the time range in the search bar without executing the search will use the new time range on the CSV export.

## Possible Solution
Rely on the data coming from `SearchStore.getOriginalSearchURLParams()` instead of getting the updated configuration from the store.

## Steps to Reproduce (for bugs)
1. Do a search, for instance a relative search for the last 5 minutes
2. Change the time range to another one, e.g. relative search in the last 1 day
3. Export CSV results
4. Results contain data from last day, and not only the last 5 minutes

## Your Environment

* Graylog Version: Graylog 2.1.0

Issue #2770, opened 2016-09-02
Search Query Flashes Login Prompt Briefly


eroji:
After upgrading to Graylog 2.1.0, all search queries flashes the login prompt briefly as it reloads the screen with the results.

We are using Graylog on an input server and a web frontend server set as slave and serving only the web interface using Apache proxy to HTTPS. I verified that even with no Apache proxy, the behavior is the same.

## Steps to Reproduce (for bugs)
Click on the magnifying glass or type in parameter to perform the search

## Your Environment
* Graylog Version: 2.1.0
* Elasticsearch Version: 2.0.5
* MongoDB Version: 3.2
* Operating System: CentOS 6.7
* Browser version: Chrome 52 / IE 11.0

twgraham:
I was getting this issue before 2.1, though I think it may be more prominent in 2.1. At first I put it down to latency since I was trying to hit my Graylog cluster on east coast USA from Australia, but now I get it when running as local docker environment too.

nilroy:
I have not seen this issue in 2.0.x but I upgraded to 2.1.0 and getting this issue.

edmundoa:
The flashing is somehow intended when you do a full reload of the page (or there is a redirection that triggers it, like the form that is submitted when searching). Anyway, I think we can mitigate at least part of the annoyance of that flashing:

- Remove the unnecessary redirections or submits we do in the web interface (#2488)
- Change the page we display while validating the user's credentials (responsible of this flashing), so it doesn't look exactly like the login form

I think the second change is small enough that we can put it in a patch release, and it can already avoid some confusion around this.

edmundoa:
This has been fixed in #2802

Issue #2488, opened 2016-07-14
app reloads on search


colmaengus:
### Problem description
app reloads when you click search 

### Steps to reproduce the problem

1. Navigate to Search tab in application
2. Click the search icon and the whole application reloads rather than using XHR to perform search

### Environment

* Graylog Version: 2.0.3-2
* Elasticsearch Version: 
* MongoDB Version:
* Operating System: docker (brower on windows)
* Browser version: chrome 51.0.2704.103 m

gruselglatz:
We are facing the same problems since 2.0.3.
Searches now take 10+ seconds, this makes it really unusable.
It doesn't depends on the amount of messages, i only have 50 Messages in this setup so far.

Environment

Graylog Version: 2.0.3-2
Elasticsearch Version:
MongoDB Version:
Operating System: CentOS 7
Browser version: chrome 50.0.2661.102 m

When it's helpful i could provide you with our timeline-log.

edmundoa:
Hi @gruselglatz,

I wonder if there is some issue in your setup when loading the page takes more than 10 seconds. Did you check where the time is being used?

gruselglatz:
I used the Chrome TimelineTool and i could share it with you, if it would be helpful. I'm not a Webdeveloper, so i can't read much out of it.

edmundoa:
@gruselglatz That would be great. I'm not sure how helpful it would be, but at least I can take a look and see if I can extract something from it. Thank you!

gruselglatz:
@edmundoa I've sent it to you via Email. I hope you could see something.

edmundoa:
@gruselglatz email received, thank you!

gruselglatz:
UDAPTE: After i dismissed the "First Start Help" the Search now loads in half the time: 4-6 seconds.

`Scripts name` **Time**
`https://xxxx.local/plugin/org.graylog.plugins.pipelineprocessor.ProcessorPlugin/plugin.org.graylog.plugins.pipelineprocessor.PipelineProcessorPlugin.124facd49c87779f691b.js`   **1,27s**

`https://xxxx.local/plugin/org.graylog.plugins.map.MapWidgetPlugin/plugin.org.graylog.plugins.map.MapWidgetPlugin.3a6e86e960b80407f74e.js`  **1,37s**

[...]

PR #2798, opened 2016-09-08
Modify actions in search page triggering a page reload

edmundoa:
## Description
Many actions in the search page trigger a full page reload, either because they use links to navigate to another page, or because they submit a form to the server.

That was needed in previous versions of the web interface (prior 2.0) but now can be replaced by a combination of react-router and a refresh of the reflux stores involved in the search. This has the advantage of avoiding the page reload, avoiding asking the server for the web interface assets (as described in #2488), and also avoiding flashes with the login page (as described in #2770).

This PR takes care of modifying some of those actions to avoid page reloads when:
- Executing searches
- Using search pagination
- Changing histogram resolution
- Selecting saved searches
- Changing message sorting

This PR intentionally does not take care of:
- Refactoring the code around searches (specially `SearchStore`), as it is too risky for a bugfix release. This is something we can move into 2.2
- Avoiding page reload when using surrounding searches. The main problem here is to reset the search view, e.g. scroll to top, reset page number, and collapse expanded message. Doing those things would require more changes that I don't think are worth pursuing right now. We can take care of fixing it in a follow-up task

## Motivation and Context
The changes were motivated by #2488, and they also fix that issue.

## How Has This Been Tested?
I have checked that these search elements still work after the changes:
- [X] Widgets use right search params
- [X] Save searches use right search params
- [X] CSV export uses right search params (but didn't solve the issue described in #2795)
- [X] Field analyzers use right params
- [X] Searching from sources tab works
- [X] Searching in streams works
- [X] Search live-loading works
- [X] Show surrounding messages works
- [X] Sorting works
- [X] Field selection works

### DIFF
--- a/graylog2-web-interface/src/components/search/SearchResult.jsx
+++ b/graylog2-web-interface/src/components/search/SearchResult.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import Immutable from 'immutable';
 import { Col, Row } from 'react-bootstrap';
 
+import { Spinner } from 'components/common';
 import { LegacyHistogram, NoSearchResults, ResultTable, SearchSidebar } from 'components/search';
 
 import StoreProvider from 'injection/StoreProvider';
@@ -23,6 +24,7 @@ const SearchResult = React.createClass({
     nodes: React.PropTypes.instanceOf(Immutable.Map),
     permissions: React.PropTypes.array.isRequired,
     searchConfig: React.PropTypes.object.isRequired,
+    loadingSearch: React.PropTypes.bool,
   },
 
   getDefaultProps() {
@@ -142,6 +144,12 @@ const SearchResult = React.createClass({
                          permissions={this.props.permissions} searchInStream={this.props.searchInStream} />
       );
     }
+
+    let loadingIndicator;
+    if (this.props.loadingSearch) {
+      loadingIndicator = <div className="loading-indicator"><Spinner text="Updating search results..."/></div>;
+    }
+
     return (
       <Row id="main-content-search">
         <Col ref="opa" md={3} sm={12} id="sidebar">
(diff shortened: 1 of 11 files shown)


graylog2-web-interface/src/components/search/SearchResult.jsx
```diff
@@ -144,6 +144,12 @@ const SearchResult = React.createClass({
                          permissions={this.props.permissions} searchInStream={this.props.searchInStream} />
       );
     }
+
+    let loadingIndicator;
+    if (this.props.loadingSearch) {
+      loadingIndicator = <div className="loading-indicator"><Spinner text="Updating search results..."/></div>;
```
dennisoelkers: I think it makes sense to extract this into a separate component straight away, that we can reuse in different places too, including moving its styling from the global stylesheet to a local one.

graylog2-web-interface/src/stores/search/SearchStore.ts
```diff
@@ -332,14 +333,18 @@ class SearchStore {
         var searchURLParams = this.getOriginalSearchURLParams();
         searchURLParams = searchURLParams.set("width", this.width);
         searchURLParams = searchURLParams.set(param, value);
-        URLUtils.openLink(this.searchBaseLocation("index") + "?" + Qs.stringify(searchURLParams.toJS()), false);
```
dennisoelkers: As `URLUtils.openLink` is not used anymore, can you please remove it?

dennisoelkers:
Besides those two remarks, it looks good to me.

graylog2-web-interface/src/components/search/SearchResult.jsx
edmundoa: Makes sense, will take care of it ✅

graylog2-web-interface/src/stores/search/SearchStore.ts
edmundoa: ✅

PR #2800, opened 2016-09-09
Updating bluebird to 3.4.x and superagent to 2.2.x.

dennisoelkers:
This PR is updating bluebird and (as it depends on it) superagent.

As bluebird now has a check for runaway promises (promises with a `.then()` handler that does not return anything) a runaway promise in the `ConfigurationStore` was detected immediately, that I fixed (along with some linter hints), although it was of no harm.


edmundoa:
I'm fine with that if at least we fix the `NodesStore`, because that one is spamming too much (it looks like every time that we fetch the list of nodes).

PR #2802, opened 2016-09-09
Do not display login form while loading

### LINKED ISSUE #2770: Search Query Flashes Login Prompt Briefly

edmundoa:
When loading Graylog, we currently display the login form while waiting for some resources to load, namely:
- Validate user session
- Load the current logged-in user

Those resources sometimes take a bit to load, and seeing the login form is confusing, as seen in #2770. This PR introduces a loading page that will avoid the confusion of seeing the login form on the screen.

I decided to use the same design used in our login page, as this provides a good transition between the loading and log-in page, which may be show directly before and/or after each other.

Fixes #2770.

### DIFF
--- a/graylog2-web-interface/src/pages/LoadingPage.jsx
+++ b/graylog2-web-interface/src/pages/LoadingPage.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { Col, Row } from 'react-bootstrap';
+
+import { Spinner } from 'components/common';
+
+import disconnectedStyle from '!style/useable!css!less!stylesheets/disconnected.less';
+import authStyle from '!style/useable!css!less!stylesheets/auth.less';
+
+const LoadingPage = React.createClass({
+  propTypes: {
+    text: React.PropTypes.string,
+  },
+
+  getDefaultProps() {
+    return {
+      text: 'Loading, please wait...',
+    };
+  },
+
+  componentDidMount() {
+    disconnectedStyle.use();
+    authStyle.use();
+  },
+
+  componentWillUnmount() {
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/pages/LoadingPage.jsx
```diff
@@ -0,0 +1,48 @@
...
+  },
+
+  componentWillUnmount() {
+    disconnectedStyle.unuse();
+    authStyle.unuse();
+  },
+
+  render() {
+    return (
+      <div>
```
dennisoelkers: Minor nitpick, do we need this `div`?

dennisoelkers:
Tested it in different scenarios, LGTM. I also like the way it improves logging in/full page refreshes a lot. Good work @edmundoa!

graylog2-web-interface/src/pages/LoadingPage.jsx
edmundoa: I don't think so, must be a leftover from some test. Will take a look 👍

PR #2803, opened 2016-09-09
Check in SearchPage if search is in progress, reuse promise then.

### LINKED ISSUE #2799: Live reload fires new request without checking if there is an existing one
## Expected Behavior

The live reload should wait for a response before executing another search request to avoid overloading the server and Elasticsearch cluster.

## Current Behavior

The live reload fires new search requests against the server according to the live reload interval even if a previous request didn't return yet. (i.e. caused by an overloaded Elasticsearch cluster or a high latency network connection to the Graylog server)

The following screenshot shows an example of this behavior.

image

## Possible Solution

- The client code should only issue a new search request if the previous one has returned
- It should also ensure to wait according to the selected live reload interval before running the next request

1. Enable live reload with a search query that takes longer than the selected live reload interval
2. Open the network tab in the developer console
3. See lots of pending search requests

## Your Environment

* Graylog Version: 2.1.1-SNAPSHOT+6e974b0
* Browser version: 53.0.2785.101

dennisoelkers:
If search parameters change, cancel promise first.

Fixes #2799.

### DIFF
--- a/graylog2-web-interface/src/pages/SearchPage.jsx
+++ b/graylog2-web-interface/src/pages/SearchPage.jsx
@@ -56,6 +56,9 @@ const SearchPage = React.createClass({
     const nextLocation = nextProps.location || {};
 
     if ((currentLocation !== nextLocation) || (currentLocation.search !== nextLocation.search)) {
+      if (this.promise) {
+        this.promise.cancel();
+      }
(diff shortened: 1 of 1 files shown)


edmundoa:
LGTM 👍

PR #2809, opened 2016-09-12
Use current search time configuration for CSV export

### LINKED ISSUE #2795: CSV export does not use current search time range

edmundoa:
Generate search CSV export link using the time range of the current search, instead of the current value of the time range inputs.

Fixes #2795


bernd:
LGTM :+1:

PR #2811, opened 2016-09-12
explicitly close okhttp response body, avoiding leak connection warning

chainkite:
## Description
As title

## Motivation and Context
I got warning below sometimes after I upgrade to 2.1.0 (around once per day)
```
WARN : okhttp3.OkHttpClient - A connection to https://versioncheck.graylog.com/ was leaked. Did you forget to close a response body?
```

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #2812, opened 2016-09-12
Properly close OkHttp Response objects to avoid resource leaks

joschi:
OkHttp Responses must be closed properly: https://github.com/square/okhttp/blob/parent-3.4.1/okhttp/src/main/java/okhttp3/ResponseBody.java#L37-L39

Refs #2811


joschi:
This PR should be cherry-picked into `2.1` when merged.

edmundoa:
LGTM 👍

sre-jon:
Does this fix address the issue that causes the browser to wait for available sockets?

joschi:
@sre-jon No. If you have found a bug in Graylog, please open another issue.

PR #2815, opened 2016-09-12
POM reorganization

joschi:
## Description

This PR extracts the dependency version information for Graylog from `org.graylog2:graylog2-parent` into Maven properties and introduces two new parent POMs, `org.graylog.plugins:graylog-plugin-parent` (for plugins without web components) and `org.graylog.plugins:graylog-plugin-web-parent` (for plugins with web components, requiring Node.js and npm), so that 3rd party projects like plugins could re-use this information without using `graylog2-parent` as Parent POM.

POM hierarchy

## Motivation and Context

Many plugins need to include dependencies that are directly or indirectly coupled to dependencies in `org.graylog2:graylog-parent`, such as additional Jersey or Jackson extensions.

The new parent POMs allow referring to all dependencies of Graylog by using them as parent POM (in this case, all defined properties are available as well).


joschi:
I've updated the PR and the description.

bernd:
LGTM :+1:

PR #2820, opened 2016-09-13
remove ldap settings check from authenticators

### LINKED ISSUE #2817: External authentication (SSO) requires LDAP
When using the SSO plugin, you also need to enable LDAP. Otherwise users are not being logged in. It throws "LDAP authentication is currently disabled." from https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/security/realm/SessionAuthenticator.java#L73

## Expected Behavior

I'd expect Graylog not to prevent other plugins (the SSO plugin here) to authenticate externally even when there is no LDAP configured. 

## Current Behavior

Login via SSO not possible due to error "LDAP authentication is currently disabled". Login form shown. 

It works after LDAP is enabled. In fact it's sufficient that LDAP is enabled, regardless if the configuration is actually working

Well …

1. Have SSO capable graylog and appropriate proxy.
2. Turn off LDAP on the LDAP configuration page
3. SSO Login doesn't work.
4.

* Graylog Version: v2.1.0+62db7e0

kroepke:
we no longer lock external accounts when ldap is disabled because other authenticators can also create external users now
if people require this we need to track which authenticator created the account in the first place, but that's too large a change for a bug fix release

fixes #2817

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/realm/AccessTokenAuthenticator.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/realm/AccessTokenAuthenticator.java
@@ -37,15 +37,12 @@ public class AccessTokenAuthenticator extends AuthenticatingRealm {
 
     private final AccessTokenService accessTokenService;
     private final UserService userService;
-    private final LdapUserAuthenticator ldapAuthenticator;
 
     @Inject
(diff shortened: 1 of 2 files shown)


kroepke:
needs backport to 2.1 branch after review, too

bernd:
LGTM :+1:

PR #2822, opened 2016-09-13
Add ClusterConfigService#get(String, Class<T>) method

bernd:
## Description

This adds a `getPayload()` method to the `ClusterConfigService` interface. It's needed if you have to convert the config payload on your own and cannot rely on the cluster config to convert it to an object.

## Motivation and Context

Using `@JsonSubTypes` does not work with `ClusterConfigService#get()` because we have to pass the concrete class to the method but the object mapper needs the super type.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/cluster/ClusterConfigService.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/cluster/ClusterConfigService.java
@@ -32,6 +32,17 @@ public interface ClusterConfigService {
     <T> T get(Class<T> type);
 
     /**
+     * Retrieve config payload of a certain type from the cluster configuration.
+     *
+     * Useful for custom handling of the config payload.
+     *
+     * @param type The {@link Class} of the Java configuration bean to retrieve.
+     * @param <T>  The type of the Java configuration bean.
+     * @return The payload of the requested type or {@code null} if it couldn't be retrieved.
+     */
+    <T> Object getPayload(Class<T> type);
+
+    /**
      * Retrieve Java class of a certain type from the cluster configuration or return a default value
      * in case that failed.
      *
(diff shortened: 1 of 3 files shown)


bernd:
An alternative would be to add a `ClusterConfigService#get(type, superType)` method. That would allow me to fix the json subtype usage.

This PR is more generic.

graylog2-server/src/main/java/org/graylog2/plugin/cluster/ClusterConfigService.java
```diff
@@ -32,6 +32,17 @@
...
     /**
+     * Retrieve config payload of a certain type from the cluster configuration.
+     *
+     * Useful for custom handling of the config payload.
+     *
+     * @param type The {@link Class} of the Java configuration bean to retrieve.
+     * @param <T>  The type of the Java configuration bean.
+     * @return The payload of the requested type or {@code null} if it couldn't be retrieved.
+     */
+    <T> Object getPayload(Class<T> type);
```
joschi: The generic `<T>` is not required in this case.

The method signature could simply be as follows:
```java
Object getPayload(Class<?> type);
```

joschi: I'd rather call this method `getRawPayload()` to make it very clear, what it's doing.

joschi: It should also be marked as `@Nullable`.

joschi:
@bernd
> An alternative would be to add a ClusterConfigService#get(type, superType) method. That would allow me to fix the json subtype usage.

I think I'd prefer that approach and introduce a method like:
```java
<T> T get(String key, Class<T> type);
```
Where `key` is the name of the configuration bean (currently the canonical name of the type by default) and `type` is the class it should be cast to.

The old `get(Class<T> type)` method could then simply delegate to `get(type.getCanonicalName(), type)`.

This would also solve your problem with polymorphic JSON, right?

bernd:
Updated the PR to use the `get(String key, Class<T> type)` version.

PR #2847, opened 2016-09-19
Update to babel 6 / simplify plugin dependency & webpack config management

dennisoelkers:
This change is:

 * Updating babel transpiler to 6.x
 * Converting babel config to new presets
 * Using `AppContainer` for `react-hot-loader` 3.x
 * Adding `add-module-exports` plugin to retain old export semantics
 * Making reexporting consistent with ES6 semantics
 * Changing webpack config generation strategy to add plugins to entry points and contexts instead of merging complete configs
 * Updating manifest IDs
 * Removing `DedupePlugin` because it breaks things currently

 * **Merging the `graylog-web-plugin` module into the server. It is not necessary to publish it anymore, as we are now building plugins along with the server repo, referencing the module directly from it. If old build style is used, it should use the module as it is currently published.**


graylog2-web-interface/webpack.combined.config.js
```diff
@@ -1,7 +1,13 @@
+const webpack = require('webpack');
 const glob = require('glob');
 const path = require('path');
 const merge = require('webpack-merge');
```
edmundoa: Small nitpick: `merge` is not used now, so we can remove the import.

graylog2-web-interface/webpack.config.js
```diff
@@ -152,7 +150,6 @@ if (TARGET === 'build') {
           except: ['$super', '$', 'exports', 'require'],
         },
       }),
-      new webpack.optimize.DedupePlugin(),
```
edmundoa: In the PR description, you mention that the `DedupePlugin` doesn't work with the changes: what are the implications? And, if we still need it, can we fix the issue so it works again?

graylog2-web-interface/package.json
```diff
@@ -90,9 +78,8 @@
     "less-loader": "^2.2.1",
     "mocha": "^2.3.4",
     "phantomjs-prebuilt": ">=1.9",
-    "react-hot-loader": "^1.3.0",
+    "react-hot-loader": "^3.0.0-beta.3",
```
edmundoa: What are the reasons to use a beta version here? I guess you already thought about it and have good reasons, but I was wondering why we want to upgrade to version 3 beta.

edmundoa: Sounds good to me, also seeing the rhythm of beta releases we will most likely use a stable version by the end of our release cycle 👍

dennisoelkers:
Closing in favor of #2906 & #2907.

PR #2864, opened 2016-09-20
Make ValidatorProvider a Singleton

bernd:
The Validator instance creation is quite expensive. Making this a Singleton reduced the CPU load by 50% and reduced the GC load from 5 GCs per second to 2 GCs per second when running a load test of the collector registration endpoint.

## Before

image
image

## After

image
image


graylog2-server/src/main/java/org/graylog2/shared/bindings/ValidatorModule.java
```diff
@@ -24,6 +24,6 @@
 public class ValidatorModule extends AbstractModule {
     @Override
     protected void configure() {
-        bind(Validator.class).toProvider(ValidatorProvider.class);
+        bind(Validator.class).toProvider(ValidatorProvider.class).asEagerSingleton();
```
joschi: It would be simpler to directly bind to the instance `Validation.buildDefaultValidatorFactory().getValidator()` instead of using a singleton provider which basically returns a static field. This way we could get rid of the `Provider<Validator>`.

https://github.com/google/guice/wiki/InstanceBindings

bernd:
Updated after review comment.

PR #2866, opened 2016-09-21
Avoid ObjectMapper injection into every RestResource subclass

bernd:
This removes a field injection of `ObjectMapper` into every `RestResource` subclass to avoid the injection overhead. This was only used in a few resources.


joschi:
@bernd Out of interest, how big is the overhead for injecting a single object (which also is a singleton and doesn't have to be instantiated)?

bernd:
> @bernd Out of interest, how big is the overhead for injecting a single object (which also is a singleton and doesn't have to be instantiated)?

I didn't measure that, sorry. This case is probably a micro optimization.

Issue #2871, opened 2016-09-21
Improve Alerting


dennisoelkers:
- [x] Refactor/Make Alert Conditions pluggable #2867
- [x] Remove Special Handling Of Email Alarm Callback #2868 
- [x] Migrate alert receivers to Email Alarm Callback #3120
- [ ] Enhance Alert Data Model #2869 
- [ ] Improve Alert Management Interface #2870

edmundoa:
Closed, as all related issues are also closed now.

Issue #2867, opened 2016-09-21
Refactor/Make Alert Conditions pluggable


dennisoelkers:
The purpose of this is to allow plugins to provide alert conditions which can be instantiated in the alerting config of a stream. Those entities provide their own alert checking strategy and a way to configure themselves dynamically.

- [x] Make Alert Conditions Pluggable
- [x] Allow dynamic configuration of alert conditions in web interface
- [x] Investigate Pipeline Usage For Pluggable Alert Conditions

PR #2873, opened 2016-09-21
Make alert conditions pluggable

dennisoelkers:
This is the first preparatory step to make alert conditions pluggable. It does:

- introduce a binder for alert conditions
- introduce a common factory to instantiate alert conditions
- add legacy bindings for ootb conditions
- lookup and use alert conditions factory instances in binder during instantiation of persisted alert conditions
- adapt tests

Refs #2871, #2867

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
@@ -76,8 +76,6 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
         this.grace = getNumber(this.parameters.get("grace")).orElse(0).intValue();
     }
 
-    protected abstract AlertCondition.CheckResult runCheck();
-
     @Override
     public String getId() {
         return id;
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
@@ -144,12 +144,6 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
         return grace;
     }
 
-    public static class NoSuchAlertConditionTypeException extends Exception {
-        public NoSuchAlertConditionTypeException(String msg) {
-            super(msg);
-        }
-    }
-
     public static class CheckResult implements AlertCondition.CheckResult {
 
         private final boolean isTriggered;
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
@@ -51,7 +51,7 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
 
     protected final String id;
     protected final Stream stream;
-    protected final Type type;
+    protected final String type;
     protected final DateTime createdAt;
     protected final String creatorUserId;
     protected final int grace;
@@ -59,7 +59,7 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
 
     private final Map<String, Object> parameters;
 
-    protected AbstractAlertCondition(Stream stream, String id, Type type, DateTime createdAt, String creatorUserId, Map<String, Object> parameters, String title) {
+    protected AbstractAlertCondition(Stream stream, String id, String type, DateTime createdAt, String creatorUserId, Map<String, Object> parameters, String title) {
         this.title = title;
         if (id == null) {
             this.id = UUID.randomUUID().toString();
@@ -83,7 +83,7 @@ public abstract class AbstractAlertCondition implements EmbeddedPersistable, Ale
         return id;
     }
 
-    public Type getType() {
+    public String getType() {
         return type;
     }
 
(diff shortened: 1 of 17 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionBindings.java
```diff
@@ -0,0 +1,43 @@
...
+import org.graylog2.alerts.types.FieldValueAlertCondition;
+import org.graylog2.alerts.types.MessageCountAlertCondition;
+import org.graylog2.plugin.alarms.AlertCondition;
+import org.graylog2.plugin.inject.Graylog2Module;
+
+public class AlertConditionBindings extends Graylog2Module {
+    @Override
+    protected void configure() {
+        MapBinder<String, AlertCondition.Factory> alertConditionBinder = alertConditionBinder();
+        installAlertConditionWithCustomName(alertConditionBinder,
```
bernd: Please add a `addAlertCondition()` (or similar) method to `PluginModule` like we have for alarm callbacks and others. You can then also just extend `PluginModule` and use that method in here.

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
```diff
@@ -51,15 +51,15 @@ public String toString() {
 
     protected final String id;
     protected final Stream stream;
-    protected final Type type;
+    protected final String type;
```
bernd: How about getting rid of the `Type` enum completely and using a static `String` field on each alert condition?

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -150,45 +141,33 @@ public AlertCondition fromPersisted(Map<String, Object> fields, Stream stream) t
...
+
+        return factory.create(stream, id, createdAt, creatorId, parameters, title);
     }
 
     @Override
-    public AbstractAlertCondition fromRequest(CreateConditionRequest ccr, Stream stream, String userId) throws AbstractAlertCondition.NoSuchAlertConditionTypeException {
+    public AlertCondition fromRequest(CreateConditionRequest ccr, Stream stream, String userId) throws AbstractAlertCondition.NoSuchAlertConditionTypeException {
         final String type = ccr.type();
         if (type == null) {
             throw new AbstractAlertCondition.NoSuchAlertConditionTypeException("Missing alert condition type");
```
bernd: This is the last usage of the `NoSuchAlertConditionTypeException`. Other places now use `IllegalArgumentException` via the `checkArgument()` in `createAlertCondition()`. I think we should either remove the exception class and use `IllegalArgumentException` or use the `NoSuchAlertConditionTypeException` instead of `IllegalArgumentException` in the other places to make it consistent.

bernd:
I was wondering if the `AbstractAlertCondition#runCheck()` method should be moved to the `AlertCondition` interface? The `AlertServiceImpl` is currently casting the condition to be able to access `#runCheck()`.

```java
public AlertCondition.CheckResult triggeredNoGrace(AlertCondition alertCondition) {
        LOG.debug("Checking alert condition [{}] and not accounting grace time.", this);
        return ((AbstractAlertCondition) alertCondition).runCheck();
}
```

dennisoelkers:
Yes, I am already doing that in a subsequent PR. But for completeness (as it actually belongs to the scope of this PR) I can move it over.

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
dennisoelkers: I would like to stick with it for now. It is used for legacy (builtin) alert conditions and removing it would provide no benefit to us currently.

I would vote for removing it later, when we go the extra mile of creating a migration job which transforms builtin alert condition types of persisted alert conditions to their full class name and we are handling their type just like all others.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
dennisoelkers: I would go for removing it, but I don't have a strong opinion about that.

graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionBindings.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
bernd: Aye, please do. Thanks!

dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -150,45 +141,31 @@ public AlertCondition fromPersisted(Map<String, Object> fields, Stream stream) t
...
-            throw new AbstractAlertCondition.NoSuchAlertConditionTypeException("Missing alert condition type");
-        }
-
-        final AbstractAlertCondition.Type alertConditionType;
-        try {
-            alertConditionType = AbstractAlertCondition.Type.valueOf(type.toUpperCase(Locale.ENGLISH));
-        } catch (IllegalArgumentException e) {
-            throw new AbstractAlertCondition.NoSuchAlertConditionTypeException("No such alert condition type: [" + type + "]");
-        }
+        checkArgument(type != null, "Milling alert condition type");
```
bernd: `Milling` -> `Missing` :smile:

Issue #2880, opened 2016-09-22
Support for multiple retention strategies


joschi:
Enable Graylog to store different types of log messages for different periods of time, based on streams.

https://graylog.ideas.aha.io/ideas/GL2E-I-356
# Steps
- [x] Create explicit default stream #2881
  - [ ] ~Add migration job adding the default stream as start page for users who don't have a start page set and have the required permissions for the default stream~
- [ ] ~Remove default output handling from `OutputBufferProcessor`~
- [x] Create "write targets" (combination of index prefix, rotation strategy, and retention strategy)
- [x] Allow to map streams onto write targets
- [x] Add migration job which creates write target from current rotation/retention config + index prefix
- [x] Make sure that Searches/IndexHelper classes uses correct index prefixes
- [x] Add Write Target management functionality to web interface (refactor indices page)
- [x] Add Stream &rarr; Write Target mapping functionality to web interface
- [x] Check if we need an UI element to mark one stream as a default stream. Might be needed if a user deletes the default stream. Otherwise there is no way back. If we create the possibility to mark a stream as the default stream, we have to ensure that **only one** stream can be a default stream.

## Open TODOs

- [x] Document stream REST API changes regarding the new mandatory `index_set_id` field - breaking change (in UPGRADING.rst) (@bernd #3341)
- [ ] ~Document indices resource API changes (in UPGRADING.rst)~ (not needed, no breaking changes)
- [x] Create unique index on `index_prefix` field to avoid "problems" (@bernd #3193)
- [x] Check if index template handling still works with multiple index sets (one template per index set?) (@bernd)
- [x] Make index prefix matching more strict. What happens if index-set-1 uses prefix `graylog` and index-set-2 uses prefix `graylog_yolo`? This is also a problem in index templates as they are using `graylog_*` to match indices! (@bernd #3193)
  - Idea: Not allowing `_` character in new index set prefixes should be enough
- [x] Rotation/retention strategies need to load the index set config from the database instead of using the old cluster config data
- [x] Check validation for `IndexSetSummary` and `IndexSetConfig` - currently it's possible to create an empty `index_prefix` field even though it's annotated with `@NotBlank` (#3188)
- [x] Make sure index template and index is created when a new index set is created (@bernd)
  - Will be created by the index rotation job that runs every 10 seconds
- [x] Check all "TODO 2.2" comments in graylog server and plugins and fix them (#3211, #3285 ongoing)
[...]

PR #2881, opened 2016-09-22
Create Default stream

joschi:
## Description

This PR introduces the "always matches" stream rule and a periodical creating a "default stream" in Graylog, which contains all newly ingested messages.

Refs #2880

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
@@ -18,8 +18,8 @@ package org.graylog2.periodical;
 
 import com.google.common.collect.ImmutableMap;
 import org.bson.types.ObjectId;
-import org.graylog2.database.NotFoundException;
 import org.graylog2.events.ClusterEventBus;
+import org.graylog2.plugin.cluster.ClusterConfigService;
 import org.graylog2.plugin.database.ValidationException;
 import org.graylog2.plugin.periodical.Periodical;
 import org.graylog2.plugin.streams.Stream;
@@ -29,6 +29,7 @@ import org.graylog2.streams.StreamImpl;
 import org.graylog2.streams.StreamRuleImpl;
 import org.graylog2.streams.StreamRuleService;
 import org.graylog2.streams.StreamService;
+import org.graylog2.streams.config.DefaultStreamCreated;
 import org.graylog2.streams.events.StreamsChangedEvent;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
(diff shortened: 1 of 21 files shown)


graylog2-web-interface/src/components/streamrules/StreamRuleForm.jsx
```diff
@@ -27,11 +27,12 @@ const StreamRuleForm = React.createClass({
     return this.props.streamRule;
   },
   FIELD_PRESENCE_RULE_TYPE: 5,
+  ALWAYS_MATCH_RULE_TYPE: 7,
   _resetValues() {
     this.setState(this.props.streamRule);
   },
   _onSubmit() {
-    if (this.state.type === this.FIELD_PRESENCE_RULE_TYPE) {
+    if (this.state.type === this.FIELD_PRESENCE_RULE_TYPE || this.state.type === this.ALWAYS_MATCH_RULE_TYPE) {
```
bernd: Please update the code so that there is no "value" form field when the always match rule is selected. (similar how it's done with the field presence rule)

graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public boolean masterOnly() {
+        return true;
+    }
+
+    @Override
+    public boolean startOnThisNode() {
+        try {
+            return streamService.load(Stream.DEFAULT_STREAM_ID) == null;
```
bernd: Since the default stream is recreated on every reboot, should we disallow the deletion of it in the UI? Otherwise a user might delete it and run into weird situations.

graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
```diff
@@ -0,0 +1,129 @@
...
+                        .put(StreamRuleImpl.FIELD_TYPE, StreamRuleType.ALWAYS_MATCH.getValue())
+                        .put(StreamRuleImpl.FIELD_FIELD, "timestamp")
+                        .put(StreamRuleImpl.FIELD_INVERTED, false)
+                        .put(StreamRuleImpl.FIELD_STREAM_ID, id)
+                        .put(StreamRuleImpl.FIELD_DESCRIPTION, "Match all messages")
+                        .build());
+        try {
+            streamService.save(stream);
+            streamRuleService.save(streamRule);
+            LOG.info("Successfully created default stream: {}", stream.getTitle());
```
bernd: We have to emit a `StreamsChangedEvent` here to make sure the stream router gets updated with the new default stream. This is needed because the periodical might run when the stream router has already been loaded.

graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
joschi: I thought about different approaches to this.

The preferred one would be to create a marker in the cluster config, so that the default stream is only created once.

It should still be possible for users to remove the default stream, if they wish so.

joschi: If addressed this by creating a marker in the cluster config.

PR #2891, opened 2016-09-26
Introduce IndexSet and IndexSetRegistry to wrap Deflector

bernd:
This replaces direct Deflector usage with IndexSetRegistry and IndexSet usage.

Add legacy implementations of both new interfaces that just use the current Deflector class for now.

Refs #2880

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
@@ -22,7 +22,7 @@ import java.util.Map;
 import java.util.Set;
 
 public interface IndexSet {
-    String[] getAllGraylogIndexNames();
+    String[] getManagedIndicesNames();
 
     String getWriteIndexAlias();
 
@@ -32,13 +32,13 @@ public interface IndexSet {
 
     String getCurrentActualTargetIndex() throws TooManyAliasesException;
 
-    Map<String,Set<String>> getAllGraylogDeflectorIndices();
+    Map<String,Set<String>> getAllDeflectorAliases();
 
     boolean isUp();
 
     boolean isDeflectorAlias(String index);
 
-    boolean isGraylogIndex(String index);
+    boolean isManagedIndex(String index);
 
     void setUp();
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
@@ -1,0 +1,50 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.indexer;
+
+import org.graylog2.indexer.indices.TooManyAliasesException;
+
+import java.util.Map;
+import java.util.Set;
+
+public interface IndexSet {
+    String[] getAllGraylogIndexNames();
+
+    String getWriteIndexAlias();
+
+    String getWriteIndexWildcard();
+
+    String getNewestTargetName() throws NoTargetIndexException;
+
+    String getCurrentActualTargetIndex() throws TooManyAliasesException;
+
+    Map<String,Set<String>> getAllGraylogDeflectorIndices();
+
+    boolean isUp();
+
+    boolean isDeflectorAlias(String indexName);
+
+    boolean isGraylogIndex(String index);
+
+    void setUp();
+
+    void cycle();
+
+    void cleanupAliases(Set<String> indices);
+
+    void pointTo(String shouldBeTarget, String currentTarget);
+}
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
@@ -36,7 +36,7 @@ public interface IndexSet {
 
     boolean isUp();
 
-    boolean isDeflectorAlias(String indexName);
+    boolean isDeflectorAlias(String index);
 
     boolean isGraylogIndex(String index);
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
@@ -21,20 +21,13 @@ import org.graylog2.indexer.indices.TooManyAliasesException;
 import java.util.List;
 import java.util.function.Consumer;
 
-public interface IndexSetRegistry {
+public interface IndexSetRegistry extends Iterable<IndexSet> {
     /**
      * Returns a list of all {@link IndexSet} instances.
      *
      * @return list of index sets
      */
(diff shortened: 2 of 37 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+
+    String getNewestTargetName() throws NoTargetIndexException;
+
+    String getCurrentActualTargetIndex() throws TooManyAliasesException;
+
+    Map<String,Set<String>> getAllGraylogDeflectorIndices();
+
+    boolean isUp();
+
+    boolean isDeflectorAlias(String indexName);
```
joschi: For consistency, the method argument should be named `index`.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+
+    String getWriteIndexAlias();
+
+    String getWriteIndexWildcard();
+
+    String getNewestTargetName() throws NoTargetIndexException;
+
+    String getCurrentActualTargetIndex() throws TooManyAliasesException;
+
+    Map<String,Set<String>> getAllGraylogDeflectorIndices();
```
joschi: Would this really return all index aliases (even the ones not in this index set)? Otherwise, the name might be misleading.

I think a small Javadoc comment would also be useful, as I literally have no idea, what the return value means.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+public interface IndexSet {
+    String[] getAllGraylogIndexNames();
+
+    String getWriteIndexAlias();
+
+    String getWriteIndexWildcard();
+
+    String getNewestTargetName() throws NoTargetIndexException;
+
+    String getCurrentActualTargetIndex() throws TooManyAliasesException;
```
joschi: "Current actual" seems to be saying the same thing, doesn't it?

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.indexer;
+
+import org.graylog2.indexer.indices.TooManyAliasesException;
+
+import java.util.Map;
+import java.util.Set;
+
+public interface IndexSet {
```
joschi: In general, I think this interface would vastly profit from Javadoc comments.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+    String[] getAllGraylogIndexNames();
```
joschi: Does this method return *all* Graylog-managed index names, including those not in this index set? If not, the method name might be misleading.

The return value could also be a `Set<String>` to make sure that each index name is only included once.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+
+    String getWriteIndexAlias();
+
+    String getWriteIndexWildcard();
```
joschi: What does this method do?

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -0,0 +1,50 @@
...
+
+    String getNewestTargetName() throws NoTargetIndexException;
```
joschi: Is this the current target index of the index alias? How is it different from `getCurrentActualTargetIndex ()`?

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+public interface IndexSetRegistry {
+    /**
+     * Returns a list of all {@link IndexSet} instances.
+     *
+     * @return list of index sets
+     */
+    List<IndexSet> getAllIndexSets();
```
joschi: Shouldn't this method return a `Set<IndexSet>` instead? Is the order of index sets important?

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Executes the given consumer function for every {@link IndexSet}.
+     *
+     * @param action the consumer function
+     */
+    void forEach(Consumer<IndexSet> action);
```
joschi: This seems to be not really necessary, as we can simply call `getAllIndexSets().forEach()`.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    boolean isUp();
+
+    /**
+     * Checks if the given index name is a current write index alias in any {@link IndexSet}.
+     *
+     * @param indexName the name of the index to check
+     * @return true when given index name is a current write index, false otherwise
+     */
+    boolean isCurrentWriteIndexAlias(String indexName);
```
joschi: Can there be more than one write index alias?

graylog2-server/src/main/java/org/graylog2/indexer/LegacyDeflectorRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+public class LegacyDeflectorRegistry implements IndexSetRegistry {
+    private final IndexSet indexSet;
+    private final List<IndexSet> indexSetRegistry;
```
joschi: We know that there is only 1 index set, so why wrap it into a `List<IndexSet>`?

graylog2-server/src/main/java/org/graylog2/indexer/LegacyDeflectorRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public List<IndexSet> getAllIndexSets() {
+        return indexSetRegistry;
+    }
+
+    @Override
+    public void forEach(Consumer<IndexSet> action) {
+        requireNonNull(action);
+        indexSetRegistry.forEach(action);
```
joschi: There is only one `IndexSet`, why call `forEach` and not simply `action.accept(indexSet)`?

graylog2-server/src/main/java/org/graylog2/indexer/LegacyDeflectorRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Inject
+    public LegacyDeflectorRegistry(IndexSet indexSet) {
+        this.indexSet = requireNonNull(indexSet);
+        this.indexSetRegistry = Lists.newArrayList(indexSet);
+    }
+
+    @Override
+    public List<IndexSet> getAllIndexSets() {
+        return indexSetRegistry;
```
joschi: Consumers of the method result could modify the internal list.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param action the consumer function
+     */
+    void forEach(Consumer<IndexSet> action);
+
+    /**
+     * Returns a list of all Graylog managed indices.
+     *
+     * @return list of managed indices
+     */
+    String[] getAllGraylogIndexNames();
```
joschi: Out of curiosity, why use an array here (and in some other methods) but not in `getAllIndexSets()`?

PR #2895, opened 2016-09-27
Support all zlib compression levels for GELF messages

chengzhicn:
Support any zlib compression level

## Description
Currently, any message with zlib compression level less than 6 will be dropped, after this commit, any compression level should be fine.

## Motivation and Context
There is no document stating which zlib compression level is supported, so I suppose any compression level should be supported.

## How Has This Been Tested?
It's just a simple change.

## Types of changes

## Checklist:

Before this commit, only information we are checking is the highest bit of second byte, thus effectively only support zlib compression level higher than 5.
This commit use caller provided second byte instead of constraint ZLIB.second(), and I believe that's a typo.


joschi:
@chengzhicn Thanks for your contribution!

PR #2897, opened 2016-09-27
Add more properties and plugins to graylog-plugin-parent

bernd:
This adds some more properties and maven plugins to the graylog-plugin-parent which is used by multiple plugins.

### DIFF
--- a/graylog-plugin-parent/pom.xml
+++ b/graylog-plugin-parent/pom.xml
@@ -40,6 +40,18 @@
 
     <name>Graylog Plugin Parent POM</name>
 
+    <properties>
+        <graylog.plugin-dir>/usr/share/graylog-server/plugin</graylog.plugin-dir>
+
+        <!-- Dependencies -->
+        <auto-service.version>1.0-rc2</auto-service.version>
+
+        <!-- Plugin versions -->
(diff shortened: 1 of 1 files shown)


graylog-plugin-parent/pom.xml
```diff
@@ -40,6 +40,18 @@
...
+    <properties>
+        <graylog.plugin-dir>/usr/share/graylog-server/plugin</graylog.plugin-dir>
+
+        <!-- Dependencies -->
+        <auto-service.version>1.0-rc2</auto-service.version>
+
+        <!-- Plugin versions -->
+        <jdeb.version>1.5</jdeb.version>
+        <rpm-maven.version>2.1.5</rpm-maven.version>
+        <license-maven.version>2.11</license-maven.version>
```
joschi: Just for information: The latest version of com.mycila:license-maven-plugin is 3.0, but that would (slightly) change the used license template, so I'll vote for keeping version 2.11 for now.

* http://code.mycila.com/license-maven-plugin/
* https://github.com/mycila/license-maven-plugin/commit/06f9c8d2ffdbfaeffe5e40b7ae23b6135bb94bf8

PR #2902, opened 2016-09-28
Add graylog.version property to graylog-plugin-parent

bernd:
This is used in the graylog-plugin.properties file.

### DIFF
--- a/graylog-plugin-parent/pom.xml
+++ b/graylog-plugin-parent/pom.xml
@@ -42,6 +42,7 @@
 
     <properties>
         <graylog.plugin-dir>/usr/share/graylog-server/plugin</graylog.plugin-dir>
+        <graylog.version>${project.parent.version}</graylog.version>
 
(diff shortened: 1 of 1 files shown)


graylog-plugin-parent/pom.xml
```diff
@@ -42,6 +42,7 @@
 
     <properties>
         <graylog.plugin-dir>/usr/share/graylog-server/plugin</graylog.plugin-dir>
+        <graylog.version>${project.parent.version}</graylog.version>
```
joschi: This is a false friend as `${project.parent.version}` might not be the `graylog-plugin-parent` POM (e. g. in case a plugin itself is using submodules). It's resolved later in the Maven build process. 😞

bernd: Yeah, true. I will close it so we have to set it in the plugins itself.

bernd:
Closing this.

PR #2907, opened 2016-09-29
Update babel packages and config

dennisoelkers:
This is an extraction from the babel 6 update PR which got too big (#2847). It is reduced to update babel packages to their babel 6 counterparts and rewriting the config for them.

It contains:

 * Updating babel transpiler to 6.x
 * Converting babel config to new presets
 * Using `AppContainer` for `react-hot-loader` 3.x
 * Changing webpack config generation strategy to add plugins to entry points and contexts instead of merging complete configs


graylog2-web-interface/webpack.combined.config.js
```diff
@@ -11,9 +17,14 @@ const pluginConfigs = process.env.disable_plugins == 'true' ? [] : glob.sync(plu
...
-  .concat(pluginConfigs)
-  .map(function(config) { return require(path.resolve(__dirname, config)); })
-  .reduce(function(config1, config2) {
-    return merge.smart(config2, config1);
-  });
+const webpackConfig = require(path.resolve(__dirname, './webpack.config.js'));
+
+pluginConfigs.forEach(function(plugin) {
+  const pluginName = plugin.split('/')[2];
+  const pluginDir = path.resolve(__dirname, '../..', pluginName, 'src/web');
```
kroepke: This needs to take into account that some plugins can have submodules, like above for `pluginPattern2`.

kroepke:
Apart from the one comment about multi-module plugins, this looks good to me.
Build still works even if the plugins aren't updated, so 👍

graylog2-web-interface/webpack.combined.config.js
dennisoelkers: ✅ in 5815a057a9b0b31954ae377a453c19644e5a6de0

PR #2914, opened 2016-09-30
DTOs as AutoValue + in memory service

### LINKED ISSUE #2913: Refactor grok code and add in-memory service
Currently the grok resource code does not use auto values for its DTOs and let's the internal `ObjectId` reference, forcing implementations to pull in MongoDb code.

In order to implement an in-memory service for use in tests, benchmarks and ad-hoc, per-request storage of grok patterns (e.g. to simulate a new set of grok patterns without replacing the current cluster state), we should update the grok support code to use the current state of writing DTOs.

kroepke:
connected to #2913 

adapt all users of code to auto value
add in-memory service and test
create new bindings for default mongodb-based grok service

fixes #2913


garybot2:

graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
+
+public class InMemoryGrokPatternService implements GrokPatternService {
+    private static final Logger LOG = LoggerFactory.getLogger(InMemoryGrokPatternService.class);
+
+    // poor man's id generator
+    private AtomicLong idGen = new AtomicLong(0);
+
+    private final Map<String, GrokPattern> store = new MapMaker().makeMap();
```
joschi: Why not simply `new HashMap<>()`?

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
```diff
@@ -17,66 +17,78 @@
...
-    }
+    @JsonProperty
+    public abstract String pattern();
+
+    @JsonProperty
+    @Nullable
+    public abstract String contentPack();
 
     @Override
     public String toString() {
```
joschi: AutoValue already produces a similar (if not identical) `toString()` method, so this one could be removed.

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
```diff
@@ -17,66 +17,78 @@
...
+                .add("pattern", pattern())
+                .add("contentPack", contentPack())
                 .toString();
     }
 
+    @JsonCreator
+    public static GrokPattern create(@Id @ObjectId @JsonProperty("_id") @Nullable String id,
+                                     @JsonProperty("name") String name,
+                                     @JsonProperty("pattern") String pattern,
+                                     @JsonProperty("contentPack") @Nullable String contentPack) {
```
joschi: The default naming strategy we're using is [snake case](https://fasterxml.github.io/jackson-databind/javadoc/2.8/com/fasterxml/jackson/databind/PropertyNamingStrategy.html#SNAKE_CASE), so the JSON field will be named `content_pack` by default.

In other words, this breaks the Graylog REST API (since `GrokPatternSummary` has been replaced by this class).

graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
kroepke: Because this is a concurrent hashmap

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
kroepke: Yup, left over, good point.

graylog2-server/src/main/java/org/graylog2/grok/GrokPattern.java
kroepke: Will fix!

Issue #2869, opened 2016-09-21
Enhance Alert Data Model


dennisoelkers:
To enhance the alerting model, alerts are becoming timespans instead of discrete events. For this to work, the data model needs to be extended.
- [x] Add End Timestamp To Data Alert Model (#3072)
- [x] Add Alert Closing Logic To Alert Scanner (#3072)
- [x] Show Old Alarm Histories In Appropriate Way (#3103)
- [ ] Introduce Closing Event In Alarm Callbacks
- [ ] Investigate If Grace Period (Pausing Between Two Alerts) Is Still Needed
- [ ] Update documentation to reflect change

phlegx:
@dennisoelkers Will these changes also enable us to set a  time period and a number of errors/warnings happening during this period? 

Meaning to only send a notification if the number of errors/warnings within a certain time frame are over the threshold an user has set? Similar what ```newrelic``` does?

edmundoa:
Closed in favour of #3178 and #3179.

PR #2937, opened 2016-10-11
Refactor alert service

dennisoelkers:
## Description
This PR is doing two things:

 * refactoring the `Alert`/`AlertService` implementation classes to use AutoValue/MongoJack
 * extracting the alert scanning and callback triggering logic into the `AlertScanner` class to cleanup the `AlertScannerThread` and `AlertService` classes and consolidate and simplify associated logic

## Motivation and Context

This is a preparation of enhancing the data model (#2869).

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -17,122 +17,113 @@
...
+            DBQuery.and(
+                DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId),
+                DBQuery.greaterThanEquals(AlertImpl.FIELD_TRIGGERED_AT, since.toDate())
+            )
+        )
+            .limit(MAX_LIST_COUNT)
+            .sort(DBSort.desc(AlertImpl.FIELD_TRIGGERED_AT))
+            .toArray()
+            .stream()
+            .collect(Collectors.toList());
```
joschi: As much as I like the new Stream API in Java 8, it is unnecessary here.

```java
return Collections.unmodifiableList(this.coll.find(
            DBQuery.and(
                DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId),
                DBQuery.greaterThanEquals(AlertImpl.FIELD_TRIGGERED_AT, since.toDate())
            )
        )
            .limit(MAX_LIST_COUNT)
            .sort(DBSort.desc(AlertImpl.FIELD_TRIGGERED_AT))
            .toArray())
```

If `Alert` and `AlertImpl` were merged, the call to `Collections.unmodifiableList()` to "cast" the result would also be unnecessary.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -17,122 +17,113 @@
...
+            DBQuery.and(
+                DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId),
+                DBQuery.is(AlertImpl.FIELD_CONDITION_ID, conditionId)
+            )
+        )
+            .sort(DBSort.desc(AlertImpl.FIELD_TRIGGERED_AT))
+            .limit(1)
+            .toArray()
+            .stream()
+            .collect(Collectors.toList());
```
joschi: Basically see previous comment:
```java
final List<AlertImpl> mostRecentAlerts = this.coll.find(
            DBQuery.and(
                DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId),
                DBQuery.is(AlertImpl.FIELD_CONDITION_ID, conditionId)
            )
        )
            .sort(DBSort.desc(AlertImpl.FIELD_TRIGGERED_AT))
            .limit(1)
            .toArray();
```

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -196,78 +170,25 @@ public boolean inGracePeriod(AlertCondition alertCondition) {
...
-        }
-
-        return alerts;
+        return this.coll.find(DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId))
+            .sort(DBSort.desc(AlertImpl.FIELD_TRIGGERED_AT))
+            .skip(skip)
+            .limit(limit)
+            .toArray()
+            .stream()
+            .collect(Collectors.toList());
```
joschi: See previous comment about using the Java 8 Stream API in this case.

graylog2-server/src/main/java/org/graylog2/rest/resources/alerts/AlertResource.java
```diff
@@ -78,7 +78,7 @@ public AlertListSummary listAll(@ApiParam(name = "since", value = "Optional para
         List<AlertSummary> alerts = streamService.loadAll().stream()
                 .filter(stream -> isPermitted(RestPermissions.STREAMS_READ, stream.getId()))
                 .flatMap(stream -> alertService.loadRecentOfStream(stream.getId(), since).stream())
-                .limit(AlertImpl.MAX_LIST_COUNT)
+                .limit(AlertService.MAX_LIST_COUNT)
```
joschi: This should be configurable, e. g. use a query parameter `int limit` with [`@DefaultValue("300")`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/DefaultValue.html).

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/alerts/StreamAlertResource.java
```diff
@@ -159,7 +159,7 @@ public AlertListSummary listPaginated(@ApiParam(name = "streamId", value = "The
         checkPermission(RestPermissions.STREAMS_READ, streamId);
 
         if (limit == 0) {
```
joschi: This whole block could be gotten rid of by using [`@DefaultParam("300")`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/DefaultValue.html) for the method parameter `int limit`.

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
```diff
@@ -191,7 +191,7 @@ public DateTime getTriggeredAt() {
 
     public static class NegativeCheckResult extends CheckResult {
         public NegativeCheckResult(AlertCondition alertCondition) {
-            super(false, alertCondition, null, null, null);
+            super(false, null, null, null, null);
```
dennisoelkers: This was changed on purpose. If the CheckResult is negative, there is no triggered alert condition. Removing the parameter makes sense though.

graylog2-server/src/main/java/org/graylog2/alerts/Alert.java
```diff
@@ -16,18 +16,28 @@
...
 import org.joda.time.DateTime;
 
 import java.util.Map;
 
 /**
  * @author Dennis Oelkers <dennis@torch.sh>
  */
-public interface Alert extends Persisted {
+public interface Alert {
+    String getId();
```
dennisoelkers: I don't see a reason that it must be part of the builder interface. It was left out on purpose, because the only part who should be responsible for creating new (not updating) new alerts should be the `AlertScanner` class.

graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
```diff
@@ -16,89 +16,91 @@
...
-        map.put("id", fields.get(FIELD_ID).toString());
-        map.put(FIELD_CONDITION_ID, fields.get(FIELD_CONDITION_ID));
-        map.put(FIELD_STREAM_ID, fields.get(FIELD_STREAM_ID));
-        map.put(FIELD_DESCRIPTION, fields.get(FIELD_DESCRIPTION));
-        map.put(FIELD_CONDITION_PARAMETERS, fields.get(FIELD_CONDITION_PARAMETERS));
-        map.put(FIELD_TRIGGERED_AT, Tools.getISO8601String(triggeredAt));
-
-        return map;
-    }
+public abstract class AlertImpl implements Alert {
```
dennisoelkers: We do that with most value/dao classes to keep the implementation separate and program to interfaces, and I think it makes sense to keep it. Especially for this PR to be consistent with the rest.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
dennisoelkers: Is this just a matter of taste or are there good arguments to change it?

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/alerts/StreamAlertResource.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -17,122 +17,113 @@
...
-            throw new RuntimeException("Tried to create alert from not triggered alert condition result.");
-        }
-
-        Map<String, Object> fields = Maps.newHashMap();
-        fields.put("triggered_at", checkResult.getTriggeredAt());
-        fields.put("condition_id", checkResult.getTriggeredCondition().getId());
-        fields.put("stream_id", checkResult.getTriggeredCondition().getStream().getId());
-        fields.put("description", checkResult.getResultDescription());
-        fields.put("condition_parameters", checkResult.getTriggeredCondition().getParameters());
+    public Alert.Builder builder() {
```
dennisoelkers: This is the extension of factory methods which we have in most of our DAO interfaces, following our pattern that value classes are constructed by DAO classes only. If we want to change this (if there are good reasons to do that), then we should do it in a separate PR and consistently.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -17,122 +17,113 @@
...
+        return AlertImpl.builder();
+    }
 
-        return new AlertImpl(fields);
+    @Override
+    public Alert factory(AlertCondition.CheckResult checkResult) {
```
dennisoelkers: See above. The naming of the method is debatable though. `create` would be more consistent.

graylog2-server/src/main/java/org/graylog2/alerts/Alert.java
joschi: Unfortunately Auto Value will fail to generate the builder class if it's missing:

```
Error:(90, 12) java: Expected a method with this signature: org.graylog2.alerts.AlertImpl.Builder id(java.lang.String)
```

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
joschi: It avoids the unnecessary instantiation of objects (such as `Stream` and the list collector).

FWIW, my rule of thumb is: Use `Stream` if you're going to act upon its items (such as filtering or mapping), but not simply to "cast" a collection type.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
joschi: These methods in the "old" service classes are simply there because the inherent coupling with MongoDB via the `PersistedImpl` and `PersistedServiceImpl` classes.

I'd vote to move those methods to where they belong for a better coherence if we touch this class.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -17,122 +17,113 @@
...
-        }
-
-        return alerts;
+        return this.coll.find(
+            DBQuery.and(
+                DBQuery.is(AlertImpl.FIELD_STREAM_ID, streamId),
+                DBQuery.greaterThanEquals(AlertImpl.FIELD_TRIGGERED_AT, since.toDate())
+            )
+        )
+            .limit(MAX_LIST_COUNT)
```
joschi: ✅ Fair enough.

graylog2-server/src/main/java/org/graylog2/rest/resources/alerts/AlertResource.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
joschi: I would imagine something like this in `AlertImpl`:
```java
public static AlertImpl fromCheckResult(AlertCondition.CheckResult) {
    return builder()
            .streamId(checkResult.getTriggeredCondition().getStream().getId())
            .conditionId(checkResult.getTriggeredCondition().getId())
            .description(checkResult.getResultDescription())
            .conditionParameters(ImmutableMap.copyOf(checkResult.getTriggeredCondition().getParameters()))
            .triggeredAt(checkResult.getTriggeredAt())
            .build();
}
```

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
joschi: ✅ Postponed for now.

PR #2938, opened 2016-10-11
Using a kinda awkward way to build web assets in production mode.

dennisoelkers:
This also prevents react-hot-loader from complaining about missing module.
Also cleaning up the webpack/karma calls, relative paths are not necessary here.

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -10,10 +10,10 @@
   },
   "readme": "../README.md",
   "scripts": {
-    "start": "./node_modules/.bin/webpack --config webpack.vendor.js && ./node_modules/.bin/webpack-dev-server --config webpack.combined.config.js",
-    "start-nohmr": "./node_modules/.bin/webpack --config webpack.vendor.js && ./node_modules/.bin/webpack-dev-server --watch --config webpack.combined.config.js",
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/package.json
```diff
@@ -10,10 +10,10 @@
   },
   "readme": "../README.md",
   "scripts": {
-    "start": "./node_modules/.bin/webpack --config webpack.vendor.js && ./node_modules/.bin/webpack-dev-server --config webpack.combined.config.js",
-    "start-nohmr": "./node_modules/.bin/webpack --config webpack.vendor.js && ./node_modules/.bin/webpack-dev-server --watch --config webpack.combined.config.js",
-    "build": "./node_modules/.bin/webpack -p --config webpack.vendor.js && ./node_modules/.bin/webpack",
-    "test": "./node_modules/.bin/webpack --config webpack.vendor.js && ./node_modules/karma/bin/karma start karma.ci.conf.js",
```
bernd: How does that work if I don't have webpack installed globally?

dennisoelkers: It doesn't use the globally installed webpack, see [the npm docs](https://docs.npmjs.com/misc/scripts#path). The scripts of installed packages are prepended to the path of scripts run by npm. BTW: I don't have webpack installed globally either.

PR #2945, opened 2016-10-12
Add migration job for index sets

joschi:
## Description

Add a migration periodical which will create the default index set configuration based on the legacy settings in Graylog.

Refs #2880

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/IndexSetsMigrationPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/IndexSetsMigrationPeriodical.java
@@ -1,0 +1,165 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/periodical/IndexSetsMigrationPeriodical.java
```diff
@@ -0,0 +1,165 @@
...
+
+    private RetentionStrategyConfig getRetentionStrategyConfig(IndexManagementConfig indexManagementConfig) {
+        final String strategyName = indexManagementConfig.retentionStrategy();
+        final Provider<RetentionStrategy> provider = retentionStrategies.get(strategyName);
+        checkState(provider != null, "Couldn't retrieve retention strategy provider for <" + strategyName + ">");
+        final RetentionStrategy strategy = provider.get();
+        @SuppressWarnings("unchecked")
+        final Class<RetentionStrategyConfig> configClass = (Class<RetentionStrategyConfig>) strategy.configurationClass();
+        final RetentionStrategyConfig config = clusterConfigService.get(configClass);
+        return firstNonNull(config, strategy.defaultConfiguration());
```
bernd: This can break if it runs before the `ConfigurationManagementPerdiodical`. But this should only be a problem if you directly update from pre-2.0 to 2.2. Not sure how our official support for that kind of update is.

joschi: True.

I think we agreed at some point to actively support direct upgrades from the last minor version only (so this would be from Graylog 2.1.x in this case), but I can't find anything written on that decision. 😔

bernd: Okay.

Btw, that was the reason I started the `ConfigurationManagementPeriodical` to bundle all data migrations in one periodical so we can actually order them. :wink:

PR #2976, opened 2016-10-20
Route messages to different index sets

bernd:
To be able to write a message to different indices/index-sets we have to map streams to index sets and handle that when writing to Elasticsearch.
- Track matching index sets in the `Message` object to we don't have to iterate over all streams when doing the mapping. This assumes that usually there will be fewer index-sets than streams for a message.
- Add `Stream#getIndexSets()` and add the legacy default index-set to each stream. This needs to be changed to load the index-sets for a stream from the database.
- Adjust the `BlockingBatchedESOutput` to write one message into all index-sets for that message.
  Depending on the number of index-sets for a message, we are actually writing multiple messages for each message. This can slow down indexing because more work has to be done.
  The batching behaviour doesn't change because we are still flushing after the configured batch size.
- The `Messages#bulkIndex()` method now takes a small `IndexAndMessage` object to decide into which index a message should be indexed.

The reasoning behind putting the index-sets into the message instead of looking them up later is that we just have to iterate over index sets instead of iterating over streams and looking up index-sets from them.

Since we are rebuilding the stream router in a background thread when stream (and soon index-set) configuration changes, the expensive work is done outside of the hot code path.

There are probably several improvements that can be implemented to make this more efficient. This is a first naive version that makes it work with minimal changes.


bernd:
Since this is an idea proposal, it's lacking tests and other things like metrics and a consistent variable naming.

If we chose to go this route, we can review it for this kind of stuff.

kroepke:
I like it, because it is simple and handles the indexsets per message in a single place as opposed to multiple ones.
At the same time we can keep our promise of the flush batch size to elasticsearch (which is important until we support multiple connections).

graylog2-server/src/main/java/org/graylog2/indexer/messages/IndexAndMessage.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.graylog2.indexer.messages;
+
+import org.graylog2.plugin.Message;
+
+public class IndexAndMessage {
+    public String indexName;
+    public Message message;
+
+    public IndexAndMessage(String indexName, Message message) {
```
joschi: Implementation detail: Should we really create a new class for this? IMHO using [`Map.Entry`](https://docs.oracle.com/javase/8/docs/api/java/util/Map.Entry.html) via [`Maps#immutableEntry()`](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Maps.html#immutableEntry(K, V)) would be sufficient.

joschi:
👍  I like the approach.

graylog2-server/src/main/java/org/graylog2/indexer/messages/IndexAndMessage.java
bernd: I am fine with using an immutable map entry. Maybe we should use a `Map.Entry<IndexSet, Message>` instead of `Map.Entry<String, Message>`. Since we only have `#getKey()` and `#getValue()` it would be hard to see what is in the key. :wink:

bernd:
@joschi I think this can go into a first review round.

PR #2980, opened 2016-10-24
Use unique IDs in user list's delete buttons

edmundoa:
Each delete button in the user list is sharing the same element ID, which is wrong, and also make it harder to create automated test. This PR changes the ID appending the user name, making each ID unique, and also making easier to find the right delete button given a user name.


joschi:
@edmundoa Would it make sense to go through the code base and check where we have other non-unique `id` attributes?

edmundoa:
@joschi I think so, but it's tricky to spot, and we also need to make sure those IDs are not referenced in some CSS. I'll add a follow-up task for that 👍

PR #2986, opened 2016-10-25
Use response status 500 if search failed but wasn't syntax error

### LINKED ISSUE #2984: Exception during quick values request returns code 400
When a quick values search request fails in elasticsearch (for example with a ReduceSearchPhaseException), the server returns a 400 Bad Request, but it should return a proper error code (500 or 503)

joschi:
Closes #2984

## Types of changes

## Checklist:


joschi:
While this PR changes the HTTP response status code, it doesn't take care of the QuickValues widget handling this error situation and neither the universal search (which basically just hangs if an error occurs).

Issue #2843, opened 2016-09-17
Search result loading indicator hard to see


lennartkoopmann:
First of all: I LOVE the new way of reloading the search result page elements instead of a full browser reload. <3

One problem I noticed: When the search is returning the same search result very fast, it looks like no search ever happened and the user can be confused. I personally thought something was wrong when the page (seemingly) did not update.

A loading indicator would probably not solve this because the search returns so fast. Maybe some other highlighting that indicates that something happened? Happy to brainstorm here.

coffee-squirrel:
Maybe an extra line near the top of the "Search result" section.  `Retrieved 2016-09-19 14:51:12.345`

jhaar:
I agree, a quick search now shows zero evidence of having worked. Similarly, I am now finding myself routinely confused with long searches. I'll do a search, then re-run it and sit there wondering if I failed to click the button, because there's no evidence at all that anything is happening. If I wait long enough suddenly the results are there - but there's simply not enough visual feedback that anything is happening. The little "popup" at the bottom right is useful, but something more fundamental like a line stating when the last search was run would be a good trick - it would always change each time it was run. Another idea: how about an "hourglass" on the mouse, or some kind of real-time timer showing how long the search has been running?

lennartkoopmann:
@kroepke Can we prioritize this into the next patch release? It is clearly a very confusing bug. Thanks!

kroepke:
Yep, I agree it's not visible enough now.

lennartkoopmann:
Thank you!

Issue #2840, opened 2016-09-16
Add "Loading" indicator on Search page refresh


hc4:
After changing search criteria it is unclear if page already loaded or load is in progress.
It will be good to have some loading indicator after changing search criteria on Search page.

## Your Environment
* Graylog Version: 2.1.1

kroepke:
There is one in the lower right hand corner (I missed it as well, because it's hard to spot).

I'll leave this open for others to discuss whether to put it in a more prominent place.

hc4:
Hm... I see it now :)

hc4:
Maybe search button could be replaced with loading animation?
![image](https://cloud.githubusercontent.com/assets/17475270/18585305/cd85c5b4-7c1e-11e6-85d5-bad2ba2eda61.png)

edmundoa:
The problem I see is that putting it in a more prominent place would annoy a lot when enabling live-reloading.

edmundoa:
@hc4 That doesn't work, because when you scroll down you don't see the search bar.

hc4:
Ok, then here:
![image](https://cloud.githubusercontent.com/assets/17475270/18585339/ffd71888-7c1e-11e6-8f9b-ee9982a3def6.png)

pdepaepe:
I agree with previous comments, new search rendering is great, but indicator by far too much not visible.

Maybe something like bootstrap alert alert-info in LoadingIndicator.css:

```
:local(.loadingIndicator) {
    color: #31708f;
    background-color: #d9edf7;
    border-color: #bce8f1;
    padding: 15px;
    margin-bottom: 20px;
    margin-right:20px;
    border: 1px solid transparent;
    border-radius: 4px;
    position:fixed;
    bottom:0;
    right:0;
    z-index:2000;
}
```

PR #2992, opened 2016-10-26
Improved search indicators

edmundoa:
This PR modifies the search indications according to some of the feedback received:

- The search update indicator is now shown centred on the top part of the screen. The colour also changed to make it more visible in different types of background. This should help with #2840
- Add the time where the search was retrieved in the search sidebar, helping with #2843

### DIFF
--- a/graylog2-web-interface/src/components/common/LoadingIndicator.css
+++ b/graylog2-web-interface/src/components/common/LoadingIndicator.css
@@ -1,12 +1,12 @@
 :local(.loadingIndicator) {
-    background-color: #FFF;
-    box-shadow: 2px 2px 10px #aaa;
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/components/common/LoadingIndicator.css
```diff
@@ -1,12 +1,12 @@
...
+    left: 50%;
+    height: 32px;
+    width: 200px;
+    margin-left: -100px; /* half of the element width */
     padding: 5px 10px;
-    border-top: 1px solid #ccc;
-    border-left: 1px solid #ccc;
-    border-top-left-radius: 4px;
+    text-align: center;
+    z-index: 2000;
```
joschi: Over 9000!

PR #3072, opened 2016-11-08
Enhance alert data model

edmundoa:
This PR covers part of the changes required in #2869 and fixes some issues introduced in previous refactorings. In summary:

- Store `triggered_at` as a Date objects in MongoDB, avoiding sorting problems with old `Alert`s
- Add `resolved_at` and `is_interval` to `Alert` model
- Add support for resolving `Alert`s once the condition is not triggered
- Simplify `Alert` and `AlertService` interfaces
- Use a `Date` object to query `Alert` objects by the `triggered_at` field, avoiding problems with `MongoJack`
- Honour `limit` parameter in `AlertResource`
- Add some tests to ensure modified and new code works as expected™

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
@@ -55,11 +55,9 @@ public abstract class AlertImpl implements Alert {
     @Override
     public abstract String getConditionId();
 
-    @JsonProperty(FIELD_TRIGGERED_AT)
     @Override
     public abstract DateTime getTriggeredAt();
 
-    @JsonProperty(FIELD_RESOLVED_AT)
     @Override
     @Nullable
     public abstract DateTime getResolvedAt();
@@ -72,6 +70,17 @@ public abstract class AlertImpl implements Alert {
     @Override
     public abstract Map<String, Object> getConditionParameters();
 
+    @JsonProperty(FIELD_TRIGGERED_AT)
+    private Date getTriggeredAtDate() {
+        return getTriggeredAt().toDate();
+    }
+
+    @JsonProperty(FIELD_RESOLVED_AT)
+    @Nullable
+    private Date getResolvedAtDate() {
+        return getResolvedAt() == null ? null : getResolvedAt().toDate();
+    }
+
     static Builder builder() {
         return new AutoValue_AlertImpl.Builder();
     }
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -189,4 +202,28 @@ public String save(Alert alert) throws ValidationException {
...
         return this.coll.save((AlertImpl)alert).getSavedId();
     }
+
+    @Override
+    public Alert resolveAlert(Alert alert) {
+        if (alert == null || isResolved(alert)) {
+            return alert;
+        }
+
+        final AlertImpl updatedAlert = AlertImpl.create(
```
bernd: How about adding a `public abstract Builder toBuilder();` to `AlertImpl.java` so you can replace the create call with the following.

```java
final Alert updatedAlert = alert.toBuilder().resolvedAt(Tools.nowUTC()).build();
```

graylog2-server/src/main/java/org/graylog2/alerts/AlertScanner.java
```diff
@@ -82,22 +83,36 @@ protected Alert handleTriggeredCheckResult(AlertCondition.CheckResult result, St
         return alert;
     }
 
+    private void handleResolveAlert(Alert alert) {
+        alertService.resolveAlert(alert);
+        // TODO: Send resolve notifications
```
edmundoa: Yes, it's the fourth point in #2869

graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
```diff
@@ -73,34 +96,75 @@ static Builder builder() {
...
                                    @JsonProperty(FIELD_CONDITION_ID) String conditionId,
-                                   @JsonProperty(FIELD_TRIGGERED_AT) Date triggeredAt,
+                                   @JsonProperty(FIELD_TRIGGERED_AT) Date triggeredAtDate,
+                                   @JsonProperty(FIELD_RESOLVED_AT) Date resolvedAtDate,
                                    @JsonProperty(FIELD_DESCRIPTION) String description,
-                                   @JsonProperty(FIELD_CONDITION_PARAMETERS) Map<String, Object> conditionParameters) {
+                                   @JsonProperty(FIELD_CONDITION_PARAMETERS) Map<String, Object> conditionParameters,
+                                   @JsonProperty(FIELD_IS_INTERVAL) boolean isInterval) {
+        final DateTime triggeredAt = new DateTime(triggeredAtDate);
+        final DateTime resolvedAt = resolvedAtDate == null ? null : new DateTime(resolvedAtDate);
```
edmundoa: I think it can be, when deserialising the object from MongoDB. I'll double check it anyway, maybe we are missing a `@Nullable` annotation in the parameter.

graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
edmundoa: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertScanner.java
edmundoa: I added the comment to "justify" the existence of that one-line method, which would not make much sense otherwise.

graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
edmundoa: The argument will be `null` when deserialising old alerts from MongoDB. Avoiding the condition, sets `resolvedAt` to `now` to all alerts that haven't got the field. I'll add the `@Nullable` annotation to the parameter and a comment to make it clear.

PR #3098, opened 2016-11-17
Assign messages to the default stream

kroepke:
When a message enters the processing stage, we need to route it to the default stream early, instead of relying on the stream router to do so, because otherwise it would not be picked up be the pipeline processor, if it runs before the stream router.

To protect the default stream from modifications all mutation is disallowed by the REST API (it returns Bad Request responses) and the corresponding UI elements are not rendered for it.


graylog2-server/src/main/java/org/graylog2/periodical/DefaultStreamMigrationPeriodical.java
```diff
@@ -113,6 +100,12 @@ public boolean masterOnly() {
 
     @Override
     public boolean startOnThisNode() {
+        try {
+            streamService.load(Stream.DEFAULT_STREAM_ID);
```
joschi: Wouldn't this make the `DefaultStreamCreated` marker obsolete?

PR #3102, opened 2016-11-18
Allow to remove message from default stream

joschi:
Refs #2880

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/streams/requests/UpdateStreamRequest.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/streams/requests/UpdateStreamRequest.java
@@ -42,15 +42,16 @@ public abstract class UpdateStreamRequest {
     @Nullable
     public abstract String matchingType();
 
+    @JsonProperty("remove_from_all_messages")
+    @Nullable
+    public abstract Boolean removeFromAllMessages();
+
     @JsonCreator
     public static UpdateStreamRequest create(@JsonProperty("title") @Nullable String title,
                                              @JsonProperty("description") @Nullable String description,
                                              @JsonProperty("matching_type") @Nullable String matchingType,
-                                             @JsonProperty("rules") @Nullable List rules) {
-        return new AutoValue_UpdateStreamRequest(title, description, matchingType);
-    }
-
-    public static UpdateStreamRequest updateMatchingType(@Nonnull String matchingType) {
-        return create(null, null, matchingType, null);
+                                             @JsonProperty("rules") @Nullable List rules,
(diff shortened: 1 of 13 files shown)


graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
```diff
@@ -203,6 +207,13 @@ public String getFingerprint() {
 
         for (Stream stream : result) {
             streamMetrics.markIncomingMeter(stream.getId());
+            if(stream.getRemoveFromAllMessages()) {
+                if(message.removeStream(defaultStream)) {
+                    LOG.trace("Successfully removed default stream <{}> from message <{}>", defaultStream.getId(), message.getId());
```
bernd: I would wrap this in `isTraceEnabled` to avoid the `getId()` calls for the stream and message objects for every message.

graylog2-server/src/main/java/org/graylog2/rest/models/streams/requests/UpdateStreamRequest.java
```diff
@@ -42,15 +42,16 @@
     @Nullable
     public abstract String matchingType();
 
+    @JsonProperty("remove_from_all_messages")
```
bernd: I personally find `remove_from_all_messages` pretty confusing. How about using `remove_matches_from_default_stream`? To me this makes more sense.

In the UI I would use "remove matches from all messages stream" or something like that, though.

graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
```diff
@@ -70,13 +72,15 @@
     public StreamRouterEngine(@Assisted List<Stream> streams,
                               @Assisted ExecutorService executorService,
                               StreamFaultManager streamFaultManager,
-                              StreamMetrics streamMetrics) {
+                              StreamMetrics streamMetrics,
+                              @DefaultStream Stream defaultStream) {
```
bernd: This needs to be a `Provider<Stream>` to avoid the problems when starting the server.

graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
joschi: ✅

graylog2-server/src/main/java/org/graylog2/rest/models/streams/requests/UpdateStreamRequest.java
joschi: ✅

bernd:
LGTM :+1:

PR #3104, opened 2016-11-18
Merge migration periodicals for default stream and index sets

bernd:
This makes it easier to execute the steps in order.


bernd:
Replaced by #3108

Issue #2946, opened 2016-10-12
Extractor Not able to load message


luckyb56:
When I try to load a message to add extractors for an input, it doesn't load any message. In chrome's inspect console I see this error:

```
Unhandled rejection Invariant Violation: findComponentRoot(..., .0.2.0.0.0.1.0.0.3.0.1.0.0.0.0.0.0.0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ".
    at n (http://1.1.1.1/vendor.js:405:9453)
    at Object.findComponentRoot (http://1.1.1.1/vendor.js:12:12296)
    at Object.findReactNodeByID (http://1.1.1.1/vendor.js:12:11592)
    at Object.u [as getNode] (http://1.1.1.1/vendor.js:12:4463)
    at y._updateDOMProperties (http://1.1.1.1/vendor.js:403:26765)
    at y.updateComponent (http://1.1.1.1/vendor.js:403:25514)
    at y.receiveComponent (http://1.1.1.1/vendor.js:403:24812)
    at Object.receiveComponent (http://1.1.1.1/vendor.js:13:24)
    at p._updateRenderedComponent (http://1.1.1.1/vendor.js:403:12139)
    at p._performComponentUpdate (http://1.1.1.1/vendor.js:403:11900)
    at p.updateComponent (http://1.1.1.1/vendor.js:403:11178)
    at p.i [as updateComponent] (http://1.1.1.1/vendor.js:12:18757)
    at p.receiveComponent (http://1.1.1.1/vendor.js:403:10216)
    at Object.receiveComponent (http://1.1.1.1/vendor.js:13:24)
    at Object.updateChildren (http://1.1.1.1/vendor.js:403:3283)
    at y._reconcilerUpdateChildren (http://1.1.1.1/vendor.js:404:14379)
    at y._updateChildren (http://1.1.1.1/vendor.js:404:15304)
    at y.updateChildren (http://1.1.1.1/vendor.js:404:15179)
    at y._updateDOMChildren (http://1.1.1.1/vendor.js:403:27437)
    at y.updateComponent (http://1.1.1.1/vendor.js:403:25552)
    at y.receiveComponent (http://1.1.1.1/vendor.js:403:24812)
[...]

PR #3106, opened 2016-11-21
Fix problems rendering object field values

edmundoa:
Some messages may contain objects as field values and we need to properly convert them into strings that can be rendered by react.

This commit changes the way we convert values into strings, relying on `JSON.stringify()` to do it.

Refs Graylog2/graylog-plugin-beats#3. It may also be related to #2946.

**Update**: As @joschi pointed out, this only fixes part of the problem in Graylog2/graylog-plugin-beats#3: being able to display messages with a hierarchical structure. The change does not affect how Graylog internally handles those messages.

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageShow.jsx
+++ b/graylog2-web-interface/src/components/search/MessageShow.jsx
@@ -28,7 +28,7 @@ const MessageShow = React.createClass({
 
   possiblyHighlight(fieldName) {
     // No highlighting for the message details view.
-    return String(this.props.message.fields[fieldName]);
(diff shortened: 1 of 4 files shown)


joschi:
@edmundoa While this fixes displaying messages with a hierarchical structure, it doesn't solve the inherent problem of Graylog2/graylog-plugin-beats#3: That Graylog can only work with flat message structures right now.

If we do indeed fix the issue with hierarchical messages, we can remove flattening from a lot of different places.

edmundoa:
You are right, this only fixes the display of those fields in the web interface, will change the text to make that more clear 👍

PR #3108, opened 2016-11-22
Create migration interface and classes to replace migration perdiodicals

bernd:
This creates a minimal `Migration` interface that can be used to create migration code.

The idea is to create a new class per migration and that they will be executed in order, oldest first. This avoids ordering problems we currently have with implementing migrations on top of periodicals.

This is a minimal first step, if this turns out to be usable we can think about improving this by using injection in the migration classes and having a migration runner that isn't a periodical.

The problem this solves is that elasticsearch, default stream and index sets migrations need to run in a specific order to work correctly. With this the order will be hardcoded in the `ConfigurationManagementPeriodical` (might be renamed).

## Implementation Details

- Migration class/file names are prefixed by the date they have been created at. This is a bit ugly, but makes it really easy to see the order of the migrations in the file system hierarchy.
- The migration interface contains a `createdAt()` method which returns a date. This is currently unused but can be used later to determine the execution order of the migrations.
- Migrations are still run by a periodical, but at least they are run in order. Previously each migration has been a separate periodical which could run in random order.


bernd:
Autodiscover options look good. :+1: Feel free to merge.

PR #3111, opened 2016-11-23
Do not log an error with stack trace when content pack dir is missing

bernd:
A warning should be enough.


kroepke:
Finally I don't have to think something's broken <3

PR #3118, opened 2016-11-24
filter indices to search in according to stream filter

### LINKED ISSUE #3117: Search in streams that use different index sets searches in all indices
Since the addition of IndexSets a stream search touches all matching index ranges (and thus even indices that aren't its index set).

The search config needs to filter out the non-matching indices, to properly limit the search request to only matching indices.
This avoids having to perform unnecessary work on the elasticsearch level.

kroepke:
Fixes #3117
Connects #3117

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.graylog2.indexer.ranges.IndexRange;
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/IndexHelper.java
```diff
@@ -83,9 +86,14 @@ public static QueryBuilder getTimestampRangeFilter(TimeRange range) throws Inval
     }
 
     public static Set<IndexRange> determineAffectedIndicesWithRanges(IndexRangeService indexRangeService,
-                                                                     TimeRange range) {
+                                                                     TimeRange range,
+                                                                     String indexPrefix) {
         final ImmutableSortedSet.Builder<IndexRange> indices = ImmutableSortedSet.orderedBy(IndexRange.COMPARATOR);
         for (IndexRange indexRange : indexRangeService.find(range.getFrom(), range.getTo())) {
+            // if we only consider a certain index set, filter the index ranges by it.
+            if (indexPrefix != null && !indexRange.indexName().startsWith(indexPrefix + "_")) {
```
bernd: We should use `IndexSet#isManagedIndex()` here because that is more strict and will be even more strict in the future.

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -646,4 +684,29 @@ public FieldTypeException(Throwable e) {
             super(e);
         }
     }
+
+
+    /**
+     * Extracts the last stream id from the filter string passed as part of the elasticsearch query. This is used later
+     * to pass to possibly existing message decorators for stream-specific configurations.
```
bernd: This extracts only the last stream id from the filter? That breaks when you do `stream_id:foo OR stream_id:bar`, no? It will then only search in the index set of the last stream?

bernd: Okay, I wasn't aware of that... :smile:

PR #3120, opened 2016-11-24
Migrate alert receivers

edmundoa:
This PR moves the alert receivers configuration from the stream model to the email alarm callback, which is the object using the receivers.

To do so, I am introducing here a list configuration field (`ListField`), and make use of them to hold the list of emails and users that should get an email notification.

The PR also cleans up parts of the code using the old alert receivers, and adapts `StreamService` to operate on the `EmailAlarmCallback` instead of the `Stream`, keeping backwards compatibility in the old add and remove receiver methods of the `StreamResource`.

**Update**: The PR contains now a migration job to move old alert receivers from streams into each `EmailAlarmCallback` that is associated to it.

Refs #2871


graylog2-server/src/main/java/org/graylog2/migrations/V20161125142400_EmailAlarmCallbackMigration.java
```diff
@@ -153,7 +127,21 @@ public void doRun() {
...
         return configurationRequest.getFields().entrySet()
-            .stream()
-            .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getDefaultValue()));
+                .stream()
+                .collect(Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().getDefaultValue()));
+    }
+
+    @AutoValue
+    @JsonAutoDetect
+    public static abstract class MigrationCompleted {
```
bernd: You moved the `EmailAlarmCallbackMigrationPeriodical` over to a migration and changed the "migration done" marker class from `EmailAlarmCallbackMigrated` to `MigrationCompleted`.

That means this migration will run again even though the `EmailAlarmCallbackMigrationPeriodical` has been run already.

Just want to make sure this is intentional and that this doesn't cause any problems. :smiley:

graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java
```diff
@@ -62,6 +65,13 @@
                 case "boolean":
                     value = "true".equalsIgnoreCase(String.valueOf(entry.getValue()));
                     break;
+                case "list":
+                    final List<?> valueList = entry.getValue() == null ? Collections.emptyList() : (List<?>) entry.getValue();
```
bernd: I guess this can be `final List<String> valueList` since you are converting everything to a string.

graylog2-server/src/main/java/org/graylog2/migrations/V20161125161400_AlertReceiversMigration.java
```diff
@@ -0,0 +1,172 @@
...
+    @Override
+    public void upgrade() {
+        // Do not run again if the migration marker can be found in the database.
+        if (clusterConfigService.get(MigrationCompleted.class) != null) {
+            return;
+        }
+
+        final Map<String, Optional<String>> streamMigrations = this.streamService.loadAll()
+                .stream()
+                .filter(stream -> !stream.getAlertReceivers().isEmpty()
```
bernd: This is wrong. It only checks if there are no keys in the map but not if the values for the keys are empty.

graylog2-server/src/main/java/org/graylog2/utilities/ConfigurationMapConverter.java
edmundoa: No, at that point the type of the elements inside the list is still unknown. The elements are converted into strings in the stream in the next line, and stored in the `value` variable, that will be then stored in the effective configuration map.

I think this is sensible at this point, as this method is used to convert data provided by the user, and it's always better being a bit more cautious about it to avoid surprises.

PR #3121, opened 2016-11-24
Adjust several helper and REST resources to index sets

bernd:
- Add `get()` and `getForIndexName()` methods to IndexSetRegistry
- Adjust all methods in Indices to index sets
- Add `indexAnalyzer()`, `indexTemplateName()`,
  `indexOptimizationMaxNumSegments()` and `indexOptimizationDisabled()`
  methods to IndexSetConfig (only static for now, needs migration)
- Adjust DeflectorResource to index sets
- Adjust IndexerOverviewResource to index sets
- Adjust IndicesResource to index sets


bernd:
I also added 2927c28 to add missing rotation/retention strategy class names to IndexSetConfig because the other changes require this.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
```diff
@@ -58,33 +53,32 @@
...
     @RequiresPermissions(RestPermissions.DEFLECTOR_READ)
     @Produces(MediaType.APPLICATION_JSON)
-    public DeflectorSummary deflector() throws TooManyAliasesException {
-        // TODO 2.2: Resource needs to be adjusted to support multiple write targets
-        final IndexSet indexSet = indexSetRegistry.getAllIndexSets().stream()
-                .findFirst()
-                .orElseThrow(() -> new NotFoundException("Unable to find current deflector alias"));
+    @Deprecated
+    public DeflectorSummary deprecatedDeflector() throws TooManyAliasesException {
+        throw new UnsupportedOperationException("Unsupported since Graylog 2.2 - use /system/deflector/{indexSetId}");
```
kroepke: Should this maybe return the default indexset information instead?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
```diff
@@ -96,12 +90,25 @@ public DeflectorSummary deflector() throws TooManyAliasesException {
     @Path("/cycle")
     @RestrictToMaster
     @AuditEvent(type = AuditEventTypes.ES_WRITE_INDEX_UPDATE_JOB_START)
-    public void cycle() {
-        final String msg = "Cycling deflector. Reason: REST request.";
+    @Deprecated
+    public void deprecatedCycle() {
+        throw new UnsupportedOperationException("Unsupported since Graylog 2.2 - use /system/deflector/{indexSetId}/cycle");
```
kroepke: same as above

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexerOverviewResource.java
```diff
@@ -80,32 +83,43 @@ public IndexerOverviewResource(DeflectorResource deflectorResource,
     @Timed
     @ApiOperation(value = "Get overview of current indexing state, including deflector config, cluster state, index ranges & message counts.")
     @Produces(MediaType.APPLICATION_JSON)
+    @Deprecated
     public IndexerOverview index() throws TooManyAliasesException {
+        throw new UnsupportedOperationException("Unsupported since Graylog 2.2 - use /system/indexer/overview/{indexSetId}");
```
kroepke: default index set: same same

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -111,17 +112,17 @@
     private static final Logger LOG = LoggerFactory.getLogger(Indices.class);
     private static final String REOPENED_INDEX_SETTING = "graylog2_reopened";
 
+    public static final String ALL_INDEX_PREFIX = "*";
```
kroepke: This, together with https://github.com/Graylog2/graylog2-server/pull/3121/files#diff-1e045693d5a6863d0e42f1f49bc4dc90R326 leads us to read the information for all indices that contain an `_` character, which might be problematic.

Can we collect all the index prefixes instead and restrict the query to the indices that we should care about instead?

graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
```diff
@@ -54,6 +55,24 @@ public MongoIndexSetRegistry(IndexSetService indexSetService,
     }
 
     @Override
+    public Optional<IndexSet> get(final String indexSetId) {
+        return findAllMongoIndexSets().stream()
```
bernd: :+1: Will fix that.

graylog2-server/src/main/java/org/graylog2/indexer/SetIndexReadOnlyJob.java
```diff
@@ -70,9 +77,9 @@ public void execute() {
 
         activityWriter.write(new Activity("Flushed and set <" + index + "> to read-only.", SetIndexReadOnlyJob.class));
 
-        if (!disableIndexOptimization) {
+        if (!indexSet.get().getConfig().indexOptimizationDisabled()) {
             try {
-                systemJobManager.submit(optimizeIndexJobFactory.create(index));
+                systemJobManager.submit(optimizeIndexJobFactory.create(index, indexSet.get().getConfig().indexOptimizationMaxNumSegments()));
```
bernd: I don't get that warning in my IDE and I am checking that at the top of the method.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
bernd: I introduced a default marker for index sets and changed the resources to use that. :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
bernd: See above

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexerOverviewResource.java
bernd: See above

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
bernd: Good catch, changed that so the `getAllMessageFields` method takes an array of index prefix wildcards.

graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
```diff
@@ -54,6 +55,24 @@ public MongoIndexSetRegistry(IndexSetService indexSetService,
...
+                .filter(mongoIndexSet -> mongoIndexSet.getConfig().id().equals(indexSetId))
+                .map(mongoIndexSet -> (IndexSet) mongoIndexSet)
+                .findFirst();
+    }
+
+    @Override
+    public Optional<IndexSet> getForIndexName(String indexName) {
+        for (MongoIndexSet indexSet : findAllMongoIndexSets()) {
```
kroepke: Can't we do a prefix comparison on an index here? Anyway, we can address this later as it is only a performance improvement, and I'm not even sure how often we do this.

graylog2-server/src/main/java/org/graylog2/indexer/SetIndexReadOnlyJob.java
kroepke: Mostly a nitpick anyway, I've set it to aggressively warn about this (and it most likely doesn't resolve the negations properly).

PR #3122, opened 2016-11-24
Fix AuditActor test

bernd:
Fixes the test for me locally.


bernd:
Seems to be a local problem...

edmundoa:
I had the same yesterday a couple of times. It fixed after cleaning and rebuilding, but not sure if there's anything else going on in there.

bernd:
@edmundoa Thanks for the info, I will test that!

bernd:
@edmundoa Yep, after clean and rebuild it works. Strange... :confused:

Issue #3128, opened 2016-11-28
Extend index ranges to include stream ids


kroepke:
When we added index sets as write target abstraction the problem arose that changing the index set of a stream causes searches to no longer find messages in previously assigned index sets.

This is because the engine filters indices to search in by the currently configured index set per stream, and does not take its history into account.
Rather than storing the entire history and using that to reconstruct the mappings over time, we will extend the index ranges to contain a set of stream ids their index contains data for.
This has several advantages:

* The existing lookup mechanics work without change.
* Restored archives (which live in index sets) can contain multiple streams, but don't necessarily contain all of them.
* When recalculating index sets we can lazily add the stream ids, because we already have to execute expensive min/max timestamp aggregations and can afford the histogram on `stream_ids`.
* The default stream also needs this handling, because otherwise it suffers from the same problem that data before the introduction of the default stream would not be found by its stream search.

The tasks therefore:
 - [x] Extend `IndexRange` to contain a set of streams (which may be missing or empty)
 - [x] Adapt `IndexHelper#determineAffectedIndicesWithRanges` to make use of the set of stream ids during lookup
 - [x] Add migration of the default stream to update index ranges
 - [x] Modify index range recalculation to set stream ids contained in the index.

bernd:
This is done.

PR #3129, opened 2016-11-28
Store streams with index ranges

kroepke:
Connect #3128

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -619,14 +619,16 @@ public class Indices {
         final DateTime min = new DateTime((long) minAgg.getValue(), DateTimeZone.UTC);
         final Max maxAgg = f.getAggregations().get("ts_max");
         final DateTime max = new DateTime((long) maxAgg.getValue(), DateTimeZone.UTC);
-        List<String> streamIds = null;
+        // make sure we return an empty list, so we can differentiate between old indices that don't have this information
+        // and newer ones that simply have no streams.
+        ImmutableList.Builder<String> streamIds = ImmutableList.builder();
         final Terms streams = f.getAggregations().get("streams");
         if (!streams.getBuckets().isEmpty()) {
-            streamIds = ImmutableList.copyOf(streams.getBuckets().stream()
+            streamIds.addAll(streams.getBuckets().stream()
                     .map(Terms.Bucket::getKeyAsString)
                     .collect(toSet()));
         }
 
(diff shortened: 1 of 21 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -589,19 +603,30 @@ public TimestampStats timestampStatsOfIndex(String index) {
...
         final Min minAgg = f.getAggregations().get("ts_min");
         final DateTime min = new DateTime((long) minAgg.getValue(), DateTimeZone.UTC);
         final Max maxAgg = f.getAggregations().get("ts_max");
         final DateTime max = new DateTime((long) maxAgg.getValue(), DateTimeZone.UTC);
+        List<String> streamIds = null;
+        final Terms streams = f.getAggregations().get("streams");
+        if (!streams.getBuckets().isEmpty()) {
+            streamIds = ImmutableList.copyOf(streams.getBuckets().stream()
+                    .map(Terms.Bucket::getKeyAsString)
+                    .collect(toSet()));
```
bernd: You use `toSet()` but assign it to a `List<String>`. How about switching `IndexRangeStats` over to `Set<String>` for the steams field?

kroepke: Mongojack didn't work with Set, so I copy it into a list for storing.
It's only important that the list does not contain duplicates, which is achieved by the intermediate set.

garybot2:

graylog2-server/src/main/java/org/graylog2/migrations/V20161130141500_DefaultStreamRecalcIndexRanges.java
```diff
@@ -0,0 +1,116 @@
...
+        final IndexSet defaultIndexSet = optDefaultIndexSet.get();
+
+        if (!cluster.isConnected()) {
+            LOG.info("Cluster not connected yet, delaying migration until it is reachable.");
+            while (true) {
+                try {
+                    cluster.waitForConnectedAndDeflectorHealthy();
+                    break;
+                } catch (InterruptedException | TimeoutException e) {
+                    e.printStackTrace();
```
bernd: Please log a proper error here. And do we need to do anything when that happens or are we just continue to the next loop iteration?

kroepke: this is left over from automatic "try-catch adding" and should be a warning, I think. timeouts and spurious interrupts can happen. I'll fix.

PR #3133, opened 2016-11-29
Use real index set ID to fix integration tests

bernd:
This should fix the integration tests.


bernd:
Yay, it finally works!

PR #3145, opened 2016-12-01
Move alert conditions and notifications to alerts page

edmundoa:
These changes move the the management of alert conditions and notifications to the new alerts page.

Most of the functionality moved is still behaving as before, for instance: assigning alarm callbacks (aka notifications) is still applied to the whole stream instead of alert conditions; and testing a notification also tests other notifications for that stream. That will be changed in follow up PRs.


graylog2-server/src/main/java/org/graylog2/alerts/types/FieldContentValueAlertCondition.java
```diff
@@ -88,10 +88,9 @@ public ConfigurationRequest getRequestedConfiguration() {
     public static class Descriptor extends AlertCondition.Descriptor {
         public Descriptor() {
             super(
-                "Field Content Value Alert Condition",
+                "Field Content Alert Condition",
                 "https://www.graylog.org/",
-                "This condition is triggered when the aggregated value of a field is higher/lower than a defined "
-                + "threshold for a given time range."
+                "This condition is triggered when the content of messages is equal to a defined value."
```
edmundoa: Yes, you are right, conditions were not pluggable before 2.2, and I also don't think we have any condition plugins around to test.

The `human_name` is only used in here as far as I know: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/components/alertconditions/AlertConditionForm.jsx#L77-L77
Condition plugins may provide some custom UI for the create/edit form, but if they don't we use `human_name` to give some context inside the generic configuration form. I don't really like how the forms look at the moment, so we can probably display the `human_name` somewhere else too.

kroepke: Ok, then let's leave them in and figure out what to do with it at a later stage.

graylog2-web-interface/src/stores/streams/StreamsStore.ts
```diff
@@ -137,6 +137,10 @@ class StreamsStore {
   sendDummyAlert(streamId: string) {
     const url = URLUtils.qualifyUrl(ApiRoutes.StreamAlertsApiController.sendDummyAlert(streamId).url);
     const promise = fetch('POST', url);
+    promise.then(
+      () => UserNotification.success('Test notification was sent successfully'),
+      (error) => UserNotification.error('Could not send test notification')
```
edmundoa: That's a good point. I thought that would raise an exception when sending the test email, but I didn't test. I want to change the test to work on a single notification instead of on all notifications of a stream, and can look into that then.

PR #3147, opened 2016-12-01
Add alarm callback titles

### LINKED ISSUE #3146: Add title to alarm callbacks
Since 2.1 it is possible to add titles to alert conditions, and I think it would be quite useful doing the same on alarm callbacks. Specially since they are becoming more visible in 2.2, and may even be shared between alert conditions in the future.

edmundoa:
This adds support for titles in alarm callbacks and allows setting them from the UI.

Fixes #3146

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/V20161125142400_EmailAlarmCallbackMigration.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/V20161125142400_EmailAlarmCallbackMigration.java
@@ -121,6 +121,7 @@ public class V20161125142400_EmailAlarmCallbackMigration extends Migration {
         final AlarmCallbackConfiguration alarmCallbackConfiguration = alarmCallbackService.create(stream.getId(),
                 CreateAlarmCallbackRequest.create(
                         EmailAlarmCallback.class.getCanonicalName(),
+                        "Email Alert Callback",
                         defaultConfig
                 ),
                 localAdminUser.getId()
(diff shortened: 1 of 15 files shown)


graylog2-server/src/main/java/org/graylog2/migrations/V20161125142400_EmailAlarmCallbackMigration.java
```diff
@@ -121,6 +121,7 @@ public void upgrade() {
         final AlarmCallbackConfiguration alarmCallbackConfiguration = alarmCallbackService.create(stream.getId(),
                 CreateAlarmCallbackRequest.create(
                         EmailAlarmCallback.class.getCanonicalName(),
+                        "Email Alert Callback",
```
kroepke: Shouldn't this be "Email Alert Notification"?

graylog2-server/src/test/java/org/graylog2/migrations/V20161125161400_AlertReceiversMigrationTest.java
```diff
@@ -147,6 +147,7 @@ public void doMigrateSingleQualifyingStream() throws Exception {
                 alarmCallbackId,
                 matchingStreamId,
                 EmailAlarmCallback.class.getCanonicalName(),
+                "Email Alert Callback",
```
kroepke: Also "Notification" ?

kroepke:
lgtm!

PR #3156, opened 2016-12-06
Dependency updates

joschi:
Just some regular dependency updates without code changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,7 +126,7 @@
         <mongojack.version>2.6.1</mongojack.version>
         <natty.version>0.12</natty.version>
         <netty.version>3.10.6.Final</netty.version>
-        <okhttp.version>3.4.2</okhttp.version>
+        <okhttp.version>3.5.0</okhttp.version>
         <opencsv.version>2.3</opencsv.version>
         <os-platform-finder.version>1.2.1</os-platform-finder.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -88,23 +88,23 @@
 
         <!-- Dependencies -->
         <airline.version>0.7</airline.version>
-        <amqp-client.version>3.6.5</amqp-client.version>
+        <amqp-client.version>4.0.0</amqp-client.version>
```
bernd: Has this been tested with a RabbitMQ server? I am hesitating a bit to upgrade to a new major version during beta.

joschi: I haven't tested it, but according to https://github.com/rabbitmq/rabbitmq-java-client#4x-series, it should be compatible:

> Starting with 4.0, this client releases are independent from RabbitMQ server releases. These versions can still be used with RabbitMQ server 3.x.

joschi: Also https://github.com/rabbitmq/rabbitmq-java-client/releases/tag/v4.0.0 says:
> RabbitMQ Java Client 4.0.0 doesn't bring any breaking changes to application APIs (e.g. Consumer), the major version upgrade rather marks the decoupling of the Java Client project from the RabbitMQ broker, at least for release cycle and versioning.

PR #3157, opened 2016-12-06
Remove quotation on string fields

### LINKED ISSUE #3155: Wrongly quoted fields in search page
As part of the work in #3106, string fields in the search page are quoted.

## Expected Behavior
Fields in search page are not quoted.

## Current Behavior
Fields in search page are quoted.

## Your Environment

* Graylog Version: 2.2.0-beta.1

edmundoa:
Only use `JSON.stringify()` to convert objects into strings, relying on `String()` for doing all other conversions. In this way we can properly display objects and arrays and not add any quotation on string fields.

Refs #3106, fixes #3155

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageShow.jsx
+++ b/graylog2-web-interface/src/components/search/MessageShow.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Col, Row } from 'react-bootstrap';
 import Immutable from 'immutable';
 import MessageDetail from './MessageDetail';
+import StringUtils from 'util/StringUtils';
 
 const MessageShow = React.createClass({
   propTypes: {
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/components/search/MessageShow.jsx
```diff
@@ -28,7 +28,8 @@ const MessageShow = React.createClass({
 
   possiblyHighlight(fieldName) {
     // No highlighting for the message details view.
-    return JSON.stringify(this.props.message.fields[fieldName]) || '';
+    const value = this.props.message.fields[fieldName];
+    return (typeof value === 'object' ? JSON.stringify(value) : String(value));
```
joschi: Would it make sense to extract this pattern into a common utility method?

edmundoa: Good point, will do that 👍

PR #3188, opened 2016-12-12
Add synthetic JavaBean getters to AutoValue classes

### LINKED ISSUE #3127: Hibernate Validator does not work with non-bean accessors (auto-value classes)
Many (all?) of our auto value classes don't use JavaBean conventions for their methods, which breaks Hibernate Validator.

As a result `@Valid` annotations in REST resources don't actually validate any of the properties of those classes, which potentially leads us to accept invalid data.

There are three possible workarounds:
1. Migrate all auto values to using JavaBean style methods
2. Write custom validator classes for each auto value.
3. Write a auto-value extension that generates proper bean-style accessors.

The third option seems the most viable for us right now and has already been prototyped.

joschi:
## Description

This PR adds an AutoValue extension based on https://github.com/Graylog2/graylog2-server/issues/3127#issuecomment-265772246 which generates synthetic getters following the JavaBean conventions.

It also adds the new `@WithBeanGetter` annotation to the relevant AutoValue classes.

Fixes #3127

## Types of changes

## Checklist:


auto-value-javabean/src/main/java/org/graylog/autovalue/WithBeanGetterExtension.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public String generateClass(Context context, String className, String classToExtend, boolean isFinal) {
+        final Generator generator = new Generator(context, className, classToExtend, isFinal);
+        final TypeSpec.Builder typeSpecBuilder = generator.classBuilder();
+        typeSpecBuilder.addMethod(generator.superConstructor());
+
+        final Map<String, ExecutableElement> properties = context.properties();
+        for (Map.Entry<String, ExecutableElement> entry : properties.entrySet()) {
+            typeSpecBuilder.addMethod(generateGetterMethod(entry.getKey(), entry.getValue()));
```
bernd: We have to check if there are `get*`/`is*` methods on the class already so we don't override them.

auto-value-javabean/src/main/java/org/graylog/autovalue/WithBeanGetterExtension.java
```diff
@@ -0,0 +1,90 @@
...
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import java.util.Map;
+
+/**
+ * Creates Java Bean getter methods for each property.
+ */
+public class WithBeanGetterExtension extends AutoValueExtension {
+    private static final ClassName ANNOTATION_JSON_PROPERTY = ClassName.get(JsonProperty.class);
```
bernd: I have been using the `ClassName.get("com.fasterxml.jackson.annotation", "JsonIgnore");` version in the original code so we don't have to depend on jackson. If we want to extract this extension into a separate project later - which I think is a good idea because this might be useful to others as well - we should probably use that code again.

graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackHistoryImpl.java
```diff
@@ -31,6 +32,7 @@
 import org.mongojack.ObjectId;
 
 @AutoValue
+@WithBeanGetter
```
bernd: Why do we want to add the annotation to every AutoValue class we have? This only makes sense for those that have validation annotations, no?

graylog2-server/src/test/java/org/graylog2/autovalue/WithBeanGetterTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class WithBeanGetterTest {
+    @Test
+    public void testBeanHasJavaBeanGetters() {
+        final TestBean bean = TestBean.create("Test", true, false);
+
+        final List<String> methodNames = Arrays.stream(bean.getClass().getMethods())
+                .map(Method::getName)
+                .collect(Collectors.toList());
+        assertThat(methodNames)
```
bernd: Please add a test that checks if the annotations are pulled over to the new method. Thanks!

graylog2-server/src/main/java/org/graylog2/alarmcallbacks/AlarmCallbackHistoryImpl.java
joschi: Yes, but I didn't want to check each class individually and used some `sed` magic to replace `@AutoValue` with `@AutoValue\n@WithBeanGetters` (minus those classes which already contained getters). 😉

graylog2-server/src/test/java/org/graylog2/autovalue/WithBeanGetterTest.java
joschi: ✅

auto-value-javabean/src/main/java/org/graylog/autovalue/WithBeanGetterExtension.java
joschi: I thought about that but decided against the check because we explicitly add the `@WithBeanGetter` annotation to classes. If these classes already contain getters, we don't add the annotation.

bernd:
Looks good to me so far. Let me know once it's ready for a final review.

graylog-project-parent/pom.xml
```diff
@@ -416,6 +417,17 @@
                 <scope>provided</scope>
             </dependency>
             <dependency>
+                <groupId>com.gabrielittner.auto.value</groupId>
+                <artifactId>auto-value-extension-util</artifactId>
+                <version>${auto-value-extension-util.version}</version>
```
bernd: Shouldn't this be marked as `provided` to avoid pulling it into the regular dependencies of the server? Or is this implicitly provided because the auto-value-javabean itself is in scope `provided`?

graylog2-server/src/main/java/org/graylog2/migrations/V20161122174500_AssignIndexSetsToStreamsMigration.java
```diff
@@ -115,6 +116,7 @@ private IndexSetConfig findDefaultIndexSet() {
 
     @JsonAutoDetect
     @AutoValue
+@WithBeanGetter
```
bernd: Nitpick: indentation is broken (for several classes)

graylog2-server/src/test/java/org/graylog2/autovalue/WithBeanGetterTest.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.autovalue;
```
bernd: I guess this file should be in the auto-value-javabean maven module instead of the the graylog2-server module?

graylog2-server/src/main/java/org/graylog2/plugin/indexer/searches/timeranges/AbsoluteRange.java
```diff
@@ -29,6 +30,7 @@
 import java.util.Map;
 
 @AutoValue
+@WithBeanGetter
```
bernd: This class, `KeywordRange` and `RelativeRange` all implement their own `#getFrom()` and `#getTo()` due to the abstract methods in `TimeRange`.

Maybe we should remove the annotation from those to avoid confusion in the future if any of those hand written implementations should change.

graylog-project-parent/pom.xml
joschi: The latter, it's a `compile` dependency for `auto-value-javabean` but will not be pulled in transitively in graylog2-server, because `auto-value-javabean` is pulled in as `provided` there.

graylog2-server/src/main/java/org/graylog2/migrations/V20161122174500_AssignIndexSetsToStreamsMigration.java
joschi: 👍

graylog2-server/src/test/java/org/graylog2/autovalue/WithBeanGetterTest.java
bernd: Bummer! Thanks! :white_check_mark:

PR #3193, opened 2016-12-13
Index set fixes

bernd:
- Create an unique index on the `index_prefix` field in MongoDB
- Handle `DuplicateKeyException` in the index sets resource
- Add a warning to the UPGRADING file about rotation/retention plugins
- Implement some checks to avoid the creation of index sets with conflicting index prefixes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
@@ -30,11 +30,11 @@ public class IndexSetValidator {
         this.indexSetRegistry = indexSetRegistry;
     }
 
-    public Optional<Error> validate(IndexSetConfig newConfig) {
+    public Optional<Violation> validate(IndexSetConfig newConfig) {
         // Build an example index name with the new prefix and check if this would be managed by an existing index set
         final String indexName = newConfig.indexPrefix() + MongoIndexSet.SEPARATOR + "0";
         if (indexSetRegistry.isManagedIndex(indexName)) {
-            return Optional.of(Error.create("Index prefix \"" + newConfig.indexPrefix() + "\" would conflict with an existing index set!"));
+            return Optional.of(Violation.create("Index prefix \"" + newConfig.indexPrefix() + "\" would conflict with an existing index set!"));
         }
 
         // Check if an existing index set has a more generic index prefix.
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
```diff
@@ -0,0 +1,60 @@
...
+            if (newConfig.indexPrefix().startsWith(indexSet.getIndexPrefix())) {
+                return Optional.of(Error.create("Index prefix \"" + newConfig.indexPrefix() + "\" would conflict with existing index set prefix \"" + indexSet.getIndexPrefix() + "\""));
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    @AutoValue
+    public static abstract class Error {
```
joschi: Maybe call it `Violation` (similar to [`ConstraintViolation`](https://docs.oracle.com/javaee/7/api/javax/validation/ConstraintViolation.html) in `javax.validation`).

I'm not totally comfortable with the naming of the class because of [`Error`](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html).

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
```diff
@@ -0,0 +1,128 @@
...
+    @JsonProperty("title")
+    @NotBlank
+    public abstract String title();
+
+    @JsonProperty("description")
+    @Nullable
+    public abstract String description();
+
+    @JsonProperty("default")
+    public abstract boolean isDefault();
```
joschi: Should it be possible to change that flag on an existing index set?

What happens if there is more than one index set with `default == true`?

joschi: Same applies for creating an index set, of course.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.graylog2.indexer;
+
+import com.google.auto.value.AutoValue;
+import org.graylog2.indexer.indexset.IndexSetConfig;
+
+import javax.inject.Inject;
+import java.util.Optional;
+
+public class IndexSetValidator {
```
joschi: 👍  for extracting this logic into its own class.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
bernd: :white_check_mark:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
bernd: Yes, it should be possible to change that. This is still an [open TODO](https://github.com/Graylog2/graylog2-server/issues/2880) and needs to be implemented. We have to ensure that only one index set can be the default one.

bernd:
Renamed the Error class to Violation.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
joschi: ✅

Issue #3178, opened 2016-12-12
Alerting open tasks


edmundoa:
These are some open tasks for alerting in 2.2:

- [x] Check permissions to manage conditions and notifications
- [x] UI refinements
    - [x] Improve conditions forms
    - [x] Ensure everything looks fine when there are no streams
    - [X] Clean up alarm callbacks code (#3205)
- [X] Allow to test a single notification (#3196)
- [X] Ensure the test action fails when email configuration is not set (#3196)
- [x] Investigate if grace period (pausing between two alerts) is still needed (#3346)

PR #3196, opened 2016-12-13
Improved alarm callback testing

edmundoa:
This PR fixes two of the open tasks for alerting #3178:
- Ensures that testing an `EmailAlarmCallback` is throwing exceptions when there are errors, that can be displayed to the user when testing a callback
- Adds a new REST resource, allowing users to test single alarm callbacks

### DIFF
--- a/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
+++ b/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
@@ -17,8 +17,16 @@ const AlertNotification = React.createClass({
   },
   mixins: [Reflux.connect(AlertNotificationsStore)],
 
+  getInitialState() {
+    return {
+      isTestingAlert: false,
+    };
+  },
+
   _onTestNotification() {
-    AlertNotificationsActions.testAlert(this.props.alertNotification.id);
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
```diff
@@ -18,8 +17,16 @@ const AlertNotification = React.createClass({
...
   mixins: [Reflux.connect(AlertNotificationsStore)],
 
+  getInitialState() {
+    return {
+      isTestingAlert: false,
+    };
+  },
+
   _onTestNotification() {
-    StreamsStore.sendDummyAlert(this.props.alertNotification.stream_id);
```
edmundoa: Yes, it can, at least we don't use it. Will take care of it 👍

edmundoa: I was wrong, we use it in some other old components that I didn't clean up yet. As we have another task in #3178 to clean up the code, I would postpone the removal until then.

bernd: :white_check_mark:

PR #3198, opened 2016-12-13
Add custom Jackson (de-) serializer for ZonedDateTime

joschi:
## Description

This PR adds a custom serializer and deserializer for `ZonedDateTime` to the `ObjectMapper` used by MongoJack, so that properties with type `ZonedDateTime` are properly serialized as a MongoDB date field.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoJackObjectMapperProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoJackObjectMapperProvider.java
@@ -22,16 +22,18 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.deser.DeserializationProblemHandler;
 import com.fasterxml.jackson.databind.jsontype.TypeIdResolver;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.type.SimpleType;
-
 import org.graylog2.indexer.retention.strategies.UnknownRetentionStrategyConfig;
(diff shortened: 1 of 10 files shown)


bernd:
Nice! We need the same for Joda DateTime because that is also wrongly serialized as a String into MongoDB.

joschi:
@bernd I wouldn't use JodaTime (`DateTime`) in new classes and migration of existing entries is (unfortunately) not that easy.

But we can look into this in another PR (possibly post-2.2.0).

bernd:
@joschi I think we have to discuss the usage Joda Time vs Java Core Time at some point. Right now we are using both (also for new code) which is confusing. But that's unrelated to this PR.

What is the problem with migration? We already have the problem that we need to migrate data after this. We are using it for the index set config which is now in a public beta release and we are also using this in Graylog Enterprise.

joschi:
> What is the problem with migration? We already have the problem that we need to migrate data after this.

I'd argue that given this is still a beta, we don't need to care for a migration process for the `index_sets` collection. It's not worth the effort.

bernd:
> I'd argue that given this is still a beta, we don't need to care for a migration process for the index_sets collection. It's not worth the effort.

But what **is** the effort? :smiley:  What needs to be done to migrate this?

joschi:
@bernd Read all entries, check the type of the `creation_date` field, update the document in case it was a `string`.

Sounds not like much but it's still code only required for people who ran Graylog 2.2.0-beta.1 or 2.2.0-beta.2 and we'll have to maintain it. I'd rather concentrate on other things. 😉

bernd:
So it turned out that only one collection in Graylog Enterprise needs a migration and others will be fixed on write and/or are not sorted/queried by date anywhere.

PR #3203, opened 2016-12-14
Index set ui fixes

bernd:
- Sort index sets by title, not by date
- Do not show the rotation/retention form for non-writable index sets


graylog2-server/src/main/java/org/graylog2/indexer/indexset/MongoIndexSetService.java
```diff
@@ -120,7 +120,7 @@ protected MongoIndexSetService(JacksonDBCollection<IndexSetConfig, ObjectId> col
         final DBQuery.Query query = DBQuery.or(idQuery.toArray(new DBQuery.Query[0]));
 
         return ImmutableList.copyOf(collection.find(query)
-                .sort(DBSort.desc("creation_date"))
+                .sort(DBSort.asc("title"))
```
bernd: Let's just leave it. Maybe we need it in the future... :smile:

PR #3207, opened 2016-12-14
Allow fetching streams by index set ID

joschi:
This PR adds checks to `IndexSetsResource` to prevent the deletion of actively used (i. e. assigned to a stream) index sets.


bernd:
@joschi There is code that does basically the same here: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/indexset/MongoIndexSetService.java#L172-L183

:smile:

joschi:
@bernd D'oh!

This being said, I think at least the method in `StreamService` from this PR should be used to address the `TODO` in the code snippet you pointed to. 😉

bernd:
> This being said, I think at least the method in StreamService from this PR should be used to address the TODO in the code snippet you pointed to. :wink:

Yes, please! :smile:

PR #3209, opened 2016-12-15
IndexSet default setting

bernd:
An index set can be marked as `default` in `IndexSetConfig`. This cannot be changed by the user at the moment, but should be. We have to ensure that only one index set is the default!

Since we are using MongoDB, there are no cross-document atomic operations or transactions so we have to take care of that ourselves.

## Idea 1: Use a timestamp

We are currently using a boolean `default` field in the `index_sets` database collection to mark the default index set. Because of the lack of transactions this is not really safe to update for several documents at once. (set `true` on the new default, set `false` on all others)

As an alternative to the boolean flag we could use a timestamp field like `default_since` or similar that will be set to the current time when a "set default index set" request comes in. To determine the default we have to load the index set with the newest timestamp. By using a unique index on the timestamp field, we would avoid duplicate timestamps.

## Idea 2: Coordinate on master

Another idea was to publish an event on the cluster event bus to set the default index set. All nodes subscribe to the event but only the master node will actually handle it. Since only one node **should** be master and the events are not processed in parallel, this should avoid interleaving writes to the index sets collection. It doesn't guarantee any order though.

This is more complicated than the timestamp idea beacuse it involves the cluster event bus and needs an event handler. It also makes the request to set the default index set async so we have to deal with that in the UI. It also fails when there is more than one active master node.

## Other Ideas?

Please let me know if you have different, better ideas. Let's discuss!


kroepke:
Can we simply use a `ClusterConfig` (or similar) setting that contains the id (or name or both) of the current default? That way there's always exactly 0 or 1 default index sets.

joschi:
@bernd Idea 3:

Store the default index set ID in cluster config. This would make queries for that more expensive, but in how many places do we really have to check if an index set is the default index set?

bernd:
@joschi @kroepke Good point! That seems to be the simplest idea, I guess. Thank you!

@joschi Right now we only need it for display purposes in the UI and when importing content packs.

bernd:
This is ready for review now.

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetService.java
```diff
@@ -38,6 +38,13 @@
     Optional<IndexSetConfig> get(String id);
 
     /**
+     * Retrieve the default index set.
+     *
+     * @return A filled {@link Optional} with the default index set, an empty {@link Optional} if there is no default.
```
joschi: Is there a situation where there is *no* default index set, which is not a serious bug?

bernd: The default index set should be there. If it's not then

- Content pack imports that contain streams will fail
- The old indices REST API endpoints without an index set parameter will fail
- The `V20161116172200_CreateDefaultStreamMigration` will fail

If we change that to not returning an optional and throwing an error when the default index set is absent, we also need a migration to migrate the default flag from the `index_sets` collection to the cluster config object.

bernd: :white_check_mark:

bernd:
I added a migration to create a `DefaultIndexSetConfig` cluster config object and delete the `default` field from the index set documents.

bernd:
Rebased against master.

graylog-project-parent/pom.xml
```diff
@@ -529,6 +529,12 @@
                 <scope>test</scope>
             </dependency>
             <dependency>
+                <groupId>com.github.fakemongo</groupId>
```
joschi: This is a duplicate: https://github.com/Graylog2/graylog2-server/blob/b9fc16676b6100df5e0c1724f9e61d8831f0dd79/graylog-project-parent/pom.xml#L531-L548

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetService.java
bernd: Ah, damn. Forgot to change the return type of `IndexSetService#getDefault()`.

bernd: Adjusted the interfaces.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
```diff
@@ -97,16 +99,19 @@ public DeflectorSummary deflector(@ApiParam(name = "indexSetId") @PathParam("ind
...
-        LOG.info(msg);
-        activityWriter.write(new Activity(msg, DeflectorResource.class));
+            final String msg = "Cycling deflector for default index set <" + indexSet.getConfig().id() + ">. Reason: REST request.";
+            LOG.info(msg);
+            activityWriter.write(new Activity(msg, DeflectorResource.class));
 
-        indexSet.cycle();
+            indexSet.cycle();
+        } catch (IllegalStateException e) {
+            throw new NotFoundException("Default index set not found");
```
joschi: I would throw an `InternalServerErrorException` here (or simply let the `IllegalStateException` bubble up which results in the same) because it's actually an error in the Graylog system, not a wrong request by the client.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
```diff
@@ -70,10 +70,12 @@ public DeflectorResource(IndexSetRegistry indexSetRegistry,
...
     public DeflectorSummary deprecatedDeflector() throws TooManyAliasesException {
-        final IndexSet indexSet = indexSetRegistry.getDefault()
-                .orElseThrow(() -> new NotFoundException("Default index set not found"));
-
-        return DeflectorSummary.create(indexSet.isUp(), indexSet.getCurrentActualTargetIndex());
+        try {
+            final IndexSet indexSet = indexSetRegistry.getDefault();
+            return DeflectorSummary.create(indexSet.isUp(), indexSet.getCurrentActualTargetIndex());
+        } catch (IllegalStateException e) {
+            throw new NotFoundException("Default index set not found");
```
joschi: I would throw an `InternalServerErrorException` here (or simply let the `IllegalStateException` bubble up which results in the same) because it's actually an error in the Graylog system, not a wrong request by the client.

bernd: :white_check_mark:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/DeflectorResource.java
bernd: :white_check_mark:

PR #3211, opened 2016-12-15
Address open TODOs for Graylog 2.2.0

joschi:
This PR addresses some open TODOs:

* Remove TODO about `ZonedDateTime` in `IndexSetConfig`
* Migrate remaining index settings from config file to `IndexSetConfig`
* Allow overriding analyzer in `MessageResource#analyze()`

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
@@ -21,8 +21,8 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
-import org.graylog.autovalue.WithBeanGetter;
 import com.google.common.collect.ComparisonChain;
+import org.graylog.autovalue.WithBeanGetter;
 import org.graylog2.plugin.indexer.retention.RetentionStrategyConfig;
 import org.graylog2.plugin.indexer.rotation.RotationStrategyConfig;
 import org.hibernate.validator.constraints.NotBlank;
@@ -128,7 +128,6 @@ public abstract class IndexSetConfig implements Comparable<IndexSetConfig> {
         return false;
     }
 
-    // TODO 2.2: creation_date is a string but needs to be a date - look at AlertImpl for example
     @JsonCreator
     public static IndexSetConfig create(@Id @ObjectId @JsonProperty("_id") @Nullable String id,
                                         @JsonProperty("title") @NotBlank String title,
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
@@ -18,7 +18,6 @@ package org.graylog2.indexer.indexset;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ComparisonChain;
@@ -33,8 +32,6 @@ import javax.annotation.Nullable;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 import java.time.ZonedDateTime;
-
-import static java.util.Objects.requireNonNull;
 
 @AutoValue
 @WithBeanGetter
@@ -100,40 +97,25 @@ public abstract class IndexSetConfig implements Comparable<IndexSetConfig> {
     @NotNull
     public abstract ZonedDateTime creationDate();
 
-    // TODO 2.2: Migrate config setting to index set!
     @JsonProperty("index_analyzer")
-    @JsonIgnore
-    public String indexAnalyzer() {
-        return "standard";
-    }
+    @NotBlank
+    public abstract String indexAnalyzer();
 
-    // TODO 2.2: Migrate config setting to index set!
     @JsonProperty("index_template_name")
-    @JsonIgnore
-    public String indexTemplateName() {
-        return "graylog-internal-" + requireNonNull(id());
-    }
+    @NotBlank
+    public abstract String indexTemplateName();
 
-    // TODO 2.2: Migrate config setting to index set!
     @JsonProperty("index_optimization_max_num_segments")
-    @JsonIgnore
-    public int indexOptimizationMaxNumSegments() {
-        return 1;
-    }
+    @Min(1L)
+    public abstract int indexOptimizationMaxNumSegments();
 
-    // TODO 2.2: Migrate config setting to index set!
     @JsonProperty("index_optimization_disabled")
-    @JsonIgnore
-    public boolean indexOptimizationDisabled() {
-        return false;
-    }
+    public abstract boolean indexOptimizationDisabled();
 
     @JsonCreator
     public static IndexSetConfig create(@Id @ObjectId @JsonProperty("_id") @Nullable String id,
                                         @JsonProperty("title") @NotBlank String title,
                                         @JsonProperty("description") @Nullable String description,
-                                        // TODO 3.0: Remove "default" field here. A migration in 2.2 removed it.
-                                        @JsonProperty("default") @Nullable Boolean isDefault, // Ignored, older objects might still have it
                                         @JsonProperty("writable") @Nullable Boolean isWritable,
                                         @JsonProperty(FIELD_INDEX_PREFIX) @NotBlank String indexPrefix,
                                         @JsonProperty("index_match_pattern") @Nullable String indexMatchPattern,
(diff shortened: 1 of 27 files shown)


graylog2-server/src/main/java/org/graylog2/migrations/V20161116172100_DefaultIndexSetMigration.java
```diff
@@ -94,6 +94,10 @@ public void upgrade() {
                 .rotationStrategy(getRotationStrategyConfig(indexManagementConfig))
                 .retentionStrategy(getRetentionStrategyConfig(indexManagementConfig))
                 .creationDate(ZonedDateTime.now(ZoneOffset.UTC))
+                .indexAnalyzer(elasticsearchConfiguration.getAnalyzer())
+                .indexTemplateName(elasticsearchConfiguration.getTemplateName())
+                .indexOptimizationMaxNumSegments(elasticsearchConfiguration.getIndexOptimizationMaxNumSegments())
+                .indexOptimizationDisabled(elasticsearchConfiguration.isDisableIndexOptimization())
```
bernd: Modifying the existing migration will break existing beta.2 setups because the default index set already exists.

Since we need a new migration anyway (see above), we should create a new one.

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
```diff
@@ -103,35 +100,21 @@
...
-        return "standard";
-    }
+    @NotBlank
+    public abstract String indexAnalyzer();
 
-    // TODO 2.2: Migrate config setting to index set!
     @JsonProperty("index_template_name")
-    @JsonIgnore
-    public String indexTemplateName() {
-        return "graylog-internal-" + requireNonNull(id());
```
bernd: This and all other settings need a migration to not break existing index sets in a 2.2 setup.

We also need UI components and/or good defaults, otherwise the creation of new index sets will not work.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
```diff
@@ -75,6 +75,21 @@
     @NotNull
     public abstract RetentionStrategyConfig retentionStrategy();
 
+    @JsonProperty("index_analyzer")
+    @NotBlank
+    public abstract String indexAnalyzer();
+
+    @JsonProperty("index_template_name")
```
bernd: I think we shouldn't allow the update of the index template name for an index set. This makes it harder to cleanup the template when an index set gets deleted.

joschi: Should this be settable at all then?

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
bernd: I think we should handle all defaults in the `@JsonCreator` method in this class because that's the place where the fields can be `null` when coming out of the database. We should remove the `@Nullable` annotation from the new field methods because they shouldn't ever be null. Otherwise we have to adjust all consumers of the methods to handle the `null` case. (they don't do that now)

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
```diff
@@ -34,11 +34,10 @@
 import javax.validation.constraints.NotNull;
 import java.time.ZonedDateTime;
 
-import static java.util.Objects.requireNonNull;
-
 @AutoValue
 @WithBeanGetter
 @JsonAutoDetect
+@JsonIgnoreProperties(ignoreUnknown = true)
```
bernd: Why do we need this?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ConfigurationResource.java
```diff
@@ -37,20 +36,17 @@
...
         this.configuration = configuration;
-        this.esConfiguration = esConfiguration;
     }
 
     @GET
     @Timed
     @ApiOperation(value = "Get relevant configuration settings and their values")
     public ExposedConfiguration getRelevant() {
-        return ExposedConfiguration.create(configuration, esConfiguration);
+        return ExposedConfiguration.create(configuration);
```
bernd: This needs a note in the UPGRADING document since we are removing some data from the API endpoint.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
```diff
@@ -98,26 +109,31 @@ public static IndexSetUpdateRequest fromIndexSetConfig(IndexSetConfig indexSet)
...
-                .indexWildcard(oldConfig.indexWildcard())
-                .shards(shards())
-                .replicas(replicas())
-                .rotationStrategyClass(rotationStrategyClass())
-                .rotationStrategy(rotationStrategy())
-                .retentionStrategyClass(retentionStrategyClass())
-                .retentionStrategy(retentionStrategy())
-                .creationDate(oldConfig.creationDate())
-                .build();
+        return IndexSetConfig.create(
```
bernd: At one point we should agree on how we do auto value object creation so we don't change it in every PR. :wink:

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
joschi: Otherwise loading the `IndexSetConfig` from the database failed due to the unknown `default` field.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/requests/IndexSetUpdateRequest.java
joschi: ✅ Reverted to builder style.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ConfigurationResource.java
joschi: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20161216123500_DefaultIndexSetMigration.java
```diff
@@ -0,0 +1,90 @@
...
+    public void upgrade() {
+        if (clusterConfigService.get(V20161216123500_Succeeded.class) != null) {
+            return;
+        }
+
+        // The default index set must have been created first.
+        checkState(clusterConfigService.get(DefaultIndexSetCreated.class) != null, "The default index set hasn't been created yet. This is a bug!");
+
+        // TODO: Migrate to new method once it's been merged
+        final IndexSetConfig indexSetConfig = indexSetService.getDefault();
```
joschi: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20161216123500_DefaultIndexSetMigration.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        // The default index set must have been created first.
+        checkState(clusterConfigService.get(DefaultIndexSetCreated.class) != null, "The default index set hasn't been created yet. This is a bug!");
+
+        for (IndexSetConfig indexSetConfig : indexSetService.findAll()) {
+            final String analyzer = elasticsearchConfiguration.getAnalyzer();
+            final IndexSetConfig updatedConfig = indexSetConfig.toBuilder()
+                    .indexAnalyzer(analyzer)
+                    .indexTemplateName(elasticsearchConfiguration.getTemplateName())
```
bernd: Because otherwise we will overwrite the same index template for every index set. Every index set has a different wildcard `template` field.

Example:

```
http :9200/_template | jq '.[].template'
"emails_*"
"yolo_*"
"restored-archive-*"
"graylog_*"
```

joschi: Ah, okay. So this is a restriction of our index set implementation. We should think about allowing to use the same index template for multiple index sets (multiple wildcards are possible).

graylog2-server/src/main/java/org/graylog2/migrations/V20161216123500_DefaultIndexSetMigration.java
```diff
@@ -73,7 +73,7 @@ public void upgrade() {
             final String analyzer = elasticsearchConfiguration.getAnalyzer();
             final IndexSetConfig updatedConfig = indexSetConfig.toBuilder()
                     .indexAnalyzer(analyzer)
-                    .indexTemplateName(elasticsearchConfiguration.getTemplateName())
+                    .indexTemplateName(indexSetConfig.indexPrefix() + "-template")
```
bernd: As written earlier, the default index set needs the template name from the old ES configuration options. It might have been modified by the user and our default template for new indices would break that.

bernd: Hmm, true. So the config option was basically a no-op?

joschi: No, it simply let users decide how to name the Graylog index template. If they needed modifications of the mapping, they could override them with a custom index template: http://docs.graylog.org/en/2.1/pages/configuration/elasticsearch.html#custom-index-mappings

PR #3215, opened 2016-12-16
Create DefaultIndexSetConfig when creating the default index set

bernd:
Fixes problems with other migrations.


bernd:
> Should the creation of DefaultIndexSetConfig be removed from `V20161215163900_MoveIndexSetDefaultConfig` to avoid duplicate creation and prevent possible conflicts?

I will change the migration to check if there is a `DefaultIndexSetConfig` already. That should avoid the possible problem you mentioned but still allow a working upgrade from an earlier beta.

PR #3224, opened 2016-12-21
Make "sender" optional in EmailAlarmCallback

### LINKED ISSUE #1512: Conflict between server.conf SMTP settings, and Alerts settings, 'test alert' message false success message
We have Graylog 1.2.1 installed on Ubuntu 14.04.  The only way our ms exchange admin will allow this linux box to send SMTP email is by jumping through all of  his anti-linux security hoops.  I have to configure graylog server thusly:

transport_email_enabled = true
transport_email_hostname = xxxx.xxxxxxxx.com
transport_email_port = 25
transport_email_use_auth = true
transport_email_use_tls = true
transport_email_use_ssl = false
transport_email_auth_username = greylog (sic)
transport_email_auth_password = xxxxxxxxx
transport_email_subject_prefix = [graylog2]
transport_email_from_email = greylog@xxxxxxxx.com (sic)

The problem is that when configuring an alert email callback, the From: header is user configurable, and does NOT default to the transport_email_from_email configured in server.conf.  In fact, if the user selects ANYTHING except that exact value from transport_email_from_email, the message will be rejected by our Exchange server.  The users configuring these email callbacks have no way to know what the correct value of this one field must be, nor do they receive any feed back about the alert not working; they simply do not receive their alerts. 

The 'send test alert' button fails to note the authentication failures and the smtp message rejection from the exchange server and claims the test was successful, even though it actually failed.  I don't know what that test is watching, because it's not certainly watching the conversation with the SMTP server.

The from: header in email callbacks (and other places where Graylog might attempt to use SMTP messaging) must honor the transport_email_from_email setting in server.conf if it is provided, otherwise SMTP transmission simply breaks with no user feedback why.

joschi:
The `sender` field in an email alert callback configuration should be optional. Graylog is using the email address configured in `` as fallback, is the `sender` field wasn't provided.

Fixes #1512

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
+++ b/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
@@ -49,6 +49,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
+
 public class EmailAlarmCallback implements AlarmCallback {
     private static final Logger LOG = LoggerFactory.getLogger(EmailAlarmCallback.class);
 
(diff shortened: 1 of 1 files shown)


bernd:
I created an issue for the missing `#checkConfiguration()` invocations: https://github.com/Graylog2/graylog2-server/issues/3236

PR #3228, opened 2016-12-21
Support syslog messages with ISO-8601 timestamps

### LINKED ISSUE #2954: Problems parsing RFC3164 Syslog Messages with ISO8601 timestamps
## Summary

Graylog's SyslogCodec mishandles older RFC3164-style syslog messages that contain an ISO8601 timestamp. Example message that causes unwanted behavior:

`<6>2016-10-12T14:10:18Z hostname testmsg[20]: {"cf_origin":"firehose","deployment":"cf","event_type":"ValueMetric","ip":"127.0.0.1","job_index":"0","level_mod":"6","msg":"","name":"RequestLatency","origin":"bbs","time":"2016-10-12T14:10:18Z","unit":"nanos","value":2.6071897e+07}"`

## Expected Behavior

Section 5.1 of [RFC3164](https://www.ietf.org/rfc/rfc3164.txt) says that implementers may wish to use an ISO8601 timestamp, so the expected behavior is that this is parsed out properly.

## Current Behavior

In SyslogCodec, messages of the above format match the regex on [L64](https://github.com/Graylog2/graylog2-server/blob/2.1.1/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java#L64), which cause it to be parsed to RFC5424 standards. This causes a mangling of the fields, with the timestamp, application name, and process id becoming random words in the message part.

## Possible Solution

On line 64 of [SyslogCodec](https://github.com/Graylog2/graylog2-server/blob/2.1.1/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java#L64), change the regex from:
`<\\d+>\\d.*`
to:
`<\\d{1,3}>[1-9]\\d{0,2}\\s.*`

RFC5424 defines the syslog header to start with `<PRI>VER`, where PRI is a 1-3 digit number from 0-191, and VER is a 1-3 digit number from 1-999. Matching this will ensure that RFC3164 syslog messages with ISO8601 timestamps aren't interpreted as RFC5424 messages.

This will also require a small update to syslog4j-graylog2 to prevent the timestamp from throwing a date parsing exception and getting replaced with Date(). I'll submit an issue to that project as well.

1. Run Graylog 2.1.1 and set up a Syslog TCP stream
2. Save the above message to a file, let's call it testmsg
3. `cat testmsg|nc grayloghost port`
4. Check Graylog, find mangled fields in the message.

## Context

This was discovered when trying to use Graylog to collect stats and other messages from Pivotal Cloud Foundry using the firehose-to-syslog nozzle provided to funnel stats into syslog messages. Due to the message content being JSON, the mangling of the message field causes the JSON extractor to fail on a large portion of messages.

## Your Environment

* Graylog Version: 2.1.1
* Elasticsearch Version: 2.3.3
* MongoDB Version: 3.2.6
* Operating System: RHEL 7.2
* Browser version: Chrome 52

joschi:
Closes #2954

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java
@@ -21,15 +21,15 @@ import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.inject.assistedinject.Assisted;
 import com.google.inject.assistedinject.AssistedInject;
-import org.graylog2.plugin.ResolvableInetSocketAddress;
-import org.graylog2.plugin.inputs.annotations.Codec;
-import org.graylog2.plugin.inputs.annotations.ConfigClass;
-import org.graylog2.plugin.inputs.annotations.FactoryClass;
 import org.graylog2.plugin.Message;
+import org.graylog2.plugin.ResolvableInetSocketAddress;
 import org.graylog2.plugin.Tools;
 import org.graylog2.plugin.configuration.Configuration;
 import org.graylog2.plugin.configuration.ConfigurationRequest;
 import org.graylog2.plugin.configuration.fields.BooleanField;
+import org.graylog2.plugin.inputs.annotations.Codec;
+import org.graylog2.plugin.inputs.annotations.ConfigClass;
+import org.graylog2.plugin.inputs.annotations.FactoryClass;
 import org.graylog2.plugin.inputs.codecs.AbstractCodec;
 import org.graylog2.plugin.inputs.codecs.CodecAggregator;
 import org.graylog2.plugin.inputs.transports.NettyTransport;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java
```diff
@@ -91,10 +89,7 @@ public Message decode(@Nonnull RawMessage rawMessage) {
                 remoteAddress = address.getInetSocketAddress();
             }
             return parse(msg, remoteAddress == null ? null: remoteAddress.getAddress(), rawMessage.getTimestamp());
-        } catch (ClassCastException e) {
```
bernd: Why are you removing this?

joschi: It didn't do anything. [`ClassCastException`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassCastException.html) is a `RuntimeException` and the `catch` clause simply propagated (i. e. re-threw) the exception.

bernd: Got it. :+1:

PR #3230, opened 2016-12-21
Remove special handling in SplitAndCountConverter

### LINKED ISSUE #3067: The split count converter does not work for count 1 unit not having a separator
## Expected Behavior

In the extractors `split_and_count` converter does not extract and count string with no separator:

**1. - For example, if I declare a separator "white space" in split_and_count converter and process this string :**
```
Word1 Word2 Word3
```
 --> This gives me the counted value: 3 **is a valid value**

**2. -  For example, if I declare a separator "white space" in split_and_count converter and process this string  (which has no white space):**
```
Word1
```
   --> This gives me the counted value: 0 **is not a valid value**

**-->The count value in my second case should be: 1**

## Context
Each customer request I try to count the number of words entered in the search engine field

## Your Environment

* AWS official Graylog AMI 
* Graylog 2.0.2 (4da1379) 
* Oracle Corporation 1.8.0_77
* Ubuntu 14  Linux 3.13.0-86-generic
* Chrome Version 54.0.2840.71 m

joschi:
Closes #3067


UPGRADING.rst
```diff
@@ -47,3 +47,22 @@ The exposed configuration settings on the ``/system/configuration`` resource of
...
+Examples:
+
++-------------+---------------+------------+------------+
+| String      | Split Pattern | Old Result | New Result |
++=============+===============+===========+=============+
+| <empty>     | ``-``         | 0         | 0           |
++-------------+---------------+-----------+-------------+
+| ``foo``     | ``-``         | 0         | 1           |
++-------------+---------------+-----------+-------------+
+| ``foo-bar`` | ``-``         | 1         | 1           |
```
edmundoa: This example would return 2 in both the new and old implementations.

PR #3242, opened 2016-12-22
Fix URL for throbber image

### LINKED ISSUE #3240: Fix throbber loading in API browser
## Expected Behavior
Image used as throbber is loaded to indicate progress of the current API call.

## Current Behavior
Fetching throbber is using a wrong URL and returning a 404.

## Your Environment
* Graylog Version: 2.2.0-beta.4-SNAPSHOT (a93ca5a55fbe0e881011b3616d1094b3809e3b93)

edmundoa:
Adds `api-browser` to image URL.

Fixes #3240


joschi:
@edmundoa I'm wondering if `graylog2-server/src/main/resources/swagger/swagger-ui.min.js` is used at all and if we should delete it if not.

edmundoa:
@joschi I think it gets compiled "automagically" and we can remove it, will take a look 👍

PR #3257, opened 2016-12-27
Fix alert condition validations

### LINKED ISSUE #3247: Alert condition configuration type issues
## Expected Behavior

The alert condition `parameters` should have the correct types.

## Current Behavior

I am currently getting an exception when the email alarm callback tries to get the `time` field from the alert condition parameters.

```
2016-12-22 14:21:56,195 ERROR: org.graylog2.alarmcallbacks.EmailAlarmCallback - Alarm callback has email recipients and is triggered, but sending emails failed
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
	at org.graylog2.alerts.StaticEmailAlertSender.buildStreamDetailsURL(StaticEmailAlertSender.java:168) ~[classes/:?]
	at org.graylog2.alerts.FormattedEmailAlertSender.getModel(FormattedEmailAlertSender.java:108) ~[classes/:?]
	at org.graylog2.alerts.FormattedEmailAlertSender.buildSubject(FormattedEmailAlertSender.java:85) ~[classes/:?]
	at org.graylog2.alerts.StaticEmailAlertSender.sendEmail(StaticEmailAlertSender.java:112) ~[classes/:?]
	at org.graylog2.alerts.StaticEmailAlertSender.sendEmails(StaticEmailAlertSender.java:224) ~[classes/:?]
	at org.graylog2.alarmcallbacks.EmailAlarmCallback.call(EmailAlarmCallback.java:97) ~[classes/:?]
	at org.graylog2.alerts.AlertScanner.handleTriggeredCheckResult(AlertScanner.java:66) ~[classes/:?]
	at org.graylog2.alerts.AlertScanner.checkAlertCondition(AlertScanner.java:101) ~[classes/:?]
	at org.graylog2.periodical.AlertScannerThread.lambda$doRun$0(AlertScannerThread.java:82) ~[classes/:?]
	at java.util.ArrayList.forEach(ArrayList.java:1249) [?:1.8.0_111]
	at org.graylog2.periodical.AlertScannerThread.doRun(AlertScannerThread.java:82) [classes/:?]
	at org.graylog2.plugin.periodical.Periodical.run(Periodical.java:77) [classes/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_111]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_111]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_111]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [?:1.8.0_111]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [?:1.8.0_111]
	at java.lang.Thread.run(Thread.java:745) [?:1.8.0_111]
```

This is because the `time` field in the `parameters` object has the wrong type. (string instead of integer)

```json
{
	"alert_conditions" : [
		{
			"creator_user_id" : "admin",
			"created_at" : "2016-12-22T12:42:10.735Z",
			"id" : "76f663ac-d5ec-4529-8455-48c63611c89c",
			"type" : "message_count",
			"title" : "Lots of yolos",
			"parameters" : {
				"grace" : "1",
				"threshold" : "5",
				"threshold_type" : "MORE",
				"time" : "1",
				"backlog" : "2"
			}
		}
	],
}
```

The code in the email alarm callback hasn't changed since 2014 so something in the creation of the alert conditions seems to be wrong.

1. Create "message count" alert condition for a stream
2. Create email alarm callback for the same stream
3. Trigger the alert condition
4. See error in logs

## Your Environment

* Graylog Version: 2.2.0-beta.4-SNAPSHOT 5e28b1fea
* MongoDB Version: 2.4.14
* Browser version: 55.0.2883.87

edmundoa:
As described in #3247, the pluggable alert condition forms introduced in 2.2 are propagating data using the wrong data types to the server, which is also failing to fix or properly validate that data, resulting in wrong objects being written into MongoDB.

To avoid this situation, this PR is doing the following:
- Convert values in the REST request according to the alert condition type definition
- Check the alert condition configuration to ensure it follows the type definition
- Standarize alert condition plugin definitions, so they look and work as other available plugins
- Remove custom alert condition forms. It sounded like a good idea having custom forms and falling back to `ConfigurationForm` if no custom form was given, but in practice this does not work as well: the code for each form is slightly different, leading to validation issues, errors due to missing methods, or holding state in different components, etc. They may also look quite different, adding inconsistencies to the user as well. IMHO we need to sort out how we want them to look, and how the API should be, and then re-introduce them if needed
- Improve alert condition descriptions, making more clear what each option do

Fixes #3247


graylog2-server/src/main/java/org/graylog2/alerts/types/FieldValueAlertCondition.java
```diff
@@ -96,19 +93,21 @@ public Config() {
         public ConfigurationRequest getRequestedConfiguration() {
             final ConfigurationRequest configurationRequest = ConfigurationRequest.createWithFields(
                     new TextField("field", "Field", "", "Field name that should be checked", ConfigurationField.Optional.NOT_OPTIONAL),
-                    new NumberField("time", "Time Range", 0, "Time span in seconds to check", ConfigurationField.Optional.NOT_OPTIONAL),
-                    new NumberField("threshold", "Threshold", 0.0, "Value which triggers an alert if crossed", ConfigurationField.Optional.NOT_OPTIONAL),
+                    new NumberField("time", "Time Range", 5, "Number of minutes to evaluate in the condition (0 evaluates condition in all messages)", ConfigurationField.Optional.NOT_OPTIONAL),
```
bernd: I had a hard time to understand the description. Maybe something like "Evaluate the condition for all messages received in the given number of minutes"?

graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionFactory.java
```diff
@@ -34,16 +40,31 @@ public AlertConditionFactory(Map<String, AlertCondition.Factory> alertConditionM
...
+         * Here we strictly use the Configuration object to verify the configuration and don't pass it down to
+         * the factory. The reason for this is that Configuration only support int values, but at least an
+         * alert condition expects a double.
+         */
+        try {
+            final ConfigurationRequest requestedConfiguration = factory.config().getRequestedConfiguration();
+            final Configuration configuration = new Configuration(parameters);
+            requestedConfiguration.check(configuration);
+        } catch (ConfigurationException e) {
+            LOG.error("Could not load alert condition <" + id + ">, invalid configuration detected.");
```
edmundoa: The only problem is that it may not be available (titles for alert conditions are only there since 2.1). Anyway, we can add it if available 👍

graylog2-server/src/main/java/org/graylog2/alerts/types/FieldValueAlertCondition.java
edmundoa: 👍

edmundoa:
I took care of the mentioned issues and also rebased against master.

Issue #3235, opened 2016-12-22
Notification details are not reloaded when updated on condition page


bernd:
## Expected Behavior

When editing the notification settings on the condition details page (e.g. `/alerts/conditions/58458e442f857c314491344e/76f663ac-d5ec-4529-8455-48c63611c89c`) the details of the notification should be updated after saving it.

## Current Behavior

After saving the notification the details for the notification stay the same and are not updated. When editing and saving a notification on the notifications page, the data is reloaded.

## Possible Solution

Reload notification settings after saving them on the conditions page.

## Steps to Reproduce (for bugs)

1. Create condition
2. Create notification
3. Go to the condition details page for the condition
4. Edit and save the settings for the notification
5. Check notification details for the updated values

## Your Environment

* Graylog Version: 2.2.0-beta.4-SNAPSHOT commit 5e28b1fea

edmundoa:
Refs https://github.com/Graylog2/graylog2-server/issues/3195

edmundoa:
Fixed in #3264

PR #3264, opened 2016-12-28
Fix reloading after notification changes

### LINKED ISSUE #3195: 2.2 UX/UI Improvement: Alerts - delete Notification from condition does no refresh
## Expected Behavior
If you look at the details of an alert condition you can **edit** or **delete** the notifications. After Edit or delete the page should be reloaded to reflect the new information.

## Current Behavior
Short notification appears at the bottom of the page and the page/information is not reloaded. You need to reload the page your self. 

When you edit the condition details the page is refreshed with the new settings for the condition but the notification stay in the old values/status.

1. create condition for a stream
2. create notification for a stream
3. look at condition details page
4. change / delete Notifications

## Your Environment

* Graylog Version: Graylog 2.2.0-beta.2+7cda131 (OVA)
* Browser: Safari Version 10.0.1 (12602.2.14.0.7)

edmundoa:
Move decision of which data to load up the component hierarchy, fixing UI inconsistencies after editing or deleting alert notifications from the alert conditions page.

Fixes #3195 and #3235

### DIFF
--- a/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
+++ b/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
@@ -3,7 +3,7 @@ import Reflux from 'reflux';
 import { Button, Col, DropdownButton, MenuItem } from 'react-bootstrap';
 
 import CombinedProvider from 'injection/CombinedProvider';
-const { AlertNotificationsStore, AlertNotificationsActions } = CombinedProvider.get('AlertNotifications');
+const { AlertNotificationsStore } = CombinedProvider.get('AlertNotifications');
 const { AlarmCallbacksActions } = CombinedProvider.get('AlarmCallbacks');
 
 import { EntityListItem } from 'components/common';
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
```diff
@@ -35,13 +37,21 @@ const AlertNotification = React.createClass({
 
   _onSubmit(data) {
     AlarmCallbacksActions.update(this.props.alertNotification.stream_id, this.props.alertNotification.id, data)
-      .then(() => AlertNotificationsActions.listAll());
+      .then(() => {
+        if (typeof this.props.onNotificationUpdate === 'function') {
```
bernd: Wouldn't a simple check for undefined/null be enough here since we specify the propTypes?

bernd:
Please rebase against master to fix merge conflict. Thanks!

graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
edmundoa: We are very inconsistent with that, but I think it is safer to ensure it is a function when using it. The prop type check may not always be there (for instance if we ever enable the production mode in react), and it only logs a warning, so it will not prevent the function to be called, and the application would be broken in that case.

edmundoa:
Rebased against master, thank you for reviewing!

PR #2972, opened 2016-10-18
Prevent n+1 query loading for Stream Rules

bjoernhaeuser:
## Description

The issue is descriped in Graylog2/graylog2-server#2805. Stream Rules are ineffienently loaded, by doing a query pro stream and then a query pro rule.

## Motivation and Context

This tries to fix this by loading all Rules for a given list of steram ids.

## How Has This Been Tested?

Tested the resulted jar file and integration tests are also still running.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Add StreamRuleService.loadForStreamIds method for efficiently loading
all Rules for a given list of streams.

**This is not the final verson. This is to get early feedback on the suggested implementation**

PR #3272, opened 2016-12-30
Micro batch loading grace periods - WIP -> Feedback requested

bjoernhaeuser:
## Description

In #2972 we made the stream page loading much faster, but it in our
setup it is still slow (3-10 seconds). After some benchmarking I figured
that most of the time is spent in the "in grace period" check.

I tried to see if that field is anywhere used, but I found not usage for
it. Nonetheless removing it from the API seems a little bit too much for
me.

Therefore I started to implement a "micro" batching for this. So, in the
first step load all "last triggered_at" datetime for all given
conditions and streams and than calculate the "in grace" from that.

## Motivation and Context

Prevent slow loading of the "streams" page and also the "streams" api.

## How Has This Been Tested?

Not really yet. Tests will also fail. I wanted to get early feedback on this before proceeding. 

Before:
screen shot 2016-12-30 at 17 48 18

After:

screen shot 2016-12-30 at 17 44 52

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
@@ -17,10 +17,13 @@
 package org.graylog2.alerts;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.mongodb.BasicDBObject;
+import com.google.common.collect.Maps;
 import com.mongodb.DBCollection;
+import org.apache.commons.lang3.time.StopWatch;
 import org.graylog2.alerts.Alert.AlertState;
 import org.graylog2.bindings.providers.MongoJackObjectMapperProvider;
 import org.graylog2.database.CollectionName;
@@ -35,11 +38,15 @@ import org.graylog2.rest.models.streams.alerts.requests.CreateConditionRequest;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Seconds;
+import org.mongojack.Aggregation;
+import org.mongojack.AggregationResult;
 import org.mongojack.DBQuery;
 import org.mongojack.DBSort;
+import org.mongojack.Id;
 import org.mongojack.JacksonDBCollection;
 
 import javax.inject.Inject;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AlertServiceImpl.java
```diff
@@ -102,6 +109,43 @@ public int triggeredSecondsAgo(String streamId, String conditionId) {
...
+
+        final AggregationResult<AlertLastTriggeredAt> aggregate = this.coll.aggregate(group, AlertLastTriggeredAt.class);
+        aggregate.results().forEach(alertLastTriggeredAt -> {
+            final Map<String, DateTime> stream = lastTriggeredAt.getOrDefault(alertLastTriggeredAt.id.get("stream_id"), Maps.newHashMap());
+
+            stream.put(alertLastTriggeredAt.id.get("condition_id"), alertLastTriggeredAt.lastTriggeredAt);
+        });
+
+        stopWatch.stop();
+        System.out.println("getLastTriggeredAt: " + stopWatch.getTime());
```
bjoernhaeuser: In our test setup this takes ~400ms.

dennisoelkers:
@bjoernhaeuser: Thank you very much for submitting this. Sorry for not getting back to you earlier, we were all rather busy with the release of 2.2.0. In order to review it, can you please rebase it on current master?

bjoernhaeuser:
@dennisoelkers I rebased to current master. Just wanted to repeat that this is only a possible solution to reduce the time the it takes to load `/streams`.

edmundoa:
@bjoernhaeuser Thank you for taking the time and effort to submit this change!

We have been discussing a bit what to do next regarding this issue and we think it would be best to remove the in grace period check from the list of streams, as we are not using those values at all. Anyway, we are still evaluating if that change would fit our next release, or if your approach would be more sensible until we reach the moment when we can do that API breaking change. We will keep the PR open until have a decision on that 🙂

bjoernhaeuser:
@edmundoa 

Thank YOU for considering this contribution. :) Do you have an idea when you will have the decision on that and in which release we could potentially see the fix for this?

The problem here (at least for us) is that this makes the web ui pretty cumbersome to use. Waiting 10 seconds (or more) for the streams and start page to load just makes everyone unhappy users of graylog :(

edmundoa:
Hi @bjoernhaeuser 

Next week we are discussing more about future releases, we should be able to tell more after those meetings and go one way or the other. I know is taking too long and it is an annoying issue, but hopefully this can be fixed soon.

bjoernhaeuser:
@edmundoa are there any news on this? People start screaming for a solution :)

dennisoelkers:
Hey @bjoernhaeuser, thanks for your patience. We are very busy with pushing forward feature wise to the next versions. Nonetheless I prepared a small PR today, which is just removing the (currently unused) `in_grace` field from the stream list API call. We are now discussing if we are including it in `2.3` or `3.0` (as it is a breaking change by nature). We will keep you posted.

Thanks for contributing and doing this more complete PR, we would love to see more contributions of that kind from you in the future. We would have merged it, if we had actual use for the flag in that place. As this is not the case right now, we decided to push for a simpler solution. I will close this now in favor of #3736.

bjoernhaeuser:
Thanks!

PR #3277, opened 2017-01-02
Add list of stream IDs to Message#toElasticSearchObject()

joschi:
The `Message#toElasticSearchObject()` only populated the `streams` field of the generated Elasticsearch document if the `Message.streams` field was filled, but not if the `streams` field in the `Message.fields` map was being used.

This lead to inconsistent "exports" of the internal `Message` representation.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -25,7 +25,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.net.InetAddresses;
@@ -38,7 +37,6 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import java.net.InetAddress;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
@@ -47,7 +45,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import static com.google.common.base.Predicates.equalTo;
 import static com.google.common.base.Predicates.not;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -216,6 +217,7 @@ public DateTime getTimestamp() {
 
         obj.put(FIELD_MESSAGE, getMessage());
         obj.put(FIELD_SOURCE, getSource());
+        obj.put(FIELD_STREAMS, getStreamIds());
```
bernd: This doesn't fix the issue with the archive restore.

In the restore case the stream IDs are in the `fields` map but the `streams` set is empty and thus the stream IDs in `fields` will be cleared.

bernd: Also, I am not sure if we really should use `getStreamIds()` here because it's using stream functions. I guess those create more objects than the simple for loop we had before?

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -466,18 +455,22 @@ public boolean removeStream(Stream stream) {
...
-            return Collections.emptyList();
+            streamField = Collections.emptyList();
+        }
+
+        final Set<String> streamIds = streamField == null ? new HashSet<>(streams.size()) : new HashSet<>(streamField);
+        for (Stream stream : streams) {
+            streamIds.add(stream.getId());
         }
+
+        return new ArrayList<>(streamIds);
```
bernd: How about changing the signature of `getStreamIds()` to return a `Set` or create a different method for this case? This would avoid the additional list creation.

I just want to be super careful here because this stuff is called for every message that's written into Elasticsearch and now we are creating one HashSet and one ArrayList for every call. Previously this was one ArrayList if the message had streams.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -466,18 +455,22 @@ public boolean removeStream(Stream stream) {
...
     public List<String> getStreamIds() {
-        if (!hasField(FIELD_STREAMS)) {
-            return streams.stream().map(Stream::getId).collect(Collectors.toList());
-        }
+        List<String> streamField;
         try {
-            @SuppressWarnings("unchecked")
-            final List<String> streamIds = getFieldAs(List.class, FIELD_STREAMS);
-            return new ArrayList<>(streamIds);
+            streamField = getFieldAs(List.class, FIELD_STREAMS);
```
bernd: I found out why this still fails with the archive restore. The archive creates a message with a `Set<String>` for the streams field in the map.

bernd: Thanks `Map<String, Object>`...

joschi: :+1:

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
joschi: I'm totally *for* changing the signature of this method, even if it's breaking the API in this case.

bernd:
LGTM :+1: Also fixes the archive issue.

PR #3284, opened 2017-01-03
Move client-side split/import of Grok pattern files to server

### LINKED ISSUE #3003: Failure to import GROK patterns from a Windows-formatted text file
## Expected Behavior

If a text file containing multiple lines with GROK patterns, one pattern per line, is created on Windows machine, it should successfully get imported in Graylog.
## Current Behavior

If the text file with patterns has CR+LF as line ending markers, Graylog web interface fails to import the file.  The page reports success, but no patterns get imported.  If the last line of the file does not have a "new line" marker, the last line gets imported, but all others fail.  Changing line ending markers from CR+LF to LF only (Linux format) allows the file to be imported successfully.
## Possible Solution

Changing line ending markers from CR+LF to LF only (Linux format) allows the file to be imported successfully.
1.  Using Windows Notepad or Notepad++ with "EOL Conversion" set to windows, create a test text file with several lines of pattern name -- space -- pattern REGEX. Save the file.
2. Attempt to import the pattern file via Graylog web GUI
3. Receive import success confirmation
4. Observe that no patterns got added to the list
## Context

Configuring custom message parsers for specialized network gear, like Cisco ASA firewalls
## Your Environment

Standard latest Graylog OVA deployed to ESXi 5.5, patched, graylog storage extended via additional virtual disk.
- Graylog Version: Graylog 2.1.1+01d50e5 on graylog (Oracle Corporation 1.8.0_101 on Linux 4.2.0-42-generic)
- Elasticsearch Version: 2.3.1 (a guess)
- MongoDB Version: 2.6.9 (a guess)
- Operating System: Windows 10 Redstone 1
- Browser version: Firefox 49.0.1

joschi:
Fixes #3003

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
@@ -140,7 +140,7 @@ public class GrokResource extends RestResource {
         return Response.accepted().build();
     }
 
-    @PUT
+    @POST
     @Consumes(MediaType.TEXT_PLAIN)
     @Timed
     @ApiOperation("Add a list of new patterns")
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
@@ -47,10 +47,20 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/logic/rest/FetchProvider.js
```diff
@@ -108,3 +144,22 @@ export default function fetch(method, url, body) {
   }
   return promise();
 }
+
+export function fetchPlainText(method, url, body) {
```
joschi: I'm a bit unhappy about this code duplication but didn't want to refactor the whole `FetchProvider` to generically support something other than `application/json` in this PR as it's being used in quite a *few* places…

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
```diff
@@ -130,6 +141,47 @@ public Response bulkUpdatePatterns(@ApiParam(name = "patterns", required = true)
     }
 
     @PUT
```
bernd: I think this should be a `@POST` because it depends on the `replace` param if this method it idempotent.

PR #3290, opened 2017-01-03
Escape curly braces in Swagger resource URIs for challenged browsers

### LINKED ISSUE #2710: On rest api browser page doc not expanding
On api browser page  when try to click on `show/hide` or `List Operations` page is not expanding. 
## Your Environment

- Graylog Version: 2.1.0-beta.3+b817da4

joschi:
Fixes #2710


joschi:
With this change, the API browser also seems to work for Safari (10.0.2) again.

PR #3291, opened 2017-01-03
Fix document counts with an empty index set

bernd:
Return 0 if there are no indices in the given index set. If we run the query with an empty index list, Elasticsearch will count all documents in all indices and thus return a wrong count.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
@@ -24,8 +24,6 @@ import org.graylog2.indexer.IndexSetRegistry;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
-import static java.util.Objects.requireNonNull;
-
 @Singleton
 public class Counts {
     private final Client c;
@@ -48,7 +46,7 @@ public class Counts {
     private long totalCount(final String[] indexNames) {
         // Return 0 if there are no indices in the given index set. If we run the query with an empty index list,
         // Elasticsearch will count all documents in all indices and thus return a wrong count.
-        if (requireNonNull(indexNames).length == 0) {
+        if (indexNames.length == 0) {
             return 0L;
         }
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -45,6 +45,12 @@ public long total() {
     public long total(final IndexSet indexSet) {
         final String[] names = indexSet.getManagedIndicesNames();
 
+        // Return 0 if there are no indices in the given index set. If we run the query with an empty index list,
+        // Elasticsearch will count all documents in all indices and thus return a wrong count.
+        if (names.length == 0) {
```
joschi: This check should probably also be added to the `Counts#total()` method.

If none of the index set contains an index and other indices (not managed by Graylog) exist in the Elasticsearch cluster, the returned value might be wrong.

bernd: Good point, fixed. :+1:

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -36,22 +38,21 @@ public Counts(Client client, IndexSetRegistry indexSetRegistry) {
...
     public long total(final IndexSet indexSet) {
-        final String[] names = indexSet.getManagedIndicesNames();
+        return totalCount(indexSet.getManagedIndicesNames());
+    }
 
+    private long totalCount(final String[] indexNames) {
         // Return 0 if there are no indices in the given index set. If we run the query with an empty index list,
         // Elasticsearch will count all documents in all indices and thus return a wrong count.
-        if (names.length == 0) {
+        if (requireNonNull(indexNames).length == 0) {
```
joschi: Why the `requireNonNull()` here?

If `indexNames` is `null`, it will throw an NPE anyway.

Issue #3254, opened 2016-12-26
Graylog-server error: "Couldn't calculate index range for index "logstash-2016.11.25"..."


deanilenko:
Hello Guys!
We have an elasticsearch node which shared across logstash, ntopng, graylog, etc. Every time we run a curator which deletes indices which are older than 30 days old we see this error message in graylog file log: 

`2016-12-26T17:24:35.139+02:00 ERROR [MongoIndexRangeService] Couldn't calculate index range for index "logstash-2016.11.25"
org.elasticsearch.index.IndexNotFoundException: no such index
	at org.elasticsearch.cluster.metadata.IndexNameExpressionResolver$WildcardExpressionResolver.resolve(IndexNameExpressionResolver.java:585) ~[graylog.jar:?]
	at org.elasticsearch.cluster.metadata.IndexNameExpressionResolver.concreteIndices(IndexNameExpressionResolver.java:133) ~[graylog.jar:?]
	at org.elasticsearch.cluster.metadata.IndexNameExpressionResolver.concreteIndices(IndexNameExpressionResolver.java:113) ~[graylog.jar:?]
	at org.elasticsearch.action.search.AbstractSearchAsyncAction.<init>(AbstractSearchAsyncAction.java:99) ~[graylog.jar:?]
	at org.elasticsearch.action.search.SearchQueryThenFetchAsyncAction.<init>(SearchQueryThenFetchAsyncAction.java:53) ~[graylog.jar:?]
	at org.elasticsearch.action.search.TransportSearchAction.doExecute(TransportSearchAction.java:93) ~[graylog.jar:?]
	at org.elasticsearch.action.search.TransportSearchAction.doExecute(TransportSearchAction.java:47) ~[graylog.jar:?]
	at org.elasticsearch.action.support.TransportAction.doExecute(TransportAction.java:149) ~[graylog.jar:?]
	at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:137) ~[graylog.jar:?]
	at org.elasticsearch.action.support.TransportAction.execute(TransportAction.java:85) ~[graylog.jar:?]
	at org.elasticsearch.client.node.NodeClient.doExecute(NodeClient.java:58) ~[graylog.jar:?]
	at org.elasticsearch.client.support.AbstractClient.execute(AbstractClient.java:359) ~[graylog.jar:?]
	at org.elasticsearch.client.FilterClient.doExecute(FilterClient.java:52) ~[graylog.jar:?]
[...]

PR #3310, opened 2017-01-06
Make sure that index ranges get deleted when an index set is removed

### LINKED ISSUE #3308: Deleting an index set and its associated indices fails to delete index ranges
When deleting an index set and its associated indices, Graylog fails to delete the corresponding index ranges, leading to orphaned entries in the `index_ranges` collection.

bernd:
The old index set cleanup job left the index ranges for the deleted indices behind.

Fixes #3308

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/jobs/IndexSetCleanupJob.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/jobs/IndexSetCleanupJob.java
@@ -21,6 +21,7 @@ import org.elasticsearch.indices.IndexTemplateMissingException;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.indexset.IndexSetConfig;
 import org.graylog2.indexer.indices.Indices;
+import org.graylog2.indexer.ranges.MongoIndexRangeService;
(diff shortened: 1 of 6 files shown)


joschi:
Wouldn't it be easier if we simply removed the check in `MongoIndexRangeService#handleIndexClosing()`?
If a non-managed index is being closed, there shouldn't be an index range for it anyway.

https://github.com/Graylog2/graylog2-server/blob/8c1922caa58dee327305fa41568ee176c79888eb/graylog2-server/src/main/java/org/graylog2/indexer/ranges/MongoIndexRangeService.java#L188-L191

bernd:
> Wouldn't it be easier if we simply removed the check in `MongoIndexRangeService#handleIndexClosing()`? If a non-managed index is being closed, there shouldn't be an index range for it anyway.

That would be easier but it makes the code even harder to understand. I actually had a hard time to find out how the index ranges get deleted because this works completely different to other `*Service` classes.  

We added the `.isManagedIndex()` calls to fix #3254, so if we remove it for `handleIndexDeletion()` it's a special case for the index range cleanup code that makes the whole thing even harder to understand in my opinion.

That's why I prefer to use an explicit `remove()` method.

PR #3315, opened 2017-01-06
Close idle GELF HTTP connections after a timeout

### LINKED ISSUE #3223: Possible file descriptor leak in Graylog Server
* Graylog Version: 2.1.2
* Elasticsearch Version: 2.3.5
* MongoDB Version: 3.2.9
* Operating System: Ubuntu 14.04.5

We're running 2.1.2 but the issue was also happening with 2.1.1 at least, updating didn't solve it. This is a one node setup.

We have the same issue as #2823 . I see an steady increase in open fds resulting in an eventual crash until the process is restarted. FWIW most errors we're seeing are not the Elasticsearch health that gitfrederic reported, instead they're mostly:

2016-12-21T10:45:46.176Z ERROR [NettyTransport] Error in Input [GELF HTTP/57b48de2bfa5c34ad4a1eed9] (channel [id: 0xeb609605, /IP:13581 => /IP:12443])
java.io.IOException: Broken pipe

But the end result is the same, ever increasing fds until a crash. Another data point is that the increase isn't monotonic, but it does keep growing as the hours pass.

bernd:
This introduces a new GELF HTTP input option to configure a timeout for idle HTTP connections.

Some clients (e.g. browsers) might not close the HTTP connection properly and effectively cause a file descriptor leak in the server. The default for all existing and new GELF HTTP inputs is 60 seconds.

Fixes #3223

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
@@ -115,7 +115,7 @@ public class HttpTransport extends AbstractTcpTransport {
         this.timer = timer;
         int maxChunkSize = configuration.intIsSet(CK_MAX_CHUNK_SIZE) ? configuration.getInt(CK_MAX_CHUNK_SIZE) : DEFAULT_MAX_CHUNK_SIZE;
         this.maxChunkSize = maxChunkSize <= 0 ? DEFAULT_MAX_CHUNK_SIZE : maxChunkSize;
-        this.idleWriterTimeout = configuration.intIsSet(CK_MAX_CHUNK_SIZE) ? configuration.getInt(CK_IDLE_WRITER_TIMEOUT, DEFAULT_IDLE_WRITER_TIMEOUT) : DEFAULT_IDLE_WRITER_TIMEOUT;
+        this.idleWriterTimeout = configuration.intIsSet(CK_IDLE_WRITER_TIMEOUT) ? configuration.getInt(CK_IDLE_WRITER_TIMEOUT, DEFAULT_IDLE_WRITER_TIMEOUT) : DEFAULT_IDLE_WRITER_TIMEOUT;
     }
 
     private static Executor executorService(final String executorName, final String threadNameFormat, final MetricRegistry metricRegistry) {
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
@@ -54,10 +54,7 @@ import org.jboss.netty.handler.codec.http.HttpResponse;
 import org.jboss.netty.handler.codec.http.HttpResponseEncoder;
 import org.jboss.netty.handler.codec.http.HttpResponseStatus;
 import org.jboss.netty.handler.codec.http.HttpVersion;
-import org.jboss.netty.handler.timeout.IdleState;
-import org.jboss.netty.handler.timeout.IdleStateAwareChannelHandler;
-import org.jboss.netty.handler.timeout.IdleStateEvent;
-import org.jboss.netty.handler.timeout.IdleStateHandler;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
```diff
@@ -184,9 +217,24 @@ public ConfigurationRequest getRequestedConfiguration() {
                                         DEFAULT_MAX_CHUNK_SIZE,
                                         "The maximum HTTP chunk size in bytes (e. g. length of HTTP request body)",
                                         ConfigurationField.Optional.OPTIONAL));
+            r.addField(new NumberField(CK_IDLE_WRITER_TIMEOUT,
```
joschi: Maybe add `NumberField.Attribute.ONLY_POSITIVE` here.

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
```diff
@@ -118,6 +131,26 @@ private static Executor executorService(final String executorName, final String
...
 
+        if (idleWriterTimeout > 0) {
+            // Install IdleState and IdleStateEvent handler to close idle connections after a timeout.
+            // This avoids dangling HTTP connections when the HTTP client does not close the connection properly.
+            // For details see: https://github.com/Graylog2/graylog2-server/issues/3223#issuecomment-270350500
+
+            baseChannelHandlers.put("idlestate", new Callable<ChannelHandler>() {
+                @Override
+                public ChannelHandler call() throws Exception {
+                    return new IdleStateHandler(timer, 0, idleWriterTimeout, 0, TimeUnit.SECONDS);
```
joschi: Why use an `IdleStateHandler` instead of [`WriteTimeoutHandler`](https://netty.io/3.10/api/org/jboss/netty/handler/timeout/WriteTimeoutHandler.html) here if only the write timeout is being used?

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
```diff
@@ -101,8 +112,10 @@ public HttpTransport(@Assisted Configuration configuration,
 
         enableCors = configuration.getBoolean(CK_ENABLE_CORS);
 
+        this.timer = timer;
         int maxChunkSize = configuration.intIsSet(CK_MAX_CHUNK_SIZE) ? configuration.getInt(CK_MAX_CHUNK_SIZE) : DEFAULT_MAX_CHUNK_SIZE;
         this.maxChunkSize = maxChunkSize <= 0 ? DEFAULT_MAX_CHUNK_SIZE : maxChunkSize;
+        this.idleWriterTimeout = configuration.intIsSet(CK_MAX_CHUNK_SIZE) ? configuration.getInt(CK_IDLE_WRITER_TIMEOUT, DEFAULT_IDLE_WRITER_TIMEOUT) : DEFAULT_IDLE_WRITER_TIMEOUT;
```
joschi: This checks for the wrong configuration key.

bernd: Oups, good catch! :+1:

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
bernd: Because I wasn't aware of it. Changed the code to use a ReadTimeoutHandler because I want to close the connection when the **client** does not write anymore. Got it mixed up. :wink:

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
bernd: :+1:

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpTransport.java
joschi: 👍

PR #3332, opened 2017-01-10
Add more metrics for extractors

### LINKED ISSUE #2959: Add timer metric to extractor condition
Currently extractors have two different timers:
- `executionTime`
- `converterExecutionTime`

However, the condition is evaluated outside of the `executionTime` and we also do not track how many extractor conditions prevented running the extractor.

Add another timer which tracks the condition execution and counts how many times the condition was evaluated.
This helps catching slow (potentially backtracking) regex conditions.

joschi:
This change adds the following metrics for extractors:

* conditionHits (number of times the condition succeeded)
* conditionMisses (number of times the condition failed)
* conditionTime (time spent in evaluating the condition)
* completeExecutionTime (total time spent in the extractor)

Fixes #2959

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
@@ -114,13 +114,16 @@ public abstract class Extractor implements EmbeddedPersistable {
 
     private final Counter conditionHitsCounter;
     private final Counter conditionMissesCounter;
+    private final Timer conditionTimer;
+    private final Timer executionTimer;
+    private final Timer converterTimer;
+    private final Timer completeTimer;
     private final String conditionHitsCounterName;
     private final String conditionMissesCounterName;
     private final String conditionTimerName;
     private final String executionTimerName;
     private final String converterTimerName;
     private final String completeTimerName;
-
 
     protected abstract Result[] run(String field);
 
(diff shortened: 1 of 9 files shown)


joschi:
The diff is best viewed with whitespace ignored: https://github.com/Graylog2/graylog2-server/pull/3332/files?w=1

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -159,115 +164,122 @@ public Extractor(MetricRegistry metricRegistry,
...
 
-        this.totalTimerName = name(getClass(), getType().toString().toLowerCase(Locale.ENGLISH), getId(), "executionTime");
-        this.converterTimerName = name(getClass(), getType().toString().toLowerCase(Locale.ENGLISH), getId(), "converterExecutionTime");
+        final String metricsPrefix = name(getClass(), getType().toString().toLowerCase(Locale.ENGLISH), getId());
+        this.conditionHitsCounterName = name(metricsPrefix, "conditionHits");
+        this.conditionMissesCounterName = name(metricsPrefix, "conditionMisses");
+        this.conditionTimerName = name(metricsPrefix, "conditionTime");
+        this.executionTimerName = name(metricsPrefix, "executionTime");
+        this.converterTimerName = name(metricsPrefix, "converterExecutionTime");
+        this.completeTimerName = name(metricsPrefix, "completeExecutionTime");
```
kroepke: Can we also pre-create these timers like the counters below?
That should save another hash lookup and instanceOf check in `metricRegister.timer`, right?

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -159,115 +164,122 @@ public Extractor(MetricRegistry metricRegistry,
...
+                        msg.addField(result.getTarget(), result.getValue());
+                    }
+                }
+
+                // Remove original from message?
+                if (cursorStrategy.equals(CursorStrategy.CUT) && !targetField.equals(sourceField) && !Message.RESERVED_FIELDS.contains(sourceField) && results[0].beginIndex != -1) {
+                    final StringBuilder sb = new StringBuilder(field);
 
-        final Result[] results = run(field);
+                    final List<Result> reverseList = Arrays.stream(results)
```
kroepke: :+1:

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -159,115 +164,122 @@ public Extractor(MetricRegistry metricRegistry,
...
+                    } else {
+                        conditionMissesCounter.inc();
+                        return;
+                    }
+                }
             }
-        }
 
-        final Timer.Context timerContext = metricRegistry.timer(getTotalTimerName()).time();
+            try (final Timer.Context executionTimer = metricRegistry.timer(getExecutionTimerName()).time()) {
```
kroepke: nitpick: I prefer to either name these `ignored` or annotate them to avoid the IDE warnings (also elsewhere in this file)

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
joschi: ✅

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
joschi: ✅

PR #3337, opened 2017-01-11
warn when neither reader nor admin roles are selected for a user

### LINKED ISSUE #2116: Remove reader role possible but not advisible
I may be missing something on my install - 2.0 Beta 3, however the users / roles system isn't entirely intuitive as it is possible to remove the reader role from a user which in essence makes the user unable to do anything unless you have manually added the necessary permissions to a new role.

It would probably make more sense if it were not possible to remove the reader role from a user, or at the very least attempting to do so flagged up a warning.
1. Create User
2. Remove reader role
3. Attempt to use a stream with new user.

kroepke:
disallow saving without at least one of these roles present (because lots of pages require at least the reader permissions)

 fixes #2116

### DIFF
--- a/graylog2-web-interface/src/components/users/EditRolesForm.jsx
+++ b/graylog2-web-interface/src/components/users/EditRolesForm.jsx
@@ -55,9 +55,9 @@ const EditRolesForm = React.createClass({
     let rolesAlert = null;
     const roles = this.state.newRoles;
     if (roles != null && roles.indexOf('Reader') === -1 && roles.indexOf('Admin') === -1) {
-      rolesAlert = <Alert bsStyle="danger" role="alert" className={EditRolesFormStyle.rolesMissingAlert}>
+      rolesAlert = (<Alert bsStyle="danger" role="alert" className={EditRolesFormStyle.rolesMissingAlert}>
         You need to select at least one of the <em>Reader</em> or <em>Admin</em> roles.
-      </Alert>;
+      </Alert>);
     }
     const externalUser = user.external ?
       (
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/components/users/EditRolesForm.jsx
```diff
@@ -39,11 +43,22 @@ const EditRolesForm = React.createClass({
...
+    this.setState({ newRoles: roles });
+  },
   render() {
     const user = this.props.user;
     if (!this.state.roles) {
       return <Spinner />;
     }
+    let rolesAlert = null;
+    const roles = this.state.newRoles;
+    if (roles != null && roles.indexOf('Reader') === -1 && roles.indexOf('Admin') === -1) {
```
edmundoa: You could replace `roles.indexOf()` with [`roles.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) for better reading experience. The polyfill we use should take care of converting it if the browser does not support it.

graylog2-web-interface/src/components/users/EditRolesForm.jsx
```diff
@@ -39,11 +43,22 @@ const EditRolesForm = React.createClass({
...
+      rolesAlert = <Alert bsStyle="danger" role="alert" className={EditRolesFormStyle.rolesMissingAlert}>
```
edmundoa: Small nitpick: this triggers a [`react/wrap-multilines`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-wrap-multilines.md) styling warning.

graylog2-web-interface/src/components/users/EditRolesForm.jsx
kroepke: Nice, didn't know about it!

graylog2-web-interface/src/components/users/EditRolesForm.jsx
kroepke: fixed

kroepke:
I opted for a simpler approach by duplication the few lines of code, rather than passing around styling props and the validation state.

PR #3339, opened 2017-01-11
Prevent setting default index set readonly.

### LINKED ISSUE #3331: Read only index sets should not be able to become default
Currently it is possible to mark read only index sets to be the default index set.

Since they are not writable (such as restored archives) this makes no sense and will likely cause indexing errors.
This option should be disabled or at least display an error when it is invoked (like trying to remove the default index set displays).

dennisoelkers:
## Motivation and Context
Without this change it was possible to update the default index set and make it read only, leading to unwanted behavior. After this change, when updating an index set we check if it is the default index set and kindly refuse setting it read only.
    
Fixes #3331.

Additionally:

Removing superfluous id property from IndexSetUpdateRequest.
    
The id property of the IndexSetUpdateRequest DTO is unneeded, as the id of the index set to update is already given in the URL. It is actually misleading and requires an additional unneeded check to see if the ids of the URL and the body are consistent. Therefore the id is removed from the IndexSetUpdateRequest and the one from the URL is used only.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
@@ -242,6 +242,10 @@ public class IndexSetsResource extends RestResource {
         final IndexSetConfig indexSet = indexSetService.get(id)
                 .orElseThrow(() -> new NotFoundException("Index set <" + id + "> does not exist"));
 
+        if (!indexSet.isWritable()) {
+            throw new ClientErrorException("Default index set must be writable.", Response.Status.CONFLICT);
+        }
+
         clusterConfigService.write(DefaultIndexSetConfig.create(indexSet.id()));
 
         final IndexSetConfig defaultIndexSet = indexSetService.getDefault();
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
```diff
@@ -211,17 +211,20 @@ public IndexSetSummary update(@ApiParam(name = "id", required = true)
...
-        }
 
         final IndexSetConfig oldConfig = indexSetService.get(id)
                 .orElseThrow(() -> new NotFoundException("Index set <" + id + "> not found"));
 
-        final IndexSetConfig savedObject = indexSetService.save(updateRequest.toIndexSetConfig(oldConfig));
         final IndexSetConfig defaultIndexSet = indexSetService.getDefault();
+        final boolean isDefaultSet = oldConfig.equals(defaultIndexSet);
+
+        if (isDefaultSet && !updateRequest.isWritable()) {
```
kroepke: This writable check here isn't enough to prevent readonly indices to become default, actually switching the current default index set is done in `setDefault` below, which is being used by the UI.

dennisoelkers:
Good points! Thanks much.

dennisoelkers:
Added requested changes.

PR #3340, opened 2017-01-11
Add migration to fix parameter value types for alert conditions

bernd:
At one point at the beginning of the 2.2 development the values for integer parameters have been stored as strings in MongoDB. This has been fixed but existing broken documents need to be fixed.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
@@ -1,0 +1,150 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.migrations;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.primitives.Ints;
+import com.mongodb.client.MongoCollection;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
```diff
@@ -0,0 +1,150 @@
...
+
+                    // No need to convert anything if the field does not exist or is already an integer
+                    if (fieldValue == null || fieldValue instanceof Integer) {
+                        continue;
+                    }
+
+                    if (!(fieldValue instanceof String)) {
+                        LOG.warn("Field <{}> in alert condition <{}> of stream <{}> is not a string but a <{}>, not trying to convert it!",
+                                field, alertConditionId, streamId, fieldValue.getClass().getCanonicalName());
+                        return;
```
edmundoa: I think in this case we should only log the situation and continue converting other fields. This situation may happen with alert conditions created before 2.0. See https://github.com/Graylog2/graylog2-server/commit/ea4aa95fb609b56dcde4adc56cf311edf4ff3c0e for more information.

graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
```diff
@@ -0,0 +1,150 @@
...
+                for (String field : intFields) {
+                    final Object fieldValue = parameters.get(field);
+
+                    // No need to convert anything if the field does not exist or is already an integer
+                    if (fieldValue == null || fieldValue instanceof Integer) {
+                        continue;
+                    }
+
+                    if (!(fieldValue instanceof String)) {
+                        LOG.warn("Field <{}> in alert condition <{}> of stream <{}> is not a string but a <{}>, not trying to convert it!",
```
edmundoa: Could we also add the alert condition title if available?

graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
```diff
@@ -0,0 +1,150 @@
...
+                                field, alertConditionId, streamId, fieldValue.getClass().getCanonicalName());
+                        return;
+                    }
+
+                    final String stringValue = parameters.get(field, String.class);
+                    final Integer intValue = Ints.tryParse(stringValue);
+
+                    LOG.info("Converting value for field <{}> from string to integer in alert condition <{}> of stream <{}>", field, alertConditionId, streamId);
```
edmundoa: I would also add the alert condition title here.

graylog2-server/src/main/java/org/graylog2/migrations/V20170110150100_FixAlertConditionsMigration.java
bernd: Oh, good catch! That should have been a `continue`.

bernd:
Addressed all comments. Thanks!

PR #3341, opened 2017-01-11
Index set upgrading docs

bernd:
Adjust the document structure in `UPGRADING` to match previous versions and add a paragraph about the index set related REST API changes for the streams endpoints.

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -1,6 +1,6 @@
-*****************
-Upgrading Graylog
-*****************
+**************************
+Upgrading to Graylog 2.2.x
+**************************
 
 .. _upgrade-from-21-to-22:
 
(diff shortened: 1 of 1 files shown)


UPGRADING.rst
```diff
@@ -66,3 +63,17 @@ Examples:
...
+
+Streams API
+-----------
+
+Due to the introduction of index sets, the payload for creating, updating and cloning of streams now requires the ``index_set_id`` field. The value for this needs to be the ID of an existing index set.
+
+Affected endpoints:
+
+* ``POST /streams``
+* ``PUT  /streams/{streamId}``
```
kroepke: In `UpdateStreamRequest` `index_set_id` is nullable, do we want to be overly specific here or not?
Or is that a bug because the other two require the field.

bernd: image

We are not sending the complete resource when updating a stream. This basically violates the PUT HTTP verb, but well. History. :smiley: 

That's why it needs to be nullable.

PR #3349, opened 2017-01-13
Allow re-configuration of shards and replicas in the UI

bernd:
This should be safe because it only affects new indices and doesn't touch existing ones.


joschi:
For the same reason we could allow changing the analyzer (currently readonly).

PR #3377, opened 2017-01-17
do not allow testing a message against the default stream

### LINKED ISSUE #3369: Testing message against default stream returns erroneous information
From the search results, it is possible to test a message against the default stream. That renders the edit stream rules page with the selected message, but the result of testing the message against the stream seems to always be negative.

screen shot 2017-01-17 at 12 21 31

Please also note that the controls to add new stream rules are displayed. I tried adding a rule to the default stream and that returned a 400 response from the server, so it seems that it's only a display issue. 

## Expected Behaviour
Testing a message against the default stream should return true, or not be allowed at all, as it would always match.

## Current Behaviour
Testing a message against the default stream returns false, and it displays controls to add new stream rules.

1. Go to the search results
2. Expand a message in the results table
3. Select test against stream, and then choose the default stream
4. See that the box down the page is red (as when it fails). Also notice that there are controls to add a new stream rule

## Your Environment

* Graylog Version: Graylog 2.2.0-rc.1-SNAPSHOT (0308a94)

kroepke:
* the default stream is disabled in the menu and has a title explaining that it cannot be used (as title for hover)
 * going to `/streams/000000000000000000000001/edit` now displays an error message (this page should not be reachable other than manually via that path)

fixes #3369

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageDetail.jsx
+++ b/graylog2-web-interface/src/components/search/MessageDetail.jsx
@@ -99,7 +99,7 @@ const MessageDetail = React.createClass({
       }
       if (stream.is_default) {
         streamList.push(
-          <MenuItem key={stream.id} disabled={true} title="Cannot test against the default stream">{stream.title}</MenuItem>
+          <MenuItem key={stream.id} disabled title="Cannot test against the default stream">{stream.title}</MenuItem>
         );
       } else {
         streamList.push(
--- a/graylog2-web-interface/src/components/search/MessageDetail.jsx
+++ b/graylog2-web-interface/src/components/search/MessageDetail.jsx
@@ -97,12 +97,19 @@ const MessageDetail = React.createClass({
       if (!streamList) {
         streamList = [];
       }
-      streamList.push(
-        <LinkContainer key={stream.id}
-                       to={Routes.stream_edit_example(stream.id, this.props.message.index, this.props.message.id)}>
-          <MenuItem>{stream.title}</MenuItem>
-        </LinkContainer>
-      );
+      if (stream.is_default) {
+        streamList.push(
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/search/MessageDetail.jsx
```diff
@@ -97,12 +97,19 @@ const MessageDetail = React.createClass({
...
       }
-      streamList.push(
-        <LinkContainer key={stream.id}
-                       to={Routes.stream_edit_example(stream.id, this.props.message.index, this.props.message.id)}>
-          <MenuItem>{stream.title}</MenuItem>
-        </LinkContainer>
-      );
+      if (stream.is_default) {
+        streamList.push(
+          <MenuItem key={stream.id} disabled={true} title="Cannot test against the default stream">{stream.title}</MenuItem>
```
edmundoa: Minor nitpick: this breaks the [`react/jsx-boolean-value`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) style rule. You can just type `disabled` and the value passed to the component will be true, no need to assign the value

graylog2-web-interface/src/pages/StreamEditPage.jsx
```diff
@@ -30,6 +31,19 @@ const StreamEditPage = React.createClass({
       return <Spinner />;
     }
 
+    let content = <StreamRulesEditor currentUser={this.state.currentUser} streamId={this.props.params.streamId}
```
edmundoa: Minor nitpick: this breaks the [`react/wrap-multilines`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-wrap-multilines.md) style rule.

kroepke:
double linting

PR #3378, opened 2017-01-17
Unregistering component callbacks from Stream(Rules)Store.

### LINKED ISSUE #3374: Error thrown after editing stream rules
Since #3352, I'm seeing this error thrown in the JS console when editing the stream rules or changing the stream rules condition type:

```
bluebird.js:1542 Unhandled rejection TypeError: Cannot read property 'filterData' of undefined
    at eval (webpack:///./src/components/streams/StreamComponent.jsx?:61:31)
    at eval (webpack:///./src/stores/streams/StreamsStore.ts?:22:13)
From previous event:
    at StreamsStore.load (webpack:///./src/stores/streams/StreamsStore.ts?:21:28)
    at StreamComponent.loadData (webpack:///./src/components/streams/StreamComponent.jsx?:56:18)
    at StreamComponent.loadData (webpack:///./~/react-proxy/modules/createPrototypeProxy.js?:44:30)
    at eval (webpack:///./src/stores/streams/StreamsStore.ts?:105:20)
    at Array.forEach (native)
    at StreamsStore._emitChange (webpack:///./src/stores/streams/StreamsStore.ts?:104:24)
From previous event:
    at StreamsStore.update (webpack:///./src/stores/streams/StreamsStore.ts?:73:62)
    at MatchingTypeSwitcher.handleTypeChange (webpack:///./src/components/streams/MatchingTypeSwitcher.jsx?:74:22)
    at MatchingTypeSwitcher.handleTypeChange (webpack:///./~/react-proxy/modules/createPrototypeProxy.js?:44:30)
    at MatchingTypeSwitcher.handleTypeChangeToOr (webpack:///./src/components/streams/MatchingTypeSwitcher.jsx?:66:12)
    at MatchingTypeSwitcher.handleTypeChangeToOr (webpack:///./~/react-proxy/modules/createPrototypeProxy.js?:44:30)
    at Object.executeOnChange (http://localhost:8080/build/vendor.js:11911:35)
    at ReactDOMComponent._handleChange (http://localhost:8080/build/vendor.js:43712:39)
    at Object.ReactErrorUtils.invokeGuardedCallback (http://localhost:8080/build/vendor.js:28148:17)
    at executeDispatch (http://localhost:8080/build/vendor.js:40974:22)
    at Object.executeDispatchesInOrder (http://localhost:8080/build/vendor.js:40997:6)
    at executeDispatchesAndRelease (http://localhost:8080/build/vendor.js:8509:23)
    at executeDispatchesAndReleaseTopLevel (http://localhost:8080/build/vendor.js:8520:11)
    at Array.forEach (native)
    at forEachAccumulated (http://localhost:8080/build/vendor.js:29115:10)
    at Object.processEventQueue (http://localhost:8080/build/vendor.js:8725:8)
    at runEventQueueInBatch (http://localhost:8080/build/vendor.js:44693:19)
    at Object.handleTopLevel [as _handleTopLevel] (http://localhost:8080/build/vendor.js:44709:6)
    at handleTopLevelWithoutPath (http://localhost:8080/build/vendor.js:44811:25)
    at handleTopLevelImpl (http://localhost:8080/build/vendor.js:44791:4)
    at ReactDefaultBatchingStrategyTransaction.perform (http://localhost:8080/build/vendor.js:10097:21)
    at Object.batchedUpdates (http://localhost:8080/build/vendor.js:27564:20)
    at Object.batchedUpdates (http://localhost:8080/build/vendor.js:6232:21)
    at dispatchEvent (http://localhost:8080/build/vendor.js:44922:21)
```

I haven't noticed any errors after the exception is raised, but trying to access a ref that is not mounted should be avoided anyway.

1. Go to the streams page
2. Create a stream and click on "Manage rules" (or do it with an existing stream)
3. Add a rule and/or change the condition type for the stream
4. See the error thrown in your console logs

## Your Environment

* Graylog Version: Graylog 2.2.0-rc.1-SNAPSHOT (0308a94)

dennisoelkers:
## Motivation and Context

Before this change, the StreamComponent was registering callbacks to be
notified when the Stream(Rules)Store state changes. Unfortunately, there
was no way to unregister these callbacks, so after unmounting the
component its callbacks were still called.

After this change, there is now a way to unregister callbacks which is
consumed in the componentWillUnmount function of the StreamComponent.

Fixes #3374.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/streams/StreamRulesStore.ts
+++ b/graylog2-web-interface/src/stores/streams/StreamRulesStore.ts
@@ -3,6 +3,7 @@ import ApiRoutes = require('routing/ApiRoutes');
 const UserNotification = require('util/UserNotification');
 const URLUtils = require('util/URLUtils');
 const fetch = require('logic/rest/FetchProvider').default;
+const lodash = require('lodash');
 
 interface StreamRuleType {
   id: number;
@@ -74,6 +75,9 @@ class StreamRulesStore {
   }
   _emitChange() {
     this.callbacks.forEach((callback) => callback());
+  }
+  unregister(callback: Callback) {
+    lodash.pull(this.callbacks, callback);
   }
 }
 
--- a/graylog2-web-interface/src/stores/streams/StreamsStore.ts
+++ b/graylog2-web-interface/src/stores/streams/StreamsStore.ts
@@ -2,6 +2,7 @@ const UserNotification = require('util/UserNotification');
 const URLUtils = require('util/URLUtils');
 import ApiRoutes = require('routing/ApiRoutes');
 const fetch = require('logic/rest/FetchProvider').default;
+const lodash = require('lodash');
 
 interface Stream {
   id: string;
@@ -148,6 +149,9 @@ class StreamsStore {
   }
   _emitChange() {
     this.callbacks.forEach((callback) => callback());
+  }
+  unregister(callback: Callback) {
+    lodash.pull(this.callbacks, callback);
   }
 }
 
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -30,6 +30,7 @@
     "immutable": "^3.7.5",
     "javascript-natural-sort": "^0.7.1",
     "jquery-ui": "1.12.x",
+    "lodash": "^4.17.4",
     "markdown": "^0.5.0",
     "md5": "^2.0.0",
     "moment-duration-format": "1.3.0",
--- a/graylog2-web-interface/src/components/streamrules/StreamRulesEditor.jsx
+++ b/graylog2-web-interface/src/components/streamrules/StreamRulesEditor.jsx
@@ -31,6 +31,10 @@ const StreamRulesEditor = React.createClass({
     StreamsStore.onChange(this.loadData);
     StreamRulesStore.onChange(this.loadData);
   },
+  componentWillUnmount() {
+    StreamsStore.unregister(this.loadData);
+    StreamRulesStore.unregister(this.loadData);
+  },
   onMessageLoaded(message) {
     this.setState({ message: message });
     if (message !== undefined) {
--- a/graylog2-web-interface/src/components/streams/StreamComponent.jsx
+++ b/graylog2-web-interface/src/components/streams/StreamComponent.jsx
@@ -30,6 +30,10 @@ const StreamComponent = React.createClass({
     StreamsStore.onChange(this.loadData);
     StreamRulesStore.onChange(this.loadData);
   },
+  componentWillUnmount() {
+    StreamsStore.unregister(this.loadData);
+    StreamRulesStore.unregister(this.loadData);
+  },
 
   loadData() {
     StreamsStore.load((streams) => {


graylog2-web-interface/src/stores/streams/StreamRulesStore.ts
```diff
@@ -75,6 +76,9 @@ class StreamRulesStore {
   _emitChange() {
     this.callbacks.forEach((callback) => callback());
   }
+  unregister(callback: Callback) {
+    lodash.pull(this.callbacks, callback);
```
edmundoa: I know we use lodash in one or two places already, but I honestly see no benefit extending its use for things like this. IMHO using `filter` from the standard library is simple enough, does not require anybody to learn anything else, and does not require the external dependency.

graylog2-web-interface/src/stores/streams/StreamsStore.ts
```diff
@@ -149,6 +150,9 @@ class StreamsStore {
   _emitChange() {
     this.callbacks.forEach((callback) => callback());
   }
+  unregister(callback: Callback) {
+    lodash.pull(this.callbacks, callback);
```
edmundoa: Same comment concerning lodash as before.

graylog2-web-interface/src/stores/streams/StreamRulesStore.ts
dennisoelkers: Using `filter` it would be `this.callbacks = this.callbacks.filter((c) => c !== callback)` which is harder to understand. As lodash also offers a lot of other useful functions that are missing in the javascript standard library, I think it's worth the (maximum) 24kb of overhead.

edmundoa: I don't doubt `lodash` is useful, but so far we only use a single function of it. As far as I know, nobody in the team knows it and we don't even have it as a direct dependency.
I'm not against adding 24kb of dependency if we need it (we already have many others, so there's not much difference), but in this case I have the feeling we are using it for some small eye candy.

dennisoelkers: The fact that it is not an explicit dependency yet is a problem, agreed. We use it in another [place](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/stores/alertconditions/AlertConditionsStore.js#L77) already, so this change is not introducing it in the first place and I think both future and existing code could benefit from it, once we are using it. Documentation is excellent too, so there is no real barrier for people in the team not knowing it yet.

edmundoa: Then please introduce the explicit dependency.

To be honest I rather stick to the standard library whenever possible, specially when there is an equivalent for the utility functions introduced in other libraries like lodash or jQuery, but we don't have a rule for that.

In my opinion the addition of lodash to the toolchain is bearable for the current team, but we will face challenges (and to be fair not just caused by lodash) when new team members arrive. The amount of technologies to learn in order to write some frontend code is really huge. We are not alone at this, as many node and JS projects have the same issue, but I think without a careful selection of the libraries we use, the problem will be even worse.

dennisoelkers: I really appreciate the thought you are putting into this, thank you. I have a differing opinion towards lodash though, because it's not like react/reflux something you need to learn to be able to write code, but it is something you can utilize when you see the need for it. So it does not raise the barrier for new team members to be productive. Instead, it allows to write more concise and succinct code when you are ready.

Long story short, I am gonna add the explicit dependency and I hope you are alright with that.

PR #3385, opened 2017-01-18
Do not show the "set as default" action for non-writable index sets

### LINKED ISSUE #3383: Disable "set as default" option for default index set
There is an option in the index sets page to set set any index set as default. That option should be disabled for the default index set, avoiding any possible confusions.
screen shot 2017-01-18 at 12 20 17

## Expected Behaviour
The option to set an index set as default is disabled on the default index set.

## Current Behaviour
The option to set an index set as default is enabled for all index sets, including the current default index set.

## Your Environment
* Graylog Version: Graylog 2.2.0-rc.1-SNAPSHOT (7bb61da)

bernd:
Fixes #3383


graylog2-web-interface/src/components/indices/IndexSetsComponent.jsx
```diff
@@ -60,18 +60,26 @@ const IndexSetsComponent = React.createClass({
   },
 
   _formatIndexSet(indexSet) {
+    const menuItems = [];
+
+    if (indexSet.writable) {
+      menuItems.push(<MenuItem onSelect={this._onSetDefault(indexSet)} disabled={!indexSet.writable}>Set as default</MenuItem>);
```
edmundoa: With this change the menu item would never be disabled, this line of code would not be reached for non-writable index sets.

By the way, I like better the idea of having some `title` saying why the item is disabled instead of just removing it from the menu. It would make it more clear imho.

bernd: I didn't read the issue careful enough and removed the menu item when not writable. I will change it to only disable it in that case.

PR #3404, opened 2017-01-20
Only show extractor actions on string fields

### LINKED ISSUE #3396: The message field "tags" cannot be extracted from
## Expected Behavior
I tried to create an extractor for the "tags" field of a message. I wanted to extract specific part of a FQDN (which was set as one of the tags for the sidecar-collectors) either via "Split and Index" or regex matching. I expected it to extract from the previewed string of the field (all tags concatinated without commas, i.e. tag1tag2tag3) in the extractor dialog.

## Current Behavior
Extraction on the "tags" field does not work with either "Split and Index" or regex matching, I haven't tried the others.
The front-end gives a generic error about the regex checker, my browser's web-console says:
```
"There was an error fetching a resource: cannot POST
https://$grayloghost/api/tools/split_and_index_tester (400)."
Additional information: Can not deserialize instance of java.lang.String out
of START_ARRAY token
 at [Source: org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseab
+leInputStream@45787142;
line: 1, column: 36] (through reference chain:
org.graylog2.rest.models.tools.requests.SplitAndIndexTestRequest["string"])
```

## Possible Solution
It looks like the "tags" field is actually an array and not a string, so it should be converted to one.

1. Add a graylog client via the sidecar collector, either leave the default tags or add your own
2. Look at any message from that host and find the "tags" field
3. Click on the downward triangle button on the right and create an extractor for the "tags" field (regex or "Split and Index")
4. Enter any values and click either "Try" or "Save"

## Context
I actually tried to get the sending host to include its FQDN with every message in a really hackish way, now I use a different workaround. Still I think this should work (at least I expected it to).

## Your Environment

* Graylog Version: 2.1.2
* Operating System: Ubuntu 16.04
* Browser version: Firefox 50

joschi:
Extractors only work on strings, so the web interface should only show the extractor actions for message fields of type "string" and hide them for all other types.

Fixes #3396

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
+++ b/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
@@ -39,7 +39,19 @@ const MessageFieldExtractorActions = React.createClass({
           </div>
         );
     } else {
-      return (<div className="message-field-actions pull-right"/>);
+        return (
+            <div className="message-field-actions pull-right">
+              <DropdownButton pullRight
+                              bsSize="xsmall"
+                              title="Select extractor type"
+                              key={1}
+                              id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
+                <MenuItem key={`menu-item-disabled`} disabled={true}>
+                    Extractors can only be used with string fields.
+                </MenuItem>
+              </DropdownButton>
+            </div>
+        );
     }
   },
 });
--- a/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
+++ b/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
@@ -25,17 +25,22 @@ const MessageFieldExtractorActions = React.createClass({
     );
   },
   render() {
-    return (
-      <div className="message-field-actions pull-right">
-        <DropdownButton pullRight
-                     bsSize="xsmall"
-                     title="Select extractor type"
-                     key={1}
-                     id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
-          {ExtractorUtils.EXTRACTOR_TYPES.map(extractorType => this._formatExtractorMenuItem(extractorType))}
-        </DropdownButton>
-      </div>
-    );
+    let messageField = this.props.message.fields[this.props.fieldName];
+    if(typeof messageField === 'string') {
+      return (
+          <div className="message-field-actions pull-right">
+            <DropdownButton pullRight
+                            bsSize="xsmall"
+                            title="Select extractor type"
+                            key={1}
+                            id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
+                {ExtractorUtils.EXTRACTOR_TYPES.map(extractorType => this._formatExtractorMenuItem(extractorType))}
+            </DropdownButton>
+          </div>
+        );
+    } else {
+      return (<div className="message-field-actions pull-right"/>);
+    }
   },
 });
 
--- a/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
+++ b/graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
@@ -25,7 +25,7 @@ const MessageFieldExtractorActions = React.createClass({
     );
   },
   render() {
-    let messageField = this.props.message.fields[this.props.fieldName];
+    const messageField = this.props.message.fields[this.props.fieldName];
     if(typeof messageField === 'string') {
       return (
           <div className="message-field-actions pull-right">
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
```diff
@@ -25,17 +25,22 @@ const MessageFieldExtractorActions = React.createClass({
...
+                            bsSize="xsmall"
+                            title="Select extractor type"
+                            key={1}
+                            id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
+                {ExtractorUtils.EXTRACTOR_TYPES.map(extractorType => this._formatExtractorMenuItem(extractorType))}
+            </DropdownButton>
+          </div>
+        );
+    } else {
+      return (<div className="message-field-actions pull-right"/>);
```
edmundoa: We should display something in this case, I would keep a disabled drop-down with a message saying that extractors cannot be applied to non-string fields.

edmundoa: For example, something like we do in here: https://github.com/Graylog2/graylog2-server/blob/66889b478a7f4113d167df85c3e2a095061c3181/graylog2-web-interface/src/components/streams/Stream.jsx#L111-L117

joschi: ✅ Personally, I find it more obvious if the dropdown list was simply missing, but for the sake of consistency I've added back the dropdown and disabled it for non-string fields.

edmundoa: 👍 In this case I think it makes a big difference if you know the system well or not. In our case, we know why the dropdown is not shown, so it may actually save us time, as we know which fields are not strings very easily. On the other hand, for somebody just starting to use Graylog, they would be confused because some fields are missing the dropdown for non-obvious reasons. So far we have (mostly) aligned with the second "persona".

graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
```diff
@@ -25,17 +25,34 @@ const MessageFieldExtractorActions = React.createClass({
...
-        <DropdownButton pullRight
-                     bsSize="xsmall"
-                     title="Select extractor type"
-                     key={1}
-                     id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
-          {ExtractorUtils.EXTRACTOR_TYPES.map(extractorType => this._formatExtractorMenuItem(extractorType))}
-        </DropdownButton>
-      </div>
-    );
+    let messageField = this.props.message.fields[this.props.fieldName];
```
edmundoa: Small nitpick: can we use `const` instead of `let` in here? The variable is never modified afterwards.

graylog2-web-interface/src/components/search/MessageFieldSearchActions.jsx
```diff
@@ -23,6 +23,18 @@ const MessageFieldSearchActions = React.createClass({
     );
   },
   render() {
+    let messageField = this.props.message.fields[this.props.fieldName];
```
edmundoa: Small nitpick: can we use `const` instead of `let` in here? The variable is never modified afterwards.

graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
```diff
@@ -25,17 +25,34 @@ const MessageFieldExtractorActions = React.createClass({
...
+        );
+    } else {
+        return (
+            <div className="message-field-actions pull-right">
+              <DropdownButton pullRight
+                              bsSize="xsmall"
+                              title="Select extractor type"
+                              key={1}
+                              id={`select-extractor-type-dropdown-field-${this.props.fieldName}`}>
+                <MenuItem key={`menu-item-disabled`} disabled={true}>
```
edmundoa: Here another couple of nitpicks: 

- `key` can be a regular string
- We don't need to pass the boolean value to the `disabled` flag (it breaks the [react/jsx-boolean-value](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) rule)

graylog2-web-interface/src/components/search/MessageFieldSearchActions.jsx
```diff
@@ -23,6 +23,18 @@ const MessageFieldSearchActions = React.createClass({
...
+    let extractors;
+    if(typeof messageField === 'string') {
+      extractors = (<li className="dropdown-submenu left-submenu">
+        <a href="#">Create extractor for field {this.props.fieldName}</a>
+        <ul className="dropdown-menu">
+            {ExtractorUtils.EXTRACTOR_TYPES.map(extractorType => this._formatExtractorMenuItem(extractorType))}
+        </ul>
+      </li>);
+    } else {
+        extractors = (<MenuItem disabled={true}>Extractors can only be used with string fields</MenuItem>);
```
edmundoa: Same as in the other file: 

- `key` can be a regular string
- We don't need to pass the boolean value to the `disabled` flag (it breaks the [react/jsx-boolean-value](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) rule)

PR #3440, opened 2017-01-27
Adding simple cache wrapper around index sets list in MongoIndexSetRegistry.

dennisoelkers:
## Motivation and Context

Before this change, every flush of the batched ES output results in MongoDB hits twice, because of status checks for all writable targets of all index sets.

This change adds a simple single-element cache in the `MongoIndexSetRegistry` class, which keeps 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
@@ -16,16 +16,24 @@
  */
 package org.graylog2.indexer;
 
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.eventbus.EventBus;
+import com.google.common.eventbus.Subscribe;
 import org.graylog2.indexer.indexset.IndexSetConfig;
 import org.graylog2.indexer.indexset.IndexSetService;
+import org.graylog2.indexer.indexset.events.IndexSetCreatedEvent;
+import org.graylog2.indexer.indexset.events.IndexSetDeletedEvent;
 import org.graylog2.indexer.indices.TooManyAliasesException;
 
 import javax.inject.Inject;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
```diff
@@ -16,32 +16,79 @@
...
+        @Inject
+        IndexSetsCache(IndexSetService indexSetService,
+                       EventBus serverEventBus) {
+            this.indexSetService = requireNonNull(indexSetService);
+            this.indexSetConfigs = new AtomicReference<>(Suppliers.memoize(this.indexSetService::findAll));
+            serverEventBus.register(this);
+        }
+
+        List<IndexSetConfig> get() {
+            return indexSetConfigs.get().get();
```
joschi: Maybe returning a copy of the list would be useful here.

graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
```diff
@@ -16,32 +16,79 @@
...
 import static java.util.Objects.requireNonNull;
 
+@Singleton
 public class MongoIndexSetRegistry implements IndexSetRegistry {
     private final IndexSetService indexSetService;
     private final MongoIndexSet.Factory mongoIndexSetFactory;
 
+    static class IndexSetsCache {
+        private final IndexSetService indexSetService;
+        private AtomicReference<Supplier<List<IndexSetConfig>>> indexSetConfigs;
```
dennisoelkers: Using the supplier guarantees it is refreshed lazily, to prevent every node hammering the collection after a change event.

graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSetRegistry.java
joschi: Good point. 👍

PR #3456, opened 2017-02-02
Cleanup web assets when calling mvn clean

bernd:
Fixes the problem that old web assets will be included in the JAR files.

### DIFF
--- a/graylog-plugin-parent/pom.xml
+++ b/graylog-plugin-parent/pom.xml
@@ -98,6 +98,22 @@
 
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


bernd:
@dennisoelkers This only does it for the server repo but not for plugins. Also, when you look at the `<directory>` configuration in the code you highlighted, this will not work for plugins at the moment.

We can of course move it to the graylog-parent pom.xml but then we need some special handling for the graylog2-web-interface directory. This PR fixes the problem for all plugins and leaves the special case for the server repo.

I think this tiny duplication is fine.

dennisoelkers:
Ah, got it. The initial description was a bit misleading.

PR #3487, opened 2017-02-13
Remove deprecated EsIndexRange and EsIndexRangeService

joschi:


dennisoelkers:
Would it make sense to rename `LegacyMongoIndexRangeService` to `MongoIndexRangeService` as it does not seem to be legacy only anymore?

joschi:
@dennisoelkers There are `org.graylog2.indexer.ranges.LegacyMongoIndexRangeService` and `org.graylog2.indexer.ranges.MongoIndexRangeService` which are not the same.

PR #3503, opened 2017-02-15
Allow version '0' for structured syslog messages

### LINKED ISSUE #3502: Graylog incorrectly processes Syslog messages with incorrect version
Since the changes for #2954, Graylog doesn't process syslog messages with an incorrect version (i. e. not `1`) anymore.

## Expected Behavior

The example message should be parsed correctly:
```
<12>0 2017-02-15T16:01:07.614260+01:00 vmlsfw test - - -  test 4
```

## Current Behavior

The example message is parsed incorrectly:
```
<12>0 2017-02-15T16:01:07.614260+01:00 vmlsfw test - - -  test 4
```

From a unit test to reproduce the behavior:
```
0 = {HashMap$Node@3116} "level" -> "4"
1 = {HashMap$Node@3117} "_id" -> "98e03610-f39b-11e6-827e-6a9c47cfab13"
2 = {HashMap$Node@3118} "source" -> "2017-02-15T16:01:07.614260+01:00"
3 = {HashMap$Node@3119} "message" -> "2017-02-15T16:01:07.614260+01:00 vmlsfw test - - -  test 4"
4 = {HashMap$Node@3120} "facility" -> "user-level"
5 = {HashMap$Node@3121} "timestamp" -> "0000-01-01T00:00:00.000+00:53:28"
```

## Possible Solution

Make parsing the syslog version more lenient (e. g. use `[0-9]` instead of `[1-9]`).

Also, using the predefined `RSYSLOG_SyslogProtocol23Format` template should work: https://github.com/rsyslog/rsyslog/blob/v7.6.7/runtime/rsconf.c#L87

## Context

https://github.com/rsyslog/rsyslog/blob/v8.24.0/runtime/msg.h#L148-L150

## Your Environment

* Graylog Version: 2.2.0
* rsyslog 7.6.7

joschi:
Although RFC 5424 clearly states that '1' is the only valid version for structured syslog
messages, some syslog daemons are using '0' as a version identifier.

https://tools.ietf.org/html/rfc5424#section-9.1

Fixes #3502

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/SyslogCodec.java
@@ -59,7 +59,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 @Codec(name = "syslog", displayName = "Syslog")
 public class SyslogCodec extends AbstractCodec {
     private static final Logger LOG = LoggerFactory.getLogger(SyslogCodec.class);
-    private static final Pattern STRUCTURED_SYSLOG_PATTERN = Pattern.compile("<\\d{1,3}>[1-9]\\d{0,2}\\s.*", Pattern.DOTALL);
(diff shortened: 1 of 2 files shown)


joschi:
This also has to be back-ported into `2.2`.

PR #3526, opened 2017-02-20
Replacing static font awesome with packaged version.

dennisoelkers:
This PR is replacing the statically included version of font awesome with a packaged one and removing the ancient assets.


graylog2-web-interface/src/routing/AppFacade.jsx
```diff
@@ -12,7 +12,7 @@ const CurrentUserStore = StoreProvider.getStore('CurrentUser');
 
 import 'javascripts/shims/styles/shim.css';
 import 'stylesheets/bootstrap.min.css';
-import 'stylesheets/font-awesome.min.css';
+import 'font-awesome/css/font-awesome.css';
```
joschi: Shouldn't this be [`font-awesome/css/font-awesome.min.css`](https://github.com/FortAwesome/Font-Awesome/blob/v4.7.0/css/font-awesome.min.css)?

dennisoelkers: Minification should be done by the toolchain, namely webpack.

joschi: ✅

Issue #3511, opened 2017-02-16
alerts - make stateful notification an option


jalogisch:
## Expected Behavior
It should be possible for the user to enable the old notification behavior - because this new way of notification breaks the default from 2.X Version

## Current Behavior
If someone updates to the Version 2.2.0 he can't work with any kind of workflow or environment around Graylog that had used the alarming feature that returns every minute the alarm.

## Possible Solution
Make the stateful notification an option and or make it possible to return to the old behavior of getting a notification as long as the condition is found.

## Context
The new stateful notification is a breaking change.  

People used the old way of notification in other ways that we - the developers did not even think of. 

Taking https://github.com/Graylog2/graylog-support/issues/9 and this https://community.graylog.org/t/alert-conditions/80 we are facing two possible workflows that we did not see during development. 

Additional none of the Testers of Beta or RC Version has such a workflow and could had stopped us.

## Your Environment
* Graylog Version: 2.2.0

hezor:
This is good news. We too have the kind of alerts that won't work properly with the stateful behavior. Any chance this option could be included in 2.2.1?

gruselglatz:
It is also important for our Workflow, that we can bound one or more condition/s and notification/s to each other. Because we need different notifications on different events in the same stream. At the moment we have to create a new stream for every new condition/notification relationship.

dennisoelkers:
@gruselglatz: Is this related to stateful alerting or a different feature? I think what you mean is coupling alert notifications to alert conditions, is that correct?

gruselglatz:
@dennisoelkers yes, like the old behaviour

edmundoa:
@gruselglatz the old behaviour was not like that, alert conditions and notifications were bounded to a stream, just like they are in now.

gruselglatz:
@edmundoa ok sorry, but you know what my point is? i only want to link notifications to a certain condition, that i can create more use cases in one stream

edmundoa:
@gruselglatz yes, I know. That's one of the future improvements that may come to alerting, but so far we didn't have enough time to do it. If you want to discuss that, feel free to open another issue, let's keep this issue on stateful notification, please 🙂

dennisoelkers:
Fixed in #3535 & #3544.

PR #3536, opened 2017-02-22
Give an option to repeat alert notifications

edmundoa:
This PR fixes the issue described in #3511 by adding an option to the alert conditions parameters to repeat notifications.

The `repeat_notifications` option is disabled by default, so all alerts will *only* send a notification when a new alert is triggered, the behaviour introduced in 2.2.0.

When users rely on repeated notifications on their alerting, they can enable `repeat_notifications` and Graylog will send a notification after each positive evaluation of the alert condition, just as we did in previous versions.

Alerts will continue having state in the UI, and users can see all sent notifications in the alert details page.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AlertNotificationsSender.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AlertNotificationsSender.java
@@ -16,7 +16,6 @@
  */
 package org.graylog2.alerts;
 
-import com.google.inject.Inject;
 import org.graylog2.alarmcallbacks.AlarmCallbackConfiguration;
 import org.graylog2.alarmcallbacks.AlarmCallbackConfigurationService;
 import org.graylog2.alarmcallbacks.AlarmCallbackFactory;
@@ -28,6 +27,7 @@ import org.graylog2.plugin.streams.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.inject.Inject;
 import java.util.List;
 
 public class AlertNotificationsSender {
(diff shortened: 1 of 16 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AlertNotificationsSender.java
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.alerts;
+
+import com.google.inject.Inject;
```
dennisoelkers: Minor nitpick, but please use `javax.inject.Inject` instead. `com.google.inject.Inject` is guice-specific, we should try to avoid that.

graylog2-web-interface/src/components/alarmcallbacks/AlarmCallbackHistoryOverview.jsx
```diff
@@ -30,7 +30,9 @@ const AlarmCallbackHistoryOverview = React.createClass({
       return <Spinner />;
     }
 
-    const histories = this.state.histories.map(this._formatHistory);
+    const histories = this.state.histories
+      .sort((h1, h2) => h1.created_at.localeCompare(h2.created_at))
```
dennisoelkers: Comparing two timestamps lexicographically might not lead to a correct result. We should use `moment().isBefore()/isAfter()` for that.

edmundoa: ✅

graylog2-server/src/main/java/org/graylog2/alerts/AlertNotificationsSender.java
edmundoa: Damn auto imports... ✅

PR #3537, opened 2017-02-22
Wait until notification types are loaded

### LINKED ISSUE #3534: Error loading alert condition details page
Sometimes loading the alert condition details page only works partially, and the alert notifications associated to the stream are always in a loading state. The error in the console is:
```
Unhandled rejection TypeError: Cannot read property 'org.graylog2.alarmcallbacks.HTTPAlarmCallback' of undefined
    at Constructor.render (webpack:///./src/components/alertnotifications/AlertNotification.jsx?:92:59)
```

## Expected Behavior
Alert notifications should load.

## Current Behavior
Alert notifications do not always load properly.

1. Go to Alerts -> Alert conditions -> Select an alert
2. Check if the notifications appear (this is a timing issue, so it may not always happen)

## Your Environment
* Graylog Version: Graylog 2.3.0-SNAPSHOT, 2.2.1 is also affected

edmundoa:
In order to easily reproduce the issue, I manually added a `Thread.sleep(1000)` in `org.graylog2.rest.resources.alarmcallbacks.AlarmCallbacksResource#available`.

Fixes #3534 

This needs to be merged into the 2.2 branch as well.

### DIFF
--- a/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
+++ b/graylog2-web-interface/src/components/alertnotifications/AlertNotification.jsx
@@ -9,7 +9,7 @@ const { AlertNotificationsStore } = CombinedProvider.get('AlertNotifications');
 const { AlarmCallbacksActions } = CombinedProvider.get('AlarmCallbacks');
 const { CurrentUserStore } = CombinedProvider.get('CurrentUser');
 
-import { EntityListItem } from 'components/common';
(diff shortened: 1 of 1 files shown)


edmundoa:
I moved the check to the component itself. The reason to put it in `ConditionAlertNotifications` before was because that component is the one requesting the data, and it feels more natural to me to group those two actions that are closely related.

Issue #2459, opened 2016-07-06
reverse quick values/Bottom-N


jalogisch:
### Problem description

if you look into your data with quick values ( http://docs.graylog.org/en/2.0/pages/queries.html#quick-values ) you only get the top values.

it would be very useful to get the bottom as an option ( lowest count in given time ) and output them as list.

kroepke:
Essentially we need to expose the order parameter of the terms aggregation to support ascending and descending sorting.

GambitK:
Yes, ordering ascending and descending would also help in searching for rare occurrences of events.

lennartkoopmann:
This will be handled in #4082,

PR #3540, opened 2017-02-23
Update Terms Method to Support Ascending and Descending Ordering

billmurrin:
Add sort order to the terms method (org.graylog2.indexer.searches.Searches) in order to provide support for ascending and descending Terms aggregation results. Relying on widgets to conduct the ascending order does not provide a "true bottom" value, it only provides the most ascending result based on a descending pull; if the possible results are greater than the size of the pull (e.g. 50) this will result in an inaccurate and misleading result.

## Description
The changes provide a sorting variable to the terms method. Subsequent functions for the method were modified to pass a default value of "descending" into the method in order to assure backwards compatibility. The changes passed the "testTerms" Test.

## Motivation and Context
This change is required to provide true ascending, or bottom, values for term based aggregations.
Relates to Issue #2459

## How Has This Been Tested?
- The same changes are employed in my plugin, the Quick Values Plus Widget, which provides ascending and descending ordering options.
- The testTerms test (based on the same method) successfully passed.
- The changes should have no impact as methods were created to ensure backwards compatibility

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -263,9 +263,17 @@ public class Searches {
         return new SearchResult(r.getHits(), indices, config.query(), request.source(), r.getTook());
     }
 
-    public TermsResult terms(String field, int size, String query, String filter, TimeRange range) {
+    public TermsResult terms(String field, int size, String query, String filter, TimeRange range, String sorting) {
+        Terms.Order termsOrder;
+
         if (size == 0) {
             size = 50;
+        }
+
+        if (sorting.equals("descending")){
+            termsOrder = Terms.Order.count(false);
+        } else {
+            termsOrder = Terms.Order.count(true);
         }
 
         SearchRequestBuilder srb;
@@ -279,7 +287,8 @@ public class Searches {
                 .subAggregation(
                         AggregationBuilders.terms(AGG_TERMS)
                                 .field(field)
-                                .size(size))
+                                .size(size)
+                                .order(termsOrder))
                 .subAggregation(
                         AggregationBuilders.missing("missing")
                                 .field(field))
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -263,11 +263,19 @@ public SearchResult search(SearchesConfig config) {
         return new SearchResult(r.getHits(), indices, config.query(), request.source(), r.getTook());
     }
 
-    public TermsResult terms(String field, int size, String query, String filter, TimeRange range) {
+    public TermsResult terms(String field, int size, String query, String filter, TimeRange range, String sorting) {
```
joschi: Please don't use a string for `sorting` here but [`Sorting.Direction`](https://github.com/Graylog2/graylog2-server/blob/2.2.1/graylog2-server/src/main/java/org/graylog2/indexer/searches/Sorting.java#L30-L33) instead.

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -263,11 +263,19 @@ public SearchResult search(SearchesConfig config) {
...
-    public TermsResult terms(String field, int size, String query, String filter, TimeRange range) {
+    public TermsResult terms(String field, int size, String query, String filter, TimeRange range, Sorting.Direction sorting) {
+        Terms.Order termsOrder;
+
         if (size == 0) {
             size = 50;
         }
 
+        if (sorting == Sorting.Direction.DESC){
+            termsOrder = Terms.Order.count(false);
```
joschi: Why not [`Terms.Order#term(boolean)`](https://static.javadoc.io/org.elasticsearch/elasticsearch/2.4.4/org/elasticsearch/search/aggregations/bucket/terms/Terms.Order.html#term(boolean)) here?

billmurrin: Hi @joschi. I am using count because I am ordering basing it on the returned document count of the aggregation instead of just the term itself.

Per the docs: https://www.elastic.co/guide/en/elasticsearch/client/java-api/2.4/_bucket_aggregations.html#_order
```
Ordering the buckets by their doc_count in an ascending manner:

AggregationBuilders
    .terms("genders")
    .field("gender")
    .order(Terms.Order.count(true))
```

Issue #3483, opened 2017-02-13
Read-only index sets should not render "Active write index"


kroepke:
For example restored archives now live in a read only index set, and thus have no writable index.

The index page still tries to render it, making the text on the page look broken:

![image](https://cloud.githubusercontent.com/assets/52014/22884622/2b80c54a-f1f6-11e6-9d74-bd1c8a55701c.png)

billmurrin:
I created this branch to help solve this, but really have no way of testing it since I don't have archiving. 

Hope this helps. If you guys think so, I'll do a pull request. 

https://github.com/billmurrin/graylog2-server/tree/3483-active-write

kroepke:
@billmurrin definitely, please do so :+1:

PR #3547, opened 2017-02-26
Read-Only Index Set "Active write index" Rendering (#3483)

billmurrin:
Updates src/pages/IndexSetPage.jsx to take into account the use of read-only indexsets as reported in issue #3483 

## Description
Does a check to see if the indexset is writeable. If not, ensures that rendering output is modified from the default state.

## Motivation and Context
The change is required to fix a cosmetic bug report in #3483 

## How Has This Been Tested?
I have been unable to test the change as I do not have archiving available.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/IndexSetPage.jsx
+++ b/graylog2-web-interface/src/pages/IndexSetPage.jsx
@@ -132,12 +132,16 @@ const IndexSetPage = React.createClass({
     if (this.state.indexerOverview && this.state.indexDetails.closedIndices) {
       const deflectorInfo = this.state.indexerOverview.deflector;
 
+      let activeIndex = ".";
+      if (indexSet.writable) {
+        activeIndex = ", current write-active index is <i>{deflectorInfo.current_target}</i>";
+      }
+
       indicesInfo = (
         <span>
           <Alert bsStyle="success" style={{ marginTop: '10' }}>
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/pages/IndexSetPage.jsx
```diff
@@ -132,12 +132,16 @@ const IndexSetPage = React.createClass({
     if (this.state.indexerOverview && this.state.indexDetails.closedIndices) {
       const deflectorInfo = this.state.indexerOverview.deflector;
 
+      let activeIndex = ".";
```
kroepke: Please use single quotes for strings (our eslint config should flag this as an error).

graylog2-web-interface/src/pages/IndexSetPage.jsx
```diff
@@ -132,12 +132,16 @@ const IndexSetPage = React.createClass({
...
+      if (indexSet.writable) {
+        activeIndex = ", current write-active index is <i>{deflectorInfo.current_target}</i>";
```
kroepke: This renders as:
image

I would wrap the text in a `<span>` instead of making it a string. Note that a template string would not work either, because of the `<i>` tag. If the span is multiline please also wrap it in `()`.

PR #3549, opened 2017-02-27
Replacing static fonts with reference to npm package.

dennisoelkers:
This change is removing static font files and stylesheet and replaces it with a reference to an npm package containing the font.


graylog2-web-interface/src/routing/AppFacade.jsx
```diff
@@ -12,7 +12,7 @@ const CurrentUserStore = StoreProvider.getStore('CurrentUser');
 
 import 'bootstrap/dist/css/bootstrap.css';
 import 'font-awesome/css/font-awesome.css';
-import 'stylesheets/newfonts.less';
+import 'opensans-npm-webfont';
```
joschi: Is this correct? The other imports all reference a specific CSS (or LESS) file while this only references the directory.

dennisoelkers: The lookup mechanism in this case works like this: find out it is a module, check the module's directory for `package.json|index.js`, if the first one is found (like in this case) check for a `main` entry and use the file it references. Which is [`style.css` in this case](https://github.com/vaeum/opensans-npm-webfont/blob/master/package.json#L5).

joschi: ✅

PR #3564, opened 2017-03-03
Clean up usage of jmte Engine and use singleton instance

joschi:
Instead of instantiating the jmte template engine all over the place, this PR cleans up a bit and injects a compiling template engine as a singleton into the relevant classes.

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/alerts/FormattedEmailAlertSenderTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/alerts/FormattedEmailAlertSenderTest.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.alerts;
 
+import com.floreysoft.jmte.Engine;
 import org.graylog2.configuration.EmailConfiguration;
 import org.graylog2.notifications.NotificationService;
 import org.graylog2.plugin.Message;
@@ -49,10 +50,11 @@ public class FormattedEmailAlertSenderTest {
     private NodeId mockNodeId;
 
     private FormattedEmailAlertSender emailAlertSender;
+    private final Engine templateEngine = new Engine();
 
     @Before
     public void setUp() throws Exception {
-        this.emailAlertSender = new FormattedEmailAlertSender(new EmailConfiguration(), mockNotificationService, mockNodeId);
+        this.emailAlertSender = new FormattedEmailAlertSender(new EmailConfiguration(), mockNotificationService, mockNodeId, templateEngine);
     }
 
     @Test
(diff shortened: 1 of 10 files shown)


graylog2-server/src/test/java/org/graylog2/decorators/FormatStringDecoratorTest.java
```diff
@@ -34,13 +35,14 @@
...
 public class FormatStringDecoratorTest {
+    private final Engine templateEngine = Engine.createDefaultEngine();
 
     @Test
     public void testFormat() {
 
         final DecoratorImpl decorator = getDecoratorConfig("${field_a}: ${field_b}", "message", true);
 
-        final FormatStringDecorator formatStringDecorator = new FormatStringDecorator(decorator);
+        final FormatStringDecorator formatStringDecorator = new FormatStringDecorator(decorator, templateEngine);
```
dennisoelkers: Why is it not possible to mock out the engine instead of supplying a real instance?

graylog2-server/src/test/java/org/graylog2/alerts/FormattedEmailAlertSenderTest.java
```diff
@@ -49,10 +50,11 @@
     private NodeId mockNodeId;
 
     private FormattedEmailAlertSender emailAlertSender;
+    private final Engine templateEngine = new Engine();
 
     @Before
     public void setUp() throws Exception {
-        this.emailAlertSender = new FormattedEmailAlertSender(new EmailConfiguration(), mockNotificationService, mockNodeId);
+        this.emailAlertSender = new FormattedEmailAlertSender(new EmailConfiguration(), mockNotificationService, mockNodeId, templateEngine);
```
dennisoelkers: Why is it not possible to mock out the engine instead of supplying a real instance?

joschi: What advantage would using a mock over the real engine have here?

graylog2-server/src/test/java/org/graylog2/decorators/FormatStringDecoratorTest.java
joschi: What advantage would using a mock over the real engine have here?

dennisoelkers: That we are only testing this class and not an external dependency.

joschi: In which test case exactly?

FWIW, I'm trying to use real objects and not mocks whenever possible, also see https://martinfowler.com/articles/mocksArentStubs.html#ClassicalAndMockistTesting

dennisoelkers: ✅

graylog2-server/src/test/java/org/graylog2/alerts/FormattedEmailAlertSenderTest.java
dennisoelkers: ✅

PR #3574, opened 2017-03-06
Prevent unwanted session extension

### LINKED ISSUE #3565: (web) session should not be updated by ajax request
## Expected Behavior
When you have opened Graylog in your webbrowser and you leave the window for the defined session without any interaction the session should become invalid and the user need to login again - is forced to re-authenticate.

## Current Behavior
All reloads in the and updated stats refresh the timeout and the session will only expire after you had closed the Browser.

## Possible Solution
use [this header](https://github.com/Graylog2/graylog2-server/blob/e7c7127a93a73e3a59f0ce840b7afdbeaeeab9d4/graylog2-server/src/main/java/org/graylog2/security/realm/SessionAuthenticator.java#L41-L41) for all updates that should not trigger a reset of the timeout.

## Context
When you run Graylog with sensitive Information you might want to have a more strict session handling. Additional does the wording "session timeout" indicate that it work like described in the *expected behavior*

Additional this is needed to make Graylog PCI-DSS Standard compliant.

dennisoelkers:
## Motivation and Context

Without this change, there are some periodical calls to the backend which extend the user's session every call, so user session expiration never kicks in.

This change sets the `X-Graylog-No-Session-Extension` header to true for those calls, prevent unnecessary session extension. As this happens in several parts of the code, a helper method is create doing this, which can be easily used in the future when there are more periodical calls added.

Fixes #3565. Should also be merged in `2.2`.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/logic/rest/FetchProvider.js
+++ b/graylog2-web-interface/src/logic/rest/FetchProvider.js
@@ -121,9 +121,30 @@ export class Builder {
     return this;
   }
 
+  noSessionExtension() {
+    this.request = this.request.set('X-Graylog-No-Session-Extension', 'true');
+
+    return this;
+  }
+
   build() {
     return this.request;
   }
+}
+
+function queuePromiseIfNotLoggedin(promise) {
+  const SessionStore = StoreProvider.getStore('Session');
+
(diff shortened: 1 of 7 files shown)


joschi:
Rebased on current master.

PR #3575, opened 2017-03-06
React grid layout

edmundoa:
This PR replaces gridster.js with [React-Grid-Layout](https://github.com/STRML/react-grid-layout).

The main advantage is that React-Grid-Layout works more smoothly, with less surprises, and lets us do nicer things than gridster.js, like having responsive dashboards or fitting the widgets properly on the screen. In the development front, React-Grid-Layout is much easier to understand, and it integrates better with React and our current application.

This PR takes care of implementing all current functionalities using React-Grid-Layout. As a new addition, I made the number of columns available variable, and depending on the user's screen size. This will allow users with large screens to fit more widgets into the dashboard, while still being able to see them when using a smaller device.


STRML:
Excellent! I am happy to help if there are questions, bug reports, or features requests in the future for RGL. Glad to see it in Graylog!

PR #3576, opened 2017-03-06
Combine vendor and app webpack builds.

dennisoelkers:
## Motivation and Context

Starting with this change, we are building vendor & app (+ plugin if not disabled) bundles in one webpack run instead of two separate ones [as before](https://github.com/Graylog2/graylog2-server/blob/def68ba/graylog2-web-interface/package.json#L13-L15).

The benefits of this are:

  * faster turnaround times
  * allows changing of the vendor bundle without the need to restart everything
  * allows us to add a chunk hash to the vendor bundle filename, effectively busting the browser's cache in case of changes to the vendor bundle

This change does not require changes to plugin configs.


edmundoa:
Another issue I found while trying to get started with these changes: running the `webpack start-nohrm` task removes the web interface assets from the server, so the server needs to be rebuilt to be able to start after the npm task.

My guess is that the newly added step of cleaning the `build` folder is deleting the assets the server relies on to start the web interface.

### Steps to reproduce

1. Run `mvn compile` on the repo and start the server
1. Run `npm start-nohrm`
1. Restart the server, see that it fails to start with this error:

```
2017-03-28 12:38:34,438 WARN : org.glassfish.jersey.internal.Errors - The following warnings have been detected: WARNING: Unknown HK2 failure detected:
MultiException stack 1 of 1
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) Error injecting constructor, java.lang.IllegalStateException: Unable to find web interface assets. Maybe the web interface was not built into server?
  at org.graylog2.web.PluginAssets.<init>(PluginAssets.java:49)
  while locating org.graylog2.web.PluginAssets
    for the 1st parameter of org.graylog2.web.IndexHtmlGenerator.<init>(IndexHtmlGenerator.java:39)
  at org.graylog2.web.IndexHtmlGenerator.class(IndexHtmlGenerator.java:39)
  while locating org.graylog2.web.IndexHtmlGenerator
```

graylog2-web-interface/templates/index.html.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const vendorFiles = Object.keys(vendorModule.files.chunks).map(function(chunk) {
+    return "\t" + scriptTagForJsFile('/' + vendorModule.files.chunks[chunk].entry);
+  });
+  const jsFiles = Object.keys(htmlWebpackPlugin.files.chunks).map(function(chunk) {
+    return "\t" + scriptTagForJsFile(htmlWebpackPlugin.files.chunks[chunk].entry);
+  }).join("\n");
+
+  return '\
+  <!DOCTYPE html>\n\
```
dennisoelkers: Same here, it looks clumsy. The difference is that now we are looking up the name of the vendor bundle (which is part of a different compiler run, so it is not available before the actual template is rendered) dynamically (because it's now being a assigned a chunkhash as part of the name), therefore we need to use a JS template instead of a static one.

dennisoelkers:
I addressed your two remarks and validated the production build, works fine now.

graylog2-web-interface/templates/index.html.js
dennisoelkers: Thanks for pushing me on to this again. I had a simple idea which helped me employing the old EJS template again. (See aecc82d)

edmundoa: Cool! Looks much nicer. Thank you for giving it another try 😄

graylog2-server/src/main/resources/web-interface/index.html.template
```diff
@@ -12,7 +12,6 @@
   </head>
   <body>
     <script src="${appPrefix}/config.js"></script>
-    <script src="${appPrefix}/assets/vendor.js"></script>
```
edmundoa: This produces a slight inconsistency between the production and development templates. Before the vendor script would always load before the polyfill chunk. Now that is only true in development.

I don't think loading the polyfill first would break anything, but I would keep the same loading order in both environments for consistency's sake.

dennisoelkers:
Did you clean your cache before? it looks like it is somehow referring to an outdated vendor bundle?

graylog2-server/src/main/resources/web-interface/index.html.template
dennisoelkers: Good point!

edmundoa:
I can't remember, as it was 2 weeks ago, but I guess I did, along removing `node_modules` and retrying a couple of times. Anyway, one of the reasons for doing this PR was to fix those caching issues with the vendor bundle in the first place, not sure if we are missing something...

edmundoa:
I just tried clearing the browser cache and I can confirm the issue persists.

PR #3585, opened 2017-03-07
Using consistent collection of non displayable fields to filter against.

### LINKED ISSUE #3584: using message decorator hides "source" field
## Expected Behavior
using a decorator should only change the visibility of fields that might be decorated.

## Current Behavior
using a decorator hides the *source* field in the field overview

## Possible Solution

some fields are not reserved before the decorator jumps in

rework https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java#L61-L89

## Your Environment

* Graylog Version: 2.2.2

dennisoelkers:
## Motivation and Context

Before this change, the "source" field was suppressed from being
returned (and therefore) displayed in decorated results. This was caused
by some fields included in the `RESERVED_FIELDS` set, which were not
[added back](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/results/SearchResult.java#L79-L84) from the decorator processor before returning the list of fields included
in the [search result](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessorImpl.java#L95-L100).

This change could have extracted the method used in `SearchResult.java`
into a commonly used helper method used in both code points. This would
have involved a lot of casting/conversion between different
(`ResultMessage`/`ResultMessageSummary`) data types for each returned
message which is decorated, so the impact of deduplication would be
bigger than its benefits. Instead, I cleaned up and (hopefully) made the
sets of fields more expressive and easier to use.

Fixes #3584.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -68,49 +68,56 @@ public class Message implements Messages {
     private static final Pattern VALID_KEY_CHARS = Pattern.compile("^[\\w\\.\\-@]*$");
     private static final char KEY_REPLACEMENT_CHAR = '_';
 
-    public static final ImmutableSet<String> RESERVED_FIELDS = ImmutableSet.of(
-            // ElasticSearch fields.
-            FIELD_ID,
-            "_ttl",
-            "_source",
-            "_all",
-            "_index",
-            "_type",
-            "_score",
+    private static final ImmutableSet<String> GRAYLOG_FIELDS = ImmutableSet.of(
+        "gl2_source_node",
+        "gl2_source_input",
+        // TODO Due to be removed in Graylog 3.x
+        "gl2_source_radio",
+        "gl2_source_radio_input",
 
-            // Our reserved fields.
-            FIELD_MESSAGE,
-            FIELD_SOURCE,
-            FIELD_TIMESTAMP,
-            "gl2_source_node",
-            "gl2_source_input",
-            "gl2_source_collector",
-            "gl2_source_collector_input",
-            "gl2_remote_ip",
-            "gl2_remote_port",
-            "gl2_remote_hostname",
-            // TODO Due to be removed in Graylog 3.x
-            "gl2_source_radio",
-            "gl2_source_radio_input"
+        "gl2_source_collector",
+        "gl2_source_collector_input",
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -68,49 +68,56 @@
...
+    public static final ImmutableSet<String> RESERVED_FIELDS = new ImmutableSet.Builder<String>()
+        .addAll(RESERVED_SETTABLE_FIELDS)
+        .addAll(ES_FIELDS)
+        .build();
+
+    public static final ImmutableSet<String> NON_DISPLAYABLE_FIELDS = new ImmutableSet.Builder<String>()
+        .addAll(GRAYLOG_FIELDS)
+        .addAll(ES_FIELDS)
+        .add(FIELD_STREAMS)
+        .add(FIELD_FULL_MESSAGE)
```
joschi: Why is `full_message` in the set for "non-displayable" fields? Shouldn't it be possible to display it?

dennisoelkers: For some reasons it was [filtered out](https://github.com/Graylog2/graylog2-server/pull/3585/files#diff-635bdd3fdeee321a9c526fc484ed8b2aL86) for the filtered fields returned by the search, also not displayed in the sidebar.

joschi: What exactly do you mean with "not displayed in the sidebar"?

Graylog sidebar

dennisoelkers: I thought so, I guess the sidebar is using a different field set. Nonetheless, please see the code reference.

dennisoelkers: Maybe the naming is misleading and we should rename `NON_DISPLAYABLE_FIELDS` to `FILTERED_FIELDS` or something like that.

dennisoelkers: The reason why you are seeing the `full_message` field is because the result is decorated. This is actually part of the bug that is fixed in this PR.

PR #3589, opened 2017-03-08
Removing output to STDERR for default profile, it irritates IntelliJ.

dennisoelkers:
Before this change, IntelliJ interpreted the output to STDERR from
ESLint sourcing the webpack config as an error. This change removes the
output for the default profile to fix this.

### DIFF
--- a/graylog2-web-interface/webpack.config.js
+++ b/graylog2-web-interface/webpack.config.js
@@ -171,6 +171,5 @@ if (TARGET === 'test') {
 }
 
 if (Object.keys(module.exports).length === 0) {
-  console.error('Running in default mode');
   module.exports = webpackConfig;
 }


graylog2-web-interface/webpack.config.js
```diff
@@ -171,6 +171,5 @@ if (TARGET === 'test') {
 }
 
 if (Object.keys(module.exports).length === 0) {
-  console.error('Running in default mode');
```
joschi: Would it make sense to log this on [`console.trace()`](https://developer.mozilla.org/en-US/docs/Web/API/Console/trace)?

dennisoelkers: I think (correct me if I am wrong) you are mistaking `trace` for the log level, while `console.trace()` just logs to STDOUT including a rudimental stack trace. We removed logging to STDOUT recently, because it gets mixed in with a potential xml (ESLint) / JSON (webpack) output that is not pipeable to other tools then.

joschi: Ah, okay. Got it.

PR #3598, opened 2017-03-09
Update to react-bootstrap 0.30

edmundoa:
We were two minor versions behind the latest react-bootstrap, and there have been quite a few changes since. This PR updates the core to the latest version, and takes care of fixing deprecations.

From the [changelog](https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md), these are the most important changes tackled in this PR:

> - Rewrite form and form control API (<FormControl>, &c.), and deprecate the old API (<Input>, &c.)

We use `Input` elements in many places. To minimize the impact of this breaking change I have added an `Input` adaptor component that understands the previous API and translates it into the new one. To avoid us copying and pasting those components further, I have added a deprecation warning in the browser console. That will hopefully raise a flag when using it by accident.

For now I am recommending using plain react-bootstrap API, but it may make sense add some abstraction over it. I am open for ideas there, although in my opinion, we should work on them outside this PR.

> - Change onSelect signature to be (eventKey: any, event: SyntheticEvent) => any on all React-Bootstrap components, instead of the old inconsistent mishmash
> - Rename <PageItem> to <Pager.Item> for consistency
> - Remove pointless navItem and navDropdown props from <Button>
> - Replace deprecated <ButtonInput> with <Button>

Following the commit history is probably the best way of reviewing the changes, unfortunately the number of changes is quite big.

This PR is not marked as `ready-to-review` until I do the follow up PRs in all affected plugins, but you can start reviewing the changes if you like.


edmundoa:
I fixed a couple of issues found while adapting all plugins, and did PRs for all of them. I think this is ready for review now.

joschi:
@edmundoa Could you please rebase this PR on current `master`?

edmundoa:
Rebased and quickly tested against the current master.

PR #3600, opened 2017-03-10
Allow "-" in the path pattern for the index range rebuild endpoint

bernd:
Since the introduction of index sets, an index name can contain a "-" character.

Fixes index range rebuild for indices with "-" character in the index name.


joschi:
@bernd Should this be back-ported to the `2.2` branch?

bernd:
> @bernd Should this be back-ported to the 2.2 branch?

Yes, because it's currently not possible to trigger a manual index range recalculation for a single index if the index contains a "-" character.

Thanks for the heads up!

PR #3623, opened 2017-03-16
Support replacing whitespace in nested keys for JSON extractor

### LINKED ISSUE #3622: JSON extractor doesn't remove whitespace from nested keys
The JSON extractor has a configuration setting to replace whitespace characters in keys.

Unfortunately this functionality doesn't include nested keys, so that a JSON payload
```json
{
  "foo": {
    "b a r": 42
  }
}
```
would lead to the flattened key name "foo_b a r" instead of "foo_b_a_r".

joschi:
Fixes #3622


joschi:

PR #3625, opened 2017-03-16
Prevent session extension in SystemJobsStore

### LINKED ISSUE #3587: Unwanted session extension on SystemJobsStore
## Expected Behavior

Polling system jobs should not extend the user session. Similar changes have been made in the following PRs:

- https://github.com/Graylog2/graylog2-server/pull/3574
- https://github.com/Graylog2/graylog2-server/pull/3583

## Current Behavior

The SystemJobStore is polling without using the special HTTP header.

## Your Environment

* Graylog Version: 2.3.0-SNAPSHOT bbeefdddf35ae6cb0b00b495e432fd538681455a

joschi:
Fixes #3587

### DIFF
--- a/graylog2-web-interface/src/stores/systemjobs/SystemJobsStore.js
+++ b/graylog2-web-interface/src/stores/systemjobs/SystemJobsStore.js
@@ -3,6 +3,7 @@ import Reflux from 'reflux';
 import URLUtils from 'util/URLUtils';
 import ApiRoutes from 'routing/ApiRoutes';
 import fetch from 'logic/rest/FetchProvider';
+import fetchPeriodically from 'logic/rest/FetchProvider';
 
 import ActionsProvider from 'injection/ActionsProvider';
 const SystemJobsActions = ActionsProvider.getActions('SystemJobs');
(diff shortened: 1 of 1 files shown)


joschi:

graylog2-web-interface/src/stores/systemjobs/SystemJobsStore.js
```diff
@@ -3,6 +3,7 @@ import Reflux from 'reflux';
 import URLUtils from 'util/URLUtils';
 import ApiRoutes from 'routing/ApiRoutes';
 import fetch from 'logic/rest/FetchProvider';
+import fetchPeriodically from 'logic/rest/FetchProvider';
```
edmundoa: This is not doing what it should. It uses the default export from `FetchProvider`, which is the `fetch` function, so there is no difference with the original code, except that the function is locally using another name.
In order to fix the issue we should use a named import here `import { fetchPeriodically }...`  to use the right function.

graylog2-web-interface/src/stores/systemjobs/SystemJobsStore.js
```diff
@@ -2,7 +2,7 @@ import Reflux from 'reflux';
 
 import URLUtils from 'util/URLUtils';
 import ApiRoutes from 'routing/ApiRoutes';
-import fetch from 'logic/rest/FetchProvider';
+import {fetch, fetchPeriodically} from 'logic/rest/FetchProvider';
```
edmundoa: In this case this is also not the recommended way of importing both functions (it breaks the [`import/named`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md) eslint rule). We should use this instead:
`import fetch, { fetchPeriodically } from 'logic/rest/FetchProvider';`

joschi: WTF JavaScript…?!

PR #3632, opened 2017-03-17
Prevent session extension for polling system messages

### LINKED ISSUE #3628: Polling system messages should not extend the session
## Expected Behaviour
Leaving the system overview page open should comply with the user's session timeout settings.

## Current Behaviour
Leaving the system overview page open do not cancel the users' session, as polling the system messages (`/api/system/messages`) is extending the session.

1. Log in and go to System -> Overview
2. Wait without any interaction
3. The user is never logged out

## Your Environment
* Graylog Version: Graylog 2.3.0-SNAPSHOT (4802bbb05c47a8416fc60a5daeb3800733bca6df), Graylog 2.2.3-SNAPSHOT is also affected

joschi:
Fixes #3628

### DIFF
--- a/graylog2-web-interface/src/stores/systemmessages/SystemMessagesStore.js
+++ b/graylog2-web-interface/src/stores/systemmessages/SystemMessagesStore.js
@@ -1,9 +1,8 @@
 import Reflux from 'reflux';
 
-import UserNotification from 'util/UserNotification';
 import URLUtils from 'util/URLUtils';
 import ApiRoutes from 'routing/ApiRoutes';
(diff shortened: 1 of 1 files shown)


joschi:

joschi:
> but on the other it does not extend the session when the user is moving through the pages.

It's rather unlikely that a user will only click through the system messages related functionality for 15 minutes (or however long the session timeout is), so I guess this will not be a problem in practice.

edmundoa:
Yes, you are probably right. Still, in that case, it's safer to invalidate the session as to leave it open forever as it was before.

PR #3634, opened 2017-03-17
Prevent useless session validation when no id or username is present.

dennisoelkers:
This change checks for the presence of a session id or username before
validating the session. This has two benefits:

  1) A round trip to the server is saved upon initial login, showing the
login dialog much faster.
  2) A race condition is prevented in automated testing when the session
id is added to localStorage between two page visits. The code does not
check if a validation is currently in progress before triggering a new
one, which leads to inconsistent session propagation.

Also needs to be merged into `2.2`.


joschi:
Since it's not a bug-fix, I wouldn't merge it into the `2.2` branch.

PR #3635, opened 2017-03-17
Make id of help menu item unique.

dennisoelkers:
This small change remedies the (copy & paste?) error of the help
dropdown in the navigation bar having the same id ('user-menu-dropdown')
as the actual user menu by changing it to 'help-menu-dropdown'.

Needs to be merged into `2.2` as well.


joschi:
Since it's not a bugfix, should we really merge it into the `2.2` branch?

PR #3651, opened 2017-03-24
Fix navigation in LDAP components

### LINKED ISSUE #3485: Unable to map LDAP groups to Graylog role - Missing plugin component
Hiya,

I have got LDAP authentication working within Graylog, and now I want to map AD groups to Graylog roles, to minimise the amount of work required to set up existing AD users with Graylog stream access.

When trying to set this up within the Graylog web interface, I get the message: 
"Plugin component missing for authenticator legacy-ldap-groups, this is an error."

This is when accessing the '/system/authentication/config/legacy-ldap-groups' URL, which is the page linked from the LDAP/Active Directory config screen within Graylog .

gl-error

Please let me know if I can provide any futher infomation.

I am running the current version of Graylog aviable from the Graylog CentOS repo (Graylog 2.2.0). I followed the instation guide below to install Graylog.
http://docs.graylog.org/en/2.2/pages/installation/os/centos.html

* Graylog Version: Graylog 2.2.0+d9681cb
* Elasticsearch Version: 2.4.4
* MongoDB Version: 3.4.2
* Operating System: CentOS 7
* Browser version: Firefox 51.0.1

edmundoa:
LDAP components do not have single URLs any more, so we use callbacks to let the parent decide which component to render.

I also fixed in here a small issue when the LDAP group component tries to render an error message object. That does not work and should also be avoided when the server returns 400 (i.e. LDAP is not configured or no groups are configured), as the default message is more helpful in that case.

Fixes #3485

Once the issue is merged I'll backport it to 2.2.3.


joschi:
@edmundoa Please rebase this PR on current master.

PR #3688, opened 2017-04-04
Revert "Dependency updates (#3672)"

dennisoelkers:
The update of auto-value from 1.4 introduced a breaking change that
disallows properties being nullable if a method exists, which returns a
builder for those. This breaks some existing plugin code which needs to
be fixed first.

This reverts commit 2a82ece7ed99d0a6fbc6894a69e12936e364f251.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,9 +88,9 @@
 
         <!-- Dependencies -->
         <airline.version>2.2.0</airline.version>
-        <amqp-client.version>4.1.0</amqp-client.version>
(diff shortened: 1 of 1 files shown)


joschi:
@dennisoelkers Would it make sense to only revert Google Auto Value to the previous version (1.3) instead of reverting the complete PR?

dennisoelkers:
Sure. This was supposed to be a quick fix to make compiling plugins work again. Feel free to modify the PR.

joschi:
Fixed by updating the relevant plugin code.

Issue #2309, opened 2016-05-30
Move Drools support into plugin


joschi:
Move the Drools rules engine from the Graylog core into a plugin to further slim down the core project.

bernd:
This has been done in https://github.com/Graylog2/graylog2-server/pull/4694

PR #3695, opened 2017-04-10
Move Drools support into plugin

fbalicchia:
Move Drools support into plugin #2309

Move the Drools rules engine from the Graylog core into a plugin to further slim down the core project.

## Description
Move Drools rules engine outside from Graylog core in a plugin

## Motivation and Context
issue #2309

## How Has This Been Tested?
After remove Drools dependency and reference in Graylog server for example in ServerBindings,
install [drools-plugin](https://github.com/fbalicchia/graylog-plugin-drools/releases/download/0.0.7/graylog-plugin-drools-0.0.7.jar)
enable rules_file server conf and use as rule this script

`
rule "Overwrite localhost"
    when
        m : Message( source == "localhost" )
    then
        m.addField("source", "localhost.hello.it" );
        log.info("[Overwrite localhost] rule fired: {}", m);
end
`
In all message the field 'source' contains 'localhost.hello.it' as expected

## Types of changes

## Checklist:

At the moment I think there are some plugins that reference 'DroolsEngine' for 
[example](https://github.com/Graylog2/graylog-plugin-pipeline-processor/blob/master/benchmarks/src/main/java/org/graylog/benchmarks/pipeline/FilterchainVsPipeline.java#L229)
and  If a patch looks good for you is ok you should change reference of other plugins.

Thanks for attention


kroepke:
Hi @fbalicchia!

Thank you for your contribution! In general I support this, but there are a couple of considerations:
1) As you noted, we will need to remove references elsewhere.
2) This is a breaking change for users and therefore needs to wait until  3.0 (we have 2.3 scheduled in between). So as of now we cannot simply merge this.
3) I would prefer to  expose the rules engine as its own `MessageProcessor` instead of the legacy `MessageFilter`. The former can be re-ordered during runtime and are not limited to returning a single message.

Thanks again!

fbalicchia:
Hi @kroepke, 
now plugin implements MessageProcessor instead of MessageFilter..

In your comment say that MessageFilter interface is legacy but in code I haven't see that is deprecated in favor of MessageProcess but probably two interface by design have  different scope ? Am I wrong ?
An other thing that I don't understand is why processBuffer.handleMessage retrive a `Message` and then before call process it become `Messages`  [processBuffer](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java#L111)

Thanks a Lot for your tip

--Filippo

kroepke:
`MessageFilter` is not deprecated yet, because we have third party plugins that depend on it right now and there's little need to remove it immediately.
However `MessageProcessor` can produce multiple messages and can be reordered at runtime so it is the preferred way.
However, as I said this change cannot be done before 3.0 as it is a breaking change for many users.

joschi:
@fbalicchia Any chance to rebase this PR on current master?

fbalicchia:
Hi @joschi, sorry for delay, yes I'll done it

fbalicchia:
@joschi  done. Now plugin point to current master. What I did not understand if I have to remove the drools references on the server

joschi:
@fbalicchia I think you forgot to push your changes. 😉

fbalicchia:
close this pull in favor of #4247

PR #3713, opened 2017-04-19
Migrate Counts class to Jest

joschi:


graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -50,13 +57,22 @@ private long totalCount(final String[] indexNames) {
             return 0L;
         }
 
+        final Count request = new Count.Builder()
+                .query("{\"query\":{\"match_all\":{}}}")
```
dennisoelkers: While this is rather easy for the `Counts` class, do we really want to hardcode static query strings instead of using some helper to build them in general for the long run?

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -50,13 +57,22 @@ private long totalCount(final String[] indexNames) {
...
-            return c.prepareSearch(indexNames)
-                    .setSize(0)
-                    .get()
-                    .getHits()
-                    .getTotalHits();
-        } catch (ElasticsearchException e) {
+            result = jestClient.execute(request);
+        } catch (IOException e) {
+            LOG.error("Couldn't fetch message count", e);
+            return -1L;
```
dennisoelkers: Do we really want to return a C-style bogus value which we would need to interpret as an error at the consumer's level (and which would be just passed on as the actual value if we forget that) instead of throwing a higher-level exception at this place?

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -50,13 +57,22 @@ private long totalCount(final String[] indexNames) {
...
+        }
+
+        if (result.isSucceeded()) {
+            return result.getCount().longValue();
+        } else {
+            LOG.error("Fetching message count failed: {}", result.getErrorMessage());
             return -1L;
```
dennisoelkers: Same here.

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
joschi: We should use the proper Elasticsearch query builders. :+1:

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
joschi: ✅ We're using `ElasticsearchException` now.

graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java
```diff
@@ -50,14 +62,25 @@ private long totalCount(final String[] indexNames) {
             return 0L;
         }
 
+        final List<String> indices = Arrays.asList(indexNames);
+        final String query = new SearchSourceBuilder()
+                .query(QueryBuilders.matchAllQuery().buildAsBytes(XContentType.JSON))
```
dennisoelkers: Isn't it enough to just call `.query(QueryBuilders.matchAllQuery())` or is it important to pass a `BytesReference`?

joschi: No, because that would only generate `{"match_all":{}}` instead of `{"query":{"match_all":{}}}`.

Issue #3722, opened 2017-04-21
API call to change index rotation strategy not respected


stenio123:
When making an API call to change the Index rotation strategy to SizeBasedRotation strategy, I get "success" http 200 code, and checking the API Browser correctly shows SizeBasedRotation strategy. However the default index strategy does not change, it is still in Message Count.

## Expected Behavior
Make API call to api/system/indices/rotation/config passing
```
{"strategy":"org.graylog2.indexer.rotation.strategies.SizeBasedRotationStrategy","config":{"type":"org.graylog2.indexer.rotation.strategies.SizeBasedRotationStrategyConfig","max_size":4294967296}}
```
Should change existing index rotation strategy.

## Current Behavior
This call changes what is show in the api browser, but not what is shown in the Indices page.

## Steps to Reproduce (for bugs)

1. Set your index to "Message Count" using the web UI
2. Go to API Browser
3. Go to System/Indices/Rotation : Index rotation strategy settings 
4. Execute PUT /system/indices/rotation/config with:
```
{
  "strategy": "org.graylog2.indexer.rotation.strategies.SizeBasedRotationStrategy",
  "config": {
    "type": "org.graylog2.indexer.rotation.strategies.SizeBasedRotationStrategyConfig",
    "max_size": 4294967296
  }
}
```

## Context
Need to automate deployment, can't rely on manual steps. Therefore I need api or configs to work.

* Graylog Version: 2.2.3+7adc951
* Elasticsearch Version: 2.2.1
* MongoDB Version: 3.0.4
* Operating System: Amazon Linux 4.4.44-39.55.amzn1.x86_64
* Browser version: Chrome  56.0.2924.87 (64-bit)

joschi:
@stenio123 The rotation and retention strategies are part of index sets as of Graylog 2.2.0, see http://docs.graylog.org/en/2.2/pages/upgrade/graylog-2.2.html#default-stream-index-sets.

You need to update your scripts which interact with the Graylog REST API, to use the correct HTTP resource (`/system/indices/index_sets`).

stenio123:
Cool, it works! Thank you @joschi . Perhaps it might be a good idea to remove the `/system/indices/rotation/config` and `/system/indices/retention/config` API endpoints if they are no longer used? Thanks!

PR #3724, opened 2017-04-21
Remove deprecated rotation/retention resources

joschi:
The rotation and retention strategies have been part of index sets since Graylog 2.2.0.

This change set removes the otherwise useless `RotationStrategyResource` and `RetentionStrategyResource` classes.

Refs #3722


bernd:
LGTM :+1:, thanks!

Issue #3366, opened 2017-01-17
Edit Extractor - Really slow (seems to depend on volumetry)


Bardack66:
Hi,

We have an input containing 3.000.000.000 messages.
This input is configured with extractors, and those needs to be adapted from time to time.
However, when we click on "Edit" for 1 of the extractor, it takes ages to load (about 103seconds).

The slowest network request (the one that takes 103sec) is : 
http://192.168.151.102:9000/api/search/universal/relative?query=gl2_source_input%3A5763b52a2696b74aaee126b8%20OR%20gl2_source_radio_input%3A5763b52a2696b74aaee126b8&limit=1

To get the  total amount of messages for that input, I checked using graylog web interface, and funny  it took also around 103 seconds to get the info.

So I suspect that when editing an extractor, it loads 1 message but basically  check for ALL before getting only 1.

Regards;

PR #3725, opened 2017-04-22
Better timerange for Show Received Messages button

lennartkoopmann:
The old behaviour was to link to a search that only looked at the last 8 hours of data. I've come across a whole bunch of users who had messages with wrong timestamps and would only find them when searching in "All messages". This change makes the button link to a search over "All messages" instead, so users with timestamps in the past can find the messages and identify the issue.

### DIFF
--- a/graylog2-web-interface/src/components/inputs/InputListItem.jsx
+++ b/graylog2-web-interface/src/components/inputs/InputListItem.jsx
@@ -60,7 +60,7 @@ const InputListItem = React.createClass({
     if (this.isPermitted(this.props.permissions, ['searches:relative'])) {
       actions.push(
         <LinkContainer key={`received-messages-${this.props.input.id}`}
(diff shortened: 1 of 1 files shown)


joschi:
@lennartkoopmann The problem is, that this will kill reasonably loaded Elasticsearch clusters.

See #3366 for a related issue.

billmurrin:
Hope it is OK to add my thoughts to this discussion. Certain solutions will tag a record upon ingestion with a receipt time and then you still have the event time of the message. While this can be confusing for new users, I can see this sort of thing helping in this particular use case.

lennartkoopmann:
@jochen I considered this and I think that the value of not confusing new users is higher. Most new users that I am seeing are sooner or later searching over "all messages" anyways.

The linked issue #3366 is having this problem with a function that is used more often and especially more often under production loads.

PR #3728, opened 2017-04-24
Update vendor-module-ids.json file

joschi:


joschi:
@bernd That's probably because #3576 was merged in the meantime.

PR #3732, opened 2017-04-25
Replace ES node client with Jest in Cluster class

joschi:


graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -17,69 +17,82 @@
...
+    @Nullable
+    private JsonObject clusterHealth(Collection<? extends String> indices) {
+        final Health request = new Health.Builder()
+                .addIndex(indices)
+                .build();
+        try {
+            return jestClient.execute(request).getJsonObject();
+        } catch (IOException e) {
+            LOG.error("Couldn't read cluster health", e);
+            return null;
```
dennisoelkers: Maybe we should throw a subclass of the `ElasticsearchException` wrapping the caught exception instead of returning `null`? This way, possibly important information about the fault gets lost.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -90,63 +103,54 @@ public ClusterHealthResponse health() {
...
+        final CatResult response;
+        try {
+            response = jestClient.execute(request);
+        } catch (IOException e) {
+            LOG.error("Unable to read Elasticsearch node file descriptors", e);
+            return new JsonArray();
         }
 
-        return builder.build();
+        return response.getJsonObject().getAsJsonArray("result");
```
dennisoelkers: Should we really return a `JsonArray` here, considering it's a public method? It's not really the most intuitive data structure to work with.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -90,63 +103,54 @@ public ClusterHealthResponse health() {
...
         try {
-            NodesInfoResponse r = c.admin().cluster().nodesInfo(new NodesInfoRequest(nodeId).all()).actionGet();
-            return r.getNodesMap().get(nodeId);
-        } catch (Exception e) {
-            LOG.error("Could not read name of ES node.", e);
+            final JestResult result = jestClient.execute(new NodesInfo.Builder().addNode(nodeId).build());
+            return result.getJsonObject().getAsJsonObject("nodes").getAsJsonObject(nodeId);
+        } catch (IOException e) {
+            LOG.error("Couldn't read information of Elasticsearch node.", e);
             return null;
```
dennisoelkers: Same as above.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -130,28 +136,48 @@ public JsonArray catNodes(String fields) {
...
-        final JsonObject nodeInfo = getNodeInfo(nodeId);
-        return nodeInfo == null ? "UNKNOWN" : nodeInfo.getAsJsonPrimitive("name").getAsString();
-
+        return getNodeInfo(nodeId)
+                .map(nodeInfo -> nodeInfo.get("name"))
+                .filter(JsonElement::isJsonPrimitive)
+                .map(JsonElement::getAsJsonPrimitive)
+                .filter(JsonPrimitive::isString)
+                .map(JsonPrimitive::getAsString)
+                .orElse("UNKNOWN");
```
dennisoelkers: Can we turn the result type into an `Optional<String>` instead? Rendering it as unknown would be up to the presentation layer then.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -130,28 +136,48 @@ public JsonArray catNodes(String fields) {
...
     public String nodeIdToHostName(String nodeId) {
-        final JsonObject nodeInfo = getNodeInfo(nodeId);
-        return nodeInfo == null ? "UNKNOWN" : nodeInfo.getAsJsonPrimitive("host").getAsString();
+        return getNodeInfo(nodeId)
+                .map(nodeInfo -> nodeInfo.get("host"))
+                .filter(JsonElement::isJsonPrimitive)
+                .map(JsonElement::getAsJsonPrimitive)
+                .filter(JsonPrimitive::isString)
+                .map(JsonPrimitive::getAsString)
+                .orElse("UNKNOWN");
```
dennisoelkers: See above, plus: "UNKNOWN" is a valid host name. If someone ever wants to actually use the result of this beyond displaying it, that could be rather nasty.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -90,64 +112,83 @@ public ClusterHealthResponse health() {
...
+        return setBuilder.build();
     }
 
     public String nodeIdToName(String nodeId) {
-        final NodeInfo nodeInfo = getNodeInfo(nodeId);
-        return nodeInfo == null ? "UNKNOWN" : nodeInfo.getNode().getName();
-
+        return getNodeInfo(nodeId)
+                .map(nodeInfo -> GsonUtils.asString(nodeInfo.get("name")))
+                .orElse("UNKNOWN");
```
dennisoelkers: Can we turn the result type into an `Optional<String>` instead? Rendering it as unknown would be up to the presentation layer then.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/Cluster.java
```diff
@@ -90,64 +112,83 @@ public ClusterHealthResponse health() {
...
     }
 
     public String nodeIdToHostName(String nodeId) {
-        final NodeInfo nodeInfo = getNodeInfo(nodeId);
-        return nodeInfo == null ? "UNKNOWN" : nodeInfo.getHostname();
+        return getNodeInfo(nodeId)
+                .map(nodeInfo -> GsonUtils.asString(nodeInfo.get("host")))
+                .orElse("UNKNOWN");
```
dennisoelkers: See above, plus: `"UNKNOWN"` is a valid host name. If someone ever wants to actually use the result of this beyond displaying it, that could be rather nasty.

PR #3733, opened 2017-04-25
Replace ES node client with Jest in ElasticsearchProbe

joschi:


graylog2-server/src/main/java/org/graylog2/system/stats/elasticsearch/ElasticsearchProbe.java
```diff
@@ -17,80 +17,116 @@
...
         this.indexSetRegistry = indexSetRegistry;
     }
 
     public ElasticsearchStats elasticsearchStats() {
-        final ClusterAdminClient adminClient = client.admin().cluster();
+        final JestResult clusterStatsResponse;
+        try {
+            clusterStatsResponse = jestClient.execute(new Stats.Builder().build());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
dennisoelkers: Maybe we could wrap it in a (subclass of the) `ElasticsearchException`? This makes it a bit easier to catch this specific case on higher levels.

PR #3736, opened 2017-04-26
Do not include 'in_grace' flag in streams list.

dennisoelkers:
## Description

This change is removing the `in_grace` field from the stream list retrieved via `StreamResource#get/getEnabled/get(id)/update`. The DTO is slightly modified, it is not including the field if it was initialized with `null`. Other parts of the code using it can still do so and it will continue to include the flag.

## Motivation and Context

The `in_grace` field is not shown in the web interface in stream lists, but it is right now costly to calculate, compared to just retrieving all/individual streams. If the flag will be used in future parts, it can be retrieved for streams individually using additional API calls.

The performance benefit of not calculating the field is massive for those parts, which are fetching the whole stream list. This is a single call to `/streams` with 39 configured streams before the change:

screen shot 2017-04-26 at 12 40 22

And after:

screen shot 2017-04-26 at 12 37 03

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/streams/alerts/AlertConditionSummary.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/streams/alerts/AlertConditionSummary.java
@@ -18,6 +18,7 @@ package org.graylog2.rest.models.streams.alerts;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 import org.graylog.autovalue.WithBeanGetter;
@@ -25,6 +26,8 @@ import org.graylog.autovalue.WithBeanGetter;
 import javax.annotation.Nullable;
 import java.util.Date;
 import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 @AutoValue
 @WithBeanGetter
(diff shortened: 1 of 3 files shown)


bernd:
@dennisoelkers Please add the change to the `UPGRADING.rst` file. Thanks!

dennisoelkers:
It was omitted on purpose to let the discussion about the version it is going to be included in happen first.

bernd:
> It was omitted on purpose to let the discussion about the version it is going to be included in happen first.

Okay, @kroepke voted for 2.3 and I do as well because this is a huge improvement for users with lots of streams. :+1: 

I will put the PR into the 2.3 milestone.

graylog2-server/src/main/java/org/graylog2/rest/models/streams/alerts/AlertConditionSummary.java
```diff
@@ -60,6 +65,17 @@ public static AlertConditionSummary create(@JsonProperty("id") String id,
                                                @JsonProperty("parameters") Map<String, Object> parameters,
                                                @JsonProperty("in_grace") Boolean inGrace,
                                                @JsonProperty("title") @Nullable String title) {
+        checkNotNull(inGrace);
```
joschi: We could save this by using the primitive `boolean` in the method signature.

joschi: Would it make sense to mark this `create()` method as deprecated?

dennisoelkers: No, it is used by the `(Stream)AlertConditionsResource`, where we want that flag to be included to show it for stream/alert condition details.

graylog2-server/src/main/java/org/graylog2/rest/models/streams/alerts/AlertConditionSummary.java
```diff
@@ -60,6 +65,17 @@ public static AlertConditionSummary create(@JsonProperty("id") String id,
...
         return new AutoValue_AlertConditionSummary(id, type, creatorUserId, createdAt, parameters, inGrace, title);
     }
+
+    @JsonCreator
```
joschi: There should be only one method annotated with `@JsonCreator`.

PR #3740, opened 2017-04-27
Migrate Indices class to Jest

joschi:


graylog2-server/src/main/java/org/graylog2/indexer/indices/IndexStatistics.java
```diff
@@ -17,27 +17,134 @@
...
+import static org.graylog2.indexer.gson.GsonUtils.asLong;
+import static org.graylog2.indexer.gson.GsonUtils.asString;
 
 @AutoValue
-@WithBeanGetter
 public abstract class IndexStatistics {
     public abstract String indexName();
 
-    public abstract CommonStats primaries();
+    public abstract JsonObject primaries();
```
dennisoelkers: Going from a structured data type to an unstructured one is a regression imo. I think we should keep the original data types and go the extra mile to extract the values in `buildIndexStats()`.

This also helps us to prevent unnecessary changes in consumers of this structure.

graylog2-server/src/main/java/org/graylog2/indexer/indices/IndexStatistics.java
```diff
@@ -17,27 +17,134 @@
...
 
-    public abstract CommonStats total();
+    public abstract JsonObject total();
```
dennisoelkers: Same here.

graylog2-server/src/main/java/org/graylog2/indexer/indices/IndexStatistics.java
```diff
@@ -17,27 +17,134 @@
...
 
-    public abstract List<ShardRouting> shardRoutings();
+    public abstract JsonObject shardRoutings();
```
dennisoelkers: Same here.

PR #3758, opened 2017-05-02
Clean up remaining unnecessary Elasticsearch classes

joschi:


graylog2-server/src/main/java/org/graylog2/indexer/results/IndexQueryResult.java
```diff
@@ -45,19 +32,10 @@ public String getOriginalQuery() {
...
-    }
-
-    public TimeValue took() {
-        return took;
+        return builtQuery;
     }
 
-    static List<ResultMessage> buildResults(SearchHits hits) {
-        return StreamSupport.stream(hits.spliterator(), false).map(ResultMessage::parseFromSource).collect(Collectors.toList());
+    public long tookMs() {
```
dennisoelkers: We need to consider that this is breaking consumers, i.e. the map widget plugin.

PR #3759, opened 2017-05-02
Implement a generic refresh for data adapters and make csvfile use it

bernd:


graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupDataAdapter.java
```diff
@@ -84,6 +107,21 @@ public void stop() {
     }
     protected abstract void doStop() throws Exception;
 
+    /**
+     * Returns the refresh interval for this data adapter. Use {@link Duration.ZERO} if refresh should be disabled.
```
kroepke: extreme nitpick: javadoc links to members must use '#' instead of '.', otherwise IDEs will flag an error:
`{@link Duration#ZERO}`

PR #3761, opened 2017-05-03
Reduce number of "n+1"-type queries in IndicesResource

joschi:


graylog2-server/src/main/java/org/graylog2/indexer/NodeInfoCache.java
```diff
@@ -0,0 +1,66 @@
...
+    private static final Duration EXPIRE_DURATION = Duration.minutes(1L);
+    private final LoadingCache<String, Optional<String>> nodeNameCache;
+    private final LoadingCache<String, Optional<String>> hostNameCache;
+
+    @Inject
+    public NodeInfoCache(Cluster cluster) {
+        this(new FunctionCacheLoader<>(cluster::nodeIdToName), new FunctionCacheLoader<>(cluster::nodeIdToHostName));
+    }
+
+    private NodeInfoCache(CacheLoader<String, Optional<String>> nodeNameCache,
```
dennisoelkers: Why do we need this constructor?

graylog2-server/src/main/java/org/graylog2/indexer/NodeInfoCache.java
```diff
@@ -0,0 +1,66 @@
...
+                CacheBuilder.newBuilder()
+                        .expireAfterWrite(EXPIRE_DURATION.getQuantity(), EXPIRE_DURATION.getUnit())
+                        .build(nodeNameCache),
+                CacheBuilder.newBuilder()
+                        .expireAfterWrite(EXPIRE_DURATION.getQuantity(), EXPIRE_DURATION.getUnit())
+                        .build(hostNameCache)
+        );
+    }
+
+    private NodeInfoCache(LoadingCache<String, Optional<String>> nodeNameCache,
```
dennisoelkers: Why do we need this constructor?

graylog2-server/src/main/java/org/graylog2/indexer/NodeInfoCache.java
joschi: ✅ The intention was to use them in test cases but since they aren't used in that way, I'll remove them.

graylog2-server/src/main/java/org/graylog2/indexer/NodeInfoCache.java
joschi: ✅

PR #3766, opened 2017-05-03
Fix dashboard layouts when width or height is 0

### LINKED ISSUE #3765: Dashboard layout broken when dashboard is created from content pack
## Expected Behaviour
Layout of dashboards created from content packs looks fine, even if the widget had no `width` and `height` information.

## Current Behaviour
Layout of dashboards created from content packs is broken when widget does not include `width` and/or `height` information.

1. Import and apply a content pack including a dashboard, like: https://github.com/Graylog2/omnibus-graylog2/blob/294decadc40183f47e738b0c9c70df93aedaaac9/files/data/contentpacks/content_pack_appliance.json
2. Go to the dashboard created from the content pack and see that the layout is broken

## Your Environment
* Graylog Version: Graylog 2.3.0-alpha.1

edmundoa:
Dashboards created from content packs always get a value for each widget dimensions, being 0 the default. This behaviour is different than what happens when a widget is created manually and is now taken into account.

Fixes #3765

### DIFF
--- a/graylog2-web-interface/src/pages/ShowDashboardPage.jsx
+++ b/graylog2-web-interface/src/pages/ShowDashboardPage.jsx
@@ -119,6 +119,9 @@ const ShowDashboardPage = React.createClass({
   _dashboardIsEmpty(dashboard) {
     return dashboard.widgets.length === 0;
   },
+  _validDimension(dimension) {
+    return dimension !== null && dimension !== undefined && dimension !== 0;
+  },
   formatDashboard(dashboard) {
     if (this._dashboardIsEmpty(dashboard)) {
       return this.emptyDashboard();
@@ -131,8 +134,8 @@ const ShowDashboardPage = React.createClass({
       positions[widget.id] = {
         col: (persistedDimensions.col === undefined ? defaultDimensions.col : persistedDimensions.col),
         row: (persistedDimensions.row === undefined ? defaultDimensions.row : persistedDimensions.row),
-        height: (persistedDimensions.height === undefined ? defaultDimensions.height : persistedDimensions.height),
-        width: (persistedDimensions.width === undefined ? defaultDimensions.width : persistedDimensions.width),
+        height: (this._validDimension(persistedDimensions.height) ? persistedDimensions.height : defaultDimensions.height),
+        width: (this._validDimension(persistedDimensions.width) ? persistedDimensions.width : defaultDimensions.width),
       };
     });
 
--- a/graylog2-web-interface/src/pages/ShowDashboardPage.jsx
+++ b/graylog2-web-interface/src/pages/ShowDashboardPage.jsx
@@ -120,7 +120,7 @@ const ShowDashboardPage = React.createClass({
     return dashboard.widgets.length === 0;
   },
   _validDimension(dimension) {
-    return dimension !== null && dimension !== undefined && dimension !== 0;
+    return Number.isInteger(dimension) && dimension > 0;
   },
   formatDashboard(dashboard) {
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/pages/ShowDashboardPage.jsx
```diff
@@ -131,8 +134,8 @@ const ShowDashboardPage = React.createClass({
       positions[widget.id] = {
         col: (persistedDimensions.col === undefined ? defaultDimensions.col : persistedDimensions.col),
         row: (persistedDimensions.row === undefined ? defaultDimensions.row : persistedDimensions.row),
-        height: (persistedDimensions.height === undefined ? defaultDimensions.height : persistedDimensions.height),
-        width: (persistedDimensions.width === undefined ? defaultDimensions.width : persistedDimensions.width),
+        height: (this._validDimension(persistedDimensions.height) ? persistedDimensions.height : defaultDimensions.height),
```
dennisoelkers: Minor nitpick: why not just use `persistedDimensions.height || defaultDimensions.height`? It is shorter, saves a function call and makes the intention even clearer.

graylog2-web-interface/src/pages/ShowDashboardPage.jsx
```diff
@@ -131,8 +134,8 @@ const ShowDashboardPage = React.createClass({
...
+        width: (this._validDimension(persistedDimensions.width) ? persistedDimensions.width : defaultDimensions.width),
```
dennisoelkers: Same here.

graylog2-web-interface/src/pages/ShowDashboardPage.jsx
edmundoa: For me this is by far more explicit, as I'm calling a function just to say this may have a different value than just `undefined`.

dennisoelkers: Using `||` would do far more than just checking for `undefined`, it also checks for `null` and `0`, so it is doing exactly the same.

edmundoa: Yes, but it does not make the intention clearer. We use this shortcut all over the place when we can't be bother to look if it's `undefined` or `null`. Using `||` also checks for `false` and probably two or three other things. In this way I'm clearly stating that this value can be `0` even if it doesn't look like it.

bernd:
@dennisoelkers @edmundoa Can we move on with this? The discussion seems to be around a minor nitpick, not a real issue.

graylog2-web-interface/src/pages/ShowDashboardPage.jsx
```diff
@@ -119,6 +119,9 @@ const ShowDashboardPage = React.createClass({
   _dashboardIsEmpty(dashboard) {
     return dashboard.widgets.length === 0;
   },
+  _validDimension(dimension) {
+    return dimension !== null && dimension !== undefined && dimension !== 0;
```
joschi: Shouldn't this be as follows (using [`Number.isInteger()`](https://www.w3schools.com/jsref/jsref_isinteger.asp)):
```javascript
return Number.isInteger(dimension) && dimension > 0;
```

edmundoa: The problem is that `Number.isInteger()` is not implemented in IE. I'll check if the polyfill we use supports it.

edmundoa: It works fine, so I changed the function 👍

PR #3768, opened 2017-05-03
Introduce total searches counter used by anonymous usage statistics.

dennisoelkers:
Supporting https://github.com/Graylog2/graylog-plugin-anonymous-usage-statistics/pull/23.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.indexer.searches;
 
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -172,6 +174,7 @@ public Searches(Configuration configuration,
 
         this.esRequestTimer = metricRegistry.timer(name(Searches.class, "elasticsearch", "requests"));
         this.esTimeRangeHistogram = metricRegistry.histogram(name(Searches.class, "elasticsearch", "ranges"));
+        this.esTotalSearchesCounter = metricRegistry.counter(name(Searches.class, "elasticsearch", "searches"));
```
joschi: The metric name should be more explicit, e. g. `total-searches` instead of `searches`.

PR #3770, opened 2017-05-04
Revert parts of #3724 to unbreak editing index sets

bernd:
The list of available rotation/retention strategies is still needed.


bernd:
It's missing a commit...

bernd:
Missing commit pushed.

PR #3776, opened 2017-05-04
refactor FileInfo to never return null or throw exceptions

kroepke:
this makes the class easier to use. if a file isn't present or inaccessible, return a null object that allows rechecking for modifications.
if the size of the file is the same after modification, this class can only detect changes if the modification time has changes. the resolution of that is typically only a second


kroepke:
Makes https://github.com/Graylog2/graylog-plugin-map-widget/pull/41 obsolete

PR #3788, opened 2017-05-05
Fixing `Indices#exists` and `Indices#aliasExists`.

dennisoelkers:
Before this change, the checks in both methods were too simple:

  * `exists` checked if a HEAD on `/<indexname>` succeeds. This returns
`true` if the supplied name is an alias (which the method is clearly
supposed not to do according to
https://github.com/Graylog2/graylog2-server/blame/master/graylog2-server/src/main/java/org/graylog2/periodical/IndexRotationThread.java#L128.
Instead we are now getting the settings for the index and check if it
actually is an alias or not.

  * `aliasExists` checked if a HEAD on `/_all/_alias/<indexname>`
succeeds. Unfortunately, this also returns true if the name is an index
and not an alias. Instead we are now doing the same call as for `exists`
and reversing the alias condition.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
@@ -298,7 +298,14 @@ public class Indices {
 
     public boolean exists(String indexName) {
         try {
-            return jestClient.execute(new IndicesExists.Builder(indexName).build()).isSucceeded();
+            final JestResult result = jestClient.execute(new GetSettings.Builder().addIndex(indexName).build());
+            if (!result.isSucceeded()) {
+                return false;
+            }
+            return Optional.of(result.getJsonObject())
+                .map(GsonUtils::entrySetAsMap)
+                .map(map -> map.containsKey(indexName))
+                .orElse(false);
         } catch (IOException e) {
             throw new ElasticsearchException("Couldn't check existence of index " + indexName, e);
         }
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -298,15 +298,29 @@ public long numberOfMessages(String indexName) throws IndexNotFoundException {
 
     public boolean exists(String indexName) {
```
joschi: Please add a short Javadoc comment what the method is supposed to do (and what not).

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
```diff
@@ -298,15 +298,29 @@ public long numberOfMessages(String indexName) throws IndexNotFoundException {
...
+            return Optional.of(result.getJsonObject())
+                .map(GsonUtils::entrySetAsMap)
+                .map(map -> map.containsKey(indexName))
+                .orElse(false);
         } catch (IOException e) {
             throw new ElasticsearchException("Couldn't check existence of index " + indexName, e);
         }
     }
 
     public boolean aliasExists(String alias) {
```
joschi: Please add a short Javadoc comment what the method is supposed to do (and what not).

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
joschi: Please add tests for the following cases:
* Index exists
* Index does not exist

Ideally, the test for an alias would be separate, but it's also okay as a part of `IndicesTest#testAliasExists()`.

graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java
joschi: Please add tests for the following cases:
* `alias` points to an index

PR #3794, opened 2017-05-08
Adding helpful exception when search tries to use nonexisting indices.

dennisoelkers:
This could happen when indices were removed manually, the cluster was
switched and/or the index ranges are out of date.

In this case, the change now emits a useful exception instead of a
generic one.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
@@ -74,8 +74,12 @@ public class JestUtils {
 
         for (JsonObject rootCause : rootCauses) {
             final String type = asString(rootCause.get("type"));
-            if ("query_parsing_exception".equals(type)) {
-                return buildQueryParsingException(errorMessage, rootCause, reasons);
+            switch(type) {
+                case "query_parsing_exception":
+                    return buildQueryParsingException(errorMessage, rootCause, reasons);
+                case "index_not_found_exception":
+                    final String indexName = asString(rootCause.get("resource.id"));
+                    return new ElasticsearchException("Index not found for query: " + indexName + ". Try recalculating your index ranges.");
             }
         }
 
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
```diff
@@ -74,8 +74,15 @@ private static ElasticsearchException specificException(Supplier<String> errorMe
...
-                return buildQueryParsingException(errorMessage, rootCause, reasons);
+            if (type == null) {
+                continue;
+            }
+            switch(type) {
+                case "query_parsing_exception":
+                    return buildQueryParsingException(errorMessage, rootCause, reasons);
+                case "index_not_found_exception":
+                    final String indexName = asString(rootCause.get("resource.id"));
+                    return new ElasticsearchException("Index not found for query: " + indexName + ". Try recalculating your index ranges.");
```
joschi: Please use [`IndexNotFoundException`](https://github.com/Graylog2/graylog2-server/blob/eee0a3ae5dd72eb4a4924120066a52683ec831cc/graylog2-server/src/main/java/org/graylog2/indexer/IndexNotFoundException.java) here.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
```diff
@@ -74,8 +75,15 @@ private static ElasticsearchException specificException(Supplier<String> errorMe
...
-                return buildQueryParsingException(errorMessage, rootCause, reasons);
+            if (type == null) {
+                continue;
+            }
+            switch(type) {
+                case "query_parsing_exception":
+                    return buildQueryParsingException(errorMessage, rootCause, reasons);
+                case "index_not_found_exception":
+                    final String indexName = asString(rootCause.get("resource.id"));
+                    return new IndexNotFoundException("Index not found for query: " + indexName + ". Try recalculating your index ranges.");
```
joschi: What do you think about using `errorMessage` here, too? This would give developers the possibility to customize the error message (also see [`Counts#totalCount()`](https://github.com/Graylog2/graylog2-server/blob/a22a712411262a676ce1d0a9cfdfc969f5a48a88/graylog2-server/src/main/java/org/graylog2/indexer/counts/Counts.java#L67)

dennisoelkers: ✅ Good idea!

PR #3795, opened 2017-05-08
Add lifecycle and error messages to lookup table domain objects

kroepke:
This change adds error handling and exposing error messages to lookup tables and data adapters.
Caches don't have lifecycle at the moment, but once they do they will share the same logic as data adapters.
Lookup tables are supposed to expose aggregate information, currently they duplicate the error message from the data adapter.

New UI shows the errors of the currently displayed page.

### DIFF
--- a/graylog2-web-interface/src/components/lookup-tables/DataAdapterTableEntry.jsx
+++ b/graylog2-web-interface/src/components/lookup-tables/DataAdapterTableEntry.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { LinkContainer } from 'react-router-bootstrap';
 
-import { Button } from 'react-bootstrap';
+import { Button, OverlayTrigger, Popover } from 'react-bootstrap';
 
 import Routes from 'routing/Routes';
 import CombinedProvider from 'injection/CombinedProvider';
@@ -12,6 +12,13 @@ const DataAdapterTableEntry = React.createClass({
 
   propTypes: {
     adapter: React.PropTypes.object.isRequired,
+    error: React.PropTypes.string,
+  },
+
+  getDefaultProps() {
+    return {
+      error: null,
+    };
   },
 
   _onDelete() {
@@ -22,10 +29,18 @@ const DataAdapterTableEntry = React.createClass({
   },
 
   render() {
+    const errorOverlay = !this.props.error ? null : (
+      <Popover id="popover-table-error" title="Lookup Table problem" style={{ maxWidth: 400 }}>
+        {this.props.error}
+      </Popover>);
     return (
       <tbody>
         <tr>
           <td>
+            {this.props.error && (
+              <OverlayTrigger trigger={['hover', 'focus']} placement="right" overlay={errorOverlay}>
+                <i className="fa fa-warning text-danger" style={{ marginRight: 5 }} />
+              </OverlayTrigger>)}
             <LinkContainer to={Routes.SYSTEM.LOOKUPTABLES.DATA_ADAPTERS.show(this.props.adapter.name)}><a>{this.props.adapter.title}</a></LinkContainer>
           </td>
           <td>{this.props.adapter.description}</td>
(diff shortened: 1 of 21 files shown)


edmundoa:
The map widget cannot compile with these changes, as they change the `LookupDataAdapter` constructor signature. Is there a branch I can use for testing? I'll disable the plugin for now.

kroepke:
The map plugin needs the referenced PR to build, yes. Sorry I forgot to click the button :)

graylog2-web-interface/src/stores/lookup-tables/LookupTablesStore.js
```diff
@@ -19,24 +19,32 @@ const LookupTablesStore = Reflux.createStore({
       count: 0,
       query: null,
     };
+    this.errorStates = {
+      tables: {},
```
edmundoa: This should be the same data type as https://github.com/Graylog2/graylog2-server/pull/3795/files#diff-b4db24176b4bf6ff7554108b3773dc89R132, otherwise consumers of the store will get different data types in some scenarios.

graylog2-web-interface/src/stores/lookup-tables/LookupTablesStore.js
```diff
@@ -103,6 +112,34 @@ const LookupTablesStore = Reflux.createStore({
...
+      request.caches = cacheNames;
+    }
+    if (dataAdapterNames) {
+      request.data_adapters = dataAdapterNames;
+    }
+
+    const promise = fetch('POST', this._url('errorstates'), request);
+
+    promise.then((response) => {
+      this.trigger({
```
edmundoa: Shouldn't this be persisted in the store's state? If it doesn't we can remove the variable from the store scope and return the initial state object in `getInitialState()`.

graylog2-web-interface/src/pages/LUTCachesPage.jsx
```diff
@@ -22,16 +23,34 @@ const LUTCachesPage = React.createClass({
 
   mixins: [
     Reflux.connect(LookupTableCachesStore),
+    Reflux.connect(LookupTablesStore, 'tableStore'),
   ],
 
   componentDidMount() {
     this._loadData(this.props);
+
+    this.errorStatesTimer = setInterval(() => {
```
edmundoa: Shouldn't we avoid querying the server for error states until we can provide it? At least from the PR description I think we are still not ready for it and we can save some API calls until then.

graylog2-web-interface/src/components/lookup-tables/DataAdapterTableEntry.jsx
```diff
@@ -22,10 +29,18 @@ const DataAdapterTableEntry = React.createClass({
   },
 
   render() {
+    const errorOverlay = !this.props.error ? null : (
+      <Popover id="popover-table-error" title="Lookup Table problem" style={{ maxWidth: 400 }}>
```
edmundoa: These `Popovers` and `OverlayTrigger` are basically the same as the ones in `LUTTableEntry`. It would help extracting them into a component that could read an error state and render it.
I see this issue more as an improvement than a big issue, but I wanted to point it out anyway.

edmundoa: Somehow related: we have been trying to avoid inline styles for a while. Could you put them in a stylesheet so we can reuse them? In this case `LUTableEntry` would also benefit from it.

graylog2-web-interface/src/pages/LUTTablesPage.jsx
```diff
@@ -96,6 +115,13 @@ const LUTTablesPage = React.createClass({
             <span>Looking things up</span>
```
edmundoa: :-)

graylog2-web-interface/src/stores/lookup-tables/LookupTablesStore.js
kroepke: yup, it doesn't need to be in the store's object. I've removed the state.

kroepke:
@edmundoa It might be beneficial to not squash this merge because that should allow easier merging of #3803 (as that contains some of the commits from this one)

PR #3801, opened 2017-05-09
Add support for ES version-specific index mappings

joschi:
Index mappings changed quite a lot from Elasticsearch 2.x to Elasticsearch 5.x, so that we cannot simply use the same mapping for both (major) versions:

* https://www.elastic.co/guide/en/elasticsearch/reference/5.4/breaking_50_mapping_changes.html
* https://www.elastic.co/guide/en/elasticsearch/reference/5.4/mapping.html

This change set adds version-specific mappings for ES 2.x and 5.x which can be controlled by the `elasticsearch_version` configuration setting.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchClientConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchClientConfiguration.java
@@ -17,6 +17,8 @@
 package org.graylog2.configuration;
 
 import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.ValidatorMethod;
 import com.github.joschi.jadconfig.validators.PositiveIntegerValidator;
 import org.graylog2.configuration.converters.URIListConverter;
 import org.graylog2.configuration.validators.ListOfURIsWithHostAndSchemeValidator;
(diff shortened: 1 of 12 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchClientConfiguration.java
```diff
@@ -45,4 +47,46 @@
...
+        return elasticsearchIdleTimeout;
+    }
+
+    public int getMaxTotalConnections() {
+        return elasticsearchMaxTotalConnections;
+    }
+
+    public int getMaxTotalConnectionsPerRoute() {
+        return elasticsearchMaxTotalConnectionsPerRoute;
+    }
```
dennisoelkers: Why do we need all those getters when they are not being used anywhere? Currently we use named injection in all places, so these getters are unnecessary. We can introduce them when we really use them.

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping2.java
```diff
@@ -0,0 +1,76 @@
...
+
+        final Map<String, Object> defaultAll = ImmutableMap.of(
+                // Match all
+                "match", "*",
+                // Analyze nothing by default
+                "mapping", ImmutableMap.of("index", "not_analyzed"));
+        final Map<String, Map<String, Object>> templateAll = ImmutableMap.of("store_generic", defaultAll);
+
+        return ImmutableList.of(templateInternal, templateAll);
+    }
```
dennisoelkers: This method is an exact duplicate of the one in [`IndexMapping5`](https://github.com/Graylog2/graylog2-server/blob/es-http-index-mapping/graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping5.java#L34-L48). I would suggest we include it in the base class and make it abstract when we are really overriding it in a subclass.

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping5.java
```diff
@@ -0,0 +1,72 @@
...
+
+        final Map<String, Object> defaultAll = ImmutableMap.of(
+                // Match all
+                "match", "*",
+                // Analyze nothing by default
+                "mapping", ImmutableMap.of("index", "not_analyzed"));
+        final Map<String, Map<String, Object>> templateAll = ImmutableMap.of("store_generic", defaultAll);
+
+        return ImmutableList.of(templateInternal, templateAll);
+    }
```
dennisoelkers: This method is an exact duplicate of the one in [`IndexMapping2`](https://github.com/Graylog2/graylog2-server/blob/es-http-index-mapping/graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping2.java#L35-L49). I would suggest we include it in the base class and make it abstract when we are really overriding it in a subclass.

PR #3804, opened 2017-05-10
Throw FieldTypeException when aggregating on non-numeric fields.

dennisoelkers:
This change is now throwing a `FieldTypeException` instead of a generic
`ElasticsearchException` again, when search requests try to aggregate on
a non-numeric field. This allows us to retry getting non-numeric
statistics or show a more accurate error message when trying to generate
charts.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
@@ -90,6 +90,7 @@ public class JestUtils {
                     if (reason != null && reason.startsWith("Expected numeric type on field")) {
                         return buildFieldTypeException(errorMessage, reason);
                     }
+                    break;
             }
         }
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
@@ -23,6 +23,7 @@ import io.searchbox.client.JestResult;
 import io.searchbox.client.http.JestHttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.graylog2.indexer.ElasticsearchException;
+import org.graylog2.indexer.FieldTypeException;
 import org.graylog2.indexer.IndexNotFoundException;
 import org.graylog2.indexer.QueryParsingException;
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -642,6 +641,11 @@ public HistogramResult fieldHistogram(String query,
                 .filter(Objects::nonNull)
                 .collect(Collectors.toList());
 
+            final Optional<String> nonNumericFieldError = errors.stream().filter(error -> error.startsWith("Expected numeric type on field")).findAny();
+            if (nonNumericFieldError.isPresent()) {
```
joschi: Would it make sense to store all errors for this operation in the `errorDetails` field of the exception?

dennisoelkers: I am not sure if it would make sense, because it could potentially include other errors which are not field type related.

graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/JestUtils.java
```diff
@@ -84,6 +85,11 @@ private static ElasticsearchException specificException(Supplier<String> errorMe
                 case "index_not_found_exception":
                     final String indexName = asString(rootCause.get("resource.id"));
                     return buildIndexNotFoundException(errorMessage, indexName);
+                case "illegal_argument_exception":
+                    final String reason = asString(rootCause.get("reason"));
+                    if (reason != null && reason.startsWith("Expected numeric type on field")) {
+                        return buildFieldTypeException(errorMessage, reason);
+                    }
```
joschi: Technically, we're missing a `break` here, which is omitted for the other cases because they return in every case, but not the `illegal_argument_exception` case.

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
dennisoelkers: I also would like to move this potentially into `JestUtils.specificException` to have a generic error handling, but this is happening after the release.

joschi: True, but couldn't there also be more than 1 numeric field error?

PR #3830, opened 2017-05-16
implement componentWillReceiveProps to properly set form state

### LINKED ISSUE #3764: Missing state reset when switching between lookup caches/data adapter edits
## Current Behavior

When I edit one lookup cache or data adapter and then edit another one, state from the previous object is show in the form.

## Possible Solution

Properly reset state.

1. Create two lookup data adapters
2. Edit lookup adapter A
3. Edit lookup adapter B
4. The edit form for adapter B shows values from adapter A

## Your Environment

* Graylog Version: 2.3.0-alpha.2-SNAPSHOT

kroepke:
the new forms did not implement componentWillReceiveProps and hung on to the old state when re-rendered.

fixes #3764

### DIFF
--- a/graylog2-web-interface/src/components/lookup-tables/CacheForm.jsx
+++ b/graylog2-web-interface/src/components/lookup-tables/CacheForm.jsx
@@ -33,7 +33,15 @@ const CacheForm = React.createClass({
   },
 
   getInitialState() {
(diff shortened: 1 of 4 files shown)


kroepke:
@edmundoa indeed the error states are loaded too often now, there's #3834 for that.

But yes, I'll check for changes explicitly as that seems the correct way anyway.

Issue #3831, opened 2017-05-16
Lookup table entities should generate unique names


kroepke:
Auto generate intial 'name' property for lookup table, cache and adapters using lowercasing and special charactors to '_' conversion.
The name should also be checked on the backend to ensure that saving will work. Append a number if necessary to make it unique.

Only do this if no name has been set manually.

PR #3836, opened 2017-05-16
Add adapter and cache config validation

kroepke:
config classes can now optionally validate their settings, e.g.
checking for missing files and non-sensical settings to raise mistakes
during creation of instances

this also allows uniform checking for unique names for entities as well
as helping plugin authors to provide a better user experience (e.g.
checking that a backend HTTP service is reachable etc)

how much or how little is done depends on the plugin author

failed validation currently does not prevent saving the configuration,
but the UI should warn the user

also auto-generates the name property from the title unless it has been modified manually

#3831
#3835

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
@@ -437,7 +437,7 @@ public class LookupTableResource extends RestResource {
             //noinspection ConstantConditions
             if (!adapterDto.id().equals(toValidate.id())) {
                 // an adapter exists with a different id, so the name is already in use, fail validation
-                validation.addError("name", "The data adapter name must be unique.");
+                validation.addError("name", "The data adapter name is already in use.");
             }
         }
 
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
@@ -16,9 +16,11 @@
  */
 package org.graylog2.rest.resources.system.lookup;
 
+import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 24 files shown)


kroepke:
This is for initial review before adding the same mechanism to lookup tables and caches.

kroepke:
rebased to master

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -421,6 +423,30 @@ public DataAdapterApi updateAdapter(@Valid @ApiParam DataAdapterApi toUpdate) {
         return DataAdapterApi.fromDto(saved);
     }
 
+    @POST
+    @Path("adapters/validate")
+    @NoAuditEvent("Validation only")
+    @ApiOperation(value = "Validate the data adapter config")
+    public ValidationResult validateAdapter(@Valid @ApiParam DataAdapterApi toValidate) {
```
edmundoa: Editing a data adapter and leaving an empty `name` does not highlight the field as not valid. The response from the server in this case is a 400 instead of 200. I guess it's because hibernate is validating the `DataAdapterApi` and it can't find a value for `name`.

I think we can live with this, as the form is in the end validated by the browser and catches that issue, but I also guess is a bit confusing when you don't know how the code underneath works.

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
```diff
@@ -46,21 +53,54 @@ const DataAdapterForm = React.createClass({
...
+  },
+
   _onChange(event) {
     const dataAdapter = ObjectUtils.clone(this.state.dataAdapter);
     dataAdapter[event.target.name] = FormsUtils.getValueFromInput(event.target);
-    this.setState({ dataAdapter: dataAdapter });
+    let generateAdapterName = this.state.generateAdapterName;
+    if (generateAdapterName && event.target.name === 'title') {
+      // generate the name
+      dataAdapter.name = this._sanitizeTitle(dataAdapter.title);
```
kroepke: The title is solely for display purposes, the name is for referring to them in pipelines (and other places), correct.

The initial sanitizing is basically just a nicer way of helping the user to fill in that field, I think once the user manually changes it, we should keep it. It also really is just a string and can be anything, the regex is just to make  it  a bit more readable.

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
```diff
@@ -76,7 +116,31 @@ const DataAdapterForm = React.createClass({
       promise = LookupTableDataAdaptersActions.update(this.state.dataAdapter);
     }
 
-    promise.then(() => { this.props.saved(); });
+    promise.then(() => {
```
kroepke: Indeed!

kroepke: Failed validation might include somthing like a missing file, which should not prevent the entity to be created, I think. Other structural things will be returned as an exception, which handling is missing yet, correct.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
kroepke: Yes, I'll see if we can make this a bit more obvious by listing the possible return codes.

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
edmundoa: Ah, then it is fine as it is imo. We should mention that the name is meant to be used from other places like pipelines in the help text (just to clarify it).

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
edmundoa: ✅

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
edmundoa: Yes, anyway as I said, in the end the browser stops you from submitting that form, so it's just a bit of inconsistency :-)

bernd:
I am running into some issues when a validation fails.

## Cache
peek 2017-05-29 15-45

## Data Adapter
peek 2017-05-29 15-48

bernd:
The state issue does not show up when using this branch but only when rebasing against master. Something in master has changed that triggers this issue.

PR #3840, opened 2017-05-17
Simplify HTTP configuration

joschi:
## Motivation and Context

Users regularly have problems finding the correct settings for the Graylog REST API and Graylog web interface listeners in their environments.

This change set tries to simplify the network settings for the Graylog HTTP interfaces (Graylog REST API and web interface).

The two separate network listeners for the Graylog REST API and the Graylog web interface have been merged, vastly reducing the necessary configuration settings.

While the the Graylog REST API and the Graylog web interface could be configured separately via the `rest_listen_uri` and `web_listen_uri` settings in the past, they now run on the same network listener which can be configured via `http_bind_address`.

The context path for the Graylog REST API is now hard-coded to `/api`.

## Types of changes

## Checklist:


philicious:
while I never had bigger problems with it, these "duplicate" settings were indeed a little confusing and I like this change 👍

graylog2-server/src/main/java/org/graylog2/configuration/HttpConfiguration.java
```diff
@@ -0,0 +1,253 @@
...
+    URI getDefaultHttpUri() {
+        return getDefaultHttpUri("/");
+    }
+
+    @VisibleForTesting
+    URI getDefaultHttpUri(String path) {
+        final HostAndPort bindAddress = getHttpBindAddress();
+
+        final URI publishUri;
+        if (WILDCARD_IP_ADDRESS.equals(bindAddress.getHost())) {
```
kroepke: This currently does not take the IPv6 wildcard address into account and will publish `[::]` as the URI, which is invalid.
I think that adding a check against `[::]` and extending `Tools.guessPrimaryNetworkAddress()` to also look for IPv6 addresses, would be the best option here.

graylog2-server/src/main/java/org/graylog2/configuration/HttpConfiguration.java
```diff
@@ -0,0 +1,253 @@
...
+    public String getUriScheme(boolean enableTls) {
+        return enableTls ? "https" : "http";
+    }
+
+    public URI getHttpPublishUri() {
+        if (httpPublishUri == null) {
+            final URI defaultHttpUri = getDefaultHttpUri();
+            LOG.debug("No \"http_publish_uri\" set. Using default <{}>.", defaultHttpUri);
+            return defaultHttpUri;
+        } else if (WILDCARD_IP_ADDRESS.equals(httpPublishUri.getHost())) {
```
kroepke: This does not take the IPv6 wildcard address into account: `[::]`. See below for another instance of that.

kroepke:
With the latest changes I get the error message:
```
We are experiencing problems connecting to the Graylog server running on http://[fe80:0:0:0:b8dd:21ff:fe2f:35c0%veth3415fad]:9000/api/. Please verify that the server is healthy and working correctly.
```
When using `http_bind_address = [::]`.
Investigating.

joschi:
@kroepke I guess `fe80:0:0:0:b8dd:21ff:fe2f:35c0%veth3415fad` is simply the first IPv6 address of the first network interface in the system.

I'm not convinced that this approach (that worked reasonably well for IPv4) is suitable for IPv6. 🤔

kroepke:
Nevermind, this was due to automatically binding to a different interface. Works as expected.

kroepke:
Looks good, thanks for the rebase and implementing this!

PR #3862, opened 2017-05-23
Check Elasticsearch version when creating index template

joschi:
Instead of using a static configuration setting or trying to discover the Elasticsearch version
on startup (which might fail if Elasticsearch is unavailable), Graylog now checks the Elasticsearch
version when trying to create an index template and select the correct index mapping.

The "ping" request of Elasticsearch should be simple enough to not negatively influence the
Elasticsearch cluster and small enough to tolerate the additional network roundtrip.

This will fail if Elasticsearch is unavailable, but so would creating the index template.


dennisoelkers:
Could you please create a PR in the [documentation](https://github.com/Graylog2/documentation) repo reflecting the changes? Other than that: LGTM.

PR #3878, opened 2017-05-30
Enable search for lookup tables, data adapters and caches

### LINKED ISSUE #3843: Migrate server-side search query language to implement filtering
Currently only the enterprise codebase has the github-like query parser for server-side filtering of entities with pagination.

That code needs to be moved into graylog2-server to be able to implement the filtering for lookup tables and related entities.

bernd:
This makes the existing search forms for lookup tables, data adapters and caches work.

Closes #3843


kroepke:
rebased, waiting for builds

PR #3873, opened 2017-05-29
Refactor lifecycle for lookup adapters and caches

kroepke:
we use auxiliary servicemanagers to start lookup tables and use listeners to maintain the dependent services' state

https://github.com/Graylog2/graylog-plugin-map-widget/pull/48 needs to be merged once this one is in.

PR #3892, opened 2017-06-06
Add content pack import/export for lookup tables, caches and adapters

### LINKED ISSUE #3833: Lookup tables should be part of content packs
To allow sharing of configurations, lookup tables and related entities (data adapters and caches) should be part of content packs.

bernd:
This adds content pack import/export for lookup tables, caches and data adapters.

Closes #3833

### DIFF
--- a/graylog2-web-interface/src/components/common/ContentPackMarker.jsx
+++ b/graylog2-web-interface/src/components/common/ContentPackMarker.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+const ContentPackMarker = React.createClass({
+  propTypes: {
+    contentPack: React.PropTypes.string.isRequired,
+    marginLeft: React.PropTypes.number,
+    marginRight: React.PropTypes.number,
+  },
+
+  getDefaultProps() {
+    return { marginLeft: 0, marginRight: 0 };
+  },
+
+  render() {
+    const style = { marginLeft: this.props.marginLeft, marginRight: this.props.marginRight };
+
+    if (this.props.contentPack) {
+      return <i className="fa fa-cube" title="Created from content pack" style={style} />;
+    }
+
+    return null;
+  },
+});
+
+export default ContentPackMarker;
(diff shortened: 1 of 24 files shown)


kroepke:
needs change for moved/renamed classes in #3873

bernd:
@kroepke I adjusted the PR for the moved/renamed classes in #3873

graylog2-web-interface/src/components/common/ContentPackMarker.jsx
```diff
@@ -0,0 +1,25 @@
...
+    marginLeft: React.PropTypes.number,
+    marginRight: React.PropTypes.number,
+  },
+
+  getDefaultProps() {
+    return { marginLeft: 0, marginRight: 0 };
+  },
+
+  render() {
+    const style = { marginLeft: this.props.marginLeft, marginRight: this.props.marginRight };
```
kroepke: @edmundoa What do you think, would it be worthwhile to give ContentPackMarker either two components as props or two children to get rid of the inline styles?
Or is that overkill?

Also see comment below for an example.

graylog2-web-interface/src/components/lookup-tables/Cache.jsx
```diff
@@ -26,7 +27,12 @@ const Cache = React.createClass({
     return (
       <Row className="content">
         <Col md={6}>
-          <h2>{cache.title} <small>({plugin.displayName})</small></h2>
+          <h2>
+            {cache.title}
+            <ContentPackMarker contentPack={cache.content_pack} marginLeft={5} />
```
kroepke: Related to above: could we give it a left and right sibling prop instead of padding with spaces and inline styles?

edmundoa: In that case we would need a different component name, maybe something like `EntityTitle` or something like that, where the `ContentPackMarker` is just another thing to be included. Otherwise it would feel odd to me, having a `ContentPackMarker` component which takes the title and other information.

kroepke: After some discussion we've decided that we can postpone the creation of additional components for now.
It's really outside the scope of this feature and ideally we would restructure the content packs anyway to make them extensible.

PR #3894, opened 2017-06-07
Cleanup auto.offset.reset setting code in Kafka transport

bernd:
- Convert to a DropdownField because there are only two valid values
- Make sure to use the default if the config does not contain the
  setting yet (old inputs)

Refs #3743

/cc @r4um


r4um:
FYI these values will change again when moving to 0.10.x.x, are called latest, earliest, none.

PR #3895, opened 2017-06-07
Use preemptive authentication for Elasticsearch if credentials are given

dennisoelkers:
## Motivation and Context
This change enables preemptive authentication towards Elasticsearch
nodes whose URIs contain credentials. This way we save half of the
requests which are otherwise bounced back before authentication is
tried.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/providers/JestClientProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/providers/JestClientProvider.java
@@ -22,6 +22,7 @@ import com.google.gson.Gson;
 import io.searchbox.client.JestClient;
 import io.searchbox.client.JestClientFactory;
 import io.searchbox.client.config.HttpClientConfig;
+import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/bindings/providers/JestClientProvider.java
```diff
@@ -70,6 +74,10 @@ public JestClientProvider(@Named("elasticsearch_hosts") List<URI> elasticsearchH
                             new AuthScope(hostUri.getHost(), hostUri.getPort(), AuthScope.ANY_REALM, hostUri.getScheme()),
                             new UsernamePasswordCredentials(username, password)
                         );
+
+                        if (!Strings.isNullOrEmpty(username) || !Strings.isNullOrEmpty(password)) {
```
bernd: Shouldn't this be `&&` instead of `||` so we only add it as preemptive auth host if username **and** password are present?

dennisoelkers: No, both `username` and `password` could be empty, so we add it if at least one of them is present. Admittedly, using _only_ `username` or `password` for authentication is seldom, but still valid and possible.

PR #3896, opened 2017-06-07
Making Elasticsearch Version injectable.

dennisoelkers:
## Motivation and Context
This allows us to reuse the mechanism of retrieving the version of the
connected Elasticsearch nodes in other parts of the code besides the
IndexMappingFactory.

## Types of changes

## Checklist:


kroepke:
Ideally we would never do any IO in providers, because the user is not necessarily aware that the injection might fail (and it might fail multiple times, block indefinitely, or otherwise throw exceptions).

It is greatly preferable to inject something that gives you a service that can be used to retrieve the information, because then the user will see a TimeoutException etc being declared.

dennisoelkers:
Good points!

PR #3897, opened 2017-06-08
Use aliases for reopened indices

### LINKED ISSUE #3871: Re-open an index does not work with Elasticsearch 5.x
## Expected Behavior

The user should be able to re-open an index.

## Current Behavior

When I try to re-open a closed index, it does not work and the request sliently fails. (no UI feedback) The browser console shows this error:

```json
{
  "message": "Couldn't update settings of index graylog_20\n\nunknown setting [index.graylog2_reopened] please check that any required plugins are installed, or check the breaking changes documentation for removed settings",
  "details": [
    "unknown setting [index.graylog2_reopened] please check that any required plugins are installed, or check the breaking changes documentation for removed settings"
  ]
}
```

1. Run Graylog 2.3.0-alpha.3 with an Elasticsearch 5.x node
2. Close an index
3. Try to re-open the closed index

## Context

Graylog is using a custom index setting as an indicator for re-opened indices. (see [Indices.java](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java#L502-L515)) As far as I can see, custom index settings are not supported anymore in Elasticsearch 5.x.

Old indices which have been upgraded from Elasticsearch 2.x moved the `index.graylog2_reopened` setting into the `archived` namespace. (see [AbstractScopedSettings.java](https://github.com/elastic/elasticsearch/blob/5.4/core/src/main/java/org/elasticsearch/common/settings/AbstractScopedSettings.java#L544-L547) in the Elasticsearch source code)

That means Graylog cannot detect existing, re-opened indices anymore and cannot set the indicator when re-opening closed indices.

```json
{
  "graylog_17": {
    "settings": {
      "archived": {
        "index": {
          "graylog2_reopened": "true"
        }
      },
      "index": {
        "routing": {
          "allocation": {
            "total_shards_per_node": "2"
          }
        },
        "number_of_shards": "4",
        "blocks": {
          "write": "true",
          "metadata": "false",
          "read": "false"
        },
        "creation_date": "1465380847575",
        "analysis": {
          "analyzer": {
            "analyzer_keyword": {
              "filter": "lowercase",
              "tokenizer": "keyword"
            }
          }
        },
        "number_of_replicas": "0",
        "uuid": "Sv8V1DFZSpWZIV6LQIedEA",
        "version": {
          "created": "2030199",
          "upgraded": "5040099"
        }
      }
    }
  }
}
```

## Your Environment

* Graylog Version: 2.3.0-alpha.4-SNAPSHOT (7bd4d1482)
* Elasticsearch Version: 5.4.0

dennisoelkers:
## Motivation and Context

Before this change, a custom index setting (`graylog2_reopened`) was
used to flag reopened indices to be exempt from retention/rotation.
Starting with Elasticsearch 5.x, custom index settings are not supported
anymore.

This change now creates a special alias (`<index name>_reopened`)
pointing to the index flagging it as being reopened.

This change is downwards-compatible in that it includes a migration job that creates aliases for every index which is marked as reopened (either by using `graylog2_reopened` for 2.x or `archived.graylog2_reopened` for 5.x). There is no automated downgrade step which allows going back.

This change requires #3896 to be merged before.

Fixes #3871.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static org.graylog2.indexer.gson.GsonUtils.asJsonObject;
+import static org.graylog2.indexer.gson.GsonUtils.asString;
+
+public class V20170607164210_MigrateReopenedIndicesToAliases extends Migration {
```
kroepke: This class does not log anything, neither failures nor successes of its operation, making it very difficult to debug failures.

Also, while I appreciate the conciseness using streams and maps, they can be very difficult to debug, especially when they transform data which happens to be of an unexpected format, like it appears to be the case below. At the very least these methods should log something about their state when returning default values to aid debugging.

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
```diff
@@ -0,0 +1,141 @@
...
+        final JestResult jestResult = JestUtils.execute(jestClient, request, () -> "Couldn't read cluster state for reopened indices " + indices);
+        final JsonObject indicesJson = getClusterStateIndicesMetadata(jestResult.getJsonObject());
+        final ImmutableSet.Builder<String> reopenedIndices = ImmutableSet.builder();
+
+        for (Map.Entry<String, JsonElement> entry : indicesJson.entrySet()) {
+            final String indexName = entry.getKey();
+            final JsonElement value = entry.getValue();
+            if (value.isJsonObject()) {
+                final JsonObject indexSettingsJson = value.getAsJsonObject();
+                final JsonObject indexSettings = getIndexSettings(indexSettingsJson, indexName);
```
kroepke: Could this be replaced with a JsonPath expression instead of nested function calls?
Essentially we are trying to figure out whether one of  `settings.index.graylog2_reopened` or `archived.settings.index.graylog2_reopened` exists or not for each index we looked at.

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
```diff
@@ -0,0 +1,141 @@
...
+    private final Indices indices;
+    private final JestClient jestClient;
+
+    @Inject
+    public V20170607164210_MigrateReopenedIndicesToAliases(Node node,
+                                                           IndexSetService indexSetService,
+                                                           MongoIndexSet.Factory mongoIndexSetFactory,
+                                                           Indices indices,
+                                                           JestClient jestClient) {
+        this.elasticsearchVersion = node.getVersion().orElseThrow(() -> new ElasticsearchException("Unable to retrieve Elasticsearch version."));
```
dennisoelkers: This is true, but it was done on purpose, so we do not have to pass the version through the whole call chain or look it up every time.

Causing the server startup to fail in case of an error was an unintended side effect though. But, isn't this what we should do in case of a failing migration, because all the rest of the code expects it to succeed and there is no way to retry it besides restarting the server?

kroepke: Personally I would always avoid IO during injection unless absolutely necessary as in general it is difficult to reason about how often it might be performed. Could be that in this case it is known, but IMHO it is an anti-pattern.
I strongly prefer to pass the version through the call chain, or even putting it into a mutable field in this case, maybe with a small comment explaining why that's ok in this class.

kroepke: As for the failing: If it is intended to fail, I think we should do that in a more obvious way. I know this is not something in the scope of this PR though.
Since we don't have a good strategy to deal with intermittent connection issues during start, I would vote for ignoring it for now.
None of the migrations can be re-run manually, I think all of them require a restart.

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
dennisoelkers: ✅

graylog2-server/src/test/java/org/graylog2/indexer/IndexMappingFactoryTest.java
```diff
@@ -18,6 +18,7 @@
 
 import com.github.zafarkhaja.semver.Version;
 import org.graylog2.indexer.cluster.Node;
+import org.junit.Rule;
```
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
```diff
@@ -0,0 +1,138 @@
...
+            LOG.debug("Received cluster state was: {}", clusterStateJson.toString());
+            return Collections.emptySet();
+        }
+
+        for (Iterator<Map.Entry<String, JsonNode>> it = indicesJson.fields(); it.hasNext(); ) {
+            final Map.Entry<String, JsonNode> entry = it.next();
+            final String indexName = entry.getKey();
+            final JsonNode value = entry.getValue();
+
+            final JsonNode indexSettings = value.path("settings").path("index");
```
kroepke: Here we don't catch the `archived` section of the index settings in case we are upgrading.
I think we need to pass in the entire `value` into `checkForReopened` or simply inline that code and perform the missing settings check after the version check.

This happens when upgrading from 2.4.2 to 5.4.2 while migrating the reopen check.

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
```diff
@@ -0,0 +1,138 @@
...
+
+        return reopenedIndices.build();
+    }
+
+    private boolean checkForReopened(@Nullable JsonNode indexSettings, Version elasticsearchVersion) {
+        final JsonNode settings;
+        if (elasticsearchVersion.satisfies(">=2.1.0 & <5.0.0")) {
+            settings = indexSettings;
+        } else if (elasticsearchVersion.satisfies("^5.0.0")) {
+            settings = indexSettings.path("archived");
```
kroepke: As per the comment above, in case of upgrading to 5.x, the `indexSettings` here does not contain the `archived` subkey anymore, so the check comes back as `false`.

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/migrations/V20170607164210_MigrateReopenedIndicesToAliases.java
dennisoelkers: ✅ Good catch, thanks!

PR #3905, opened 2017-06-12
Fix and centralize lookup adapter/cache error handling

bernd:
- During adapter refresh, only clear the error once the refresh is done
- Force refresh when an error is set, even when the file info didn't change
- Make sure that `#startUp()` and `#shutDown()` never throw an error
- Run validation when opening adapter/cache form to show problems immediately

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
@@ -79,18 +79,13 @@ public class CSVFileDataAdapter extends LookupDataAdapter {
         if (isNullOrEmpty(config.path())) {
             throw new IllegalStateException("File path needs to be set");
         }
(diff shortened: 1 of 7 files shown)


kroepke:
For the HTTPJsonPath adapter I get a rather ugly message when trying to create it:

```
(400). Additional information: Null multiValueJSONPath
 at [Source: org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream@102f1aac; line: 1, column: 148] (through reference chain: org.graylog2.rest.models.system.lookup.$AutoValue_DataAdapterApi$Builder["config"]->org.graylog2.lookup.adapters.$AutoValue_HTTPJSONPathDataAdapter_Config$Builder["multi_value_jsonpath"])
```

bernd:
@kroepke Good catch, thanks! The forms should only validate on load when edited, new entities will always be invalid...

Fixed in 4b0f23d

PR #3911, opened 2017-06-14
Fix remove link placement on tag-remove elements

edmundoa:
Just a small fix of something I saw in the new filter used in the collector plugin.

### DIFF
--- a/graylog2-web-interface/public/stylesheets/graylog2.less
+++ b/graylog2-web-interface/public/stylesheets/graylog2.less
@@ -2661,7 +2661,7 @@ ul.pill-list > li {
   color: #333;
   cursor: pointer;
   margin-left: 5px;
-  vertical-align: middle;
+  vertical-align: text-bottom;
 }
 
 .tag-remove:before,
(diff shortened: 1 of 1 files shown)


mariussturm:
Works on all browsers but Chrome stable 59 (Version 59.0.3071.86 (Official Build) (64-bit)).
The 'X' is a little bit over the middle line of the text.

edmundoa:
As the change looks good in Chrome Canary, we are waiting a bit to see if the next Chrome version is not showing that issue. Otherwise we will work the PR a bit and try to fix the placement on all browsers.

bernd:
@edmundoa @mariussturm Any progress on this?

bernd:
@edmundoa @mariussturm ping

edmundoa:
@bernd @mariussturm this is looking good for me in Chrome 61.0.3163.91, so I think we can look into it again.

bernd:
@edmundoa Thank you!

edmundoa:
@mariussturm I think the latest code should fix the issue you were still seeing in Chrome for Linux. Please take a look and let me know if it still looks odd!

PR #3914, opened 2017-06-14
Add lookup table/cache/adapter permissions

bernd:
This adds custom permissions to lookup tables, caches and adapters. We decided to only use `lookuptables` for now, instead of separate permissions for tables, caches and adapters.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
@@ -31,6 +31,7 @@ import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import org.apache.shiro.authz.annotation.RequiresAuthentication;
+import org.apache.shiro.authz.annotation.RequiresPermissions;
 import org.graylog2.audit.AuditEventTypes;
 import org.graylog2.audit.jersey.AuditEvent;
 import org.graylog2.audit.jersey.NoAuditEvent;
@@ -63,6 +64,7 @@ import org.graylog2.search.SearchQuery;
 import org.graylog2.search.SearchQueryField;
 import org.graylog2.search.SearchQueryParser;
 import org.graylog2.shared.rest.resources.RestResource;
+import org.graylog2.shared.security.RestPermissions;
 import org.hibernate.validator.constraints.NotEmpty;
 import org.mongojack.DBQuery;
 import org.mongojack.DBSort;
@@ -176,6 +178,7 @@ public class LookupTableResource extends RestResource {
     @GET
     @Path("tables/{name}/query")
     @ApiOperation(value = "Query a lookup table")
+    @RequiresPermissions(RestPermissions.LOOKUP_TABLES_READ)
     public LookupResult performLookup(@ApiParam(name = "name") @PathParam("name") @NotEmpty String name,
                                       @ApiParam(name = "key") @QueryParam("key") @NotEmpty String key) {
         return lookupTableService.newBuilder().lookupTable(name).build().lookup(key);
@@ -184,6 +187,7 @@ public class LookupTableResource extends RestResource {
     @GET
     @Path("tables")
     @ApiOperation(value = "List configured lookup tables")
+    @RequiresPermissions(RestPermissions.LOOKUP_TABLES_READ)
     public LookupTablePage tables(@ApiParam(name = "page") @QueryParam("page") @DefaultValue("1") int page,
                                   @ApiParam(name = "per_page") @QueryParam("per_page") @DefaultValue("50") int perPage,
                                   @ApiParam(name = "sort",
@@ -243,6 +247,7 @@ public class LookupTableResource extends RestResource {
     public LookupTablePage get(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
                                @ApiParam(name = "resolve") @QueryParam("resolve") @DefaultValue("false") boolean resolveObjects) {
 
+        checkPermission(RestPermissions.LOOKUP_TABLES_READ, idOrName);
         Optional<LookupTableDto> lookupTableDto = dbTableService.get(idOrName);
         if (!lookupTableDto.isPresent()) {
             throw new NotFoundException();
@@ -270,6 +275,7 @@ public class LookupTableResource extends RestResource {
     @Path("tables")
     @AuditEvent(type = AuditEventTypes.LOOKUP_TABLE_CREATE)
     @ApiOperation(value = "Create a new lookup table")
+    @RequiresPermissions(RestPermissions.LOOKUP_TABLES_CREATE)
     public LookupTableApi createTable(@ApiParam LookupTableApi lookupTable) {
         try {
             LookupTableDto saved = dbTableService.save(lookupTable.toDto());
@@ -289,6 +295,7 @@ public class LookupTableResource extends RestResource {
     @ApiOperation(value = "Update the given lookup table")
     public LookupTableApi updateTable(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
                                       @Valid @ApiParam LookupTableApi toUpdate) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_EDIT, idOrName);
         LookupTableDto saved = dbTableService.save(toUpdate.toDto());
         clusterBus.post(LookupTablesUpdated.create(saved));
 
@@ -300,6 +307,8 @@ public class LookupTableResource extends RestResource {
     @AuditEvent(type = AuditEventTypes.LOOKUP_TABLE_DELETE)
     @ApiOperation(value = "Delete the lookup table")
     public LookupTableApi removeTable(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_DELETE, idOrName);
+
         // TODO validate that table isn't in use, how?
         Optional<LookupTableDto> lookupTableDto = dbTableService.get(idOrName);
         if (!lookupTableDto.isPresent()) {
@@ -346,6 +355,7 @@ public class LookupTableResource extends RestResource {
     @GET
     @Path("adapters")
     @ApiOperation(value = "List available data adapters")
+    @RequiresPermissions(RestPermissions.LOOKUP_TABLES_READ)
     public DataAdapterPage adapters(@ApiParam(name = "page") @QueryParam("page") @DefaultValue("1") int page,
                                 @ApiParam(name = "per_page") @QueryParam("per_page") @DefaultValue("50") int perPage,
                                 @ApiParam(name = "sort",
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -243,6 +247,7 @@ public LookupTablePage tables(@ApiParam(name = "page") @QueryParam("page") @Defa
     public LookupTablePage get(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
                                @ApiParam(name = "resolve") @QueryParam("resolve") @DefaultValue("false") boolean resolveObjects) {
 
+        checkPermission(RestPermissions.LOOKUP_TABLES_READ, idOrName);
```
kroepke: This (and the other references to `idOrName`) can become problematic, if we ever grant individual permissions, because then it is not clear which one to use when editing the permission: the id or the name.

I think it is better to load the DTO before and then check the permission on the id and never the names.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -284,10 +290,12 @@ public LookupTableApi createTable(@ApiParam LookupTableApi lookupTable) {
     }
 
     @PUT
-    @Path("tables")
+    @Path("tables/{idOrName}")
```
kroepke: Is this strictly required? I tried to avoid putting the ID in there twice,  because then we should also check that the ID in the payload is the same as in  the URL.

Also `idOrName` in the permission.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -284,10 +290,12 @@ public LookupTableApi createTable(@ApiParam LookupTableApi lookupTable) {
...
     @AuditEvent(type = AuditEventTypes.LOOKUP_TABLE_UPDATE)
     @ApiOperation(value = "Update the given lookup table")
-    public LookupTableApi updateTable(@Valid @ApiParam LookupTableApi toUpdate) {
+    public LookupTableApi updateTable(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
+                                      @Valid @ApiParam LookupTableApi toUpdate) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_EDIT, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -299,6 +307,8 @@ public LookupTableApi updateTable(@Valid @ApiParam LookupTableApi toUpdate) {
     @AuditEvent(type = AuditEventTypes.LOOKUP_TABLE_DELETE)
     @ApiOperation(value = "Delete the lookup table")
     public LookupTableApi removeTable(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_DELETE, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -423,6 +436,7 @@ public ErrorStates errorStates(@ApiParam(name = "request") @Valid ErrorStatesReq
     @Path("adapters/{idOrName}")
     @ApiOperation(value = "List the given data adapter")
     public DataAdapterApi getAdapter(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_READ, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -467,6 +483,7 @@ public DataAdapterApi createAdapter(@Valid @ApiParam DataAdapterApi newAdapter)
     @AuditEvent(type = AuditEventTypes.LOOKUP_ADAPTER_DELETE)
     @ApiOperation(value = "Delete the given data adapter", notes = "The data adapter cannot be in use by any lookup table, otherwise the request will fail.")
     public DataAdapterApi deleteAdapter(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_DELETE, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -483,10 +500,12 @@ public DataAdapterApi deleteAdapter(@ApiParam(name = "idOrName") @PathParam("idO
     }
 
     @PUT
-    @Path("adapters")
+    @Path("adapters/{idOrName}")
```
kroepke: path  param vs body

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -483,10 +500,12 @@ public DataAdapterApi deleteAdapter(@ApiParam(name = "idOrName") @PathParam("idO
...
     @AuditEvent(type = AuditEventTypes.LOOKUP_ADAPTER_UPDATE)
     @ApiOperation(value = "Update the given data adapter settings")
-    public DataAdapterApi updateAdapter(@Valid @ApiParam DataAdapterApi toUpdate) {
+    public DataAdapterApi updateAdapter(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
+                                        @Valid @ApiParam DataAdapterApi toUpdate) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_EDIT, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -588,6 +610,7 @@ public CachesPage caches(@ApiParam(name = "page") @QueryParam("page") @DefaultVa
     @Path("caches/{idOrName}")
     @ApiOperation(value = "List the given cache")
     public CacheApi getCache(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_READ, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -614,6 +638,7 @@ public CacheApi createCache(@ApiParam CacheApi newCache) {
     @AuditEvent(type = AuditEventTypes.LOOKUP_CACHE_DELETE)
     @ApiOperation(value = "Delete the given cache", notes = "The cache cannot be in use by any lookup table, otherwise the request will fail.")
     public CacheApi deleteCache(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_DELETE, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -630,10 +655,12 @@ public CacheApi deleteCache(@ApiParam(name = "idOrName") @PathParam("idOrName")
     }
 
     @PUT
-    @Path("caches")
+    @Path("caches/{idOrName}")
```
kroepke: path vs body

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -630,10 +655,12 @@ public CacheApi deleteCache(@ApiParam(name = "idOrName") @PathParam("idOrName")
...
     @AuditEvent(type = AuditEventTypes.LOOKUP_CACHE_UPDATE)
     @ApiOperation(value = "Update the given cache settings")
-    public CacheApi updateCache(@ApiParam CacheApi toUpdate) {
+    public CacheApi updateCache(@ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
+                                @ApiParam CacheApi toUpdate) {
+        checkPermission(RestPermissions.LOOKUP_TABLES_EDIT, idOrName);
```
kroepke: `idOrName` permission

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
bernd: @kroepke I added this for consistency with other resources in our system. We are using this everywhere, basically. When addressing a single resource, we put the ID in the URL. I think it's a good practice and common in many other HTTP APIs.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
bernd: Good point. I will change this to load the DTO and check against that. :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
kroepke: Ok fair enough!

kroepke: Then I'd say we should at least check the ids for equality to prevent mistakes when passing different ids.

bernd: > Then I'd say we should at least check the ids for equality to prevent mistakes when passing different ids.

Yes! I will do that. :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -173,9 +176,29 @@ public LookupTableResource(DBLookupTableService dbTableService,
         this.cacheSearchQueryParser = new SearchQueryParser(CacheDto.FIELD_TITLE, CACHE_SEARCH_FIELD_MAPPING);
     }
 
+    private void checkLookupTableId(String idOrName, LookupTableApi toUpdate) {
+        requireNonNull(idOrName, "idOrName parameter cannot be null");
+        if (!idOrName.equals(toUpdate.id()) || !idOrName.equals(toUpdate.name())) {
```
kroepke: There's still the edge case that the name changed and is used as the identifier for updating.
Our code doesn't do that, but I think we can live with it  for now.

If that's not the case, we can load the DTO first, and then check the old version of them.

bernd: Agreed. Changing the name will cause havoc anyway because we do not check if it's used...

PR #3917, opened 2017-06-15
Collect and show metrics for lookup caches and adapters

bernd:
Extend lookup caches and data adapters to collect runtime metrics and expose some of them in the UI.

This also fixes an issue with the ContentPackMarker component (6eef181) and the MetricComponent (4640593).

https://github.com/Graylog2/graylog-plugin-map-widget/pull/50 needs to be merged once this is merged.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
@@ -16,6 +16,9 @@
  */
 package org.graylog2.lookup.caches;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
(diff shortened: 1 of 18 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
```diff
@@ -198,4 +215,90 @@ public static Builder builder() {
             public abstract Config build();
         }
     }
+
+    private static class InstrumentedCache<K, V> implements Cache<K, V> {
```
kroepke: Same as https://github.com/Graylog2/graylog2-server/blob/4e8aace5b3ca7a9d19495ef1f1437e553e41d70c/graylog2-server/src/main/java/org/graylog2/metrics/CacheStatsSet.java#L29-L29 ?

kroepke: Or am I missing something here?

graylog2-web-interface/src/components/metrics/MetricsMapper.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import Reflux from 'reflux';
+//import { Row, Col } from 'react-bootstrap';
```
kroepke: left over?

graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
bernd: I actually didn't know about the CacheStatsSet...

The problem here is that this would work for the Guava based cache, but not for others. The cache metrics are defined in the LookupCache class so every cache gets the same metrics. We rely on those to show the count and hit rate in the UI.

graylog2-web-interface/src/components/metrics/MetricsMapper.jsx
bernd: Yes, will remove.

graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
kroepke: yeah let's stick to the generic one for now

PR #3925, opened 2017-06-20
Migrate to Jackson-based release of Jest 2.4.5

joschi:
## Description

Jackson turned out to be substantially faster for our use cases than Gson.

This PR migrates to our custom Jackson-based release of Jest 2.4.5 and adapts all existing Gson-related usages.


bernd:
@joschi The usage stats plugin also needs an update for this.

PR #3941, opened 2017-06-26
Not retrying in case of `isSuccessful()` being false.

### LINKED ISSUE #3938: Bulk indexer retries on "mapper_parsing_exception" error
## Current Behavior

Since we merged https://github.com/Graylog2/graylog2-server/pull/3929, the bulk indexer retries all errors because we trigger a retry if `BulkResult#isSucceeded` is `false`.

This also retries for errors like:

```json
{
  "type":"mapper_parsing_exception",
  "reason":"failed to parse [__woo2]",
  "caused_by":{
    "type":"number_format_exception",
    "reason":"For input string: \"foo\""
  }
} 
```

This would basically stop indexing into Elasticsearch if there is a message with a wrong data type. We cannot do that.

## Possible Solution

We either do not retry when `BulkResult#isSucceeded` is `false` or we go through all `BulkResult#getFailedItems()`, remove certain messages (might be configurable), and retry the rest.
This has several implications, we need to talk about this.

Not retrying at all would also throw away messages if we run into a `cluster_block_exception`. (as we did with Graylog <2.3)

1. Send message that has a number field `foo`
2. Send message that has a string field `foo`
3. Check server log, Graylog will retry forever

## Your Environment

* Graylog Version: 2.3.0-alpha.4-SNAPSHOT 0c4df85 
* Elasticsearch Version: 2.3.5

dennisoelkers:
If `isSuccessful()` returns false for a bulk indexing result, retrying
the whole bulk will lead to infinite loops in case of index mapping
errors or other permanent errors.

This change also adds an informational logging message whenever a bulk
was indexed successfully on the n-th (n > 1) attempt, indicating that
indexing has recovered from a formerly erroneous condition.

Fixes #3938.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -65,13 +65,14 @@ public class Messages {
     private static final Duration MAX_WAIT_TIME = Duration.seconds(30L);
     private static final Retryer<BulkResult> BULK_REQUEST_RETRYER = RetryerBuilder.<BulkResult>newBuilder()
             .retryIfException(t -> t instanceof IOException)
-            .retryIfResult((result) -> result == null || !result.isSucceeded())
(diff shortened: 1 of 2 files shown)


dennisoelkers:
Wait before merging, I am adding tests.

Issue #1632, opened 2015-12-14
Unable to send boolean values in Gelf


pdepaepe:
Hello,

Such call will show all fields in graylog

```
echo -ne '{"version": "1.1","host":"localhost","short_message":"ABC","full_message":"ABC","level":1,"_some_user_id":9001,"_some_info":"foo","_boolValueTrue":"true","_boolValueFalse":"false"}\0' | nc localhost 12203
```

When this will hide bool2ValueTrue/False fields

```
echo -ne '{"version": "1.1","host":"localhost","short_message":"ABC","full_message":"ABC","level":1,"_some_user_id":9001,"_some_info":"foo","_bool2ValueTrue":true,"_bool2ValueFalse":false}\0' | nc localhost 12203
```

joschi:
The GELF protocol currently (as of version 1.1) only allows string and numeric values, see https://www.graylog.org/resources/gelf/ for details.

Refs #668

PR #3958, opened 2017-06-30
Upgrade to gelfclient 1.4.2.1

joschi:
Older versions of gelfclient didn't handle boolean values correctly and only generated the Java type string (such as "[B@1234abc") instead of `true` or `false`.

Refs Graylog2/gelfclient#29

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <drools.version>6.5.0.Final</drools.version>
         <elasticsearch.version>2.4.4</elasticsearch.version>
         <jest.version>2.4.5+jackson</jest.version>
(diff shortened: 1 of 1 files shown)


bernd:
Our GELF codec does not support boolean values and will drop those fields from the message.

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java#L198-L213

Also the GELF spec does not support boolean values at the moment. (as you said [here](https://github.com/Graylog2/graylog2-server/issues/1632#issuecomment-164696290) :wink:)

joschi:
The GELF specification is a clusterf*ck…

Yeah, so let's close this.

PR #3972, opened 2017-07-03
Fail fast and loud for invalid GELF messages

### LINKED ISSUE #3970: GELF message with missing fields is dropped
## Expected Behavior
If a message send via *GELF* (TCP) is missing some specifications or a field that should not only be visible when switching the log to *debug*.

https://github.com/Graylog2/graylog2-server/blob/2.2.3/graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/DecodingProcessor.java#L179-L186

## Current Behavior
GELF Message missing a field - `short_message` for example the message silently disappear.

## Possible Solution
Change the logging level in case the decoding does not work.

1. Create GELF TCP Input 
2. send a message with whitespaced short message
```
echo -e '{"version": "1.1","host":"example.org","short_message":" ","full_message":"Backtrace here\n\nmore stuff","level":1,"_user_id":9001,"_some_info":"foo","_some_env_var":"bar"}\0' | nc -w 1 GRAYLOG_SERVER 12201
```
3.  To verify send a working message
```
echo -e '{"version": "1.1","host":"example.org","short_message":"Short","full_message":"Backtrace here\n\nmore stuff","level":1,"_user_id":9001,"_some_info":"foo","_some_env_var":"bar"}\0' | nc -w 1 GRAYLOG_SERVER 12201
```
4. Change log level to debug to see that the message with empty `short_message` is discarded

## Your Environment

* Graylog Version: 2.2.3

joschi:
Instead of waiting for a later stage and "silently" dropping (logged on DEBUG)
invalid messages, `GelfCodec` now actively checks for the existence and validity
of mandatory GELF message fields (such as "version", "host", "short_message", and
"timestamp", according to the GELF spec).

Refs http://docs.graylog.org/en/2.2/pages/gelf.html
Fixes #3970

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -228,13 +228,13 @@ public class GelfCodec extends AbstractCodec {
         final JsonNode hostNode = jsonNode.path("host");
         if (hostNode.isMissingNode()) {
             log.warn(prefix + "is missing mandatory \"host\" field.");
-            return;
-        }
-        if (!hostNode.isTextual()) {
-            throw new IllegalArgumentException(prefix + "has invalid \"host\": " + hostNode.asText());
-        }
-        if (StringUtils.isBlank(hostNode.asText())) {
-            throw new IllegalArgumentException(prefix + "has empty mandatory \"host\" field.");
+        } else {
+            if (!hostNode.isTextual()) {
+                throw new IllegalArgumentException(prefix + "has invalid \"host\": " + hostNode.asText());
+            }
+            if (StringUtils.isBlank(hostNode.asText())) {
+                throw new IllegalArgumentException(prefix + "has empty mandatory \"host\" field.");
+            }
         }
 
         final JsonNode shortMessageNode = jsonNode.path("short_message");
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -221,14 +221,6 @@ public class GelfCodec extends AbstractCodec {
     }
 
     private void validateGELFMessage(JsonNode jsonNode) {
-        final JsonNode versionNode = jsonNode.path("version");
-        if (versionNode.isMissingNode()) {
-            throw new IllegalArgumentException("GELF message is missing mandatory \"version\" field.");
-        }
-        final String version = versionNode.asText();
-        if (!"1.1".equals(version)) {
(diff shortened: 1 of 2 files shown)


joschi:
@bernd An alternative would be to allow messages with *any* fields in `DecodingProcessor`, so that invalid (GELF) message, e. g. the ones with missing "short_message" or "message" field respectively, won't be dropped at a later stage.

Anyway, we should clarify this in the GELF "specification" in the documentation because right now it's non-obvious (and impossible to find out without reading code of each respective Graylog release) to know why a message was discarded.

bernd:
@joschi Can't we use the check you implemented for the `short_message` and `message` field? That would already help #3970, no?

I am not sure about adjusting the GELF spec. The problem is that we did a really bad implementation of the spec. (the irony) Telling people that our implementation accepts basically everything as long as a `short_message` or `message` field exists might motivate people to actually do that instead of following the spec. :wink: IMHO this makes the problem worse for them in the future when we have GELF 2.0 and do a proper implementation. I would rather use the check and log message you implemented to check for the `short_message` and `message` field. (as mentioned above)

bernd:
@joschi Any comments?

graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
```diff
@@ -218,6 +222,47 @@ public Message decode(@Nonnull final RawMessage rawMessage) {
         return message;
     }
 
+    private void validateGELFMessage(JsonNode jsonNode, UUID id, ResolvableInetSocketAddress remoteAddress) {
+        final String prefix = "GELF message <" + id + "> " + (remoteAddress == null ? "" : "(received from <" + remoteAddress + ">) ");
+
+        final JsonNode hostNode = jsonNode.path("host");
+        if (hostNode.isMissingNode()) {
+            log.warn(prefix + "is missing mandatory \"host\" field.");
+            return;
```
bernd: Shouldn't we check for the `short_message` / `message` fields first before returning? As it is now, a missing `host` field would skip the `short_message` / `message` field checks.

joschi: Correct, this is missing the alternative case.

bernd:
@joschi Thank you! :+1:

NicholasMeacoe:
Hi, when trying out the RC1 build noticed this was now stopping pretty much all of the traffic shipped over UDP via Gelf4Net.  The datetime object it passes over contains the timezone, and the code change to only accept a number kicks the message out.  If we get Gelf4Net changed, will we need to send in another field to indicate the timezone?  We have packets coming over the wire from UK, US and APAC.  

Cheers,
Nick

joschi:
@NicholasMeacoe This has been addressed in #4027/#4031/#4034.

PR #3974, opened 2017-07-04
Add support for boolean values in GelfCodec

joschi:
This PR should serve as a base for discussion for extending the official GELF specification to support boolean values.

`GelfCodec` (and the GELF message format as specified by http://docs.graylog.org/en/2.2/pages/gelf.html) currently don't support boolean values.

Since the payload format of GELF messages is just JSON with some mandatory fields, there's no reason *not* to support booleans.

Graylog itself has no problems whatsoever in processing boolean values and other message inputs, as well as extractors and the processing pipelines, already do support boolean values.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -207,6 +207,8 @@ public class GelfCodec extends AbstractCodec {
                 continue;
             } else if (value.isTextual()) {
                 fieldValue = value.asText();
+            } else if (value.isBoolean()) {
+                fieldValue = value.asBoolean();
             } else {
                 log.debug("Field [{}] has unknown value type. Skipping.", key);
                 continue;
--- a/graylog2-server/src/test/java/org/graylog2/inputs/codecs/GelfCodecTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/inputs/codecs/GelfCodecTest.java
@@ -94,6 +94,34 @@ public class GelfCodecTest {
     }
 
     @Test
+    public void decodeSupportsAllDataTypes() throws Exception {
+        final String json = "{"
+                + "\"version\": \"1.1\","
+                + "\"host\": \"example.org\","
+                + "\"short_message\": \"A short message that helps you identify what is going on\","
+                + "\"_boolean\": true,"
+                + "\"_long\": 42,"
+                + "\"_double\": 3.14,"
+                + "\"_string\": \"Foobar\","
+                + "\"_map\": {\"key\":\"value\"},"
+                + "\"_array\": [\"val1\",\"val2\"]"
+                + "}";
+
+        final RawMessage rawMessage = new RawMessage(json.getBytes(StandardCharsets.UTF_8));
+        final Message message = codec.decode(rawMessage);
+
+        assertThat(message).isNotNull();
+        assertThat(message.getField("boolean")).isEqualTo(true);
+        assertThat(message.getField("long")).isEqualTo(42L);
+        assertThat(message.getField("double")).isEqualTo(3.14D);
+        assertThat(message.getField("string")).isEqualTo("Foobar");
+
+        // Container nodes are converted into their JSON string equivalent
+        assertThat(message.getField("map")).isEqualTo("{\"key\":\"value\"}");
+        assertThat(message.getField("array")).isEqualTo("[\"val1\",\"val2\"]");
+    }
+
+    @Test
     public void decodeBuildsValidMessageObject() throws Exception {
         final String json = "{"
                 + "\"version\": \"1.1\","


joschi:
Refs Graylog2/gelfclient#30, Graylog2/gelfclient#31

dennisoelkers:
This PR is now 9 months old. How to proceed with it? The code looks good to me and is very small, so there should be no problem merging it in `3.0`. It would also be a good first step for including boolean values in what could be GELF 1.2/2.0/....

joschi:
@dennisoelkers The problem is that it would lead to mapping exceptions if people have been using boolean values before, so there's no (soft) migration path for this change.

dennisoelkers:
Got it. Couldn't that be mitigated by documenting it (suggesting a deflector cycle, documenting the implications for search queries) in the upgrade notes to `3.0`? Having booleans handled properly and not just as strings is an improvement imo.

joschi:
> Having booleans handled properly and not just as strings is an improvement imo.

In my opinion as well but there seems to be relatively little interest in this particular change right now.

Up for discussion with @Graylog2/engineers.

bernd:
@dennisoelkers @joschi If we want to do this we have to carefully think about the implications. Apart from clearly documenting the change we probably have to offer migration paths for existing data. Otherwise old queries and alerts will break.

I think this is a lot of work and risk for the little benefit the user will get from this change.

An alternative could be to cut a new GELF version with this (and maybe other) change(s) and implement a codec for each GELF version. GELF inputs can then select the version that should be used. That way we don't break existing setups but users can still use new GELF versions and features.

joschi:
Closing due to lack of discussion and interest in the PR.

bernd:
@dennisoelkers Thanks for picking this up again! :+1: Good to see that this isn't a problem on the server side.

I think we can still run into problems with backwards compatibility when we and others start to update client libraries like [gelfclient](https://github.com/Graylog2/gelfclient/blob/master/src/main/java/org/graylog2/gelfclient/encoder/GelfMessageJsonEncoder.java#L97-L104) to support boolean values. That could result in mapping exceptions when the same field gets sent as `boolean` instead of `String` after a client library update.
That's why I am still in favor of creating a new GELF version so users can choose what version to accept (and also send from clients) and to keep backwards compatibility. (as written [here](https://github.com/Graylog2/graylog2-server/pull/3974#issuecomment-369315222))

danotorrey:
@bernd Will this really result in mapping exceptions since boolean fields are currently being dropped (as described by @dennisoelkers here https://github.com/Graylog2/graylog2-server/pull/3974#issuecomment-454305974)? I believe new boolean fields would just start coming through that don't exist in the index.

This codec was copied in the Forwarder, but I am revisiting this to see if we can close up this PR.

bernd:
@danotorrey The problem is with the client libraries. Our gelfclient for example is converting boolean values to a string. If we would just change the client to send booleans instead of strings, the user would see mapping errors when indexing messages. Because the same fields now gets sent as boolean (and now accepted by the server) where they have previously been sent as string instead.

See: [GelfMessageJsonEncoder.java](https://github.com/Graylog2/gelfclient/blob/87cf261e81addaf4e8c93f75020def0c32f9e69f/src/main/java/org/graylog2/gelfclient/encoder/GelfMessageJsonEncoder.java#L97-L104) in https://github.com/Graylog2/gelfclient.

```java
                if (field.getValue() instanceof Number) {
                    // Let Jackson figure out how to write Number values.
                    jg.writeObjectField(realKey, field.getValue());
                } else if (field.getValue() == null) {
                    jg.writeNullField(realKey);
                } else {
                    jg.writeStringField(realKey, field.getValue().toString());
                }
```

And that is only our own client. There are other 3rd party client implementations we don't control so we can't just change this, unfortunately. Luckily we have versioning capabilities in GELF. :smile: 

Releasing a new GELF version (1.2) that supports booleans would solve this issue. The clients can continue to send version 1.1 by default and the server is choosing a codec based on the version. If users want to send boolean values, they have to explicitly switch to the new GELF version in their client.
With the next major client release we could change the default to the new GELF version.

danotorrey:
@bernd Ooo I see. The encoding side is sending booleans as strings. Makes sense. If the encoder emits as a string, then we can't change it in the same version without breaking 3rd party usages. 😀 I was thinking of the Graylog Server implementation of GelfCodec, which ignores boolean values.

https://github.com/Graylog2/graylog2-server/blob/50b565dcead6e0a372236d5c2f8530dc5726fa9b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java#L207

Why is GelfCodec in the `Graylog2-server` repo instead of `gelfclient`?

Can I go ahead and transfer this issue to the `gelfclient` repo and proceed with this change?

bernd:
@danotorrey 

> Why is GelfCodec in the `Graylog2-server` repo instead of `gelfclient`? Is this because client implementations do not belong in the `gelfclient` repo?

Because the gelfclient library is younger than the server. :wink: I think we should keep it that way for now. We can discuss if we want to move all GELF de/serializing code into the gelfclient library in the future.
 
> Can I go ahead and transfer this issue to the `gelfclient` repo and proceed with this change? We could target this for 3.1.

If "proceed with this change" means to create a new GELF codec version with this PR, then yes. :smiley:  We can create an issue in the gelfclient to support the new GELF version. But that can only happen after the server supports the new GELF version.

PR #4013, opened 2017-07-17
Issue 4003

### LINKED ISSUE #4003: ExceptionUtils#getRootCauseMessage can NPE
When constructing the root cause message, the Exception's message can be null which leads to an NPE.

kroepke:
Guard against NPE exception in ExceptionUtils#formatMessageCause

fixes #4003

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/shared/utilities/ExceptionUtilsTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/shared/utilities/ExceptionUtilsTest.java
@@ -1,0 +1,36 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog2/shared/utilities/ExceptionUtilsTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class ExceptionUtilsTest {
+    @Test
+    public void formatMessageCause() throws Exception {
+
+        final NullPointerException npe = new NullPointerException();
+
+        try {
+            assertThat(ExceptionUtils.formatMessageCause(npe)).isNotBlank();
+        } catch (Exception e) {
+            fail(e.getMessage());
```
joschi: Random nitpick: The test fail anyway if an exception was thrown and not caught in `ExceptionUtils#formatMessageCause(…)`.

PR #4038, opened 2017-07-27
Return NaN for non-present fields of FieldStatsResult.

### LINKED ISSUE #4026: NPE computing statistical widget
After the Graylog server has been stopped for a while, I'm getting a NPE when computing a statistical widget:

```
java.lang.NullPointerException: null
	at org.graylog2.indexer.results.FieldStatsResult.<init>(FieldStatsResult.java:50) ~[classes/:?]
	at org.graylog2.indexer.searches.Searches.fieldStats(Searches.java:508) ~[classes/:?]
	at org.graylog2.dashboards.widgets.strategies.StatisticalCountWidgetStrategy.compute(StatisticalCountWidgetStrategy.java:150) ~[classes/:?]
	at org.graylog2.dashboards.widgets.WidgetResultCache$ComputationResultSupplier.get(WidgetResultCache.java:89) ~[classes/:?]
	at org.graylog2.dashboards.widgets.WidgetResultCache$ComputationResultSupplier.get(WidgetResultCache.java:73) ~[classes/:?]
	at com.google.common.base.Suppliers$ExpiringMemoizingSupplier.get(Suppliers.java:227) ~[guava-21.0.jar:?]
	at org.graylog2.dashboards.widgets.WidgetResultCache.getComputationResultForDashboardWidget(WidgetResultCache.java:65) ~[classes/:?]
```

Looking into the value of the `ExtendedStatsAggregation` object that raises the exception, this is what I see:
screen shot 2017-07-21 at 12 00 16

This is the widget that the code tries to compute, as seen in MongoDB:
```
{
  "creator_user_id": "hans", 
  "cache_time": 10,
  "description": "Statistical value",
  "id": "773a22ce-3c17-467e-86cf-ac0f409702e7",
  "type": "STATS_COUNT",
  "config": {
    "timerange": {
      "type": "relative",
      "range": 300
    },
    "field": "took_ms",
    "trend": true,
    "query": "",
    "stats_function": "std_deviation",
    "lower_is_better": true
  }
}
```

## Expected Behavior
Computing statistical widgets should work, even if there are no results because there's no data.

## Current Behavior
Computing statistical widgets does not work when there's no data available. At least that's what I read from the results I see.

1. Stop the server for double the time the widget is querying (ensuring there's no results for the statistical value)
2. Start the server and try to get a computation of the widget value
3. See NPE

## Your Environment
* Graylog Version: 2.4.0-SNAPSHOT, but probably also present in 2.3.0-rc.2-SNAPSHOT
* Elasticsearch Version: 2.4.4
* Operating System: macOS Sierra

dennisoelkers:
Before this change, an ExtendedStatsAggregation could include an
arbitrary number of fields that are null. Assigning them a non-boxed
field type leads to an NPE and a 500 is being returned to the caller
when the result of a extended field stats widget is requested.

This change properly assigns a valid value for those fields, so a result
(albeit possibly containing NaN for one or more fields) is being
returned to the caller.

Fixes #4026.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
@@ -45,7 +45,7 @@ public class FieldStatsResult extends IndexQueryResult {
                             long tookMs) {
         super(query, source, tookMs);
         this.count = getValueCount(valueCountAggregation, extendedStatsAggregation);
-        this.cardinality = cardinalityAggregation == null ? Long.MIN_VALUE : cardinalityAggregation.getCardinality();
+        this.cardinality = cardinalityAggregation == null || cardinalityAggregation.getCardinality() == null ? Long.MIN_VALUE : cardinalityAggregation.getCardinality();
 
         if (extendedStatsAggregation != null) {
             sum = Optional.ofNullable(extendedStatsAggregation.getSum()).orElse(Double.NaN);
@@ -83,9 +83,9 @@ public class FieldStatsResult extends IndexQueryResult {
     }
 
     private long getValueCount(ValueCountAggregation valueCountAggregation, ExtendedStatsAggregation extendedStatsAggregation) {
-        if (valueCountAggregation != null) {
+        if (valueCountAggregation != null && valueCountAggregation.getValueCount() != null) {
             return valueCountAggregation.getValueCount();
-        } else if (extendedStatsAggregation != null) {
+        } else if (extendedStatsAggregation != null && extendedStatsAggregation.getCount() != null) {
             return extendedStatsAggregation.getCount();
         }
         return Long.MIN_VALUE;
--- a/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
@@ -21,6 +21,7 @@ import io.searchbox.core.search.aggregation.ExtendedStatsAggregation;
 import io.searchbox.core.search.aggregation.ValueCountAggregation;
 
 import java.util.List;
+import java.util.Optional;
 
 public class FieldStatsResult extends IndexQueryResult {
     private List<ResultMessage> searchHits;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
```diff
@@ -47,13 +48,13 @@ public FieldStatsResult(ValueCountAggregation valueCountAggregation,
...
-            max = extendedStatsAggregation.getMax();
-            variance = extendedStatsAggregation.getVariance();
-            stdDeviation = extendedStatsAggregation.getStdDeviation();
+            sum = Optional.ofNullable(extendedStatsAggregation.getSum()).orElse(Double.NaN);
+            sumOfSquares = Optional.ofNullable(extendedStatsAggregation.getSumOfSquares()).orElse(Double.NaN);
+            mean = Optional.ofNullable(extendedStatsAggregation.getAvg()).orElse(Double.NaN);
+            min = Optional.ofNullable(extendedStatsAggregation.getMin()).orElse(Double.NaN);
+            max = Optional.ofNullable(extendedStatsAggregation.getMax()).orElse(Double.NaN);
+            variance = Optional.ofNullable(extendedStatsAggregation.getVariance()).orElse(Double.NaN);
+            stdDeviation = Optional.ofNullable(extendedStatsAggregation.getStdDeviation()).orElse(Double.NaN);
```
joschi: This is kind of an abuse for the `Optional<T>` type. Please use Guava's [`MoreObjects.firstNonNull(...)`](https://google.github.io/guava/releases/22.0/api/docs/com/google/common/base/MoreObjects.html#firstNonNull-T-T-) or alternatively (without a 3rd party library) the usual `x == null ? Double.NaN : x` idiom.

dennisoelkers: I prefer this one, because it neither requires a 3rd party library nor is it as verbose as the ternary operator in this case. What are the downsides/the "abuse" of using it?

joschi: The downside is the additional creation of the `Optional<T>` objects which isn't necessary when using one of the two alternatives.

Also, why is the first snippet less verbose than the second?
```java
min = Optional.ofNullable(extendedStatsAggregation.getMin()).orElse(Double.NaN);
```
more verbose than
```java
min = firstNonNull(extendedStatsAggregation.getMin(), Double.NaN);
```

graylog2-server/src/test/java/org/graylog2/indexer/results/FieldStatsResultTest.java
```diff
@@ -0,0 +1,47 @@
+package org.graylog2.indexer.results;
+
+import edu.emory.mathcs.backport.java.util.Collections;
```
joschi: Wrong import.

graylog2-server/src/main/java/org/graylog2/indexer/results/FieldStatsResult.java
dennisoelkers: > nor is it as verbose as the _ternary operator_ in this case.

`min = Optional.ofNullable(extendedStatsAggregation.getMin()).orElse(Double.NaN);`
`min = extendedStatsAggregation.getMin() == null ? Double.NaN : extendedStatsAggregation.getMin();`

dennisoelkers: You are correct regarding the additional object instantiation, though. Will fix.

PR #4053, opened 2017-08-03
Allow empty/blank "short_message" field if "message" exists

joschi:
Because every fucking GELF library seems to be broken. 😠


joschi:
This should be backported to the `2.3` branch.

PR #4057, opened 2017-08-03
Updating to React v15.6.1.

dennisoelkers:
This change is bumping React to the most versions. General functionality was tested, there should be no semantical changes. We have new warnings in the console (mostly about https://fb.me/react-warning-dont-call-proptypes) which will be remediated in later pull requests.


edmundoa:
I also see lots of deprecations being logged for `isMounted`. After #4059, I think that should be the next thing to tackle.

PR #4059, opened 2017-08-04
Correcting usage of PropTypes, switching to new separate package.

dennisoelkers:
## Motivation and Context

PropTypes have moved into a separate package (prop-types), so all
imports and usages of the old way (being part of the React module) are
changed.

For more details see
https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes

This change was fabricated using a codemod
(https://github.com/reactjs/react-codemod/blob/master/transforms/React-PropTypes-to-prop-types.js)
which was executed like this:

`jscodeshift -t transforms/React-PropTypes-to-prop-types.js graylog2-server/graylog2-web-interface/src/**/*.jsx`

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/packages/graylog-web-plugin/package.json
+++ b/graylog2-web-interface/packages/graylog-web-plugin/package.json
@@ -33,6 +33,7 @@
     "jquery": "2.1.x",
     "moment": "2.14.1",
     "moment-timezone": "0.5.7",
+    "prop-types": "^15.5.10",
     "react": "^15.6.1",
     "react-addons-linked-state-mixin": "^15.6.0",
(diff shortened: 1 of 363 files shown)


edmundoa:
Other than the remark above, the changes look good to me.

We also need to do this replacement in all plugins. Should I add issues for that, @dennisoelkers?

dennisoelkers:
Absolutely correct, good catch!

dennisoelkers:
Regarding the changes for the plugins: If you have the time, I would be happy if you could add issues, otherwise I can do it too.

edmundoa:
I added issues in all affected plugins and referenced to this one to get to them easily.

PR #4062, opened 2017-08-07
Return index set wildcard instead of individual indices in Searches#termineAffectedIndices()

### LINKED ISSUE #4054: Search query fails with large number of indices
Graylog (or rather Jest) is sending HTTP requests with a large initial line (URI path and query string) to the Elasticsearch HTTP API if a large number of indices is included in the search query (e. g. when searching in "All messages").

Related topic: https://community.graylog.org/t/unable-to-search-in-all-messages/1922

## Expected Behavior

Search queries covering a lot of indices should work.

## Current Behavior

Search queries covering a lot of indices fail with an internal server error (HTTP 500) and produce an error message in the Elasticsearch logs:

```
[WARN ][http.netty               ] [ElasticsearchNodeName] Caught exception while handling client http traffic, closing connection [id: 0xecc07e39, /10.1.2.3:54321 => /10.1.2.3:9200]
org.jboss.netty.handler.codec.frame.TooLongFrameException: An HTTP line is larger than 4096 bytes.
```

## Possible Solution

Patch Jest to send index names in the POST body.

1. Create lots of indices (so that the list of index names is longer than 4 KB)
2. Run search query covering all indices
3. ???
4. Profit!

## Your Environment

* Graylog Version: 2.3.0
* Elasticsearch Version: 2.x, 5.x

joschi:
This PR changes the behavior of the `Searches` class to use a [Terms Query](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/query-dsl-terms-query.html) filtering for the `_index` field instead of listing all indices in the URI path if a defined number of indices (see `Searches.MAX_INDICES_PER_QUERY`) has been reached.

While this leads to more load required on the side of Elasticsearch (because all index shards have to be visited instead of just the ones of the specified indices), it circumvents the problem with the URI being larger than 4 KB.

Using partitioned parallel queries when the maximum URI size has been reached would unfortunately only be feasible for a subset of queries currently supported by the `Searches` class. 😞 

Fixes #4054

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -21,6 +21,7 @@ import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSortedSet;
 import io.searchbox.client.JestClient;
 import io.searchbox.client.JestResult;
@@ -50,6 +51,7 @@ import org.graylog2.indexer.FieldTypeException;
 import org.graylog2.indexer.IndexHelper;
 import org.graylog2.indexer.IndexMapping;
 import org.graylog2.indexer.IndexSet;
+import org.graylog2.indexer.IndexSetRegistry;
 import org.graylog2.indexer.cluster.jest.JestUtils;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.indexer.ranges.IndexRange;
@@ -86,8 +88,8 @@ import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 import static com.codahale.metrics.MetricRegistry.name;
-import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.util.Objects.requireNonNull;
 import static org.elasticsearch.index.query.QueryBuilders.matchAllQuery;
 import static org.elasticsearch.index.query.QueryBuilders.queryStringQuery;
 
@@ -102,6 +104,9 @@ public class Searches {
     public static final String AGG_CARDINALITY = "gl2_field_cardinality";
     public static final String AGG_VALUE_COUNT = "gl2_value_count";
     private static final Pattern filterStreamIdPattern = Pattern.compile("^(.+[^\\p{Alnum}])?streams:([\\p{XDigit}]+)");
+
+    @VisibleForTesting
+    static final int MAX_INDICES_PER_QUERY = 100;
 
     public enum TermsStatsOrder {
         TERM,
@@ -161,6 +166,7 @@ public class Searches {
 
     private final Configuration configuration;
     private final IndexRangeService indexRangeService;
+    private final IndexSetRegistry indexSetRegistry;
     private final Timer esRequestTimer;
     private final Histogram esTimeRangeHistogram;
     private final Counter esTotalSearchesCounter;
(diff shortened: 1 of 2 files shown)


bernd:
@joschi Thanks for the PR!

One thing I don't understand yet: Why do we need the terms query which the index list? We are already using a timerange filter in the query, how does the terms query improve things?

joschi:
> Why do we need the terms query with the index list? We are already using a timerange filter in the query, how does the terms query improve things?

Since users can have multiple index sets which contain distinct data sets, specifying the time range alone isn't sufficient.

Just think of a search query covering firewall and network appliance logs in a number of index sets which shouldn't include application logs from the same time frame.

bernd:
> Just think of a search query covering firewall and network appliance logs in a number of index sets which shouldn't include application logs from the same time frame.

Sorry, I forgot to mention that we also filter on streams. Index sets are bound to streams so this will be handled as far as I can see. Currently, I can search in either one or all streams. If I search in one stream, we have a stream filter in addition to the timerange. If I search in all streams, it's okay to look into all index sets.

Not sure if there is another scenario where selecting the indices is important. If not, we can greatly simply the code (as far as I can see) and save some cpu cycles processing the query.

hc4:
@bernd  agree.
Just use pattern as index name: `graylog2_*`
E.g. that is what doing Grafana in it's requests to ES. And ES handles such queries quiet fast.
No sense in making Terms query, just more work for ES.

joschi:
@bernd I've implemented a variant returning the write index aliases instead of using a terms filter.

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -805,16 +813,27 @@ private QueryBuilder standardAggregationFilters(TimeRange range, String filter)
...
         final Set<IndexRange> indexRanges = determineAffectedIndicesWithRanges(range, filter);
-        return indexRanges.stream()
-                .map(IndexRange::indexName)
-                .collect(Collectors.toSet());
+        if(indexRanges.size() > MAX_INDICES_PER_QUERY) {
+            return indexRanges.stream()
+                    .map(IndexRange::indexName)
+                    .map(indexSetRegistry::getForIndex)
+                    .flatMap(o -> o.map(java.util.stream.Stream::of).orElseGet(java.util.stream.Stream::empty))
+                    .map(IndexSet::getWriteIndexAlias)
```
bernd: I guess you wanted to use `IndexSet#getIndexWildcard()` here? Using `graylog_deflector` doesn't make sense here. AFAIK

joschi: ✅ You're right. That was bollocks.

graylog2-server/src/test/java/org/graylog2/indexer/searches/SearchesTest.java
```diff
@@ -607,7 +607,7 @@ public void determineAffectedIndicesOnlyReturnsAliasesIfTooManyIndicesAreFound()
 
         final TimeRange absoluteRange = AbsoluteRange.create(now, now.plusDays(numberOfIndices + 1));
 
-        assertThat(searches.determineAffectedIndices(absoluteRange, null)).containsExactly(indexSet.getWriteIndexAlias());
+        assertThat(searches.determineAffectedIndices(absoluteRange, null)).containsExactly(indexSet.getIndexWildcard());
```
hc4: Good practice is to use constant values for expectations, insead of building it using actual code.
This is great example, why ;)

bernd:
Also please check if there are other search methods that are not using the correct index list. Thanks! :smiley:

joschi:
> I guess the search method should use the regular determineAffectedIndices() method here, not sure if there was a reason to not use it, though.

The idea was not to duplicate the work done in `determineAffectedIndicesWithRanges()` and this went well until `determineAffectedIndices(...)` started to include more complex logic than just extracting the index name from the set of index ranges.

joschi:
> I noticed another place where we are not using the alias. `indicesContainingField()` returns a list of indices.
@bernd The `Searches#indicesContainingField()` method in `Searches#fieldStats()` already receives a reduced list of indices from `Searches#determineAffectedIndices()`.

bernd:
@joschi Yes, but `Indices#getAllMessageFieldsForIndices()` returns a full list of indices again.

PR #4065, opened 2017-08-08
Make scroll value configurable instead of using a hardcoded value

bernd:
This replaces the hardcoded value of "1m" with a parameter.


graylog2-server/src/main/java/org/graylog2/indexer/results/ScrollResult.java
```diff
@@ -41,12 +41,13 @@
     };
 
     public interface Factory {
-        ScrollResult create(SearchResult initialResult, String query, List<String> fields);
+        ScrollResult create(SearchResult initialResult, @Assisted("query") String query, @Assisted("scroll") String scroll, List<String> fields);
```
joschi: Is it possible to keep the old factory method and simply mark it as `@Deprecated`?
Otherwise third party plugins might break due to this change.

bernd: I will check if that's possible. If not, we still might want to do it. The factory has been introduced in 2.3 and is probably not widely used by plugin authors.

bernd:
I added a commit to keep backwards compatibility.

PR #4096, opened 2017-08-16
Adding DSV over HTTP data adapter.

dennisoelkers:
## Motivation and Context

This adapter allows us to fetch delimiter-separated value files over
HTTP to back a lookup table with their content. The user is able to
specify the delimiter, (a) character(s) indicating lines which are
supposed to be ignored (comments) and if the lookup table should only
check for the presence of a key (in which case only true/false is
returned instead of the value).

Unlike the CSV adapter, this adapter does not support a starting line
naming the columns and specifying the key/value column.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
@@ -19,6 +19,7 @@ package org.graylog2.lookup;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
 
 import org.graylog2.lookup.adapters.CSVFileDataAdapter;
+import org.graylog2.lookup.adapters.DSVHTTPDataAdapter;
 import org.graylog2.lookup.adapters.HTTPJSONPathDataAdapter;
 import org.graylog2.lookup.caches.GuavaLookupCache;
 import org.graylog2.lookup.caches.NullCache;
@@ -49,6 +50,11 @@ public class LookupModule extends Graylog2Module {
                 HTTPJSONPathDataAdapter.class,
                 HTTPJSONPathDataAdapter.Factory.class,
                 HTTPJSONPathDataAdapter.Config.class);
+
+        installLookupDataAdapter(DSVHTTPDataAdapter.NAME,
+                DSVHTTPDataAdapter.class,
+                DSVHTTPDataAdapter.Factory.class,
+                DSVHTTPDataAdapter.Config.class);
     }
 
 }
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/DSVHTTPDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/DSVHTTPDataAdapter.java
@@ -1,0 +1,298 @@
+/**
+ * This file is part of Graylog.
+ *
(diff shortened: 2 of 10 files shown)


joschi:
The `DSVParser` class currently doesn't support using the delimiter character inside a quoted string:
```diff
diff --git a/graylog2-server/src/test/java/org/graylog2/lookup/adapters/dsvhttp/DSVParserTest.java b/graylog2-server/src/test/java/org/graylog2/lookup/adapters/dsvhttp/DSVParserTest.java
index 1db1a6b6e..93cb70a58 100644
--- a/graylog2-server/src/test/java/org/graylog2/lookup/adapters/dsvhttp/DSVParserTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/lookup/adapters/dsvhttp/DSVParserTest.java
@@ -72,7 +72,9 @@ public class DSVParserTest {
         final String input = "# Sample file for testing\n" +
                 "\"foo\":\"23\"\n" +
                 "\"bar\":\"42\"\n" +
-                "\"baz\":\"17\"";
+                "\"baz\":\"17\"\n" +
+                "\"qux\":\"42:23\"\n" +
+                "\"qu:ux\":\"42\"";
         final DSVParser dsvParser = new DSVParser("#", ":", "\"", false, false, 0, Optional.of(1));

         final Map<String, String> result = dsvParser.parse(input);
@@ -80,11 +82,13 @@ public class DSVParserTest {
         assertThat(result)
                 .isNotNull()
                 .isNotEmpty()
-                .hasSize(3)
+                .hasSize(5)
                 .containsExactly(
                         new AbstractMap.SimpleEntry<>("foo", "23"),
                         new AbstractMap.SimpleEntry<>("bar", "42"),
-                        new AbstractMap.SimpleEntry<>("baz", "17")
+                        new AbstractMap.SimpleEntry<>("baz", "17"),
+                        new AbstractMap.SimpleEntry<>("qux", "42:23"),
+                        new AbstractMap.SimpleEntry<>("qu:ux", "42")
                 );
     }
```

Test result:
```
java.lang.AssertionError: 
Expecting:
  <{"bar"="42", "baz"="17", "foo"="23", "qu"="ux", "qux"="42"}>
to contain exactly (and in same order):
  <[foo=23, bar=42, baz=17, qux=42:23, qu:ux=42]>
but some elements were not found:
  <[MapEntry[key="qux", value="42:23"], MapEntry[key="qu:ux", value="42"]]>
and others were not expected:
  <[MapEntry[key="qux", value="42"], MapEntry[key="qu", value="ux"]]>


	at org.graylog2.lookup.adapters.dsvhttp.DSVParserTest.parseQuotedStrings(DSVParserTest.java:86)
```

I would expect this to work, otherwise what's the point of having quoted keys/values?

joschi:
@dennisoelkers Would it make sense to make the entry-separator (currently "\n") configurable?

Some sources might use "\r" or "\r\n" for line separation, and some might not even use newline characters but something like `k1:v1;k2:v2;k3:v2;...`.

(Inspired by the [`Splitter`](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Splitter.html) class from Guava)

dennisoelkers:
Thanks, @joschi. Addressed both of your remarks.

PR #4125, opened 2017-09-05
Migrate Elasticsearch tests to NoSQLUnit Elasticsearch HTTP adapter

joschi:
Tests using Elasticsearch were previously relying on the NoSQLUnit Elasticsearch 2.x adapter to create an embedded Elasticsearch 2.x instance.

Unfortunately embedded Elasticsearch nodes are unsupported since Elasticsearch 5.0.0 (see https://www.elastic.co/blog/elasticsearch-the-server) so that these tests now have to rely on an externally spawned Elasticsearch instance.

In this pull request, we're utilizing the [elasticsearch-maven-plugin](https://github.com/alexcojocaru/elasticsearch-maven-plugin/tree/elasticsearch-maven-plugin-5.11) which spawns an Elasticsearch node during Maven's `verify` lifecycle phase (`integration-test` to be precise) which then can be used by Graylog's Elasticsearch tests (ending in `*IT` instead of `*Test`).


graylog2-server/pom.xml
```diff
@@ -62,6 +62,14 @@
         <maven.build.timestamp.format>yyyyMMddHHmmss</maven.build.timestamp.format>
         <!-- to work around filtering bug, which makes maven.build.timestamp inaccessible -->
         <build.timestamp>${maven.build.timestamp}</build.timestamp>
+
+        <it.es.httpPort>9200</it.es.httpPort>
```
dennisoelkers: Using the standard port might interfere with an ES instance that is running locally. We should either use a fixed but different port or (even better) use a dynamically assigned port.

graylog2-server/src/test/java/org/graylog2/ElasticsearchBase.java
```diff
@@ -57,6 +60,8 @@
 
 public abstract class ElasticsearchBase {
     private static final String PROPERTIES_RESOURCE_NAME = "elasticsearch.properties";
+    private static final String DEFAULT_PORT = "9200";
```
dennisoelkers: Shouldn't it be `19200` now?

joschi: No, because if you run the tests from the IDE, you usually want to use the default Elasticsearch HTTP port.

When the tests run in context of the `integration-test` phase of Maven, the property with the port of the Elasticsearch HTTP API of the "test" instance will be there and override the deafult.

dennisoelkers: :+1:

PR #4149, opened 2017-09-14
Use Multi Search in Searches to prevent large HTTP request headers

joschi:
Instead of artificially restricting the number of indices in a search request and resorting to using index wildcards if the list of indices threatens to hit the default HTTP header size limit of Elasticsearch (4 KB), the `Searches` class is now using the Multi Search API of Elasticsearch which allows specifying the list of indices in the request body instead of the URI path.

Refs https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-multi-search.html
Refs elastic/elasticsearch#26360
Refs #4054


bernd:
@joschi Can you please rebase against master? Thanks!

PR #4152, opened 2017-09-14
Allowing non-admin users to create dashboards.

dennisoelkers:
## Motivation and Context
Before this change,  non-admin users possessing the `dashboards:create` permission were able to create new dashboards, but were never able to access them if they do not have the global `dashboards:read` permission (which is not desired in many cases, because it allows users to access _all_ dashboards).

With this change, users creating a dashboard which lack the permission to read/edit that dashboard get the required permissions assigned dynamically. Upon dashboard deletion the user list is traversed and permissions for this specific dashboard are revoked.

## Types of changes

## Checklist:


bernd:
There will be no 2.3.2.

PR #4159, opened 2017-09-17
Expose a syncLdapUser method which can be called by other authenticators or to sync signed-in and new users from Ldap

gaspardpetit:
## Description
This changelist exposes a syncLdapUser method which creates or sync a user from Ldap. 

## Motivation and Context
At the moment, SSO authentication creates a user based on HTTP headers and default values. Instead, it would be useful to be able to authenticate a user using SSO but then sync the user roles and information from Ldap if Ldap is available.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
@@ -87,11 +87,12 @@ public class LdapUserAuthenticator extends AuthenticatingRealm {
         // safe, we only handle this type
         final UsernamePasswordToken token = (UsernamePasswordToken) authtoken;
 
-        final LdapSettings ldapSettings = ldapSettingsService.load();
-        if (ldapSettings == null || !ldapSettings.isEnabled()) {
+        if (isEnabled() == false) {
             LOG.trace("LDAP is disabled, skipping");
             return null;
         }
+
+        final LdapSettings ldapSettings = ldapSettingsService.load();
 
         final String principal = (String) token.getPrincipal();
         final char[] tokenPassword = firstNonNull(token.getPassword(), new char[0]);
@@ -175,11 +176,12 @@ public class LdapUserAuthenticator extends AuthenticatingRealm {
     @Nullable
     public User syncLdapUser(String principal) {
 
-        final LdapSettings ldapSettings = ldapSettingsService.load();
-        if (ldapSettings == null || !ldapSettings.isEnabled()) {
+        if (isEnabled() == false) {
             LOG.trace("LDAP is disabled, skipping");
             return null;
         }
+
+        final LdapSettings ldapSettings = ldapSettingsService.load();
 
         // do not try to look a token up in LDAP if there is no principal or password
         if (isNullOrEmpty(principal)) {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
```diff
@@ -159,11 +141,77 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken authtok
...
+                ldapSettings.getGroupIdAttribute(),
+                ldapSettings.getGroupSearchPattern());
+    }
+
     public boolean isEnabled() {
         final LdapSettings ldapSettings = ldapSettingsService.load();
         return ldapSettings != null && ldapSettings.isEnabled();
     }
 
+    public void syncLdapUser(String principal) {
```
joschi: The method signature should allow users to at least verify if the synchronization was successful, so instead of returning`void`, it should return a boolean value (with `true` == successful, `false` == unsuccessful).

Alternatively return the `User` instance which was synced and mark the method as `@Nullable`.

graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
```diff
@@ -159,12 +141,83 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken authtok
...
     public boolean isEnabled() {
         final LdapSettings ldapSettings = ldapSettingsService.load();
         return ldapSettings != null && ldapSettings.isEnabled();
     }
 
     @Nullable
+    public User syncLdapUser(String principal) {
+
+        final LdapSettings ldapSettings = ldapSettingsService.load();
+        if (ldapSettings == null || !ldapSettings.isEnabled()) {
```
joschi: This could be simplified to:
```
if (!isEnabled()) {
    // ...
}
```

graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
```diff
@@ -159,12 +141,83 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken authtok
...
+            LOG.trace("LDAP is disabled, skipping");
+            return null;
+        }
+
+        // do not try to look a token up in LDAP if there is no principal or password
+        if (isNullOrEmpty(principal)) {
+            LOG.debug("Principal wsa empty. Not trying to sync user with LDAP.");
```
joschi: Not originally from your changes, but it would be a great opportunity to fix this typo ("wsa" ➡️ "was"). 😉

graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
gaspardpetit: Completely agree.

graylog2-server/src/main/java/org/graylog2/security/realm/LdapUserAuthenticator.java
gaspardpetit: Agreed, I tried to follow the same style as in doGetAuthenticationInfo - would you be ok if I changed it in both places?

joschi: @gaspardpetit Yes, that would be great! 👍

PR #4160, opened 2017-09-18
Improving widget/grid stability and prop consistency

dennisoelkers:
This PR is cherry-picking a number of commits from the `post-2.3-changes` branch which aid in improving widget and grid stability and consistency.

### DIFF
--- a/graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
+++ b/graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
@@ -20,12 +20,16 @@ const MessageTablePaginator = React.createClass({
     children: PropTypes.oneOfType([
       PropTypes.arrayOf(PropTypes.element),
       PropTypes.element,
-    ]),
+    ]).isRequired,
+    onPageChange: PropTypes.func.isRequired,
+    pageSize: PropTypes.number,
     position: PropTypes.string,
   },
+
   getInitialState() {
     return {
       paginationWidth: 0,
+      pageSize: UniversalSearchStore.DEFAULT_LIMIT,
     };
   },
 
@@ -59,7 +63,7 @@ const MessageTablePaginator = React.createClass({
     }
   },
   _numberOfPages() {
-    return Math.ceil(this.props.resultCount / UniversalSearchStore.DEFAULT_LIMIT);
+    return Math.ceil(this.props.resultCount / this.props.pageSize);
   },
   _minPage() {
     const currentTenMin = Math.floor(this.props.currentPage / 10) * 10;
@@ -83,7 +87,7 @@ const MessageTablePaginator = React.createClass({
       newPage = Number(page);
     }
 
-    SearchStore.page = newPage;
+    this.props.onPageChange(newPage);
   },
   render() {
     const pages = [];
--- a/graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
+++ b/graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
@@ -6,7 +6,6 @@ import { Page } from 'components/common';
 import EventHandlersThrottler from 'util/EventHandlersThrottler';
 
 import StoreProvider from 'injection/StoreProvider';
-const SearchStore = StoreProvider.getStore('Search');
 const UniversalSearchStore = StoreProvider.getStore('UniversalSearch');
 
 import $ from 'jquery';
--- a/graylog2-web-interface/src/components/visualizations/GraphVisualization.jsx
+++ b/graylog2-web-interface/src/components/visualizations/GraphVisualization.jsx
@@ -160,6 +160,9 @@ const GraphVisualization = React.createClass({
     } else {
       this.graph.redraw();
     }
+    if (this.props.config.threshold) {
+      this.renderThreshold();
+    }
   },
   // Draws a horizontal threshold line in the graph
   renderThreshold() {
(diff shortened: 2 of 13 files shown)


graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
```diff
@@ -83,7 +87,7 @@ const MessageTablePaginator = React.createClass({
       newPage = Number(page);
     }
 
-    SearchStore.page = newPage;
```
edmundoa: We should remove the import on `SearchStore`, it seems unused since this change.

graylog2-web-interface/src/util/FormsUtils.js
```diff
@@ -14,6 +14,11 @@ const FormUtils = {
         return input.value;
     }
   },
+
+  // Emulates a minimal input event that can be used with FormUtils#getValueFromInput()
+  inputEvent(name, value) {
```
edmundoa: This does not seems to be used at the moment, and I'm also not sure what is the reasoning for adding this code. If we need a fake event for some reason, I guess we should use at least an instance of [`CustomEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent).

graylog2-web-interface/src/components/visualizations/StackedGraphVisualization.jsx
```diff
@@ -31,20 +33,21 @@ const StackedGraphVisualization = React.createClass({
...
   componentWillReceiveProps(nextProps) {
     if (deepEqual(this.props, nextProps)) {
       return;
     }
 
+    this._updateSeriesNames(nextProps);
+    this.dataPoints = this._formatData(nextProps);
     if (nextProps.height !== this.props.height || nextProps.width !== this.props.width) {
       this._resizeVisualization(nextProps.width, nextProps.height);
+      this.renderGraph(nextProps);
```
edmundoa: Do we need to call `renderGraph()` here? It looks like it is updating the series names and resizing the visualization (which are the two things dependent on props as far as I can see). Both those properties are also being updated with custom methods, so I guess we can save a re-render on this case.

graylog2-web-interface/src/components/visualizations/GraphVisualization.jsx
```diff
@@ -154,6 +160,87 @@ const GraphVisualization = React.createClass({
...
+
+      // Put circles on the threshold line to make it easier to show the tooltip
+      const dots = chartBody.selectAll('circle.threshold').data(xValues);
+      dots.enter()
+        .append('circle')
+        .attr('class', 'threshold')
+        .attr('r', thresholdDotRadius)
+        .attr('rel', 'tooltip') // To make the tooltip helper pick it up
+        .attr('data-original-title', () => {
+          return `<div class="datapoint-info">${thresholdTooltip}</div>`;
```
edmundoa: Will `thresholdTooltip` ever have some user defined value? Here we use d3 directly and are away from React's protection against code injection.

graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
```diff
@@ -27,15 +27,25 @@ const QuickValuesVisualization = React.createClass({
...
     displayAnalysisInformation: PropTypes.bool,
     displayAddToSearchButton: PropTypes.bool,
+    sortOrder: PropTypes.string,
+    dataTableTitle: PropTypes.string,
+    limit: PropTypes.number,
+  },
+  getDefaultProps() {
+    return {
+      dataTableTitle: 'Top values',
+      limit: 50,
```
edmundoa: This setting is being currently misused as far as I can see. We use it to set the table size, which I guess it's what we want. At the same time, we also use it to calculate the top values and as a cap for the number of slices in the pie chart.

edmundoa: By the way, the changes removed all usages of the `NUMBER_OF_TOP_VALUES` variable, but the variable itself is still in the code. We should remove it if we don't need it any longer.

graylog2-web-interface/src/components/search/MessageTablePaginator.jsx
dennisoelkers: ✅

graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
dennisoelkers: Actually all three are required. The idea is to provide a Top-N widget, where n is the limit of the number of top items. I do not know if the `.size(limit)` statement is required, because the top group should also include limiting the table size, obviously.

Re: `NUMBER_OF_TOP_VALUES`: ✅

graylog2-web-interface/src/components/visualizations/StackedGraphVisualization.jsx
dennisoelkers: What do you mean with "custom methods"? From my tests the graph is not properly resized without the following `renderGraph()` call.

graylog2-web-interface/src/components/visualizations/GraphVisualization.jsx
dennisoelkers: Very good catch. Escaping this one.

graylog2-web-interface/src/components/visualizations/StackedGraphVisualization.jsx
edmundoa: I mean that it looks like `renderGraph()`, `_updateSeriesNames()` and `_resizeVisualization()` are updating the same things, just at different times. If `_renderGraph()` is the one that is really needed to update the values, then I guess the other two methods are somehow redundant.

graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
edmundoa: Okay, I understand that all three are required, but I think there are two different things we want to customize:
1. Number of rows in the table (set in `size(limit)`)
2. Number of top values inside the visualization (which is used to highlight the _n_ top values in the table, and is also used to select which fields will be in the pie chart). Those settings are handled by `this.group.top(limit)` and `slicesCap(this.props.limit)`

Currently with only one setting for everything, and being 50 the default, the table will only display 50 values, all of them will be marked as "top values", and the pie chart will display 51 values (the 50 "top values" plus the others categories, if we don't limit the amount of data somewhere else). I think this may be a particular case of the settings, but using the visualization as it was before should still be possible and makes sense in my opinion.

bernd:
@dennisoelkers @edmundoa Any progress on this?

dennisoelkers:
No, not yet.

edmundoa:
Cherry-picked in 2.4 branch in 6db7ae3ad9c6ded6963165d3246926af944c7e09

PR #4162, opened 2017-09-19
Replacing karma+mocha with jest+enzyme.

dennisoelkers:
## Motivation and Context

This change is:

  * Replacing previously little used karma + mocha with jest + enzyme, which allows us to do snapshot testing
  * Updating enzyme
  * Converting existing tests to work with jest
  * Making maven run jest for tests
  * Establishing consistent directory structure for tests (tests for `src/foo/bar/Baz.jsx` are in `test/foo/bar/Baz.test.js`)

## Types of changes

## Checklist:


edmundoa:
This looks good to me and I think it can come in handy in some cases, good work!

The only thing I would change is the placement of tests. I think having another top level folder with the whole tree structure for tests is a bit annoying, specially in JS, where the IDE only helps you so much. Editing a test would require jumping on the tree, and moving a file would require moving the test as well, recreating folders and so on as you go. I don't know about others, but I will most likely forget about it and the tests tree will probably end up in an inconsistent state.

Did you consider putting the test file along the code is testing? We do that for styles now and I think it works quite well. Here I represent both ways with a couple of trees, in case I didn't make my point clear.

Current folder structure:
```
├── src
│   ├── components
│   │   ├── common
│   │   │   ├── Spinner.jsx
│   │   ├── users
│   │   │   ├── UserPreferencesButton.jsx
│   ├── logic
│   │   └── search
│   │       ├── queryParser.ts
├── test
│   ├── components
│   │   ├── common
│   │   │   ├── Spinner.test.jsx
│   │   │   └── __snapshots__
│   │   │       └── Spinner.test.jsx.snap
│   │   └── users
│   │       ├── UserPreferencesButton.test.js
│   │       └── __snapshots__
│   │           └── UserPreferencesButton.test.js.snap
│   └── logic
│       └── search
│           └── queryParser.test.js
```

Having both tests and source in the same level:
```
├── src
│   ├── components
│   │   ├── common
│   │   │   ├── Spinner.jsx
│   │   │   ├── Spinner.test.jsx
│   │   │   └── __snapshots__
│   │   │       └── Spinner.test.jsx.snap
│   │   ├── users
│   │   │   ├── UserPreferencesButton.jsx
│   │       ├── UserPreferencesButton.test.js
│   │       └── __snapshots__
│   │           └── UserPreferencesButton.test.js.snap
│   ├── logic
│   │   └── search
│   │       ├── queryParser.ts
│           └── queryParser.test.js
```

Any opinions on this?

bernd:
I like the second example with the colocation of source code and tests. I already get lost with our current system where we group web source code by type and not by sub-system. Having another code tree would be even more confusing, indeed.

dennisoelkers:
Great to hear that you prefer putting it next to the SUT, because this is actually a best practice in JS/jest-world. I hesitated from doing this, because I actually thought people would expect a directory layout that is consistent with our Java code base (separate `tests` directory).

:+1: for tests next to the SUT! I would be happy to do the change for this.

edmundoa:
Good to hear that! 👍 

This is also why I started the conversation, in the end it's a matter of taste more than anything else. @dennisoelkers let me know if you need some help with it.

dennisoelkers:
Implemented the change.

PR #4168, opened 2017-09-21
Invert filter condition to fix numeric field errors

bernd:
This broke in #4149


graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -579,7 +579,7 @@ public HistogramResult fieldHistogram(String query,
         if (failedShards > 0) {
             final List<String> errors = StreamSupport.stream(shards.path("failures").spliterator(), false)
                 .map(failure -> failure.path("reason").path("reason").asText())
-                .filter(String::isEmpty)
+                .filter(s -> !s.isEmpty())
```
joschi: A test for this would be great.

PR #4183, opened 2017-09-26
Remove embedded Auto Value JavaBean extension

joschi:
Instead of having a Maven sub-module with the Auto Value JavaBean extension,
use the split out project from https://github.com/Graylog2/auto-value-javabean

This prevents unnecessary clutter in the graylog2-server repository.


bernd:
Apart from reducing clutter, this will probably fix some build problems we had in the past.

That said, I would like to move this to 3.0 because we are coming closer to the 2.4 release and I would like to avoid touching the build process. Okay for you, @joschi?

joschi:
@bernd Fine with me.

PR #4186, opened 2017-09-27
Migrate stream matching from MessageFilter to MessageProcessor

joschi:
Instead of matching stream rules as part of the legacy message filter chain, this change set moves it into a message processor which can be independently configured and moved inside the list of message processors as the users require.

For example it wasn't possible to use the GeoIP processor for retrieving the geo location of a client and sort the message into a country-specific stream (which requires the country code) without using a pipeline rule which means running the processing pipeline last which also affects other messages (which might require a different order).

Also see:
* https://community.graylog.org/t/is-it-possible-to-make-a-stream-with-geoip-filter/2605
* https://community.graylog.org/t/stream-pipeline-and-input-extractor-clarification/5655/8

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/MessageFilterBindings.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/MessageFilterBindings.java
@@ -22,7 +22,6 @@ import com.google.inject.multibindings.Multibinder;
 import org.graylog2.filters.ExtractorFilter;
 import org.graylog2.filters.RulesFilter;
 import org.graylog2.filters.StaticFieldFilter;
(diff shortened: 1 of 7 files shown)


bernd:
Two thoughts about this:

1. We should remove the legacy message filter chain and convert all old message filters into message processors
2. We have to be really careful to make sure the order doesn't change when users upgrade because they might have changed the processor order. If we do this in a major release, it would also be okay to document the change to make sure users see it in the upgrade notes.

dennisoelkers:
As this PR has turned stale and there was no activity for almost a year, I am closing this for now.

PR #4193, opened 2017-09-29
Close response body in HTTPFileRetriever if request was unsuccessful. (master)

dennisoelkers:
## Motivation and Context

When a request was not successful but has a body, its response body was
not closed. This change checks if the response has a response body in
case it was unsuccessful and closes the response body before throwing an
exception.

Refs Graylog2/graylog-plugin-threatintel#59.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java
@@ -69,6 +69,9 @@ public class HTTPFileRetriever {
             }
         } else {
             if (response.code() != 304) {
+                if (response.body() != null) {
+                    response.close();
+                }
                 throw new IOException("Request failed: " + response.message());
             }
         }
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java
@@ -55,24 +55,21 @@ public class HTTPFileRetriever {
         }
         final Call request = client.newCall(requestBuilder.build());
 
-        final Response response = request.execute();
+        try (final Response response = request.execute()) {
+            if (response.isSuccessful()) {
+                final String lastModifiedHeader = response.header("Last-Modified", DateTime.now(DateTimeZone.UTC).toString());
+                final Map<String, String> newLastModified = new HashMap<>(this.lastLastModified.get());
+                newLastModified.put(url, lastModifiedHeader);
+                this.lastLastModified.set(ImmutableMap.copyOf(newLastModified));
 
-        if (response.isSuccessful()) {
-            final String lastModifiedHeader = response.header("Last-Modified", DateTime.now(DateTimeZone.UTC).toString());
-            final Map<String, String> newLastModified = new HashMap<>(this.lastLastModified.get());
-            newLastModified.put(url, lastModifiedHeader);
-            this.lastLastModified.set(ImmutableMap.copyOf(newLastModified));
-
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java
```diff
@@ -69,6 +69,9 @@ public HTTPFileRetriever(OkHttpClient httpClient) {
             }
         } else {
             if (response.code() != 304) {
+                if (response.body() != null) {
+                    response.close();
```
joschi: What about closing the response body in all other cases?

To cite  the [`ResponseBody` Javadoc](https://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html):
> # The response body must be closed.

dennisoelkers: That would produce an [ISE](https://github.com/square/okhttp/blob/parent-3.9.0/okhttp/src/main/java/okhttp3/Response.java#L280-L282).

joschi: I'm pretty sure that the `ResponseBody` is not `null` in line 68. 😉

Additionally, according to the developers of OkHttp, a real network response should always have a non-null body (see https://github.com/square/okhttp/pull/3277#issuecomment-292810136), so it should be fairly safe to close the response in every case.

dennisoelkers: I don't get it. Line 68 is not even reached in this case. Regarding your other point: the inline documentation of `Response#close` says clearly that `It is an error to close a response that is not eligible for a body.` and throws an ISE if it is called with a `null` body, so the library takes that case into account and therefore we should also.

joschi: > Line 68 is not even reached in this case.

Correct, but it can be reached and the response isn't closed. That's what I meant with "What about closing the response body in all other cases?" in my first comment.

https://github.com/square/okhttp/pull/3277#issuecomment-292810136
> Responses returned from calls always have a body.

dennisoelkers: Oh wow, that was misleading. Pointing out that you are referring to the successful case would have cleared this up from the very beginning.

In case of a successful request, the body is [already closed properly](https://github.com/Graylog2/graylog2-server/blob/91c34204ce5c325bffbe257537c26673af345390/graylog2-server/src/main/java/org/graylog2/lookup/adapters/dsvhttp/HTTPFileRetriever.java#L67) [according to the docs](https://github.com/square/okhttp/blob/parent-3.9.0/okhttp/src/main/java/okhttp3/ResponseBody.java#L44-L54).

joschi: The body isn't closed if the request was unsuccessful ([`Response#isSuccessful()`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Response.html#isSuccessful--)) and doesn't have 304 as return code, or if any statement before line 67 throws an exception.

All this can be solved by declaring the `Response` (via [`Call#execute()`](https://square.github.io/okhttp/3.x/okhttp/okhttp3/Call.html#execute--)) in a [try-with-resource](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) block.

dennisoelkers: Got it now. Thanks!

PR #4197, opened 2017-09-29
Waiting for data adapter/cache creation in bundle importer.

dennisoelkers:
## Motivation and Context

Before this change, lookup data adapters, caches and tables were created
sequentially during content pack import, leading to errors when lookup
tables were created before caches or data adapters were successfully
created and started. Imported lookup tables were usable only after the
server was restarted at least once after a content pack import.

After this change, the bundle importer waits for the successful creation
of data adapters/caches before creating lookup tables.

Fixes Graylog2/graylog-plugin-threatintel#57.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
+++ b/graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
@@ -17,7 +17,6 @@
 package org.graylog2.bundles;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -198,7 +197,7 @@ public class BundleImporter {
             if (!rollback()) {
                 LOG.error("Rollback unsuccessful.");
             }
-            Throwables.propagate(e);
+            throw new RuntimeException(e);
         }
     }
 
@@ -718,7 +717,7 @@ public class BundleImporter {
         }
     }
 
-    private void createLookupCaches(String bundleId, Set<LookupCacheBundle> lookupCaches) throws InterruptedException {
+    private void createLookupCaches(String bundleId, Set<LookupCacheBundle> lookupCaches) {
         for (LookupCacheBundle bundle : lookupCaches) {
             final CacheDto dto = dbCacheService.save(CacheDto.builder()
                     .title(bundle.getTitle())
@@ -738,10 +737,14 @@ public class BundleImporter {
         final CountDownLatch latch = new CountDownLatch(caches.size());
         caches.forEach(c -> c.addListener(new LatchUpdaterListener(latch), scheduler));
 
-        latch.await(30, TimeUnit.SECONDS);
+        try {
+            latch.await(30, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.warn("Starting imported Lookup Table Caches did not finish within 30 seconds. A server restart might be required for imported Lookup Tables to function.");
+        }
     }
 
-    private void createLookupDataAdapters(String bundleId, Set<LookupDataAdapterBundle> lookupDataAdapters) throws InterruptedException {
+    private void createLookupDataAdapters(String bundleId, Set<LookupDataAdapterBundle> lookupDataAdapters) {
         for (LookupDataAdapterBundle bundle : lookupDataAdapters) {
             final DataAdapterDto dto = dbDataAdapterService.save(DataAdapterDto.builder()
                     .title(bundle.getTitle())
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
```diff
@@ -709,9 +729,19 @@ private void createLookupCaches(String bundleId, Set<LookupCacheBundle> lookupCa
...
+
+        if (!createdLookupCaches.isEmpty()) {
+            clusterBus.post(CachesUpdated.create(createdLookupCaches.keySet()));
+        }
+
+        final Collection<LookupCache> caches = lookupTableService.getCaches(createdLookupCaches.keySet());
+        final CountDownLatch latch = new CountDownLatch(caches.size());
+        caches.forEach(c -> c.addListener(new LatchUpdaterListener(latch), scheduler));
+
+        latch.await(30, TimeUnit.SECONDS);
```
joschi: We should at least log a short message on either INFO or WARN if the timeout expired (which means that not all caches were refreshed).

graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
```diff
@@ -722,5 +752,15 @@ private void createLookupDataAdapters(String bundleId, Set<LookupDataAdapterBund
...
+
+        if (!createdLookupDataAdapters.isEmpty()) {
+            clusterBus.post(DataAdaptersUpdated.create(createdLookupDataAdapters.keySet()));
+        }
+
+        final Collection<LookupDataAdapter> dataAdapters = lookupTableService.getDataAdapters(createdLookupDataAdapters.keySet());
+        final CountDownLatch latch = new CountDownLatch(dataAdapters.size());
+        dataAdapters.forEach(da -> da.addListener(new LatchUpdaterListener(latch), scheduler));
+
+        latch.await(30, TimeUnit.SECONDS);
```
joschi: We should at least log a short message on either INFO or WARN if the timeout expired (which means that not all caches were refreshed).

dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
```diff
@@ -738,10 +737,14 @@ private void createLookupCaches(String bundleId, Set<LookupCacheBundle> lookupCa
         final CountDownLatch latch = new CountDownLatch(caches.size());
         caches.forEach(c -> c.addListener(new LatchUpdaterListener(latch), scheduler));
 
-        latch.await(30, TimeUnit.SECONDS);
+        try {
+            latch.await(30, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.warn("Starting imported Lookup Table Caches did not finish within 30 seconds. A server restart might be required for imported Lookup Tables to function.");
```
joschi: The `InterruptedException` is not thrown when the latch's timeout expires but only when the thread is being interrupted.

If the timeout expired, [`CountDownLatch#await()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html#await-long-java.util.concurrent.TimeUnit-) will return `false`.

graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
```diff
@@ -761,6 +764,10 @@ private void createLookupDataAdapters(String bundleId, Set<LookupDataAdapterBund
         final CountDownLatch latch = new CountDownLatch(dataAdapters.size());
         dataAdapters.forEach(da -> da.addListener(new LatchUpdaterListener(latch), scheduler));
 
-        latch.await(30, TimeUnit.SECONDS);
+        try {
+            latch.await(30, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            LOG.warn("Starting imported Lookup Table Data Adapters did not finish within 30 seconds. A server restart might be required for imported Lookup Tables to function.");
```
joschi: The `InterruptedException` is not thrown when the latch's timeout expires but only when the thread is being interrupted.

If the timeout expired, [`CountDownLatch#await()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html#await-long-java.util.concurrent.TimeUnit-) will return `false`.

graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
dennisoelkers: ✅ (Should read javadocs properly)

graylog2-server/src/main/java/org/graylog2/bundles/BundleImporter.java
dennisoelkers: ✅ (Should read javadocs properly)

PR #4198, opened 2017-09-30
Introducing support to Java8

vgdsouza:
I'm doing a project with my professor. We are change old java code for new ones. We do that using a software that he developed.


joschi:
@vgdsouza Sorry, we won't replace loops with "functional" streams just for the sake of it.

Please also read the following references:

* https://blog.jooq.org/2015/12/08/3-reasons-why-you-shouldnt-replace-your-for-loops-by-stream-foreach/
* https://dzone.com/articles/stream-performance-3

Issue #4082, opened 2017-08-14
Quick Values improvements


lennartkoopmann:
- [x] Allow to sort on value, percentage or count
- [x] Allow to build a chart of the values over time. This is important to see how values changed over time. This needs a "Only chart top 10[/30/100] values" to avoid huge charts with too many lines.
- [x] Stacking of data based on another field. For example, you could run a quick values analysis on
a source address field and then select to stack a source address against a destination address. This would then give you the top destination addresses the specific source IP has been talking to. You
could also run this without selecting a source IP first and get all conversation pairs, together with their count.

## Open Tasks

- [ ] Disable stacked field form if Elasticsearch is < 5.x
- [x] Split stacked terms result into key value pairs to fix https://github.com/Graylog2/graylog2-server/issues/4216
  - Use [WORD SEPARATOR MIDDLE DOT](http://www.fileformat.info/info/unicode/char/2e31/index.htm) unicode character to join and split the stacked fields (but also make the character configurable)

edmundoa:
Points 1 and 3 are implemented in #4205

PR #4205, opened 2017-10-04
Implement QuickValue improvements

bernd:
This PR implements the features outlined in #4082.

WIP

quick-values


graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+                     required
+                     onChange={this._onChange}
+                     value={this.state.tableSize}
+                     labelClassName="col-sm-3"
+                     wrapperClassName="col-sm-9" />
+              <Input type="radio"
+                     name="order"
+                     label="Top values"
+                     checked={this.state.order === 'desc'}
+                     onChange={this._onOrderChange('desc')}
```
edmundoa: We don't need to use a custom change method for this input. In this case we can use `value=("asc"|"desc")` in the top and bottom `Input`s to get the same effect. In that way the labels of the radio inputs also work as expected, i.e. change the values.

graylog2-web-interface/src/util/FormsUtils.js
```diff
@@ -14,6 +14,11 @@ const FormUtils = {
         return input.value;
     }
   },
+
+  // Emulates a minimal input event that can be used with FormUtils#getValueFromInput()
+  inputEvent(name, value) {
```
edmundoa: This looks quite nasty and as far as I see it is safe to remove. If we really need it, then we should use `CustomEvent` to get the same result but with an event object.

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+  _onSave(e) {
+    e.preventDefault();
+    this.props.onSave(this.state);
+  },
+
+  render() {
+    return (
+      <Row>
+        <form className="form form-horizontal" onSubmit={this._onSave}>
```
edmundoa: The `form` and `fieldset` tags should be inside `Col`, as is intended by bootstrap. Otherwise we may ran in some CSS issues.

edmundoa: There's no cancel button on this form and no obvious way of getting back to the previous screen. I would add a `Cancel` button at the end, to make it more obvious.

edmundoa: I would also like to get rid of the `form-horizontal` here and make the column size smaller. In my opinion the single column form avoids making your eyes jump from the label to the field and in this case we don't have that many inputs anyway:

screen shot 2017-10-04 at 15 28 06

At the very least we should reduce the column size, because the current inputs don't require that much space, and I think long inputs look even worse than some white empty space.

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+              <Input type="number"
+                     id="tableSize"
+                     name="tableSize"
+                     label="Total table size"
+                     required
+                     onChange={this._onChange}
+                     value={this.state.tableSize}
+                     labelClassName="col-sm-3"
+                     wrapperClassName="col-sm-9" />
+              <Input type="radio"
```
edmundoa: These two radio buttons need a label. I would use "Sort options", but feel free to go with something else you find more appropriate.

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
edmundoa: In the example I did above, I also used a bit of padding (15px), so it looks aligned with the pie chart in the quick values visualization.

graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
```diff
@@ -71,25 +80,44 @@ const FieldQuickValues = React.createClass({
...
+  },
+
   render() {
     let content;
 
     let inner;
-    if (this.state.data.length === 0) {
+    if (this.state.showVizOptions) {
+      inner = (
+        <QuickValuesOptionsForm limit={this.state.options.limit}
```
edmundoa: Adding this component as children of the `div` below, which sets `overflow: auto`, breaks the layout and makes this scroll horizontally.

graylog2-web-interface/src/util/FormsUtils.js
bernd: I will remove it. :smile:

graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
bernd: I will fix that.

graylog2-web-interface/src/stores/field-analyzers/FieldQuickValuesStore.js
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    FieldQuickValuesActions.get.promise(promise);
+  },
+
+  _errorHandler(message, title, cb) {
+    return (error) => {
+      let errorMessage;
+      try {
+        errorMessage = error.additional.body.message;
```
edmundoa: Are we always going to set a user-friendly body message on responses to this request? I know the way we handle errors is far from good, but this pattern we repeat in many places (also duplicating code all over the place) is also not that great. Error messages returned from the server are often too technical and obscure, and other times too vague for being shown to a non-technical operator.

To put an example, and I know we discussed this before. Selecting a stacked field in ES 2.x returns: `Loading quick values failed with message: Unable to perform terms query script_lang not supported [painless]` which is not helping me at all to know what went wrong with that, just making me confused.

At least I think we should have this kind of error handling abstracted somewhere and not repeating it in every store.

edmundoa: Also now that I look at it, we even accept a callback there for some reason, I guess c&p 😄

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+              </Input>
+
+              <Input label="Stacked fields">
+                <MultiSelect options={fieldOptions}
+                             value={this.state.stackedFields}
+                             onChange={this._onStackedFieldChange} />
+              </Input>
+
+              <Button type="submit" bsStyle="success" bsSize="small">Update</Button>
+              {' '}
```
edmundoa: We should use `<ButtonToolbar>` for this. From react-bootstrap docs:
>Because React doesn't output newlines between elements, buttons on the same line are displayed flush against each other. To preserve the spacing between multiple inline buttons, wrap your button group in <ButtonToolbar />.

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+  _onChange(event) {
+    this._changeConfig(event.target.name, FormsUtils.getValueFromInput(event.target));
+  },
+
+  _onStackedFieldChange(values) {
+    this._changeConfig('stackedFields', values);
+  },
+
+  _onCancel(event) {
+    event.preventDefault();
```
edmundoa: I think we don't need to prevent the `click` event here.

graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
```diff
@@ -71,28 +88,75 @@ const FieldQuickValues = React.createClass({
...
+
   render() {
     let content;
 
     let inner;
-    if (this.state.data.length === 0) {
-      inner = <Spinner />;
+    if (this.state.showVizOptions) {
+      inner = (
+        <div style={{ marginTop: 10 }}>
```
edmundoa: Can we add a CSS file to handle all these inline styles, please? :)

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+      .filter(field => field.name !== this.props.field) // Do not include the main QuickValues field
+      .map((field) => {
+        return { value: field.name, label: field.name };
+      });
+
+    return (
+      <Row>
+        <Col md={6}>
+          <form className="form" onSubmit={this._onSave}>
+            <fieldset style={{ paddingLeft: 15 }}>
```
edmundoa: Can we add this into a CSS file, please?

graylog2-web-interface/src/stores/field-analyzers/FieldQuickValuesStore.js
bernd: > Are we always going to set a user-friendly body message on responses to this request?

Nope we don't. I agree that we need some kind of centralized error handling where we can map error codes from the backend to user friendly messages in the UI. (or somthing similar)

Until we have this showing a message like in the code snippet below makes it really hard to debug the problem because our exception handling in REST resources is mostly automatic and does not log anything to the server log. 

```javascript
UserNotification.error(`Loading quick values failed with status: ${error}`,
          'Could not load quick values');
```

So I personally rather have a message that the user might not understand correctly but which helps me to debug the problem. :smiley: 

I completely agree regarding the code duplication. If we want to use this kind of error handling until we have a better one, we should put this into a helper function that can be shared.

edmundoa: Those errors may help solving the problem in some cases, but they are overly complicated, sometimes even include some Java exceptions and may even leak some information. Today for instance I was playing today with the email delivery and some details about the email server and so on may be leak from the API really easily without knowing it. As I said before, in my opinion that is no error message for a non-technical operator, which may operate Graylog.

Anyway, if we really want to change that behaviour or abstract it, we already have a helper for it:
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/logic/rest/FetchProvider.js#L14

We could add a method there without changing the default behaviour to make easier to access the body error message if it's available.

bernd: Okay, I removed the custom error handler. This needs a centralized solution. :smiley:

graylog2-web-interface/src/components/field-analyzers/QuickValuesOptionsForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+import { Input } from 'components/bootstrap';
+import { MultiSelect } from 'components/common';
+import FormsUtils from 'util/FormsUtils';
+
+const QuickValuesOptionsForm = React.createClass({
+  propTypes: {
+    limit: PropTypes.number.isRequired,
+    tableSize: PropTypes.number.isRequired,
+    order: PropTypes.string.isRequired,
+    field: PropTypes.string.isRequired,
```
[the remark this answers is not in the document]
bernd: Fixed it.

graylog2-web-interface/src/stores/field-analyzers/FieldQuickValuesStore.js
edmundoa: Hehe, ok. I'll add a task to my to do to at least discuss it or make it easier to access the response message. 👍

PR #4207, opened 2017-10-04
Add noop rotation strategy

joschi:
Users might want to control the rotation of indices externally (i. e. with programs such as [Curator](https://www.elastic.co/guide/en/elasticsearch/client/curator/5.1/index.html)) instead of relying on Graylog's built-in rotation strategies.

This change set adds a noop rotation strategy which simply does nothing.


lennartkoopmann:
Hi! I'm closing this because it does not fit into the current roadmap and company goals.

Issue #4204, opened 2017-10-03
Show total size of all Graylog indices on the index sets overview page


lennartkoopmann:
Currently, if you want to know how much disk space all indices are taking up together in total, you have to manually add up all the index set sizes you can see on the index sets overview page.

This can become very annoying if you have a lot of index sets and we should display the total size of all indices managed by Graylog on that page.

jalogisch:
could be something like the output from 

```
curl 'http://ELASTICSEARCHIP:9200/_cat/nodes?v&h=n,ip,disk*'
```

liberomic:
@jalogisch 
Do you have a command for specific indexes ? your command is great for all indexes

jalogisch:
@liberomic as this is the **cat api** of Elasticsearch, it can be shifted to the context of [indices](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-indices.html#cat-indices) very easy.

```
curl 'http://ELASTICSEARCHIP:9200/_cat/indices/graylog*?v'
```

might need a little more tuning but you are able to get that information. Please note that this is already visible in Graylog. The feature request is only to show the disk usage in some kind of information/overview.

liberomic:
@jalogisch 
THANKS
Do you have options to sum option for specifix indexes as a first command. I wand send this information in graylog to capture the storage utilizzation for specific source.

Issue #1503, opened 2015-10-22
Elasticsearch 2.0.0-rc1.


ScottJEhas:
Looks like Graylog does not want to parse the Elasticsearch string.  Does anyone know if Elasticsearch 2.0 is supported?
```
2015-10-22T13:39:00.023-05:00 WARN  [netty] [Graylog.prod.gigenet.com] Message not fully read (response) for [7] handler org.elasticsearch.discovery.zen.ping.unicast.UnicastZenPing$4@6b1f095, error [true], resetting
2015-10-22T13:39:00.512-05:00 INFO  [IndexerSetupService] Checking Elasticsearch HTTP API at http://10.201.6.221:9200/
2015-10-22T13:39:00.697-05:00 ERROR [ServiceManager] Service IndexerSetupService [FAILED] has failed in the STARTING state.
java.lang.IllegalArgumentException: unable to parse version 2.0.0-rc1
	at org.elasticsearch.Version.fromString(Version.java:539)
	at org.graylog2.initializers.IndexerSetupService.startUp(IndexerSetupService.java:170)
	at com.google.common.util.concurrent.AbstractIdleService$2$1.run(AbstractIdleService.java:54)
	at com.google.common.util.concurrent.Callables$3.run(Callables.java:95)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NumberFormatException: For input string: "0-rc1"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:492)
	at java.lang.Integer.parseInt(Integer.java:527)
	at org.elasticsearch.Version.fromString(Version.java:518)
	... 4 more
```

joschi:
Graylog is not supporting Elasticsearch 2.x for now and we won't add support for it in the 1.x version line.

PR #4211, opened 2017-10-05
Show stats of all index sets without pagination

lennartkoopmann:
Add information and stats about all indices on the "Index Sets" page without accounting for pagination. The reason for this is, that finding out the total size of all indices can be really annoying if you have a lot of index sets. See also #4204.

This change was initiated by customer request #1503. 

This introduces a new REST API endpoint at "/system/indices/index_sets/stats" to retrieve global, aggregated stats without pagination or other information about the index sets.

## How Has This Been Tested?
Tested on a local setup and also a v2.3.1 with about 500GB of data in over 400 indices and 3 index sets. The performance impact to calculate the stats with an initial concern but my benchmarks showed that it only adds about 300ms to the *Index Sets* page load time, which is acceptable IMO. Adding a caching layer or getting creative with the already computed `stats` of the paginated index sets is probably not worth the overhead.

## Screenshots:

screen shot 2017-10-05 at 2 54 38 pm

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
@@ -154,6 +154,29 @@ public class IndexSetsResource extends RestResource {
     }
 
     @GET
+    @Path("stats")
+    @Timed
+    @ApiOperation(value = "Get stats of all index sets")
+    @ApiResponses(value = {
+            @ApiResponse(code = 403, message = "Unauthorized"),
+    })
+    public IndexSetStats globalStats() {
+        checkPermission(RestPermissions.INDEXSETS_READ);
+
+        long indices = 0;
+        long documents = 0;
+        long size = 0;
+        for (IndexSetStats stats : indexSetRegistry.getAll().stream()
+                .collect(Collectors.toMap(indexSet -> indexSet.getConfig().id(), indexSetStatsCreator::getForIndexSet)).values()) {
+            indices += stats.indices();
+            documents += stats.documents();
+            size += stats.size();
+        }
+
+        return IndexSetStats.create(indices, documents, size);
+    }
+
+    @GET
     @Path("{id}")
     @Timed
     @ApiOperation(value = "Get index set")
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
@@ -174,7 +174,8 @@ public class IndexSetsResource extends RestResource {
                 .map(IndexSet::getIndexWildcard)
                 .collect(Collectors.toSet());
         final Set<IndexStatistics> indicesStats = indices.getIndicesStats(indexWildcards);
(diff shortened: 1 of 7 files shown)


kroepke:
I'm not really comfortable with this approach, because it causes one elasticsearch query for each configured index set (via the index wildcard).
It also does so sequentially, which means that if the user has so many index sets that pagination becomes a problem, the stats query will take significant time, too, for each page that's visited.

I would much rather get the indices stats for multiple index sets in parallel short of storing this information in the database for read-only indices and only asking for the latest deflector stats whenever necessary.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
```diff
@@ -154,6 +154,29 @@ public IndexSetResponse list(@ApiParam(name = "skip", value = "The number of ele
...
+            @ApiResponse(code = 403, message = "Unauthorized"),
+    })
+    public IndexSetStats globalStats() {
+        checkPermission(RestPermissions.INDEXSETS_READ);
+
+        long indices = 0;
+        long documents = 0;
+        long size = 0;
+        for (IndexSetStats stats : indexSetRegistry.getAll().stream()
+                .collect(Collectors.toMap(indexSet -> indexSet.getConfig().id(), indexSetStatsCreator::getForIndexSet)).values()) {
```
dennisoelkers: Can you please extract this (the map you are iterating over) into a local variable? This makes it way easier to understand what is going on and the type is explicitly visible.

graylog2-web-interface/src/stores/indices/IndexSetsStore.jsx
```diff
@@ -128,6 +128,26 @@ const IndexSetsStore = Reflux.createStore({
     IndexSetsActions.setDefault.promise(promise);
   },
 
+  stats() {
+    const url = URLUtils.qualifyUrl(ApiRoutes.IndexSetsApiController.stats().url);
+    const promise = fetch('GET', url);
```
dennisoelkers: If this is a long-running operation (which it probably becomes when the number of indices is big enough), we should cache the promise and return it if it still running instead of creating a new one. The general logic for this would look like this:

```
if (this.promise) {
  return this.promise;
}
this.promise = fetch(..);
this.promise.then(...).finally(() => { this.promise = undefined; });
```

This pattern avoids building up too many requests in the browser.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetsResource.java
joschi: Also, why collect into a map if you're only interested in the values anyway?

joschi: Additionally, this will run two remote queries for each index set for a piece of information which could be fetched with a single request for all index sets via the [Cat Indices API](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/cat-indices.html).

lennartkoopmann:
@kroepke I agree with your point about the parallelity. Let me take a second attempt at that part.

Thanks for the other comments @dennisoelkers! :) Will approach those, too.

lennartkoopmann:
Thank you @joschi! Came here to fix it and was pleasantly surprised. :)

PR #4233, opened 2017-10-11
Provide details about missing permissions when unauthorized

joschi:
Instead of just providing the information that some HTTP request of a user failed due to insufficient permissions, Graylog now provides helpful hints which permission was missing to successfully fulfill the request.

Example responses:
```
HTTP/1.1 403 Forbidden
X-Graylog-Node-ID: cd03ee44-b2a7-4824-be16-bb7456149dbd
Content-Type: application/json
Date: Wed, 11 Oct 2017 09:44:14 GMT
Content-Length: 123

{"type":"ApiError","message":"Not authorized. User <username> is missing permissions [users:list] to perform <GET api/users>"}
```
```
HTTP/1.1 403 Forbidden
X-Graylog-Node-ID: cd03ee44-b2a7-4824-be16-bb7456149dbd
X-Runtime-Microseconds: 81841
Content-Type: application/json
Date: Wed, 11 Oct 2017 09:39:12 GMT
Content-Length: 91

{"type":"ApiError","message":"Not authorized. User <username> is missing permission <indexsets:read>"}
```
Refs http://blog.koehntopp.info/index.php/2659-a-sad-security-user-story/ 😉

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/RestResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/RestResource.java
@@ -25,7 +25,6 @@ import org.apache.shiro.subject.Subject;
 import org.graylog2.configuration.HttpConfiguration;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.IndexSetRegistry;
-import org.graylog2.plugin.BaseConfiguration;
 import org.graylog2.plugin.database.users.User;
 import org.graylog2.shared.security.ShiroPrincipal;
 import org.graylog2.shared.users.UserService;
@@ -99,7 +98,7 @@ public abstract class RestResource {
 
     protected void checkPermission(String permission) {
         if (!isPermitted(permission)) {
-            throw new ForbiddenException("Not authorized");
+            throw new ForbiddenException("Not authorized. User is missing permission <" + permission + ">");
         }
     }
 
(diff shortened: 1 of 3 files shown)


bernd:
Do we really want to expose this information? From a security perspective, I thought it's always better to expose as _few_ information as possible. I totally see the point for user friendliness, but I think we should carefully think about this before we expose more information.

joschi:
> Do we really want to expose this information? From a security perspective, I thought it's always better to expose as few information as possible.

What's the possible attack vector?

* If users are not *authenticated*, they will receive a different message.
* If users are not *authorized*, they already received a message saying so, just without listing the individual permissions required to successfully perform the action.
* If an attacker is able to forge permissions, they will use `*` and not individual permissions included in the error message.

packs:
In general, there is a preference to display generic "Access Denied" messages such as @bernd recommends. Personally, I would prefer the user receive a generic "You do not have permissions to access this resource" and an application log be generated that would inform an administrator who tried to access what and which permission check failed.

Providing the list of failed permissions to an authenticated, but unauthorized, user is a lower risk than to unauthenticated users, however, it is information that can still be used by less (or untrusted) parties to map out application configuration and security roles. Since we can't make any assumptions as to the trustworthiness of the authenticated user, consider the case of an MSP or outsourced helpdesk, we should still limit what information is provided to them.

PR #4238, opened 2017-10-11
Add periodical to check free disk space in journal directory

joschi:
Users regularly run into problems with their Graylog setups due to the disk journal filling up the whole disk and thus corrupting the journal.

This change set adds a periodical which checks the available disk space in the journal directory and if it's insufficient, creates an urgent system notification and switches the load-balancer status of the Graylog node to DEAD.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/PeriodicalBindings.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/PeriodicalBindings.java
@@ -40,6 +40,7 @@ import org.graylog2.periodical.ThrottleStateUpdaterThread;
 import org.graylog2.periodical.UserPermissionMigrationPeriodical;
 import org.graylog2.periodical.VersionCheckThread;
 import org.graylog2.plugin.periodical.Periodical;
+import org.graylog2.shared.journal.KafkaJournalDiskCheckPeriodical;
 
 public class PeriodicalBindings extends AbstractModule {
     @Override
@@ -66,5 +67,6 @@ public class PeriodicalBindings extends AbstractModule {
         periodicalBinder.addBinding().to(ConfigurationManagementPeriodical.class);
         periodicalBinder.addBinding().to(LdapGroupMappingMigration.class);
         periodicalBinder.addBinding().to(IndexFailuresPeriodical.class);
+        periodicalBinder.addBinding().to(KafkaJournalDiskCheckPeriodical.class);
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/notifications/Notification.java
+++ b/graylog2-server/src/main/java/org/graylog2/notifications/Notification.java
@@ -60,6 +60,7 @@ public interface Notification extends Persisted {
         GC_TOO_LONG,
         JOURNAL_UTILIZATION_TOO_HIGH,
         JOURNAL_UNCOMMITTED_MESSAGES_DELETED,
+        JOURNAL_INSUFFICIENT_DISK_SPACE,
         OUTPUT_DISABLED,
         INDEX_RANGES_RECALCULATION,
         GENERIC
--- a/graylog2-server/src/main/java/org/graylog2/notifications/NotificationImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/notifications/NotificationImpl.java
@@ -32,7 +32,7 @@ import java.util.Locale;
 import java.util.Map;
 
 @CollectionName("notifications")
-public class NotificationImpl extends PersistedImpl implements Notification {
+public class    NotificationImpl extends PersistedImpl implements Notification {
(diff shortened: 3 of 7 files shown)


bernd:
I like the idea to help the user not running into a corrupt journal. :+1: I have a few thoughts about this.

**1.)** I think we should use an absolute value for the free space. In my experience that is always better than using a percentage. Also, we have a `max_journal_size` setting which we can use to calculate a sane default for the free space requirement. The journal might actually become larger than the max size setting for a short period of time because of the cleanup delay so we have to account for that and add some buffer.

**2.)** The periodical job should create metrics for free/used space so users can integrate that into their monitoring and act proactively when a disk fills up. (not sure if we have that already)

**3.)** I think we should consider a more visible warning for this like having a persistent notification on the top of every page so users __will__ see it. This small red dot is easy to miss. We can also use this for more warnings. (this is probably out of scope for this PR, but we should consider opening a ticket for this)

**4.)** In my opinion things like changing the LB status and/or shutting down inputs should be opt-in. If this works fine for one or two releases, we can consider making it more strict. I am still on the fence about the LB status, though. This might be okay to do by default. Things like stopping inputs should definitely be opt-in in my opinion.

**5.)** I think there should be an override for all actions that will be taken when the limit is crossed. Otherwise a user doesn't have any chance to continue processing when there is a temporary disk space issue or similar things.

**6.)** The free space limit and the opt-in/opt-out for the actions should be dynamically configurable per node via the web interface. Again, this helps with temporary issues and also allows the user to change limits when a disk gets resized online. Also I don't really see why we want to add more and more settings to the config file, which requires a server restart to change, when we can make this configurable at runtime.

joschi:
@bernd Thanks for your remarks. I'll go over them one by one.

> I think we should use an absolute value for the free space. In my experience that is always better than using a percentage.

I'm not sure about this one. Having 100 MB of free disk space is fine on a partition sized 1 GB (10%), but is a red flag on a partition sized 1 TB (0,01%).

> The periodical job should create metrics for free/used space so users can integrate that into their monitoring and act proactively when a disk fills up.

That's already part of the filesystem stats we expose on `/system/stats/fs`.

> I think we should consider a more visible warning for this like having a persistent notification on the top of every page so users will see it.

Isn't that what "urgent" system notifications are for? I'd rather have another PR which handles "urgent" system notifications in the way you've described.

> In my opinion things like changing the LB status and/or shutting down inputs should be opt-in.

Stopping inputs is already optional (but enabled by default). People can consciously disable this action, but I'd keep it enabled as a default. Otherwise there's no advantage over having nothing at all in single node setups (where the LB status isn't used for anything).

> I think there should be an override for all actions that will be taken when the limit is crossed.

What do you mean by override? Users can already reset the LB status and start the inputs again if they want.

> The free space limit and the opt-in/opt-out for the actions should be dynamically configurable per node via the web interface.

I would prefer it when these settings were in the same place as the general Kafka journal settings. This way they can't be missed when changing the configuration of the Kafka journal.

bernd:
> I'm not sure about this one. Having 100 MB of free disk space is fine on a partition sized 1 GB (10%), but is a red flag on a partition sized 1 TB (0,01%).

Well, our current limit is 5% on a 1TB partition this is 50GB. The default journal max size is 5 GB. That would mean processing stops once less than 50GB is free. Not sure if that's a good thing either. ;)
That's why I wanted to base the default value on the max size setting and not using an arbitrary value. If we use percentage or an actual size value in the end is debatable. Since we use a size value for the journal size, a size value for the threshold makes sense in my opinion.

> Stopping inputs is already optional (but enabled by default). People can consciously disable this action, but I'd keep it enabled as a default. Otherwise there's no advantage over having nothing at all in single node setups (where the LB status isn't used for anything).

Now that I thought about this some more, I am actually against doing anything like removing the node from the LB and/or stopping inputs. This should only notify the user that something is wrong and that's it. We already had bad experiences with automatically disabling stuff in the server. (e.g. automatically deactivating stream rules) I don't want to introduce something similar again.

> I would prefer it when these settings were in the same place as the general Kafka journal settings. This way they can't be missed when changing the configuration of the Kafka journal.

This is a setting that can easily be changed at runtime so I don't see any reason to make this a setting that needs a server restart to change it. Sure, it's more work to build UI and APIs for that. But it's totally worth it in my opinion.

jalogisch:
> We already had bad experiences with automatically disabling stuff in the server. (e.g. automatically deactivating stream rules) I don't want to introduce something similar again.

Im personal a big Fan of that - as this is one of the best indicators that people should think about what they do and the support around that isn't really present. At least from my point of view.

Just to have another voice for that feature, i'll want to raise the hand. The support is harder to make for a destroyed journal than for explain why processing stopped.

bernd:
Closing due to inactivity.

PR #4244, opened 2017-10-13
Implement QuickValues over time widget

bernd:
This implements a new visualization to show quick values over time as a stacked bar chart.

Refs #4082 

peek 2017-10-13 15-42


kroepke:
image

The Top-2 value is actually the most common term here, so somewhere the ordering goes wrong.

kroepke:
There are a couple of improvements we can do, but in the interest of getting broader feedback we'll postpone those and talk about them in subsequent PRs.

Rest lgtm!

PR #4268, opened 2017-10-19
updated default/example graylog.conf

jalogisch:
## Description
I realized after questions in the community and training sessions that ppl. tend to change the elasticsearch setting in the configuration file and ask why they are not changed after a restart of graylog. 

My intention here is that ppl can actually read where to find the setting in the web interface. It might be technical not 100% correct (I did not checked that) but it will point the ppl to the point where they can change them in a running Graylog environment.

## Types of changes

## Checklist:

### DIFF
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -555,8 +555,9 @@ mongodb_threads_allowed_to_block_multiplier = 5
 #
 # ATTENTION: These settings have been moved to the database in Graylog 2.2.0. When you upgrade, make sure to set these
 #            to your previous settings so they will be migrated to the database!
-#            This setting is only used once on first startup after that 
-#            all settings are made online in “System > Indices” in the Graylog web interface.
+#            This configuration setting is only used on the first start of Graylog. After that,
+#            index related settings can be changed in the Graylog web interface on the 'System / Indices' page.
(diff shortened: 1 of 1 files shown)


misc/graylog.conf
```diff
@@ -239,6 +239,8 @@ plugin_dir = plugin
 #
 # ATTENTION: These settings have been moved to the database in 2.0. When you upgrade, make sure to set these
 #            to your previous 1.x settings so they will be migrated to the database!
+#            This setting is only used once on first startup after that 
+#            all settings are made online in “System > Indices” in the Graylog web interface.
```
joschi: Please use "normal" double quotes (`"`) instead of `“`.

misc/graylog.conf
```diff
@@ -239,6 +239,8 @@ plugin_dir = plugin
 #
 # ATTENTION: These settings have been moved to the database in 2.0. When you upgrade, make sure to set these
 #            to your previous 1.x settings so they will be migrated to the database!
+#            This setting is only used once on first startup after that
```
joschi: "This configuration setting is only used on the first start of Graylog."

misc/graylog.conf
joschi: "After that, index related settings can be changed in the Graylog web interface on the "System / Indices" page. Also see http://docs.graylog.org/en/2.3/pages/configuration/index_model.html#index-set-configuration."

PR #4302, opened 2017-10-30
React router update

edmundoa:
This PR updates our current `react-router` version from 1.0.3 to 2.8.1. Due to this update, we also need to update `history` to 2.1.2 and `react-router-bootstrap` to .

Changes:
- [react-router](https://github.com/ReactTraining/react-router/blob/v2.8.1/CHANGES.md)
- [history](https://github.com/ReactTraining/history/blob/v2/CHANGES.md)
- [react-router-bootstrap](https://github.com/react-bootstrap/react-router-bootstrap/blob/v0.23.2/CHANGELOG.md)

Looking at the changes, there are two big changes that we needed to take care here:
- Remove usages of `history` through react-router's props, as this is deprecated now. Instead we use our `History` util, which let us access the global react-router history
- Replace usages of `pushState()` and `replaceState()` with `push()` and `replace()`. The method signature also changed, and we no longer need to pass `null` as first argument

Plugins also need some adjustments, but as changes are deprecations and the code will continue working as before, you can start reviewing (and even merging) if you want to. I'll link changes in plugins to this PR for future reference.


edmundoa:
Rebased to the latest master.

PR #4311, opened 2017-11-02
Update ES6 classes formatting

edmundoa:
As ES6 React components are here to stay, update existing ES6 classes to ensure other people have a more or less good example to look at:

- Use arrow functions instead of manually binding functions in constructor
- Add `propTypes`, `defaultProps` and state directly inside the class, making components look more like those created with `createClass`. Unfortunately creating an ESLint rule for this requires more time that I'm willing to spend on it right now
- Use function `ref`s
- Remove superfluous functions


edmundoa:
FYI: Most of the new style applied to the ES6 components came from https://babeljs.io/blog/2015/06/07/react-on-es6-plus and that seems to be the "standard" way of using it these days.

PR #4322, opened 2017-11-07
Improve robustness of QuickValues with stacked fields

### LINKED ISSUE #4289: Improve behaviour when stacking field on quick values fails
Stacking fields on quick values is not supported in ES 2, but there is no easy way of disabling the controls in that case. When trying to stack a field in ES 2, we end up with the quick values component stuck in a loading state and an error notification showing up, which is quite confusing.

## Expected Behavior
Capture errors when stacking fields in quick values and show an error message in the div. At least we should update the component state, to avoid that it keeps on that loading state.

## Current Behavior
The component stays in a broken state until the configuration changes or the quick values analyzer is dismissed and added again.

1. Use ES 2
2. Calculate quick values for a field
3. Stack another field in the quick values configuration
4. See the component spin until the configuration changes or quick values are dismissed

## Your Environment
* Graylog Version: 2.4.0-beta.1
* Elasticsearch Version: ES 2.x

bernd:
- Disable the stacked field form in QuickValues configuration when Elasticsearch version is < 5
- Reset loading state in case of an error so the user can change settings and try again
- Reset QuickValues visualization config to defaults when analyzing a new field

Closes #4289 

Note: This needs to go into the 2.4 branch as well

### DIFF
--- a/graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
+++ b/graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
@@ -164,11 +164,11 @@ const FieldQuickValues = React.createClass({
     if (this.state.field !== undefined) {
       this.setState({ loadingData: true });
       if (this.state.showHistogram) {
-        FieldQuickValuesActions.getHistogram(this.state.field, this.state.fieldQueryObjects, this.state.options).then(() => {
+        FieldQuickValuesActions.getHistogram(this.state.field, this.state.fieldQueryObjects, this.state.options).finally(() => {
           this.setState({ loadingData: false });
         });
       } else {
-        FieldQuickValuesActions.get(this.state.field, this.state.options).then(() => {
+        FieldQuickValuesActions.get(this.state.field, this.state.options).finally(() => {
           this.setState({ loadingData: false });
         });
       }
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/components/field-analyzers/FieldQuickValues.jsx
```diff
@@ -157,11 +166,11 @@ const FieldQuickValues = React.createClass({
     if (this.state.field !== undefined) {
       this.setState({ loadingData: true });
       if (this.state.showHistogram) {
-        FieldQuickValuesActions.getHistogram(this.state.field, this.state.fieldQueryObjects, this.state.options).then(() => {
+        FieldQuickValuesActions.getHistogram(this.state.field, this.state.fieldQueryObjects, this.state.options).finally(() => {
```
edmundoa: Unfortunately this is not enough to remove the loading spinner if there was an error getting the histogram, as there will be no data at that point. As this is disabled now for ES 2.x users, I guess it's not a blocker, but still annoying if someone runs into it.

edmundoa: I'm just wondering, why don't we use the `loading` state from the store in here?

PR #4328, opened 2017-11-08
Replace webpack-dev-server with custom express server

edmundoa:
Many of us (if not everyone) has problems running `webpack-dev-server` with many plugins, as the server keeps disconnecting from the web browser.

In order to solve that, but keeping the setup easy to install and start, this PR adds a custom express server, using `webpack-dev-middleware` and `webpack-hot-middleware` to achieve something similar to what we had before with the `start-nohmr` script. This works by enabling webpack hot module replacement, but not accepting any changes, forcing webpack to reload the page when a change is detected.

Apart from performing a full page reload on each change, the new setup also helps highlighting errors in files, producing a result like this:
screen shot 2017-11-08 at 14 55 21

Regarding build times, they seem to be roughly the same, but this needs more testing and also some external experiences to see if it's any good.

### DIFF
--- a/graylog2-web-interface/devServer.js
+++ b/graylog2-web-interface/devServer.js
@@ -1,0 +1,37 @@
+const express = require('express');
+const webpack = require('webpack');
+const compress = require('compression');
+const history = require('connect-history-api-fallback');
+const webpackDevMiddleware = require('webpack-dev-middleware');
+const webpackHotMiddleware = require('webpack-hot-middleware');
+const webpackConfig = require('./webpack.bundled');
+
+const app = express();
+const vendorConfig = webpackConfig[0];
+const appConfig = webpackConfig[1];
+// Use two compilers to avoid re-compiling the vendor config on every change.
+// We assume dependencies won't change while the server is running.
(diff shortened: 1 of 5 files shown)


edmundoa:
Forgot to mention:

- The collector and map widgets will not reload by default, as their entry point is already accepting HMR changes. If we continue with these changes, we would need to adapt those plugins
- The changes in here only affect the `web-nohmr` run script, all other npm scripts should continue working as before

graylog2-web-interface/devServer.js
```diff
@@ -0,0 +1,37 @@
...
+app.use(webpackDevMiddleware(vendorCompiler, {
+  publicPath: appConfig.output.publicPath,
+  lazy: false,
+  noInfo: true,
+}));
+
+app.use(webpackDevMiddleware(appCompiler, {
+  publicPath: appConfig.output.publicPath,
+  lazy: false,
+  noInfo: true,
```
edmundoa: FYI: Removing this will give you more information about the bundle update.

PR #4333, opened 2017-11-09
Define REPLACE_MODULES in webpack production build as well

bernd:
Fixes an issue with running the production build.


joschi:
@bernd Which issue does it fix?

bernd:
@joschi Unbreaks the web interface by fixing this:
pasted image at 2017_11_09 03_41 pm

edmundoa:
Just to add a bit more context, this was introduced in #4328, as I added the `REPLACE_MODULES` variable to keep the old HMR development build working.

Issue #4171, opened 2017-09-21
Fix JavaScript warnings


bernd:
## Current Behavior

Since we updated several JavaScript dependencies to newer versions (incl. React), we are getting a lot of warnings and deprecation notes when using the web interface. This makes it really hard to see real problems while developing new features or debugging problems.

![image](https://user-images.githubusercontent.com/461/30700475-060611c2-9ee8-11e7-8a5a-31601ea3c2d0.png)

## Your Environment

* Graylog Version: 2.4.0-SNAPSHOT

bernd:
I guess this has been fixed in the meantime.

PR #4334, opened 2017-11-09
Add missing Input ids

edmundoa:
Following the effort of reduce the console warnings (#4171), this PR adds ids to all `Input` elements.

Since the react-bootstrap upgrade in https://github.com/Graylog2/graylog2-server/pull/3598, id props are required in form controls, which we tried to enforce by adding warnings in the console. After all this time, is time to fix those usages and enforce the id prop on our own `Input` component.

I also changed a few components API to ensure we can pass a reasonable id to them.

Plugins should also be updated, although the changes in here should not break them, just spit a slightly noisier PropType validation error during development. I'll take care of them in follow up PRs.


edmundoa:
The linter errors are here I'm afraid unavoidable. I touched so many files that I can't really reformat them all to make the job pass without cluttering the already hard to review PR.

PR #4353, opened 2017-11-15
Add pluggable global notification area to App component

bernd:
A plugin can register a "globalNotifications" UI extension that will be rendered into the global notification area on each page.

Note: This needs to be cherry-picked into 2.4

## Example notification

image

### Component

```jsx
import React from 'react';
import { Alert } from 'react-bootstrap';

const GlobalTestNotification = React.createClass({
  render() {
    return (
      <Alert bsStyle="danger">
        <h4><strong>Alert</strong></h4>
        <p>This is an example alert.</p>
      </Alert>
    );
  },
});

export default GlobalTestNotification;
```

### Register

```jsx
import packageJson from '../../package.json';
import { PluginManifest, PluginStore } from 'graylog-web-plugin/plugin';
import GlobalTestNotification from 'GlobalTestNotification';

PluginStore.register(new PluginManifest(packageJson, {
  globalNotifications: [
    {
      key: 'org.graylog.plugins.test.GlobalTestNotification',
      component: GlobalTestNotification,
    },
  ],
}));
```

### DIFF
--- a/graylog2-web-interface/src/routing/AppGlobalNotifications.jsx
+++ b/graylog2-web-interface/src/routing/AppGlobalNotifications.jsx
@@ -1,8 +1,6 @@
 import React from 'react';
 import { PluginStore } from 'graylog-web-plugin/plugin';
 
-import styles from './AppGlobalNotifications.css';
-
 const AppGlobalNotifications = React.createClass({
   render() {
     const globalNotifications = PluginStore.exports('globalNotifications')
@@ -22,7 +20,7 @@ const AppGlobalNotifications = React.createClass({
       .filter(component => !!component);
 
     return (
-      <div className={`container-fluid ${styles.globalNotifications}`} id="global-notifications">
+      <div id="global-notifications">
         {globalNotifications}
       </div>
     );
--- a/graylog2-web-interface/src/routing/AppGlobalNotifications.jsx
+++ b/graylog2-web-interface/src/routing/AppGlobalNotifications.jsx
@@ -8,11 +8,13 @@ const AppGlobalNotifications = React.createClass({
     const globalNotifications = PluginStore.exports('globalNotifications')
       .map((notification) => {
         if (!notification.component) {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/routing/AppGlobalNotifications.jsx
```diff
@@ -0,0 +1,32 @@
...
+        if (!notification.key) {
+          // eslint-disable-next-line no-console
+          console.warn('Missing "key" for globalNotification plugin:', notification);
+        }
+        return <Component key={notification.key} />;
+      })
+      .filter(component => !!component);
+
+    return (
+      <div className={`container-fluid ${styles.globalNotifications}`} id="global-notifications">
```
edmundoa: Could we wrap the component's children in a react-bootstrap's `Row` and `Col`, avoiding to add styles in here? In the end the styling is just mimicking what those components do.

bernd: I tried that before. The problem is that we get padding and margins even when there is no active notification.

Maybe you have an idea how to fix that when using Row/Col?

## With Row/Col
image

## With div and custom style

image

edmundoa: Oh, I see, that's why you are adding `:not(:empty)` to the CSS selector.

There is afaik no way for a parent to know if its children are rendering something or not. This is mostly due to how React works: you try to put all the logic you can upper in the tree and pass things down, not the other way around, which is what this situation would need.

Then I would suggest: why don't we leave the styling to the children? In the end we are not adding much styling in here, so we can also let plugin authors decide how the notification should look like. In the end they could also override anything we set in here.

bernd: @edmundoa Good point. I will remove the styling and leave it to the children. :+1:

bernd:
@edmundoa I removed the styling.

PR #4383, opened 2017-11-29
Frontend documentation and components gallery

edmundoa:
This PR introduces a frontend documentation site built with [React Styleguidist](https://react-styleguidist.js.org).

The documentation and site is still in a very early stage, but I think it's better to merge the changes sooner than later, since there are already lots of them.

As a summary, I took care of two things in this PR:

1. Install and do a basic configuration of React Styleguidist. I am still not satisfied with the styling, and we need to find a way of properly mocking stores, as they currently try to do one request
2. Document all our `common` components. For now I also included other components in the gallery, but they are still not documented. I will take care of that in further PRs
3. Write some very basic global frontend documentation. We need to expand that in the future as well

To see the new documentation site, we have two new yarn scripts that let you do that:

- `yarn run docs:server` will run a development server with the documentation
- `yarn run docs:build` will generate HTML that can be used to serve the site from the local system or another server

I'm sorry about the length of the PR 🐙 I can split it up if the reviewer thinks that will make the task easier.


edmundoa:
As I change many files in here, it's normal the linter is complaining. I didn't see any warnings in IntelliJ while writing the documentation, so

PR #4416, opened 2017-12-13
Don't render partial page content when missing permissions

### LINKED ISSUE #4407: Missing function level access control allows access to pages by unauthorised users
**Description**

Some pages of the application can be accessed by unauthorised users.

**Impact**

Currently no sensitive data is available on these pages that can’t be access via another
method. However, access to these pages should be restricted.

The existence of this vulnerability can be verified by navigating to these URL’s 
```text
https://<hostname>/system/authentication/config
https://<hostname>/system/grokpatterns
https://<hostname>/system/logging
```

_Graylog 2.2.0 on Redhat 7_

kroepke:
Hides various content when the user does not have the correct permission to avoid accidentally leaking information in the future.

Added a new `authentication:read` permission to disambiguate the `clusterconfigentry:read` check

fix #4407

### DIFF
--- a/graylog2-web-interface/src/components/grok-patterns/GrokPatterns.jsx
+++ b/graylog2-web-interface/src/components/grok-patterns/GrokPatterns.jsx
@@ -8,6 +8,7 @@ import PageHeader from 'components/common/PageHeader';
 import EditPatternModal from 'components/grok-patterns/EditPatternModal';
 import BulkLoadPatternModal from 'components/grok-patterns/BulkLoadPatternModal';
 import DataTable from 'components/common/DataTable';
+import IfPermitted from 'components/common/IfPermitted';
 
 const GrokPatterns = React.createClass({
   getInitialState() {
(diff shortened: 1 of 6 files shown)


kroepke:
Against master, should be backported to 2.4 for next release.

kroepke:
linter errors are due to old codebase

bernd:
@edmundoa Can you please review this? :smiley:

graylog2-web-interface/src/pages/GrokPatternsPage.jsx
```diff
@@ -1,13 +1,15 @@
...
-import { DocumentTitle } from 'components/common';
+import { DocumentTitle, IfPermitted } from 'components/common';
 
 const GrokPatternsPage = React.createClass({
   render() {
     return (
       <DocumentTitle title="Grok patterns">
-        <GrokPatterns />
+        <IfPermitted permissions="inputs:read">
+          <GrokPatterns />
```
edmundoa: Shouldn't we leave the permissions check to the component that renders the entity? With your changes, we are already checking in `GrokPatterns` for the right permissions, so I guess this other check can be avoided.

graylog2-web-interface/src/components/grok-patterns/GrokPatterns.jsx
```diff
@@ -87,26 +88,30 @@ const GrokPatterns = React.createClass({
...
-            <DataTable id="grok-pattern-list"
-                       className="table-striped table-hover"
-                       headers={headers}
-                       headerCellFormatter={this._headerCellFormatter}
-                       sortByKey={'name'}
-                       rows={this.state.patterns}
-                       dataRowFormatter={this._patternFormatter}
-                       filterLabel="Filter patterns"
-                       filterKeys={filterKeys} />
+            <IfPermitted permissions="inputs:read">
```
edmundoa: This `DataTable` component also renders `delete` and `edit` buttons for each grok pattern. They don't seem to work, but it feels "wrong". I guess we should either disable (or remove) those buttons, or not render the table unless the user can edit it.

graylog2-web-interface/src/pages/GrokPatternsPage.jsx
kroepke: True!

kroepke:
@edmundoa Fixed, I also cleaned up some more obvious eslint warnings.

PR #4417, opened 2017-12-13
check node id file permissions as part of parameter validation

### LINKED ISSUE #4410: Permission issues on node-id file lead to not very helpful error messages and exit
The server repeatedly prints an error that no node-id file exists and that it created a new fil before exit if the node-id file cannot be written, **but no error about a file permission issue is printed**. This happened to me on initial setup. 

I moved this into the 2.4.0 milestone for now but am OK with moving it into 3.0.0 if it's hard to fix and if it is not a recent regression.

## Your Environment

* Graylog Version: v2.4.0-beta.4-SNAPSHOT
* Operating System: Ubuntu 17.10

kroepke:
check node id file permissions as part of parameter validation, instead of failing during injection

doing IO during object construction is brittle and leads to unintelligible error messages

fixes #4410

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -18,6 +18,7 @@ package org.graylog2;
 
 import com.github.joschi.jadconfig.Parameter;
 import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.Validator;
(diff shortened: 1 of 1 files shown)


kroepke:
Could be backported to 2.4 if someone feels it is important enough.

kroepke:
I didn't use the built-in `File` validator classes, because unfortunately the config parameter is a string, not a file and I didn't want to make too many changes for 2.4.

PR #4418, opened 2017-12-13
allow to pass custom renderers for key and value titles in histograms

### LINKED ISSUE #4395: Traffic graphs have wrong data type in overlay
The new traffic graphs on the Cluster Overview page re-use the `<TrafficGraph />` component but do not change the word "messages" in the hover overlay to "bytes":

screen shot 2017-12-04 at 12 48 08 pm

This should say "48,083,000 bytes", or, even better, "48.08 MB".

kroepke:
the default renderers can now be replaced individually by setting props

used in TrafficGraphs to display nicer values

fixes #4395


kroepke:
Should be backported to 2.4 once merged

PR #4419, opened 2017-12-13
Add source editor component

edmundoa:
This PR moves the source code editor we used in the pipelines to the web interface, so other plugins can reuse it if they need it.

I manage to make a few improvements in there, hopefully they will improve the experience of using it:
- Updated to latest version of ReactAce
- Fixed issue with undo command after load: https://github.com/Graylog2/graylog-plugin-pipeline-processor/issues/224. I'm not fully satisfied with the fix, so I'll try to do a better fix and get it upstream
- Allow editor to be user-resizable, fixing https://github.com/Graylog2/graylog-plugin-pipeline-processor/issues/229
- Add read-only flag, so we can also pre-visualize code with the component
- Add toolbar to make more visibles some actions that can be done in the editor. For now only copying, pasting*, undo, and redo are available.

Bonus: Add dark theme support! 
screen shot 2017-12-13 at 17 36 42

*Dispatching a pasting event from JS doesn't really work for privacy reasons, so the user must trigger an action in the OS. The button therefore only shows a small tooltip indicating it can be done in the Edit menu.

### DIFF
--- a/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
+++ b/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
@@ -82,7 +82,6 @@ class SourceCodeEditor extends React.Component {
     return (
       <Resizable height={height}
                  width={width}
-                 axis="y"
                  minConstraints={[200, 200]}
                  onResize={this.handleResize}>
         <div className={containerStyle} style={{ height: height, width: Number.isNaN(width) ? '100%' : width }}>
--- a/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
+++ b/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
@@ -109,13 +109,13 @@ class SourceCodeEditor extends React.Component {
     }
   }
 
-  isRedoDisabled = () => {
-    return this.props.readOnly || !this.reactAce || !this.reactAce.editor.getSession().getUndoManager().hasRedo();
-  }
+  isCopyDisabled = () => this.props.readOnly || this.state.selectedText === '';
 
-  isUndoDisabled = () => {
-    return this.props.readOnly || !this.reactAce || !this.reactAce.editor.getSession().getUndoManager().hasUndo();
-  }
+  isPasteDisabled = () => this.props.readOnly;
+
+  isRedoDisabled = () => this.props.readOnly || !this.reactAce || !this.reactAce.editor.getSession().getUndoManager().hasRedo();
+
+  isUndoDisabled = () => this.props.readOnly || !this.reactAce || !this.reactAce.editor.getSession().getUndoManager().hasUndo();
 
   handleRedo = () => {
     this.reactAce.editor.redo();
@@ -154,9 +154,9 @@ class SourceCodeEditor extends React.Component {
                                  onSuccess={this.focusEditor}
                                  text={this.state.selectedText}
                                  buttonTitle="Copy (Ctrl+C / &#8984;C)"
-                                 disabled={this.state.selectedText === ''} />
+                                 disabled={this.isCopyDisabled} />
                 <OverlayTrigger placement="top" trigger="click" overlay={overlay} rootClose>
-                  <Button bsStyle="link" bsSize="sm" title="Paste (Ctrl+V / &#8984;V)">
+                  <Button bsStyle="link" bsSize="sm" title="Paste (Ctrl+V / &#8984;V)" disabled={this.isPasteDisabled}>
                     <i className="fa fa-paste fa-fw" />
                   </Button>
                 </OverlayTrigger>
--- a/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
+++ b/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
@@ -57,6 +57,7 @@ class SourceCodeEditor extends React.Component {
     this.state = {
       height: props.height,
       width: props.width,
+      selectedText: '',
     };
   }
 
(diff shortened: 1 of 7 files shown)


edmundoa:
PS: For now the dark theme is only supported when specified via props. I guess we can do something better with that once we do some other improvements 🙂

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -0,0 +1,205 @@
...
+ * editors.
+ */
+class SourceCodeEditor extends React.Component {
+  static propTypes = {
+    /**
+     * Annotations to show in the editor's gutter. The format should be:
+     * `[{ row: 0, column: 2, type: 'error', text: 'Some error.'}]`
+     * The type value must be one of `error`, `warning`, or `info`.
+     */
+    annotations: PropTypes.array,
```
kroepke: Is it worth using a `PropTypes.shape` here or do we not bother because this typically comes from the backend anyway?

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -0,0 +1,205 @@
...
+                   width={width}
+                   minConstraints={[200, 200]}
+                   onResize={this.handleResize}>
+          <div className={containerStyle} style={{ height: height, width: validCssWidth }}>
+            <AceEditor ref={(c) => { this.reactAce = c; }}
+                       annotations={this.props.annotations}
+                       editorProps={{ $blockScrolling: 'Infinity' }}
+                       focus={this.props.focus}
+                       fontSize={this.props.fontSize}
+                       mode="text"
```
kroepke: Do we want to extract this into an optional prop, too?
Might come in handy if we ever use this to display/edit JSON or other modes.

graylog2-web-interface/src/components/common/SourceCodeEditor.md
```diff
@@ -0,0 +1,67 @@
...
+  { row: 3, column: -1, text: 'info!', type: 'info' },
+];
+<SourceCodeEditor id="editor-2"
+                  annotations={annotations}
+                  readOnly
+                  theme="dark"
+                  value={code} />
+```
+
+Non-resizable editor without toolbar and with custom height and width:
```
kroepke: States that the editor isn't resizable, when in fact it is.

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
edmundoa: The main problem is that each mode requires an `import` from `brace`. I couldn't find a way of importing all the modes brace provides, and I'm unsure if it's even safe to do so. I can play a bit more with it and see how it goes. Otherwise I would vote for extending the component when we need it.

graylog2-web-interface/src/components/common/SourceCodeEditor.md
edmundoa: That was me changing props and texts last minute 😅 Will fix it.

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
edmundoa: I'm most of the time against those shape objects, as you need to define the whole schema of an object (or an array of objects in this case) in a way that is not very readable and it's easy to make you waste more time removing warnings than helping catching errors. Additionally, in this case the shape is highly dependant on ace, and they don't even have good documentation for it. For instance, the `column` property didn't seem to have any effect when I tested it, as all annotations appear in the gutter on the right of the editor. I also had to go to the source code to see what types were supported.
In my opinion, having an example and the vales the `type` property accepts is good enough.

kroepke: Works for me :)

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
kroepke: Yeah makes sense, let's not overcomplicated it for now.

edmundoa:
@kroepke addressed your comments and other things I saw:
- Now we support JSON, Lua, Markdown, and YAML languages with the `mode` prop
- Improved the logic to disabled copy and paste buttons
- Avoided resizing editor and text selection when they don't make sense
- Added a note regarding uncontrolled usages of the component, also updated examples to always keep the state themselves. React-Ace is supposed to keep the text value if you don't update it by hand, but that seems to fail in some situations

edmundoa:
I rebased this against master and also removed the hack to reset the undo history. The issue was fixed upstream.

edmundoa:
The failing test is unrelated to the changes, but to the new year :tada: I'll submit another PR fixing it.

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -0,0 +1,205 @@
...
+          <div className={style.toolbar} style={{ width: validCssWidth }}>
+            <ButtonToolbar>
+              <ButtonGroup>
+                <ClipboardButton title={<i className="fa fa-copy fa-fw" />}
+                                 bsStyle="link"
+                                 bsSize="sm"
+                                 onSuccess={this.focusEditor}
+                                 text={this.state.selectedText}
+                                 buttonTitle="Copy (Ctrl+C / &#8984;C)"
+                                 disabled={this.isCopyDisabled} />
```
bernd: This passes the `isCopyDisabled` function instead of executing it.

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -0,0 +1,205 @@
...
+                <OverlayTrigger placement="top" trigger="click" overlay={overlay} rootClose>
+                  <Button bsStyle="link" bsSize="sm" title="Paste (Ctrl+V / &#8984;V)" disabled={this.isPasteDisabled}>
```
bernd: This passes the `isPasteDisabled` function instead of executing it.

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
edmundoa: Good finding! This got broken in one of the changes I did.

PR #4423, opened 2017-12-15
Convert timestamp quickvalues to formatted date

### LINKED ISSUE #4288: Add timestamp field to search query through quick values does not work
## Expected Behavior
Using the "add to query" button in quick values to add a timestamp field should add the right date time to the search bar.

## Current Behavior
Using the "add to query" button in quick values to add a timestamp field adds "Invalid date" to the search bar.

1. Run quick values analyzer on `timestamp` field
2. Add one of the values to the query by using the "add to query" button
3. Observe that no date or timestamp was added, but "Invalid date"

## Your Environment
* Graylog Version: Graylog 3.0.0-SNAPSHOT (aada4c64b6e8fae20bf972aa40fec330d5686137). I could not reproduce it in two 2.4.0 setups
* Elasticsearch Version: 5.6.2

kroepke:
otherwise the search store will end up searching for the unix timestamp which is not supported

fixes #4288

### DIFF
--- a/graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
+++ b/graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
@@ -7,6 +7,8 @@ import dc from 'dc';
 import d3 from 'd3';
 import deepEqual from 'deep-equal';
 import $ from 'jquery';
+import DateTime from 'logic/datetimes/DateTime';
+import moment from 'moment';
 
 import D3Utils from 'util/D3Utils';
 import StringUtils from 'util/StringUtils';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
```diff
@@ -192,7 +194,10 @@ const QuickValuesVisualization = React.createClass({
           // different terms in the stacked value more visible.
           formattedTerm = this.props.data.terms_mapping[d.term].map(t => t.value).join(' <strong style="color: #999999;">&mdash;</strong> ');
         }
-
+        if (this.props.field === 'timestamp') {
+          // convert unix timestamp to proper formatted value, so that add to search button works correctly
+          formattedTerm = moment.utc(Number(d.term)).format(DateTime.Formats.TIMESTAMP);
```
edmundoa: This does not display date times in the user time zone, providing inconsistent values with the search results. We should use `new DateTime(Number(d.term)).toString(DateTime.Formats.TIMESTAMP)` to get the time formatted in the user's time zone.

graylog2-web-interface/src/components/visualizations/QuickValuesVisualization.jsx
```diff
@@ -205,7 +210,12 @@ const QuickValuesVisualization = React.createClass({
     ];
 
     if (this.props.displayAddToSearchButton) {
-      columns.push(d => this._getAddToSearchButton(d.term));
+      if (this.props.field === 'timestamp') {
+        // convert unix timestamp to proper formatted value, so that add to search button works correctly
+        columns.push(d => this._getAddToSearchButton(moment.utc(Number(d.term)).format(DateTime.Formats.TIMESTAMP)));
```
edmundoa: This also has problems with the user TZ, as the value from here is later converted into UTC (see gif below) and when pressing the search button we can't find any results. We should use, as in the other case `new DateTime(Number(d.term)).toString(DateTime.Formats.TIMESTAMP)` to convert the unix time in the user's selected time zone.

time

edmundoa: By the way, as both date times should be formatted the same way, I would extract the conversion into a function for extra clarity.

bernd:
Moving to 3.0.0 milestone.

PR #4424, opened 2017-12-15
remove some traffic graphs

kroepke:
We don't need the incoming and decoded traffic graphs anymore.


kroepke:
Should be backported to 2.4 as well

PR #4430, opened 2017-12-18
Restore ability to run behind reverse proxy

### LINKED ISSUE #4315: Issues with new HTTP config handling
In #3840 we simplified the HTTP config handling to make it easier for users. I noticed a few issues.

## Expected Behavior

Previously, when the server was configured to listen on `127.0.0.1:9000`, I could also use `localhost:9000` to access the server.

## Current Behavior

This doesn't work anymore because we have a new `<base/>` tag in our `index.html.template` file.

https://github.com/Graylog2/graylog2-server/blob/0c3c7f9907f69e2611ee4a9d162d4f5b3b975788/graylog2-server/src/main/resources/web-interface/index.html.template#L8

That one somehow trips up the react-router history in production mode:

image

I also noticed that when we use the `X-Graylog-Server-URL` HTTP header, the `gl2AppPathPrefix` in `config.js` and the URL in the `<base/>` tag changes. The URL in the HTTP header contains the `/api` suffix, but that shouldn't end up in `gl2AppPathPrefix` in `config.js` and the `<base/>` tag. (at least as far as I understand it)

```
$ http localhost:9000/config.js
HTTP/1.1 200 OK
Content-Length: 118
Content-Type: application/javascript
Date: Fri, 03 Nov 2017 12:43:44 GMT
X-Graylog-Node-ID: 28164cbe-4ad9-4c9c-a76e-088655aa7889
window.appConfig = {
  gl2ServerUrl: 'http://127.0.0.1:9000/api/',
  gl2AppPathPrefix: '/',
  rootTimeZone: 'UTC',
};
```
```
$ http localhost:9000/config.js X-Graylog-Server-URL:http://devd.io:9000/api
HTTP/1.1 200 OK
Content-Length: 119
Content-Type: application/javascript
Date: Fri, 03 Nov 2017 12:43:50 GMT
X-Graylog-Node-ID: 28164cbe-4ad9-4c9c-a76e-088655aa7889
window.appConfig = {
  gl2ServerUrl: 'http://devd.io:9000/api/',
  gl2AppPathPrefix: '/api',
  rootTimeZone: 'UTC',
};
```

```
$ http localhost:9000 | fgrep '<base'
    <base href="http://127.0.0.1:9000/">
```
```
$ http localhost:9000 X-Graylog-Server-URL:http://devd.io:9000/api | fgrep '<base'
    <base href="http://devd.io:9000/api">
```

## Possible Solution

- Check if we really need the `<base/>` tag in the index template
- Fix changing of the `gl2AppPathPrefix` when `X-Graylog-Server-URL` is set

1. Run a production artifact build
2. Keep default `http_bind_address = 127.0.0.1:9000`
3. Access http://127.0.0.1:9000/
4. Access http://localhost:9000/ and see it fail

/cc @edmundoa @joschi 

## Your Environment

* Graylog Version: 3.0.0-SNAPSHOT (0c3c7f990)

kroepke:
Due to changes in the http config handling, the ability to run behind a proxy with a custom path prefix was lost.

This PR restores the application prefix handling.

fixes #4315

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
@@ -27,7 +27,6 @@ import org.graylog2.utilities.IpSubnet;
 import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 import javax.ws.rs.container.ContainerRequestContext;
-import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.SecurityContext;
 import java.net.URI;
@@ -112,7 +111,13 @@ public class RestTools {
                     .findFirst();
         }
 
-        return externalUri.orElse(defaultUri);
+        final URI uri = externalUri.orElse(defaultUri);
+
+        // Make sure we return an URI object with a trailing slash
+        if (!uri.toString().endsWith("/")) {
+            return URI.create(uri.toString() + "/");
+        }
+        return uri;
     }
 
     public static String getPathFromResource(Resource resource) {
--- a/graylog2-server/src/main/resources/web-interface/index.html.template
+++ b/graylog2-server/src/main/resources/web-interface/index.html.template
@@ -5,16 +5,15 @@
     <meta name="robots" content="noindex, nofollow">
     <meta charset="UTF-8">
     <title>${title}</title>
-    <base href="${baseUri}">
-    <link rel="shortcut icon" href="assets/favicon.png">
+    <link rel="shortcut icon" href="${appPrefix}/assets/favicon.png">
     ${foreach cssFiles cssFile}
-    <link href="assets/${cssfile}" rel="stylesheet">
+    <link href="${appPrefix}/assets/${cssfile}" rel="stylesheet">
     ${end}
   </head>
   <body>
-    <script src="config.js"></script>
+    <script src="${appPrefix}/config.js"></script>
     ${foreach jsFiles jsFile}
-    <script src="assets/${jsFile}"></script>
+    <script src="${appPrefix}/assets/${jsFile}"></script>
     ${end}
   </body>
 </html>
--- a/graylog2-server/src/main/resources/web-interface/index.html.template
+++ b/graylog2-server/src/main/resources/web-interface/index.html.template
@@ -5,15 +5,15 @@
     <meta name="robots" content="noindex, nofollow">
     <meta charset="UTF-8">
     <title>${title}</title>
-    <link rel="shortcut icon" href="${appPrefix}/assets/favicon.png">
+    <link rel="shortcut icon" href="${appPrefix}assets/favicon.png">
     ${foreach cssFiles cssFile}
-    <link href="${appPrefix}/assets/${cssfile}" rel="stylesheet">
+    <link href="${appPrefix}assets/${cssfile}" rel="stylesheet">
     ${end}
   </head>
   <body>
-    <script src="${appPrefix}/config.js"></script>
+    <script src="${appPrefix}config.js"></script>
     ${foreach jsFiles jsFile}
-    <script src="${appPrefix}/assets/${jsFile}"></script>
+    <script src="${appPrefix}assets/${jsFile}"></script>
     ${end}
   </body>
 </html>
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -139,7 +139,7 @@ plugin_dir = plugin
 # The maximum size of the HTTP request headers in bytes.
 #http_max_header_size = 8192
 
-# The size of the thread pool used exclusively for serving the HTTP interface and REST API.
+# The size of the thread pool used exclusively for serving the HTTP interface.
 #http_thread_pool_size = 16
 
 ################
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -139,9 +139,10 @@ plugin_dir = plugin
 # The maximum size of the HTTP request headers in bytes.
 #http_max_header_size = 8192
 
-# The size of the thread pool used exclusively for serving the HTTP interface.
+# The size of the thread pool used exclusively for serving the HTTP interface and REST API.
 #http_thread_pool_size = 16
 
+################
 # HTTPS settings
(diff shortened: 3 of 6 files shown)


bernd:
@kroepke Can we do this or something similar? This avoids entries in `index.html` that have a duplicate `/` in them. Like: `http://127.0.0.1//config.js`

This makes the proxy config requirements a little less strict.

```diff
diff --git graylog2-server/src/main/java/org/graylog2/rest/RestTools.java graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
index 664ccab0d..48cd85fe5 100644
--- graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
+++ graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
@@ -115,6 +114,16 @@ public class RestTools {
         return externalUri.orElse(defaultUri);
     }
 
+    public static String buildExternalUriStringWithoutTrailingSlash(@NotNull MultivaluedMap<String, String> httpHeaders, @NotNull URI defaultUri) {
+        final String uriString = buildExternalUri(httpHeaders, defaultUri).toString();
+
+        if (uriString.endsWith("/")) {
+            return uriString.substring(0, uriString.length() - 1);
+        }
+
+        return uriString;
+    }
+
     public static String getPathFromResource(Resource resource) {
         String path = resource.getPath();
         Resource parent = resource.getParent();
diff --git graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java
index 9ed5f69ef..a9da23b87 100644
--- graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java
+++ graylog2-server/src/main/java/org/graylog2/web/IndexHtmlGenerator.java
@@ -69,7 +69,7 @@ public class IndexHtmlGenerator {
                 .put("title", "Graylog Web Interface")
                 .put("cssFiles", cssFiles)
                 .put("jsFiles", sortedJsFiles)
-                .put("appPrefix", RestTools.buildExternalUri(headers, httpConfiguration.getHttpExternalUri()))
+                .put("appPrefix", RestTools.buildExternalUriStringWithoutTrailingSlash(headers, httpConfiguration.getHttpExternalUri()))
                 .build();
         return templateEngine.transform(template, model);
     }
```

kroepke:
@bernd that probably works, and in fact I had a sllightly different version of this previously but removed it because it wasn't strictly necessary.

Are the `//` still a problem?

bernd:
@kroepke The `//` are not a problem when I configure the proxy server accordingly.

The problem I see is, that we basically rely on the proxy server to normalize the URI. If the proxy server passes the URI unmodified to Graylog, it wouldn't work anymore.

In nginx it's basically `proxy_pass http://127.0.0.1:9000/;` vs `proxy_pass http://127.0.0.1:9000;`. If we use the second version without a trailing `/`, nginx passes the URL unmodified to Graylog.

If a user is using a proxy server that cannot normalize the URI or if the user is using a proxy server that doesn't provide access to such configuration options (e.g. Heroku, AWS, Kubernetes), Graylog will not work anymore. I checked the nginx-ingress controller in Kubernetes and it's using the `proxy_pass` directive without a trailing `/` to pass the URI unnormalized to the upstream.

That's why I think Graylog should behave correctly and not generating URIs that wouldn't work with a proxy that passes the URI without normalization. And I think that most proxy servers will not touch the URI by default.

bernd:
@kroepke As discussed, I updated this PR to fix the issues I reported earlier.

@joschi It would be cool if you can review and merge this. :smiley:

graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
```diff
@@ -112,7 +111,13 @@ public static URI buildExternalUri(@NotNull MultivaluedMap<String, String> httpH
                     .findFirst();
         }
 
-        return externalUri.orElse(defaultUri);
+        final URI uri = externalUri.orElse(defaultUri);
+
+        // Make sure we return an URI object with a trailing slash
+        if (!uri.toString().endsWith("/")) {
```
joschi: What happens if `externalUri` already has *multiple* trailing slashes, i. e. `http://example.org///´?

misc/graylog.conf
```diff
@@ -139,9 +139,10 @@ plugin_dir = plugin
 # The maximum size of the HTTP request headers in bytes.
 #http_max_header_size = 8192
 
-# The size of the thread pool used exclusively for serving the HTTP interface.
+# The size of the thread pool used exclusively for serving the HTTP interface and REST API.
```
joschi: FWIW, "HTTP interface" is not synonymous with "web interface".

graylog2-server/src/main/resources/web-interface/index.html.template
```diff
@@ -5,16 +5,15 @@
     <meta name="robots" content="noindex, nofollow">
     <meta charset="UTF-8">
     <title>${title}</title>
-    <base href="${baseUri}">
-    <link rel="shortcut icon" href="assets/favicon.png">
+    <link rel="shortcut icon" href="${appPrefix}assets/favicon.png">
```
joschi: Why not use `${appPrefix}/assets/favicon.png` here (and in the other `script` tags)?
Then we could simply strip all trailing slashes from `http_external_uri`.

graylog2-server/src/main/java/org/graylog2/rest/RestTools.java
bernd: In this case it doesn't add another trailing slash. But multiple trailing slashes are not cleaned up. That didn't change.

misc/graylog.conf
bernd: Aye, I misread that. :white_check_mark:

graylog2-server/src/main/resources/web-interface/index.html.template
bernd: I looked at the usages of `RestTools#buildExternalUri()` and saw that we expect the URI to have a trailing slash. That's why I decided to make sure a URI returned from `#buildExternalUri()` has a trailing slash and adjusted the template to remove the `/` to avoid duplicate ones.

PR #4464, opened 2018-01-08
Enabling No Response bot for repository.

dennisoelkers:
This bot is automatically closing issues which are marked to be in need
for more information, but this information was not provided in a
reasonable time frame.

More information about the bot is provided here:
https://probot.github.io/apps/no-response/

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,2 @@
+daysUntilClose: 30
+responseRequiredLabel: needs-input


.github/no-response.yml
```diff
@@ -0,0 +1,2 @@
+daysUntilClose: 30
```
joschi: 👍

.github/no-response.yml
```diff
@@ -0,0 +1,2 @@
...
+responseRequiredLabel: needs-input
```
joschi: Shouldn't we also provide a short comment via `closeComment`?

dennisoelkers: The default comment is pretty good imo:

```
  This issue has been automatically closed because there has been no response
  to our request for more information from the original author. With only the
  information that is currently in the issue, we don't have enough information
  to take action. Please reach out if you have or find the answers we need so
  that we can investigate further.
```
(from https://probot.github.io/apps/no-response/)

joschi: ✅

PR #4467, opened 2018-01-09
Change Permission Buttons from Toggle to Set and Clear

### LINKED ISSUE #3210: "Roles" UI/UX improvement: "Set/Clear all"
In the "Roles" UI we currently have as "mass operations" for "Filter Streams" only the operations "Toggle read permissions" and "Toggle edit permissions". This does not make any sense at all for me in the generic case. Who would want to "invert" permissions?

## Expected Behavior
I would expect that you could "Set" or "Clear" all permissions with one click, so please implement "Set read permissions", "Set edit permissions", "Clear read permissions", and "Clear edit permissions".

## Current Behavior
The difference is that instead of "toggling" an arbitrary existing set of permissions, that generally seems not to be useful, you can mass-enable or disable permissions.

## Your Environment

* Graylog Version: Graylog 2.1.2+50e449a

kmerz:
## Description
So I changed the label of the buttons from "Toggle" to
"Set" and "Clear" depending on which state the streams
permissions are. If all permissions are "Allow" the button
will read "Clear" and remove the permissions. If one or
more will have no "Allow" permission, the button will read
"Set" and will set the permission to "Allow" for all selected
streams.

## Motivation and Context
After selecting two streams (or dashboards) in the Edit Role
view, two buttons appeared in the header of the table.
"Toggle read permissions" and "Toggle edit permissions"
which would invert the permissions of a selected streams.

The user would rather set or clear the permissions of the
selected streams.

Fixes #3210

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/users/PermissionSelector.jsx
+++ b/graylog2-web-interface/src/components/users/PermissionSelector.jsx
@@ -19,19 +19,27 @@ const PermissionSelector = React.createClass({
       const isRead = this.props.permissions.contains(`streams:read:${stream.id}`);
       const isEdit = this.props.permissions.contains(`streams:edit:${stream.id}`);
       return (<ButtonGroup bsSize="small">
-        <Button bsStyle={isRead ? 'info' : 'default'} onClick={() => this._toggleStreamReadPermissions(stream)}
+        <Button bsStyle={isRead ? 'info' : 'default'}
+                onClick={() => this._toggleStreamReadPermissions(stream)}
                 active={isRead}>Allow reading</Button>
-        <Button bsStyle={isEdit ? 'info' : 'default'} onClick={() => this._toggleStreamEditPermissions(stream)}
+        <Button bsStyle={isEdit ? 'info' : 'default'}
+                onClick={() => this._toggleStreamEditPermissions(stream)}
                 active={isEdit}>Allow editing</Button>
       </ButtonGroup>);
     };
 
     const multiStreamButtons = (streamIds) => {
+      const selectedStreams = this.props.streams.filter(stream => streamIds.contains(stream.id));
+      const allRead = selectedStreams.every(stream => this.props.permissions.contains(`streams:read:${stream.id}`));
+      const allEdit = selectedStreams.every(stream => this.props.permissions.contains(`streams:edit:${stream.id}`));
+      const readActionLabel = allRead ? 'Clear' : 'Set';
+      const editActionLabel = allEdit ? 'Clear' : 'Set';
+
       return (
         <div className="pull-right" style={{ marginTop: 10, marginBottom: 10 }}>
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/users/PermissionSelector.jsx
```diff
@@ -90,67 +105,67 @@ const PermissionSelector = React.createClass({
...
+  _setReadPermissions(target, idList, allUnset = true) {
     let added = Immutable.Set.of();
     let deleted = Immutable.Set.of();
 
     idList.forEach((id) => {
       const readTarget = `${target}:read:${id}`;
       const editTarget = `${target}:edit:${id}`;
 
-      if (this.props.permissions.contains(readTarget)) {
+      if (this.props.permissions.contains(readTarget) && allUnset) {
```
edmundoa: This is a small (and subjective) thing, but in my opinion `shouldClear` or `clearPermissions` would be easier to read than `allUnset`.

graylog2-web-interface/src/components/users/PermissionSelector.test.jsx
```diff
@@ -0,0 +1,145 @@
...
+
+import PermissionSelector from 'components/users/PermissionSelector';
+
+/* https://github.com/facebook/react/issues/7371
+ *
+ * findDomNode with refs is not supported by the react-test-renderer.
+ * So we need to mock the findDOMNode function for TableList respectievly
+ * for its child component TypeAheadDataFilter.
+ */
+jest.mock('react-dom', () => ({
```
edmundoa: As we may need this function in other tests, maybe we can extract it into a mocking helper in `graylog2-web-interface/test/helpers/mocking`. We can also ignore it for now, but I guess it's likely that we end up copy-pasting this code here and there :)

dennisoelkers:
An interesting effect of the PR that I encountered is that you cannot unset the "Allow editing" permission once it is set for an entity. When you click on "Allow editing" on e.g. a stream, clicking it again does not remove the permission. Instead you have to click on "Allow reading", which removes both.

graylog2-web-interface/src/components/users/PermissionSelector.jsx
```diff
@@ -90,67 +105,67 @@ const PermissionSelector = React.createClass({
...
-  _toggleDashboardReadPermissions(dashboard) {
-    this._toggleReadPermissions('dashboards', Immutable.Set.of(dashboard.id));
+  _setDashboardReadPermissions(dashboard) {
+    this._setReadPermissions('dashboards', Immutable.Set.of(dashboard.id));
   },
 
-  _toggleDashboardEditPermissions(dashboard) {
-    this._toggleEditPermissions('dashboards', Immutable.Set.of(dashboard.id));
+  _setDashboardEditPermissions(dashboard) {
+    this._setEditPermissions('dashboards', Immutable.Set.of(dashboard.id));
```
dennisoelkers: The naming of the single permission change functions was better before the change, because they are not setting an explicit permission, they are toggling a given permission.

PR #4483, opened 2018-01-16
Hide replay button if user has no permission

### LINKED ISSUE #1874: "Replay search" should be evaluated against stream acls
Hello,

On Graylog 1.3..3, "replay search" button on dashboard widgets are always visible.
If user doesn't have access to stream, it raises an ACL error.

I think more user friendly to display this button only if connected user have also the stream read access.

Thanks.

### LINKED ISSUE #4120: remove "play" icon from no-search dashboard
Hi there

We have a "public" dashboard that users who are not in any real Role are associated with. It shows them some graphs, but they don't have any stream access. (ie all streams and dashboard besides "Public" are restricted to Roles)

When such an 'unprivileged' user sees the dashboard, the "play" button is still visible. If they mouse-over it, it pops up 'the stream is not available' - which is correct. However, since it doesn't work, wouldn't it be best to *remove* the 'play' icon? Otherwise it's confusing - some people complain that it's broken when it's actually working as expected

 
1.
2.
3.
4.

## Your Environment

* Graylog Version: 2.3.1
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kmerz:
## Description
A widget on a dashboard has a play button which directs the
user to the search on which the widget is based on.
If the user had no permission to look on the streams (no read
permission on the stream) he got the message:

'The stream is not available, cannot replay search.'

Some user thought that the button was broken, since they assumed
to trigger a update with this button.

To fix the problem we remove the button for users without
the necessary permissions.

Fixes #4120 
Fixes #1874

## How Has This Been Tested?
The tests generate snapshots which i diffed manually to ensure that
the expected result was rendered.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/widgets/Widget.jsx
+++ b/graylog2-web-interface/src/components/widgets/Widget.jsx
@@ -1,17 +1,17 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import $ from 'jquery';
-import { PluginStore } from 'graylog-web-plugin/plugin';
+import Reflux from 'reflux';
+
 import { WidgetConfigModal, WidgetEditConfigModal, WidgetFooter, WidgetHeader, WidgetVisualizationNotFound } from 'components/widgets';
+import { PluginStore } from 'graylog-web-plugin/plugin';
 import StoreProvider from 'injection/StoreProvider';
-import ActionsProvider from 'injection/ActionsProvider';
 import Routes from 'routing/Routes';
 import EventHandlersThrottler from 'util/EventHandlersThrottler';
-import Reflux from 'reflux';
 import PermissionsMixin from 'util/PermissionsMixin';
+import CombinedProvider from '../../injection/CombinedProvider';
 
-const WidgetsStore = StoreProvider.getStore('Widgets');
-const WidgetsActions = ActionsProvider.getActions('Widgets');
+const { WidgetsStore, WidgetsActions } = CombinedProvider.get('Widgets');
 const CurrentUserStore = StoreProvider.getStore('CurrentUser');
 
 const Widget = React.createClass({
--- a/graylog2-web-interface/src/components/widgets/Widget.jsx
+++ b/graylog2-web-interface/src/components/widgets/Widget.jsx
@@ -216,11 +216,8 @@ const Widget = React.createClass({
                              onUpdate={this.updateWidget} />
     );
 
-    let disabledReplay = false;
-    if (this.props.streamIds != null && this.props.widget.config.stream_id &&
-        !this.props.streamIds[this.props.widget.config.stream_id]) {
-      disabledReplay = true;
-    }
+    const disabledReplay = this.props.streamIds != null && this.props.widget.config.stream_id &&
+      !this.props.streamIds[this.props.widget.config.stream_id];
     return (
       <div ref="widget" className="widget" data-widget-id={this.props.widget.id}>
         <WidgetHeader ref="widgetHeader"
--- a/graylog2-web-interface/src/components/widgets/Widget.jsx
+++ b/graylog2-web-interface/src/components/widgets/Widget.jsx
@@ -1,21 +1,22 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import ReactDOM from 'react-dom';
 import $ from 'jquery';
 import { PluginStore } from 'graylog-web-plugin/plugin';
-
 import { WidgetConfigModal, WidgetEditConfigModal, WidgetFooter, WidgetHeader, WidgetVisualizationNotFound } from 'components/widgets';
-
 import StoreProvider from 'injection/StoreProvider';
-const WidgetsStore = StoreProvider.getStore('Widgets');
-
 import ActionsProvider from 'injection/ActionsProvider';
-const WidgetsActions = ActionsProvider.getActions('Widgets');
-
 import Routes from 'routing/Routes';
 import EventHandlersThrottler from 'util/EventHandlersThrottler';
+import Reflux from 'reflux';
+import PermissionsMixin from 'util/PermissionsMixin';
+
+const WidgetsStore = StoreProvider.getStore('Widgets');
+const WidgetsActions = ActionsProvider.getActions('Widgets');
+const CurrentUserStore = StoreProvider.getStore('CurrentUser');
 
 const Widget = React.createClass({
+  mixins: [Reflux.connect(CurrentUserStore), PermissionsMixin],
+
   propTypes: {
     widget: PropTypes.object.isRequired,
     dashboardId: PropTypes.string.isRequired,
@@ -69,14 +70,14 @@ const Widget = React.createClass({
     return ('stream_id' in this.props.widget.config) && (this.props.widget.config.stream_id !== null);
   },
   _getWidgetNode() {
-    return ReactDOM.findDOMNode(this.refs.widget);
+    return this.node;
   },
   _loadValue() {
     if (this.state.deleted || (this.state.result !== undefined && !this.props.shouldUpdate)) {
       return;
     }
 
-    const width = this.refs.widget.clientWidth;
+    const width = this.node.clientWidth;
 
     WidgetsStore.loadValue(this.props.dashboardId, this.props.widget.id, width).then((value) => {
       // Avoid updating state if the result didn't change
@@ -111,6 +112,7 @@ const Widget = React.createClass({
   },
   _calculateWidgetSize() {
     const $widgetNode = $(this._getWidgetNode());
+    if (!$widgetNode) { return; }
     // .height() give us the height of the whole widget without counting paddings, we need to remove the size
     // of the header and footer from that.
     const availableHeight = $widgetNode.height() - (this.WIDGET_HEADER_HEIGHT + this.WIDGET_FOOTER_HEIGHT);
@@ -189,9 +191,10 @@ const Widget = React.createClass({
     this.refs.editModal.open();
   },
   updateWidget(newWidgetData) {
-    newWidgetData.id = this.props.widget.id;
+    const realNewWidgetData = newWidgetData;
+    realNewWidgetData.id = this.props.widget.id;
 
-    WidgetsStore.updateWidget(this.props.dashboardId, newWidgetData);
+    WidgetsStore.updateWidget(this.props.dashboardId, realNewWidgetData);
   },
   deleteWidget() {
     if (window.confirm(`Do you really want to delete "${this.props.widget.description}"?`)) {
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -216,10 +216,10 @@ const Widget = React.createClass({
                              onUpdate={this.updateWidget} />
     );
 
-    let disabledTooltip = null;
+    let disabledReplay = false;
     if (this.props.streamIds != null && this.props.widget.config.stream_id &&
         !this.props.streamIds[this.props.widget.config.stream_id]) {
-      disabledTooltip = 'The stream is not available, cannot replay search.';
+      disabledReplay = true;
```
dennisoelkers: FWIW, I would change this to 
`const disabledReplay = this.props.streamIds != null && this.props.widget.config.stream_id && !this.props.streamIds[this.props.widget.config.stream_id];`

to get rid of the `let`/`if` and make it a bit more concise.

graylog2-web-interface/src/components/widgets/WidgetFooter.jsx
```diff
@@ -47,25 +47,16 @@ const WidgetFooter = React.createClass({
     }
 
     // if we have a tooltip, we disable the button link and instead show a tooltip on hover
-    const title = this.props.replayToolTip ? null : 'Replay search';
-    const href = this.props.replayToolTip ? null : this.props.replayHref;
-    let replay = (
-      <Button bsStyle="link" className="btn-text" title={title} href={href}>
-        <i className="fa fa-play" />
-      </Button>
+    let replay = this.props.replayDisabled ? <div /> : (
```
dennisoelkers: Is there a reason why you used `let` instead of `const` here?

graylog2-web-interface/src/components/widgets/WidgetFooter.test.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import WidgetFooter from 'components/widgets/WidgetFooter';
+
+describe('<WidgetFooter />', () => {
+  const date = new Date('December 17, 1995 03:24:00');
+  const lockWidget = true;
+  const disableReplay = true;
```
dennisoelkers: This is unneeded imo.

graylog2-web-interface/src/components/widgets/WidgetFooter.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  it('should render a widget footer locked and disabled replay', () => {
+    const wrapper = renderer.create(<WidgetFooter
+      locked={lockWidget}
+      onShowConfig={() => {}}
+      onEditConfig={() => {}}
+      onDelete={() => {}}
+      replayHref={'http://example.org'}
+      replayDisabled={disableReplay}
```
dennisoelkers: You can just use `replayDisabled` here.

graylog2-web-interface/src/components/widgets/WidgetFooter.test.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import WidgetFooter from 'components/widgets/WidgetFooter';
+
+describe('<WidgetFooter />', () => {
+  const date = new Date('December 17, 1995 03:24:00');
+  const lockWidget = true;
```
dennisoelkers: This is unneeded imo.

graylog2-web-interface/src/components/widgets/WidgetFooter.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+  const disableReplay = true;
+
+  it('should render a widget footer locked and enabled replay', () => {
+    const wrapper = renderer.create(<WidgetFooter
+      locked={lockWidget}
```
dennisoelkers: As `lockWidget` is always `true` in this scope, we can just write `locked` here.

graylog2-web-interface/src/components/widgets/WidgetFooter.test.jsx
```diff
@@ -0,0 +1,52 @@
...
+      replayDisabled={false}
+      calculatedAt={date.getTime()}
+      error={{}}
+      errorMessage={''} />);
+    expect(wrapper.toJSON()).toMatchSnapshot();
+  });
+
+  it('should render a widget footer locked and disabled replay', () => {
+    const wrapper = renderer.create(<WidgetFooter
+      locked={lockWidget}
```
dennisoelkers: As `lockWidget` is always `true` in this scope, we can just write `locked` here.

graylog2-web-interface/src/components/widgets/WidgetFooter.jsx
```diff
@@ -47,25 +47,16 @@ const WidgetFooter = React.createClass({
     }
 
     // if we have a tooltip, we disable the button link and instead show a tooltip on hover
-    const title = this.props.replayToolTip ? null : 'Replay search';
-    const href = this.props.replayToolTip ? null : this.props.replayHref;
-    let replay = (
-      <Button bsStyle="link" className="btn-text" title={title} href={href}>
-        <i className="fa fa-play" />
-      </Button>
+    const replay = this.props.replayDisabled ? <div /> : (
```
dennisoelkers: Is there a reason why you have chosen to render an empty `<div>` instead of `null`? We do the latter in most of the code base when we want to render nothing but need to assign a value to a variable. If there is a reason why `null` is a bad idea in this case, we should still use a `<span>` instead of a `<div>`, as the former has no semantic implications.

graylog2-web-interface/src/components/widgets/WidgetFooter.jsx
```diff
@@ -47,25 +47,16 @@ const WidgetFooter = React.createClass({
...
     // if we have a tooltip, we disable the button link and instead show a tooltip on hover
-    const title = this.props.replayToolTip ? null : 'Replay search';
-    const href = this.props.replayToolTip ? null : this.props.replayHref;
-    let replay = (
-      <Button bsStyle="link" className="btn-text" title={title} href={href}>
-        <i className="fa fa-play" />
-      </Button>
+    const replay = this.props.replayDisabled ? null : (
+      <div className="widget-replay">
+        <Button bsStyle="link" className="btn-text" title={'Replay search'} href={this.props.replayHref}>
```
edmundoa: In the `title` prop we should use double-quotes which read a bit better, e.g. `title="Replay search"`.

graylog2-web-interface/src/components/widgets/WidgetFooter.jsx
```diff
@@ -47,25 +47,16 @@ const WidgetFooter = React.createClass({
     }
 
     // if we have a tooltip, we disable the button link and instead show a tooltip on hover
```
edmundoa: The comment is outdated and can also go now.

edmundoa:
By the way, this will also fix https://github.com/Graylog2/graylog2-server/issues/1874 🎉

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -204,37 +207,34 @@ const Widget = React.createClass({
...
-    let disabledTooltip = null;
-    if (this.props.streamIds != null && this.props.widget.config.stream_id &&
-        !this.props.streamIds[this.props.widget.config.stream_id]) {
-      disabledTooltip = 'The stream is not available, cannot replay search.';
-    }
+    const disabledReplay = !((!this.props.widget.config.stream_id &&
+      this.isPermitted(this.state.currentUser.permissions,
+        ['searches:absolute', 'searches:keyword', 'searches:relative'])) ||
+      (this.props.widget.config.stream_id && this.props.streamIds != null &&
+      this.props.streamIds[this.props.widget.config.stream_id]));
```
edmundoa: I personally find this sentence hard to read. Could we maybe extract it into a couple of variables or functions? In that way we can split the condition and be able to understand it easier.

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -204,37 +207,34 @@ const Widget = React.createClass({
...
-                         dashboardId={this.props.dashboardId}
+      <WidgetConfigModal dashboardId={this.props.dashboardId}
                          widget={this.props.widget}
                          boundToStream={this._isBoundToStream()} />
     );
 
     const editConfigModal = (
-      <WidgetEditConfigModal ref="editModal"
-                             widget={this.props.widget}
+      <WidgetEditConfigModal widget={this.props.widget}
```
edmundoa: We still use the `editModal` ref to open the edit modal dialog, so it can't be removed.

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -204,37 +207,34 @@ const Widget = React.createClass({
       return <span />;
     }
     const showConfigModal = (
-      <WidgetConfigModal ref="configModal"
-                         dashboardId={this.props.dashboardId}
+      <WidgetConfigModal dashboardId={this.props.dashboardId}
```
edmundoa: We still use `configModal` ref to open the info modal dialog, so it can't be removed.

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -1,21 +1,22 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import ReactDOM from 'react-dom';
 import $ from 'jquery';
 import { PluginStore } from 'graylog-web-plugin/plugin';
-
```
dennisoelkers: We tend to separate imports of modules and imports of components in the same code base by an empty line.

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -1,21 +1,22 @@
...
 import StoreProvider from 'injection/StoreProvider';
-const WidgetsStore = StoreProvider.getStore('Widgets');
-
 import ActionsProvider from 'injection/ActionsProvider';
-const WidgetsActions = ActionsProvider.getActions('Widgets');
-
 import Routes from 'routing/Routes';
 import EventHandlersThrottler from 'util/EventHandlersThrottler';
+import Reflux from 'reflux';
```
dennisoelkers: Module imports are placed at the top of the imports. Usually it's react, followed by reflux (if required), then others.

graylog2-web-interface/src/components/widgets/Widget.jsx
```diff
@@ -1,21 +1,22 @@
...
+import PermissionsMixin from 'util/PermissionsMixin';
+
+const WidgetsStore = StoreProvider.getStore('Widgets');
+const WidgetsActions = ActionsProvider.getActions('Widgets');
```
dennisoelkers: You can reduce these two lines by using the `CombinedProvider`.

PR #4499, opened 2018-01-22
Add support for Java 8 lambda expressions to MetricUtils#map()

### LINKED ISSUE #4493: Progress bars on node and node details page are broken
## Expected Behavior

The progress bars for JVM memory and buffers should show the current values.

## Current Behavior

The progress bars don't change and don't show any values. The metrics response includes the JVM data, though.

p

1. Navigate to "System / Nodes"
2. Watch JVM memory progress bar
3. Navigate to a node details page
4. Watch the progress bars

## Your Environment

* Graylog Version: 3.0.0-SNAPSHOT+6762727
* Browser version: Chrome 63.0.3239.132

joschi:
Lambda expressions in Java 8+ are synthetic classes and return that synthetic class when `#getClass()` is being used, so the "type discovery" in `MetricUtils#map()` didn't work as expected.

Also see https://stackoverflow.com/questions/16827262/how-will-java-lambda-functions-be-compiled and https://gist.github.com/dgageot/5840050 for an example.

Fixes #4493

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/metrics/MetricUtils.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/metrics/MetricUtils.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.shared.metrics;
 
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/shared/metrics/MetricUtils.java
```diff
@@ -72,38 +73,41 @@
...
+        } else if(metric instanceof Counter) {
+            metricMap.put("metric", ((Counter) metric).getCount());
+            metricMap.put("type", "counter");
+        } else if(metric instanceof Gauge) {
+            metricMap.put("metric", ((Gauge) metric).getValue());
+            metricMap.put("type", "gauge");
         } else {
             metricMap.put("metric", metric);
+            // best guess
+            metricMap.put("type", "gauge");
```
joschi: Not sure if we should keep this "catch all" or simply return an error. 🤔

bernd: What other types are there? If there aren't any other metric types, I think we should throw an error.

joschi: AFAIK there aren't any others used by Graylog.

The Graylog web interface also only has support for Counter, Gauge, Histogram, Meter, and Timer.

https://github.com/Graylog2/graylog2-server/blob/0f97411137b2f5b36f3a24d0a74c409ef469df39/graylog2-web-interface/src/components/metrics/MetricDetails.jsx#L27-L36

PR #4507, opened 2018-01-23
Upgrade jQuery to 3.3.1

edmundoa:
Also bump dependencies relying on jQuery (currently only toastr has a new bugfix version).

As this is a long version jump and there are many changes in jQuery as you can see in the release notes for [3.0](https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/), [3.1](https://blog.jquery.com/2016/07/07/jquery-3-1-0-released-no-more-silent-errors/), [3.2](https://blog.jquery.com/2017/03/16/jquery-3-2-0-is-out/), and [3.3](https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/).

Fortunately for us, 3.0 was developed with backwards compatibility in mind, so most of the changes that affect us are deprecations. On the other hand, most of the jQuery code we run is coming through 3rd party plugins, so we may have some problems to get fixes in them.

As recommended, I'm running the web interface with the [jQuery Migrate plugin](https://github.com/jquery/jquery-migrate/#README) to be able to see deprecations and changes we need to address.

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -138,7 +138,7 @@
     "yargs": "^10.0.3"
   },
   "resolutions": {
-    "@types/jquery": "2.0.48",
-    "jquery": "2.1.4"
+    "@types/jquery": "3.3.0",
+    "jquery": "3.3.1"
   }
 }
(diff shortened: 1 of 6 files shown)


edmundoa:
Apart from the changes in the PR, I saw deprecation warnings using the jQuery migrate plugin in toastr, typeahead, and jquery-ui. So far we cannot address those warnings since we are already using the latest version of those libraries.

From my point of view this is ready to be reviewed, I think we should focus our efforts in moving away from the few jQuery plugins we use instead of fixing those deprecations.

kmerz:
Like what you did here :)

PR #4508, opened 2018-01-23
Fixed CSS rules for main-nav dropdown

razvanphp:
## Description
After merging graylog-web interface into the server and React rewrite, the main navbar is not wrapped in `div#navigation-bar` anymore, so those CSS rules don't match. This triggers the default bootstrap rules with bigger font (14px vs 12px) and smaller paddings.

## Motivation and Context

I've been in a long-term relationship with Graylog, so the first thing I noticed when we were finally able to upgrade, was the different dropdown appearance... this will make it beautiful again.

## Screenshots (1.3.4 vs 2.4.1):

screen shot 2018-01-24 at 00 29 38
screen shot 2018-01-24 at 00 29 54

## Types of changes

### DIFF
--- a/graylog2-web-interface/public/stylesheets/graylog2.less
+++ b/graylog2-web-interface/public/stylesheets/graylog2.less
@@ -2234,20 +2234,20 @@ dl.message-details dd div.message-field-actions {
   z-index: 20;
 }
 
-div#navigation-bar ul.dropdown-menu li a {
+nav.navbar-fixed-top ul.dropdown-menu li a {
   font-size: 12px;
 }
 
-div#navigation-bar ul.dropdown-menu {
+nav.navbar-fixed-top ul.dropdown-menu {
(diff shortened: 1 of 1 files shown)


dennisoelkers:
Thanks for the observation, @razvanphp!

I like the change/revert, for illustration, these are the differences:

Before:

screen shot 2018-01-24 at 08 14 53

After:

screen shot 2018-01-24 at 08 14 28

dennisoelkers:
Anyone from @Graylog2/engineers (and everybody else on github as well) who has an opinion is invited: Please add a 👍 if you like this change or add a 👎 if you prefer the current styling. This PR will be merged or closed depending on the outcome in about seven days.

razvanphp:
I thought this was a bug, since the styles were there, just not matching since the FE rewrite.

Hope to see this back in.

Issue #4509, opened 2018-01-24
Quick Values aggregation converts timestamp to long


lennartkoopmann:
When running a stacked quickvalues analysis, where the second field is a timestamp, the values are returned as `long` / UNIX timestamps and no a human readable format like ISO8601.

![screenshot from 2018-01-23 19-56-22](https://user-images.githubusercontent.com/35022/35310609-d11b7a64-0077-11e8-95a5-0430939099e0.png)

![screenshot from 2018-01-23 19-56-44](https://user-images.githubusercontent.com/35022/35310612-d30ae0ee-0077-11e8-82c1-842aa21385bd.png)

## Steps to Reproduce (for bugs)

1. Run a "Quick Values" analysis on any field
2. Add "timestamp" as stacked field

## Your Environment

* Graylog Version: 2.4.0
* Elasticsearch Version: 5.6.4

lennartkoopmann:
In fact, this happens for a single quickvalues analysis on timestamp fields, too.

edmundoa:
@lennartkoopmann this is related to https://github.com/Graylog2/graylog2-server/issues/4288, which is already fixed in master. I guess the question is to see if this is important enough to be included in 2.4.2 /cc: @bernd

bernd:
@edmundoa This should go into 2.4.2, yes. :+1:

PR #4519, opened 2018-01-24
Fix timezone for message list add to query in field timestamp

kroepke:
Refs #4509


kroepke:
Needs backport to 2.4

PR #4521, opened 2018-01-25
Display Timestamp in User Timezone

### LINKED ISSUE #2689: Graylog2 shows different timestamp
## Context

I have install the graylog2 on linux redhat 7,everything is fine but i have a small promble about the 'timestamp'.
I set timezone "root_timezone = Asia/Hong_Kong",then i can see the timestamp shows the correct time in the dashboards,but when i click one log and shows the details of the log,I find that the timestamp are different,the outside timestamp shows the correct time(Asia/Hong_Kong) and there shows a incorrect timestamp(UTC time,not what i want),if i export as csv,the excel show the timestamp is UTC time,I wish both timestamp show the Asia/Hong_Kong time ,but i can't find anyother place can set that.
how can i fixup that?
timestamp
setting
## Your Environment

- Graylog Version:graylog-server-2.0.3-1.noarch
- Elasticsearch Version:elasticsearch-2.3.4-1.noarch
- MongoDB Version:mongodb-org-server-3.2.8-1.el7.x86_64
- Operating System:linux redhat 7
- Browser version:firefox

kmerz:
## Description
Display timestamps on the search page in the timezone of the user. The timezone of the user
is either the user setting or the timezone of the browser.

Additionally make it possible for a user to unset his timezone setting so the Graylog gui
can use the browser timezone.

## Motivation and Context
The user saw timestamps in UTC even though he set his browser and user setting to
his timezone.

## How Has This Been Tested?
Started browser in timezone US Pacific with:
 TZ="US/Pacific" google-chrome "--user-data-dir=$HOME/chrome-profile"
Set and unset the timezone of the user.

Partially Fixes #2689 
For the CSV Export I'll open an other issue.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
@@ -273,10 +273,9 @@ public class UserImpl extends PersistedImpl implements User {
         DateTimeZone dateTimeZone = null;
         if (timeZone != null) {
             try {
-                dateTimeZone = DateTimeZone.forID(firstNonNull(timeZone, DateTimeZone.UTC.getID()));
+                dateTimeZone = DateTimeZone.forID(timeZone);
             } catch (IllegalArgumentException e) {
-                LOG.info("Invalid timezone \"{}\", falling back to UTC.", timeZone);
-                dateTimeZone = DateTimeZone.UTC;
+                LOG.error("Invalid timezone \"{}\", falling back to UTC.", timeZone);
             }
         }
         setTimeZone(dateTimeZone);
--- a/graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
@@ -270,20 +270,21 @@ public class UserImpl extends PersistedImpl implements User {
 
     @Override
     public void setTimeZone(final String timeZone) {
-        DateTimeZone dateTimeZone;
-        try {
-            dateTimeZone = DateTimeZone.forID(firstNonNull(timeZone, DateTimeZone.UTC.getID()));
-        } catch (IllegalArgumentException e) {
-            LOG.info("Invalid timezone \"{}\", falling back to UTC.", timeZone);
-            dateTimeZone = DateTimeZone.UTC;
+        DateTimeZone dateTimeZone = null;
+        if (timeZone != null) {
+            try {
+                dateTimeZone = DateTimeZone.forID(firstNonNull(timeZone, DateTimeZone.UTC.getID()));
+            } catch (IllegalArgumentException e) {
+                LOG.info("Invalid timezone \"{}\", falling back to UTC.", timeZone);
+                dateTimeZone = DateTimeZone.UTC;
+            }
         }
-
         setTimeZone(dateTimeZone);
     }
 
     @Override
     public void setTimeZone(final DateTimeZone timeZone) {
-        fields.put(TIMEZONE, timeZone.getID());
+        fields.put(TIMEZONE, timeZone == null ? null : timeZone.getID());
     }
 
     @Override
--- a/graylog2-web-interface/src/components/search/MessageDetail.jsx
+++ b/graylog2-web-interface/src/components/search/MessageDetail.jsx
@@ -23,7 +23,7 @@ const MessageDetail = React.createClass({
     expandAllRenderAsync: PropTypes.bool,
     showTimestamp: PropTypes.bool,
     disableFieldActions: PropTypes.bool,
-    possiblyHighlight: PropTypes.func,
+    renderForDisplay: PropTypes.func,
     inputs: PropTypes.object,
     nodes: PropTypes.object,
     message: PropTypes.object,
@@ -263,7 +263,7 @@ const MessageDetail = React.createClass({
         <Col md={9}>
           <div ref="messageList">
             <MessageFields message={this.props.message}
-                           possiblyHighlight={this.props.possiblyHighlight}
+                           renderForDisplay={this.props.renderForDisplay}
                            disableFieldActions={this.props.disableFieldActions}
                            customFieldActions={this.props.customFieldActions}
                            showDecoration={this.state.showOriginal}
--- a/graylog2-web-interface/src/components/search/MessageTableEntry.css
+++ b/graylog2-web-interface/src/components/search/MessageTableEntry.css
@@ -1,0 +1,3 @@
+:local .timezoneInfo {
+    color: #cccccc;
+}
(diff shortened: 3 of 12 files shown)


graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
```diff
@@ -270,20 +270,21 @@ public DateTimeZone getTimeZone() {
...
-        } catch (IllegalArgumentException e) {
-            LOG.info("Invalid timezone \"{}\", falling back to UTC.", timeZone);
-            dateTimeZone = DateTimeZone.UTC;
+        DateTimeZone dateTimeZone = null;
+        if (timeZone != null) {
+            try {
+                dateTimeZone = DateTimeZone.forID(firstNonNull(timeZone, DateTimeZone.UTC.getID()));
+            } catch (IllegalArgumentException e) {
+                LOG.info("Invalid timezone \"{}\", falling back to UTC.", timeZone);
+                dateTimeZone = DateTimeZone.UTC;
```
dennisoelkers: I do not think we should do this. Using an incorrect timezone name indicates a programming error/misuse of the API, so we should handle this accordingly and return an error to the caller without changing the timezone of the user. If we leave it that way, the caller has no way to find out from the result of the call if the timezone was set properly. The message logged is also not helpful, it is only visible to the server admin and does not even include the user name.

I know this is the way the code was before, but if you are at it, I would be glad if you could improve that part too :)

graylog2-web-interface/src/components/search/MessageDetail.jsx
```diff
@@ -23,7 +23,7 @@ const MessageDetail = React.createClass({
     expandAllRenderAsync: PropTypes.bool,
     showTimestamp: PropTypes.bool,
     disableFieldActions: PropTypes.bool,
-    possiblyHighlight: PropTypes.func,
+    renderForDisplay: PropTypes.func,
```
dennisoelkers: :+1: for prop rename!

graylog2-web-interface/src/components/search/MessageTableEntry.jsx
```diff
@@ -102,6 +113,12 @@ const MessageTableEntry = React.createClass({
   _toggleDetail() {
     this.props.toggleDetail(`${this.props.message.index}-${this.props.message.id}`);
   },
+
+  _toUserTime(value) {
+    const dateTime = new DateTime(value);
+    return dateTime.toString(DateTime.Formats.TIMESTAMP);
```
dennisoelkers: Maybe it makes sense to wrap the `value` in a `Timestamp` component instead and pass the timezone of the user as a prop? This way we have a more central handling of the time conversion.

edmundoa: After the changes the `timestamp` field in messages is in the user defined time zone (or the one set in the browser), which is what the user expects. Unfortunately, other timestamps may still come in UTC, e.g. if there's one in the `full_message` field. Additionally, when adding the `timestamp` field to the search query, the value we set there is using UTC (because we already format the time to be usable against ES), and I find that a bit confusing. Here's an screenshot to illustrate the problem:

image 2018-01-26 13-44-46

To reduce this confusion I would use `DateTime.Formats.TIMESTAMP_TZ` to format the `timestamp` field, making the user aware that this date time is displayed in a certain time zone, and others may not. We could also add a little informative text somewhere next to the value, but only if we find a way while not making it too prominent.

dennisoelkers: Yes, I agree. This would also mean that for consistency we should format the timestamp column in the message table (!= message details) accordingly.

kmerz: > Maybe it makes sense to wrap the value in a Timestamp component instead and pass the timezone of the user as a prop?
I do not know how to archive this without breaking `possiblyHighlight` which needs a String to operate.

> We could also add a little informative text somewhere next to the value
I would like that. At first I thought about a mouseover popup, but you would need to know that, to get the information.
Then I thought about to add the text somewhere on the right side in bright gray but that seems to distract more then I thought :(

edmundoa:
I cannot add a comment on https://github.com/Graylog2/graylog2-server/pull/4521#discussion_r164436933, so I'll continue the conversation here. Something like that is what we currently do with decorated fields, as you can see in the `severity` field here:

screen shot 2018-01-29 at 15 51 01

I think the approach you mentioned can be good. I think we could add a small information icon (from [fontawesome icon](http://fontawesome.io/icons)) in a soft colour next to the `timestamp`'s `<dt>` element that pops over when you hover or click there. I would use a react-bootstrap `Popover` element with the information. That looks a bit more appealing than the usual `title` popover you get from the browser.

kmerz:
Tada!
graylog - search 1

graylog2-web-interface/src/components/search/MessageTableEntry.jsx
```diff
@@ -1,10 +1,14 @@
...
 import React from 'react';
 import Immutable from 'immutable';
+import { Popover, OverlayTrigger } from 'react-bootstrap';
 
 import MessageDetail from './MessageDetail';
 import { Timestamp } from 'components/common';
 import StringUtils from 'util/StringUtils';
+import DateTime from 'logic/datetimes/DateTime';
+
+const style = require('!style!css!./MessageTableEntry.css');
```
edmundoa: This produces an [`import/no-webpack-loader-syntax`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-webpack-loader-syntax.md) error in ESLint. You can replace it with a normal import:
`import style from './MessageTableEntry.css';`

graylog2-web-interface/src/components/search/MessageTableEntry.css
```diff
@@ -0,0 +1,3 @@
+:local .timezoneInfo {
+    color: #cccccc;
```
edmundoa: This is highly personal, but you can play with it and see how you like it:

I would make the icon stand a little bit more, maybe using `#aaa`, as the decorator marker. I also think the icon is too close to the timestamp, so I would add a slight `margin-left` to avoid that. This is what I used:

```
:local .timezoneInfo {
    color: #aaa;
    margin-left: 0.2em;
}
```

Before:
screen shot 2018-01-30 at 13 45 05

After:
screen shot 2018-01-30 at 13 44 42

edmundoa: This is of course more a suggestion than a change, so you can ignore it if you feel like it :)

kmerz: I had it before in ```#aaa``` but since you meant not to prominent I cooled it down a bit ;)
Will change it to ```#aaa``` :+1:

edmundoa:
@dennisoelkers any other comments on this PR?

PR #4526, opened 2018-01-25
Prevent ldap server password leak

### LINKED ISSUE #3763: Password leak through LDAP configuration page
On the LDAP/AD configuration page, the LDAP/AD password is actually returned back to the frontend in plain text. This is a minor security leak, since you need administrator privileges to access this page, but can be easily mitigated.

## Expected Behavior
I cannot view the LDAP/AD password in the frontend/web interface of the application.

## Current Behavior
I can view the LDAP/AD password in the frontend/web interface of the application by using the browser's inspector.

## Possible Solution
The LDAP/AD password is not send back or rendered to the frontend, but a normally invalid placeholder is sent, such as "********". When a user saves the configuration, the application can check whether or not the placeholder is present. If the placeholder is present in the form, the LDAP/AD password will not be changed, otherwise it will.

1. Open the System > Authentication > LDAP Settings page
2. Fill in the LDAP authentication parameters
3. Save the page
4. Reload the page
5. Open your browser's inspector on the LDAP password field
6. View the password in plain text

## Context
Trying to set up an easy to use and secure logging environment. 

## Your Environment
* Graylog Version: 2.2.1+4829190 
* Elasticsearch Version: 2 (Docker, latest)
* MongoDB Version: 3 (Docker, latest)
* Operating System: Windows 10
* Browser version: Chrome (stable)

kmerz:
## Description
The ldap configuration page was setting the input value of the
password field to the actuall password.

This change prevents reading the password via "inspect" in browser,
by hiding the password input field and instead show a string "Password is set".

If the user wants to set a new password he now needs to click the "Reset Password" button.

There is now no way anymore to clear the password anymore, since we drop empty
passwords now. That is needed since the GUI does not receive and therefore cannot send
the password any longer.

graylog - ldap settings

## Motivation and Context
 A attacker could read the ldap password, if he gained admin access to graylog.

## How Has This Been Tested?
Password set and updated and checked via inspect.

Fixes #3763 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/system/ldap/requests/LdapSettingsRequest.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/system/ldap/requests/LdapSettingsRequest.java
@@ -39,6 +39,7 @@ public abstract class LdapSettingsRequest {
     public abstract String systemUsername();
 
     @JsonProperty
+    @Nullable
     public abstract String systemPassword();
 
     @JsonProperty
@@ -91,7 +92,7 @@ public abstract class LdapSettingsRequest {
     @JsonCreator
     public static LdapSettingsRequest create(@JsonProperty("enabled") boolean enabled,
                                              @JsonProperty("system_username") @NotEmpty String systemUsername,
-                                             @JsonProperty("system_password") @NotEmpty String systemPassword,
+                                             @JsonProperty("system_password") @Nullable String systemPassword,
                                              @JsonProperty("system_password_set") boolean isSystemPasswordSet,
                                              @JsonProperty("ldap_uri") URI ldapUri,
                                              @JsonProperty("use_start_tls") boolean useStartTls,
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/system/ldap/requests/LdapSettingsRequest.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/system/ldap/requests/LdapSettingsRequest.java
@@ -42,6 +42,9 @@ public abstract class LdapSettingsRequest {
     public abstract String systemPassword();
 
     @JsonProperty
+    public abstract boolean isSystemPasswordSet();
+
+    @JsonProperty
     public abstract URI ldapUri();
 
     @JsonProperty
@@ -89,6 +92,7 @@ public abstract class LdapSettingsRequest {
     public static LdapSettingsRequest create(@JsonProperty("enabled") boolean enabled,
                                              @JsonProperty("system_username") @NotEmpty String systemUsername,
                                              @JsonProperty("system_password") @NotEmpty String systemPassword,
+                                             @JsonProperty("system_password_set") boolean isSystemPasswordSet,
                                              @JsonProperty("ldap_uri") URI ldapUri,
                                              @JsonProperty("use_start_tls") boolean useStartTls,
                                              @JsonProperty("trust_all_certificates") boolean trustAllCertificates,
@@ -105,6 +109,7 @@ public abstract class LdapSettingsRequest {
         return new AutoValue_LdapSettingsRequest(enabled,
                                                  systemUsername,
                                                  systemPassword,
+                                                 isSystemPasswordSet,
                                                  ldapUri,
                                                  useStartTls,
                                                  trustAllCertificates,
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog2/security/ldap/LdapSettingsImpl.java
```diff
@@ -84,6 +84,8 @@
     public static final String LDAP_GROUP_MAPPING_NAMEKEY = "group";
     public static final String LDAP_GROUP_MAPPING_ROLEKEY = "role_id";
 
+    public static final String SYSTEM_PASSWORD_PLACEHOLDER = "*************";
```
dennisoelkers: I am a bit opposed against the idea to return a magic string to mask the password. Maybe it is a better idea to use a boolean flag (`system_password_set`) to indicate that there is an ldap system password configured, which is returned in (and only in) the DTO used by `LdapResource#getLdapSettings`? The DTOs used for setting/getting LDAP settings are already separate, so the effort should not be too big.

graylog2-server/src/main/java/org/graylog2/security/ldap/LdapSettingsImpl.java
```diff
@@ -140,8 +142,18 @@ public String getSystemPassword() {
...
+    @Override
+    public String getSystemPasswordPlaceholder() {
+        final Object o = fields.get(SYSTEM_PASSWORD);
+        if (o == null) return "";
+        return SYSTEM_PASSWORD_PLACEHOLDER;
+    }
+
     @Override
     public void setSystemPassword(String systemPassword) {
+        if (systemPassword == SYSTEM_PASSWORD_PLACEHOLDER) {
```
dennisoelkers: The correct way to check for equality in java is using `Object#equals`. This way (using `==`) we check for reference equality, not value equality, which works only for primitive types or by accident (when equal values are managed as singletons).

graylog2-server/src/main/java/org/graylog2/rest/models/system/ldap/requests/LdapSettingsRequest.java
```diff
@@ -89,6 +92,7 @@
     public static LdapSettingsRequest create(@JsonProperty("enabled") boolean enabled,
                                              @JsonProperty("system_username") @NotEmpty String systemUsername,
                                              @JsonProperty("system_password") @NotEmpty String systemPassword,
```
kroepke: I'm getting
```
There was an error fetching a resource: cannot PUT http://localhost:9000/api/system/ldap/settings (400). Additional information: Can not construct instance of org.graylog2.rest.models.system.ldap.requests.LdapSettingsRequest, problem: Null systemPassword
```
when trying to save an updated LDAP settings page with leaving the system password empty (i.e. to the "set" placeholder).

This change needs to allow for the `systemPassword` to be empty/null and check that the password has already beend set.

bernd:
@kmerz Okay, I found the issues:

1. My LDAP test server accepted anonymous connections so I got a successful connection even though my password was wrong :wink: 
2. The LDAP test resource expects the password to be present all the time

A fix for the second issue could be the following. If the submitted config doesn't include a password, check if we have one in the database and use that.

```diff
diff --git graylog2-server/src/main/java/org/graylog2/rest/resources/system/ldap/LdapResource.java graylog2-server/src/main/java/org/graylog2/rest/resources/system/ldap/LdapResource.java
index 875333990..3230298aa 100644
--- graylog2-server/src/main/java/org/graylog2/rest/resources/system/ldap/LdapResource.java
+++ graylog2-server/src/main/java/org/graylog2/rest/resources/system/ldap/LdapResource.java
@@ -137,9 +137,21 @@ public class LdapResource extends RestResource {
             config.setTrustManagers(new TrustAllX509TrustManager());
         }
 
-        if (!isNullOrEmpty(request.systemUsername()) && !isNullOrEmpty(request.systemPassword())) {
+        if (!isNullOrEmpty(request.systemUsername())) {
             config.setName(request.systemUsername());
-            config.setCredentials(request.systemPassword());
+
+            if (!isNullOrEmpty(request.systemPassword())) {
+                // Use the given password for the connection test
+                config.setCredentials(request.systemPassword());
+            } else {
+                // If the config request has a username but no password set, we have to use the password from the database.
+                // This is because we don't expose the plain-text password through the API anymore and the settings form
+                // doesn't submit a password.
+                final LdapSettings ldapSettings = ldapSettingsService.load();
+                if (ldapSettings != null && !isNullOrEmpty(ldapSettings.getSystemPassword())) {
+                    config.setCredentials(ldapSettings.getSystemPassword());
+                }
+            }
         }
 
         LdapNetworkConnection connection = null;
```

Issue #4527, opened 2018-01-25
Improve handling of concurrent actions in persistence layer


edmundoa:
In order to avoid data inconsistencies, we need to improve our handling of concurrent actions in the persistence layer.

Currently there is no easy way of preventing concurrent actions to modify the same MongoDB document with unexpected consequences, as you can see in #4525. This is specially problematic when dealing with embedded documents, as the same document may be updated more often and only partially. In the best case this situation may lead to an action being reverted (as it is the case in the example), but this could potentially lead to data inconsistencies, specially when other parts of the system react to those data changes.

This is a complex topic and we need to analyse it and look for a solution that help us avoiding this kind of problem in the whole persistence layer, so I want to open this issue here and we can update it once we have some ideas and know how to solve the issue.

PR #4530, opened 2018-01-25
Only update dashboard positions when necessary

### LINKED ISSUE #4525: Concurrent requests when removing widget may persist it again
Deleting a widget from the web interface may not end up deleting the widget as expected. The reason for this is that two requests are made in very short time when deleting a widget:

1. User clicks on "Delete" button on a widget
2. React Grid Layout calls the `onDragStop` callback, which triggers an update to the server to update the dashboard positions
3. The JS code eventually stops, since we ask for user confirmation before sending the delete request to the server
4. Both the update positions and delete requests are sent to the server (if the user confirmed the deletion)
5. Depending on which of the requests is executed first, this situation potentially leads to a widget being removed from the dashboard document just to be stored again in the other request
6. Subsequent requests to the dashboard resource will include the "removed" widget, since it is still in MongoDB, and that's why the widget does not disappear from the dashboard

Here is a gif of how it looks like:

widget

And here is the only log available in the server (there are no logs in browser JS console):
```
2018-01-25 14:19:05,794 INFO : org.graylog2.rest.resources.dashboards.DashboardWidgetsResource - Deleted widget <7778a1fa-983f-42d5-a8a7-b2196ccf8f06> from dashboard <5a68b21258ddc87f325a5b7a>. Reason: REST request.
```

## Expected Behavior
Deleting a widget from the web interface works as expected.

## Current Behavior
Deleting a widget from the web interface may not delete the widget after all.

1. Go to a dashboard and enter "edit" mode
2. Delete a widget and confirm the action
3. If you are (un)lucky the widget comes back again (see gif above)

## Your Environment
* Graylog Version: Graylog 3.0.0-SNAPSHOT(c83918210356422d4e1f4738365f04254b5fbd64) and Graylog 2.4.3

edmundoa:
This PR makes some modifications in the web interface to be smarter calculating and updating dashboard positions:

1. Avoid callbacks when clicking on an action button in a widget. This is what made #4525 too likely to happen
2. Only compute layout when positions change, instead of doing it on every render
3. Ensure there are layout changes before calling the `onLayoutChange` callback, avoiding doing unnecessary updates in the server

In order to fix the issue in 2.4, I think we should only backport change number 1, avoiding to introduce any other potential issues in the stable release. I can take care of that once this PR is reviewed.

Before a solution for #4527 lands, these changes fix #4525.

### DIFF
--- a/graylog2-web-interface/src/components/common/ReactGridContainer.jsx
+++ b/graylog2-web-interface/src/components/common/ReactGridContainer.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Responsive, WidthProvider } from 'react-grid-layout';
+import lodash from 'lodash';
 
 import 'react-grid-layout/css/styles.css';
 import 'react-resizable/css/styles.css';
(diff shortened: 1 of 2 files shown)


edmundoa:
Oh, I forgot: I tried to write some tests to check how well the changes worked but it end up being too complicated. To test the functionality we need that react-grid-layout works but with enzyme it can only go to a point. I gave up after I needed to stub a couple of `HTMLElement` properties that do not seem to be implemented in the DOM implementation used for the tests.

graylog2-web-interface/src/components/common/ReactGridContainer.jsx
```diff
@@ -107,7 +115,39 @@ const ReactGridContainer = React.createClass({
...
+        w: width || 1,
+      };
+    });
+  },
+
   _onLayoutChange(newLayout) {
+    // `onLayoutChange` may be triggered when clicking somewhere in a widget, check before propagating the change.
+    // Filter out additional Object properties in nextLayout, as it comes directly from react-grid-layout
+    const filteredNewLayout = newLayout.map(item => ({ i: item.i, x: item.x, y: item.y, h: item.h, w: item.w }));
+    if (lodash.isEqual(this.state.layout, filteredNewLayout)) {
```
dennisoelkers: Would it make sense to move this kind of check into `shouldComponentUpdate` and keep the component stateless?

edmundoa: I don't think so, as this callback is used by `react-grid-layout` to notify when a widget is dragged or resized and it's not modifying the state or deciding if the component should re-render. The check it's avoiding to call a prop when the layout didn't change.

edmundoa:
@dennisoelkers do you have any other comments on this? :)

edmundoa:
@dennisoelkers did you find any time to look at this? I'm afraid more changes will come soon in these components and we will have to solve conflicts in them.

PR #4537, opened 2018-01-29
Ignore `LookupCacheKey#isPrefixOnly()` when (de-) serializing with Jackson

joschi:
The derived `isPrefixOnly()` method should be ignored when the `LookupCacheKey` instance is being serialized.

Otherwise deserialization will fail due to the unknown "prefix_only" field.


joschi:
This PR should be back-ported into the `2.4` branch.

PR #4538, opened 2018-01-29
Simplify `ByteBufferUtils#readBytes()`, document side-effects, add tests

joschi:
The `ByteBufferUtils#readBytes(ByteBuf, int, int)` method failed for non-array-backed byte buffers whose reader index didn't start at index 0 and failed to mention side effects (e. g. consumption of the given `ByteBuffer`) in the Javadocs.

Additionally, the "optimization" in `ByteBufferUtils#readBytes(ByteBuf, int, int)` was unnecessary because the relevant `ByteBuffer` implementations are already using the backing array if available.


joschi:
This PR should probably also be back-ported into the `2.4` branch.

PR #4541, opened 2018-01-29
Fix entry count metric for lookup caches

### LINKED ISSUE #4540: Entry count for Guava lookup cache is broken
## Expected Behavior

The statistics (entry count) of a Guava lookup cache should be updated if the number of entries changes.

## Current Behavior

The statistics (entry count) of a Guava lookup cache are always 0.

## Context

https://github.com/Graylog2/graylog2-server/blob/2c418975989da2fff589e35cdf8c00ff41983e52/graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupCache.java#L63

https://github.com/Graylog2/graylog2-server/blob/2c418975989da2fff589e35cdf8c00ff41983e52/graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java#L85-L91

Either `LookupCache#entryCount()` should return the number of entries in the cache (`long` instead of `Gauge<Long>`) or `GuavaLookupCache#entryCount()` has to return a dynamically changing `Gauge<Long>` instead of one always returning `0L` if the cache hasn't been started yet.

joschi:
The way the `LookupCache#entryCount()` method signature was defined, caused implementers to write incorrect implementations.

For example `GuavaLookupCache#entryCount()` returned a constant `Gauge<Long>` on the first call (in the constructor of `LookupCache`) which was never updated.

The signature of `LookupCache#entryCount()` has been changed to prevent such incorrect usages and implementations by simply returning a `long` value and adding a dynamic `Gauge<Long>` in the constructor of `LookupCache`.

Additionally, the `CacheTableEntry` tried to access the entry count metric through an invalid object path which resulted in the metric always being displayed as "NaN" in the web interface.

Fixes #4540

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/caches/GuavaLookupCache.java
@@ -16,7 +16,6 @@
  */
 package org.graylog2.lookup.caches;
 
-import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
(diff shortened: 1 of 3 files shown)


joschi:
This PR should probably also be back-ported into the `2.4` branch.

PR #4542, opened 2018-01-29
Add JSON factory method for `LookupResult`

### LINKED ISSUE #4534: LookupResult has no constructor or factory method usable by Jackson
While every property of [`LookupResult`](https://github.com/Graylog2/graylog2-server/blob/2.4.3/graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java) is annotated with a `@JsonProperty` annotation so that instances can be serialized with Jackson, it lacks a suitable constructor or factory method (with `@JsonCreator`) to **deserialize** and create an instance.

This makes it unnecessary hard to use `LookupResult` with anything that needs to serialize and deserialize the actual object graph, e. g. any external cache (instead of Guava Cache).

joschi:
This commit adds a factory method to `LookupResult` which can be used by Jackson to deserialize an instance of `LookupResult` from JSON.

Additionally, the derived field "empty" (via `LookupResult#isEmpty()`) is being ignored and thus not part of the serialized or deserialized representation of `LookupResult`.

Fixes #4534

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupCacheKey.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupCacheKey.java
@@ -39,6 +39,12 @@ import javax.annotation.Nullable;
  * // Key with prefix only
  * LookupCacheKey.prefix(dataAdapter.id());
  * }</pre>
+ * <p>
+ * For convenience, this class can be serialized and deserialized with Jackson (see
+ * {@link com.fasterxml.jackson.databind.ObjectMapper}, but we strongly recommend implementing your own
+ * serialization and deserialization logic if you're implementing a lookup cache.
+ * <p>
+ * There are <em>no guarantees</em> about binary compatibility of this class across Graylog releases!
  */
 @AutoValue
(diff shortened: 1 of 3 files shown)


joschi:
This PR should also be back-ported into the `2.4` branch.

bernd:
@joschi I was wondering if we should advise cache authors to create separate DTO objects for serialization instead of serializing the `LookupResult` so they have better control over the data that is stored in external databases. (same goes for the `LookupCacheKey` in https://github.com/Graylog2/graylog2-server/pull/4537)

Using DTOs can also avoid data migrations in the cache store when the `LookupResult` and `LookupCacheKey` objects change.

What do you think?

joschi:
@bernd Fair enough, but given that the only official cache implementation (the Guava-based lookup cache) is using exactly these DTOs, that fact should very clearly be stated in the (Javadoc) documentation of the related classes.

bernd:
@joschi Agreed! Can you add the comments? :smiley:

joschi:
@bernd I've added a Javadoc comment for `LookupResult` and `LookupCacheKey`.

PR #4547, opened 2018-01-31
Allow using colon character (":") in password

joschi:
`ShiroSecurityContextFilter` incorrectly parsed the HTTP Basic Auth header and prevented users from using the colon character (":") in their passwords.

Refs https://community.graylog.org/t/graylog-behind-nginx-sub-directory-how-to-configure/3968

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContext.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContext.java
@@ -19,11 +19,13 @@ package org.graylog2.shared.security;
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.shiro.authc.AuthenticationException;
 import org.apache.shiro.authc.AuthenticationToken;
+import org.apache.shiro.authc.UsernamePasswordToken;
 import org.apache.shiro.subject.Subject;
 import org.apache.shiro.util.ThreadContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


joschi:
This PR should be back-ported into the `2.4` branch.

davama:
Thank you for submitting this!
Hope it gets merged soon :+1:

graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContext.java
```diff
@@ -49,18 +51,22 @@ public ShiroSecurityContext(Subject subject, AuthenticationToken token, boolean
...
         if (token == null || token.getPrincipal() == null) {
             return null;
         }
         return token.getPrincipal().toString();
     }
 
+    @Nullable
     public String getPassword() {
-        if (token == null || token.getCredentials() == null) {
+        if (token instanceof UsernamePasswordToken) {
```
bernd: Hmm, this breaks once we implement another `AuthenticationToken` which returns non-null in `#getCredentials()`.

What's the reason for the type check here?

joschi: @bernd For now, `UsernamePasswordToken` is the only type which has a sensible value for `#getCredentials()`. We can add handling for other tokens if the need for that comes up.

The previous implementation of the `ShiroSecurityContext#getPassword()` method was simply wrong (and returned `char[].toString()` in case of a `UsernamePasswordToken`).

PR #4550, opened 2018-01-31
Make API Tokens configurable in web interface

### LINKED ISSUE #3318: Show/Create Access Token in Webinterface
Hi

I'm a starter in graylog, so close my case, if you don't experiment this or I missunderstand something. It is also not so important to me, I just want let you know about my experiment.

I created token with curl, no problem.
So maybe I mixed to different function, ot this part of the web interface has big bugs.
I can't create token for non admin user (eg. monitoring user). Maybe good if you can create tokens for other users.

## Expected Behavior

With disable Access Token Authenticator I can't use tokens.
The website shows my current tokens.

## Current Behavior

I don't see the token at the web interface, under System/Authentication/Access Token Authenticator.
If I disable this module, I also can use the api token with success.
And the  "last_access" doesn't updated.
curl  -u  XX:XX  -X GET 'http://IP:9000/api/users/XX/tokens/?pretty=true'
{
  "tokens" : [ {
    "name" : "XX",
    "token" : "XX",
    "last_access" : "1970-01-01T00:00:00.000Z"
  } ]
}
kep

## Possible Solution

I don't see the token at the web interface, under System/Authentication/Access Token Authenticator.
If I disable this module, I also can use the api token with success.

1. Create token
2. Check on the web page

## Your Environment

* Graylog Version: 2.1.2
* Elasticsearch Version:
* MongoDB Version:
* Operating System: CentOS Linux release 7.3.1611 (Core)
* Browser version: Firefox 50.0.1

kmerz:
## Description
API Tokens are configurable via REST API but there is till now no way to
create, delete or list them in the web interface.

- I fixed the permissions handling for API Tokens which till now only worked, if the user requested his own tokens regardless of his role (like admin). That meant that even the admin could not alter the API Tokens. Now a normal user can only manage his own Tokens and a user with role *admin* can manage the tokens of other people additionally.
- Add a new TokenList Component which list, create and deletes tokens. I did not wire the UserStore into that component but rather gave handlers as arguments and implemented them by the caller. This is easier to test and easier to use IMHO.

## Motivation and Context
To use tokens a user needed to use the REST API manually to create delete or list
them.

## How Has This Been Tested?
I tested the REST API to ensure that the permissions worked as I described above.
I tested the TokenList manually.

graylog - edit tokens of user admin

Fixes #3318 
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -86,7 +86,12 @@ import java.util.Set;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.maxBy;
-import static org.graylog2.shared.security.RestPermissions.*;
+import static org.graylog2.shared.security.RestPermissions.USERS_EDIT;
+import static org.graylog2.shared.security.RestPermissions.USERS_PERMISSIONSEDIT;
+import static org.graylog2.shared.security.RestPermissions.USERS_ROLESEDIT;
+import static org.graylog2.shared.security.RestPermissions.USERS_TOKENCREATE;
+import static org.graylog2.shared.security.RestPermissions.USERS_TOKENREMOVE;
+import static org.graylog2.shared.security.RestPermissions.USERS_TOKENLIST;
 
 @RequiresAuthentication
 @Path("/users")
@@ -231,10 +236,7 @@ public class UsersResource extends RestResource {
                            @Valid @NotNull ChangeUserRequest cr) throws ValidationException {
         checkPermission(USERS_EDIT, username);
 
-        final User user = userService.load(username);
-        if (user == null) {
-            throw new NotFoundException("Couldn't find user " + username);
-        }
+        final User user = loadUser(username);
 
         if (user.isReadOnly()) {
             throw new BadRequestException("Cannot modify readonly user " + username);
@@ -424,11 +426,11 @@ public class UsersResource extends RestResource {
     @ApiOperation("Retrieves the list of access tokens for a user")
     public TokenList listTokens(@ApiParam(name = "username", required = true)
                                 @PathParam("username") String username) {
-        final User user = _tokensLoadUser(username);
-
         if (!isPermitted(USERS_TOKENLIST, username)) {
             throw new ForbiddenException("Not allowed to list tokens for user " + username);
         }
+
+        final User user = loadUser(username);
 
         final ImmutableList.Builder<Token> tokenList = ImmutableList.builder();
         for (AccessToken token : accessTokenService.loadAll(user.getName())) {
@@ -446,11 +448,12 @@ public class UsersResource extends RestResource {
             @ApiParam(name = "username", required = true) @PathParam("username") String username,
             @ApiParam(name = "name", value = "Descriptive name for this token (e.g. 'cronjob') ", required = true) @PathParam("name") String name,
             @ApiParam(name = "JSON Body", value = "Placeholder because POST requests should have a body. Set to '{}', the content will be ignored.", defaultValue = "{}") String body) {
-        final User user = _tokensLoadUser(username);
-
         if (!isPermitted(USERS_TOKENCREATE, username)) {
             throw new ForbiddenException("Not allowed to create tokens for user " + username);
         }
+
+        final User user = loadUser(username);
+
 
         final AccessToken accessToken = accessTokenService.create(user.getName(), name);
 
@@ -464,8 +467,6 @@ public class UsersResource extends RestResource {
     public void revokeToken(
             @ApiParam(name = "username", required = true) @PathParam("username") String username,
             @ApiParam(name = "token", required = true) @PathParam("token") String token) {
-        final User user = _tokensLoadUser(username);
-
         if (!isPermitted(USERS_TOKENREMOVE, username)) {
             throw new ForbiddenException("Not allowed to remove tokens for user " + username);
         }
@@ -479,7 +480,7 @@ public class UsersResource extends RestResource {
         }
     }
 
-    private User _tokensLoadUser(String username) {
+    private User loadUser(String username) {
         final User user = userService.load(username);
         if (user == null) {
             throw new NotFoundException("Unknown user " + username);
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -423,11 +423,14 @@ public class UsersResource extends RestResource {
 
     @GET
     @Path("{username}/tokens")
-    @RequiresPermissions(RestPermissions.USERS_TOKENLIST)
     @ApiOperation("Retrieves the list of access tokens for a user")
     public TokenList listTokens(@ApiParam(name = "username", required = true)
                                 @PathParam("username") String username) {
-        final User user = _tokensCheckAndLoadUser(username);
+        final User user = _tokensLoadUser(username);
+
+        if (!getSubject().isPermitted(RestPermissions.USERS_TOKENLIST + ":" + user.getName())) {
+            throw new ForbiddenException("Not allowed to list tokens for user " + username);
+        }
 
         final ImmutableList.Builder<Token> tokenList = ImmutableList.builder();
         for (AccessToken token : accessTokenService.loadAll(user.getName())) {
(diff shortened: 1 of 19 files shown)


dennisoelkers:
Great work! Maybe it makes more sense to use the [`KeyValueTable](https://graylog2.github.io/frontend-documentation/#keyvaluetable) for listing/modifying the tokens?

graylog2-web-interface/src/components/authentication/AuthenticationComponent.jsx
```diff
@@ -115,9 +115,14 @@ const AuthenticationComponent = React.createClass({
...
-      authenticators = [<LinkContainer key="profile-edit" to={Routes.SYSTEM.AUTHENTICATION.USERS.edit(encodeURIComponent(this.state.currentUser.username))}>
-        <NavItem title="Edit User">Edit User</NavItem>
-      </LinkContainer>];
+      authenticators = [
+        <LinkContainer key="profile-edit" to={Routes.SYSTEM.AUTHENTICATION.USERS.edit(encodeURIComponent(this.state.currentUser.username))}>
+          <NavItem title="Edit Profile">Edit Profile</NavItem>
+        </LinkContainer>,
+        <LinkContainer key="profile-edit-tokens" to={Routes.SYSTEM.AUTHENTICATION.USERS.edit_tokens(encodeURIComponent(this.state.currentUser.username))}>
+          <NavItem title="Edit Tokens">Edit Tokens</NavItem>
+        </LinkContainer>
```
edmundoa: Another linter nitpick: Missing trailing comma.

graylog2-web-interface/src/pages/EditTokensPage.jsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <DocumentTitle title={`Edit tokens of user ${this.props.params.username}`}>
+        <span>
+          <PageHeader title={<span>Edit tokens of user <em>{this.props.params.username}</em></span>} subpage>
+            <span>You can create new tokens or delete old ones.</span>
+            {null}
+          </PageHeader>
+          <TokenList tokens={this.state.tokens}
+                     delete={this._deleteToken}
+                     create={this._createToken} />
```
edmundoa: We usually name props expecting function handlers with "on" + verb (e.g. `onDelete` or `onCreate`) to differentiate them from other props at first glance.

graylog2-web-interface/src/components/users/TokenList.jsx
```diff
@@ -0,0 +1,137 @@
...
+    this.setState({ token_name: '' });
+  }
+
+  itemActionsFactory(token) {
+    const copyButtonProps = {};
+    copyButtonProps['data-clipboard-text'] = token.token;
+
+    return (
+      <span>
+        <Button data-clipboard-button {...copyButtonProps} bsSize="xsmall">
```
edmundoa: Could we use the [`ClipboardButton`](https://graylog2.github.io/frontend-documentation/#clipboardbutton) component here?

graylog2-web-interface/src/components/users/TokenList.jsx
```diff
@@ -0,0 +1,137 @@
...
+                id="create-token-input"
+                type="text"
+                placeholder="e.g ServiceName"
+                value={this.state.token_name}
+                onChange={this._onNewTokeChanged} />
+            </Col>
+            <Col sm={2}>
+              <Button id="create-token"
+                      disabled={this.state.token_name === ''}
+                      onClick={this._createToken}
```
edmundoa: We should move this function to the `onSubmit` event handler of the `form` element, and stop the event with `preventDefault()` in the handler. Otherwise clicking the button does a full page reload.

graylog2-web-interface/src/stores/users/UsersStore.ts
```diff
@@ -95,6 +101,43 @@ export const UsersStore = {
...
+    return promise;
+  },
+
+  deleteToken(username: string, token_name: string): Promise<string[]> {
+    const  url = URLUtils.qualifyUrl(ApiRoutes.UsersApiController.delete_token(encodeURIComponent(username),
+      encodeURIComponent(token_name)).url, {});
+    const  promise = fetch('DELETE', url);
+
+    promise.then(() => {
+      UserNotification.success("Token \"" + token_name + "\" of user \"" + username + "\" was deleted successfully");
```
edmundoa: Could we display in this notification the token name instead of the token itself?

graylog2-web-interface/src/pages/EditTokensPage.jsx
```diff
@@ -0,0 +1,79 @@
...
+      [`users:tokenlist:${username}`]);
+  },
+
+  _deleteToken(token) {
+    const promise = UsersStore.deleteToken(this.props.params.username, token);
+    promise.then(() => this._loadTokens(this.props.params.username));
+  },
+
+  _createToken(tokenName) {
+    const promise = UsersStore.createToken(this.props.params.username, tokenName);
```
edmundoa: I think we should add a loading indicator when creating and deleting tokens, in case the API takes a while to return in here. Probably changing the button name and disabling it until the list is updated should suffice.

What do you think?

graylog2-web-interface/src/components/users/TokenList.jsx
```diff
@@ -0,0 +1,137 @@
...
+    this.props.create(this.state.token_name);
+    this.setState({ token_name: '' });
+  }
+
+  itemActionsFactory(token) {
+    const copyButtonProps = {};
+    copyButtonProps['data-clipboard-text'] = token.token;
+
+    return (
+      <span>
```
edmundoa: I would use a `ButtonToolbar` in here to separate both buttons. If you think putting them together would make more sense, then I would use a `ButtonGroup`, which styles them a bit better for that.

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -86,9 +86,7 @@
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.maxBy;
-import static org.graylog2.shared.security.RestPermissions.USERS_EDIT;
-import static org.graylog2.shared.security.RestPermissions.USERS_PERMISSIONSEDIT;
-import static org.graylog2.shared.security.RestPermissions.USERS_ROLESEDIT;
+import static org.graylog2.shared.security.RestPermissions.*;
```
dennisoelkers: Please avoid wildcard imports. You can disable automatic insertion of wildcard imports in your Intellij config (`Code Style`/`Java`).

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -423,11 +421,14 @@ public void changePassword(
 
     @GET
     @Path("{username}/tokens")
-    @RequiresPermissions(RestPermissions.USERS_TOKENLIST)
     @ApiOperation("Retrieves the list of access tokens for a user")
     public TokenList listTokens(@ApiParam(name = "username", required = true)
                                 @PathParam("username") String username) {
-        final User user = _tokensCheckAndLoadUser(username);
+        final User user = _tokensLoadUser(username);
```
dennisoelkers: I think the naming of that method is suboptimal, for two reasons:

  * Method names are not prefixed with `_` in Java. The signature already reveals that it is a private method.
  * The name `_tokensLoadUser` implies that it is specific to the Token management, which it is not. If it would be called `loadUser`, it would describe accurately what it does and it could be reused by the `changeUser` method.

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -423,11 +421,14 @@ public void changePassword(
...
+
+        if (!isPermitted(USERS_TOKENLIST, username)) {
```
dennisoelkers: I think we should do the check before fetching the user, so the call cannot be used to find out existing usernames.

graylog2-web-interface/src/routing/Routes.jsx
```diff
@@ -93,6 +93,8 @@ const Routes = {
       USERS: {
         CREATE: '/system/authentication/users/new',
         edit: username => `/system/authentication/users/edit/${username}`,
+        edit_tokens: username => `/system/authentication/users/tokens/${username}`,
```
dennisoelkers: I think it makes sense to group them as a `TOKENS` subkey, as tokens are a subresource of a user.

graylog2-web-interface/src/routing/Routes.jsx
```diff
@@ -93,6 +93,8 @@ const Routes = {
...
+        TOKENS: '/system/authentication/users/tokens/',
```
dennisoelkers: This would become `TOKENS.LIST` then.

dennisoelkers:
Would it make sense to allow the user to temporarily disable tokens? That means, a token is not deleted but still not usable for authenticating.

edmundoa:
@dennisoelkers I think it would be a nice to have, although I would not include it in the scope of this PR. I think the goal here is to make authentication tokens visible (and usable) to users, and that is already achieved with the functionality in here. We could of course add that functionality in the future.

edmundoa:
Looks good to me too! 👍

laszlo-laszlo:
Great work!

PR #4555, opened 2018-02-01
Add upgrading note for proxy setups using X-Graylog-Server-URL

bernd:
Users who use a proxy setup and are upgrading to 3.0 need to adjust the `X-Graylog-Server-URL` HTTP header config.


bernd:
Oups, the formatting is wrong. Fixing it.

PR #4573, opened 2018-02-13
Add body writer for serializing errors (GenericError et al) as CSV

### LINKED ISSUE #4568: CSV Export: Error messages are swallowed
Refs https://community.graylog.org/t/graylog-2-3-1-export-to-csv-via-api-call/4102

## Expected Behavior

The HTTP response of `GET /search/universal/relative/export` for exporting CSV data sets should contain an explanation for an occurred error instead of swallowing the error message.

## Current Behavior

The HTTP response is empty and the error message is swallowed.
The logs of the Graylog node contain the following error messages:

```text
ERROR: org.glassfish.jersey.message.internal.WriterInterceptorExecutor - MessageBodyWriter not found for media type=text/csv, type=class org.graylog2.rest.resources.search.responses.AutoValue_SearchError, genericType=class org.graylog2.rest.resources.search.responses.AutoValue_SearchError.
ERROR: org.glassfish.jersey.server.ServerRuntime$Responder - Error occurred when processing a response created from an already mapped exception.
```

## Possible Solution

Create a mapper for `SearchError` (or `ApiError`) for the `text/csv` content type.

## Your Environment

* Graylog Version: 2.4.3

joschi:
Fixes #4568

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/GenericErrorCsvWriter.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/GenericErrorCsvWriter.java
@@ -1,0 +1,81 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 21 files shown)


graylog2-server/src/main/java/org/graylog2/rest/GenericErrorCsvWriter.java
```diff
@@ -0,0 +1,81 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Provider
+@Produces(MoreMediaTypes.TEXT_CSV)
+public class GenericErrorCsvWriter implements MessageBodyWriter<GenericError> {
+    private final CsvMapper mapper;
+
+    public GenericErrorCsvWriter() {
+        this.mapper = new CsvMapper();
```
joschi: We could probably move that out of this class, make it a singleton, and also use it in `ScrollChunkWriter`.

PR #4576, opened 2018-02-13
Reduce development startup time by improving globbing.

dennisoelkers:
## Motivation and Context

With a higher number (>5) of plugins in development, running `yarn
start` takes significant time during the globbing step which finds all webpack configs of the installed plugins.

By using a different globbing module (fast-glob) which supports a number of options not present for glob, the startup time can be significantly reduced.

I wrote a small script[1] to compare the times for both scenarios, its results are (on my machine):

```
fast-glob: 0.427s
glob: 48.122s
```

[1]: https://gist.github.com/dennisoelkers/aca165c209f0e6d5b54b2513fa25e243

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/webpack.combined.config.js
+++ b/graylog2-web-interface/webpack.combined.config.js
@@ -8,10 +8,15 @@ const MANIFESTS_PATH = path.resolve(ROOT_PATH, 'manifests');
 const VENDOR_MANIFEST_PATH = path.resolve(MANIFESTS_PATH, 'vendor-manifest.json');
 const TARGET = process.env.npm_lifecycle_event;
 
-const pluginPrefix = '../../graylog-plugin-*/**/';
-const pluginConfigPattern = `${pluginPrefix}webpack.config.js`;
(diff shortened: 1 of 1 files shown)


edmundoa:
Looking at the script you linked in the description and the `glob` documentation, it looks like the `ignore` pattern for `glob` is not right and it should add a wildcard at the end to match all files in the directory:

```diff
-  ignore: '**/node_modules',
+  ignore: '**/node_modules/**',
```

Could you please take a look and see if that helps already with the issue? If results are similar I think we would be better off sticking to `glob`, since it's much more used than `fast-glob`.

dennisoelkers:
I tried it both ways, no difference.

dennisoelkers:
Interesting. I did not find any description of how the ignore patterns work for `glob`, but it seems like they are not matched on a directory before descending. Adding a wildcard at the end almost brings it en par with `fast-glob` (although still 2x as slow). I second sticking to `glob` though and will adapt the PR.

dennisoelkers:
Closed in favor of #4582.

PR #4579, opened 2018-02-14
TableList improvements

edmundoa:
This PR aims to fix some shortcomings on the `TableList` component. We need to do those in order to reuse the component in the new collector UI and to fix some old usability issues that have been there for a while now. These are the main changes:

- Introduce `ControlledTableList` component, that allows more flexibility to consumers that only want the basis of a `TableList` but nothing else. This component is now used by `TableList` itself
- Improve feedback when selecting one or more elements in the list to perform bulk actions
- Update selection when filter changes
- Update filtered items and selection when `items` prop changes
- Make bulk actions optional, by adding an `enableBulkActions` prop
- Simplify how to disable filtering, by adding an `enableFilter` prop

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -121,6 +121,7 @@
     "less": "^2.5.3",
     "less-loader": "^4.0.5",
     "phantomjs-prebuilt": ">=1.9",
+    "react-immutable-proptypes": "^2.1.0",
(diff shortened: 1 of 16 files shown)


kmerz:
:+1:

edmundoa:
I have rebased to master and adapted the `TokenList` component to changes in this PR.

PR #4581, opened 2018-02-14
Collect and expose index field type information

bernd:
This PR introduces a system to collect and expose field type information for Elasticsearch indices.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
@@ -122,7 +122,7 @@ public class IndexFieldTypePollerPeriodical extends Periodical {
 
             // Cleanup orphaned field type entries that haven't been removed by the event handler
             dbService.streamForIndexSet(indexSetId)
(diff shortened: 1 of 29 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+
+        // Make sure there is no existing polling job running for this index set
+        cancel(futures.get(indexSetId));
+
+        LOG.debug("Schedule index field type updating for index set <{}/{}> every {} ms", indexSetId, indexSetTitle,
+                refreshInterval.getMillis());
+        final ScheduledFuture<?> future = scheduler.scheduleAtFixedRate(() -> {
+            try {
+                // Only check the active write index on a regular basis, the others don't change anymore
```
kroepke: What happens if the deflector cycles quicker than our refresh interval?
Could we miss one index in that case?

bernd: We calculate the types on index rotation in `SetIndexReadOnlyAndCalculateRangeJob`. That should catch all indices.

image

PR #4582, opened 2018-02-15
Reduce development startup time by improving globbing.

dennisoelkers:
## Motivation and Context

With a higher number (>5) of plugins in development, running yarn start takes
significant time during the globbing step which finds all webpack configs of
the installed plugins.

By using better glob options (ignoring the `node_modules` directories of
installed plugins and setting the working directory of the glob
explicitly instead of implicitly by including it in the glob pattern),
we are able to reduce globbing times by the factor of 100.

I wrote a small script[1] to compare the times for both options, its
results are (on my machine):

```
glob: 53.336s
glob w/ ignoring + cwd: 0.581s
```

[1]: https://gist.github.com/dennisoelkers/682b8fd25233ec0d2df03107b929e860

## Types of changes

## Checklist:


dennisoelkers:
Oh my. Thanks for catching this, I cherry-picked a bit too sloppily.

PR #4584, opened 2018-02-15
Adjust resolution of dashboard layout for wide screens

### LINKED ISSUE #4306: Widescreen dashboards configuration
I upgraded Graylog from version 2.2.0 to version 2.3.2 yesterday. Unfortunately, one dashboard I use a lot on an Ultra-wide monitor (3440x1440) became sort-of degraded. The dashboard is now limited to only 6 columns and the handful of wastefully wide count widgets causes many of the other widgets I used to see all in one screen to became off-screen with necessity to scroll. To illustrate the problem, I have 13 cm wide widgets on my screen just to see one zero in them, where just half of that width would suffice greatly.

I think this "new feature" introduced somewhere along the way between the two versions mentioned was not thoroughly thought of. Could you please point me to the discussion on this topic, if such one took place, so I can understand the motivation?

**As a solution for this inconvenience**: I would like
* either the old behaviour available (unlimited columns count with fixed cell width),
* configurable maximal columns count,
* more intelligent dynamic column count limitation (e.g. based on the browser window width) or
* smaller minimum width of a single dashboard grid cell with twice as many columns as they are now.

_Thank you very much for considering this change request._

kmerz:
## Description
Double the resolution of the dashboard layout grid. And add a migration
to adjust the existing dashboard widgets positions and settings. 

## Motivation and Context
If you have a big screen one column takes 1/6 of the screen. If you have
a static number widget than 1/6 is a huge widget just for showing one number.

## How Has This Been Tested?
- Add a dashboard with a couple of widgets and place them so their positions get stored to the database
- Apply migration and load new code by restarting the server
- Go to the dashboard page and the widgets position and sizes shouldn't have changed
- Ensure that the amount of columns have doubled
- Restart server to ensure that the migration did not apply a second time

## Types of changes

## Checklist:

Fixes #4306


graylog2-server/src/main/java/org/graylog2/migrations/V20180214093600_AdjustDashboardPositionToNewResolution.java
```diff
@@ -0,0 +1,123 @@
...
+    public void upgrade() {
+        if (clusterConfigService.get(MigrationCompleted.class) != null) {
+            LOG.debug("Migration already completed!");
+            return;
+        }
+
+        Map<String, String> dashboardIds = Maps.newHashMap();
+        for (Dashboard dashboard : dashboardService.all()) {
+            Map<String, Object> fields = dashboard.getFields();
+            BasicDBObject positions = (BasicDBObject) fields.get(DashboardImpl.EMBEDDED_POSITIONS);
```
dennisoelkers: Having to use parts of an implementation class (`Dashboard_Impl_`) on an upper abstraction level is always a sign that there is missing functionality in the implemented interface. I think we should implement the functionality to get the position of a widget in a structured/typed format (this way we can also get rid of all the following casts and direct field accesses), modify the positions and save them back using `DashboardService#updateWidgetPositions`.

graylog2-server/src/main/java/org/graylog2/dashboards/Dashboard.java
```diff
@@ -26,6 +26,10 @@
 
     void setTitle(String title);
 
+    Map<String, Map<String, Integer>> getPositions();
+
+    void setPostions(Map<String, Map<String, Object>> positions);
```
dennisoelkers: - Typo: `setPostions` -> `setPositions`
- Instead of `Map<String, Map<String, Object>>` we should use a structured & stricter typed parameter type here, e.g. an auto-value based value class.

graylog2-server/src/main/java/org/graylog2/dashboards/Dashboard.java
```diff
@@ -26,6 +26,10 @@
 
     void setTitle(String title);
 
+    Map<String, Map<String, Integer>> getPositions();
```
dennisoelkers: Instead of `Map<String, Map<String, Object>>` we should use a structured & stricter typed result type here, e.g. an auto-value based value class.

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardImpl.java
```diff
@@ -72,6 +76,32 @@ public void setDescription(String description) {
         this.fields.put(FIELD_DESCRIPTION, description);
     }
 
+    @Override
+    public Map<String, Map<String, Integer>> getPositions() {
+        final Map<String, Map<String, Integer>> result = new HashMap<>();
+
+        final BasicDBObject positions = (BasicDBObject) fields.get(DashboardImpl.EMBEDDED_POSITIONS);
+        if (positions == null) {
+            return null;
```
dennisoelkers: Instead of returning `null`, we should return an empty result instead.

graylog2-server/src/main/java/org/graylog2/dashboards/widgets/WidgetPosition.java
```diff
@@ -0,0 +1,33 @@
...
+@AutoValue
+public abstract class WidgetPosition {
+
+    public abstract String id();
+    public abstract Integer width();
+    public abstract Integer height();
+    public abstract Integer col();
+    public abstract Integer row();
+
+    public static WidgetPosition create(String id, Integer width, Integer height, Integer col, Integer row) {
```
dennisoelkers: Instead of using a static factory method, I would suggest using a builder to create `WidgetPosition` instances. We have four attributes of the same type here, making it prone for mixups. E.g.:

While reading the following usage of the factory method:

`WidgetPosition.create(id, 3, 2, 1, 4)`

Can you tell me which numeric parameter is which, from the top of your head? Or:

`WidgetPosition.create(id, height, width, row, col)`

Can you tell if the parameters are in the right order without looking up the definition?

By using a builder pattern, it would like this:

```
WidgetPosition.builder()
    .height(height)
    .width(width)
    .row(row)
    .column(col)
    .build()
```

Making it possible to understand it right on the spot.

graylog2-server/src/main/java/org/graylog2/bundles/Dashboard.java
```diff
@@ -35,6 +36,8 @@
     @NotNull
     private List<DashboardWidget> dashboardWidgets = Collections.emptyList();
 
+    private List<WidgetPosition> positions;
```
dennisoelkers: Why did you add this? From what I can see, the getter/setter is never used.

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardImpl.java
```diff
@@ -72,6 +79,42 @@ public void setDescription(String description) {
...
+            return;
+        }
+        final Map<String, Map<String, Object>> positions = new HashMap<>(widgetPositions.size());
+        for (WidgetPosition widgetPosition : widgetPositions) {
+            Map<String, Object> position = new HashMap<>(4);
+            position.put("width", widgetPosition.width());
+            position.put("height", widgetPosition.height());
+            position.put("col", widgetPosition.col());
+            position.put("row", widgetPosition.row());
+            positions.put(widgetPosition.id(), position);
```
dennisoelkers: While this would work, it is unnecessarily complex and error-prone. What would happen if an attribute is added to the `WidgetPosition` class, but forgotten to transfer it in here?

My suggestion is this: Make the `WidgetPosition` class serializable to JSON by adding `@JsonProperty` annotations to all getters (and `@JsonIgnore` `id()`) and just put the value class in the dashboard field.

kmerz: After a short discussion with @bernd we came to the conclusion that for now it would be too much work to use the ```JsonProperty``` here since the underlying ```PersistendImpl``` cannot use it.

dennisoelkers: :+1:

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardServiceImpl.java
```diff
@@ -107,20 +110,19 @@ public Dashboard load(String id) throws NotFoundException {
 
     @Override
     public void updateWidgetPositions(Dashboard dashboard, WidgetPositionsRequest positions) throws ValidationException {
-        Map<String, Map<String, Object>> map = Maps.newHashMap();
+        final List<WidgetPosition> widgetPositions = new ArrayList<>();
```
dennisoelkers: We can use the size of `positions.positions()` to allocate the correct size of the array right from the start.

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardServiceImpl.java
```diff
@@ -107,20 +110,19 @@ public Dashboard load(String id) throws NotFoundException {
...
 
         for (WidgetPositionsRequest.WidgetPosition position : positions.positions()) {
```
dennisoelkers: What happens if `positions` is `null`?

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardServiceImpl.java
```diff
@@ -107,20 +110,19 @@ public Dashboard load(String id) throws NotFoundException {
...
+                    .width(position.width())
+                    .height(position.height())
+                    .col(position.col())
+                    .row(position.row())
+                    .build());
         }
 
-        dashboard.getFields().put(DashboardImpl.EMBEDDED_POSITIONS, map);
-
+        dashboard.setPositions(widgetPositions);
```
dennisoelkers: What happens if `dashboard` is `null`?

graylog2-server/src/main/java/org/graylog2/dashboards/DashboardImpl.java
```diff
@@ -72,6 +79,49 @@ public void setDescription(String description) {
...
+        final Map<String, Map<String, Integer>> positions = new HashMap<>(widgetPositions.size());
+        for (WidgetPosition widgetPosition : widgetPositions) {
+            Map<String, Integer> position = new HashMap(4);
+            position.put("width", widgetPosition.width());
+            position.put("height", widgetPosition.height());
+            position.put("col", widgetPosition.col());
+            position.put("row", widgetPosition.row());
+            positions.put(widgetPosition.id(), position);
+        }
+        getFields().put(DashboardImpl.EMBEDDED_POSITIONS, positions);
```
dennisoelkers: What would happen if `getFields()` returns `null`? (`fields` can be set directly using the constructor)

PR #4598, opened 2018-02-20
Remove Jest snapshots

joschi:
This PR removes some accidentally committed snapshots directories of Jest and adds them to the `.gitignore` file to prevent this from happening in the future.

Refs facebook/jest#1650
Refs https://facebook.github.io/jest/docs/en/snapshot-testing.html


edmundoa:
Those snapshots are used in frontend tests and need to be in the repository. Snapshot tests compare results of generating the DOM tree for the component under test with the stored snapshot, helping us to spot unintentional changes in the component tree.

PR #4606, opened 2018-02-22
Retain legacy behavior of Beats input

joschi:
Instead of changing the behavior of the old Beats input (`BeatsCodec`), this change set reverts the changes of Graylog2/graylog-plugin-beats#29 and creates a new input ( `Beats2Input`, `Beats2Codec`) instead.

This way, existing users don't need to change anything and new users can use the new Beats input with the improved field handling introduces in Graylog2/graylog-plugin-beats#29.

Closes Graylog2/graylog-plugin-beats#33

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java
@@ -1,0 +1,178 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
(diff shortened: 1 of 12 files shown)


joschi:
## Example messages (sent with Filebeat)

*Beats Legacy*
beats-legacy

*Beats (new)*
beats-new

pastukhov:
Hi!
Is it possible to use new verison with graylog 2.4.5?

joschi:
@pastukhov No, unless you're going to backport the changes yourself and create a custom build. But nothing official and nothing we'd support in any way.

PR #4619, opened 2018-02-27
Add SelectPopover component

edmundoa:
This PR introduces a `SelectPopover` component that we want to use on actions and filters in `TableList` components:

screen shot 2018-02-27 at 17 26 24

Some features of this component:

- Customise how to display elements via a function
- Consumers of this component can control its selected item and get notified when it changes
- Optional text filter element
- Scrollable list if there are more than 10 items

### DIFF
--- a/graylog2-web-interface/src/components/common/SelectPopover.md
+++ b/graylog2-web-interface/src/components/common/SelectPopover.md
@@ -1,0 +1,46 @@
+```js
+const createReactClass = require('create-react-class');
+const Button = require('react-bootstrap').Button;
+
+const items = [
+  'Black',
+  'Blue',
+  'Green',
+  'Red',
+  'White',
+  'Yellow',
+];
+
+const SelectPopoverExample = createReactClass({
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/components/common/SelectPopover.md
```diff
@@ -95,7 +95,7 @@ const SelectPopoverFormattedExample = createReactClass({
                          filterPlaceholder="Filter by colour" />
         </div>
         
-        {selectedItem ? `You have selected ${selectedItem}` : 'Please select a colour!'}
+        {selectedItem ? `You have selected ${selectedItem}` : 'Please select another colour!'}
```
mariussturm: colour -> color

edmundoa: http://www.dictionary.com/browse/colour?s=t

edmundoa: 😄

PR #4627, opened 2018-03-01
Add NullAway static code analysis

joschi:
This commit adds Uber's NullAway Error Prone plugin.

> NullAway is a tool to help eliminate NullPointerExceptions (NPEs) in your Java code.

See also:
* https://github.com/uber/NullAway
* https://eng.uber.com/nullaway/

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
@@ -58,7 +58,7 @@ public class NodeImpl extends PersistedImpl implements Node {
 
     @Override
     public DateTime getLastSeen() {
-        return new DateTime(((Integer) fields.get("last_seen")) * 1000L, DateTimeZone.UTC);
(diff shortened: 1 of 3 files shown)


joschi:
@dennisoelkers Splitting the activation of NullAway and the fixes into multiple PRs would lead to a non-buildable master – unless we deactivate NullAway which kind of contradicts the idea of this pull request.

As a compromise, I could split the activation of NullAway and the fixes into multiple *commits* and you could [rebase and merge](https://help.github.com/articles/about-pull-request-merges/#rebase-and-merge-your-pull-request-commits) this PR which retains the individual commits.

dennisoelkers:
The idea was to create/merge the PR fixing the problems _first_, then the NullAway PR.

joschi:
@dennisoelkers I've created a separate PR for the NPE fix in `NodeImpl`.

Upon thinking about the usefulness of NullAway, I think we're fine without it right now (plus it would require more changes to `graylog-parent`/`graylog-plugin-parent` so that plugins using Google Auto Service will still work).

PR #4636, opened 2018-03-05
Remove Cobertura Maven plugin

joschi:
The Cobertura Maven plugin (and Cobertura itself) doesn't support Java 8.

Additionally, code coverage metrics are currently not used anywhere in the build, so we can as well remove it.

Refs mojohaus/cobertura-maven-plugin#21
Refs cobertura/cobertura#166

### DIFF
--- a/graylog-project-parent/pom.xml
+++ b/graylog-project-parent/pom.xml
@@ -820,10 +820,6 @@
                             </execution>
                         </executions>
                     </plugin>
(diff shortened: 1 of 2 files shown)


joschi:
If at one point we decide to reinstate code coverage checks, we could use [JaCoCo](http://www.jacoco.org/), which supports Java 8 and 9 at the time of writing.

PR #4644, opened 2018-03-09
Create assembly in package phase automatically

bernd:
To make sure the shade plugin runs before the assembly plugin in the
package phase, we have to move the assembly plugin definition after the
shade plugin definition in pom.xml.


bernd:
This needs to be backported to 2.4 once merged.

Issue #4459, opened 2018-01-05
Sporadic HttpConfiguratonTest failures on Jenkins and Travis


bernd:
## Expected Behavior

The tests for the new HTTP configuration should work.

## Current Behavior

From time to time we are seeing errors when running the `HttpConfigurationTest` class in CI environments. I saw the same issue on Jenkins and Travis so it doesn't seem to be related to a specific CI environment.

Example Travis job: https://travis-ci.org/Graylog2/graylog2-server/builds/325398290

Excerpt:

```
  HttpConfigurationTest.testHttpBindAddressIsIPv6AddressWithoutBrackets 
Expected: (an instance of com.github.joschi.jadconfig.ValidationException and exception with message a string containing "Possible bracketless IPv6 literal: 2001:db8::1")
     but: exception with message a string containing "Possible bracketless IPv6 literal: 2001:db8::1" message was "Couldn't run validator method"
Stacktrace was: com.github.joschi.jadconfig.ValidationException: Couldn't run validator method
	at com.github.joschi.jadconfig.JadConfig.invokeValidatorMethods(JadConfig.java:227)
	at com.github.joschi.jadconfig.JadConfig.process(JadConfig.java:100)
	at org.graylog2.configuration.HttpConfigurationTest.testHttpBindAddressIsIPv6AddressWithoutBrackets(HttpConfigurationTest.java:70)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.rules.ExpectedException$ExpectedExceptionStatement.evaluate(ExpectedException.java:239)
[...]

PR #4650, opened 2018-03-13
Disable unreliable tests.

dennisoelkers:
Some tests of the HttpConfigurationTest suite keep on failing
unpredictably and have proven to be unreliable, causing the need to
manually restart tests on Travis or Jenkins.
(See https://travis-ci.org/Graylog2/graylog2-server/builds/352299567 as
an example)

Therefore, this PR disables these tests until they are fixed.

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/configuration/HttpConfigurationTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/configuration/HttpConfigurationTest.java
@@ -62,7 +62,7 @@ public class HttpConfigurationTest {
     }
 
     @Test
-    @Ignore
+    @Ignore // Disabled test due to being unreliable (see https://github.com/Graylog2/graylog2-server/issues/4459)
     public void testHttpBindAddressIsIPv6AddressWithoutBrackets() throws RepositoryException, ValidationException {
         expectedException.expect(ValidationException.class);
         expectedException.expectMessage("Possible bracketless IPv6 literal: 2001:db8::1");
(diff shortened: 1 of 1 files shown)


graylog2-server/src/test/java/org/graylog2/configuration/HttpConfigurationTest.java
```diff
@@ -61,6 +62,7 @@ public void testHttpBindAddressIsIPv6Address() throws RepositoryException, Valid
     }
 
     @Test
+    @Ignore
```
bernd: Please add a text description to the `@JsonIgnore` annotation that includes the link to the issue:

https://github.com/Graylog2/graylog2-server/issues/4459

bernd:
Refs https://github.com/Graylog2/graylog2-server/issues/4459

graylog2-server/src/test/java/org/graylog2/configuration/HttpConfigurationTest.java
dennisoelkers: 👍

graylog2-server/src/test/java/org/graylog2/configuration/HttpConfigurationTest.java
```diff
@@ -61,6 +62,7 @@ public void testHttpBindAddressIsIPv6Address() throws RepositoryException, Valid
     }
 
     @Test
+    @Ignore // Disabled test due to being unreliable (see https://github.com/Graylog2/graylog2-server/issues/4459)
```
joschi: The [`@Ignore`](http://junit.sourceforge.net/javadoc/org/junit/Ignore.html) annotation has an argument which can be used to display the reason while running the test.
This might be more useful than a code comment.

dennisoelkers: Indeed!

PR #4656, opened 2018-03-14
Add generic data and script dir settings

bernd:
The data dir will be used as base dir for settings like `message_journal_dir` and `content_packs_dir`.

The `bin_path` setting will point to the relative `bin` path by default. The `data_dir` setting will point to the `data` directory by default and acts as the base directory for the `message_journal_dir` and `content_pack_dir` settings.

Config values can be injected into objects by using the `@GraylogDataDir` and `@GraylogBinPath` annotations.

#### `@GraylogBinPath`

```java
     public class ScriptExec {
       @Inject
       ScriptExec(@GraylogBinPath Path binPath, String scriptName) {
           this.scriptName = binPath.resolve(scriptName);
       }
     }
```

#### `@GraylogDataDir` -- with `data_dir` as base directory

```java
     public class Config {
       @Parameter("data_dir")
       private Path dataDir = Paths.get("data");

       @Parameter("message_journal_dir")
       @GraylogDataDir
       private Path messageJournalDir = Paths.get("journal");
     }
  
     public class Journal {
       @Inject
       Journal(@GraylogDataDir("message_journal_dir") Path journalDir) {
           // journalDir == "data/journal"
       }
     }
```

#### `@GraylogDataDir` -- without `data_dir` as base directory

```java
     public class Config {
       @Parameter("data_dir")
       private Path dataDir = Paths.get("data");

       @Parameter("message_journal_dir")
       @GraylogDataDir
       private Path messageJournalDir = Paths.get("/mnt/graylog-journal");
     }
  
     public class Journal {
       @Inject
       Journal(@GraylogDataDir("message_journal_dir") Path journalDir) {
           // journalDir == "/mnt/graylog-journal"
       }
     }
```

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -140,3 +140,13 @@ The following HTTP API paths changed due to the plugin merge:
 +---------------------------------------------------------------------------------------------+-----------------------------------------------+
 | ``/plugins/org.graylog.plugins.pipelineprocessor/system/pipelines/simulate``                | ``/system/pipelines/simulate``                |
 +---------------------------------------------------------------------------------------------+-----------------------------------------------+
+
+New "bin_path" and "data_dir" configuration parameters
+======================================================
+
+We introduced two new configuration parameters related to file system paths.
+
+- ``bin_path`` config option points to the directory that contains scripts like ``graylogctl``.
+- ``data_dir`` option configures the base directory for other data directories like ``message_journal_dir``.
+
+Please check the updated default ``graylog.conf`` configuration file for required changes to your existing file.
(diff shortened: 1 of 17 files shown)


bernd:
@joschi @kroepke The behavior with `data_dir` as base dir for config settings with relative paths can be confusing for users.

I am open to other suggestions.

Maybe it would be enough to have a single configurable `data_dir` and remove separate path settings like `message_journal_dir` and `content_packs_path`.

bernd:
@joschi @kroepke After going away from the computer to take a walk I decided to close this PR in favor of the way simpler https://github.com/Graylog2/graylog2-server/pull/4657.

I think this one is way too complicated and introduces too many possibilities for misconfiguration by users and misuse by developers.

The [new PR](https://github.com/Graylog2/graylog2-server/pull/4657) only adds the `bin_path` and `data_dir` settings without changing the behavior of the existing `message_journal_dir` and `content_packs_dir` settings.

PR #4657, opened 2018-03-14
Add "bin_path" and "data_dir" settings

### LINKED ISSUE #2057: Create generic data directory for Graylog to use
Graylog needs to store a few things on the local filesystem during its operation, e.g.
- journal
- pid file
- node file
- elasticsearch autogenerated files
- self-generated TLS files

Currently each of these is relative to the CWD of the server, but we don't have a common parent directory.
For simplification of setting Graylog up introduce a default data-dir which the other paths use as a relative base path, unless overridden.

bernd:
The `data_dir` config setting can be usd by Graylog subsystems to store data instead of creating separate config settings for that.

Closes #2057

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
@@ -30,6 +30,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
 @SuppressWarnings("FieldMayBeFinal")
 public abstract class BaseConfiguration {
@@ -52,6 +54,12 @@ public abstract class BaseConfiguration {
 
     @Parameter(value = "inputbuffer_wait_strategy", required = true)
     private String inputBufferWaitStrategy = "blocking";
+
+    @Parameter(value = "bin_path", required = true)
+    private Path binPath = Paths.get("bin");
+
+    @Parameter(value = "data_dir", required = true)
+    private Path dataDir = Paths.get("data");
 
     @Parameter(value = "plugin_dir")
     private String pluginDir = "plugin";
@@ -127,6 +135,14 @@ public abstract class BaseConfiguration {
 
     public WaitStrategy getInputBufferWaitStrategy() {
         return getWaitStrategy(inputBufferWaitStrategy, "inputbuffer_wait_strategy");
+    }
+
+    public Path getBinPath() {
+        return binPath;
+    }
+
+    public Path getDataDir() {
+        return dataDir;
     }
 
     public String getPluginDir() {
--- a/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
@@ -26,6 +26,7 @@ import com.lmax.disruptor.BusySpinWaitStrategy;
 import com.lmax.disruptor.SleepingWaitStrategy;
 import com.lmax.disruptor.WaitStrategy;
 import com.lmax.disruptor.YieldingWaitStrategy;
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
```diff
@@ -129,6 +137,14 @@ public WaitStrategy getInputBufferWaitStrategy() {
         return getWaitStrategy(inputBufferWaitStrategy, "inputbuffer_wait_strategy");
     }
 
+    public Path getBinPath() {
+        return binPath;
+    }
+
+    public Path getDataDir() {
```
joschi: Shouldn't this be used as a parent directory for `message_journal_dir` and `content_packs_dir` by default?

joschi:
@bernd Would it make sense to distinguish between data directories which may grow during runtime (such as `message_journal_dir`) and constant size data directories (such as `content_packs_dir`)?

graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
bernd: As written in https://github.com/Graylog2/graylog2-server/pull/4656#issuecomment-373128648 I deliberately didn't touch the existing settings. I tried to do that in #4656 but there was lots of room for error for users and developers.

Using the `data_dir` as a parent for e.g. `message_journal_dir` requires either injecting both values in every place and manually building the real path or something like I build in #4656 which turned out to be complicated and brittle.

Having e.g. `message_journal_dir` use `data_dir` as base directory if it's a relative path but not using `data_dir` as base if it's an absolute path is pretty confusing for users. (IMHO) Always using `data_dir` as base path also doesn't make sense because then we could just remove the `message_journal_dir` setting. (as mentioned in the second paragraph in https://github.com/Graylog2/graylog2-server/pull/4656#issuecomment-373109961) But that would probably be too strict because the user would have to move all data if the journal should be on a separate disk partition.

Please let me know if you have a better idea. :smiley:

joschi: I simply meant using `getDataDir().resolve("journal")` as default value for `message_journal_dir`. No fancy annotations or magic.

bernd: The `getDataDir()` accessor is not reachable from the `KafkaJournalConfiguration`.

joschi: That's a problem we can solve, once we decide what the desired outcome is. 😉

kroepke: Can we run code to set sensible defaults or is that too late?

bernd: @kroepke I don't think so. That's what led me down this annotation magic path in #4656.

bernd: @joschi True. We have to keep in mind that we also need this for plugins, though.

kroepke: Personally I like setting the default as @joschi suggested, but if it is problematic right now I could live without it for now.

bernd: We can use the default when we move the `message_journal_dir` setting from the `KafkaJournalConfiguration` into `BaseConfiguration`.

bernd:
@joschi I think constant sized data can be stored in the configuration directory and the growing data in the `data_dir`.

bernd:
@joschi @kroepke How do we want to continue here? I need this soon.

bernd:
@joschi @kroepke ping

joschi:
@bernd See https://github.com/Graylog2/graylog2-server/pull/4687 for proposed changes.

Issue #4360, opened 2017-11-17
Extract utility module from various modules/repos


kroepke:
There are helpful utilities, abstractions and customizations all across the various repositories in Graylog.

Some are inaccessible because they live in plugins, but could be useful in the core, some are not usable because they live in the enterprise code base, but are generic enough that they should be usable in the OSS code as well. In other places we have similar helpers, that could easily be unified.

The objective is to create an independent maven module, that lives in the server repository, but can be used without pulling in the entire server. By nature it will require some of our core dependencies, namely Guava, Guice, Jackson, Mongojack/MongoDB and others. Depending on the work we want to invest, we could split the module up, or just make a first step in extracting the utility code.

If you know a candidate for extraction, please comment on this issue and we'll update the following list:

 - [ ] NameAndIdDbService from legacy enterprise repository, pattern is followed in the [lookup tables](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/lookup/db/DBCacheService.java), too. It allows naming DB objects and can load objects via both their internal ObjectId and the name, which is more user friendly for URLs etc.
 - [ ]  Extract entity search options into a more reusable component. Newer interfaces have the ability to perform server-side searches on domain objects with a mini github-like language. This is used across the enterprise plugins and as well in the lookup tables. This should be possible to do on any domain object, including standard options for pagination.
 - [ ] Make Jackson AutoValueTypeIdResolver available. It allows using JsonSubTypes with AutoValue builders and is commonly used in enterprise products to implement versioned value classes. This is generally useful for APIs so it should be extracted.

PR #4668, opened 2018-03-19
Add PaginatedDbService class

bernd:
This class can be used as base class for Mongojack DB services. It provides implementations for CRUD actions, paginated searches and stream queries.

I moved PaginatedList from rest models to the database package because PaginatedDbService is using it now.

Refs #4360


graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * This class is a helper to implement a basic Mongojack-based database service that allows CRUD operations on a
+ * single DTO type and offers paginated access.
+ * <p>
+ *     It makes only a few assumptions, which are common to many Graylog entities:
+ *     <ul>
+ *         <li>The DTO class has a name which is unique</li>
```
kroepke: This comment is obsolete, right?

bernd:
graylog-project/pr failure is due to required changes in the threatintel plugin. See: https://github.com/Graylog2/graylog-plugin-threatintel/pull/100

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -0,0 +1,220 @@
...
+     */
+    protected PaginatedList<DTO> findPaginatedWithQueryFilterAndSort(DBQuery.Query query,
+                                                                     Predicate<DTO> filter,
+                                                                     DBSort.SortBuilder sort,
+                                                                     int page,
+                                                                     int perPage) {
+        final AtomicInteger total = new AtomicInteger(0);
+
+        // First created a filtered stream
+        final Stream<DTO> dtoStream = streamQueryWithSort(query, sort)
```
joschi: What's the impact of the intermediary stream (and the following flat-mapping) on memory consumption?
I could imagine that peeking into each result item (and this loading every document from MongoDB) has some effect on the overall memory consumption.

Could it make sense to execute two MongoDB queries here, one returning the total number of documents matching the query and one returning the requested page with limit and offset in the MongoDB query itself instead of being filtered afterwards?

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Returns a stream of database entries for the given {@link DBQuery.Query} sorted by the give {@link DBSort.SortBuilder}.
+     *
+     * @param query the query to execute
+     * @param sort the sort order for the query
+     * @return stream of database entries that match the query
+     */
+    protected Stream<DTO> streamQueryWithSort(DBQuery.Query query, DBSort.SortBuilder sort) {
+        return Streams.stream((Iterable<DTO>) db.find(query).sort(sort));
```
joschi: When would this cursor be closed?

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
joschi: And do we require the total number of documents in every case or could this be an additional database request if required?

graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
```diff
@@ -80,13 +79,8 @@ public String toString() {
 
     @JsonAutoDetect
     public class PaginationInfo {
```
joschi: Could this be converted to an Auto Value class in order to get rid of the custom `equals()`/`hashCode()`/ `toString()` implementations?

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
bernd: Yes, this will increase memory consumption because it needs to iterate over all objects. The main reason we do this is not the counting of total objects but to execute the filter `Predicate` for each object returned by the `DBQuery`.

One example where we have to use a filter `Predicate` are permission checks, where we query a list of objects from the database and need to filter that list for objects the user has access to.
In that case we cannot use the `limit()` and `skip()` methods on the `DBCursor` but have to do that in memory to get correct total counts and page counts.

Consumers should use `findPaginatedWithQueryAndSort()` if possible to avoid this overhead. But in some cases there is no way around that.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Returns an unordered stream of database entries for the given {@link DBQuery.Query}.
+     *
+     * @param query the query to execute
+     * @return stream of database entries that match the query
+     */
+    protected Stream<DTO> streamQuery(DBQuery.Query query) {
+        return Streams.stream((Iterable<DTO>) db.find(query));
```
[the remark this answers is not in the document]
bernd: Good point. I will update the code to add a `Stream#onClose()` handler to close the `DBCursor` and will document the need for closing the stream.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
bernd: See above.

graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
bernd: Aye, will do.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -0,0 +1,234 @@
...
+        return save.getSavedObject();
+    }
+
+    /**
+     * Deletes the {@link DTO} for the given ID from the database.
+     *
+     * @param id ID of the {@link DTO} to delete
+     */
+    public void delete(String id) {
+        db.removeById(new ObjectId(id));
```
joschi: Please return the `WriteResult` in this method so that callers can check it if required.

bernd: Do you think we should return the complete `WriteResult<DTO, ObjectId` or would it be enough to return the value of `WriteResult.getN()`? This is what we do in other parts of the server and it would avoid leaking the `WriteResult` object to the caller.

bernd: Sorry, somehow missed the second part of your sentence. I will return the number of deleted documents. :+1:

joschi: ✅

Issue #1914, opened 2016-03-08
How to turn the value of a link to a hyperlink


123dev:
### Problem description

Not sure if I'm asking this in the right area, mods, please kindly move this if it is in the wrong area.

We use Graylog across multiple applications, our data crosses multiple systems, and all our logs include a custom field: correlation_id
This id is carried over when data flows from one application to another, so that we can easily find and correlate log entries.
This has proven to be extremely important to our operations and can't imagine working efficiently without it.

The way we currently use this is copy the text field
open another graylog browser window
search for correaltion_id:<copied_value>
to get to the list of log entries with the matching id.

We'd like to take this to the next level and have a smooth navigation and are looking for guidance on how we can  turn this specific field value into a hyperlink
We have created an application (kind of url shortener / mapper) that takes this in simple form
https://123hub/?cid=C58MyDc
and redirects to:
https://graylog/search?rangetype=relative&fields=source%2Ccorrelation_id%2Chttp_status%2Clevel%2Cmessage%2Curi_stem&width=1465&relative=0&q=correlation_id%3AC58MyDc#fields=source%2Ccorrelation_id%2Chttp_status%2Clevel%2Cmessage%2Curi_stem

This is of-course customizable and can be used with different keywords.

So all we're looking for is a way to turn a field value to a hyperlink.

Has something similar have already been done?
by a plugin? or do we need to modify the graylog source and build it oursleves?

Any guidance is greatly appreciated.
Thanks for an awesome product.
### Environment
- Graylog Version: 1.3.2 (AWS Image)

kroepke:
Hi, yes, this is the right place to ask :)

Currently Graylog does not support this kind of plugin yet, but we are looking into it and are planning to support this in the future.
However, it will likely not make it into 2.0.
If you have more requirements, please feel free to either file additional feature requests or add details to this issue if they are relevant to hyperlinks.

I'll leave this open until the time we are spec'ing out the feature.

Thanks,
Kay

123dev:
Thank Kay for the follow up.
Happy to hear that there is interest in supporting this in the future.

I will definitely not hesitate to filing other feature requests or adding details to this if our requirements change.

Would this ticket be updated when this feature is supported? so that we get to be notified?

In the meantime, if you can think of recommending any other mechanism (browser addons or anything) that we can use / code to get a temporary solution would be welcome.
Would making changes to this code be in the right area?
[...]

PR #4669, opened 2018-03-19
Adding hyperlinked fields in search result details

pbr0ck3r:
## Description
This PR allows graylog to have hyperlinked search result fields. A user can create a decorator for a field that they want to allow to be a hyperlink.

Closes this issue/feature request: https://github.com/Graylog2/graylog2-server/issues/1914

## How Has This Been Tested?
I built graylog via `mvn compile`
I then ran the frontend and backend servers.
Tested saving config with several fields, not fields, etc.
Made sure that the search page details rendered accordingly.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldConverter.java
+++ b/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldConverter.java
@@ -1,0 +1,104 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.decorators;
+
+import com.floreysoft.jmte.Engine;
+import com.floreysoft.jmte.template.Template;
+import com.floreysoft.jmte.template.VariableDescription;
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.assistedinject.Assisted;
+import org.graylog2.plugin.Message;
+import org.graylog2.plugin.configuration.ConfigurationRequest;
+import org.graylog2.plugin.configuration.fields.BooleanField;
+import org.graylog2.plugin.configuration.fields.TextField;
+import org.graylog2.plugin.decorators.SearchResponseDecorator;
+import org.graylog2.rest.models.messages.responses.ResultMessageSummary;
+import org.graylog2.rest.resources.search.responses.SearchResponse;
+
+import javax.inject.Inject;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.requireNonNull;
+
+public class LinkFieldConverter implements SearchResponseDecorator {
+
+    private static final String CK_LINK_FIELD = "link_field";
+
+    private final String linkField;
+
+    public interface Factory extends SearchResponseDecorator.Factory {
+        @Override
+        LinkFieldConverter create(Decorator decorator);
+
+        @Override
+        LinkFieldConverter.Config getConfig();
+
+        @Override
+        LinkFieldConverter.Descriptor getDescriptor();
+    }
+
+    public static class Config implements SearchResponseDecorator.Config {
+
+        @Override
+        public ConfigurationRequest getRequestedConfiguration() {
+            return new ConfigurationRequest() {
+                {
+                    addField(new TextField(
+                            CK_LINK_FIELD,
+                            "Link field",
+                            "message",
+                            "The field that will be transformed into a hyperlink."
+                    ));
+                }
+            };
+        }
+    }
+
+    public static class Descriptor extends SearchResponseDecorator.Descriptor {
+        public Descriptor() {
+            super("Hyperlink String", "http://docs.graylog.org/", "Hyperlink string");
+        }
+    }
+
+    @Inject
+    public LinkFieldConverter(@Assisted Decorator decorator, Engine templateEngine) {
+        this.linkField = (String) requireNonNull(decorator.config().get(CK_LINK_FIELD),
+                                                   CK_LINK_FIELD + " cannot be null");
+    }
+
+    @Override
+    public SearchResponse apply(SearchResponse searchResponse) {
+        final List<ResultMessageSummary> summaries = searchResponse.messages().stream()
+                .map(summary -> {
+                    final Message message = new Message(ImmutableMap.copyOf(summary.message()));
+                    if (summary.message().containsKey(linkField)) {
+                        final String href = (String) summary.message().get(linkField);
+                        final LinkedHashMap object = new LinkedHashMap<String, String>();
+                        object.put("type", "a");
+                        object.put("href", href);
+                        message.addField(linkField, object);
+                    }
+                    return summary.toBuilder().message(message.getFields()).build();
+                })
+                .collect(Collectors.toList());
+
+        return searchResponse.toBuilder().messages(summaries).build();
+    }
+}
--- a/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
+++ b/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
@@ -37,7 +37,7 @@ import java.util.stream.Collectors;
 
 import static java.util.Objects.requireNonNull;
 
-public class LinkFieldConverter implements SearchResponseDecorator {
+public class LinkFieldDecorator implements SearchResponseDecorator {
 
     private static final String CK_LINK_FIELD = "link_field";
 
@@ -78,7 +78,7 @@ public class LinkFieldConverter implements SearchResponseDecorator {
     }
 
     @Inject
-    public LinkFieldConverter(@Assisted Decorator decorator, Engine templateEngine) {
+    public LinkFieldConverter(@Assisted Decorator decorator) {
         this.linkField = (String) requireNonNull(decorator.config().get(CK_LINK_FIELD),
                                                    CK_LINK_FIELD + " cannot be null");
     }
@@ -87,14 +87,15 @@ public class LinkFieldConverter implements SearchResponseDecorator {
     public SearchResponse apply(SearchResponse searchResponse) {
         final List<ResultMessageSummary> summaries = searchResponse.messages().stream()
                 .map(summary -> {
-                    final Message message = new Message(ImmutableMap.copyOf(summary.message()));
-                    if (summary.message().containsKey(linkField)) {
-                        final String href = (String) summary.message().get(linkField);
-                        final LinkedHashMap object = new LinkedHashMap<String, String>();
-                        object.put("type", "a");
-                        object.put("href", href);
-                        message.addField(linkField, object);
+                    if (!summary.message().containsKey(linkField)) {
+                      return summary;
                     }
+                    final Message message = new Message(ImmutableMap.copyOf(summary.message()));
+                    final String href = (String) summary.message().get(linkField);
+                    final Map<String, String> decoratedField = new HashMap<>();
+                    decoratedField.put("type", "a");
+                    decoratedField.put("href", href);
+                    message.addField(linkField, decoratedField);
                     return summary.toBuilder().message(message.getFields()).build();
                 })
                 .collect(Collectors.toList());
--- a/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
+++ b/graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
@@ -31,6 +31,7 @@ import org.graylog2.rest.resources.search.responses.SearchResponse;
 
 import javax.inject.Inject;
 import java.util.HashMap;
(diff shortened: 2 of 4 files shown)


graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -54,6 +55,13 @@ const MessageFieldDescription = React.createClass({
 
     return termsMarkup;
   },
+  _getFieldValue() {
+    if (this.props.linkedFields.includes(this.props.fieldName)) {
+      return <a href={this.props.possiblyHighlight(this.props.fieldName)}>{this.props.possiblyHighlight(this.props.fieldName)}</a>
```
joschi: You probably don't want to use the highlighted version of the field name as hyperlink reference. 😉

pbr0ck3r: Good catch. Will make that fix.

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -54,6 +55,13 @@ const MessageFieldDescription = React.createClass({
 
     return termsMarkup;
   },
+  _getFieldValue() {
+    if (this.props.linkedFields.includes(this.props.fieldName)) {
+      return <a href={this.props.fieldName}>{this.props.fieldName}</a>
```
joschi: Ah, now you're missing the highlight in the content of the `<a>` tag. 😉

```javascript
return <a href={this.props.fieldName}>{this.props.possiblyHighlight(this.props.fieldName)}</a>
```

graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
```diff
@@ -63,6 +63,8 @@
             .add("file")
             .add("source_file")
             .build();
+    private static final Set<String> DEFAULT_LINKED_SEARCH_FIELDS = ImmutableSet.<String>builder()
```
joschi: You could simply write `ImmutableSet.of()` or `Collections.emptySet()`.

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
pbr0ck3r: Fixed this!

graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
pbr0ck3r: Changed!

graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java
```diff
@@ -63,6 +63,7 @@
             .add("file")
             .add("source_file")
             .build();
+    private static final Set<String> DEFAULT_LINKED_SEARCH_FIELDS = Collections.emptySet();
```
joschi:
```text
graylog2-server/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchesClusterConfig.java:66: error: cannot find symbol
    private static final Set<String> DEFAULT_LINKED_SEARCH_FIELDS = Collections.emptySet();
```
This won't work unless you import the `Collections` class.

pbr0ck3r: Fixed!

edmundoa:
Hi @pbr0ck3r,

In first place, thank you very much for your contribution! 🙂 

We would love to add this feature to the product, but we have been discussing a bit about the implementation and we think this would be a better fit for a [frontend decorator](http://docs.graylog.org/en/2.4/pages/plugins/decorators.html). Would you be willing to rework the feature so it uses a decorator to convert a field into a link?

Please also create future PRs based on the current master branch, as that contains the latest code under development, making it easier to integrate your changes into the product. We also do not add new features to version branches, they are kept to let people build a certain Graylog version exactly as it was. Thank you!

pbr0ck3r:
Sorry I ddin't branch off master. I was using the gitlab-cli which had me by default on 2.4 branch. I will branch off master and push the new code to this PR when I have time to re-write it in the coming weeks.

pbr0ck3r:
I am assuming you all want me to make a decorator plugin that I would maintain myself. If so then this PR can then be closed.

edmundoa:
@pbr0ck3r we are happy to merge the decorator in our code and maintain it (we already have [a few decorators](https://github.com/Graylog2/graylog2-server/tree/master/graylog2-server/src/main/java/org/graylog2/decorators))! I'll keep the PR open if you want to work on it. You are of course free to create a plugin with the decorator if you want to.

pbr0ck3r:
How would users specific specific fields they want decorated? Currently I have a list in the configuration page that allows users to specific fields. In a decorator, that list would have to be hard coded in the config. Since decorators don't have any UI components, how would a user change the fields?

pbr0ck3r:
Never mind... forgot how decorators work. A user can specify the field they want decorated in the pipeline rule. I will work on refactoring this code.

pbr0ck3r:
@edmundoa After starting to refactor my code into a Decorator. I noticed that even if I create a decorator to make a field a link. It does not render as a link. If you look at the current PR I have. I had to modify the MessageField.jsx file in order to get fields to render as hyperlinks (anchor tags). From the decorator even if I pass back a anchor tag for the field, react does not render it as that. I just renders it as a string.

edmundoa:
@pbr0ck3r decorators are not related to pipeline rules (except for `Pipelines Processor Decorator`), and are not configured in pipeline rules, but in the search (or stream search) UI:
screen shot 2018-04-03 at 11 49 12

I would start by creating a new decorator in Java, i.e. _Link Field Converter_, that takes a single field as parameter and converts it into something we can use to render an anchor with the field value used as text and `href` of the link. In the future we could allow users to customise the title or href of the link, maybe using other fields in the message, but I would leave it out of the first version.

As you said, React will not set any strings as HTML for security reasons. Formatting the link as HTML in the backend and rendering it with [`dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) would be an option, but for me it still feels too insecure as that will interpret _any_ HTML in the field and potentially do very bad things.

I think it would be safer to use [`React.createElement()`](https://reactjs.org/docs/react-api.html#createelement) to create the anchor element and set it's `href` and text, as it will at least escape the children and (potentially) only execute code when you click on the link. An example of rendering a link with `createElement` would be: `React.createElement('a', { href: 'http://example.org' }, 'http://example.org')`. So, the _Link Field Converter_ could provide an object like `{decoratedField: {type: 'a', props: { href: 'http://example.org' }, children: 'http://example.org' }}` that we can use to call to `createElement` with the right arguments and create the link.

What do you think?

pbr0ck3r:
@edmundoa before I go and refactor my frontend react code. Is it fine for to be modifying the MessageField.jsx to use React.createElement instead of just creating a anchor tag like I was doing? I see you all's point of making a decorator instead of the configuration list I had before.

edmundoa:
@pbr0ck3r I'm not sure if I understood what you meant, so please ask again if it's not clear :) Modifying `MessageField` is necessary to go with either option, so it's fine changing it. As I commented before, I think using `React.createElement` would be a safer choice than creating the anchor tag directly from the message field as innerHTML. It would also enable other decorators to create other kind of elements as well.

pbr0ck3r:
@edmundoa.. I have pushed up the new `LinkFieldConvertor` decorator and also made the changes we discussed to have the reactjs use `React.createElement`.

edmundoa:
@pbr0ck3r sorry it's taking so long for me to review this, I hope I can make by the end of this week. Thank you for your patience!

graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldConverter.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public SearchResponse apply(SearchResponse searchResponse) {
+        final List<ResultMessageSummary> summaries = searchResponse.messages().stream()
+                .map(summary -> {
+                    final Message message = new Message(ImmutableMap.copyOf(summary.message()));
+                    if (summary.message().containsKey(linkField)) {
+                        final String href = (String) summary.message().get(linkField);
+                        final LinkedHashMap object = new LinkedHashMap<String, String>();
```
edmundoa: Here I see a couple of improvements:

- Use a regular `HashMap`, as we don't really care to maintain order in this case, and it's going to be a bit faster
- Use generics in the local variable you are storing the `Map`, avoiding unchecked calls to `put()` afterwards
- The name `object` is too generic for my taste, I would go with `decoratedField` or something similar

In summary, I would change this line into:
```
final Map<String, String> decoratedField = new HashMap<>();
```

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -50,6 +50,15 @@ class MessageFieldDescription extends React.Component {
     SearchStore.addSearchTerm(this.props.fieldName, this.props.fieldValue);
   };
 
+  _getFieldValue() {
+    if (this.props.isDecorated && ('type' in this.props.fieldValue) && (this.props.fieldValue.type === 'a')) {
+      const link = this.props.fieldValue.href;
+      return React.createElement('a', {href: link}, link);
+    } else {
+      return this.props.renderForDisplay(this.props.fieldName)
```
edmundoa: Here there's a missing semicolon at the end of the line.

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -50,6 +50,15 @@ class MessageFieldDescription extends React.Component {
     SearchStore.addSearchTerm(this.props.fieldName, this.props.fieldValue);
   };
 
+  _getFieldValue() {
```
edmundoa: Please use an arrow function here, to ensure the method is bound to `this` in any case.

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -50,6 +50,15 @@ class MessageFieldDescription extends React.Component {
...
+    if (this.props.isDecorated && ('type' in this.props.fieldValue) && (this.props.fieldValue.type === 'a')) {
+      const link = this.props.fieldValue.href;
+      return React.createElement('a', {href: link}, link);
```
edmundoa: Please add spaces between the object braces: https://eslint.org/docs/rules/object-curly-spacing

graylog2-web-interface/src/components/search/MessageFieldDescription.jsx
```diff
@@ -50,6 +50,15 @@ class MessageFieldDescription extends React.Component {
...
+    } else {
```
edmundoa: The `else` case is not really needed (since we return in the `if` branch) and can be deleted.

pbr0ck3r: Fixed all the requested changes.

graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
```diff
@@ -0,0 +1,105 @@
...
+
+public class LinkFieldDecorator implements SearchResponseDecorator {
+
+    private static final String CK_LINK_FIELD = "link_field";
+
+    private final String linkField;
+
+    public interface Factory extends SearchResponseDecorator.Factory {
+        @Override
+        LinkFieldConverter create(Decorator decorator);
```
edmundoa: There are a bunch of the old `LinkFieldConverter` class name that must be renamed here and in `DecoratorBindings`.

graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public SearchResponse apply(SearchResponse searchResponse) {
+        final List<ResultMessageSummary> summaries = searchResponse.messages().stream()
+                .map(summary -> {
+                    if (!summary.message().containsKey(linkField)) {
+                      return summary;
+                    }
+                    final Message message = new Message(ImmutableMap.copyOf(summary.message()));
+                    final String href = (String) summary.message().get(linkField);
+                    final Map<String, String> decoratedField = new HashMap<>();
```
edmundoa: We need to add imports for `Map` and `HashMap`.

graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
```diff
@@ -0,0 +1,105 @@
...
+import org.graylog2.plugin.Message;
+import org.graylog2.plugin.configuration.ConfigurationRequest;
+import org.graylog2.plugin.configuration.fields.BooleanField;
+import org.graylog2.plugin.configuration.fields.TextField;
+import org.graylog2.plugin.decorators.SearchResponseDecorator;
+import org.graylog2.rest.models.messages.responses.ResultMessageSummary;
+import org.graylog2.rest.resources.search.responses.SearchResponse;
+
+import javax.inject.Inject;
+import java.util.LinkedHashMap;
```
edmundoa: This import is not needed any longer.

graylog2-server/src/main/java/org/graylog2/decorators/LinkFieldDecorator.java
```diff
@@ -30,9 +30,8 @@
 import org.graylog2.rest.resources.search.responses.SearchResponse;
 
 import javax.inject.Inject;
-import java.util.LinkedHashMap;
-import java.util.List;
```
edmundoa: `List` is still used in `apply()`, so this import should not be removed.

pbr0ck3r:
@edmundoa Thanks for making the time to work through this with me! Much appreciated!

edmundoa:
Thank you very much for your contribution @pbr0ck3r! 🎉

PR #4674, opened 2018-03-21
Add HOC generator connecting ES6 class components to stores.

dennisoelkers:
## Motivation and Context

This generator allows ES6 class components to be connected to stores,
although the current version of Reflux allows this only using a mixin,
which is not usable for ES6 classes.

The change provides a `connect()` function which takes the supplied
component class and wraps it in a generated component which connects
itself to the stores defined in the provided mapping. The state supplied
by the stores is then passed to the component as props.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/connect.jsx
+++ b/graylog2-web-interface/src/stores/connect.jsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import { ListenerMethods } from 'reflux';
+import { isFunction } from 'lodash';
+
+/**
+ * Generating a higher order component wrapping an ES6 React component class, connecting it to the supplied stores.
+ * The generated wrapper class passes the state it receives from the stores to the component it wraps as props.
+ *
+ * The `stores` parameter should consist of an object mapping desired props keys to stores. i.e.:
+ * {
+ *  samples: SamplesStore
+ * }
+ *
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/stores/connect.jsx
```diff
@@ -0,0 +1,75 @@
...
+      // Listening to each store.
+      Object.keys(stores).forEach((key) => {
+        const store = stores[key];
+        const cb = (v) => {
+          const newState = {};
+          newState[key] = v;
+          this.setState(newState);
+        };
+
+        this.listenTo(store, cb);
```
edmundoa: I know the changes are already merged, but I'm wondering about it: shouldn't we stop listening to store changes when the component will unmount? At least the Reflux connect mixin is doing that: https://github.com/reflux/refluxjs/blob/v0.2.x/src/connect.js#L24

dennisoelkers: Yes, we are doing it in the same way: https://github.com/Graylog2/graylog2-server/pull/4674/files#diff-1221e180d6167b0a46e7111816d08ac0R41

edmundoa: Thank you, didn't see that 👍

dennisoelkers: Thank you for taking care of this! :+1:

PR #4675, opened 2018-03-21
Add activePage prop to PaginatedList.jsx

bernd:
Previously there was no way to control the active page from outside the component.

This also cleans up some strange code from the automatic create-react-class conversion and fixes some eslint errors.


dennisoelkers:
What kind of strange code do you mean?

bernd:
> What kind of strange code do you mean?

This:

image

Because the codemod generated a function for the `defaultProps`, eslint couldn't validate that non-required props have a default prop defined.

dennisoelkers:
Ah, that's because it was a function with more than a single return-statement before: https://github.com/Graylog2/graylog2-server/blob/84843a8ac5f7d1076f01dae52f6e9a66f69ce802/graylog2-web-interface/src/components/common/PaginatedList.jsx#L30-L37

PR #4684, opened 2018-03-23
Some SearchForm.jsx improvements and fixes

bernd:
- Fix eslint errors and fix deprecation warnings
- Fix potential race condition with the loading state update
- Don't reset loading indicatorin `componentWillReceiveProps()`. This fixes an issue where the loading indicator never showed up.
- Convert the query string input to a controlled component and add `query` and `onQueryChange` props

### DIFF
--- a/graylog2-web-interface/src/components/common/SearchForm.jsx
+++ b/graylog2-web-interface/src/components/common/SearchForm.jsx
@@ -54,6 +54,8 @@ class SearchForm extends React.Component {
   };
 
   static defaultProps = {
+    onReset: null,
+    label: null,
     placeholder: 'Enter search query...',
     wrapperClass: 'search',
     queryWidth: 'auto',
@@ -62,7 +64,9 @@ class SearchForm extends React.Component {
     searchBsStyle: 'default',
     searchButtonLabel: 'Search',
     resetButtonLabel: 'Reset',
+    useLoadingState: false,
     loadingLabel: 'Loading...',
+    children: null,
   };
 
   state = {
@@ -72,6 +76,8 @@ class SearchForm extends React.Component {
   componentWillReceiveProps() {
     this._resetLoadingState();
   }
+
+  queryRef = null;
 
   _setLoadingState = () => {
     if (this.props.useLoadingState) {
@@ -89,12 +95,12 @@ class SearchForm extends React.Component {
     e.preventDefault();
 
     this._setLoadingState();
-    this.props.onSearch(this.refs.query.value, this._resetLoadingState);
+    this.props.onSearch(this.queryRef.value, this._resetLoadingState);
   };
 
   _onReset = () => {
     this._resetLoadingState();
-    this.refs.query.value = '';
+    this.queryRef.value = '';
     this.props.onReset();
   };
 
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -54,6 +62,10 @@ class SearchForm extends React.Component {
   };
 
   static defaultProps = {
+    query: '',
+    onQueryChange: null,
```
edmundoa: I would use `() => {}` as default prop, in that way we don't need to check if the function is `null` or not.

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -88,27 +103,37 @@ class SearchForm extends React.Component {
...
   _onReset = () => {
     this._resetLoadingState();
-    this.refs.query.value = '';
-    this.props.onReset();
+    this.setState({ query: '' }, this.props.onReset);
+  };
+
+  handleQueryChange = (e) => {
+    const query = e.target.value;
+    this.setState({ query: query }, () => {
```
edmundoa: Is there any reason to wait for the state to be updated before calling `this.props.onQueryChange()`? The prop is in the end not getting anything from the state, so if this is really needed we should add a comment.

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -88,27 +103,37 @@ class SearchForm extends React.Component {
   _onSearch = (e) => {
     e.preventDefault();
 
-    this._setLoadingState();
-    this.props.onSearch(this.refs.query.value, this._resetLoadingState);
+    this._setLoadingState().then(() => {
+      this.props.onSearch(this.state.query, this._resetLoadingState);
```
edmundoa: I would add a little comment here or in `_setLoadingState` to clarify why we use a promise in here.

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -88,27 +103,37 @@ class SearchForm extends React.Component {
...
+    });
   };
 
   _onReset = () => {
     this._resetLoadingState();
-    this.refs.query.value = '';
-    this.props.onReset();
+    this.setState({ query: '' }, this.props.onReset);
```
edmundoa: As `this.state.query` is initialised with the value from `this.props.query`, I guess resetting the value should also use `this.props.query`, shouldn't it? If no query was given as prop, in the end it would also use `''` as default.

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -62,21 +74,24 @@ class SearchForm extends React.Component {
     searchBsStyle: 'default',
     searchButtonLabel: 'Search',
     resetButtonLabel: 'Reset',
+    useLoadingState: false,
     loadingLabel: 'Loading...',
+    children: null,
   };
 
   state = {
+    query: this.props.query,
```
edmundoa: I think we should also update the `query` in the state when the prop changes. This is specially needed if there is another way to reset the query, other than the one provided by this component.

graylog2-web-interface/src/components/common/SearchForm.jsx
edmundoa: Resetting the query should also call `this.props.onQueryChange()`.

graylog2-web-interface/src/components/common/SearchForm.jsx
bernd: Good point!

graylog2-web-interface/src/components/common/SearchForm.jsx
bernd: Aye, fixed.

bernd:
@edmundoa I addressed your PR feedback. Thank you!

bernd:
@edmundoa Oh, forgot the changes in `SearchForm.md`

bernd:
@edmundoa Done

PR #4685, opened 2018-03-23
Avoid DOM structure changes when there are no pages to show

bernd:
This fixes a long standing issue with the SearchForm component where the query string disappears when a search doesn't return any results.

The problem was that we had a shortcut in the render function that returned only the children and no page selectors when there are no items to show. This apparently changed the DOM structure in a way that the children of the PaginatedList component have been replaced with new instances.
In the case of a SearchForm component as a PaginatedList child, this reset the query string.


bernd:
> When there are no elements to display, I would expect the consumer of PaginatedList handling the case and displaying some informational text instead of a list of elements, but keeping the pagination controls makes the UI more consistent with other cases, specially if there are no items because a filter didn't match any elements.

Yeah, sounds good to me. :+1:

PR #4691, opened 2018-03-28
Add new Wizard Component to common

kmerz:
## Description
Add new wizard component which will allow a step by step
configuration.

## Motivation and Context
For some more complex configurations it is easier for a user to have a step by step
guide through the configuration. This component should help the developer to create
such a step by step configuration

## How Has This Been Tested?
Add a test which covers all functionallity

screenshot-2018-3-28 graylog - content packs

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Wizard.jsx
+++ b/graylog2-web-interface/src/components/common/Wizard.jsx
@@ -46,34 +46,30 @@ class Wizard extends React.Component {
     this.state = {
       selectedStep: props.steps[0].key,
     };
-
-    this._wizardChanged = this._wizardChanged.bind(this);
-    this._onNext = this._onNext.bind(this);
-    this._onPrevious = this._onPrevious.bind(this);
   }
 
-  _wizardChanged(eventKey) {
+  _wizardChanged = (eventKey) => {
     this.props.onStepChange(eventKey);
     this.setState({ selectedStep: eventKey });
-  }
+  };
 
-  _disableButton(direction) {
+  _disableButton = (direction) => {
     const len = this.props.steps.length;
     const position = direction === 'next' ? (len - 1) : 0;
     return this.props.steps[position].key === this.state.selectedStep;
-  }
+  };
 
-  _onNext() {
+  _onNext = () => {
     this._wizardChanged(this.props.steps[this._getSelectedIndex() + 1].key);
-  }
+  };
 
-  _onPrevious() {
+  _onPrevious = () => {
     this._wizardChanged(this.props.steps[this._getSelectedIndex() - 1].key);
-  }
+  };
 
-  _getSelectedIndex() {
+  _getSelectedIndex = () => {
     return this.props.steps.map(step => step.key).indexOf(this.state.selectedStep);
-  }
+  };
 
   render() {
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/components/common/Wizard.jsx
```diff
@@ -0,0 +1,109 @@
...
+      <Row className="content">
+        <Col md={2} className={WizardStyle.subnavigation}>
+          <Nav stacked bsStyle="pills" activeKey={this.state.selectedStep} onSelect={this._wizardChanged}>
+            {this.props.steps.map((navItem) => {
+              return (<NavItem key={navItem.key} eventKey={navItem.key}>{navItem.title}</NavItem>);
+            })}
+          </Nav>
+          <br />
+          <Row>
+            <Col md={6}>
```
edmundoa: Small issue: On an [`md` screen size](https://getbootstrap.com/docs/3.3/css/#grid-options), the previous and next buttons are overlapped. I think we could make both buttons smaller, i.e. `bsSize="small"`, avoiding the overlapping and still looking alright.

Normal buttons:
screen shot 2018-04-02 at 14 04 14

Small buttons:
screen shot 2018-04-02 at 14 04 51

edmundoa: We should also use `sm` or `xs` to set the grid size for these buttons instead of `md`. In that way both buttons will keep their grid layout even if the screen is a bit smaller, instead of expanding into different rows. We usually don't care too much about how things look on small screens, but this one is easy to solve and I thought it was worth mentioning.

graylog2-web-interface/src/components/common/Wizard.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  _wizardChanged(eventKey) {
+    this.props.onStepChange(eventKey);
+    this.setState({ selectedStep: eventKey });
+  }
+
+  _disableButton(direction) {
+    const len = this.props.steps.length;
+    const position = direction === 'next' ? (len - 1) : 0;
+    return this.props.steps[position].key === this.state.selectedStep;
```
edmundoa: Sometimes wizards need the user to complete the current step before going any further. I guess that's something we will need to take care of sooner or later, but I'm okay if we wait until we really need it.

graylog2-web-interface/src/components/common/Wizard.jsx
```diff
@@ -0,0 +1,105 @@
...
+      <Row className="content">
+        <Col md={2} className={WizardStyle.subnavigation}>
+          <Nav stacked bsStyle="pills" activeKey={this.state.selectedStep} onSelect={this._wizardChanged}>
+            {this.props.steps.map((navItem) => {
+              return (<NavItem key={navItem.key} eventKey={navItem.key}>{navItem.title}</NavItem>);
+            })}
+          </Nav>
+          <br />
+          <Row>
+            <Col md={6} sm={6} xs={6}>
```
edmundoa: If you want the same layout in all devices you can use `xs={6}`. Bootstrap will apply that for that screen size and larger ones :)

PR #4713, opened 2018-04-09
Move expanded list from enterprise to server

kmerz:
## Description
This change adds **ExpandableList** and **ExpandableListItem** from the enterprise plugin.
It also supplies documentation and tests for the components.

## Motivation and Context
These components are needed for content selection in Content Pack.

## How Has This Been Tested?
I add new tests, and ensured that the reports content selection still worked.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/ExpandableListItem.jsx
+++ b/graylog2-web-interface/src/components/common/ExpandableListItem.jsx
@@ -11,8 +11,13 @@ class ExpandableListItem extends React.Component {
   static propTypes = {
     /** Is the Item checked */
     checked: PropTypes.bool,
-    /** TODO: i do not know what this means */
-    undetermined: PropTypes.bool,
+    /**
+     * Indicates whether the checkbox on this item should be in an indetermined state or not.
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/common/ExpandableListItem.jsx
```diff
@@ -0,0 +1,102 @@
...
+import style from './ExpandableListItem.css';
+
+/**
+ * The ExpandableListItem is needed to render a ExpandableList.
+ */
+class ExpandableListItem extends React.Component {
+  static propTypes = {
+    /** Is the Item checked */
+    checked: PropTypes.bool,
+    /** TODO: i do not know what this means */
```
edmundoa: Now that I think of it, the name is actually not correct, it should be called `indetermined` instead. I would document it as:

>Indicates whether the checkbox on this item should be in an indetermined state or not. This is mostly helpful to represent cases where the element is only partially checked, for instance when `ExpandableListItem`'s child is an `ExpandableList` and some of its items are checked, but others are not.

PR #4723, opened 2018-04-15
Allow passing measureBeforeMount to ReactGridContainer.

dennisoelkers:
This PR introduces a new prop for the `ReactGridContainer` component
named `measureBeforeMount`, which is passed to the WidthProvider HOC. It
triggers that the width of the grid is measured before any widgets are
rendered, so the grid is initialized initially with the correct width
(instead of defaulting to 1280 for the first render).

Without using the prop, it defaults to `false` which was the case
before, so no existing code is affected by it.


kmerz:
:+1: LGTM

Issue #4548, opened 2018-01-31
Grok Debugger in web interface


joschi:
The Graylog web interface should contain a "Grok Debugger" inspired by https://grokdebug.herokuapp.com/ and using the existing Grok patterns in Graylog so that users can play around and test Grok patterns.

The output should be sufficiently detailed so that bugs in Grok patterns or missing Grok patterns can be easily spotted.

jalogisch:
https://www.elastic.co/guide/en/kibana/current/grokdebugger-getting-started.html

PR #4720, opened 2018-04-12
More detailed error message for missing Grok patterns

joschi:
If a Grok pattern couldn't be resolved, the java-grok library emitted an error message which didn't include the name of the missing pattern.

After upgrading to the latest version of (our repackaged) java-grok, the error message includes the name of the missing Grok pattern to help users to solve the problem.

Fixes #3949

PR #4725, opened 2018-04-16
Grok editor

kmerz:
## Description
Add helpful error messages and add a test grok pattern button.

## Motivation and Context
Creating or editing grok patterns was hard for a unknowleged user.
The form did not have help text and you could not test your grok patterns.

## How Has This Been Tested?
Provieded wrong or broken grok patterns and saw the errors.
Add sample data and tested grok patterns with it.

screenshot-2018-4-16 graylog - grok patterns 1

## Types of changes

## Checklist:

Fixes: #4548

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
+++ b/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
@@ -69,7 +69,6 @@ public class AuditEventTypes implements PluginAuditEventTypes {
     public static final String EXTRACTOR_UPDATE = PREFIX + "extractor:update";
     public static final String GETTING_STARTED_GUIDE_OPT_OUT_CREATE = PREFIX + "getting_started_guide_opt_out:create";
     public static final String GROK_PATTERN_CREATE = PREFIX + "grok_pattern:create";
-    public static final String GROK_PATTERN_TEST = PREFIX + "grok_pattern:test";
     public static final String GROK_PATTERN_DELETE = PREFIX + "grok_pattern:delete";
     public static final String GROK_PATTERN_IMPORT_CREATE = PREFIX + "grok_pattern_import:create";
     public static final String GROK_PATTERN_UPDATE = PREFIX + "grok_pattern:update";
@@ -195,7 +194,6 @@ public class AuditEventTypes implements PluginAuditEventTypes {
             .add(GETTING_STARTED_GUIDE_OPT_OUT_CREATE)
             .add(GROK_PATTERN_CREATE)
             .add(GROK_PATTERN_DELETE)
-            .add(GROK_PATTERN_TEST)
             .add(GROK_PATTERN_IMPORT_CREATE)
             .add(GROK_PATTERN_UPDATE)
             .add(INDEX_SET_CREATE)
--- a/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
+++ b/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
@@ -195,6 +195,7 @@ public class AuditEventTypes implements PluginAuditEventTypes {
             .add(GETTING_STARTED_GUIDE_OPT_OUT_CREATE)
             .add(GROK_PATTERN_CREATE)
             .add(GROK_PATTERN_DELETE)
+            .add(GROK_PATTERN_TEST)
             .add(GROK_PATTERN_IMPORT_CREATE)
             .add(GROK_PATTERN_UPDATE)
             .add(INDEX_SET_CREATE)
--- a/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
+++ b/graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
@@ -69,6 +69,7 @@ public class AuditEventTypes implements PluginAuditEventTypes {
     public static final String EXTRACTOR_UPDATE = PREFIX + "extractor:update";
     public static final String GETTING_STARTED_GUIDE_OPT_OUT_CREATE = PREFIX + "getting_started_guide_opt_out:create";
     public static final String GROK_PATTERN_CREATE = PREFIX + "grok_pattern:create";
+    public static final String GROK_PATTERN_TEST = PREFIX + "grok_pattern:test";
     public static final String GROK_PATTERN_DELETE = PREFIX + "grok_pattern:delete";
     public static final String GROK_PATTERN_IMPORT_CREATE = PREFIX + "grok_pattern_import:create";
     public static final String GROK_PATTERN_UPDATE = PREFIX + "grok_pattern:update";
--- a/graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
+++ b/graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
@@ -19,8 +19,11 @@ package org.graylog2.grok;
 import com.google.common.base.Strings;
 import com.google.common.collect.MapMaker;
 import com.google.common.collect.Sets;
+import io.thekraken.grok.api.Grok;
 import io.thekraken.grok.api.GrokCompiler;
+import io.thekraken.grok.api.Match;
 import io.thekraken.grok.api.exception.GrokException;
+import org.apache.logging.log4j.core.util.Patterns;
 import org.graylog2.database.NotFoundException;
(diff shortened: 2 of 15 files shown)


kmerz:
Either @edmundoa or @dennisoelkers

joschi:
FWIW, I'd prefer having the Grok test page separate from creating Grok patterns.

Typically, users don't create individual Grok patterns for each log pattern they have but use the "basic" Grok patterns provided by Graylog.

edmundoa:
@kmerz @joschi I'm not sure how people use the page, but from the top of my head there are two use cases for using the test:

- Ensure the created/edited pattern works with some examples, and modify the pattern if does not
- Test a pattern when the extracted text is not what the user expect, and modifying the pattern if needed

I agree that people don't usually create patterns themselves, but I guess one would expect to be able to modify the pattern in the same view as the test occurs. What do you think if we would hide the test options by default and display them on demand? In that way people can still check grok patterns without having a separate view for that.

kmerz:
@joschi @edmundoa It is really useful to have it on edit or create pattern page. I would rather leave it there. The first time I came here I did not understand anything and would have liked to have something to play around with. It is now quite easy to make a Debugger Page but I do not see quite the workflow. Perhaps you can provide a workflow by answering the following questions: 

- When do you use the Grok-Debugger aside from editing or creating a Grok Pattern?
- What do you do with the Grok Pattern you tested when not Saving it in either a Extractor or a new Grok Pattern?

I would also say that people do not usually create their own patterns because it is so damn hard without something to test on.

What do you miss especially on this Implementation from the Grok Debugger you provided? I see that a auto completion would be helpful but also a lot of work.

joschi:
> When do you use the Grok-Debugger aside from editing or creating a Grok Pattern?

Whenever people want to find out how to parse a given log message they've encountered.

Recent examples:
* https://community.graylog.org/t/cant-manage-particular-type-of-log-extractof-dont-works/4997
* https://community.graylog.org/t/grok-extractor-with-pipes-and-spaces/4832

> What do you do with the Grok Pattern you tested when not Saving it in either a Extractor or a new Grok Pattern?

Using it in a Grok extractor or in a pipeline rule with `grok()`.

graylog2-server/src/main/java/org/graylog2/audit/AuditEventTypes.java
```diff
@@ -72,6 +72,7 @@
     public static final String EXTRACTOR_UPDATE = PREFIX + "extractor:update";
     public static final String GETTING_STARTED_GUIDE_OPT_OUT_CREATE = PREFIX + "getting_started_guide_opt_out:create";
     public static final String GROK_PATTERN_CREATE = PREFIX + "grok_pattern:create";
+    public static final String GROK_PATTERN_TEST = PREFIX + "grok_pattern:test";
```
joschi: Do we really need an audit event for that?

Typically, we only log modifying actions in the audit log, but not read actions.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
```diff
@@ -100,6 +109,23 @@ public GrokPattern listPattern(@ApiParam(name = "patternId", required = true)
...
+    @Path("/test")
+    @ApiOperation(value = "Test pattern with sample data")
+    @AuditEvent(type = AuditEventTypes.GROK_PATTERN_TEST)
+    public Response testPattern(@ApiParam(name = "pattern", required = true) GrokPatternTestRequest request) {
+        String result;
+        try {
+            result = grokPatternService.match(request.grokPattern(), request.sampleData());
+        } catch(GrokException e) {
+            Map<String, String> error = new HashMap(1);
+            error.put("message", e.getMessage());
```
joschi:
```java
ImmutableMap.of("message", e.getMessage());
// or
Collections.singeltonMap("message", e.getMessage());
```

graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
```diff
@@ -34,6 +34,8 @@
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.Preconditions.checkNotNull;
```
joschi: Minor nitpick: This *could* be replaced by [`Objects#requireNonNull()`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-).

graylog2-server/src/main/java/org/graylog2/rest/resources/system/GrokResource.java
```diff
@@ -17,26 +17,34 @@
 package org.graylog2.rest.resources.system;
 
 import com.codahale.metrics.annotation.Timed;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
+import net.minidev.json.JSONObject;
```
joschi: Wrong import.

joschi:
FWIW, there's also some overlap with #4720.

kmerz:
@joschi Then I suggest that your PR should go in first and I will rebase on it.

graylog2-server/src/main/java/org/graylog2/grok/InMemoryGrokPatternService.java
```diff
@@ -19,21 +19,27 @@
 import com.google.common.base.Strings;
 import com.google.common.collect.MapMaker;
 import com.google.common.collect.Sets;
+import io.thekraken.grok.api.Grok;
 import io.thekraken.grok.api.GrokCompiler;
+import io.thekraken.grok.api.Match;
 import io.thekraken.grok.api.exception.GrokException;
+import org.apache.logging.log4j.core.util.Patterns;
```
joschi: Wrong import?

graylog2-server/src/test/java/org/graylog2/rest/resources/system/GrokResourceTest.java
```diff
@@ -30,6 +31,7 @@
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
+import scala.collection.parallel.ParIterableLike;
```
joschi: Wrong import?

dennisoelkers:
Another thing I have noticed: Exceptions generated by testing a pattern end up in the server's log. As these are in a way expected exceptions which do not have negative effect on processing, these should not be logged.

dennisoelkers:
I can still reproduce pattern testing errors ending up in the server log. E.g. when a pattern contains an undefined key, an `IllegalArgumentException` is thrown, which is not catched. The stack trace looks like this:

```
java.lang.IllegalArgumentException: No definition for key 'USER' found, aborting
	at io.thekraken.grok.api.GrokCompiler.compile(GrokCompiler.java:152) ~[grok-0.1.8-graylog.jar:?]
	at io.thekraken.grok.api.GrokCompiler.compile(GrokCompiler.java:102) ~[grok-0.1.8-graylog.jar:?]
	at io.thekraken.grok.api.GrokCompiler.compile(GrokCompiler.java:98) ~[grok-0.1.8-graylog.jar:?]
	at org.graylog2.grok.MongoDbGrokPatternService.match(MongoDbGrokPatternService.java:114) ~[classes/:?]
	at org.graylog2.rest.resources.system.GrokResource.testPattern(GrokResource.java:121) ~[classes/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_131]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_131]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_131]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_131]
[...]
```

dennisoelkers:
Another improvement: When an error is thrown, it is shown in a toastr notification, which vanishes after a few seconds. I think this is unfortunate, because you will miss it if you are not fast enough. Showing any error in a separate element _inside_ the form would be an improvement imo.

dennisoelkers:
Unfortunately, this change prevents importing a pattern file with patterns depending on each other. It seems like before this change, only the syntax was checked during a bulk import, but not if used keys had been defined before.

PR #4729, opened 2018-04-17
Kafka 1.0.1 compatibility changes

muralibasani:
Currently Graylog2 is limited to run with Kafka 0.9.0.1 or lower versions. This change overcomes it, by being comptabile with newer version of kafka (1.0.1)., but downside is it is not compatible with old versions of Kafka (0.9.0.1)  or lower.

## Description

- Changes do not work with Kafka older versions (0.9.0.1 or lower)
- Frontend Kafka inputs have new fields to accept SSL parameters, bootstrap servers, auto.offset.reset to earliest/latest instead of largest/smallest
- In Frontend : based on if SSL i set to true or false, corresponding ssl parameters will be accepted. 

## Motivation and Context
Change is required to make Graylog2 run with new Kafka version 1.0.1, and avail new features of it like ssl and better streaming capabilities.

## How Has This Been Tested?
Tested with Multiple nodes, multiple partitions
Installed in Ubuntu and tested all changes
Only affects Kafka classes

## Types of changes

- [x ] New feature  New feature - Kafka new version 1.0.1, and SSL, with no backwards compatibility

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] I have updated the documentation accordingly.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] I have added tests to cover my changes.
- [ x] All new and existing tests passed.


graylog2-server/src/main/java/org/graylog2/commands/journal/JournalShow.java
```diff
@@ -75,7 +75,7 @@ private void appendSegmentDetails(KafkaJournal journal, StringBuffer sb) {
             sb.append("\t\t").append("Segment ").append(i++).append("\n");
             sb.append("\t\t\t").append("Base offset: ").append(segment.baseOffset()).append("\n");
             sb.append("\t\t\t").append("Size in bytes: ").append(segment.size()).append("\n");
-            sb.append("\t\t\t").append("Created at: ").append(new DateTime(segment.created(), DateTimeZone.UTC)).append("\n");
+            sb.append("\t\t\t").append("Created at: ").append(new DateTime(segment.size(), DateTimeZone.UTC)).append("\n");
```
joschi: I somehow doubt that this is correct. 😉

joschi:
@muralibasani I'm not sure about that but would it be possible to use a current version of org.apache.kafka:kafka-streams (or org.apache.kafka:kafka-clients) while keeping org.apache.kafka:kafka_2.11:0.9.0.1 to retain compatibility with the on-disk message journal?

joschi:
Also see the discussion in #2582.

graylog2-server/src/main/java/org/graylog2/commands/journal/JournalShow.java
pbr0ck3r: Segment.created() has been moved from a public method to a private method.

joschi: @pbr0ck3r This doesn't make it more correct.

What's the meaning of a timestamp generated out of the size of a segment on disk? 😉

pbr0ck3r: @joschi I wasn't inferring that it was correct. I was just stating that from looking at the Segment library code that it is no longer a public method and has been moved to private. It does not look like they have created any helper/alternative methods to get the create datetime.

muralibasani: .size() is indeed not correct. Corrected it.

muralibasani:
Regarding your comment, Kafka current version client could not consume messages from 0.9. 

At somepoint of time, migration has to happen to keep it rolling with new Kafka features. Replicator or mirror maker could be used as an option for customers. ?

graylog2-server/src/main/java/org/graylog2/commands/journal/JournalShow.java
```diff
@@ -75,7 +75,7 @@ private void appendSegmentDetails(KafkaJournal journal, StringBuffer sb) {
             sb.append("\t\t").append("Segment ").append(i++).append("\n");
             sb.append("\t\t\t").append("Base offset: ").append(segment.baseOffset()).append("\n");
             sb.append("\t\t\t").append("Size in bytes: ").append(segment.size()).append("\n");
-            sb.append("\t\t\t").append("Created at: ").append(new DateTime(segment.created(), DateTimeZone.UTC)).append("\n");
+            sb.append("\t\t\t").append("Created at: ").append(new DateTime(segment.time().milliseconds(), DateTimeZone.UTC)).append("\n");
```
joschi:
```text
[ERROR] /var/lib/jenkins/workspace/graylog-project-pr-snapshot/.repos/graylog2-server/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java:[391,68] cannot find symbol
  symbol:   method time()
  location: variable segment of type kafka.log.LogSegment
```

joschi:
@muralibasani Since Kafka 0.9.0.1 and Kafka Streams 1.x seem happily live together in the same process, why not create a new Kafka transport so that in the end we have a "Kafka Legacy Input" (which is the old input) and a new "Kafka Input" which works with newer versions of Kafka and supports TLS?

This way we could also keep using the message journal (Kafka Log) as it is and keep it backward compatible.

graylog2-server/pom.xml
```diff
@@ -460,6 +460,11 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka_2.11</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams</artifactId>
+            <version>1.0.1</version>
```
joschi: Why not use the latest version here?
https://mvnrepository.com/artifact/org.apache.kafka/kafka-streams/1.1.0

pbr0ck3r: @joschi... @muralibasani and I worked on this together and we choose to upgrade to 1.0.1 as it was the easiest to do from 9.0.1 and also provided the most value. Upgrading to 1.1.0 requires some more code rewriting on the graylog side as more API calls changes and code is deprecated/refactored. While it would provide some value, I see it as something that can be done in a future PR.

joschi: @pbr0ck3r I'd prefer to use the latest stable version right away instead of having multiple PRs.

muralibasani:
- time() method fixed and upgraded to 1.1.0
- backwards compatibility yet to work upon.

graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java
```diff
@@ -231,7 +234,20 @@ public KafkaJournal(@Named("message_journal_dir") Path journalDirectory,
...
+
+        java.util.Set<String> hs = new java.util.HashSet<>();
+
+        List<String> testHs = new ArrayList<String>();
+        //JavaConversions.asScalaBuffer(testHs).thisCollection().
+
+        //JavaConversions.asScalaBuffer(testHs);
+
+        //scala.Predef.Set[scala.Predef.String]
+        scala.collection.immutable.Set<String> overridenConfig = null;//JavaConversions.asScalaBuffer(testHs);
```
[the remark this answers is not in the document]
pbr0ck3r: Seems that in the 1.1.0 upgrade the time() method is no longer private. So this issue is irrelevant. Time() method is only private when running 1.0.1 kafka

muralibasani:
Closing this request and opened a new pull request https://github.com/Graylog2/graylog2-server/pull/4770 to accomodate the changes mentioned.

PR #4733, opened 2018-04-18
Add support for int values in SearchQuery

edmundoa:
This PR adds support for querying `int` values in `SearchQuery`, and adapts the default search operator, so that users can still find `int` and `Date` values even if they don't specify an operator.

With the changes, these are the default operators:
- `foo` -> Behaves like before the changes
- `string_field:foo` -> Behaves like before the changes
- `int_field:0` -> Uses an `equals` operator to find `0` in `int_field`
- `date_field:2018-04-18T12:15:00Z` -> Uses an `equals` operator to find `2018-04-18T12:15:00Z` in `date_field`

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/search/SearchQueryParser.java
+++ b/graylog2-server/src/main/java/org/graylog2/search/SearchQueryParser.java
@@ -238,6 +238,8 @@ public class SearchQueryParser {
                 return new FieldValue(parseDate(pair.getLeft()), pair.getRight(), negate);
             case STRING:
                 return new FieldValue(pair.getLeft(), pair.getRight(), negate);
+            case INT:
+                return new FieldValue(Integer.parseInt(pair.getLeft()), pair.getRight(), negate);
             default:
                 throw new IllegalArgumentException("Unhandled field type: " + field.getFieldType().toString());
         }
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/search/SearchQueryParser.java
```diff
@@ -231,15 +233,18 @@ private DateTime parseDate(String value) {
...
+        final Pair<String, SearchQueryOperator> pair = extractOperator(value, fieldType == STRING ? DEFAULT_STRING_OPERATOR : DEFAULT_OPERATOR);
 
-        switch (field.getFieldType()) {
+        switch (fieldType) {
             case DATE:
                 return new FieldValue(parseDate(pair.getLeft()), pair.getRight(), negate);
             case STRING:
                 return new FieldValue(pair.getLeft(), pair.getRight(), negate);
+            case INT:
+                return new FieldValue(Integer.parseInt(pair.getLeft()), pair.getRight(), negate);
```
bernd: Can we also use a `long` instead of an `int` here? This would make it work with very large values as well.

If MongoDB works with both, we could just parse the value into a `long` and rename the data type to `NUMBER` to make it more generic.

edmundoa: So far I only needed an `int` and I tried to only implement that. I can look into it, though.

Anyway, I think `NUMBER` is not really good even with `long` support, because then I would also expect the method to work with `double`, `float`, and maybe other number types in Java. Also makes parsing the number a bit more of a challenge 😄

edmundoa: That didn't work when parsing a number as `long` and then trying to serialize it to a POJO that has an `int` field. Mongojack tries to serialize the `long` value as an `int` and it fails with:
`java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer`

edmundoa:
As we discussed offline, I added a `LONG` type additionally, so users can choose the right one to match their model object.

PR #4739, opened 2018-04-19
Set dashboard title and description with React

dennisoelkers:
*Backport of #4730 for 2.4*

Handle editing of dashboard title and description through React and
Reflux, avoiding the use of jquery and specially `.html`

(cherry picked from commit 5fbb7d326cc55d617cb8fd8106ddb116e37390ad)

### DIFF
--- a/graylog2-web-interface/src/components/dashboard/Dashboard.jsx
+++ b/graylog2-web-interface/src/components/dashboard/Dashboard.jsx
@@ -69,7 +69,7 @@ const Dashboard = React.createClass({
       <li className="stream">
         <h2>
           <LinkContainer to={Routes.dashboard_show(this.props.dashboard.id)}>
-            <a><span ref="dashboardTitle">{this.props.dashboard.title}</span></a>
+            <a>{this.props.dashboard.title}</a>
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/dashboard/Dashboard.jsx
```diff
@@ -68,16 +68,14 @@ const Dashboard = React.createClass({
     return (
       <li className="stream">
         <h2>
-          <LinkContainer to={Routes.dashboard_show(this.props.dashboard.id)}>
-            <a><span ref="dashboardTitle">{this.props.dashboard.title}</span></a>
-          </LinkContainer>
+          <Link to={Routes.dashboard_show(this.props.dashboard.id)}>{this.props.dashboard.title}</Link>
```
edmundoa: `Link` is not imported in the file. Since 2.4 uses a different react-router version than 3.0, I would keep the old `LinkContainer` component and play it safe.

dennisoelkers: Good point! 👍

PR #4742, opened 2018-04-19
Link decorator fixes

edmundoa:
After merging #4669, I realised that there were a couple of problems left:

- Fields decorated with decorators other than `LinkFieldDecorator` threw an error, as we used the `in` decorator to see if we needed to render a link for them, but that operator is only available in objects. I fixed that condition in here, to ensure it works in other decorated fields
- Fields using the `LinkFieldDecorator` were not decorated when being rendered in the message table, as the logic to decorate the field was in the message details. To fix this, I moved the logic up the component tree, and used the already existing `renderForDisplay` function to render the field

screen shot 2018-04-19 at 18 04 05


edmundoa:
The linting errors are not related to the changes, so I'll leave them be to not add more noise in here.

PR #4746, opened 2018-04-20
Sort output config fields stably

### LINKED ISSUE #4745: Output config options ordered randomly
## Expected Behavior

My plugin's output's config options are show in the same order as I added them in the output config class independent of their amount.

## Current Behavior

My plugin's output's config options are show in a random order if there are more than 10.

## Possible Solution

Use a stable sort algorithm while sorting by optionality.

1. Create an output plugin with more than 10 non-optional config options.
2. View the config in the web UI.

## Context

@N-o-X and me are developing a Graylog plugin.

## Your Environment

* Graylog Version: 2.4.3
* Operating System: CentOS 7

Al2Klimov:
## Description

fixes #4745

## Motivation and Context

#4745

## How Has This Been Tested?

Not at all. Unfortunately I couldn't build Graylog independent of my change.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
+++ b/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
@@ -83,7 +83,14 @@ const ConfigurationForm = createReactClass({
   },
 
   _sortByOptionality(x1, x2) {
-    return (this.state.configFields[x1].is_optional - this.state.configFields[x2].is_optional);
+    let diff = this.state.configFields[x1.name].is_optional - this.state.configFields[x2.name].is_optional;
+
+    if (!diff) {
+      // Sort equal fields stably
+      diff = x1.pos - x2.pos;
+    }
+
+    return diff;
   },
 
   _save() {
@@ -153,9 +160,12 @@ const ConfigurationForm = createReactClass({
       shouldAutoFocus = false;
     }
 
-    const configFieldKeys = $.map(this.state.configFields, (v, k) => k).sort(this._sortByOptionality);
+    const configFieldKeys = $.map(this.state.configFields, (field, name) => name)
+      .map((name, pos) => ({ name: name, pos: pos }))
+      .sort(this._sortByOptionality);
+
     const configFields = configFieldKeys.map((key) => {
-      const configField = this._renderConfigField(this.state.configFields[key], key, shouldAutoFocus);
+      const configField = this._renderConfigField(this.state.configFields[key.name], key.name, shouldAutoFocus);
       if (shouldAutoFocus) {
         shouldAutoFocus = false;
       }


Al2Klimov:
PS: I (and @N-o-X and @bobapple) really would like you to include this in v2.4.x.

Al2Klimov:
Unfortunately I seem not to have access to your Jenkins so I can't even figure out my mistake.

Al2Klimov:
As discussed w/ @jalogisch (offline):

* Command executed by Jenkins (effectively): `node node_modules/.bin/eslint -c packages/graylog-web-plugin/.eslintrc -f checkstyle src/components/configurationforms/ConfigurationForm.jsx`
* I fixed my code style errors, but there are some more (of others)
* I'll open a dedicated PR for 2.4.x

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -153,9 +153,12 @@ const ConfigurationForm = createReactClass({
       shouldAutoFocus = false;
     }
 
-    const configFieldKeys = $.map(this.state.configFields, (v, k) => k).sort(this._sortByOptionality);
+    const configFieldKeys = $.map(this.state.configFields, (v, k) => k)
+      .map((v, i) => ({ key: v, pos: i }))
```
edmundoa: I know you didn't wrote that line, but I would vote for replacing the `$.map(this.state.configFields, (v, k) => k)` with the much clearer, in my opinion, `Object.keys(this.state.configFields)`.

Please also use a clearer name for `v`. I suggest `key` instead. In the end we minify the code, so better make it easier to understand.

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -83,7 +83,7 @@ const ConfigurationForm = createReactClass({
   },
 
   _sortByOptionality(x1, x2) {
-    return (this.state.configFields[x1].is_optional - this.state.configFields[x2].is_optional);
+    return this.state.configFields[x1.key].is_optional - this.state.configFields[x2.key].is_optional || x1.pos - x2.pos;
```
edmundoa: I think we should rewrite this in a more explicit way.
Subtracting two booleans is already hard to understand without deep understanding of JS internals, and using the `or` operator to do another comparison when the previous result was 0 only makes it harder to understand.

edmundoa:
I got carried on by the review and I forgot to mention: thank you very much for your contribution! Let's work through the code and get this fix for good 🙂

Al2Klimov:
Hello @edmundoa and thank you for responding!

> we should use another strategy, for instance sorting (...) by key name.

This would be a breaking change for our output plugin and for all the others.
So far plugin authors registered the config options in the desired order (e.g. host, user, password, ...), but this would (1) shuffle all options (of all plugins) (2) during a patchlevel release.

Besides that.. AFAIK no one complained about that.

So I won't change that – at least not in this PR as it doesn't solve the problem.

But you can expect the other changes soon.

Best,
AK

edmundoa:
@Al2Klimov sounds reasonable to me. For what I read, most browsers respect the `Object` property order, even if the standard does not require that. The way I see it, at least the situation will be better than it was before for most people.

To further guarantee that the inputs will appear in the order specified by the author (even if it is implicitly), we would need to explicitly export the order of each key in the API, allowing to read that property from the JS code and sorting by it. That change would be fine for 3.0, but it would be a no-go for 2.4.x.

edmundoa:
@Al2Klimov to be honest, I don't know the reason for it and it probably happened organically. I _guess_ the reason was to let users see the mandatory fields they had to fill out first, in case they didn't pay much attention to the form. Also sometimes configuration fields are put together from different classes and maybe the sorting looked somehow "wrong".

Feel free to open an issue for that so we can discuss it and maybe also change it.

edmundoa:
Thank you for your contribution @Al2Klimov! Do you want to update the 2.4 branch with the changes or should I?

Al2Klimov:
I'll do.

PR #4750, opened 2018-04-23
Implement HTTP proxy server authentication in OkHttpClientProvider

### LINKED ISSUE #4594: Graylog/JVM proxy settings not working when proxy basic auth is needed
My test installation is complaining about “Graylog Enterprise License Violation” and I have to use a proxy to access https://api.graylog.com/report.

## Expected Behavior
Access to https://api.graylog.com/report should be possible when proxy with basic-auth is configured.

## Current Behavior
The BlueCoat proxy uses basic auth and I didn’t get it working via the ```http_proxy_uri``` in the ```server.conf```. **Is there a special format to use when proxy-basic-auth is required?** My ```yum.conf``` for example is successfully using the proxy configuration.

I added the following to the default java options for the JVM:
```GRAYLOG_SERVER_JAVA_OPTS="... -Djdk.http.auth.tunneling.disabledSchemes= -Dhttp.proxyUser=user -Dhttp.proxyPassword=password -Dhttp.proxyHost=myproxy -Dhttp.proxyPort=8080 -Dhttps.proxyUser=user -Dhttps.proxyPassword=password -Dhttps.proxyHost=myproxy -Dhttps.proxyPort=8080"```

I already added ```-Djdk.http.auth.tunneling.disabledSchemes=``` since Java 8u111 basic auth has been disabled by default (http://www.oracle.com/technetwork/java/javase/8u111-relnotes-3124969.html).

But I still keep getting HTTP-407:
```2018-02-13T21:38:56.046+01:00 WARN  [LicenseReportPeriodical] Unable to connect to license server: Failed to authenticate with proxy.```
```2018-02-13T21:43:55.889+01:00 WARN  [LicenseChecker] License violation - Failed to report license status to Graylog, Inc. - consecutive failures: 85, limit: 72```

1. Configure proxy with basic auth in server.conf or via JVM options
2. Restart graylog-server

## Your Environment

* Graylog Version: 2.4.3
* Elasticsearch Version: 5.6.7
* MongoDB Version: 3.2
* Operating System: Oracle Linux 7.4
* Browser version: Firefox Quantum 58.0.2

### PR DESCRIPTION
**Pull request by bernd**
This makes it possible to use a proxy server in `http_proxy_uri` that requires authentication.

Fixes #4594 

**Note:** This needs to be cherry picked into 2.4

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
@@ -20,6 +20,7 @@ import com.github.joschi.jadconfig.util.Duration;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import okhttp3.Authenticator;
+import okhttp3.Challenge;
 import okhttp3.Credentials;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -43,6 +44,8 @@ import java.net.SocketAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.net.HttpHeaders.PROXY_AUTHORIZATION;
@@ -122,6 +125,9 @@ public class OkHttpClientProvider implements Provider<OkHttpClient> {
     }
 
     public static class ProxyAuthenticator implements Authenticator {
+        private static final Logger LOG = LoggerFactory.getLogger(ProxyAuthenticator.class);
+        private static final String AUTH_BASIC = "Basic";
+
         private final String credentials;
 
         ProxyAuthenticator(String user, String password) {
@@ -131,6 +137,16 @@ public class OkHttpClientProvider implements Provider<OkHttpClient> {
         @Nullable
         @Override
         public Request authenticate(@Nonnull Route route, @Nonnull Response response) throws IOException {
+            final Set<String> authenticationMethods = response.challenges().stream()
+                    .map(Challenge::scheme)
+                    .collect(Collectors.toSet());
+
+            if (!authenticationMethods.contains(AUTH_BASIC)) {
+                LOG.warn("Graylog only supports the \"{}\" authentication scheme but the proxy server asks for one of the following: {}",
+                        AUTH_BASIC, authenticationMethods);
+                return null;
+            }
+
             if (response.request().header(PROXY_AUTHORIZATION) != null) {
                 return null; // Give up, we've already failed to authenticate.
             }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java [hunk 1]
```diff
@@ -97,8 +105,36 @@ public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
...
+            this.credentials = Credentials.basic(requireNonNull(user), requireNonNull(password));
+        }
+
+        @Nullable
+        @Override
+        public Request authenticate(@Nonnull Route route, @Nonnull Response response) throws IOException {
+            if (response.request().header(PROXY_AUTHORIZATION) != null) {
+                return null; // Give up, we've already failed to authenticate.
+            }
+            return response.request().newBuilder().addHeader(PROXY_AUTHORIZATION, credentials).build();
```
joschi: We're using basic authentication here, no matter what mechanism the `Proxy-Authentication` request header mentions.

We should at least issue a warning if the proxy requests an authorization mechanism which we don't support (which is everything except "Basic") and fall back to no credentials.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authenticate

graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
```diff
@@ -97,8 +105,36 @@ public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
...
 
         return clientBuilder.build();
     }
+
+    public static class ProxyAuthenticator implements Authenticator {
+        private static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
+        private final String credentials;
+
+        ProxyAuthenticator(String user, String password) {
+            this.credentials = Credentials.basic(requireNonNull(user), requireNonNull(password));
```
joschi: Using `requireNonNull()` is not too useful here, in my opinion. Also see above regarding the edge cases for `Proxy-Authentication`.

If you want to keep the null-check, at least add a message which mentions the parameter which was `null`.

```java
this.credentials = Credentials.basic(requireNonNull(user, "user"), requireNonNull(password, "password"));
```

graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
```diff
@@ -97,8 +105,36 @@ public void connectFailed(URI uri, SocketAddress sa, IOException ioe) {
             };
 
             clientBuilder.proxySelector(proxySelector);
+
+            if (!isNullOrEmpty(httpProxyUri.getUserInfo())) {
+                final List<String> list = Splitter.on(":")
+                        .limit(2)
+                        .omitEmptyStrings()
+                        .splitToList(httpProxyUri.getUserInfo());
```
joschi: Technically, the "user" and the "password" components of the HTTP/1.1 user info can be empty but neither variant is supported if we use this `Splitter` configuration.

Do we want to support these edge cases? They're arguably valid, the question is if we want to support them.

graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/OkHttpClientProvider.java
bernd: Good catch! :+1:

bernd:
@joschi I addressed you comments.

PR #4770, opened 2018-05-02
Kafka upgrade - compatible with old and new versions

muralibasani:
Currently Graylog2 is limited to run with Kafka 0.9.0.1 or lower versions. This change overcomes it, by being comptable with old 0.9.0.1 to 1.1.0 versions.

## Description
Keep org.apache.kafka:kafka_2.11:0.9.01, so that the message journal stays compatible with previous versions of Graylog. (No changes to pom and kafka versions.)

Created a new Kafka transport instead of changing the existing KafkaTransport class and use the new transport to create a new Kafka input. The old input still works with older Kafka brokers.
This way users can decide which version of Kafka they want to support and existing setups won't break.

## Motivation and Context
Change is required to make Graylog2 run with newer kafka versions like 1.1.0 and also does not break with old kafka version 0.9.0.1

## How Has This Been Tested?
Tested with different versions of kafka brokers and tested.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My code follows the code style of this project.
- [ x] I have read the **CONTRIBUTING** document.
- [ x] All new and existing tests passed.


florianvolle:
@muralibasani If you want the PR to be reviewed we kindly ask you to sign the CLA.

muralibasani:
Can someone look at this pr? Not sure if I am missing something else.

bernd:
@muralibasani Thank you, I will take a look shortly.

bernd:
@muralibasani Not sure if I am missing something, but to me it looks like the `KafkaNewTransport` class is still using the old 0.9.x consumer class. Also there is no dependency on Kafka 1.1.0. Is this PR missing some commits? :smiley:

pbr0ck3r:
@bernd Kafka 0.9.0.1 dependency library was kept in the pom as per request from @joschi to allow backwards compatibility. The change in the KafkaNewTransport is really the format Kafka 1.x uses for journals and how graylog consumes them and also the ability to use SSL.

pbr0ck3r:
Just to make sure we are on the same page here @bernd. You want us to include both dependencies for the old comsumer API (9.0.x) and the new consumer API (1.x). We then integrate/use the new libraries in the KafkaNewTransport.
```
<dependencies>
  <dependency>
      <groupId>org.apache.kafka</groupId>
      <artifactId>kafka-consumer</artifactId>
      <version>1.1.0</version>
  </dependency>
</dependencies>```

**Comment by muralibasani** [comment]
Guys, that is not going to work. Changing pom, to include only new kafka version does not work for old brokers. They don't use entirely different packages. kafka.consumer package is used  in both old and new versions. I have ofcourse tested this and knew very well. You can try changing the same and check it too.
New kafka version api fails to fetch the message (from old broker) because of message format changes.

**Comment by muralibasani** [comment]
Adding to my previous comment, aim of the PR is to not break existing brokers, and also giving the flexibility to work with new brokers with new transport/ui. It is only possible with this solution.

Please see Joschi comments on Apr 20 in old PR https://github.com/Graylog2/graylog2-server/pull/4729

**Comment by bernd** [comment]
@muralibasani @pbr0ck3r 

> They don't use entirely different packages. kafka.consumer package is used in both old and new versions.

I didn't know about that. I thought the new Java API (`org.apache.kafka:kafka-clients`) was not used in the 0.9.0.1 broker package. What a pity.

That means we have to stay on the 0.9.0.1 version for `org.apache.kafka:kafka_2.11` and `org.apache.kafka:kafka-clients` for now. Understood.

I have some questions to avoid further confusion:

1. `kafka.consumer.Consumer` v0.9.0.1 works with 0.9.0.1 brokers but **not** with 1.x brokers?
2. `org.apache.kafka.clients.consumer.KafkaConsumer` v0.9.0.1 works with 0.9.0.1 brokers and **also** with 1.x brokers?

**Comment by muralibasani** [comment]
@bernd 
For your questions.
kafka.consumer.Consumer v0.9.0.1 works with 0.9.0.1 brokers but not with 1.x brokers?
Yes.,so in new transport we had to use clients api.

org.apache.kafka.clients.consumer.KafkaConsumer v0.9.0.1 works with 0.9.0.1 brokers and also with 1.x brokers?
Yes. But other way round, 1.x api cannot connect/fetch from old broker.

**Comment by WolfspiritM** [comment]
When and how will this be updated? We're currently having a problem with graylog and kafka in docker cause we can't use the "new" seperation of internal and external traffic that was introduced in Kafka 2 years ago!

https://cwiki.apache.org/confluence/display/KAFKA/KIP-103%3A+Separation+of+Internal+and+External+traffic

The error we get:
```
kafka.common.KafkaException: Failed to parse the broker info from zookeeper: {"listener_security_protocol_map":{"INSIDE":"PLAINTEXT","OUTSIDE":"PLAINTEXT"},"endpoints":["INSIDE://kafka:29092","OUTSIDE://10.10.10.10:29094"],"jmx_port":-1,"host":"kafka","timestamp":"1547825661392","port":29092,"version":4},
	at kafka.cluster.Broker$.createBroker(Broker.scala:85) ~[graylog.jar:?],
	at kafka.utils.ZkUtils$$anonfun$getCluster$1.apply(ZkUtils.scala:563) ~[graylog.jar:?],
```

According to google an update of the kafka lib will fix this...

**Comment by pbr0ck3r** [comment]
@bernd Know that this is marked as ready for review. This PR includes support for 1.0.x. But now that 1.1.x and also 2.x are out. What are GL thoughts on upgrading to 1.1.x or even 2.x? Just am trying to get a feel of how GL is going to support kafka in the future.

**Comment by bernd** [comment]
@pbr0ck3r Thanks for the heads up! Yeah, we definitely want to support newer versions of Kafka in the future. We didn't have enough resources to do it yet, unfortunately. We also need to figure out an upgrade path for users that currently use older Kafka setups with Graylog.

**Comment by pbr0ck3r** [comment]
@bernd Cool! I might start looking into kafka 1.1.x and the amount of work it would take to upgrade this PR to 1.1.x. As well as looking into 2.x and how much work it would be to upgrade to that. Also am currently running this PR code in our production instance with no known issues.

**Comment by pbr0ck3r** [comment]
Just a note that this PR works with kafka 1.1.1 (which is the latest 1.1.x) release.

**Comment by pbr0ck3r** [comment]
@bernd Kafka 2.1 works as well with this PR.

**Comment by marcobelmonte** [comment]
For the love of god would someone approve this?

**Comment by arsiesys** [comment]
Hello, does this PR have any chance to be reviewed then merged ? :) 
If the support of Kafka is planned to be removed, can we have an official statement ?

Thanks !

**Comment by aarnaud** [comment]
It will be useful to have a official build with this integration instead of build our custom version. 
@pbr0ck3r use this PR in production since February. 

Thanks guys,

**Comment by kroepke** [comment]
Hi!
We definitely want to support Kafka going forward, but a few things need to happen on our end for integrating this:
 * figure out the compatibility matrix
 * establish the upgrade paths and clearly document them
 * make sure our users' environments don't suddenly break when we make changes
 * and finally having enough resources and people to test all of this

Would splitting the Kafka inputs out into plugins (and thus making it easier to replace/update them) help you at all?

**Comment by arsiesys** [comment]
Hello @kroepke,
Thanks for your answer, that's reassuring :).
Having the Kafka input as plugin could be an excellent alternative and would be really helpful.

**Comment by arsiesys** [comment]
Hello @muralibasani @kroepke, I made this repo with your work as a plugin (dirty probably but working :D).
https://github.com/arsiesys/graylog-kafka-plugin

**Comment by hgnoyke** [comment]
Hi,

is there any timeline for supporting this feature? Maybe only for a limited range of Kafka versions, >= 1.1.x would be fine :-)

We are waiting for this PR because of missing SSL/TLS support and otherwise need alternatives. We would like to stick with Graylog, so maybe there is also the chance to officially provide the proposed graylog-kafka-plugin?

Thanks for a heads-up!

**Comment by bernd** [comment]
@hgnoyke Thanks for the heads-up. We will check if we can include it in the 3.2 release.

**Comment by marcobelmonte** [comment]
The other day this really beautiful girl started talking to me and I could tell she was really interested in me but then I remembered that whenever something really great is about to happen to a person if you can look into the future you'll probably find only pain and agony so despite almost thinking "you know what - this girl is so hot and intelligent maybe I'll be one of the lucky few" and running with it I quickly pulled out my Android and brought up this GitHub issue and quickly found my center and got jerked back into reality. Thank you!

File: graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaNewTransport.java [hunk 1]
```diff
@@ -0,0 +1,491 @@
...
+        ZkClient zkClient = new ZkClient(zkConnect, 30000, 30000, ZKStringSerializer$.MODULE$);
+        ZkUtils zkUtils = ZkUtils.apply(zkClient, false);
+
+        List<String> subscribedTopics = new ArrayList<String>();
+
+        List<String> topics = JavaConversions.seqAsJavaList(zkUtils.getAllTopics());
+
+        topics.stream()
+            .forEach(topicName->{
+                if(topicName!=null && topicName.matches(configuration.getString(CK_TOPIC_FILTER))) {
```
cbsmith: This is duplicating functionality in KafkaClient, which already supports pattern matching with the KafkaClient.subsribe(Pattern) API call, and iterating through all topics like this is potentially really inefficient (and doesn't handle updates to available topics.

It'd be better to avoid requiring Zookeeper as a parameter at all and just let the KafkaClient library handle this.

marcobelmonte:
image

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaNewTransport.java
cbsmith: Sent in a pull request to address this.

muralibasani:
It was long time ago. I will take a look.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaNewTransport.java
muralibasani: I will take a look

mpfz0r:
Thank you @muralibasani and all the others for your help and for your patience.
This is finally done with https://github.com/Graylog2/graylog2-server/pull/7504 , and will be released with 3.3. and backported to 3.2.3

I'd appreciate tests and feedback.
Closing this PR now.

marcobelmonte:
image

pbr0ck3r:
@mpfz0r @bernd awesome to see this get merged in!

PR #4774, opened 2018-05-07
Add `trigger_alert()` function

### LINKED ISSUE #4772: Allow Notifications to be called directly from pipeline
So today in our Graylog environment we witnessed several messages, all properly tagged, not trigger an alert.  The issue is that the messages were delivered late to the cluster (they were delivered late, thus processed a significant amount of time beyond their timestamp).  Thus, as Graylog is running the searches every minute **based on the timestamp of the message** (field content, field 'alert' = 'autotick'), they were not seen in the Graylog alert engine and thus an alert was not fired for those messages.

To prevent such tragedies from occurring in the future, I would like to ask for a new feature - allowing alert notifications to be triggered directly from the pipeline (and allowing notifications to not be bound explicitly to streams).  We are already writing complex, compound conditionals in the pipelines prior to inserting the messages into streams, only to subject them to yet another arbitrary conditional before firing the alert.  Why can we not skip that secondary condition process and call one of our notifications directly from the pipeline?  Sure it's not stateful, but it's surely better than missing an important event.

* Graylog Version: 2.4.4+4659dbe
* Elasticsearch Version: 5.6.9-1
* MongoDB Version: 3.4.14-1.el7
* Operating System: Red Hat Enterprise Linux Server release 7.5 (Maipo)
* Browser version: Google Chrome 66.0.3359.139 on Windows 10

joschi:
Alert conditions in Graylog are evaluated in a fixed interval by `AlertScanner` and `AlertScannerThread`, which might be too inflexible for some users who want to trigger an immediate alert if a specific message is encountered.

In order to enable users to manually trigger alerts in a pipeline rule, this change set adds the `trigger_alert()` function.

The function also enables a primitive kind of "real-time alerting" utilizing pipeline rules.

Fixes #4772.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ProcessorFunctionsModule.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ProcessorFunctionsModule.java
@@ -21,6 +21,7 @@ import com.google.inject.Scopes;
 import com.google.inject.TypeLiteral;
 import com.google.inject.multibindings.MapBinder;
 import org.graylog.plugins.pipelineprocessor.ast.functions.Function;
+import org.graylog.plugins.pipelineprocessor.functions.alerts.TriggerAlert;
 import org.graylog.plugins.pipelineprocessor.functions.conversion.BooleanConversion;
 import org.graylog.plugins.pipelineprocessor.functions.conversion.DoubleConversion;
 import org.graylog.plugins.pipelineprocessor.functions.conversion.IsBoolean;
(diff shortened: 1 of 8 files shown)


dennisoelkers:
Refs #4772.

bernd:
@lennartkoopmann Please check if this is aligns with the alerting roadmap.

lennartkoopmann:
Closing this because it does not align with the upcoming alerting roadmap. Thanks everyone!

dannosaur:
Really? Given my issue of messages being delayed coming in , this would be incredibly handy to have available. Where is this roadmap you speak of, and how does one request that something (this) be added to it?

If this is not on your "roadmap", how might you plan on resolving #4772 then? You've not closed that down, and it's directly related to this.

jasonkeller:
@lennartkoopmann any details on how #4772 gets fixed if not by something like this?  @dannosaur asked but didn't tag you so I presume you haven't seen his question.

Issue #2403, opened 2016-06-22
Ability to filter on Inputs page


coffee-squirrel:
### Problem description

We are using multiple inputs to help separate teams, environments, and applications.  Although we've used a standard naming format and can use a browser's "find" functionality to find the input(s) we're looking for on the Inputs page, it would be useful to have a filtering/searching mechanism similar to what exists on the "Streams" and "User accounts" pages.
### Environment
- Graylog Version: 2.0.3
- Elasticsearch Version:  2.2.0
- MongoDB Version:  2.6.12
- Operating System:  Red Hat Enterprise Linux Server release 6.7
- Browser version:  Firefox 47

PR #4803, opened 2018-05-23
Add filter for inputs

kmerz:
## Description
Add a filter input for inputs to filter inputs on the inputs list page.

## Motivation and Context
If a user has lots of inputs finding the right one can be tiresome with the browser search function.

## How Has This Been Tested?
Add local inputs and filtered them.

screenshot-2018-5-24 graylog - inputs 1

## Types of changes

## Checklist:

Fixes: #2403


kmerz:
@edmundoa Yes you are right, the ```SearchForm``` does what I need, I do not know why I did not take it in the first place even though I found it.

edmundoa:
@kmerz it's hard to find all components that we have, the frontend documentation helps a little, but not sure how much. We also lack some information regarding where to use certain components or others. I guess that would be something to add in the component gallery, too.

graylog2-web-interface/src/components/inputs/InputsList.jsx
```diff
@@ -83,15 +136,28 @@ const InputsList = createReactClass({
...
 
+        <Row id="filter-input" className="content input-new">
+          <Col>
+            <SearchForm onSearch={this._onFilterInputs}
+                        label="Filter inputs"
+                        topMargin={0}
+                        onReset={this._onFilterReset}
+                        wrapperClass={InputsListStyle.filterWrapper}
+                        searchButtonLabel="Filter"
+                        placeholder="Search by title"
```
edmundoa: I think having an input above without any label and another one below it with a label looks kind of funny. Maybe should we try to make both inputs look similar? This is a quick attempt:

Before:
before

After:
after

edmundoa: In any case we should only use one verb in the `SearchForm`, either "search" or "filter".

graylog2-web-interface/src/components/inputs/InputsList.jsx
```diff
@@ -55,7 +65,12 @@ const InputsList = createReactClass({
       localInputs = localInputs.filter(input => input.node === this.props.node.node_id);
     }
 
-    this.setState({ globalInputs: globalInputs, localInputs: localInputs });
+    this.setState({
+      globalInputs: globalInputs,
+      filteredGlobalInputs: globalInputs,
+      localInputs: localInputs,
+      filteredLocalInputs: localInputs,
+    });
```
edmundoa: We also need to update the filtered inputs when props change. For instance, when a new input is added, we should keep the filter or at least reset it. Right now the filter seems to be in place, but the list is not filtered.

edmundoa: Sorry, I meant when the state propagated from the store changes :)

graylog2-web-interface/src/components/inputs/InputsList.jsx
```diff
@@ -83,26 +144,33 @@ const InputsList = createReactClass({
...
+        <Row id="filter-input" className="content input-list">
           <Col md={12}>
-            <h2>
+            <SearchForm onSearch={this._onFilterInputs}
+                        topMargin={0}
+                        onReset={this._onFilterReset}
+                        searchButtonLabel="Filter"
+                        placeholder="Filter by title" />
+            <br />
+            <h1>
```
edmundoa: Looking at this again, `h1` feels too big. Could we make it `h2` again and increase the spacing between global and local inputs? I think that could make the section a bit less prominent, while keeping some separation between them.

graylog2-web-interface/src/components/inputs/InputsList.jsx
```diff
@@ -70,6 +85,52 @@ const InputsList = createReactClass({
...
+    }
+
+    const filterMethod = (input) => {
+      return regExp.test(input.title);
+    };
+    const filteredGlobalInputs = this.state.globalInputs.filter(filterMethod);
+    const filteredLocalInputs = this.state.localInputs.filter(filterMethod);
+    this.setState({
+      filteredGlobalInputs: filteredGlobalInputs,
+      filteredLocalInputs: filteredLocalInputs,
```
edmundoa: I think we are missing to store the `filter` in the state in here.

graylog2-web-interface/src/components/rules/RuleHelper.jsx
```diff
@@ -170,7 +179,11 @@ end`,
...
                 <Row>
                   <Col sm={12}>
-                    <FilterInput onChange={this._filterDescriptors} />
+                    <SearchForm onSearch={this._filterDescriptors}
+                                label="Filter rules"
+                                topMargin={0}
+                                searchButtonLabel="Filter"
+                                onReset={this._onFilterReset} />
                     <div className={`table-responsive ${RuleHelperStyle.marginTab}`}>
                       <PaginatedList totalItems={functionDescriptors.length} pageSize={this.state.pageSize} onChange={this._onPageChange} showPageSizeSelect={false}>
```
edmundoa: I don't think is very important, or if it's related to this change, but the save/reset page is not working as I expected. Examples:

1. Go to page 3 of functions
2. Filter `is`, as that's the function I want to use
3. The pagination says I'm in page 3, but I am actually in page 1

--

1. Filter items with `is`
2. Go to page 2
3. Reset the filter
4. The pagination says I'm in page 2, but I'm in page 1

---

After looking at the code a bit, I think all we are missing is to pass the `activePage` prop to `PaginatedList`. That should make the pager change the page when changed externally.

edmundoa:
LGTM 👍

PR #4818, opened 2018-06-04
Merge sidecar plugin

edmundoa:
Merge updated version of sidecar plugin. This will coexist for a while with the old collector plugin, as we need to give users some time to migrate their collectors.

Refs https://github.com/Graylog2/graylog-plugin-collector/issues/72


bernd:
There seems to be a styling issue with the dropdown menu when selecting a collector for a configuration:

image

bernd:
When filtering for a specific configuration, the user can clear the selection to show all configurations again. The "clear selection" disappears if the user selects one collector/backend in the list while a configuration selection is active. (it's the same for other filter options like "collector")

peek 2018-06-06 18-44

bernd:
There is an issue when the sidecar configuration setting `send_status` is `false`. Some parts of the system cannot handle this.

Rendering a configuration doesn't work:

```
2018-06-06 19:07:32,010 ERROR: org.graylog2.shared.rest.exceptionmappers.AnyExceptionClassMapper - Unhandled exception in REST resource
java.lang.NullPointerException: null
	at org.graylog.plugins.sidecar.services.ConfigurationService.renderConfigurationForCollector(ConfigurationService.java:128) ~[classes/:?]
	at org.graylog.plugins.sidecar.rest.resources.ConfigurationResource.renderConfiguration(ConfigurationResource.java:180) ~[classes/:?]
```

The sidecar details page doesn't render and fails with:

```
warning.js:33 Warning: Failed prop type: The prop `status` is marked as required in `StatusIndicator`, but its value is `null`.
```

```
bluebird.js:1486 Unhandled rejection TypeError: Cannot read property 'cpu_idle' of null
    at Object.formatNodeDetails (webpack:///./src/components/sidecars/sidecars/SidecarStatus.jsx?:87:51)
    at Object.render (webpack:///./src/components/sidecars/sidecars/SidecarStatus.jsx?:191:16)
```

PR #4826, opened 2018-06-08
Preventing index size/time input to hide rotation strategy dropdown.

### LINKED ISSUE #4769: Select Rotation Strategy Dropdown
## Expected Behavior
When I click the 'Select rotation strategy' drop down in the 'Configure index set' page I expect to see all of the options.

## Current Behavior
The drop down falls behind the 'Rotation period (ISO8601 Duration) drop down.

image

1.Click 'System' drop down
2.Select 'indices'
3. Select an index set
4.'Edit Index Set
5.Scroll down and click 'Select rotation strategy'

## Your Environment

* Graylog Version: 2.4.3
* Elasticsearch Version:
* MongoDB Version:
* Operating System: Win 10
* Browser version: Chrome: Version 66.0.3359.139 (Official Build) (64-bit

dennisoelkers:
## Motivation and Context

Before this change, other elements (namely react-bootstrap input
components with addons) overlapped with `Select` inputs due to a higher
`z-index` of the first.

This change is increasing the `z-index` of the `Select`/`react-select`
menu to `3` (react-bootstrap input addons are on `2`) to prevent this.

Fixes #4769.

Should be backported to `2.4`.

Before:
screen shot 2018-06-08 at 10 22 14
screen shot 2018-06-08 at 10 22 25

After:
screen shot 2018-06-08 at 10 22 57
screen shot 2018-06-08 at 10 23 05

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.css
+++ b/graylog2-web-interface/src/components/common/Select.css
@@ -9,3 +9,8 @@
 .select-sm .Select-input > input {
     padding: 0;
 }
+
+:local(.increaseZIndex) .Select-menu-outer {
+    z-index: 3;
+}
+
--- a/graylog2-web-interface/src/components/common/Select.css
+++ b/graylog2-web-interface/src/components/common/Select.css
@@ -11,6 +11,6 @@
 }
 
 :local(.increaseZIndex) .Select-menu-outer {
-    z-index: 3;
+    z-index: 5;
 }
 
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -165,7 +165,7 @@ const Select = createReactClass({
     }
 
     return (
-      <div className={size === 'small' ? 'select-sm' : ''}>
+      <div className={`${size === 'small' ? 'select-sm' : ''} ${this.reactSelectSmStyles.locals.increaseZIndex}`}>
         <SelectComponent ref={(c) => { this._select = c; }}
                          onChange={onReactSelectChange || this._onChange}
                          {...reactSelectProps}


dennisoelkers:
> Is there a way to fix this in a more generic way?

Maybe, maybe not. This happens only when two things are combined:

a) a `react-select` input
b) an input with a addon below a)

To "fix" in a general way, we would have to normalize the z-index of b) to 0 globally. I would not do that right now, because we have only a single form where it broke, and I am sure it has been set to 2 in `react-bootstrap` for a reason.

edmundoa:
I was looking into a similar issue that @bernd found in #4828, and investigating a bit I found the culprit of this behaviour, which is a change in react-select: https://github.com/JedWatson/react-select/commit/b0bc3e33d5155f971c2310cf3e8eb5184f0795d8#diff-895b3c670e6ebd76e91dbce59ff14fdb

I think using a low z-index value by default in the library makes sense, so I think the best way we can solve this issue is increasing the react-select's z-index in our component, as they suggest there. In that way we can set a higher value than the one used by react-bootstrap or ace editor, ensuring the option list will appear above other inputs.

dennisoelkers:
I reimplemented it by increasing the `z-index` for the `react-select`/`Select` menu.

graylog2-web-interface/src/components/common/Select.css
```diff
@@ -8,4 +8,9 @@
 
 .select-sm .Select-input > input {
     padding: 0;
-}
\ No newline at end of file
+}
+
+:local(.increaseZIndex) .Select-menu-outer {
+    z-index: 3;
```
edmundoa: This is unfortunately not good enough to make the options appear on top of the ace gutter, which has a `z-index` of 4. Could you increase the value a bit more, please? In the end I don't think increasing it is a problem, since opening the select options should be displayed on top of other elements in all cases I can think of.

dennisoelkers: Hm, I tested that in the extended search UI which uses a `Select` component next to an Ace input as well and 3 was enough for that. Where do you encounter the problem in the UI, so I can verify the problem/the fix?

dennisoelkers: Ah, that one does not use the gutter, which explains why it works there. :)

PR #4830, opened 2018-06-08
Load pipeline actions and stores through a provider

### LINKED ISSUE #4829: Stores and actions of old pipelines plugin are not injected through provider
While working on the sidecar plugin merge, I saw that stores and actions from the processor pipelines plugin are not injected using one of the providers we use. E.g:
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/pages/RuleDetailsPage.jsx

This may result in components creating more than one instance of the store, potentially leading to data inconsistencies.

## Expected Behavior
Stores and actions are singletons and injected through a provider, e.g. `CombinedProvider`.

## Current Behavior
Stores and actions are imported as usual

edmundoa:
Inject actions and stores from `CombinedProvider`, ensuring they are all singleton and behaving like other actions and stores.

Fixes #4829


edmundoa:
Done.

PR #4846, opened 2018-06-15
Make "keyword" field type enumerable instead of full-text searchable

### LINKED ISSUE #4841: Incorrect field type mapping for "keyword" type
According to https://github.com/Graylog2/graylog2-server/blob/59598088c1c185a81b8ab871be21a862434cf22a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/FieldTypeMapper.java#L55 it is searchable, which isn't true. It is only enumerable.

bernd:
Also add a TODO comment to revisit the ES 2.x "string" data type.

Fixes #4841

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/FieldTypeMapper.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/FieldTypeMapper.java
@@ -55,8 +55,6 @@ public class FieldTypeMapper {
     private static final ImmutableMap<String, FieldTypes.Type> TYPE_MAP = ImmutableMap.<String, FieldTypes.Type>builder()
             .put("keyword", STRING_TYPE) // since ES 5.x
             .put("text", STRING_FTS_TYPE) // since ES 5.x
-            // TODO: Depending on the value of the "index" field, a "string" can be similar to "keyword" or "text".
-            .put("string", STRING_FTS_TYPE) // until ES 2.x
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/FieldTypeMapper.java
```diff
@@ -53,8 +54,9 @@
      */
     private static final ImmutableMap<String, FieldTypes.Type> TYPE_MAP = ImmutableMap.<String, FieldTypes.Type>builder()
             .put("keyword", STRING_TYPE) // since ES 5.x
-            .put("text", STRING_TYPE) // since ES 5.x
-            .put("string", STRING_TYPE) // until ES 2.x
+            .put("text", STRING_FTS_TYPE) // since ES 5.x
+            // TODO: Depending on the value of the "index" field, a "string" can be similar to "keyword" or "text".
+            .put("string", STRING_FTS_TYPE) // until ES 2.x
```
joschi: With Graylog 3.0.0 supporting only ES 5.x and 6.x, I'm not sure we need to support ES 2.x data types. ES 6.x won't be able to read indices written by ES 2.x anyway.

bernd: Good point, yes. :+1: I will remove it.

Issue #4883, opened 2018-07-04
org.graylog2.alerts.AlertConditionFactory - Could not load alert condition '邮件报警' <ee25fba0-ad5f-4a6f-a4f6-9921889bb847>, invalid configuration detected.


ZhangKe4042602:
**I add a graylog AlertCondition "邮件报警" to test graylog plugin alert notify.but when i delete this alertCondition after done test.graylog service log always print error info like this:**
`2018-07-04 02:45:12,150 ERROR: org.graylog2.alerts.AlertConditionFactory - Could not load alert condition '邮件报警' <ee25fba0-ad5f-4a6f-a4f6-9921889bb847>, invalid configuration detected.
2018-07-04 02:45:12,150 ERROR: org.graylog2.streams.StreamServiceImpl - Skipping alert condition.
org.graylog2.plugin.configuration.ConfigurationException: Mandatory configuration field query is missing or has the wrong data type
	at org.graylog2.plugin.configuration.ConfigurationRequest.check(ConfigurationRequest.java:117) ~[graylog.jar:?]
	at org.graylog2.alerts.AlertConditionFactory.createAlertCondition(AlertConditionFactory.java:63) ~[graylog.jar:?]
	at org.graylog2.alerts.AlertServiceImpl.fromPersisted(AlertServiceImpl.java:170) ~[graylog.jar:?]
	at org.graylog2.streams.StreamServiceImpl.getAlertConditions(StreamServiceImpl.java:285) [graylog.jar:?]
	at org.graylog2.periodical.AlertScannerThread.doRun(AlertScannerThread.java:55) [graylog.jar:?]
	at org.graylog2.plugin.periodical.Periodical.run(Periodical.java:77) [graylog.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_151]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_151]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_151]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_151]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_151]
[...]

PR #4884, opened 2018-07-04
Improve log messages for alert conditions and configuration requests

joschi:
Improve error and warning messages for alert conditions by always including the alert condition ID and the stream ID it's connected to.

Refs #4883

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionFactory.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionFactory.java
@@ -48,8 +48,10 @@ public class AlertConditionFactory {
                                                Map<String, Object> parameters,
                                                String title) throws ConfigurationException {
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AlertConditionFactory.java
```diff
@@ -62,8 +64,7 @@ public AlertCondition createAlertCondition(String type,
             final Configuration configuration = new Configuration(parameters);
             requestedConfiguration.check(configuration);
         } catch (ConfigurationException e) {
-            final String conditionTitle = isNullOrEmpty(title) ? "" : "'" + title + "' ";
-            LOG.error("Could not load alert condition " + conditionTitle + "<" + id + ">, invalid configuration detected.");
+            LOG.error("Could not load alert condition {}<{}> on stream <{}>, invalid configuration detected.", conditionTitle, id, stream.getId());
```
dennisoelkers: It would be helpful to include the stream title in here as well.

Issue #4854, opened 2018-06-20
Graylog fails (not fail-fast) to start if mongodb is not initially reachable


dkhamitov:
Hi guys,

**My scenario is as follows:**
1. MongoDB is DOWN
2. Graylog is starting
3. Graylog shows in logs that it fails to connect to MongoDB
4. Graylog behaves as if it would recover once the MongoDB is UP. It can be observed in logs
5. MongoDB is started
6. Graylog opens a connection to MongoDB but then fails with Guice error while trying to dependency-inject
7. Graylog says Terminating and kills itself

**My environment:**
OS: SLES 11.4 (sles)
JRE: Oracle Corporation 1.8.0_144 on Linux 3.0.101-108.35-default
Graylog server 2.4.3+2c41897
Arch: amd64
MongoDB: 3.0.0
Elasticsearch: 5.6.9
MongoDB + Elasticseacrh + Graylog are all on the same host
Every URL in Graylog configuration is pointing to 127.0.0.1

[Here](https://github.com/Graylog2/graylog2-server/files/2119681/issue.log) is the log of the above scenario

I would expect that Graylog should eventually connect successfully to MongDB or fail-fast in this case. What do you think? Is this possible at all?

kmerz:
I can confirm the problem!

Once mongodb is reachable graylog will terminate!

PR #4891, opened 2018-07-09
Fail-fast if mongodb is not initially reachable (#4854)

dkhamitov:
Hi guys,
It's a fix for #4854. As it's turned out, Graylog eventually fails to start but it takes about 60 errors with a duration of 30 sec each while Guice tries to inject the instance of [MongoConnection](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/database/MongoConnectionImpl.java) into all the dependents such as [DBDataAdapterService](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/lookup/db/DBDataAdapterService.java) and so on. So it is about 60 * 30 = 1800 sec = 30 min in total until we see all the Guice errors. As far as I've learned recently, Guice doesn't have the ability to fail-fast. ~~So I've tweaked [MongoConnection](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/database/MongoConnectionImpl.java) to fail-fast if the first attempt is a fail. In the previous implementation we just had to wait 30 min until we realise that it's actually a fail. I've also made small changes to [MongoConnectionProvider](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoConnectionProvider.java) so [MongoConnection](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/database/MongoConnectionImpl.java) is a singleton at the level of Guice, not as a static field of the provider.~~ To summarise, I think it's better to fail-fast than wait quite some time until Guice has done its work or Mongo is up but the result is the same

~~I also want to add some unit tests to [MongoConnectionImpl](https://github.com/Graylog2/graylog2-server/blob/7b3717728b4ff8f2ec5d34808c43cf4d516724ce/graylog2-server/src/main/java/org/graylog2/database/MongoConnectionImpl.java) if you are fine with this kind of fix.~~ See my next comment

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoConnectionProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoConnectionProvider.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.bindings.providers;
 
+import com.mongodb.MongoException;
 import org.graylog2.configuration.MongoDbConfiguration;
 import org.graylog2.database.MongoConnection;
 import org.graylog2.database.MongoConnectionImpl;
@@ -28,23 +29,27 @@ import javax.inject.Provider;
 public class MongoConnectionProvider implements Provider<MongoConnection> {
     private static final Logger LOG = LoggerFactory.getLogger(MongoConnectionProvider.class);
     private static MongoConnection mongoConnection = null;
+    private static Exception mongoException;
 
     @Inject
     public MongoConnectionProvider(MongoDbConfiguration configuration) {
         if (mongoConnection == null) {
             try {
                 mongoConnection = new MongoConnectionImpl(configuration);
-
                 mongoConnection.connect();
             } catch (Exception e) {
                 LOG.error("Error connecting to MongoDB: {}", e.getMessage());
-                throw e;
+                mongoException = e;
             }
         }
     }
 
     @Override
     public MongoConnection get() {
-        return mongoConnection;
+        if (mongoException == null) {
+            return mongoConnection;
+        } else {
+            throw MongoException.fromThrowable(mongoException);
+        }
     }
 }


dkhamitov:
Can someone look at this PR?

kroepke:
Sorry for the delay, I've just now found time to look at this.

I believe the underlying reason that the provider gets called again and again is that its binding is configured incorrectly.
https://github.com/Graylog2/graylog2-server/blob/20ab7325eb82ec09c08ece0433e183412f32a05d/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java#L152 suggests that it is a singleton but it isn't bound in that scope, so the provider gets called for each injection site. That's the reason for the explosion of timeouts.

Essentially this patch makes it a singleton provider, but the correct (and much more concise way of doing so) is to simply change the binding:

```diff
diff --git a/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java b/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
index e21aa4f53..9eaa17f98 100644
--- a/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
+++ b/graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
@@ -36,8 +36,6 @@ import org.graylog2.bindings.providers.SystemJobFactoryProvider;
 import org.graylog2.bindings.providers.SystemJobManagerProvider;
 import org.graylog2.bundles.BundleService;
 import org.graylog2.cluster.ClusterConfigServiceImpl;
-import org.graylog2.rest.GenericErrorCsvWriter;
-import org.graylog2.users.UserPermissionsCleanupListener;
 import org.graylog2.dashboards.widgets.WidgetCacheTime;
 import org.graylog2.dashboards.widgets.WidgetEventsListener;
 import org.graylog2.database.MongoConnection;
@@ -62,6 +60,7 @@ import org.graylog2.plugin.inject.Graylog2Module;
 import org.graylog2.plugin.streams.DefaultStream;
 import org.graylog2.plugin.streams.Stream;
 import org.graylog2.rest.ElasticsearchExceptionMapper;
+import org.graylog2.rest.GenericErrorCsvWriter;
 import org.graylog2.rest.GraylogErrorPageGenerator;
 import org.graylog2.rest.NotFoundExceptionMapper;
 import org.graylog2.rest.QueryParsingExceptionMapper;
@@ -149,7 +148,7 @@ public class ServerBindings extends Graylog2Module {
     }
 
     private void bindSingletons() {
-        bind(MongoConnection.class).toProvider(MongoConnectionProvider.class);
+        bind(MongoConnection.class).toProvider(MongoConnectionProvider.class).in(Scopes.SINGLETON);
 
         if (configuration.isMessageJournalEnabled()) {
             install(new KafkaJournalModule());
```

Now, I still believe that this isn't the correct way of injecting something that can fail, but without a huge reorganization that's the best we can do.

Can you confirm that the above diff matches your expectations?

Thanks for your input!

dkhamitov:
I've done some testing and now I remember why it didn't work. Actually, making this provider a singleton was my first commit and I just struck through my first comment above related to singleton things :) https://github.com/dkhamitov/graylog2-server/commit/e6fb44f3d7d76a9f7c26c6ea66ee55eaf25e0a7d. The problem with just singleton is that Guice will create a new instance of Provider anyway if the first call to Provider#get results in error. It works as follows:
1. Once MongoConnectionProvider#get is called for the first time, it throws an error but static MongoConnectionProvider#mongoConnection is now not NULL
2. As the previous call has resulted in error, Guice instantiates a new instance of MongoConnectionProvider and calls get() again. But this call doesn't result in error as MongoConnectionProvider#mongoConnection is not NULL anymore (https://github.com/Graylog2/graylog2-server/blob/4e5f528682299bdd288000a80305e56f8c35828d/graylog2-server/src/main/java/org/graylog2/bindings/providers/MongoConnectionProvider.java#L34)
3. Now, we wait until Guice eventually fails to wire up all the beans. We will wait for the fail to happen until Mongo is up or 30 min otherwise (com.mongodb.MongoClientOptions.Builder#serverSelectionTimeout * 60 errors). Though, it can be fail-fast in case of initially unavailable Mongo if we just cache the first Mongo error as the PR suggests and avoid further long connection attempts.

> Now, I still believe that this isn't the correct way of injecting something that can fail, but without a huge reorganization that's the best we can do.

Agree, there are other edge cases in the current injecting approach which are still not covered

kroepke:
Hm ok, it worked for me as (at least I) expected, but I will take another look.

What I looked at additionally was to use a throwing provider, but to be useful this would require all call sites to catch the resulting exception.

I think we have two conflicting requirements:
1. We want to fail fast if MongoDB isn't there.
2. The provider (which essentially is proving a connection pool and not the actual connection) should not fail in its constructor, because that makes guice throw up. But changing this wouldn't make it fail fast anymore.

They are conflicting because if we changed the code to satisfy 2, we can't achieve 1. Even if we checked before, the connection can still fail during constructing 2.

I think the only non-hacky way of solving this is to create the connection before starting the injector and then injecting the instance directly. The underlying error here is to have the constructor do IO.

I'll have to think about this a bit :)

kroepke:
cleaning up old PRs we won't be getting to

PR #4903, opened 2018-07-11
Sidecar collector rename

mariussturm:
Related to https://github.com/Graylog2/collector-sidecar/pull/262
Fixes https://github.com/Graylog2/collector-sidecar/issues/255

## Description
After merging https://github.com/Graylog2/collector-sidecar/pull/262 the Sidecar will only report the Collector-ID in the corresponding status response. So we have to change the response model as well as the frontend to resolve the ID to an actual collector name on the status page.

### DIFF
--- a/graylog2-web-interface/src/components/sidecars/sidecars/SidecarStatus.jsx
+++ b/graylog2-web-interface/src/components/sidecars/sidecars/SidecarStatus.jsx
@@ -1,17 +1,24 @@
 import React from 'react';
 import createReactClass from 'create-react-class';
+import Reflux from 'reflux';
 import PropTypes from 'prop-types';
 import lodash from 'lodash';
 import { Col, Row } from 'react-bootstrap';
 
+import CombinedProvider from 'injection/CombinedProvider';
 import SidecarStatusFileList from './SidecarStatusFileList';
 
+const { CollectorsStore, CollectorsActions } = CombinedProvider.get('Collectors');
+
 const SidecarStatus = createReactClass({
+  mixins: [Reflux.connect(CollectorsStore)],
+
   propTypes: {
     sidecar: PropTypes.object.isRequired,
   },
 
   componentDidMount() {
+    this.loadCollectors();
     this.style.use();
   },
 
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/pages/SidecarStatusPage.jsx
```diff
@@ -38,8 +40,13 @@ class SidecarStatusPage extends React.Component {
     SidecarsActions.getSidecar(this.props.params.sidecarId).then(sidecar => this.setState({ sidecar }));
   };
 
+  reloadCollectors() {
```
edmundoa: As this is an ES6 class, this should use an arrow function, to ensure `this` is properly bound, even if the function calling it is not autobound by React.

From https://reactjs.org/docs/react-without-es6.html#autobinding:

> If you’d rather play it safe, you have a few options:
>  - Bind methods in the constructor.
>  - Use arrow functions, e.g. `onClick={(e) => this.handleClick(e)}`.
>  - Keep using `createReactClass`.

graylog2-web-interface/src/components/sidecars/sidecars/SidecarStatus.jsx
```diff
@@ -9,6 +9,7 @@ import SidecarStatusFileList from './SidecarStatusFileList';
 const SidecarStatus = createReactClass({
   propTypes: {
     sidecar: PropTypes.object.isRequired,
+    collectors: PropTypes.array.isRequired,
```
edmundoa: Going to a sidecar details page, I'm seeing the following in my JS console:

```
Warning: Failed prop type: The prop `collectors` is marked as required in `SidecarStatus`, but its value is `undefined`.
    in SidecarStatus (created by SidecarStatusPage)
    in SidecarStatusPage (created by ReactProxy)
```

graylog2-web-interface/src/pages/SidecarStatusPage.jsx
```diff
@@ -25,6 +26,7 @@ class SidecarStatusPage extends React.Component {
 
   componentDidMount() {
     this.reloadSidecar();
+    this.reloadCollectors();
```
edmundoa: Do we also need to automatically reload collectors, or is it fine to only load them on page load? I guess the second option is fine, but want to double-check :)

mariussturm: Loading them once is good enough.

mariussturm:
I think the `CollectorsAdministration` page is not affected because we don't use the `node_details.collectors` list. Instead we do it exactly like in this PR, call `CollectorsActions.all()` and from that result we can use `collector.name`.

PR #4913, opened 2018-07-16
Changes to test MessageList

kmerz:
## Description
Some changes necessary to test MessageList in enterprise plugin:
- use 'import' for css
- remove unused Store

## Motivation and Context
To test MessageList in enterprise-plugin, there a small changes necessary to
make it easier.

## Types of changes
- [x] Non-Breaking refactoring

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -5,6 +5,9 @@ import ReactSelect from 'react-select';
 import lodash from 'lodash';
 
 import AppConfig from 'util/AppConfig';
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/common/Select.jsx
```diff
@@ -135,8 +129,6 @@ const Select = createReactClass({
   },
 
   _select: undefined,
-  reactSelectStyles: require('!style/useable!css!react-select/dist/react-select.css'),
```
dennisoelkers: Instead of removing this and not using the (cleaner) way of `use()/unuse()`, we should check if it is possible to mock the `require` here in the test. Just returning a dummy object with the accessed fields/functions should be sufficient.

PR #4914, opened 2018-07-16
Add new StoreProviderMock and add custom Function to StoreMock

kmerz:
## Description
Add a new StoreProviderMock which can be used as a mock
for StoreProvider. It's a factory which takes a map of stores
with the store names as keys.

Also enhance StoreMock so it can add a custom function
in stead of jest.fn(). This is useful when the function
should return something specific or you want to add a
`expect` to the function outcome.

## Motivation and Context
StoreProviderMock is needed to make a meaningful mock of StoreProvider
(just like CombinedProviderMock)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/test/helpers/mocking/StoreProviderMock.js
+++ b/graylog2-web-interface/test/helpers/mocking/StoreProviderMock.js
@@ -1,0 +1,14 @@
+export default class StoreProviderMock {
+  constructor(stores = {}) {
+    this.stores = stores;
+  }
+
+  getStore(name) {
+    const result = this.stores[name] || {};
+    if (!result[`${name}Store`]) {
+      result[`${name}Store`] = {};
+    }
+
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/test/helpers/mocking/StoreProviderMock.js
```diff
@@ -0,0 +1,14 @@
+export default class StoreProviderMock {
+  constructor(stores = {}) {
+    this.stores = stores;
+  }
+
+  getStore(name) {
+    const result = this.stores[name] || {};
+    if (!result[`${name}Store`]) {
+      result[`${name}Store`] = {};
```
dennisoelkers: In order to make this consistent with `CombinedProviderMock`, it should return a configurable default store mock when there is no store registered for it.

PR #4962, opened 2018-07-30
ContentPack: use header for navigation and put actions on page

### LINKED ISSUE #4941: Content packs page header buttons change
## Expected Behavior

The buttons in the content packs page header shouldn't change depending on the page the user navigates to. This has been an issue for users in the past and we fixed it on the lookup tables pages, for example.

## Current Behavior

The page header buttons for content packs change when the user navigates to the details page of a content pack.

### Overview page

image

### Content pack details page

image

## Context

First: we have other pages that have the same "issue". :wink: I think it depends on what the buttons in the page header are seen as. Are they _navigation_ or _actions_? On the current content pack pages the buttons are _actions_. On the lookup tables pages, for example, the page header buttons are _navigation_.

What do we want to use on the content pack pages? TBD

## Possible Solution

If we decide to only have navigation buttons on the content pack pages, we should place an overview button in the header. The "Upload" and "Create a content pack" buttons on the overview page could be moved down into the page body.

Alternative placement for the _action_ buttons could be something like we do for lookup tables or inputs:

image

image

The edit button on the content pack details page should probably be moved to the "Version" table:

image

kmerz:
## Description
Rearrange Header Buttons and rename Edit to 'Edit clone'
 
## Motivation and Context
Buttons should be easy to find for a user.

screenshot_2018-08-08 graylog - content packs 1
screenshot_2018-08-08 graylog - content packs

Fixes #4941 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/content-packs/ContentPackVersions.jsx
+++ b/graylog2-web-interface/src/components/content-packs/ContentPackVersions.jsx
@@ -1,8 +1,10 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import Routes from 'routing/Routes';
 import { DataTable } from 'components/common';
-import { Button } from 'react-bootstrap';
+import { Button, ButtonToolbar } from 'react-bootstrap';
+import { LinkContainer } from 'react-router-bootstrap';
 import ContentPackDownloadControl from 'components/content-packs/ContentPackDownloadControl';
 
 import './ContentPackVersions.css';
@@ -51,9 +53,13 @@ class ContentPackVersions extends React.Component {
         </td>
         <td>{pack.rev}</td>
         <td className="text-right">
-          <Button bsStyle="primary" bsSize="small" onClick={() => { this.props.onDeletePack(pack.id, pack.rev); }}>Delete</Button>
-          &nbsp;
-          <Button bsStyle="info" bsSize="small" onClick={() => { downloadRef.open(); }}>Download</Button>
+          <ButtonToolbar className="pull-right">
+            <Button bsStyle="primary" bsSize="small" onClick={() => { this.props.onDeletePack(pack.id, pack.rev); }}>Delete</Button>
+            <Button bsStyle="info" bsSize="small" onClick={() => { downloadRef.open(); }}>Download</Button>
+            <LinkContainer to={Routes.SYSTEM.CONTENTPACKS.edit(encodeURIComponent(pack.id), encodeURIComponent(pack.rev))}>
+              <Button bsStyle="primary" bsSize="small">Edit</Button>
+            </LinkContainer>
+          </ButtonToolbar>
         </td>
         {downloadModal}
       </tr>
--- a/graylog2-web-interface/src/components/content-packs/ContentPackVersions.jsx
+++ b/graylog2-web-interface/src/components/content-packs/ContentPackVersions.jsx
@@ -3,7 +3,7 @@ import React from 'react';
 
 import Routes from 'routing/Routes';
 import { DataTable } from 'components/common';
-import { Button, ButtonToolbar } from 'react-bootstrap';
+import { Button, DropdownButton, ButtonToolbar, MenuItem } from 'react-bootstrap';
 import { LinkContainer } from 'react-router-bootstrap';
 import ContentPackDownloadControl from 'components/content-packs/ContentPackDownloadControl';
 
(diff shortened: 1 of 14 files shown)


edmundoa:
I was reviewing the changes in here and reading the underlying issue (#4941), and in my opinion the new position of buttons and navigation is in most cases worse than before.

From the original issue:
>The buttons in the content packs page header shouldn't change depending on the page the user navigates to. This has been an issue for users in the past and we fixed it on the lookup tables pages, for example.

This have been a problem indeed, but because on those cases every sub-page inside the section used different header buttons, or different ordering for those buttons. Having different header buttons for lists of entities and the entity itself is not a problem and sometimes is even helpful. In the case of content packs, there are some inconsistencies (as in most our sections), but it wasn't as confusing as in the other places we fixed it in previous versions.

I'll go through the different pages and give my opinion on the changes:

## Content packs page

Before:
content-packs-before

After:
content-packs-after

I have two problems with this:

- Putting actions next to a data filter is confusing and in my opinion we should avoid that unless those actions are related to the data filter itself. The reason for this is that it is not clear what will be affected by those actions, or when one should use them
- Having a filter and the pagination next to each other will create some design issues, as there are many elements together in a small place

I would personally use this header for this page:
screen shot 2018-08-02 at 13 26 30

Upload looks a bit funny in there, as it's not a page but a modal, but create and upload should go together in my opinion. The "Content packs" button should be marked as active, so it's colour is a bit darker.

Another solution that would also be fine is to move the filter input above the pagination controls, and then put the create and upload buttons above the page size selector, aligned to the right, something like:
screen shot 2018-08-02 at 13 36 36

## Create content pack page

Nothing changed on this page, but I would keep the button to "Create content pack" on the header, if it is still in the content packs page.

## Content pack details

Before:
content-pack-details-before

After:
content-pack-details-after

In this case, with this change we add a new "feature": editing old versions of a content pack. I think this may harm more than help, since changes since a certain version will be lost if that version is edited.

The styling also looks a bit funny to me with so many buttons in a small place, but maybe we can get away with it by using "actions" dropdowns to hide the three buttons in there, as those actions are not going to be used that often.

Other than that, I think "Edit" should remain in the header only if "Create content pack" remains in the header in the content packs page, keeping both pages consistent.

## Edit content pack

Nothing changed on this page, but I would keep the button to "Edit" on the header, if we decide to keep it in the content packs details page.

---

What do you think @kmerz and @bernd?

bernd:
@edmundoa Thank you very much for the feedback!

> This have been a problem indeed, but because on those cases every sub-page inside the section used different header buttons, or different ordering for those buttons. Having different header buttons for lists of entities and the entity itself is not a problem and sometimes is even helpful. In the case of content packs, there are some inconsistencies (as in most our sections), but it wasn't as confusing as in the other places we fixed it in previous versions.

Fair enough! :smile: 

I like you suggestion for the content packs page.

@kmerz What do you think?

> In this case, with this change we add a new "feature": editing old versions of a content pack. I think this may harm more than help, since changes since a certain version will be lost if that version is edited.

I agree that the possibility to edit old content packs is probably confusing for the user. Editing is not possible, though. "Editing" a content pack will create a new revision when it's saved.

Heh, naming it "Edit" is probably confusing to because it's not an edit but a clone + edit.

@kmerz What do you think?

kmerz:
Thanks for the input @edmundoa and thank @bernd for the response.

I agree with the content packs page and already implemented that now.

I also agree with renaming the `Edit` button. How about `Clone for edit` or just `Edit clone`?

bernd:
> I also agree with renaming the Edit button. How about `Clone for edit` or just `Edit clone`?

Hmm, how about `Create new revision` or something similar? Cloning or copying content packs sounds like creating a completely new one. But "editing" it just creates a new revision of an existing content pack.

edmundoa:
I like the changes, I think the navigation is now more consistent.

One thing I'm still missing is a way to install a content pack from the content pack details page. Should we add an "Install" option to each version? Or maybe add an "Install" button in the header?

Regarding the "Edit" option, I'm not convinced about naming it "Edit clone". So far what I think would describe the functionality better is "Edit from revision".

Another thing I saw and I didn't find a better place to write it (since editing any revision is a new functionality). Editing an old revision does not work at the moment, since we don't use the "latest" version number, but the version number following the picked revision. As an example, editing version `2` of a content pack, will always try to save the next version as number `3`, even if that version already exists.

bernd:
> One thing I'm still missing is a way to install a content pack from the content pack details page. Should we add an "Install" option to each version? Or maybe add an "Install" button in the header?

@edmundoa I think it makes sense to have an "install" button for each version so the user can decide to install a different version if a new version is broken, for example.

kmerz:
> @edmundoa I think it makes sense to have an "install" button for each version so the user can decide to install a different version if a new version is broken, for example.

I like it. Will do so!

graylog2-web-interface/src/components/content-packs/ContentPackVersions.jsx
```diff
@@ -51,9 +100,17 @@ class ContentPackVersions extends React.Component {
...
-          &nbsp;
-          <Button bsStyle="info" bsSize="small" onClick={() => { downloadRef.open(); }}>Download</Button>
+          <ButtonToolbar className="pull-right">
+            <Button bsStyle="success" bsSize="small" onClick={() => { downloadRef.open(); }}>Download</Button>
+            <DropdownButton id={`action-${pack.rev}`} bsStyle="info" title="Actions" bsSize="small">
+              <LinkContainer to={Routes.SYSTEM.CONTENTPACKS.edit(encodeURIComponent(pack.id), encodeURIComponent(pack.rev))}>
+                <MenuItem >Edit clone</MenuItem>
+              </LinkContainer>
+              <MenuItem onClick={() => { this.props.onDeletePack(pack.id, pack.rev); }}>Delete</MenuItem>
+              <MenuItem onClick={openFunc}>Install</MenuItem>
```
edmundoa: I know it's a small thing, but please avoid putting delete in the middle of the dropdown, as it is much easier to accidentally click it. As far as I know delete goes always last in our dropdowns, and most of the time has a separator before it. I would order them this way:

```
Install
Edit clone – I'm still not convinced by the name, but that's another topic ;)
---
Delete
```

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -95,9 +100,6 @@ const ShowContentPackPage = createReactClass({
               <LinkContainer to={Routes.SYSTEM.CONTENTPACKS.LIST}>
                 <Button bsStyle="info">Content Packs</Button>
               </LinkContainer>
-              <LinkContainer to={Routes.SYSTEM.CONTENTPACKS.edit(encodeURIComponent(lastPack.id), encodeURIComponent(lastPack.rev))}>
```
edmundoa: I think something went wrong during the merge conflict, `lastPack` is not used any more, so we can remove the variable.
We should also remove `lastVersion`, as it is only used to get `lastPack` and the function it references is removed in this change.

PR #4987, opened 2018-08-09
Add custom request header to prevent CSRF

edmundoa:
Improve our protection against CSRF by requiring a custom request header (`X-Requested-By`) in all non-GET requests sent to our API. This is mentioned as a way of CSRF prevention [1] and it is particularly suitable for REST APIs, since let requests to remain stateless.

1: https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Protecting_REST_Services:_Use_of_Custom_Request_Headers

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/RemoteInterfaceProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/RemoteInterfaceProvider.java
@@ -21,6 +21,7 @@ import com.google.common.net.HttpHeaders;
 import com.google.common.net.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
+import org.glassfish.jersey.client.filter.CsrfProtectionFilter;
 import org.graylog2.cluster.Node;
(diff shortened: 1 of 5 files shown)


kmerz:
I only add this for documentation purpos, because I got a better understanding of that fix by reading this:

https://seclab.stanford.edu/websec/csrf/csrf.pdf

> Custom HTTP headers can be used to prevent CSRF because
> the browser prevents sites from sending custom HTTP
> headers to another site but allows sites to send custom HTTP
> headers to themselves using XMLHttpRequest. For example,
> the prototype.js JavaScript library [45] uses this approach
> and attaches the X-Requested-By header with the
> value XMLHttpRequest. Google Web Toolkit also recommends
> [16] that web developers defend against CSRF attacks
> by attaching a X-XSRF-Cookie header to XMLHttpRequets
> that contains a cookie value. The cookie value is not actually
> required to prevent CSRF attacks: the mere presence
> of the header is sufficient.

bernd:
I will test this today.

PR #4990, opened 2018-08-10
Add entityId to NativeEntityDescriptor

kmerz:
## Description
Add the origin entity id to the installation model. 

## Motivation and Context
With that the installed entity has a relation with the entity from which it was created from the
content pack.

## How Has This Been Tested?
Installed a new content pack and took a look in the database.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/model/entities/NativeEntityDescriptor.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/model/entities/NativeEntityDescriptor.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.contentpacks.model.entities;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
 import org.graylog2.contentpacks.model.Identified;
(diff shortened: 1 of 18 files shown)


graylog2-server/src/main/java/org/graylog2/contentpacks/model/entities/NativeEntityDescriptor.java
```diff
@@ -29,26 +30,44 @@
...
     }
 
     /**
-     * Shortcut for {@link #create(ModelId, ModelType)}
+     * Shortcut for {@link #create(String, String, ModelType)}
      */
-    public static NativeEntityDescriptor create(String id, ModelType type) {
-        return create(ModelId.of(id), type);
+    public static NativeEntityDescriptor create(String entityId, String nativeId, ModelType type) {
+        return create(ModelId.of(entityId), ModelId.of(nativeId), type);
```
bernd: How about creating a `NativeEntityDescriptor.create(ModelId entityId, ...)` method to avoid the `ModelId.of(entityId)` call? (that also removes some `entity.id().toString()` calls.

graylog2-server/src/test/resources/org/graylog2/contentpacks/ContentPackInstallationPersistenceServiceTest.json
```diff
@@ -20,14 +20,17 @@
         {
           "id": "5b4c920b4b900a0024af2b5d",
           "type": "collector"
+          "entity_id": "5b1a49eb3d274631fe07c86a"
```
bernd: There is a missing `,` after the `type` line. (same for the next two entries)

PR #5005, opened 2018-08-19
Show warning if configured journal size does not fit on disk

### LINKED ISSUE #4933: Show warning if configured journal size does not fit on disk
## Expected Behavior
Graylog should warn if the journal would not fit at the configured location. 

## Current Behavior
You can configure any size at any location for the message journal and Graylog will accept this. That could potential damage the setup if the space is not enough for the configured journal size - messages will get lost when the journal is damaged.

## Possible Solution

- Graylog could warn on startup if the configured journal size will not fit into the configured disk (in the logfile)
- Graylog could warn at the nodes overview if the journal holding partition gets filled not leaving enough space for the journal.
- Graylog could give a notification like for journal usage if the partition would not fit.

## Context

Most users did not notice the setting of the journal size until they run into issue - but then it becomes really fast really messy. Users could have created multiple small partitions where /var is not big enough to hold the journal and the logfiles at once.

## Your Environment

* Graylog Version: 2.4.6

radykal-com:
## Description
Logs and shows a notification if the journal file can't fit in the partition (partition too small, or other files filling up the disk)

## Motivation and Context
Fixes #4933 

## How Has This Been Tested?
Configuring an small partition for the journal file and setting the journal file to almost the full size of the partition. Then moving some other files to the partition (making imposible to reach the journal max file without getting out of space) will trigger the notification.
If restarting graylog with these changes, it will also log a warning during start.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/notifications/Notification.java
+++ b/graylog2-server/src/main/java/org/graylog2/notifications/Notification.java
@@ -62,7 +62,8 @@ public interface Notification extends Persisted {
         JOURNAL_UNCOMMITTED_MESSAGES_DELETED,
         OUTPUT_DISABLED,
         INDEX_RANGES_RECALCULATION,
-        GENERIC
+        GENERIC,
+        JOURNAL_WONT_FIT_IN_PARTITION
     }
 
     enum Severity {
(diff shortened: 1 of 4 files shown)


radykal-com:
@jalogisch any update on this review?

Issue #4958, opened 2018-07-27
Implement reindex facility to migrate indices of older Elasticsearch versions.


kroepke:
Because versions created prior to Elasticsearch 5 need to be reindexed before a 6.x cluster can start, Graylog needs to be able to identify and reindex indices as needed.

Reindexing should make sure the process has completely finished before allowing to remove old indices.

PR #5064, opened 2018-09-05
Add Jest actions for use with ES Task API

kroepke:
In preparation to expose reindex capabilities, we need to be able to access the status of the long running tasks (to show progress) and also be able to cancel them.

related to #4958

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/tasks/CancelTask.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/tasks/CancelTask.java
@@ -14,28 +14,15 @@
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
-/**
- * This file is part of Graylog.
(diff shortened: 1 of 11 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/cluster/jest/tasks/CancelTask.java
```diff
@@ -0,0 +1,82 @@
...
+ * General Public License as published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Graylog.  If not,
+ * see <http://www.gnu.org/licenses/>.
+ */
```
bernd: Small nitpick: duplicate licenses

PR #5067, opened 2018-09-05
Fix editing of content packs

kmerz:
## Description
   * Fix revision bump when editing a new content pack revision
   * Fix selection of entities from former content pack revision

## Motivation and Context
   * When a content pack was created from revision `1` but there was already a revision `2`, the create would fail, since the new content pack would have a content pack `1`,
   * When editing a content pack the entities were not considering the new entity type format 

## How Has This Been Tested?
Edit a content pack

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/EditContentPackPage.jsx
+++ b/graylog2-web-interface/src/pages/EditContentPackPage.jsx
@@ -35,11 +35,16 @@ const EditContentPackPage = createReactClass({
   },
 
   componentDidMount() {
-    ContentPacksActions.getRev(this.props.params.contentPackId, this.props.params.contentPackRev).then(() => {
+    ContentPacksActions.get(this.props.params.contentPackId).then(() => {
+      const originContentPackRev =  this.props.params.contentPackRev;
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/pages/EditContentPackPage.jsx
```diff
@@ -35,11 +35,16 @@ const EditContentPackPage = createReactClass({
   },
 
   componentDidMount() {
-    ContentPacksActions.getRev(this.props.params.contentPackId, this.props.params.contentPackRev).then(() => {
+    ContentPacksActions.get(this.props.params.contentPackId).then(() => {
+      const originContentPackRev = this.props.params.contentPackRev;
+      const newContentPack = this.state.contentPack[originContentPackRev];
+      const nextContentPackRev = Math.max(...Object.keys(this.state.contentPack).map(x => parseInt(x, 10))) + 1;
```
bernd: Bumping the revision in the UI can go wrong when two people clone the same content pack. But since we have unique indices in place in the database, the worst thing that can happen is that the user gets an ugly error. :wink:

PR #5073, opened 2018-09-05
Allow kafka group.id to be configured

pdepaepe:
Signed-off-by: Pierre De Paepe <pierre.de-paepe@corp.ovh.com>

## Description

This little patch will allow graylog administrators to define a custom Kafka group_id in related input panel.
To avoid update issue, current group "graylog2" remains the default value.

## Motivation and Context

If you have a big Kafka cluster with multiple graylog clusters subscribed on, each time a unhealthy consumer flaps, the rebalancing order is sent to all graylog nodes introducing unecessary stress on log ingestion.
An easy way to avoid it, consists to define for each graylog cluster a specific consumer group name.

## How Has This Been Tested?
In production : ~20 graylog clusters for one kafka.

## Screenshots (if appropriate):

group id

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x ] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #5001, opened 2018-08-17
kafka input maximum fetch size


caiwenhao:
## Expected Behavior

```
2018-08-16 23:33:06,033 ERROR: org.graylog2.inputs.transports.KafkaTransport - Kafka consumer error, stopping consumer thread.
kafka.common.MessageSizeTooLargeException: Found a message larger than the maximum fetch size of this consumer on topic online_lxyd partition 0 at fetch offset 55010910. Increase the fetch size, or decrease the maximum message size the broker will allow.
```
## Context
a message larger than the maximum fetch size, I can't find this configuration to Increase the fetch size.

## Your Environment
* Graylog Version: 2.4.6-1

Issue #4481, opened 2018-01-15
Setting fetch.message.max.bytes for Kafka Inputs


edsonmarquezani:
(This may be a feature request.)

How can I set the `fetch.message.max.bytes` property for Kafka consumers on Graylog?

I ran into an issue today with large messages on my topics, which caused my input to freeze. The only way to work around this was forcing the consumer to skip the messages, increasing offset values directly on zookeeper.

Is there a way to set custom Kafka client options like this?

Issue #3960, opened 2017-07-01
SSL/TLS for Kafka inputs


ghost:
Apparently it is not possible to use SSL/TLS for Kafka inputs in latest version or there is not documentation for it.

## Expected Behavior
Kafka inputs should allow SSL for accessing to server. 

## Current Behavior
There is no option related to SSL/TLS in Kafka inputs

## Possible Solution
No Idea :)

## Context
This feature is important when you want to transfer log data across the globe and don't want unwanted access from wiretappers. 

## Your Environment
We have multiple sites in different locations, each site stores all data in a Kafka server and one graylog server connects to each site and reads log data.

* Graylog Version: Graylog 2.2.3

tokred:
Would be highly appreciated to fulfill regulatory and compliance requirements!

Best regards

k1-hedayati:
@tokred sorry I didn't understand what you mean

joschi:
@k1-hedayati I think @tokred simply wants to express some approval of this feature request.

k1-hedayati:
So I guess this doesn't evolve me, ok :)

madiTG:
+1

pbr0ck3r:
@joschi I have worked with a few other devs on upgrading Kafka to 1.0.1 and have it successfully working. This allows users to enable/disable SSL/TLS. As I am looking to make a PR to master branch for this so you all can enjoy the benefits as well. Is backwards capability a requirement? Or is it more of a `it would be nice to have`. From version 9 to 1.01 there was a lot of deprecated API calls, so adding backwards capability may take a lot more work.

joschi:
@pbr0ck3r Unfortunately it's not that easy since Graylog is using Kafka's journal implementation for its own message journal (see [`KafkaJournal`](https://github.com/Graylog2/graylog2-server/blob/2.4.3/graylog2-server/src/main/java/org/graylog2/shared/journal/KafkaJournal.java)) and Kafka 0.10.x and later broke backwards-compatibility with it.

This being said, keeping Kafka 0.9.x in Graylog and adding a new Kafka transport using the newer Kafka Java client libraries *could* work.

pbr0ck3r:
@joschi I have kafka 1.0.1 working and have modified the KafkaJournal. If I submitted this code in a PR would it be most likely accepted for 3.0?

pbr0ck3r:
[...]

PR #5088, opened 2018-09-07
Load consumer properties from file

jsargiot:
Load custom Kafka Consumer properties from a file

## Description

This change allows to define a custom properties file that will be loaded and overwrites the default configuration of the KafkaTransport while also allowing to set other Consumer properties (like TLS configuration certs and passwords).

The order of the properties also changes to give default values the lowest precedence, then the properties from the consumer properties file are loaded to override default values, and later input configuration to override consumer properties. Consumer `client.id` is also moved to the latest stage to avoid being overrided.

Empty values for the consumer properties file are ignored.

Failing to read the consumer properties file stops the KafkaTransport in the assumption that missing properties could affect the connection in unexpected ways.

> Note: I'm creating the PR against branch 2.5, please let me know if it shouldn't, I was initially going for 2.4 but I guess that ship has sailed (happy to move the PR to 2.4 if you say so).

## Motivation and Context

The provided fields in the Kafka Input configuration do not cover TLS configuration, we could add those, but then there are other configurations that people need to tweak depending on their environment needs, this is a never ending story of adding fields everytime there's someone with a specific need, I on the other hand, propose to load Kafka Consumer configuration from a config file, just like many other tools (for example kafka-console-consumer.sh with `--consumer.config` argument), this way we allow for an configuration of the Kafka Consumer (https://kafka.apache.org/documentation/#newconsumerconfigs) without modifying Graylog.

There are a number of issues that could be solved by this approach:

https://github.com/Graylog2/graylog2-server/issues/5001: By setting `fetch.message.max.bytes` in the consumer.properties the Consumer would be able to fetch larger messages.

https://github.com/Graylog2/graylog2-server/issues/4481: Same as the previous.

https://github.com/Graylog2/graylog2-server/issues/3960: Setting `ssl.*` family for configuration would automatically add SSL support to the client (once new client is merged, old client does not support SSL).

https://github.com/Graylog2/graylog2-server/pull/5073: Because of the order in which properties are loaded, we're allowing to modify the `group.id` from the consumer.properties file. So PR #5073 would be a nice addition but we can achieve the same without touching the UI.

https://github.com/Graylog2/graylog2-server/pull/4770: The use of a consumer properties file would serve as base for this PR. The addition of SSL would have to be removed (in my opinion), because there are several `security.protocol`s not just `SSL` and having options for each is just mad.

## How Has This Been Tested?

Created Travis build to ensure the build is ok:
    https://travis-ci.org/jsargiot/graylog2-server/builds/427290269

Tested locally-built package against:
    Kafka 0.10
    Kafka 1.1.1

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
@@ -58,6 +58,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Named;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -79,6 +82,7 @@ public class KafkaTransport extends ThrottleableTransport {
     public static final String CK_TOPIC_FILTER = "topic_filter";
     public static final String CK_THREADS = "threads";
     public static final String CK_OFFSET_RESET = "offset_reset";
+    public static final String CK_CONSUMER_PROPERTIES = "consumer_properties";
 
     // See https://kafka.apache.org/090/documentation.html for available values for "auto.offset.reset".
     private static final Map<String, String> OFFSET_RESET_VALUES = ImmutableMap.of(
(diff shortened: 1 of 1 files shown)


mpfz0r:
this got implemented in #7504

PR #5104, opened 2018-09-13
Adapt default winlogbeat paths

edmundoa:
This PR adapts the initial Winlogbeat collector paths to match the default sidecar install paths in Windows.

As we didn't release this code yet, it should be fine modifying the original migration instead of doing another one on top of it.

Fixes https://github.com/Graylog2/collector-sidecar/issues/289

---
Edit: I extended the PR to also modify the default paths for Filebeat and NXLog, as requested. After some discussion, we also moved the default configurations to `/var/lib` by default.


bernd:
Sorry for hijacking this PR, but can you also fix the default path to filebeat? We changed the installation location to `/usr/lib/graylog-sidecar/filebeat` recently and that needs to be reflected in the default config.

https://github.com/Graylog2/graylog2-server/blob/12b9db93238976520a4f3edd090af7da61c95a8d/graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java#L50

Thank you! :smiley:

PR #5121, opened 2018-09-17
Using loadable-components to split out pages and brace.

dennisoelkers:
This change is employing `loadable-components` to replace `react-proxy-loader` because of the two shortcomings it has:

  * It does not work with React 16+, effectively blocking the update
  * It requires a static webpack rule, making it match imports against a
file pattern which makes it hard to use/unusable for plugins

`loadable-components` is remedying both, as it relies on the generic dynamic import function, which triggers webpack to split out chunks.

To enable this, the following steps have been performed:

 * Removing react-proxy-loader.
 * Updating yarn.lock
 * Create wrapper function to load component asynchronously.
 * Loading new sidecar/content pack pages async.
 * Updating loadable-components and adding dynamic-import-node plugin for jest.
 * Removing deprecated setConfig call.
 * Updating yarn.lock.
 * Including simple error component.
 * Loading stores in tests again
 * Using CJS syntax in export.

Additionally, the following bigger changes were required:

**Separate stores into own chunk, loaded before plugin/app bundles (e317761).**

Before this change, any (explicit or implicit) consumer (including plugins!) of the `CombinedProvider` module inevitably included all actions/stores in core.

With this change, the providers are separated from the inclusion of actions/stores. Actions/stores are now in a separate 'builtins' bundle which also registers these in the `CombinedProvider` module upon entry. This leads to slimmer plugin bundles, as they reference, but not include stores from core. It also prevents plugins from including outdated stores from core.

**Hashing module ids instead of assigning incremented integers (2de3428).**

Before this change, modules were assigned incremented integers, leading to duplication in plugins or other assets which are built in separate compilations.

After this change, the module id is created using a hash, decreasing the risk of collisions by magnitudes.

**Adding webpack plugin generating unique chunk ids (fe38cbf).**

Before this change, chunk ids in webpack were just integers starting with 0, incremented by 1 for every chunk. Due to their simple nature, collisions were unavoidable when building assets in multiple
independent compilations. This lead to failing chunk loading for plugins loading chunks asynchronously using `import()`.

This change adds a simple webpack hook that generates a random build id that is appended to each chunk id, making chunk id collisions way less probable.


graylog2-server/src/main/java/org/graylog2/web/PluginAssets.java
```diff
@@ -115,6 +117,14 @@ public PluginAssets(Set<Plugin> plugins,
                     if (file2.equals(polyfillJsFile)) {
                         return 1;
                     }
+                    // Builtins JS script goes second
```
kmerz: `Builtins JS script goes third`?

dennisoelkers: :+1:

kmerz:
LGTM! :+1: Nice :package:ing

PR #5130, opened 2018-09-18
Fix content pack serialisation

### LINKED ISSUE #5085: *AutoValue_ContentPackV1* in ContentPack version
## Expected Behavior
A downloaded and newly created content pack should be up-loadable.

## Current Behavior
A downloaded content pack has property values like `AutoValue_ContentPackV1` which can't
be recognized by the backend when loading up.

1. Create new content pack
2. Download content pack
3. Delete content pack (to prevent rev already exists error)
4. Upload content pack

kmerz:
## Description
A downloaded content pack had *AutoValue_ContentPackV1* as version
serialized in the json. That lead to errors when the content pack
was downloaded and later uploaded again.

The problem was that `JsonTypeInfo.as` was using `PROPERTY`
which would read the version to determine the version of an
content pack, but also write the class name into the version.

https://fasterxml.github.io/jackson-annotations/javadoc/2.3.0/com/fasterxml/jackson/annotation/JsonTypeInfo.As.html
We now use `EXISTING_PROPERTY` so `JsonTypeInfo.as` will only
read this value but not write it itself.

Fixes #5085 

## Types of changes

## Checklist:


dennisoelkers:
Can you add tests for these cases to make sure serialization/deserialization is working?

graylog2-server/src/test/java/org/graylog2/contentpacks/model/ContentPackTest.java
```diff
@@ -115,6 +118,50 @@ public void serializeContentPackV1() {
...
+                        LongParameter.builder().name("MY_LONG").title("My Long").description("Some description").defaultValue(Optional.of(42L)).build(),
+                        StringParameter.builder().name("MY_STRING").title("My String").description("Some description").defaultValue(Optional.of("Default Value")).build()))
+                .entities(ImmutableSet.of(
+                        EntityV1.builder().id(ModelId.of("fafd32d1-7f71-41a8-89f5-53c9b307d4d5")).type(ModelTypes.INPUT_V1).version(ModelVersion.of("1")).data(entityData).build()))
+                .build();
+
+        final String jsonTxt = objectMapper.writeValueAsString(contentPack);
+        assertThat(jsonTxt).doesNotContain("AutoValue_");
+
+        final URL contentPackURL = ContentPackTest.class.getResource("expected_content_pack.json");
```
dennisoelkers: This test block should be split out into a separate test  method named like "testSerializationFormatConsistency" or something of that kind. The reason is that if e.g. jackson changes serialization semantics (or we replace the JSON serialization library) so the test fails, it is not related to AutoValue and we prevent a red herring by containing it in an isolated test case.

We can extract and reuse a method to generate the `ContentPack` instance though.

graylog2-server/src/test/java/org/graylog2/contentpacks/model/ContentPackTest.java
```diff
@@ -115,6 +118,50 @@ public void serializeContentPackV1() {
...
+        Path path = Paths.get(contentPackURL.toURI());
+        String expectedJSON = new String(Files.readAllBytes(path)).replace("\n", "").replace("\r", "");
+
+        assertThat(jsonTxt).isEqualTo(expectedJSON);
+
+        final ContentPack readContentPack = objectMapper.readValue(jsonTxt, ContentPack.class);
```
dennisoelkers: Same here. I think testing deserialization is different from making sure that there are not `AutoValue_` prefixes.

PR #5154, opened 2018-09-25
Implement Error Boundary for Navigation Bar children.

dennisoelkers:
## Motivation and Context

    This change implements a error boundary component wrapping children of
    the navigation bar in the `App` component.

    Before this change, any errors in children components of the navigation
    bar caused the application to malfunction, errors being shown on the
    js console only and (for React 16) the DOM to be unmounted, requiring
    the user to do a full page refresh to be able to work with the
    application again.

    After this change, errors occuring in components which are children of
    the navigation bar leave the application in a functional state, while
    showing the error message, stack trace and component tree in the
    browser. Users are still able to navigate the application afterwards.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/ErrorPage.jsx
+++ b/graylog2-web-interface/src/pages/ErrorPage.jsx
@@ -29,34 +29,36 @@ class ErrorPage extends React.Component {
     const { error, info } = this.props;
 
     return (
-      <DocumentTitle title="Something went wrong.">
-        <Row className="jumbotron-container">
-          <Col mdOffset={2} md={8}>
-            <Jumbotron>
-              <h1>Something went wrong.</h1>
-              <p>It seems like the page you navigated to contained an error.</p>
-              <p>You can use the navigation to reach other parts of the product, refresh the page or submit an error report.</p>
-              <p>The details of the error are below:</p>
-              <dl className={errorPageStyles.errorMessage}>
-                <dt>Error:</dt>
-                <dd>
-                  <pre>{error.message}</pre>
-                </dd>
+      <div className="container-fluid">
+        <DocumentTitle title="Something went wrong.">
+          <Row className="jumbotron-container">
+            <Col mdOffset={2} md={8}>
+              <Jumbotron>
+                <h1>Something went wrong.</h1>
+                <p>It seems like the page you navigated to contained an error.</p>
+                <p>You can use the navigation to reach other parts of the product, refresh the page or submit an error report.</p>
+                <p>The details of the error are below:</p>
+                <dl className={errorPageStyles.errorMessage}>
+                  <dt>Error:</dt>
+                  <dd>
+                    <pre>{error.message}</pre>
+                  </dd>
 
-                <dt>Stack:</dt>
-                <dd>
-                  <pre>{error.stack}</pre>
-                </dd>
+                  <dt>Stack:</dt>
+                  <dd>
+                    <pre>{error.stack}</pre>
+                  </dd>
 
-                <dt>Component Tree:</dt>
-                <dd>
-                  <pre>{info.componentStack}</pre>
-                </dd>
-              </dl>
-            </Jumbotron>
-          </Col>
-        </Row>
-      </DocumentTitle>
+                  <dt>Component Tree:</dt>
+                  <dd>
+                    <pre>{info.componentStack}</pre>
+                  </dd>
+                </dl>
+              </Jumbotron>
+            </Col>
+          </Row>
+        </DocumentTitle>
+      </div>
     );
   }
 }
--- a/graylog2-web-interface/src/pages/ErrorPage.jsx
+++ b/graylog2-web-interface/src/pages/ErrorPage.jsx
@@ -60,7 +60,7 @@ class ErrorPage extends React.Component {
                     <dt>
                       Error:
                       <div className={`pull-right ${errorPageStyles.toggleDetails}`}>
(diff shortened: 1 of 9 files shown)


graylog2-web-interface/src/pages/ErrorPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  componentWillUnmount() {
+    style.unuse();
+  }
+
+  render() {
+    const { error, info } = this.props;
+
+    return (
+      <DocumentTitle title="Not Found">
```
kmerz: I would write "Something went wrong" here as well. Since something went wrong and not something was not found.

graylog2-web-interface/src/pages/ErrorPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+        <Row className="jumbotron-container">
+          <Col mdOffset={2} md={8}>
+            <Jumbotron>
+              <h1>Something went wrong.</h1>
+              <p>It seems like the page you navigated to contained an error.</p>
+              <p>You can use the navigation to reach other parts of the product, refresh the page or submit an error report.</p>
```
kmerz: Just a suggestion: Would here not be a good place for the issue page on github?
It could be a bit crowded, but than we could also ask to the user to add the stacktrace to the issue.

graylog2-web-interface/src/pages/ErrorPage.jsx
dennisoelkers: True!

graylog2-web-interface/src/pages/ErrorPage.jsx
dennisoelkers: I was also thinking about this, but I was hesitating because the error could also be the result of a 3rd party plugin that is faulty.

kmerz: That is true, but don't we not still know if one of our used libraries make graylog unusable?

dennisoelkers: Sure. But we cannot differentiate where it happened and we might produce a lot of issues in the wrong repository.

Nevertheless, my suggestion is to include the same component that we show when a search term returns no results/fails:

screen shot 2018-09-26 at 11 45 42

This way we offer the user some options to proceed with the error and not just "open an issue here".

edmundoa: I think adding such a component would make sense. I would also display more clearly the Graylog version being used, and by default collapse the stack trace and component tree, as they may be too prominent and scary, specially if users are not used to debugging computer applications.

Additionally I think we should also consider making it easier to submit this information to us by adding a "copy to clipboard" button to get all the information (maybe even formatted as markdown).

graylog2-web-interface/src/pages/ErrorPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+  componentWillUnmount() {
+    style.unuse();
+  }
+
+  render() {
+    const { error, info } = this.props;
+
+    return (
+      <DocumentTitle title="Something went wrong.">
+        <Row className="jumbotron-container">
```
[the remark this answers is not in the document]
dennisoelkers: ✅

dennisoelkers:
- I have made the error details collapsible now. They are collapsed per default on production and expanded by default for development. 
- A "Copy to clipboard" button now exists to the right of the error message/details which allows the user to copy the error message + details to clipboard
- The error page now shows our common measures to get support

Thanks for the valuable input on this one, guys!

graylog2-web-interface/src/pages/ErrorPage.jsx
```diff
@@ -0,0 +1,89 @@
...
+                <p>You can use the navigation to reach other parts of the product, refresh the page or submit an error report.</p>
+                <div className={errorPageStyles.errorMessage}>
+                  <div className="content" style={{ padding: '2em' }}>
+                    <SupportSources />
+                  </div>
+                  <dl>
+                    <dt>
+                      Error:
+                      <div className={`pull-right ${errorPageStyles.toggleDetails}`}>
+                        <a role="link" tabIndex={0} onClick={this._toggleDetails}>{showDetails ? 'Show less' : 'Show more'}</a>
```
edmundoa: Links without an `href` do not render the `pointer` cursor, so users have no feedback if that is a text or a link that can be clicked. We should either add some CSS to do that or to add an `href` attribute to the link.

graylog2-web-interface/src/routing/__snapshots__/AppErrorBoundary.test.jsx.snap
```diff
@@ -0,0 +1,134 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
edmundoa: It looks like the snapshot needs some updating after the last changes.

graylog2-web-interface/src/pages/ErrorPage.jsx
dennisoelkers: ✅

graylog2-web-interface/src/routing/__snapshots__/AppErrorBoundary.test.jsx.snap
dennisoelkers: ✅

edmundoa:
Looks good to me, so let's wait to hear from @kmerz :)

dennisoelkers:
Thanks for the thorough review, @edmundoa!

PR #5160, opened 2018-09-27
Enforce newlines at the end of rendered configurations

mpfz0r:
If a user entered a Sidecar configuration without an empty last line,
we would render the last configuration line without a newline terminator.
This might confuse some config parsers, looks weird when using `cat` and
is against my personal taste ;-)

Also add a small unit test for the ConfigurationService.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ConfigurationService.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ConfigurationService.java
@@ -194,9 +194,6 @@ public class ConfigurationService extends PaginatedDbService<Configuration> {
         }
 
         template = writer.toString();
-        if (!template.substring(template.length() -1).equals("\n")) {
-            return template + "\n";
-        }
-        return template;
+        return template.endsWith("\n") ? template : template + "\n";
     }
 }
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ConfigurationService.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ConfigurationService.java
@@ -181,6 +181,7 @@ public class ConfigurationService extends PaginatedDbService<Configuration> {
 
     private String renderTemplate(String configurationId, Map<String, Object> context) throws RenderTemplateException {
         Writer writer = new StringWriter();
+        String template;
         try {
             Template compiledTemplate = templateConfiguration.getTemplate(configurationId);
             compiledTemplate.process(context, writer);
@@ -192,6 +193,10 @@ public class ConfigurationService extends PaginatedDbService<Configuration> {
             throw new RenderTemplateException(e.getMessage(), e);
         }
 
-        return writer.toString();
+        template = writer.toString();
+        if (!template.substring(template.length() -1).equals("\n")) {
+            return template + "\n";
+        }
+        return template;
     }
 }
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog/plugins/sidecar/collectors/ConfigurationServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.List;
+
+import static com.lordofthejars.nosqlunit.mongodb.InMemoryMongoDb.InMemoryMongoRuleBuilder.newInMemoryMongoDbRule;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(JukitoRunner.class)
```
edmundoa: Small nitpicking:

We (almost) only use Jukito in Sidecar tests. I think the only advantage we get is to inject the `ObjectMapperModule` we use in `setUp()` (we also inject a `ValidatorModule` that is never used. I vote for doing like most of our tests and instead of using `JukitoRunner`, create a local instance of `MongoJackObjectMapperProvider` in `setUp()`. Here is an example of that: https://github.com/Graylog2/graylog2-server/blob/269bd12111609d4acad8fcc24c16e7ffb25cf1f5/graylog2-server/src/test/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceServiceTest.java#L54

graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ConfigurationService.java
```diff
@@ -192,6 +193,10 @@ private String renderTemplate(String configurationId, Map<String, Object> contex
             throw new RenderTemplateException(e.getMessage(), e);
         }
 
-        return writer.toString();
+        template = writer.toString();
+        if (!template.substring(template.length() -1).equals("\n")) {
```
edmundoa: I think we can simplify the condition with `!template.endsWith("\n")`.

I also think it's clearer adding the conditional into the return sentence (but this is more of a personal thing, so please disregard if you don't agree):

`return template.endsWith("\n") ? template : template + "\n";`

graylog2-server/src/test/java/org/graylog/plugins/sidecar/collectors/ConfigurationServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+@UseModules({ObjectMapperModule.class, ValidatorModule.class})
+@CustomComparisonStrategy(comparisonStrategy = MongoFlexibleComparisonStrategy.class)
+public class ConfigurationServiceTest {
+    private final String FILEBEAT_CONF_ID = "5b8fe5f97ad37b17a44e2a34";
+    @Mock
+    private CollectorService collectorService;
```
edmundoa: This mock is never used and we should remove it.

graylog2-server/src/test/java/org/graylog/plugins/sidecar/collectors/ConfigurationServiceTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+import javax.validation.Validator;
+import java.util.List;
```
edmundoa: There are many imports here that are not used and we should remove. The best way of doing that is to check the "Optimize imports on the fly" option in your IDE:

screen shot 2018-09-27 at 12 44 37

mpfz0r: cool. thanks :)

Issue #5012, opened 2018-08-21
CSRF protection does not return proper error message


dennisoelkers:
In #4987 / #4998 simple protection against CSRF attacks was introduced. This breaks all consumers of the API which do not send the `X-Requested-By` for non-`GET` requests. Unfortunately, a consumer of the API which does not conform to this receives a very generic error looking like this:

```
{
	"type": "AutoValue_ApiError",
	"message": "HTTP 400 Bad Request"
}
```

which does not include any indication about the source of the problem. This makes it very hard for users of the API to fix the underlying issue.

PR #5182, opened 2018-10-05
Return helpful error message when CSRF protection fails. (#5177)

edmundoa:
Port of https://github.com/Graylog2/graylog2-server/pull/5177 into 2.5.


edmundoa:
Fixes #5012.

PR #5196, opened 2018-10-10
Fix failing content pack tests

bernd:
Broken tests have been introduced in https://github.com/Graylog2/graylog2-server/pull/5141.


dennisoelkers:
Why did the status check did not catch the failing tests? Do we have to adjust something else so this would have been caught in the original PR already?

bernd:
> Why did the status check did not catch the failing tests? Do we have to adjust something else so this would have been caught in the original PR already?

The tests in #5141 failed anyway because of a required change in the enterprise plugin. That's why we didn't catch this error.

PR #5202, opened 2018-10-12
Small fixes to content packs regarding UX

### LINKED ISSUE #5186: Content pack details rendering ugly if the description is empty
## Expected Behavior

It should indicate that the description is actually empty.

## Current Behavior
image

## Possible Solution

Either render a placeholder text, or hide the entire box and rendering a placeholder for the markdown output.
Right now it looks more broken than empty.

### LINKED ISSUE #5188: Inconsistent button size/placement for entities and parameters page
## Expected Behavior
Currently the page shown below uses various different button sizes, which looks odd if all of them are shown.
The presented data should be more visible with respect to navigational elements.

## Current Behavior
image

## Possible Solution
Use a single "mini" size for the action/edit buttons.
Button alignment is also off and line breaks happen, even though there would be enough space to show them inline.

### LINKED ISSUE #5189: Content pack Create vs Download button
At the very end of the content pack creation wizard, the user is given the option to create or download the content pack.

This confused me, because I wasn't sure what would happen if I clicked "Download". Does it also create the content pack? Will it discard what I had done?

I would either reword the buttons or always create the content pack and then off to download it, e.g. with a "Create" and "Create and Download" button.

### LINKED ISSUE #5191: Content pack "More action" menu confusing
## Expected Behavior
I previously installed a content pack and expected the action menu to allow me to uninstall it, but it only allows me to "Remove all".
From the action buttons it's also not clear how to get to the uninstall process (you need to click on the title on the other side of the screen and then uninstall).

## Current Behavior
image

The action menu has a "Remove all" entry, which is confusing, because it's unclear what "all" is. All content packs? All installed entities?

## Possible Solution

Rename the "Remove all" entry to "Delete content pack".
Either add "Uninstall..." option to show the detail page, or otherwise make it clear how to get there.

kmerz:
* Fix inconsistent button size. (Fixes #5188)
* Add a "Create and Download" Button (Fixes #5189)
* Hide description well when no description was entered (Fixes #5186)
* Show Intalled badge in content pack list (Fixes #5191)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
@@ -18,6 +18,8 @@ package org.graylog2.contentpacks;
 
 import com.google.common.collect.ImmutableSet;
 import com.mongodb.BasicDBObject;
+import com.mongodb.BasicDBObjectBuilder;
+import com.mongodb.DBObject;
 import org.bson.types.ObjectId;
 import org.graylog2.bindings.providers.MongoJackObjectMapperProvider;
 import org.graylog2.contentpacks.model.ContentPackInstallation;
@@ -36,6 +38,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 @Singleton
 public class ContentPackInstallationPersistenceService {
@@ -61,6 +64,16 @@ public class ContentPackInstallationPersistenceService {
         try (final DBCursor<ContentPackInstallation> installations = dbCollection.find()) {
             return ImmutableSet.copyOf((Iterator<ContentPackInstallation>) installations);
         }
+    }
+
+    public Set<ContentPackInstallation> findByContentPackIds(Set<ModelId> ids) {
+        final Set<String> stringIds = ids.stream().map(x -> x.toString()).collect(Collectors.toSet());
+        final DBObject query = BasicDBObjectBuilder.start()
+                .push(ContentPackInstallation.FIELD_CONTENT_PACK_ID)
+                .append("$in", stringIds)
+                .get();
+        final DBCursor<ContentPackInstallation> result = dbCollection.find(query);
+        return ImmutableSet.copyOf((Iterable<ContentPackInstallation>) result);
     }
 
     public Optional<ContentPackInstallation> findById(ObjectId id) {
@@ -94,9 +107,8 @@ public class ContentPackInstallationPersistenceService {
         return writeResult.getN();
     }
 
-    public Map<ModelId, Map<Integer, ContentPackMetadata>> getInstallationMetadata() {
-        final Set<ContentPackInstallation> contentPackInstallations = loadAll();
-        Map <Integer, ContentPackMetadata> contentPackMetadataMap = new HashMap();
+    public Map<ModelId, Map<Integer, ContentPackMetadata>> getInstallationMetadata(Set<ModelId> ids) {
+        final Set<ContentPackInstallation> contentPackInstallations = findByContentPackIds(ids);
         Map<ModelId, Map<Integer, ContentPackMetadata>> installationMetaData = new HashMap();
         for (ContentPackInstallation installation : contentPackInstallations) {
             Map<Integer, ContentPackMetadata> metadataMap = installationMetaData.get(installation.contentPackId());
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
@@ -109,11 +109,11 @@ public class ContentPackInstallationPersistenceService {
 
     public Map<ModelId, Map<Integer, ContentPackMetadata>> getInstallationMetadata(Set<ModelId> ids) {
         final Set<ContentPackInstallation> contentPackInstallations = findByContentPackIds(ids);
-        Map<ModelId, Map<Integer, ContentPackMetadata>> installationMetaData = new HashMap();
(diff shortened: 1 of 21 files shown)


graylog2-web-interface/src/components/content-packs/ContentPacksList.jsx
```diff
@@ -99,15 +104,17 @@ class ContentPacksList extends React.Component {
         contentPackId={item.id}
         revision={item.rev}
       />);
-      const states = item.states || [];
+
+      const installationCount = this.props.contentPackInstallCount[item.id];
```
bernd: This doesn't seem to work. I have an installed content pack but the overview doesn't show the badge.

You are trying to get the value for the `item.id` from the `this.props.contentPackInstallCount` object, but the `propTypes` defines this as an object with `contentPackId` and `installCount` fields. Also, the `ContentPacksList` component doesn't seem to receive the `contentPackInstallCount` props because it's always empty for me.

kmerz: Thanks for reviewing @bernd. You are right about the `PropTypes` shape, I used it wrongly. As you can see in the test the format should be `{ 'd3c72e5e-ce15-11e8-a8d5-f2801f1b9fd1': 3 }` (so uid: count).

But the rest seems to be a problem of the server. Did you recompile it?

bernd: You are right, of course. I didn't restart the server... :see_no_evil: 

The badge looks a bit misplaced though.

image

How about putting it in front of the content pack title? Or closer to the buttons on the right?
That way we wouldn't increase the row size of the content pack entry.

kmerz: I agree. I will play arround and try to find a better solution!

graylog2-server/src/main/java/org/graylog2/rest/models/system/contenpacks/responses/ContentPackMetaData.java
```diff
@@ -0,0 +1,23 @@
+package org.graylog2.rest.models.system.contenpacks.responses;
```
bernd: This file is missing the license header and that makes the build fail. You can fix that by running `mvn license:format` in the server repository.

graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
```diff
@@ -90,4 +94,25 @@ public int deleteById(ObjectId id) {
         final WriteResult<ContentPackInstallation, ObjectId> writeResult = dbCollection.removeById(id);
         return writeResult.getN();
     }
+
+    public Map<ModelId, ContentPackMetaData> getInstallationMetaData() {
```
bernd: This creates the following structure in the `ContentPackList`:

```json
 "content_packs_meta_data": {
    "1794d39d-077f-7360-b92b-95411b05fbce": {
      "installation_count": {
        "1": 1
      }
    }
  }
```

I think it would be better if we have one metadata object for each content pack revision instead of one metadata object per content pack.
So it would look like this:

```json
 "content_packs_meta_data": {
    "1794d39d-077f-7360-b92b-95411b05fbce": {
      "1": {
         "installation_count": 1
      }
    }
  }
```

Otherwise we would need duplicate the revision for future metadata values.

graylog2-server/src/main/java/org/graylog2/rest/models/system/contenpacks/responses/ContentPackList.java
```diff
@@ -36,8 +38,13 @@
     @JsonProperty
     public abstract Set<ContentPack> contentPacks();
 
+    @JsonProperty
+    public abstract Map<ModelId, ContentPackMetaData> contentPacksMetaData();
```
bernd: I checked the rest of the server code base and we usually use "metadata" instead of "meta data". So I think we can change the name of the field to `contentPacksMetadata` so the generated field name in JSON is `content_packs_metadata`.

graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
bernd: I think the name of this method should be `getContentPackMetadata()` because it returns a map of metadata objects.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/contentpacks/ContentPackResource.java
```diff
@@ -115,7 +117,8 @@ public ContentPackList listLatestContentPacks() {
         checkPermission(RestPermissions.CONTENT_PACK_READ);
 
         Set<ContentPack> contentPacks = contentPackPersistenceService.loadAllLatest();
```
bernd: This endpoint is loading the latest revisions only. The `getInstallationMetaData()` method is loading all content packs though and will include the metadata for all content pack revisions.

How about passing the set of content packs into the `getInstallationMetaData()` method and only include the metadata for those content packs? That would avoid loading and including the metadata for all content packs and revisions all the time.

kmerz: We want to display if a content pack is installed regardless the revision. The `latest` endpoint will show all content packs but only the latest revision. So passing the ContentPacks to filter would not do any good.

bernd: We are currently loading all content packs from the backend and do the pagination in the frontend. Ususally we use [PaginatedDbService](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java) to do the pagination and filtering in the backend to avoid loading all entries from the database all the time.

Once we use that, we don't have to load all installations but only the ones we need for the content packs on the active page.

PR #5212, opened 2018-10-15
Make search query in alert conditions configurable in the UI

### LINKED ISSUE #3966: Custom Query For Alert Conditions
## Expected Behavior
Be able to configure the Query for an alert condition. 

## Current Behavior
The Query for an alert condition is `*` on a stream. 

## Possible Solution
Add one additional field to the alert condition that the experienced user is able to make a more specific search on a stream for the alert condition.

## Context
This way it is not needed to create lots of streams just to be able to alert on a certain type of message by adjusting the query that is done.

bernd:
This exposes a new query config field in the alert condition
configuration. The value defaults to "*".

The query is used to build a filter string for the alert condition
search. We add the query string to the query filter for performance
reasons.

This PR also adds two API endpoints to test new and existing alert conditions.

Fixes #3966

**Attention:** This needs to be cherry-picked into 2.5 once merged

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
+++ b/graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
@@ -38,7 +38,6 @@ import java.util.Map;
 import java.util.UUID;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Strings.isNullOrEmpty;
 
 public abstract class AbstractAlertCondition implements EmbeddedPersistable, AlertCondition {
     protected static final String CK_QUERY = "query";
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
```diff
@@ -155,6 +161,25 @@ public boolean shouldRepeatNotifications() {
...
+     * @param streamId the stream ID
+     * @param query    the query string (might be null or empty)
+     * @return the combined filter string
+     */
+    protected String buildQueryFilter(String streamId, String query) {
+        checkArgument(!isNullOrEmpty(streamId), "streamId parameter cannot be null or empty");
+
+        final StringBuilder builder = new StringBuilder().append("streams:").append(streamId);
+
+        if (!isNullOrEmpty(query) && !"*".equals(query)) {
```
kroepke: `CK_QUERY_DEFAULT_VALUE` and maybe trim?

bernd: I don't want to use the configuration constant here because the filter builder would break if the default value changes. I updated the code to trim the streamId and query, though. Good catch, thanks! :+1:

graylog2-server/src/main/java/org/graylog2/alerts/AbstractAlertCondition.java
```diff
@@ -155,6 +160,32 @@ public boolean shouldRepeatNotifications() {
...
+        final String trimmedStreamId = streamId.trim();
+
+        checkArgument(!trimmedStreamId.isEmpty(), "streamId parameter cannot be empty");
+
+        final StringBuilder builder = new StringBuilder().append("streams:").append(trimmedStreamId);
+
+        if (query != null) {
+            final String trimmedQuery = query.trim();
+            if (!trimmedQuery.isEmpty() && !"*".equals(trimmedQuery)) {
+                builder.append(" ").append(trimmedQuery);
```
edmundoa: Shouldn't we join both filters with an `AND` operator? Otherwise this does not further filter messages in the stream, as the user would expect.

bernd:
> Changing the alert condition query to AND POST worked as I initially expected, so to me it looks like we are missing to use the right operator between the stream query filter and the query the user provided.

@edmundoa Yes, you are right. I rebased and updated the PR to use the `AND` operator to combine the streams query and the user defined query.

PR #5214, opened 2018-10-16
Add Grok Pattern Dependency Resolution for Content Packs

kmerz:
## Description
Implement `resolveNativeEntity` and `resolveForInstallation` for GrokPatternFacade.

## Motivation and Context
To install grok patterns which depend on another we need to create a dependency resolution
with a graph.

## How Has This Been Tested?
* Installed Grok Patterns with Dependecies (and uninstalled them successfully)
* Created a content pack with grok pattern which dependen on each other.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/GrokPatternFacade.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/GrokPatternFacade.java
@@ -175,6 +175,33 @@ public class GrokPatternFacade implements EntityFacade<GrokPattern> {
     }
 
     @Override
+    public Graph<EntityDescriptor> resolveNativeEntity(EntityDescriptor entityDescriptor) {
+        final MutableGraph<EntityDescriptor> mutableGraph = GraphBuilder.directed().build();
+        mutableGraph.addNode(entityDescriptor);
+
+        final ModelId modelId = entityDescriptor.id();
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/contentpacks/facades/GrokPatternFacade.java
```diff
@@ -166,4 +173,69 @@ public EntityExcerpt createExcerpt(GrokPattern grokPattern) {
...
+                                                Map<EntityDescriptor, Entity> entities) {
+        if (entity instanceof EntityV1) {
+            return resolveForInstallationV1((EntityV1) entity, parameters, entities);
+        } else {
+            throw new IllegalArgumentException("Unsupported entity version: " + entity.getClass());
+        }
+    }
+
+    private Graph<Entity> resolveForInstallationV1(EntityV1 entity,
+                                                   Map<String, ValueReference> parameters,
```
bernd: The `parameters` are currently not used. We need to use them to resolve parameterized grok patterns. A user could parameterize the grok pattern iteself. If that happens we need to resolve the parameter. For example:

```java
final String namedPattern = grokPatternEntity.pattern().asString(parameters);
```

I doesn't really make sense to parameterize that kind of data, but I think we still have to do it correctly. :wink:

PR #5216, opened 2018-10-16
Add Bouncy Castle dependency

mpfz0r:
Having BC as a crypto provider offers several advantages over JCE.
Most importantly, it supports keys in PKCS#1 format, which is
still widely used, and the default output of `openssl genrsa`.

### DIFF
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -320,6 +320,11 @@
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-validator</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
+            <version>${bouncycastle.version}</version>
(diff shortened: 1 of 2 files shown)


dennisoelkers:
Why do we need the dependency if we have no code yet that uses it?

mpfz0r:
Because it is automatically added as a Security Provider and allows
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/security/tls/PemKeyStore.java#L102
to accept PKCS#1 encoded private Keys

dennisoelkers:
Do you know if it is possible to let it register as a security provider explicitly and prohibit registering others?

mpfz0r:
Why would you wanna do that?
If you want to give BC preference over other cyrpto providers we could insert it with the highest priority
using:

https://docs.oracle.com/javase/7/docs/api/java/security/Security.html#insertProviderAt(java.security.Provider,%20int)

Issue #5245, opened 2018-10-27
Timezone information for America/Sao_Paulo wrong


kikoreis:
## Expected Behavior

The timezone America/Sao_Paulo is currently at UTC-3. This was a change from the originally planned timezone defined in 2017 (which by now would have had DST kick in, putting us at UTC-2); on Ubuntu the change forced an update in the system timezones which happened about 2 montths ago.

On the local system, here is what I see:

    root@bdillogger:~# TZ=America/Sao_Paulo date
    Sáb Out 27 14:42:10 -03 2018

I believe the Java timezone is also correct, because a test program that does 

      LocalTime localTime = LocalTime.now();
      System.out.println(localTime);

Shows me 14:43:13.023 with an offset of -180s.

My user is configured to use the TZ "Sao_Paulo", and my server.conf has:

    root_timezone = America/Sao_Paulo

So my assumption is that Searches and the System Overview page should be showing the right timezone offset, at -0300.

## Current Behavior

But they are not, and instead are at -0200. If I do a search, a message with the following timestamp:

    2018-10-27T17:39:11.209Z

Is displayed in local time as:

    2018-10-27 15:39:11.209

which is -0200. The overview page confirms this:

 ```
User kiko:
2018-10-27 15:46:29 -02:00
Your web browser:
2018-10-27 14:46:29 -03:00
Graylog server:
2018-10-27 15:46:29 -02:00
```

I'm trying to figure out where Graylog is getting this value of -0200. Perhaps my tzdb.dat is wrong, but wouldn't that cause my test program to print the wrong time too?

## Steps to Reproduce (for bugs)
1. Configure the America/Sao_Paulo timezone for a user
2. Go to Graylog's System/Overview page
3. See -0200 for the user.

## Your Environment

* Graylog Version: 2.4.6-1
* Elasticsearch Version: 5.6.12
* MongoDB Version: 3.6.8
* Operating System: Ubuntu 16.04 LTS with latest updates applied
* Browser version: Any recent FF/Chromium

kikoreis:
The commit answered my original question -- this is due to the moment and moment-timezone packages which the UI depends on. Is it possible to update them manually in an installed Graylog?

edmundoa:
@kikoreis No, it is unfortunately not possible to update them manually, at least not without rebuilding the whole web interface. I'm afraid you will have to wait until 2.5 is out to get this issue resolved.

In the meantime you could workaround the issue by using UTC-3 as time zone.

[...]

PR #5247, opened 2018-10-29
Upgrade moment and moment-timezone

edmundoa:
This updates the time zone database we use in the web interface, fixing some incorrect time zones, including issue #5245.

These changes need to be cherry-picked into 2.5.

For a list of changes in both libraries since we last upgraded them, expand details.

<details>

### moment:
https://github.com/moment/moment/blob/develop/CHANGELOG.md

```
2.22.2 See full changelog
    Release May 31, 2018
    #4564 [bugfix] Avoid using trim()
    #4453 [bugfix] Treat periods as periods, not regex-anything period, for weekday parsing in strict mode.
    Minor locale improvements (pa-in, be, az)

2.22.1 See full changelog
    Release Apr 14, 2018
    #4495 [bugfix] Added HTML5_FMT to moment.d.ts
    Minor locale improvements
    QUnit upgrade and coveralls reporting

2.22.0 See full changelog
    Release Mar 30, 2018
    #4423 [new locale] Added Mongolian locale mn
    Various locale improvements
    Minor misc changes
```

### moment-timezone:
https://github.com/moment/moment-timezone/blob/develop/changelog.md

```
0.5.23 2018-10-28
    Fix minor issue with tz guessing in Russia #691

0.5.22 2018-10-28
    Updated data to IANA TZDB 2018g #689
    Fix issue with missing LMT entries for some zones, and fix data builds on Linux and Windows #308

0.5.21 2018-06-23
    Bugfix: revert breaking change introduced in 0.5.18

0.5.20 2018-06-18
    Bugfix: accidentally commented code

0.5.19 2018-06-18
    Revert: moved moment to peerDependencies

0.5.18 2018-06-18
    Return error when timezone name is not a string.
    Moved moment to peerDependencies #628
    Prefer nodejs to amd declaration #573

0.5.17 2018-05-12
    Updated data to IANA TZDB 2018d. #616

0.5.16 2018-04-18
    Fixed Etc/UTC timezone recognition, updated tests. #599
    Updated minified files to contain IANA TZDB 2018d data

0.5.15 2018-04-17
    Updated data to IANA TZDB 2018d. #596

0.5.14 2017-10-30
    Ensure Intl response is valid when guessing time zone. #553
    Updated data to IANA TZDB 2017c. #552
    Convert to tz keeping wall time #505
    Make all time zones available for guessing. #483
    zone.offset has been deprecated in favor of zone.utcOffset #398
    Check for timestamp formats when parsing #348

0.5.13 2017-04-04
    Bumped version to address Bower cache issues with last release. #474
    (No actual changes otherwise)

0.5.12 2017-04-02
    Updated data to IANA TZDB 2017b. #422
    Build the truncated data file as 2012-2022 (+/- 5 years).

0.5.11 2016-12-23
    Remove log statement when data is loaded twice. #352

0.5.10 2016-11-27
    Updated data to IANA TZDB 2016j. #422

0.5.9 2016-11-03
    Fixed the output of moment.tz.version. #413

0.5.8 2016-11-03
    Updated data to IANA TZDB 2016i. #411
```

</details>


edmundoa:
In my opinion it does not. First, we would be testing an external dependency, nothing in our product. Second, the country (in this case Brazil) may decide to revert the change that introduced the issue at some point, and we would have a failing test for no good reason.

kikoreis:
The annoying part about Brazil's DST changes is that they are not on a fixed calendar and change year over year. That is annoying but has evolved to getting handled reasonably well in the timezone ecosystem in general, but this year there was an additional complication, which is that the DST dates were changed a second time, extraordinarily, due to the realization that they would add risk to our presidential elections.

For a more permanent fix, I guess I'm more interested in knowing how we can have moment/moment-timezone updated in an installed Graylog. It's annoying that this is yet another timezone database, since the system already has two installed.

dennisoelkers:
@edmundoa: Can you please do a PR for `2.5` as well? Thanks!

PR #5255, opened 2018-10-31
Fix collector name validator when called on itself

mpfz0r:
If a sidecar collector name is edited on an existing
collector and then changed back to its original name,
the validation would falsely claim that this collector
already exists.
We need to exclude ourself from the query.

### DIFF
--- a/graylog2-web-interface/src/stores/sidecars/CollectorsStore.js
+++ b/graylog2-web-interface/src/stores/sidecars/CollectorsStore.js
@@ -178,11 +178,15 @@ const CollectorsStore = Reflux.createStore({
   },
 
   validate(name, collectorId) {
-    let url = `${this.sourceUrl}/collectors/validate/?name=${name}`;
+    const search = {
+      name: name,
+    };
     if (collectorId) {
-      url += `&id=${collectorId}`;
+      search.id = collectorId;
     }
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/stores/sidecars/CollectorsStore.js
```diff
@@ -177,8 +177,12 @@ const CollectorsStore = Reflux.createStore({
     CollectorsActions.copy.promise(promise);
   },
 
-  validate(name) {
-    const promise = fetch('GET', URLUtils.qualifyUrl(`${this.sourceUrl}/collectors/validate/?name=${name}`));
+  validate(name, collectorId) {
+    let url = `${this.sourceUrl}/collectors/validate/?name=${name}`;
+    if (collectorId) {
+      url += `&id=${collectorId}`;
```
edmundoa: This works, but I'd rather use URI.js to handle the URL search parameters, now that the logic in there is a bit more complex. This is how it would look like:

```jsx
const search = {
  name: name,
};

if (collectorId) {
  search.id = collectorId;
}

const url = URI(`${this.sourceUrl}/collectors/validate`).search(search).toString();
const promise = fetch('GET', URLUtils.qualifyUrl(url));
```

What do you think?

PR #5269, opened 2018-11-06
support for decimal IPv4 representation in to_ip

### LINKED ISSUE #5268: Add pipeline function support for integer to IPv4 conversion
Some systems represent IPv4 addresses in their decimal representation.
Currently the `to_ip` function only supports the string representation of an IP address, but it could easily try to convert integers as well.

## Expected Behavior
`to_ip("5.59.95.221") == to_ip(87777245)` should be `true`.

## Current Behavior
`to_ip(87777245)` fails to convert the address and returns the default IP address (`0.0.0.0` if not specified).

## Possible Solution
Implement integer support for converting IP addresses. Only support IPv4 addresses, because we don't have bigint support and IPv6 addresses usually are given in their hex form.

## Context
Forcepoint logs IP addresses this way.

kroepke:
if the argument is a number, try to convert its intValue into an IP address
otherwise the previous behavior of taking the string value and converting that applies

fixes #5268

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ips/IpAddressConversion.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ips/IpAddressConversion.java
@@ -46,6 +46,7 @@ public class IpAddressConversion extends AbstractFunction<IpAddress> {
         final Object ip = ipParam.required(args, context);
         try {
             if (ip instanceof Number) {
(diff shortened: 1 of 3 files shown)


ChristopherKBlack:
I don't know how to interpret the messages below, but it looks like we have hit a snag in this process. 

If/when this gets merged successfully, when will this feature make it into the product? I'm asking version number, not date. I would like to give the customer some idea when to expect it.

Thanks.

kroepke:
@ChristopherKBlack by default 3.0, but since it is trivial, I'll look at backporting it to 2.5

PR #5274, opened 2018-11-06
Backport DNS Lookup adapter to version 2.5

danotorrey:
Backport DNS Lookup adapter to version 2.5. 
    
Add/update dependencies to allow backport:
- Add Netty 4.1.31.FINAL dependency
- Upgrade Google Guava from 23 to 23.6.1-jre

See https://github.com/Graylog2/graylog2-server/commit/5d755477c46150345ff111d54f45ae5cf1b87938 for original commit in master.

### DIFF
--- a/graylog2-web-interface/src/components/lookup-tables/adapters/index.js
+++ b/graylog2-web-interface/src/components/lookup-tables/adapters/index.js
@@ -1,8 +1,12 @@
 import { PluginManifest, PluginStore } from 'graylog-web-plugin/plugin';
 
+import {} from 'components/maps/adapter';
 import CSVFileAdapterFieldSet from './CSVFileAdapterFieldSet';
 import CSVFileAdapterSummary from './CSVFileAdapterSummary';
 import CSVFileAdapterDocumentation from './CSVFileAdapterDocumentation';
+import DnsAdapterFieldSet from './DnsAdapterFieldSet';
+import DnsAdapterSummary from './DnsAdapterSummary';
+import DnsAdapterDocumentation from './DnsAdapterDocumentation';
 import DSVHTTPAdapterFieldSet from './DSVHTTPAdapterFieldSet';
 import DSVHTTPAdapterSummary from './DSVHTTPAdapterSummary';
 import DSVHTTPAdapterDocumentation from './DSVHTTPAdapterDocumentation';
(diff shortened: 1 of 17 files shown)


danotorrey:
I am investigating the CI failures.

I am also planning to review this with Bernd next week.

graylog2-web-interface/src/components/lookup-tables/adapters/index.js
```diff
@@ -1,8 +1,12 @@
 import { PluginManifest, PluginStore } from 'graylog-web-plugin/plugin';
 
+import {} from 'components/maps/adapter';
```
bernd: This is why the build fails on CI. It looks like this is an unrelated change that snuck in by accident.

danotorrey: Good eye!! Thanks. graylog.conf was also erroneously committed. I've removed both. Let's see if the CI builds successfully now.

danotorrey:
I also added a missing netty-buffer dependency, which was causing a build failure. Let's see if the build passes this time.

danotorrey:
@bernd CI build is not passing. This is ready for review.

mpfz0r:
If you update Netty in 2.5 make sure to include
https://github.com/Graylog2/graylog2-server/pull/5279

danotorrey:
Thank you @mpfz0r! I definitely will include that change in 2.5.

danotorrey:
The `netty-tcnative-boringssl-static.version` dependency was added in the original Netty 4 upgrade in early 2018 (see https://github.com/Graylog2/graylog2-server/commit/6762727721a8a40679cd98992a3d4c756815d4ed). Though, I am not sure specifically what the dependency is for. It is not included in 2.5. I think we're OK without it.  

@bernd Do you know anything about `netty-tcnative-boringssl-static.version`? Are we ok without it here? I think so.

bernd:
@danotorrey If we don't have the tcnative dependency in the 2.5 pom.xml, we don't need it. I think that one has been added to 3.0 to improve TLS performance for our message inputs.

danotorrey:
@bernd Should I wait until the Gelfclient is released and update the version in this same PR? Seems like that might be ok, since the `netty4` version is being added here too.

bernd:
> @bernd Should I wait until the Gelfclient is released and update the version in this same PR? Seems like that might be ok, since the `netty4` version is being added here too.

The new gelfclient version has been release and should be available. So yes, feel free to just do it in this PR. :+1:

danotorrey:
Ok, will do.

danotorrey:
@bernd `gelfclient` has been updated. Ready for review.

comnam90:
When will this be released?

danotorrey:
@comnam90 This is available in Graylog version 2.5, which has actually already been released. More information is available in the following page in the Graylog docs.

http://docs.graylog.org/en/2.5/pages/lookuptables.html?highlight=dns#dns-lookup-adapter

Please let me know if you have any questions.

comnam90:
@danotorrey Thanks! must've missed it when i was looking!

PR #5278, opened 2018-11-08
Remove Sidecar collector configuration path

mpfz0r:
Instead of a GUI option, we let the sidecar autogenerate the path
to ".../generated/<collector_name>.conf"

Add validations for the collector name, so it can be used as a filename.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
@@ -16,8 +16,14 @@
  */
 package org.graylog.plugins.sidecar.migrations;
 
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -96,19 +108,32 @@ public void upgrade() {
...
                 "-v -c %s",
                 ""
         );
     }
 
+    private void removeConfigPath() {
+        final FindIterable<Document> documentsWithConfigPath = collection.find(exists("configuration_path"));
+        for (Document document : documentsWithConfigPath) {
+            final ObjectId objectId = document.getObjectId("_id");
+            System.out.print( document);
```
mariussturm: Debug code!

PR #5279, opened 2018-11-08
Update netty-tcnative to 2.0.19

### LINKED ISSUE #5263: Beats TLS Inputs fail to Start on Graylog 3.0 Alpha 3
I built the current 3.0 alpha and it appears that beats input or beats input depreciated fails to start when TLS is specified.  This issue occurs on both openjdk and oracle jdk.  If you disable TLS the input starts as expected.  I checked to see if self signed certificates were being generated and they were.  Also specifying a self signed certificate has the same behavior with the input stuck on starting.  I checked permissions of the certificate folders and files, graylog:graylog has access.

mpfz0r:
The netty update to 4.1.31 prevented TLS enabled inputs to start.
The newest netty-tcnative version fixes that.

Fixes #5263
Refs #5250

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,7 +128,7 @@
         <mongojack.version>2.8.2</mongojack.version>
         <natty.version>0.13</natty.version>
         <netty.version>4.1.31.Final</netty.version>
-        <netty-tcnative-boringssl-static.version>2.0.12.Final</netty-tcnative-boringssl-static.version>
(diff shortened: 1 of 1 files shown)


danotorrey:
The `netty-tcnative-boringssl-static.version` dependency was added in the original Netty 4 upgrade in early 2018 (see https://github.com/Graylog2/graylog2-server/commit/6762727721a8a40679cd98992a3d4c756815d4ed). Though, I am not sure specifically what the dependency is for. It is not included in 2.5. I think we're OK without it. I will confirm with @bernd.

PR #5296, opened 2018-11-20
Redirect to content pack overview if no content pack can be found

kmerz:
## Description
If the user requests a content pack which is not there anymore we now redirect to the
content packs overview.

## Motivation and Context
If a user deleted the last content pack the show content pack site was throwing an error.

## How Has This Been Tested?
Removing the last revision of an content pack.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
+++ b/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
@@ -35,7 +35,9 @@ const ShowContentPackPage = createReactClass({
 
 
   componentDidMount() {
-    ContentPacksActions.get(this.props.params.contentPackId);
+    ContentPacksActions.get(this.props.params.contentPackId).catch(() => {
+      window.location = '/system/contentpacks';
+    });
     ContentPacksActions.installList(this.props.params.contentPackId);
   },
 
@@ -47,16 +49,16 @@ const ShowContentPackPage = createReactClass({
     if (window.confirm('You are about to delete this content pack, are you sure?')) {
       ContentPacksActions.deleteRev(contentPackId, revision).then(() => {
         UserNotification.success('Content Pack deleted successfully.', 'Success');
-        ContentPacksActions.get(contentPackId);
+        ContentPacksActions.get(contentPackId).catch(() => {
+          window.location = '/system/contentpacks';
+        });
       }, (error) => {
-        /* eslint-disable camelcase */
-        let err_message = error.message;
-        const err_body = error.additional.body;
-        /* eslint-enable camlecase */
-        if (err_body && err_body.message) {
-          err_message = error.additional.body.message;
+        let errMessage = error.message;
+        const errBody = error.additional.body;
+        if (errBody && errBody.message) {
+          errMessage = error.additional.body.message;
         }
-        UserNotification.error(`Deleting bundle failed: ${err_message}`, 'Error');
+        UserNotification.error(`Deleting bundle failed: ${errMessage}`, 'Error');
       });
     }
   },
--- a/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
+++ b/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
@@ -36,7 +36,7 @@ const ShowContentPackPage = createReactClass({
 
 
   componentDidMount() {
-    ContentPacksActions.get(this.props.params.contentPackId).catch((error) => {
+    ContentPacksActions.get(this.props.params.contentPackId).catch(() => {
       UserNotification.error("Cannot find a matching Content Pack.");
       history.push(Routes.SYSTEM.CONTENTPACKS.LIST);
     });
--- a/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
+++ b/graylog2-web-interface/src/pages/ShowContentPackPage.jsx
@@ -53,9 +53,9 @@ const ShowContentPackPage = createReactClass({
   },
 
   _deleteContentPackRev(contentPackId, revision) {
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -47,16 +51,16 @@ const ShowContentPackPage = createReactClass({
...
-        /* eslint-disable camelcase */
-        let err_message = error.message;
-        const err_body = error.additional.body;
-        /* eslint-enable camlecase */
-        if (err_body && err_body.message) {
-          err_message = error.additional.body.message;
+        let errMessage = error.message;
+        const errBody = error.additional.body;
+        if (errBody && errBody.message) {
+          errMessage = error.additional.body.message;
```
edmundoa: You can use `error.responseMessage` to get the body message if there's any.

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -35,7 +36,10 @@ const ShowContentPackPage = createReactClass({
 
 
   componentDidMount() {
-    ContentPacksActions.get(this.props.params.contentPackId);
+    ContentPacksActions.get(this.props.params.contentPackId).catch(() => {
+      UserNotification.error('Cannot find a matching Content Pack.');
```
edmundoa: In case the backend returns any other error than a 404, we will tell the user we couldn't find the content pack, where that is not really what happened. I think we should check the error status is a 404 before doing that.

edmundoa: The error message is somehow confusing for me, I think because of the "matching" in there. As far as I could think, most times the user will find a 404 it is because the content pack was deleted, so maybe we can change that message into something like:

```
UserNotification.error(`The Content Pack with ID "${id}" was not found and may have been deleted.`, 'Cannot find the Content Pack')
```

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -47,16 +56,18 @@ const ShowContentPackPage = createReactClass({
...
-        const err_body = error.additional.body;
-        /* eslint-enable camlecase */
-        if (err_body && err_body.message) {
-          err_message = error.additional.body.message;
+        let errMessage = error.message;
+        if (error.responseMessage) {
+          errMessage = error.responseMessage;
         }
-        UserNotification.error(`Deleting bundle failed: ${err_message}`, 'Error');
+        UserNotification.error(`Deleting bundle failed: ${errMessage}`, 'Error');
```
edmundoa: "Bundle" was one of the 3 or so terms we used for content packs, so I guess we need to rename that :)

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -47,16 +56,18 @@ const ShowContentPackPage = createReactClass({
     if (window.confirm('You are about to delete this content pack, are you sure?')) {
       ContentPacksActions.deleteRev(contentPackId, revision).then(() => {
         UserNotification.success('Content Pack deleted successfully.', 'Success');
```
edmundoa: I think this feedback message is somehow outdated, right? Otherwise I don't understand why we delete a content pack and then try to fetch it.

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
```diff
@@ -47,16 +56,18 @@ const ShowContentPackPage = createReactClass({
...
-        ContentPacksActions.get(contentPackId);
+        ContentPacksActions.get(contentPackId).catch((error) => {
+          if (error.status !== 404) {
```
edmundoa: Should we display a message when the error is a 404?

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
kmerz: Because we delete a content pack revision. The content pack remains. We refetch the content pack to show the updated list of revisions. But if you delete the last content pack revision we want to redirect to the overview.

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
kmerz: No because we deleted the last content pack revision and that is okay, so we get to the content packs list silently.

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
kmerz: :+1: fixed!

graylog2-web-interface/src/pages/ShowContentPackPage.jsx
edmundoa: Alright, then the message should say something like "Content Pack revision deleted successfully". I would be confused if I see the original message when I'm deleting some old version of a Content Pack.

edmundoa: Oh, I saw you changed it already. Sorry about the noise 😄

edmundoa:
I rebased this PR with the changes in master before testing and I tried to create a Content Pack that I could delete to test the 404 behaviour. There is an error on the last step of the creation wizard after clicking "Create":

```
TypeError: this.state.contentPack.toObject is not a function
	_onSave CreateContentPackPage.jsx:58
	_onSave (index):974
```

Not sure if is related to the changes in here, but I just let you know so you can take a look and see if it fits in the changes for this PR or not.

PR #5297, opened 2018-11-20
[WIP] Add ES reindex API

bernd:
This is WIP.

It was initially supposed to go into 2.5.0 but we moved it out to save some time. This will eventually go into 3.x so we should keep this PR open for now.


bernd:
Closing this outdated PR. I would need a major overhaul to support ES7 and we are also didn't need it so far.

PR #5299, opened 2018-11-21
Calculate necessary uninstall for content pack and display it for confirmation

### LINKED ISSUE #5111: Show list of entities that will be removed on content pack uninstallation
The content pack uninstall dialog is currently not showing the list of entities that will be removed.

This should be implemented to make sure the user knows what will be removed.

### LINKED ISSUE #5259: Uninstall a content pack with dependency might lead to broken configuration
## Current Behavior
A content pack with a input and a lookup table as a dependency gets installed two times
with different parameters. The first one gets uninstalled, which also uninstalls the the lookup table needed by the second input, leaving a broken input configuration for 2.

## Expected Behavior
The lookup table should only be removed, if no other content pack installation is using it.

## Possible Solution
Before uninstalling we should check in the content pack installations if the entity is still referenced.

1. Create a content pack with a input and a lookup table extractor.
2. Install content pack two times.
3. Look that the lookup table was installed.
4. Uninstall first content pack installation.
5. Look that the lookup table was deleted.

kmerz:
## Description
When uninstalling a content pack it is important to only uninstall what is no longer needed.
e.g a Lookup Table can only be installed once but refereed to by many content pack installations.
So we should only install such content packs when it is the last reference.

To make it clearer for a user what will be uninstalled, we ask the server to calculate beforehand
and show the result in the confirmation dialog.

Fixes #5259 
Fixes #5111 

## How Has This Been Tested?
Installed a content pack with a lookup table and a input two times and the first uninstall only
removed the input and the second uninstall the input and the lookup table.

graylog - content packs 10

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackInstallationPersistenceService.java
```diff
@@ -126,4 +126,19 @@ public int deleteById(ObjectId id) {
...
+    public int countInstallationOfEntityById(ModelId contentPackEntityId) {
+        final Set<ContentPackInstallation> allInstallations = loadAll();
+        Optional<Integer> result = allInstallations.stream().map(contentPackInstallation -> {
+            return contentPackInstallation.entities().stream().
+                    filter(nativeEntityDescriptor -> nativeEntityDescriptor.contentPackEntityId()
+                            .equals(contentPackEntityId)).collect(Collectors.toSet()).size();
+        }).reduce((acc, next) -> acc + next );
+
+
+        if (result.isPresent())
```
bernd: Please use `{}` around if-statements. But I think this whole method can be simplified to the following to move the counting operation into the database:

```java
    /**
     * Returns the number of installations the given content pack entity ID is used in.
     * @param contentPackEntityId the content pack entity ID
     * @return number of installations
     */
    public long countInstallationOfEntityById(ModelId contentPackEntityId) {
        final String field = String.format(Locale.ROOT, "%s.%s", ContentPackInstallation.FIELD_ENTITIES, NativeEntityDescriptor.FIELD_ENTITY_ID);

        return dbCollection.getCount(DBQuery.is(field, contentPackEntityId));
    }
```

This returns the number of installations that use the given entity ID. Before it returned the number of entities in all installations that use the given entity ID.

I think the new behavior is even more correct because in [ContentPackService#uninstallDetails()](https://github.com/Graylog2/graylog2-server/pull/5299/files#diff-dbbfb51fd730e758a0e91273028b7407R206)  we compare this result to `<= 1`. With the old behavior, if a single content pack installation would use the entity ID in several entities, that test would fail but it should not. (same issue [here](https://github.com/Graylog2/graylog2-server/pull/5299/files#diff-dbbfb51fd730e758a0e91273028b7407R261))

kmerz: Since a `content_pack_entity_id` can only be appear once in a installation the error case described above is not valid. Anyhow is the described solution much more efficient than mine. So I adapted the code accordingly.

PR #5303, opened 2018-11-21
Add Sidecar collector configuration variables

mpfz0r:
This supersedes the '<#inlcude "config-snippet">' mechanism.

Refactor the former "Filebeat Reference" box to a more
generic "Collector Configuration Reference".


graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+        <br />
+        <br />
+        <br />
+        <div className={`table-responsive ${ConfigurationHelperStyle.tableMaxHeight}`}>
+          <Table responsive>
+            <thead>
+              <tr>
+                <th>Name</th>
+                <th>Description</th>
+                <th style={{ textAlign: 'right' }} >Actions</th>
```
edmundoa: You should use the `actions` class name to make this consistent with other tables.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+        ConfigurationVariableActions.delete(configVar)
+          .then(() => this._onSuccessfulUpdate());
+      };
+
+      const escapedName = `\${${configVar.name}}`;
+      variableRows.push(
+        <tr key={this._getId(configVar.id)}>
+          <td><code>{escapedName}</code></td>
+          <td>{configVar.description}</td>
+          <td style={{ textAlign: 'right' }}>
```
edmundoa: Please avoid inline styles and move this to `ConfigurationHelperStyle.css` instead.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+            <BootstrapModalConfirm ref={(c) => { deleteConfirmationModal = c; }}
+                                   title="Delete Configuration Variable?"
+                                   onConfirm={_handleDeleteConfirm}
+                                   onCancel={() => {}}>
+              <p>Are you sure you want to remove the configuration variable <strong>{configVar.name}</strong>?</p>
+            </BootstrapModalConfirm>
+            <Button bsStyle="primary" bsSize="xsmall" onClick={_handleDeleteCheck}>
+              Delete
+            </Button>
```
edmundoa: The delete and edit button have no space between them. You could either set a `margin` in one of them, or add a `&nbsp;` to avoid that.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+      {configurations.map(conf => <li key={conf.id}><a href={conf.id}>{conf.name}</a></li>)}
+    </ul>);
+  };
+
+  _configurationVariableListFormatter = () => {
+    const variableRows = [];
+
+    Object.values(this.state.configurationVariables).forEach((configVar) => {
+      let deleteConfirmationModal;
+      const _handleDeleteCheck = () => {
```
edmundoa: I think it would make the code clearer moving this function and `_handleDeleteConfirm` out of the loop and into the class level. They could accept arguments with the variables they need to do their thing, and then make them return a function the consumers can call.

graylog2-web-interface/src/components/sidecars/configuration-forms/EditConfigurationVariableModal.jsx
```diff
@@ -0,0 +1,146 @@
...
+import { Button } from 'react-bootstrap';
+import { BootstrapModalForm, Input } from 'components/bootstrap';
+import CombinedProvider from 'injection/CombinedProvider';
+import ConfigurationHelperStyle from './ConfigurationHelper.css';
+
+const { ConfigurationVariableActions } = CombinedProvider.get('ConfigurationVariable');
+
+class EditConfigurationVariableModal extends React.Component {
+  static propTypes = {
+    create: PropTypes.boolean,
```
edmundoa: I'm getting this when loading the edit collector page:

```
Warning: Failed prop type: EditConfigurationVariableModal: prop type `create` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
    in EditConfigurationVariableModal (created by ConfigurationVariablesHelper)
    in ConfigurationVariablesHelper (created by ConfigurationHelper)
```

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+            <thead>
+              <tr>
+                <th>Name</th>
+                <th>Description</th>
+                <th style={{ textAlign: 'right' }} >Actions</th>
+              </tr>
+            </thead>
+            <tbody>
+              {this._configurationVariableListFormatter()}
+              <BootstrapModalWrapper ref={(modal) => { this.errorModal = modal; }}>
```
edmundoa: I would move the modal out of the table to make the markup a bit cleaner. In the end react-bootstrap takes care of picking the container for the modal, so it does not need to be in here.

edmundoa: I was playing a bit with the modal and have a version that I like a bit better, what do you think?

Before:
screenshot 2018-11-26 at 14 32 48


After:
screenshot 2018-11-26 at 14 32 18


```js
              <BootstrapModalWrapper ref={(modal) => { this.errorModal = modal; }}>
                <Modal.Header>
                  <Modal.Title>Error deleting configuration variable</Modal.Title>
                </Modal.Header>
                <Modal.Body>
                  <Alert bsStyle="warning">
                    <p>
                      Cannot delete this configuration variable as it is still in use. Please remove the variable from
                      the following configurations and try again.
                      {this.state.errorModalContent}
                    </p>
                  </Alert>
                </Modal.Body>
                <Modal.Footer>
                  <Button onClick={() => { this.errorModal.close(); }}>Close</Button>
                </Modal.Footer>
              </BootstrapModalWrapper>
```

graylog2-web-interface/src/components/sidecars/configuration-forms/EditConfigurationVariableModal.jsx
```diff
@@ -0,0 +1,146 @@
...
+  openModal = () => {
+    this.modal.open();
+  };
+
+  _getId = (prefixIdName) => {
+    return prefixIdName + this.state.name;
+  };
+
+  _closeModal = () => {
+    this.modal.close();
```
edmundoa: We should reset the modal contents after closing it, at least for consistency with other modals in the product.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationVariablesHelper.jsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  _reloadConfiguration = () => {
+    ConfigurationVariableActions.all()
+      .then((configurationVariables) => {
+        this.setState({ configurationVariables: configurationVariables });
+      });
+  };
+
+  _getId = (idName, index) => {
```
edmundoa: Do we need this method? We never pass an index to it, so I think we can remove it.

graylog2-web-interface/src/components/sidecars/configuration-forms/EditConfigurationVariableModal.jsx
```diff
@@ -0,0 +1,146 @@
...
+  _hasErrors = () => {
+    return !lodash.isEmpty(this.state.errors);
+  };
+
+  openModal = () => {
+    this.modal.open();
+  };
+
+  _getId = (prefixIdName) => {
+    return prefixIdName + this.state.name;
```
edmundoa: Could we use the ID instead of name to append to all IDs in the component? The name can change and with it the IDs inside the component, which may not be ideal.

As the ID will not be set when creating a new variable, we could use `this.state.id || 'new'`.

graylog2-web-interface/src/components/sidecars/configuration-forms/TemplatesHelper.jsx
```diff
@@ -5,8 +5,7 @@ class TemplatesHelper extends React.Component {
   render() {
     return (
       <div>
-        <h3>Variables</h3>
-        <h5 style={{ marginBottom: 10 }}>Operating System</h5>
+        <h5>These variables will be filled with the runtime information from each Sidecar</h5>
```
edmundoa: I would use a paragraph in here, otherwise this text is larger than most fonts on the page, even if it is not that important.

graylog2-web-interface/src/stores/sidecars/ConfigurationVariableStore.js
```diff
@@ -0,0 +1,114 @@
...
+    const uri = URI(baseUrl).toString();
+
+    return fetch('GET', URLUtils.qualifyUrl(uri));
+  },
+
+  all() {
+    const promise = this._fetchConfigurationVariables();
+    promise
+      .then(
+        () => {},
```
edmundoa: If you don't want to use the `then` in the promise, you can call `catch` instead:

`promise.catch((error) => {...})`

graylog2-web-interface/src/stores/sidecars/ConfigurationVariableStore.js
```diff
@@ -0,0 +1,114 @@
...
+
+const ConfigurationVariableStore = Reflux.createStore({
+  listenables: [ConfigurationVariableActions],
+  sourceUrl: '/sidecar/configuration_variables',
+
+  _fetchConfigurationVariables() {
+    const baseUrl = `${this.sourceUrl}`;
+    const uri = URI(baseUrl).toString();
+
+    return fetch('GET', URLUtils.qualifyUrl(uri));
```
edmundoa: Couldn't we skip the previous two sentences and use this last one instead?

graylog2-web-interface/src/stores/sidecars/ConfigurationVariableStore.js
```diff
@@ -0,0 +1,114 @@
...
+      method = 'POST';
+    } else {
+      url += `/${configurationVariable.id}`;
+      method = 'PUT';
+    }
+
+    const promise = fetch(method, url, request);
+    promise
+      .then(() => {
+        const action = configurationVariable.id === '' ? 'created' : 'updated';
```
edmundoa: I think setting this variable in the if above would make it clearer, as all values depending on the type of action would be set in the same place.

graylog2-web-interface/src/stores/sidecars/ConfigurationVariableStore.js
```diff
@@ -0,0 +1,114 @@
...
+      });
+
+    ConfigurationVariableActions.save.promise(promise);
+  },
+
+  getConfigurations(configurationVariable) {
+    const url = URLUtils.qualifyUrl(`${this.sourceUrl}/${configurationVariable.id}/configurations`);
+    const promise = fetch('GET', url);
+    promise.then(
+      () => {},
```
edmundoa: Same as above, I would use `catch` in here instead of an empty `then`.

graylog2-web-interface/src/stores/sidecars/ConfigurationVariableStore.js
```diff
@@ -0,0 +1,114 @@
...
+      description: configurationVariable.description,
+      content: configurationVariable.content,
+    };
+    const url = URLUtils.qualifyUrl(`${this.sourceUrl}/validate`);
+    const method = 'POST';
+
+    const promise = fetch(method, url, request);
+    promise
+      .then(() =>
+        response => response,
```
edmundoa: I'm not sure what you are trying to do here :)

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationVariableResource.java
```diff
@@ -0,0 +1,199 @@
...
+    @GET
+    @Path("/{id}/configurations")
+    @RequiresPermissions(SidecarRestPermissions.CONFIGURATIONS_READ)
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Show configurations using this variable")
+    public List<Configuration> getConfigurationVariablesConfigurations(@ApiParam(name = "id", required = true)
+                                                                       @PathParam("id") String id) {
+        final ConfigurationVariable configurationVariable = configurationVariableService.find(id);
+        if (configurationVariable == null) {
+            throw new NotFoundException("Could not find ConfigurationVariable <" + id + ">.");
```
edmundoa: I would extract loading and checking a variable into a function, as we seem to be using it in some other places.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationVariableResource.java
```diff
@@ -0,0 +1,199 @@
...
+import javax.ws.rs.core.Response;
+import java.util.List;
+import java.util.regex.Pattern;
+
+@Api(value = "Sidecar/ConfigurationVariables", description = "Manage collector configuration variables")
+@Path("/sidecar/configuration_variables")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+@RequiresAuthentication
+public class ConfigurationVariableResource extends RestResource implements PluginRestResource {
```
edmundoa: Do we need to implement the `PluginRestResource` interface? Not sure how we have it in other resources, so just double-checking :)

graylog2-web-interface/src/components/sidecars/configuration-forms/DeleteConfirmButton.jsx
```diff
@@ -0,0 +1,27 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Button } from 'react-bootstrap';
+
+class DeleteConfirmButton extends React.Component {
```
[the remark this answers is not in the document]
mpfz0r: Oh, this is actually not needed anymore. I used that in an older development version.
`git rm`ing this :)

graylog2-web-interface/src/components/sidecars/configuration-forms/DeleteConfirmButton.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+  handleClick = () => {
+    if (window.confirm(`You are about to delete ${this.props.type} "${this.props.entity.name}". Are you sure?`)) {
+      this.props.onClick(this.props.entity);
+    }
+  };
+
+  render() {
+    return (
+      <Button bsStyle="primary" bsSize="xsmall" onClick={this.handleClick}>
```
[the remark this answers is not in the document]
mpfz0r: see above

graylog2-web-interface/src/components/sidecars/configuration-forms/DeleteConfirmButton.jsx
```diff
@@ -0,0 +1,27 @@
...
+        Delete
```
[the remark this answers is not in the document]
mpfz0r: see above

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationVariableResource.java
bernd: @edmundoa The plugins we moved into the server are still acting like plugins (not sure why that decision has been made, though) so the resources need to implement `PluginRestResource` to work with `addRestResource()` in the bindings.

edmundoa: Makes sense, thank you for the clarification! 👍

bernd:
I will also take a look at the backend part.

graylog2-web-interface/src/pages/SidecarEditConfigurationPage.jsx
```diff
@@ -52,6 +52,10 @@ const SidecarEditConfigurationPage = createReactClass({
     return !this.state.configuration || !this.state.configurationSidecars;
   },
 
+  _configurationVariableRenameHandler(oldname, newname) {
+    this.configurationForm._replaceConfigurationVariableName(oldname, newname);
```
edmundoa: I see a couple of problems with this:
- This works but does not feel like react. Ideally you want to let `ConfigurationForm` handle its own state, not force a change from outside the component. I think there is a piece of state that should defined in this component, and it's the variable list. In that way you could pass the list down to both `ConfigurationForm` and `ConfigurationHelper`, allowing `ConfigurationForm` to rename variables when they change in its props.
- Here you are using a private method of the configurationForm component (at least that's the convention of using the `_` prefix on a function name).

graylog2-web-interface/src/pages/SidecarNewConfigurationPage.jsx
```diff
@@ -12,6 +12,10 @@ import ConfigurationHelper from 'components/sidecars/configuration-forms/Configu
 const SidecarNewConfigurationPage = createReactClass({
   displayName: 'SidecarNewConfigurationPage',
 
+  _configurationVariableRenameHandler(oldname, newname) {
+    this.configurationForm._replaceConfigurationVariableName(oldname, newname);
```
edmundoa: Same as above.

graylog2-web-interface/src/components/sidecars/configuration-forms/TemplatesHelper.jsx
```diff
@@ -16,31 +14,31 @@ class TemplatesHelper extends React.Component {
           </thead>
           <tbody>
             <tr>
-              <td><code>{'${'}operatingSystem{'}'}</code></td>
+              <td><code>{'${sidecar.'}operatingSystem{'}'}</code></td>
```
edmundoa: Why not moving this logic into a method? In that way we would only need to replace it once if we ever change the prefix.

graylog2-web-interface/src/components/sidecars/configuration-forms/EditConfigurationVariableModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+    this.setState(this._cleanState());
+    this.modal.open();
+  };
+
+  _getId = (prefixIdName) => {
+    return prefixIdName + this.state.id || 'new';
+  };
+
+  _closeModal = () => {
+    this.setState(this._cleanState());
```
edmundoa: We don't need to set a clean state both on opening and closing the modal.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationHelper.jsx
```diff
@@ -8,12 +8,12 @@ import DocsHelper from 'util/DocsHelper';
...
 import FilebeatHelper from './FilebeatHelper';
 import TemplatesHelper from './TemplatesHelper';
-import IncludesHelper from './IncludesHelper';
+import ConfigurationVariablesHelper from './ConfigurationVariablesHelper';
 import ConfigurationHelperStyle from './ConfigurationHelper.css';
 
 class ConfigurationHelper extends React.Component {
   static propTypes = {
-    type: PropTypes.string.isRequired,
+    variableRenameHandler: PropTypes.func.isRequired,
```
edmundoa: I would rename this to better match our most usual conventions: `onVariableRename` would be the prop name, `variableRenameHandler` would be the function that handles the "event" in the end.

graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationHelper.jsx
```diff
@@ -62,22 +62,36 @@ class ConfigurationHelper extends React.Component {
...
 
         <Row className="row-sm">
           <Col md={12}>
             <Tabs id="configurationsHelper" defaultActiveKey={1} animation={false}>
-              <Tab eventKey={1} title="Reference">
-                <br />
+              <Tab eventKey={1} title="Variables">
+                <p className={ConfigurationHelperStyle.marginQuickReferenceText}>
+                  Use variables to share text snippets across multiple configurations.<br />
+                  Unexpanded literals like <code>$&#123;foo&#125;</code> can be written as
```
edmundoa: I think we need to clarify this sentence to make it easier to understand.

PR #5306, opened 2018-11-22
Fix index stats collection (#5226)

bernd:
Use the correct fields for indexing and merges.

(cherry picked from commit 229b11edc8350ff899d64cd9ee29ad20682e7f58)


bernd:
This fixes the index stats for the detail index view:

## Before

image

## After

image

PR #5307, opened 2018-11-22
Fix react CSS warning

mpfz0r:
Number.isNan(Infinity) is false. Compare against Infinity instead.

### DIFF
--- a/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
+++ b/graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import lodash from 'lodash';
 import { PropTypes } from 'prop-types';
 import { Resizable } from 'react-resizable';
 import 'brace';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -134,7 +134,7 @@ class SourceCodeEditor extends React.Component {
   render() {
     const { height, width } = this.state;
     const { theme, resizable } = this.props;
-    const validCssWidth = Number.isNaN(width) ? '100%' : width;
+    const validCssWidth = (width === Infinity) ? '100%' : width;
```
edmundoa: When resizing the editor, I'm seeing this warning in the console:

```
Warning: `NaN` is an invalid value for the `width` css style property.
```

Reading a bit around, I guess we should use [`lodash.isFinite(width)`](https://lodash.com/docs/4.17.11#isFinite) to check the width. That method checks if the value is a number and not `Infinity`, `-Infinity` or `NaN`.

mpfz0r: Yep, makes sense. Thx!

PR #5309, opened 2018-11-23
Create value classes for Content Packs

kmerz:
## Description
Replace the big content pack hash with a value class.

## Motivation and Context
Value classes help to ensure default values, encapsulate accessors and help find bugs while writing code.

## How Has This Been Tested?
Clicked (hopefully) through all scenarios. 

## Types of changes

## Checklist

### DIFF
--- a/graylog2-web-interface/src/components/content-packs/ContentPackEdit.test.jsx
+++ b/graylog2-web-interface/src/components/content-packs/ContentPackEdit.test.jsx
@@ -3,6 +3,7 @@ import renderer from 'react-test-renderer';
 import { mount } from 'enzyme';
 import 'helpers/mocking/react-dom_mock';
 
+import ContentPack from 'logic/content-packs/ContentPack';
 import ContentPackEdit from 'components/content-packs/ContentPackEdit';
 
 jest.mock('react-overlays', () => ({
@@ -12,19 +13,10 @@ jest.mock('react-overlays', () => ({
 
 
 describe('<ContentPackEdit />', () => {
-  const emptyContentPack = {
-    v: 1,
-    id: '9950ba5a-0887-40a9-2b8f-8b50292cc7fb',
-    rev: 1,
-    requires: [],
-    parameters: [],
-    entities: [],
-    name: '',
-    summary: '',
-    description: '',
-    vendor: '',
-    url: '',
-  };
+  const emptyContentPack = ContentPack.builder()
+    .id('9950ba5a-0887-40a9-2b8f-8b50292cc7fb')
+    .build();
+
   const enrichedEntity = {
     id: '111-beef',
     v: '1.0',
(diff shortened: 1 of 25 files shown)


graylog2-web-interface/src/components/content-packs/ContentPackSelection.test.jsx
```diff
@@ -43,25 +44,17 @@ describe('<ContentPackSelection />', () => {
...
-        name: 'name',
-        summary: 'summary',
-        description: 'descr',
-        vendor: 'vendor',
-        url: 'url',
-      };
       called += 1;
-      if (called === Object.keys(contentPack).length) {
-        expect(state.contentPack).toEqual(contentPack);
+      if (called === 5) {
```
dennisoelkers: What happens when one of the change callbacks does not fire? The following `expect()` calls will never be executed then.

graylog2-web-interface/src/logic/content-packs/ContentPack.js
```diff
@@ -0,0 +1,249 @@
...
+      vendor,
+      url,
+      requires,
+      parameters,
+      entities,
+    }));
+  }
+
+  toObject() {
+    return this._value;
```
dennisoelkers: I think we should never expose the internal state of the object directly, it makes the whole object mutable. We should either clone the state before returning it, or (way better in this case) implement a `toJSON` function that returns a new object containing the desired serialization format for JSON (As AFAIK `toObject` is only used in conjunction with `JSON.stringify`).

graylog2-web-interface/src/components/content-packs/ContentPackEdit.test.jsx
```diff
@@ -46,19 +38,16 @@ describe('<ContentPackEdit />', () => {
   const selectedEntities = { dashboard: [{ id: '111-beef', type: 'dashbaord' }] };
 
   const parameter = { title: 'Port', name: 'PORT', type: 'integer', default_value: '23' };
```
dennisoelkers: Maybe it is beyond the scope of this PR, but maybe we should introduce a `Parameter` value class as well?

kmerz: I did a value class for Entity in a following PR already. And will continue to transfer at need. (Which will come sooner or later!)

graylog2-web-interface/src/components/content-packs/ContentPackSelection.test.jsx
kmerz: So this expect will fail:
`expect(changeFn.mock.calls.length).toBe(5);`

graylog2-web-interface/src/components/content-packs/ContentPackPreview.jsx
```diff
@@ -18,7 +18,7 @@ class ContentPackPreview extends React.Component {
   };
 
   _renderDownload() {
-    const contentPackEncoded = encodeURIComponent(JSON.stringify(this.props.contentPack, null, 2));
+    const contentPackEncoded = encodeURIComponent(JSON.stringify(this.props.contentPack.toJSON(), null, 2));
```
dennisoelkers: You do not need to call `toJSON` explicitly, `JSON.stringify` is doing it [implicitly](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description).

graylog2-web-interface/src/components/content-packs/ContentPackSelection.test.jsx
dennisoelkers: Why don't you check the state of `state.contentPack` immediately after that expect? To me it looks like there is an unnecessary level of asynchronicity.

graylog2-web-interface/src/components/content-packs/ContentPackPreview.jsx
kmerz: I did not know that! Thank you!

graylog2-web-interface/src/components/content-packs/ContentPackVersions.test.jsx
```diff
@@ -4,39 +4,39 @@ import { mount } from 'enzyme';
...
+    .summary('This is a summary')
+    .vendor('graylog.com')
+    .url('www.graylog.com');
+
   const contentPack = {
-    1: buildPack(1),
-    2: buildPack(2),
-    3: buildPack(3),
-    4: buildPack(4),
+    1: contentPackRev.rev(1).build().toJSON(),
```
dennisoelkers: I think we don't need `toJSON` here either, right?

kmerz: You are absolutely right!!

PR #5316, opened 2018-11-27
Add test button to alert conditions

edmundoa:
The last changes in the alerting API introduced an endpoint to test alert conditions. This PR adds UI for it, so users can test if a certain alert condition would trigger an alert or not at that point of time.

This needs to be cherry-picked into 2.5.

### DIFF
--- a/graylog2-web-interface/src/components/alertconditions/AlertConditionTestModal.jsx
+++ b/graylog2-web-interface/src/components/alertconditions/AlertConditionTestModal.jsx
@@ -73,14 +73,23 @@ class AlertConditionTestModal extends React.Component {
   renderSatisfiedCondition = (testResults) => {
     return (
       <span>
-        <p>Condition was satisfied and an Alert would be triggered.</p>
-        <p><b>Details</b>: {testResults.description}</p>
+        <i className={`fa fa-bell ${style.testResultIcon}`} />
+        <p className={style.testResultText}>Condition was satisfied and an Alert would be triggered.<br />
+          <b>Details</b>: {testResults.description}
+        </p>
       </span>
     );
   };
 
   renderUnsatisfiedCondition = () => {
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/components/alertconditions/AlertConditionTestModal.jsx
```diff
@@ -0,0 +1,123 @@
...
+      </span>
+    );
+  };
+
+  renderSatisfiedCondition = (testResults) => {
+    return (
+      <span>
+        <p>Condition was satisfied and an Alert would be triggered.</p>
+        <p><b>Details</b>: {testResults.description}</p>
+      </span>
```
bernd: The background color for the "condition satisfied" state is the same as for the "condition not satisfied". This makes it hard to differentiate between the two. I had to look twice to see that my condition matched after changing the query because I somehow expected a different visual. :wink: 

image

image

Do you think it would make sense to use different background colors for the two states?

edmundoa: We could use different colours, but usually each colour has a connotation: green is "good", yellow is "meh", red is "bad".
On this case we already use red when the condition throws an error. These two cases are a bit more complicated because none of them really fit into any of the cases above, it's more like testing an extractor (where we also use blue afaik).
I can play a bit with it and see if we can do something about it.

bernd: Okay, makes sense. Let's keep it as it is for now. Thank you!

edmundoa: I added a small icon next to it, to have at least a visual hint.

bernd: Thank you

Issue #5087, opened 2018-09-07
Rotate active write index (200) on non-master


TenGbps:
When am doing a 'Rotate active write index' on web interface, the index not rotation and no error is return.

On graylog log we have :
`org.graylog2.shared.security.RestrictToMasterFilter - Rejected request that is only allowed against master nodes. Returning HTTP 403.`

Please add option in config file to disable this security

* Graylog Version: 2.4.6
* Elasticsearch Version: 5.5.1
* MongoDB Version: 2.6.12
* Operating System: CentOS7
* Browser version: Firefox 60 ESR

jalogisch:
might be related to: https://github.com/Graylog2/graylog2-server/issues/5031

Issue #5031, opened 2018-08-29
Archive Plugin: Rebuild archive catalog did not update the list and the count of archives in the backend


xtruthx:
## Expected Behavior
Trigger the Button "Rebuild" on the "Archive Plugin Webfrontend" should rebuild a catalog from the configured archive backend to  represent the actual state of files on disk in this archive.
Same is expected when using the API with curl for this action

## Current Behavior
If you delete archived data from the archive backend the trigger of the "rebuild catalog action" is reporting an "OK: 200" and is running on the right graylog node the master node.
After finishing the SystemJob there is no change in the count (number of archives) and listed (listed archives in the webfrontend page of the plugin) archives.

Log from the master node:
>  2018-08-28T15:56:54.024+02:00 INFO  [SystemJobManager] Submitted SystemJob <38588080-aaca-11e8-b312-30e17154a810> [org.graylog.plugins.archive.job.BuildCatalogSystemJob]
2018-08-28T15:56:54.025+02:00 INFO  [BuildCatalogJob] Rebuilding backend <Archive on Filesystem>
2018-08-28T15:56:54.354+02:00 INFO  [SystemJobManager] SystemJob <38588080-aaca-11e8-b312-30e17154a810> [org.graylog.plugins.archive.job.BuildCatalogSystemJob] finished in 329ms.

If you compare the node_id used in the API request with the one of the master node they are equal. So the action/systemjob is started and finished on the correct node in the cluster.

## Possible Solution
NONE

## Steps to Reproduce (for bugs)

1. Delete archive dir for a index from disk e.g
2. Trigger rebuild catalog via webfrontend or API
3. Retrive a 200 return code
4. look at the webfrontend to starr on the same list and count of archives as before the action

## Context
We try to establish a valid retention for archives done by Gralyog.

## Your Environment
* Graylog Version:
Version: 2.4.6+ceaa7e4, codename Wildwuchs
JVM: PID 23279, Oracle Corporation 1.8.0_171 on Linux 4.9.0-6-amd64
* Elasticsearch Version: 5.6.9

* MongoDB Version: 3.2.11

* Operating System:
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
[...]

PR #5317, opened 2018-11-28
Add "requestOnMaster()" method to ProxiedResource

### LINKED ISSUE #5120: Extend ProxiedResource to include a "forwardToMaster()" method
The [ProxiedResource#getForAllNodes()](https://github.com/Graylog2/graylog2-server/blob/12b9db93238976520a4f3edd090af7da61c95a8d/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java#L66-L102) method allows us to forward a request to all Graylog nodes in the cluster.

There are several use cases where we need to do the opposite, forward a request to the master node. See https://github.com/Graylog2/graylog2-server/issues/5087 and https://github.com/Graylog2/graylog2-server/issues/5031 for examples. (we also have use for this in enterprise)

To avoid writing this over and over again in several subsystems, I propose the addition of a `forwardToMaster()` (or some similar name) method in `ProxiedResource`.  That method takes a retrofit client class and and an interface provider just like `getForAllNodes()` does. But then it checks the database for the current master node and forwards the request to that node. (and also returns the response to the requesting client)

bernd:
This can be used to proxy API requests to the active master node. It
works the same way as the existing "getForAllNodes()" method by using
retrofit remote interface definitions for the HTTP requests to the
master.

Prerequisite to be able to work on #5031, #5087 and some issues in Graylog Enterprise.

Fixes #5120


graylog2-server/src/test/java/org/graylog2/shared/rest/resources/ProxiedResourceTest.java
```diff
@@ -0,0 +1,45 @@
...
+import org.graylog2.shared.rest.resources.ProxiedResource.MasterResponse;
+import org.junit.Test;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class ProxiedResourceTest {
+    @Test
+    public void masterResponse() {
```
dennisoelkers: What are we testing here? I does not make sense to test AutoValue. If we are testing `create`, this implies that we are unsure about the parameter type/order of the factory method, which means that we should just use a builder instead.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -112,4 +114,82 @@ protected ProxiedResource(@Context HttpHeaders httpHeaders,
...
+         * @return HTTP status code
+         */
+        public abstract int code();
+
+        /**
+         * Returns the typed response object if the request was successful. Otherwise it returns {@code null}.
+         *
+         * @return typed response object or {@code null}
+         */
+        @Nullable
```
dennisoelkers: Instead of making it `Nullable`, we should use an `Optional` here.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -112,4 +114,82 @@ protected ProxiedResource(@Context HttpHeaders httpHeaders,
...
+        public abstract ResponseType entity();
+
+        /**
+         * Returns the error response if the request wasn't successful. Otherwise it returns {@code null}.
+         * @return error response or {@code null}
+         */
+        @Nullable
```
dennisoelkers: Instead of making it Nullable, we should use an Optional here.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -112,4 +114,82 @@ protected ProxiedResource(@Context HttpHeaders httpHeaders,
...
+
+        final RemoteInterfaceType remoteInterfaceType = remoteInterfaceProvider.apply(masterNode.getNodeId())
+                .orElseThrow(() -> new IllegalStateException("Master node " + masterNode.getNodeId() + " not found"));
+
+        final Call<RemoteCallResponseType> call = remoteInterfaceFunction.apply(remoteInterfaceType);
+        final Response<RemoteCallResponseType> response = call.execute();
+
+        final byte[] errorBody = response.errorBody() == null ? null : response.errorBody().bytes();
+
+        return MasterResponse.create(response.isSuccessful(), response.code(), response.body(), errorBody);
```
dennisoelkers: I do not really see the point of using the `MasterResponse` class here. Do we expect the requests to come up with a negative response regularly? If not, we should just throw exceptions wrapping the status. This way we also force the consumer to handle those instead of relying that the response object is parsed.

graylog2-server/src/test/java/org/graylog2/shared/rest/resources/ProxiedResourceTest.java
bernd: We are testing that the `#body()` method returns the correct value depending on the value of `#entity()`.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
bernd: We are using the `requestOnMaster()` method mostly in cases where we want to transparently proxy the response of the master request to the calling client. My first implementation returned a `javax.ws.rs.core.Response` here but I found that to be not flexible enough when using it because the caller might want to access the typed response object. But I also didn't want to return the `retrofit2.Response` here because the callers would have to write some code all the time to build a `Response` depending on the success status of the request.

This is how you can use the `MasterResponse` in a resource:

```java
final MasterResponse<T> response = requestOnMaster(...);

return Response.status(response.code()).entity(response.entity()).build();
```
The `MasterResponse#entity()` method returns the correct value depending on the success status of the response.

If we would return the `retrofit2.Response` here, you would have to check if we are dealing with an error or a success to select the correct value for the `Response#entity()`.

I didn't want to throw an exception here because I want to be able to return the error response from the master as it is without wrapping it in some exception. I guess we could also do that with a custom exception implementation but that would also require us to write some code.

So I decided to use a small `MasterResponse` object.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
bernd: :heavy_check_mark:

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
bernd: :heavy_check_mark:

PR #5218, opened 2018-10-16
Add throttling to the AWS Flow Logs/Logs inputs

danotorrey:
This change adds a few methods to ThrottlableTransport needed to support throttling in the AWS Log/FlowLogs inputs.

PR #5321, opened 2018-11-28
Add throttling to the AWS Flow Logs/Logs inputs (#5218)

bernd:
* Add blockUntilUnthrottled method with timeout

This allows an input to block while throttled during a timeout period. If the block latch is not cleared in the timeout period, the latch will expire and allow execution to proceed.

* Add throttled state change callback

* Document return for method ThrottleableTransport.blockUntilUnthrottled()

* Fixed documentation typo

(cherry picked from commit 72c6126d29ecb366a3f02a854362ec51c2a31c7c)


danotorrey:
LGTM 👍

PR #5357, opened 2018-12-03
Keep Content Pack entities when creating a new version

kmerz:
## Description
When selecting the entities for a new version of a content pack, we now offer the previous entities of the
content pack. That way a user can modify the original content pack without having the original entities on his system.

When creating a new version of the content pack the user sees now icons behind the entities to distinguish between entities from the server (blue) and original entities from the content pack.

## Motivation and Context
Before when editing a content pack you needed the original entities on your system other wise the
selection would not work, since Graylog could not find them in the database. 

## How Has This Been Tested?
Created a new content pack, and created a new version for it.

graylog - content packs 1

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/ExpandableListItem.css
+++ b/graylog2-web-interface/src/components/common/ExpandableListItem.css
@@ -1,4 +1,4 @@
-:local(.listItem) {
+:local(.listItemPadding) {
     padding: 10px 5px;
 }
 
--- a/graylog2-web-interface/src/components/common/ExpandableListItem.jsx
+++ b/graylog2-web-interface/src/components/common/ExpandableListItem.jsx
@@ -35,6 +35,8 @@ class ExpandableListItem extends React.Component {
       PropTypes.element,
       PropTypes.arrayOf(PropTypes.element),
     ]),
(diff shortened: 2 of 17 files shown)


bernd:
@kmerz What do you think about moving the icon in front of the name and only coloring the ones from the content pack?
image

Putting the icon after the title makes it look pretty awkward to me, sorry. :wink: Wrapping the icons in `<Label />` gives a bit too much emphasis on the icons in my opinion. Also using a transparent background color and only change `color` for the icons makes the icons a bit clearer.

I think we also need a legend to explain the icons and/or hover text to make sure users understand the meaning of the icons.

kmerz:
> @kmerz What do you think about moving the icon in front of the name and only coloring the ones from the content pack?
> image
> 
> Putting the icon after the title makes it look pretty awkward to me, sorry.  Wrapping the icons in `<Label />` gives a bit too much emphasis on the icons in my opinion. Also using a transparent background color and only change `color` for the icons makes the icons a bit clearer.
> 
> I think we also need a legend to explain the icons and/or hover text to make sure users understand the meaning of the icons.

Excellent suggestions! Thank you very much!

PR #5359, opened 2018-12-04
clarify definition of `web_endpoint_uri`

eightnoneone:
Definition provided in context of web endpoint is vague as REST is not clarified.

We spent way too long chasing `rest_transport_uri` and `rest_listen_uri` as solutions to what controlled what was promoted into the browser for REST calls. Searching "REST" in the config misses `web_endpoint_uri`.

## Types of changes

## Checklist:

### DIFF
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -131,7 +131,7 @@ rest_listen_uri = http://127.0.0.1:9000/api/
 # for the application.context configuration parameter in pre-2.0 versions of the Graylog web interface.
 #web_listen_uri = http://127.0.0.1:9000/
 
-# Web interface endpoint URI. This setting can be overriden on a per-request basis with the X-Graylog-Server-URL header.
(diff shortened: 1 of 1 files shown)


misc/graylog.conf
```diff
@@ -131,7 +131,7 @@ rest_listen_uri = http://127.0.0.1:9000/api/
 # for the application.context configuration parameter in pre-2.0 versions of the Graylog web interface.
 #web_listen_uri = http://127.0.0.1:9000/
 
-# Web interface endpoint URI. This setting can be overriden on a per-request basis with the X-Graylog-Server-URL header.
+# Web interface REST endpoint URI. Used by browsers to receive steams. This setting can be overriden on a per-request basis with the X-Graylog-Server-URL header.
```
dennisoelkers: It is not only `Used by browsers to receive steams.`, it does way more. I would recommend changing this to something like `Must be reachable from the user's browser.`

eightnoneone: Good point. I committed the change.

dennisoelkers:
Thanks for the contribution, @eightnoneone!

PR #5376, opened 2018-12-05
Sidecar migration helper

mariussturm:
## Description

This adds an additional workflow that allows a user to fetch rendered configurations from a remote host running a 0.x Sidecar into the Graylog server. From there the user can create new configurations for 1.x Sidecars/Collectors

## Motivation and Context

A fully automated migration between both systems is not possible so we want to make the process as smooth as possible.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationForm.jsx
```diff
@@ -147,6 +148,14 @@ const ConfigurationForm = createReactClass({
     this.setState({ formData: nextFormData });
   },
 
+  _onTemplateImport(nextTemplate) {
+    const nextFormData = lodash.cloneDeep(this.state.formData);
+    if (!nextFormData.template || window.confirm('Do you want to overwrite your current work with this Configuration?')) {
+      nextFormData.template = nextTemplate;
+      this.setState({ formData: nextFormData });
```
edmundoa: This does not reload the template validations, so we should replace this two lines with:

```js
this._onTemplateChange(nextTemplate);
```

To reproduce the issue, use an empty source code editor and then import a configuration. The field will still be red because it thinks it is empty.

The same issue happens when adding the default collector configuration, so you could also change it there if you want to 🙂

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+          <Button bsStyle="info" bsSize="xsmall" onClick={e => this._onApplyButton(upload.rendered_configuration, e)}>
+            Apply
+          </Button>
+        </td>
+      </tr>
+    );
+  }
+
+  render() {
+    if (this._isLoading()) {
```
edmundoa: As most of the modal is not changing, we should only change what we render in the `Modal.Body`.

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+            <Button type="button" onClick={this.hide}>Close</Button>
+          </Modal.Footer>
+        </BootstrapModalWrapper>
+      );
+    }
+
+    return (
+      <BootstrapModalWrapper ref={(c) => { this.uploadsModal = c; }}>
+        <Modal.Header closeButton>
+          <Modal.Title><span>Configuration Imports</span></Modal.Title>
```
edmundoa: We should reflect in the title that these are old configurations from the old collector plugin.

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+        </Modal.Header>
+        <Modal.Body>
```
edmundoa: I think it would help to add a few words on how the user should do after they apply a configuration. At least we should mention that they need to review them and adapt them, maybe using the new variables to replace some hardcoded IDs.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/ImportService.java
```diff
@@ -0,0 +1,87 @@
...
+import org.mongojack.JacksonDBCollection;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ImportService extends PaginatedDbService<CollectorUpload> {
+    private static final String COLLECTION_NAME = "collector_uploads";
+    private final JacksonDBCollection<CollectorUpload, ObjectId> dbCollection;
```
edmundoa: File can be converted to local variable in the constructor, says IntelliJ 😄

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationResource.java
```diff
@@ -124,6 +130,29 @@ public ConfigurationListResponse listConfigurations(@ApiParam(name = "page") @Qu
...
+                                                   @ApiParam(name = "per_page") @QueryParam("per_page") @DefaultValue("10") int perPage,
+                                                   @ApiParam(name = "query") @QueryParam("query") @DefaultValue("") String query,
+                                                   @ApiParam(name = "sort",
+                                                           value = "The field to sort the result on",
+                                                           required = true,
+                                                           allowableValues = "id,collector_id, created")
+                                                   @DefaultValue(Configuration.FIELD_NAME) @QueryParam("sort") String sort,
+                                                   @ApiParam(name = "order", value = "The sort direction", allowableValues = "asc, desc")
+                                                   @DefaultValue("asc") @QueryParam("order") String order) {
+        final SearchQuery searchQuery = searchQueryParser.parse(query);
```
edmundoa: We don't use any queries and this parser is not made for that collection. We also don't use sorting or ordering. I think we should remove all things we will never use from here, even if this method will most likely go away in some months.

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+      <tr key={upload.id}>
+        <td>
+          <OverlayTrigger placement="top" overlay={tooltip} rootClose>
+            <span>{upload.node_id}</span>
+          </OverlayTrigger>
+        </td>
+        <td>{upload.collector_name}</td>
+        <td><Timestamp dateTime={upload.created} format={'YYYY-MM-DD HH:mm:ss z'} /></td>
+        <td>
+          <Button bsStyle="info" bsSize="xsmall" onClick={e => this._onApplyButton(upload.rendered_configuration, e)}>
```
mpfz0r: we don't need to pass the event to _onApplyButton

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,154 @@
...
+      return (
+        <BootstrapModalWrapper ref={(c) => { this.uploadsModal = c; }}>
+          <Modal.Header closeButton>
+            <Modal.Title><span>Configuration Imports</span></Modal.Title>
+          </Modal.Header>
+          <Modal.Body>
+            <Alert bsStyle="info">
+              <i className="fa fa-info-circle" />&nbsp;
+              There are no configuration uploads available. Please go to<br />
+              <strong>System-&gt;Collectors-&gt;Details-&gt;Import Configuration</strong><br />
```
mpfz0r: The menu is called `Collectors (legacy)` now. We might also provide a link :)

graylog2-web-interface/src/components/sidecars/configuration-forms/ImportsViewModal.jsx
```diff
@@ -0,0 +1,145 @@
...
+    const pageSize = this.PAGE_SIZE;
+    const { uploads, totalUploads } = this.state;
+    const formattedUploads = uploads.map(upload => (this._formatUpload(upload)));
+
+    if (totalUploads === 0) {
+      return (
+        <Alert bsStyle="info">
+          <i className="fa fa-info-circle" />&nbsp;
+          There are no configuration uploads available. Please go to<br />
+          <strong>System-&gt;Collectors (legacy)-&gt;Details-&gt;Import Configuration</strong><br />
```
edmundoa: This line is hard to read, as the whole line is bold and there are almost no spaces between words. We should add spaces between arrows and words, and maybe make it cursive instead.

edmundoa: I think we should also remove those `br`s around this line. In the end we are already highlighting the way to get the configurations imported, so let's let the browser to break the line if it should.

edmundoa:
Tests are failing, but it is not related to this PR.

PR #5389, opened 2018-12-10
 Remove nonexistent "consistency" parameter from index/bulk requests

bernd:
The parameter has been removed from Elasticsearch about 2.5 years ago
before the release of 5.0 and requests have it will fail.

This hasn't been noticed so far because it only happens when the
Indices#move() method runs. That method is only called by the
FixDeflectorByMoveJob periodical. The job doesn't run that often,
apparently.

Also the only test that is using it is the MessagesIT test and that is
using "assumeTrue()" so it doesn't fail but gets skipped.


mpfz0r:
LGTM :+1: 

Just for reference: `consistency=one` is now `wait_for_active_shards=1` which is the default:
https://www.elastic.co/guide/en/elasticsearch/reference/5.0/docs-index_.html#index-wait-for-active-shards

PR #5409, opened 2018-12-14
Add NXlog collectors and add new documentation reference

mpfz0r:
Add NXlog default collectors and templates for Linux and windows

Replace inline filebeat documentation with external links
We don't want to maintain the documentation of various collectors.
Link to the official documentation instead.


graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -104,13 +104,154 @@ public void upgrade() {
                         "   - name: Security"
         );
         ensureCollector(
-                "nxlog",
+                "nxlog-linux",
```
mariussturm: Unfortunately this resolves UI-wise to `nxlog-linux on Linux` when you select the collector on the configuration page.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -104,13 +104,154 @@ public void upgrade() {
...
+                        "<Extension syslogExt>\n" +
+                        "  Module xm_syslog\n" +
+                        "</Extension>\n" +
+                        "\n" +
+                        "User nxlog\n" +
+                        "Group nxlog\n" +
+                        "\n" +
+                        "Moduledir /usr/lib/nxlog/modules\n" +
+                        "CacheDir /var/spool/collector-sidecar/nxlog\n" +
+                        "PidFile /var/run/graylog-sidecar/nxlog.pid\n" +
```
mariussturm: Do we want to use Sidecar specific paths for collectors or maybe rely completely on default paths?

graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -104,13 +104,154 @@ public void upgrade() {
...
+                        "\tReadFromLast True\n" +
+                        "\tRecursive False\n" +
+                        "\tRenameCheck False\n" +
+                        "\tExec $FileName = file_name(); # Send file name with each message\n" +
+                        "</Input>\n" +
+                        "\n" +
+                        "<Input syslog-udp>\n" +
+                        "\tModule im_udp\n" +
+                        "\tHost 127.0.0.1\n" +
+                        "\tPort 514\n" +
```
mariussturm: I wouldn't start a syslog input by default. We should either use an unprivileged port for this or comment the input if we really want to show how this works.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -104,13 +104,154 @@ public void upgrade() {
...
+                        "  Path file => gelf\n" +
+                        "</Route>\n" +
+                        "<Route route-2>\n" +
+                        "  Path syslog-udp => gelf\n" +
+                        "</Route>\n" +
+                        "\n" +
+                        "\n"
+        );
+        ensureCollector(
+                "nxlog-windows",
```
mariussturm: same as above

graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
mpfz0r: I took these from the config generated by the old sidecar.
Never change a running system ;-)

graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
mpfz0r: Ack. I'll comment this out

PR #5414, opened 2018-12-14
Rename requires to constraints and move it from top level to entity

### LINKED ISSUE #5335: Edit contraints with content pack
## Expected Behavior
After editing the entity selection of a content pack the resulting constraints should match
the selection.

## Current Behavior
If you remove a entity from a content pack there is now way to tell if a constraint need to be removed
since the there is no relation between constraint and entity.

## Possible Solution
1. Every entity stores his constraint
2. The constraint gets a list of entities which depend on the constraint

kmerz:
## Description
The is PR will move the constraints from content pack top level to each entity.
 
## Motivation and Context
When editing a content pack we did not have knowledge which constraint belongs
to which entity. When editing the content pack entities we therefore could not
edit the constraints accordingly to the add or removed entities.

## How Has This Been Tested?
Created multiple versions of a content pack
installed and uninstalled it.

## Types of changes

## Checklist:

Fixes #5335


graylog2-server/src/main/java/org/graylog2/contentpacks/model/ContentPackV1.java
```diff
@@ -84,10 +83,18 @@
 
     public static Builder builder() {
         return new AutoValue_ContentPackV1.Builder()
-                .requires(ImmutableSet.of())
                 .parameters(ImmutableSet.of());
     }
 
+    public Set<Constraint> constraints() {
+        return entities().asList().stream()
```
bernd: What's the reason for using `.asList()` here? Doesn't the following just work as well?
```suggestion
        return entities().stream()
```

graylog2-server/src/main/java/org/graylog2/rest/resources/system/contentpacks/CatalogResource.java
```diff
@@ -25,6 +26,7 @@
 import org.graylog2.audit.jersey.NoAuditEvent;
 import org.graylog2.contentpacks.ContentPackService;
 import org.graylog2.contentpacks.model.entities.EntitiesWithConstraints;
```
bernd: It looks like the `EntitiesWithConstraints` class is only referenced in some unused imports in `ContentPackService` and `CatalogResource` and can be removed.

graylog2-server/src/main/java/org/graylog2/contentpacks/facades/EntityFacade.java
```diff
@@ -20,6 +20,8 @@
 import com.google.common.graph.GraphBuilder;
 import com.google.common.graph.ImmutableGraph;
 import com.google.common.graph.MutableGraph;
+import org.graylog2.contentpacks.model.constraints.Constraint;
+import org.graylog2.contentpacks.model.constraints.GraylogVersionConstraint;
```
bernd: These are unused imports. Please run "Optimize Imports" on the complete "contentpacks" package in IntelliJ to remove all of them. :smiley:

PR #5415, opened 2018-12-17
Avoid failing JSON deser. if ext. type is present, but prop is not.

dennisoelkers:
This change allows deserializing classes which fulfill the following:

  * They contain a property `V` which is subtyped
  * They contain an additional property `T` defining the type of the
property `V` (and maybe the types of additional properties or of the
class itself)
  * `V` is optional, so it might be present or not
  * `T` is always present

The important point is the optionality of `V`. Jackson assumes that if a property that is used as an external property with `@JsonTypeInfo` is present, that the property annotated with `@JsonTypeInfo` must be present as well. This is not always the case, especially when `T` is used to indicate of the class itself and `V`'s type depends on `T` but the property itself is optional.

This change is disabling Jackson's [`FAIL_ON_MISSING_CREATOR_PROPERTIES`](https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/DeserializationFeature.html#FAIL_ON_MISSING_CREATOR_PROPERTIES) feature, which allows us to use the forementioned constellation.

All current usages of `@JsonTypeInfo` were investigated and the change is safe as we do not even have any usages of `@JsonTypeInfo` for properties of a class, it is used only to annotate complete classes.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ObjectMapperProvider.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ObjectMapperProvider.java
@@ -71,6 +71,7 @@ public class ObjectMapperProvider implements Provider<ObjectMapper> {
         this.objectMapper = mapper
                 .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
                 .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE)
+                .disable(DeserializationFeature.FAIL_ON_MISSING_EXTERNAL_TYPE_ID_PROPERTY)
                 .setPropertyNamingStrategy(new PropertyNamingStrategy.SnakeCaseStrategy())
                 .setSubtypeResolver(subtypeResolver)
                 .setTypeFactory(typeFactory)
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/webpack/vendor-module-ids.json
```diff
@@ -889,7 +889,10 @@
       "node_modules/react/cjs/react.production.min.js": 24,
       "node_modules/moment/locale sync /^\\.\\/.*$/": 25,
       "node_modules/moment-timezone/node_modules/moment/locale sync /^\\.\\/.*$/": 32,
-      "node_modules/moment/locale/mn.js": 13
+      "node_modules/moment/locale/mn.js": 13,
+      "node_modules/hoist-non-react-statics/dist/hoist-non-react-statics.cjs.js": 18,
+      "node_modules/scheduler/index.js": 20,
+      "node_modules/scheduler/cjs/scheduler.production.min.js": 29
```
bernd: @dennisoelkers The change in this file is unrelated to this PR, no?

dennisoelkers: Oh yeah, that slipped in accidentally.

dennisoelkers:
Thanks, @bernd!

Issue #5408, opened 2018-12-14
Password Leak through System/Inputs


mbirch-ca:
When loading the System Inputs, the returned data includes passwords in plain text making them visible to any user with access to the system.

## Expected Behavior
It should not be possible for any user (reader or admin) to see any password.

## Current Behavior
When loading the web interface, the browser network tools can see passwords in plain text.  While the browser displays '********', the JSON includes the password

The "inputs" screen is a default permission for a *Reader* making these passwords visible to any user.

## Possible Solution
A similar issue was reported and resolved relating to LDAP credentials being visible in #3763

To protect passwords, it probably makes more sense not to the send any passwords when viewing or modifying settings.  Keeping this stored on the server would ensure its secure and only update when an admin makes a change.

## Steps to Reproduce (for bugs)
1.  From your web browser, enable the network monitor
2.  Open the System > Inputs page
3.  From the browser network monitor, search for "inputs"
4.  Display the response and look in the JSON object for the passwords

## Context
Information leakage of passwords would be seen as a security concern depending upon what it can access.  (Many standards require protection of log data.)

Depending upon the type of input, the security exposure can be increased if a message bus (like Kafka) is used and contains other data.  Given that the screen also provides information about the TLS certificate store, there is additional information exposure if the certificate store could be retrieved.  (Which is easier based on having the location and password.)

## Your Environment

* Graylog Version:   2.4.3
* Operating System:  Windows 7/10
* Browser version:    Firefox (latest),  Chrome (latest)

dennisoelkers:
Thanks for reporting this, @mbirch-ca! We will address this asap.

mbirch-ca:
@kmerz , do you know if this will be resolved in the next release?  (Maybe in the 2.5 branch if 3.0 is not going to be released?)

Was looking for a method to mitigate this risk but found I was blocked given the potential impact of issue #5420 as the default permission of Reader provides access to System/Inputs.

edmundoa:
@mbirch-ca We plan to release 2.5.2 including a fix for this issue in the next days, so please stay tuned!

edmundoa:
Refs #5550

PR #5432, opened 2018-12-20
Mask password fields of inputs returned by the REST API.

dennisoelkers:
## Motivation and Context

Before this change, input details returned by the REST API would contain
all configuration fields without any modification. This implies that
password fields are also contained using their original value, showing
configured password for inputs in clear text.

For users which are not allowed to edit an input, this change now iterates over configuration fields checking for the presence of password fields and replace their content with `<password set>` instead of the original value if they are not empty.

For users which possess the required permission to edit an input, the original value is returned as it is used to set the value of the password field and would be used for updating the input if left unchanged.

This is a middle ground between a quick fix improving the situation (not everybody who is able to view inputs can also see passwords) and a proper fix that would require more work and attention to consistency, which will happen in a future version.

Refs #5408 and partially fixes it.

## Types of changes

## Checklist:


dennisoelkers:
@kmerz: Can you also check again and merge if you are happy?

PR #5448, opened 2019-01-03
Add Etag caching for Sidecar UpdateRegistration results

mpfz0r:
The response to the registration PUT request contains
configuration assignments, which can be cached as well.

This requires us to invalidate the sidecar Etag cache in more
situations:
 - Configuration assignments
 - User triggered collector actions (stop, start, restart)


mariussturm:
* We also need to invalidate the cache if the runtime configuration in System->Configuration->Collector gets an update.

In general we want to cache all API calls but because of technical limitations of the current cache implementation we will postpone this PR till we have a finer grained cache to prevent side effects.

bernd:
@mpfz0r @mariussturm What's the status on this? Should we merge it or can this be closed? Thank you!

mpfz0r:
Like Marius said, "postponed" :-)

mpfz0r:
fixed with another PR

PR #5462, opened 2019-01-08
Add the CONTAINS rule as one not requiring the field to be present

JeremyEinfeld:
## Description
The CONTAINS rule evaluates the match to false when the rule is inverted (I.E. "field does not contain value") and the field is missing (which can be seen in [tests](https://github.com/Graylog2/graylog2-server/blob/0fb0b2ee3de22a1258abe5787e19e79bc9e84eff/graylog2-server/src/test/java/org/graylog2/streams/matchers/ContainsMatcherTest.java#L76-L83)), but because the rule is not in the [`ruleTypesNotNeedingFieldPresence` list](https://github.com/Graylog2/graylog2-server/blob/0fb0b2ee3de22a1258abe5787e19e79bc9e84eff/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java#L57) the rule is [not tested when the field doesn't exist](https://github.com/Graylog2/graylog2-server/blob/0fb0b2ee3de22a1258abe5787e19e79bc9e84eff/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java#L176-L179).

## Motivation and Context
Having an error stream from which we wanted to exclude some noisy errors, we added a rule "Field Exception must not contain Server cannot set status after HTTP headers have been sent". Following that change, messages that did not contain the Exception field were no longer being added to the stream.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -54,7 +54,7 @@ import java.util.concurrent.TimeUnit;
 public class StreamRouterEngine {
     private static final Logger LOG = LoggerFactory.getLogger(StreamRouterEngine.class);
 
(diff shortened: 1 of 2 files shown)


bernd:
Thank you for the PR!

We have to check if this has any implications for backwards compatibility.

PR #5463, opened 2019-01-08
Update plugin list in System -> Enterprise page

### LINKED ISSUE #4600: Adapt enterprise page to changes in plugins
As we are merging plugins, we should ensure the status displayed in the enterprise page shows the right plugins and their status in the system. As this is an ongoing process, I think it's better to leave this issue open and fix the issue once we are closer to the release.

## Expected Behaviour
Only available plugins appear in enterprise page.

## Current Behaviour
Some merged plugins appear in the enterprise page, even if they are not independent any more.

1. Go to _System -> Enterprise_
2. See that some plugins in there are not independent now and their status is incorrect

## Your Environment

* Graylog Version: Graylog 3.0.0-SNAPSHOT (a3da5916cbb8ac5924933336560de4eaec246322)

dennisoelkers:
The plugin list in the _System -> Enterprise_ page was outdated, since we now ship all enterprise modules in a single plugin. This PR takes care of correcting that and fixes a few linter issues we had in there.

Fixes #4600

### DIFF
--- a/graylog2-web-interface/src/components/enterprise/PluginList.css
+++ b/graylog2-web-interface/src/components/enterprise/PluginList.css
@@ -1,4 +1,4 @@
-.enterprise-plugins {
+:local(.enterprisePlugins) {
   margin-left: 0;
   padding-left: 0;
 }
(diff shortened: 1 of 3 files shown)


edmundoa:
Refs #5459

PR #5495, opened 2019-01-11
Improve Sidecar API validations

mariussturm:
Sidecar validates most entities on the server side. This PR cleans up the code there and brings the API calls in a consistent state. It also adds validations for empty fields. Collector names now are unique per operating systems. This allows to have a `nxlog` collector for linux as well as for windows as an example.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/CollectorResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/CollectorResource.java
@@ -29,25 +29,46 @@ import org.graylog.plugins.sidecar.rest.models.Collector;
 import org.graylog.plugins.sidecar.rest.models.CollectorSummary;
 import org.graylog.plugins.sidecar.rest.responses.CollectorListResponse;
 import org.graylog.plugins.sidecar.rest.responses.CollectorSummaryResponse;
-import org.graylog.plugins.sidecar.rest.responses.ValidationResponse;
 import org.graylog.plugins.sidecar.services.CollectorService;
 import org.graylog.plugins.sidecar.services.ConfigurationService;
 import org.graylog.plugins.sidecar.services.EtagService;
 import org.graylog2.audit.jersey.AuditEvent;
+import org.graylog2.audit.jersey.NoAuditEvent;
 import org.graylog2.database.PaginatedList;
 import org.graylog2.plugin.database.ValidationException;
 import org.graylog2.plugin.rest.PluginRestResource;
+import org.graylog2.plugin.rest.ValidationResult;
 import org.graylog2.search.SearchQuery;
 import org.graylog2.search.SearchQueryField;
 import org.graylog2.search.SearchQueryParser;
 import org.graylog2.shared.rest.resources.RestResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
-import javax.ws.rs.*;
-import javax.ws.rs.core.*;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.DELETE;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.POST;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.CacheControl;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.EntityTag;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
 import java.util.List;
+import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 @Api(value = "Sidecar/Collectors", description = "Manage collectors")
@@ -56,6 +77,11 @@ import java.util.stream.Collectors;
 @Produces(MediaType.APPLICATION_JSON)
 @RequiresAuthentication
 public class CollectorResource extends RestResource implements PluginRestResource {
+    private static final Logger LOG = LoggerFactory.getLogger(CollectorResource.class);
+
+    private static final Pattern VALID_COLLECTOR_NAME_PATTERN = Pattern.compile("^[A-Za-z0-9_.-]+$");
+    private static final Pattern VALID_PATH_PATTERN = Pattern.compile("^[^;:*?\"<>|&]+$");
+//everything but ; : * ? " < > | &
     private final CollectorService collectorService;
     private final ConfigurationService configurationService;
     private final EtagService etagService;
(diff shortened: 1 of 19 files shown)


graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/CollectorResource.java
```diff
@@ -236,31 +257,94 @@ public Response deleteCollector(@ApiParam(name = "id", required = true)
...
-            collector = collectorService.findByNameExcludeId(name, id);
+            collectorOptional = Optional.ofNullable(collectorService.findByName(toValidate.name()));
         }
-        if (collector == null) {
-            return ValidationResponse.create(false, null);
+        if (collectorOptional.isPresent()) {
+            collector = collectorOptional.get();
+            if (!collector.id().equals(toValidate.id())) {
+                // a collector exists with a different id, so the name is already in use, fail validation
+                validation.addError("name", "Collector with name \"" + toValidate.name() + "\" already exists");
```
edmundoa: As the combination of name and OS is what we validate, we should mention that the name exists for that OS.

graylog2-web-interface/src/components/sidecars/configuration-forms/CollectorForm.jsx
```diff
@@ -135,55 +155,58 @@ const CollectorForm = createReactClass({
...
             <Input type="text"
                    id="validationParameters"
                    label="Parameters for Configuration Validation"
                    onChange={this._onInputChange('validation_parameters')}
                    help={<span>Parameters that validate the configuration file. <strong> %s will be replaced by the path to the configuration file.</strong></span>}
-                   value={validationParameters} />
+                   value={validationParameters || ''} />
 
             <FormGroup controlId="defaultTemplate">
               <ControlLabel>Default Template</ControlLabel>
```
edmundoa: The default template field does not show any validation state, but it is currently required.

I have found the same that @mpfz0r did, when filling the collector form, the create button is disabled until I fill the default template. From the network requests I can see that the backend returns:

```JSON
{"errors":{"default_template":["Collector default template cannot be empty."]},"failed":true}
```

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationResource.java
```diff
@@ -332,6 +339,50 @@ public Response deleteConfiguration(@ApiParam(name = "id", required = true)
...
+
+        if (toValidate.collectorId().isEmpty()) {
+            validation.addError("collector_id", "Associated collector ID cannot be empty.");
+        }
+
+        if (toValidate.color().isEmpty()) {
+            validation.addError("color", "Collector color cannot be empty.");
+        }
+
+        if (toValidate.template().isEmpty()) {
```
edmundoa: This does not trim the input, so adding a space to it would make the validation fail. I guess that should be done in all required fields.

graylog2-web-interface/src/components/sidecars/configuration-forms/CollectorForm.jsx
```diff
@@ -135,55 +155,58 @@ const CollectorForm = createReactClass({
...
                    label="Executable Path"
                    onChange={this._onInputChange('executable_path')}
-                   help="Path to the collector executable"
-                   value={this.state.formData.executable_path}
+                   bsStyle={this._validationState('executable_path')}
+                   help={this._formatValidationMessage('executable_path', 'Path to the collector executable')}
+                   value={this.state.formData.executable_path || ''}
                    required />
 
             <Input type="text"
```
edmundoa: We should mark this field as optional, maybe by changing the label into: 
```<span>Executable Parameters <small className="text-muted">(Optional)</small></span>```

File: graylog2-web-interface/src/components/sidecars/configuration-forms/CollectorForm.jsx [hunk 5]
```diff
@@ -135,55 +155,58 @@ const CollectorForm = createReactClass({
...
                    id="executeParameters"
                    label="Execute Parameters"
                    onChange={this._onInputChange('execute_parameters')}
                    help={<span>Parameters the collector is started with.<strong> %s will be replaced by the path to the configuration file.</strong></span>}
-                   value={executeParameters} />
+                   value={executeParameters || ''} />
 
             <Input type="text"
```
edmundoa: We should mark this field as optional.

graylog2-web-interface/src/components/sidecars/configurations/CopyConfigurationModal.jsx
```diff
@@ -46,16 +46,20 @@ class CopyConfigurationModal extends React.Component {
...
     }
   };
 
   _changeName = (event) => {
     const nextName = event.target.value;
     this.setState({ name: nextName });
-    this.props.validateConfiguration(nextName).then(validation => (
-      this.setState({ error: validation.error, error_message: validation.error_message })
-    ));
+    this.props.validateConfiguration({ name: nextName }).then((validation) => {
```
edmundoa: Trying to clone a valid configuration fails with:

```JSON
{"errors":{"template":["Collector template cannot be empty."],"color":["Collector color cannot be empty."],"collector_id":["Associated collector ID cannot be empty."]},"failed":true}
```

I guess the error is here, as we only pass the name into the validation function (which defaults every property into an empty string).

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationResource.java
mariussturm: We can't use trim here as a single whitespace is used as default value for incomplete object validations.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/resources/ConfigurationVariableResource.java
```diff
@@ -100,7 +105,8 @@ public Response createConfigurationVariable(@ApiParam(name = "JSON body", requir
                                              @Valid @NotNull ConfigurationVariable request) {
         ValidationResult validationResult = validateConfigurationVariableHelper(request);
         if (validationResult.failed()) {
-            return Response.status(Response.Status.BAD_REQUEST).entity(validationResult).build();
```
mpfz0r: The former code returned a structured API error like:
```

{
  "errors": {
    "name": [
      "A variable with that name already exists."
    ]
  },
  "failed": true
}
```
And it was identical to the response from the validation.

Now this would return:
```
{
  "type": "ApiError",
  "message": "Cannot update variable. {name=[A variable with that name already exists.]}"
}
```

mariussturm: This is based on feedback from @edmundoa, the point was that an Exception makes it more clear that something went wrong. But I can also see your point of a better structured API error. I don't have a strong opinion, we should just have the same type of error under the `/sidecars/*` API namespace.

edmundoa: I thought we could pass `validationResults` to the exception to get that into the response body. If that's not possible, we can leave it as it was.

bernd: Throwing an exception or returning a `Response.status(Response.Status.BAD_REQUEST)` doesn't make a huge difference because the exception mapper will return it into a 400 or 500 anyway.

So I think we can use `Response.status(Response.Status.BAD_REQUEST)` with the proper `validationResults` to get a structured response.

PR #5533, opened 2019-01-16
Run default collector migration with nodeOperatingSystem

mpfz0r:
Collectors uniqeness is now checked by combining the name
with the nodeOperatingSystem.
Change the migration to allow two "nxlog" collectors:
One for linux and one for windows.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
@@ -279,14 +279,14 @@ public class V20180212165000_AddDefaultCollectors extends Migration {
                                    String defaultTemplate) {
         Collector collector = null;
         try {
-            collector = collectorService.findByName(collectorName);
+            collector = collectorService.findByNameAndOs(collectorName, nodeOperatingSystem);
             if (collector == null) {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog/plugins/sidecar/migrations/V20180212165000_AddDefaultCollectors.java
```diff
@@ -279,14 +279,14 @@ private String ensureCollector(String collectorName,
...
         Collector collector = null;
         try {
-            collector = collectorService.findByName(collectorName);
+            collector = collectorService.findByNameAndOs(collectorName, nodeOperatingSystem);
             if (collector == null) {
-                final String msg = "Couldn't find collector '" + collectorName + "' fixing it.";
-                LOG.error(msg);
+                final String msg = "Couldn't find collector '{} on {}' fixing it.";
+                LOG.error(msg, collectorName, nodeOperatingSystem);
                 throw new IllegalArgumentException(msg);
```
bernd: The `msg` here is just a log template and doesn't have the `collectorName` and `nodeOperatingSystem` in the exception message.

mpfz0r: hmm, it is ignored anyway :-)
but yeah, better fix it

PR #5564, opened 2019-01-21
Add Script Alert Notification section to graylog.conf

danotorrey:
Add a new section to the default graylog.conf file for Script Alert Notification that includes the available parameters and default values.

@bernd Did I make this change in the correct place?

This change is needed for the Script Alert Notification feature https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/11

I will cherry-pick into 3.0 branch once merged.


bernd:
@danotorrey We usually don't add any plugin configuration options from plugins to the server config template. Especially not Enterprise related ones.

I think this should be highlighted in the integrations documentation and we should choose a good default value that works with the OS packages.

danotorrey:
@bernd Thanks for the info. Makes sense! Will do. Closing this PR.

PR #5603, opened 2019-01-28
Add the CONTAINS rule as one not requiring the field to be present

bernd:
* Add the CONTAINS rule as one not requiring the field to be present
* Add a regression test
* Ensure that the contains match returns false without the field

(cherry picked from commit 48ebfad187e1dca2ac79d775183ce4ac60b6f5bc)


bernd:
Moving this to the 3.0.1 milestone.

bernd:
Closing this as it hasn't been merged in a year and we most probably don't do another 3.0.x release anyway.

PR #5614, opened 2019-01-29
Fix selection of entities for creating a new version

kmerz:
## Description
The selection of entities in new versions of content packs was broken in production.

## Motivation and Context
Prior this PR the code relied on the class names to be correct with the instances constructor names.
But in production this names seem to differ. As it seemed `instanceof` did have similar problems
in development already.
So this PR uses `instanceof` with its own implementation.

## How Has This Been Tested?
- Created a content pack
- Created a new version and tested that all desired entities are part of it.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/content-packs/ContentPackConstraints.jsx
```diff
@@ -43,7 +43,7 @@ class ContentPackConstraints extends React.Component {
       return newConstraint;
     });
 
-    if (constraints instanceof 'Set') {
+    if (constraints.constructor.name === 'Set') {
```
bernd: Why are you removing the `instanceof` call here again?

graylog2-web-interface/src/logic/content-packs/Entity.js
```diff
@@ -64,6 +64,18 @@ export default class Entity {
     return this.getValueFromData('description') || '';
   }
 
+  /* eslint-disable-next-line class-methods-use-this */
+  get isEntity() {
+    return true;
+  }
+
+  static [Symbol.hasInstance](obj) {
```
bernd: Can we add a comment to this (and other occurrences) of this construct? I have no idea what it's doing. :smile:
Thank you!

Issue #4533, opened 2018-01-26
Show recent messages for an input never loads


edmundoa:
@TheJCamping wrote on https://github.com/Graylog2/graylog2-server/issues/4510#issuecomment-360869685:

On the /system/inputs page selecting "show recent messages" for an input will result in a page that never loads or an error message:
 Error Message: Unable to perform search query Details: Search status code: 500 Search response: cannot GET http://192.168.13.37:12900/search/universal/relative?query=gl2_source_input%3A54863813a78e39c792b058d1&range=0&limit=150&sort=timestamp%3Adesc (500)

In the developer tools for chrome. the network page shows a request to this address: http://graylog-01:12900/search/universal/relative?query=gl2_source_input%3A54863813a78e39c792b058d1&range=0&limit=150&sort=timestamp%3Adesc times out. If I change the range to 3000, I can get that request to load.

edmundoa:
This is a design decision we did in https://github.com/Graylog2/graylog2-server/pull/3725. I guess we need to think about that and decide if we want to revert it or take another measure to mitigate this issue.

TheJCamping:
Could this default to the current behavior and be changeable via the config file or a setting somewhere?

edmundoa:
As different people think the button should behave in different ways, my proposal would be to let people decide what they want without (more) configuration options. I think changing the button into a dropdown with options may help:

<img width="1655" alt="screen shot 2018-01-30 at 11 37 04" src="https://user-images.githubusercontent.com/716185/35561946-2b4e73da-05b2-11e8-998e-c09b23782424.png">

Any opinions? /cc: @lennartkoopmann

PR #5621, opened 2019-01-30
Reduce time range on 'Show received messages' button

### LINKED ISSUE #5620: 'Show received messages' link can trigger OOM with field charts query
Once a chart is created on the search page, it keeps its initially configured resolution
and query. Extending the search time range to `all messages` can create a query
that runs either Graylog or Eleasticsearch out of memory.
(See #5559)

This is easily achieved by accident, since the 'Show received messages' link
on inputs defaults to `range=0`.

## Possible Solution
Change the link to use a range of `last 7 days` instead.

chart-oom

* Graylog Version: 3.0 rc1
* Elasticsearch Version: 6.6

mpfz0r:
Defaulting to search in all messages is expensive and might trigger
OOM scenarios much easier.

Instead of searching all messages, we are searching only
messages in a range of past and future 26 hours.

26 hours because that's the largest possible timezone mistake a user can make.

Fixes #5620

### DIFF
--- a/graylog2-web-interface/src/util/TimeRangeHelper.ts
+++ b/graylog2-web-interface/src/util/TimeRangeHelper.ts
@@ -25,7 +25,7 @@ import type { AbsoluteRangeQueryParameter } from 'views/logic/TimeRange';
 const recentMessagesTimeRange = (): AbsoluteRangeQueryParameter => {
   const now = Date.now();
   // The biggest possible time difference on earth is 26 hours.
-  // It's between Kiribati is UTC+14 and the Howland Islands UTC-12
+  // It's between Kiribati (UTC+14) and the Howland Islands (UTC-12)
   // So we are going to create an absolute range
   // from 26 hours in the past to 26 hours into the future.
   const fromDate = new Date(now - 26 * 60 * 60000).toISOString();
--- a/graylog2-web-interface/src/util/TimeRangeHelper.ts
+++ b/graylog2-web-interface/src/util/TimeRangeHelper.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


bernd:
Refs https://github.com/Graylog2/graylog2-server/issues/4533

mpfz0r:
> I think @edmundoa's suggestion of using a drop-down to offer different values is a reasonable compromise between the possibility of killing you ES cluster and not finding your messages. slightly_smiling_face
> 
> [#4533 (comment)](https://github.com/Graylog2/graylog2-server/issues/4533#issuecomment-361553051)
> 
> What do you think?

From a discussion I had (forgot with whom) the reason why it was changed to "all messages" was that
this would show messages from the future. Apparently some of our users haven't discovered NTP yet..

So couldn't we just set the time range to 'last 7 days - now + 1 day' or so?

bernd:
> So couldn't we just set the time range to 'last 7 days - now + 1 day' or so?

That could be done by using an absolute time range parameter instead of the relative one, yes.

graylog2-web-interface/src/util/TimeRangeHelper.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Creates an absolute time range that can be used to look up recently received messages.
+ * It accommodates for the eventuality that messages can have timestamps from the future
+ * due to wrong timezone parsing or wrong system clocks.
+ */
+const recentMessagesTimeRange = (): AbsoluteRangeQueryParameter => {
+  const now = Date.now();
+  // The biggest possible time difference on earth is 26 hours.
+  // It's between Kiribati is UTC+14 and the Howland Islands UTC-12
```
mpfz0r, line 28:
```suggestion
  // It's between Kiribati (UTC+14) and the Howland Islands (UTC-12)
```

ousmaneo (approved):
Tested and works!

bernd (approved):

PR #5665, opened 2019-02-08
Fix auto-completion in several drop-down fields

### LINKED ISSUE #5659: Auto-complete on Lookup Table creation page is broken
Not going to write a long description, because I think this shows the problem well:

autocomplete

* Graylog Version: Graylog 3.0.0-rc.2+2667ff2
* Browser version: Firefox 65.0 on Ubuntu 18.04

bernd:
In all these cases the matchProp needs to be set to "label" and not "value"
because the value contains the object ID or another identifier.

After this, the auto-completion is looking at same data that the user
is seeing.

Fixes #5659


danotorrey:
I am going to work on fixing up these linter errors.

bernd:
@danotorrey I think we shouldn't do that for this PR. There are a lot of unrelated linter errors and fixing them here would only mixup the actual issue fix with linter fixes.

danotorrey:
@bernd Good point. Thanks. I'll ignore those for now.

PR #5669, opened 2019-02-11
Fix slow loading UI after server startup

mpfz0r:
Avoid registering ShiroSecurityContextFilter as a DynamicFeature.
Doing so, triggers Jersey to create a ServiceLocator for all
existing (~1600) Resources. [1]
Furthermore, each created ServiceLocator will have its own Guice/HK2 Bridge,
whose JIT resolver cache is cold upon startup and needs to be filled.

Instead of registering the ShiroSecurityContextFilter dynamically, we
register it statically in the JerseyService because every Resource is filtered
through it anyway.
We need to give it a higher priority so it's executed before the authentication filter.

[1]
https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/model/ResourceMethodInvoker.java#L187


lennartkoopmann:
yesssssssss

deeshe:
We decided to remove this from the Stable Backports for now until we are more confident that it doesn't have any negative implications.

PR #5670, opened 2019-02-11
Server changes for Forwarder

danotorrey:
The Graylog Server will need several changes in order for the Graylog Forwarder to be implemented.

1) The Graylog Forwarder feature uses the same Journaling capability that exists in the Graylog Server currently. The `KafkaJournal` class has been reworked to allow support for instance-based metric names.

2) Output implementations do not currently have access to the output id. A change has been made that allows the output implementation to access the output instance and therefore the id.

See https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/30 for the core Forwarder PR.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
@@ -48,7 +48,7 @@ public class MessageOutputFactory {
 
         Preconditions.checkArgument(factory != null, "Output type is not supported: %s!", outputType);
 
-        return factory.create(stream, configuration);
+        return factory.create(output, stream, configuration);
     }
 
 
--- a/graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
@@ -29,11 +29,16 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class MessageOutputFactory {
+    // We have two output lists for backwards compatibility.
+    // See comments in MessageOutput.Factory and MessageOutput.Factory2 for details
     private final Map<String, MessageOutput.Factory<? extends MessageOutput>> availableOutputs;
+    private final Map<String, MessageOutput.Factory2<? extends MessageOutput>> availableOutputs2;
 
     @Inject
-    public MessageOutputFactory(Map<String, MessageOutput.Factory<? extends MessageOutput>> availableOutputs) {
+    public MessageOutputFactory(Map<String, MessageOutput.Factory<? extends MessageOutput>> availableOutputs,
+                                Map<String, MessageOutput.Factory2<? extends MessageOutput>> availableOutputs2) {
         this.availableOutputs = availableOutputs;
+        this.availableOutputs2 = availableOutputs2;
     }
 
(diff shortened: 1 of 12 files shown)


danotorrey:
@bernd Have you seen this test error before?

image

bernd:
> @bernd Have you seen this test error before?
> 

Yes, it's unrelated to your changes. And I think it's already been fixed in the server.

danotorrey:
Note that #5738 also contains changes needed for the Forwarder to work. The `forwarder` branch has temporarily been rebased off of `graceful-shutdown-service` (to allow dev/testing of the Forwarder), which we can undo before this PR is ready for merging.

danotorrey:
@bernd These Forwarder server changes are ready for review.

danotorrey:
@bernd I had to add handling to teardown `KafkaJournal` metrics (see https://github.com/Graylog2/graylog2-server/pull/5670/commits/f75f1c8653bcea2b524c0b80d57aea4ef3f075be) when the journal is shutdown to ensure that restarts of the Forwarder output are successful and don't result in a duplicate metric error (when editing and saving the output, for example). This PR is ready for review.

graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
```diff
@@ -44,35 +49,56 @@ public MessageOutput fromStreamOutput(Output output, final Stream stream, Config
         final String outputType = output.getType();
         Preconditions.checkArgument(outputType != null);
 
+        // We have two output lists for backwards compatibility.
+        // See comments in MessageOutput.Factory and MessageOutput.Factory2 for details
+        final MessageOutput.Factory2<? extends MessageOutput> factory2 = this.availableOutputs2.get(outputType);
```
danotorrey: @bernd Minor question: Is there any reason why `Factory2` is before `Factory` here? Just wondering if we want `Factory` to come first for consistency. Same question for the if-else block that follows.

danotorrey:
@bernd `MessageOutput.Factory2` changes LGTM (except for one minor question https://github.com/Graylog2/graylog2-server/pull/5670/files#r262732042).

Also wondering: What is the long-term plan for `Factory2`? Will we keep it forever, or is there a path to consolidate in a future major version.

danotorrey:
@bernd All metrics in `KafkaJournal` are now torn down properly on restart. I left the initialization in two steps in the constructor, since I think this was deliberate in original implementation (possibly some metrics used while constructor is executing).

graylog2-server/src/main/java/org/graylog2/outputs/MessageOutputFactory.java
bernd: @danotorrey The idea was that plugins using the newer factory interface should be preferred if the same plugin is registered twice for some reason. I will add a comment that explains it.

bernd:
@danotorrey 

> @bernd the `MessageOutput.Factory2` changes LGTM (except for one minor question above). Can you please verify that the constructor changes in [Graylog2/graylog-plugin-enterprise-integrations@9add4a5](https://github.com/Graylog2/graylog-plugin-enterprise-integrations/commit/9add4a5c980ed3ff6583208d7f27515849a81dc7) are valid?

Looks good and works for me. :+1: 
 
> Also wondering: What is the long-term plan for `Factory2`? Will we keep it forever, or is there a path to consolidate in a future major version.

I guess we will have it until we refactor the `MessageOutput`. This will be a breaking change for all output plugins, so I guess it will take some time. So yeah, we will probably keep it for now.

PR #5672, opened 2019-02-12
Fix gui test after release

kmerz:
Prior to this change, the release did change the doc path
and a snapshot test would fail.

This change will update the failing snapshot.


kmerz:
Dup from #5671

PR #5675, opened 2019-02-12
Add pagination to stream overview page

kmerz:
## Description
Add pagination to the stream overview page.

## Motivation and Context
A good overview page should have: Search and Pagination.
This PR add pagination and uses the search on the backend site.
The code is mainly copied from lookup table.
The list all API endpoint is kept since it makes sense when using  scripts to avoid stepping
through all streams with pages.

## How Has This Been Tested?
Add more than 10 streams and stepped through the pages and searched for streams.

graylog - streams

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -189,6 +209,41 @@ private boolean ensureUserHasPermissionsForStream(User user, String id) throws V
...
+        @ApiParam(name = "order", value = "The sort direction", allowableValues = "asc, desc")
+        @DefaultValue("asc") @QueryParam("order") String order) {
+
+        SearchQuery searchQuery;
+        try {
+            searchQuery = searchQueryParser.parse(query);
+        } catch (IllegalArgumentException e) {
+            throw new BadRequestException("Invalid argument in search query: " + e.getMessage());
+        }
+        final Predicate<StreamDTO> permissionFilter = streamDTO -> isPermitted(RestPermissions.STREAMS_READ, streamDTO.id());
```
dennisoelkers: We are duplicating a bit of code here vs. `StreamResource#get`. We have several options to address this:

  a) Reuse `getPage` in `get` and returning `getPage(...).streams()` in `get`. This reduces duplication to a minimum, but involves unnecessary boxing/instantiation.
  b) Extract a method that returns `streams` (of line 237), which is used in both methods.
  c) Extract `permissionFilter` and only reuse that one in both.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -189,6 +209,41 @@ private boolean ensureUserHasPermissionsForStream(User user, String id) throws V
         return permissionsChanged;
     }
 
+    @GET
+    @Timed
+    @Path("/page")
```
dennisoelkers: What about `paginated` instead?

graylog2-web-interface/src/routing/ApiRoutes.js
```diff
@@ -183,6 +183,7 @@ const ApiRoutes = {
   },
   StreamsApiController: {
     index: () => { return { url: '/streams' }; },
+    indexPage: () => { return { url: '/streams/page' }; },
```
dennisoelkers: Maybe `indexPage` -> `paginated`?

graylog2-web-interface/src/components/streams/StreamComponent.jsx
```diff
@@ -75,51 +93,32 @@ class StreamComponent extends React.Component {
...
-                                 onDataFiltered={this._updateFilteredStreams} />
+            <SearchForm onSearch={this._onSearch} onReset={this._onReset} useLoadingState />
           </Col>
         </Row>
         <Row>
           <Col md={12}>
-            {streamsList}
+            <PaginatedList onChange={this._onPageChange}
+                           totalItems={pagination.total}>
+              <br />
```
kyleknighted: We should be using CSS for margin

graylog2-web-interface/src/util/PaginationHelper.js
```diff
@@ -0,0 +1,13 @@
+const PaginationHelper = {
+  urlGenerator: (destUrl, page, perPage, query) => {
```
kyleknighted: Maybe we could do something like this so there is less repeated code?

```js
  urlGenerator: (destUrl, page, perPage, query) => {
    const url = `${destUrl}?page=${page}&per_page=${perPage}&resolve=true`;
    
    if (query) {
      return `${url}&query=${encodeURIComponent(query)}`;
    }

    return url;
  },
```

graylog2-web-interface/src/stores/streams/StreamsStore.ts
```diff
@@ -44,6 +45,28 @@ class StreamsStore {
...
+        const pagination = {
+          count: response.pagination.count,
+          total: response.pagination.total,
+          page: response.pagination.page,
+          perPage: response.pagination.per_page,
+          query: response.pagination.query,
+        };
+        return {
+          streams: response.streams,
+          pagination: pagination,
```
kyleknighted: With [object shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015) you don't need to specify the variable if the key is the same.

```suggestion
          pagination,
```

graylog2-web-interface/src/util/PaginationHelper.js
```diff
@@ -0,0 +1,12 @@
+// @flow strict
+const PaginationHelper = {
+  urlGenerator: (destUrl: string, page: number, perPage: number, query: string, resolve: boolean = true): string => {
```
dennisoelkers, line 3: This function could use a better name. Maybe something like `paginateUrl`?

graylog2-web-interface/src/util/PaginationHelper.js
```diff
@@ -0,0 +1,12 @@
+// @flow strict
+const PaginationHelper = {
```
dennisoelkers, line 2: We should get rid of the enclosing object. It does not add benefit, but prevents tree shaking potentially, because we cannot import only a single function.

graylog2-web-interface/src/util/PaginationHelper.js
```diff
@@ -0,0 +1,12 @@
...
+  urlGenerator: (destUrl: string, page: number, perPage: number, query: string, resolve: boolean = true): string => {
+    const url = `${destUrl}?page=${page}&per_page=${perPage}&resolve=${resolve.toString()}`;
```
dennisoelkers, line 4: Instead of manual string handcrafting, we could use the [URI.js](https://medialize.github.io/URI.js/) builder to help with escaping et al.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -189,6 +209,41 @@ private boolean ensureUserHasPermissionsForStream(User user, String id) throws V
...
+        try {
+            searchQuery = searchQueryParser.parse(query);
+        } catch (IllegalArgumentException e) {
+            throw new BadRequestException("Invalid argument in search query: " + e.getMessage());
+        }
+        final Predicate<StreamDTO> permissionFilter = streamDTO -> isPermitted(RestPermissions.STREAMS_READ, streamDTO.id());
+        final PaginatedList<StreamDTO> result = paginatedStreamService
+                .findPaginated(searchQuery, permissionFilter, page, perPage, sort, order);
+        final List<StreamDTO> streams = result.stream().map(streamDTO -> {
+            List<StreamRule> rules = streamRuleService.loadForStreamId(streamDTO.id());
```
dennisoelkers, line 238: Instead of loading rules for each stream separately, we should use `StreamRuleService#loadForStreamIds` so we are generating 2 queries instead of n+1.

UPGRADING.rst
```diff
@@ -4,6 +4,12 @@ Upgrading to Graylog 4.0.x
 
 .. _upgrade-from-33-to-40:
 
+Mark 'List of streams' in Streams API als deprecated
```
dennisoelkers, line 7: Maybe we should phrase this as `Deprecation of API endpoint for unpaginated listing of streams`?

UPGRADING.rst
```diff
@@ -4,6 +4,12 @@ Upgrading to Graylog 4.0.x
...
+====================================================
+
+In 4.0 we introduce a new API endpoint to retrieve streams from the backend: '/streams/paginated'.
```
dennisoelkers, line 10: We should mention that this one allows pagination parameters.

UPGRADING.rst
```diff
@@ -4,6 +4,12 @@ Upgrading to Graylog 4.0.x
...
+We therefore mark '/streams' as deprecated. This endpoint will be removed in the next version.
```
dennisoelkers, line 11: It would be helpful to a) mention that `/streams` will continue to exist, but will have the request & response format of `/streams/paginated` and b) advise the user to start preparing any consumer of the `/streams` format to accept the new response format so only the endpoint URL needs to be switched for the next version.

PR #5681, opened 2019-02-13
Also return ES "caused_by" error message for failed searches

mpfz0r:
If you search on a field that is mapped to a numeric,
you won't see the real reason on why this has failed.
I assume the error format has changed in recent ES versions,
and the info is now hidden in the `caused_by` field on shard failures.

For reference, this is the response:
```
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 64,
    "failures" : [
      {
        "shard" : 0,
        "index" : "graylog_0",
        "reason" : {
          "type" : "query_shard_exception",
          "reason" : "failed to create query: {\n [....]  }\n}",
          "index" : "graylog_0",
          "caused_by" : {
            "type" : "number_format_exception",
            "reason" : "For input string: \"foo\""
          }
        }
      }
    ]
  },
  "hits" : {
    "total" : 0,
    "max_score" : null,
    "hits" : [ ]
  }
}
```

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchFailure.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/SearchFailure.java
@@ -1,3 +1,19 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


dennisoelkers:
Could you also add a test case for this?

dennisoelkers:
@mpfz0r: Only the license headers are missing, then we are good to go.

PR #5679, opened 2019-02-13
Paginated dashboard overview list

kmerz:
## Description
Add pagination to the dashboard overview page together with filtering.

## Motivation and Context
A good overview page should have: Search and Pagination.
This PR add pagination and uses the search on the backend site.
The code is mainly copied from lookup table.
The list all API endpoint is kept since it makes sense when using  scripts to avoid stepping
through all dashboards with pages.

## How Has This Been Tested?
Add more than 10 dashboards and stepped through the pages and searched for streams.
Created, updated and deleted dashboards and ensured that table refreshes accordingly.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

# Caution
This PR is based on #5675 and must be merged **after** that PR.

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #5686, opened 2019-02-14
Paginated pipelines

kmerz:
## Description
Add pagination to the pipelines overview page together with filtering.

## Motivation and Context
A good overview page should have: Search and Pagination.
This PR add pagination and uses the search on the backend site.
The code is mainly copied from lookup table.
The list all API endpoint is kept since it makes sense when using  scripts to avoid stepping
through all dashboards with pages.

## How Has This Been Tested?
Add more than 10 pipelines and stepped through the pages and searched for pipelines.
Created, updated and deleted pipelines.

## Types of changes

# Caution
This PR is based on #5675 and #5679 and must be merged **after** these PRs.

## Checklist:


graylog2-web-interface/src/components/pipelines/ProcessingTimelineComponent.tsx
```diff
@@ -99,7 +137,7 @@ const ProcessingTimelineComponent = () => {
     const stageNumbers = stages.map((stage) => stage.stage);
 
     return pipelines
-      .map((p) => p.stages.map(({ stage }) => stage))
+      .map(({ stages: pStages }) => pStages.map(({ stage }) => stage))
```
kyleknighted, line 140: what does `p` stand up in `pStages` and why can't it just be `stages`?

graylog2-web-interface/src/components/pipelines/ProcessingTimelineComponent.tsx
```diff
@@ -119,20 +157,24 @@ const ProcessingTimelineComponent = () => {
...
-        PipelinesActions.delete(pipeline.id);
+        PipelinesActions.delete(pipeline.id).then(() => {
+          setPagination({ page: DEFAULT_PAGINATION.page, perPage, query });
+        });
       }
     };
   };
 
   const _pipelineFormatter = (pipeline) => {
+    const { id, title, description, stages: pStages } = pipeline;
```
kyleknighted, line 168: can we all it `pipelineStages` or `stages`? `pStages` has no real meaning out of context of this destructure

kyleknighted (changes requested):
Upon test driving, if I go to Page 2 of pipelines and then click my browsers back button, it takes me to what I was viewing before Page 1. I would expect it to return me to Page 1

graylog2-web-interface/src/components/pipelines/ProcessingTimelineComponent.tsx
edmundoa: Unfortunately ESLint complains about that because it's shadowing the already defined `stages` argument. However, I have replaced it to the more verbose `pipelineStages` since I think it makes clearer what that is (and where the `p` comes from).

edmundoa:
> Upon test driving, if I go to Page 2 of pipelines and then click my browsers back button, it takes me to what I was viewing before Page 1. I would expect it to return me to Page 1

@kyleknighted I agree that this would be a desired behaviour but I don't think we do this anywhere in the product, so we will introducing an inconsistent behaviour here. In all paginated lists I tried, going through pages does not store a new entry in the browser's history, and even on the search page the back button does not take you to the previous page in the message table (at least it didn't for me when I tried).

This does not mean we can't change how pagination works but I'm not sure if this PR is the right place to do it.

We should also think a bit more how to do it, because saving that state internally and using it will become confusing at some point. I think the clearer way of doing it would be to put pagination and query information in the URL, and also saving that in the browser's history (e.g. Pipelines - Page 2). In my opinion, doing that consistently would also make more sense when hitting the browser's reload button: the information is in the URL, so it's clearer why you are landing on a certain page and with a certain search query.

What do you think? Should we open a new issue to document this and take care of it?

edmundoa:
@kyleknighted in the end I created a custom hook that we could use for getting and setting pagination details in the browser's query and also used it in the PR. I did a quick and dirty research on how some websites do pagination and I couldn't see that any sets pages in the document title, so I dropped the idea for now.

I would appreciate if you take the changes for another drive and let me know how you like it. Thank you!

kyleknighted (approved):
looks great, sorry for the delay! 
happy to wait on the back button fix as a wide spread solution instead of focused on this one. Thanks to you and @kmerz for this!

PR #5699, opened 2019-02-20
Add new pipeline function "grok_exists"

### LINKED ISSUE #5689: Test for existing Grok-Pattern during pipeline-processing
The idea is to add either a check if a string is a valid name of a Grok-Pattern, or generally is a Grok-Pattern that can be used, or to allow the grok() function to return without a processing error if it can't resolve a name to a pattern.
The reason for such feature is as follows:
A few programs/devices write log messages that all start the same, but then have different syntaxes, based on what they exactly log in that message. But you will be able to tell based on the common syntax at the start. An example for this would be the Cisco asa, every message starts containing something like  "%ASA-4-722004:", and this number can be used to identify the syntax and data following it.
To extract the data from the variable part of said logs, you will need to make one Grok-Pattern per Log-Pattern. But said Grok-Pattern needs a rule to be applied in a pipeline. This can lead to a stage with 70 rules and more, all which look the same except a small check that a value is equal to something, and a different Grok-Pattern.
To reduce processing, and the ammount of needed rules, it would be usefull to be able to just take the value you are testing against, and instead append it to the Grok-Pattern-Name, which would result in 1 rule for every Grok-Pattern. This can be done if the existing patterns are there for every possible logging pattern, but if one is missing, it will cause a processing error, which at least adds the field with the error message to the processed log, and probably causes a performance hit from handling the error.
It's not always possible to provide a 100% coverage of the possible patterns, so a check if the corresponding Grok-Pattern exists would allow the usage of a single rule without a 100% coverage.

kmerz:
## Description
Prior to this change, a missing grok pattern would
raise a error in the pipeline processor when using the "grok"
function. But the user would like to able to make one rule
which uses a grok pattern dynamicaly depending on if
a grok pattern exists or not.

This change adds a new function "grok_exists" which will
return true or false depending if a grok pattern exists.
Additionally it will make a entry to the graylog-server.log
if the second argument of the function is true and
the pattern was not found.

## How Has This Been Tested?
- Add a grok pattern named "KONRAD"
- Add a new pipeline rule:
```
rule "grok exists true"
when
    grok_exists("KONRAD")
then
    debug("grok exists");
end
```
- Check that debug log got raised
- Changed Pattern name and checked that debug log got not raised.

Fixes #5689 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/GrokExists.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/GrokExists.java
@@ -1,0 +1,79 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog.plugins.pipelineprocessor.functions;
+
+import io.krakens.grok.api.Grok;
+import org.graylog.plugins.pipelineprocessor.EvaluationContext;
+import org.graylog.plugins.pipelineprocessor.ast.functions.AbstractFunction;
+import org.graylog.plugins.pipelineprocessor.ast.functions.FunctionArgs;
+import org.graylog.plugins.pipelineprocessor.ast.functions.FunctionDescriptor;
+import org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor;
+import org.graylog2.grok.GrokPatternRegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.collect.ImmutableList.of;
+
+import javax.inject.Inject;
+
+public class GrokExists extends AbstractFunction<Boolean> {
+
+    private static final Logger log = LoggerFactory.getLogger(GrokExists.class);
+    public static final String NAME = "grok_exists";
+
+    private final ParameterDescriptor<String, String> patternParam;
+    private final ParameterDescriptor<Boolean, Boolean> doLog;
+
+    private final GrokPatternRegistry grokPatternRegistry;
+
+    @Inject
+    public GrokExists(GrokPatternRegistry grokPatternRegistry) {
+        this.grokPatternRegistry = grokPatternRegistry;
+
+        patternParam = ParameterDescriptor.string("pattern")
+                .description("The Grok Pattern which is to be tested for existance.").build();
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/GrokExists.java
```diff
@@ -0,0 +1,79 @@
...
+        return patternExists;
+    }
+
+    @Override
+    public FunctionDescriptor<Boolean> descriptor() {
+       return FunctionDescriptor.<Boolean>builder()
+               .name(NAME)
+               .returnType(Boolean.class)
+               .params(of(patternParam, doLog))
+               .description("Tests if the searched Grok Pattern exists.")
```
mpfz0r: `Pattern` in lowercase please

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/GrokExists.java
```diff
@@ -0,0 +1,79 @@
...
+        final String pattern = patternParam.required(args, context);
+        final boolean logWhenNotFound = doLog.optional(args, context).orElse(false);
+
+        if (pattern == null) {
+            return null;
+        }
+
+        final boolean patternExists = grokPatternRegistry.grokPatternExists(pattern);
+        if (!patternExists && logWhenNotFound) {
+           log.info("Grok Pattern " + pattern + " was not found in pipeline rule function execution");
```
mpfz0r: I wonder why you want this...
No other pipeline function logs.
This could potentially create a crazy volume of log messages.
Couldn't we at least rate limit this?

bernd: I think it's okay here because it only logs the message if the log parameter is set to true. We might want to include a warning in the parameter description, though. Without the optional logging it's hard to debug this.

@kmerz How about logging something like `log.info("Grok pattern <{}> doesn't exist", pattern);`? Using "not found in pipeline rule function execution" sounds a bit like an issue with the pipeline system instead of a missing Grok pattern.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/GrokExists.java
```diff
@@ -0,0 +1,79 @@
...
+    private final GrokPatternRegistry grokPatternRegistry;
+
+    @Inject
+    public GrokExists(GrokPatternRegistry grokPatternRegistry) {
+        this.grokPatternRegistry = grokPatternRegistry;
+
+        patternParam = ParameterDescriptor.string("pattern")
+                .description("The Grok Pattern which is to be tested for existance.").build();
+        doLog = ParameterDescriptor.bool("do_log").optional()
+                .description("Log if the Grok Pattern is missing.").build();
```
bernd: Please add a warning to the description that enabling the logging can produce a high volume of logs.

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternRegistry.java
```diff
@@ -83,6 +83,10 @@ public void grokPatternsDeleted(GrokPatternsDeletedEvent event) {
         reload();
     }
 
+    public boolean grokPatternExists(String patternName) {
+        return grokPatternService.loadByName(patternName).isPresent();
```
bernd: This is hitting the database for every pipeline function call. The `GrokPatternRegistry` is using a cache for caching `Grok` instances. I think we need to add a cache for this method as well. Otherwise this can put too much load on the database.

Issue #5680, opened 2019-02-13
Graylog cannot authenticate on MongoDB if password contains the sign +


frantz45:
## Expected Behavior
Graylog should authenticate on MongoDB

## Current Behavior
Graylog does not authenticate on MongoDB

## Possible Solution
Choose a password without the sign +

## Steps to Reproduce (for bugs)
1. Enable authentication on MongoDB
2. Restart MongoDB service
3. Create a MongoDB user with the sign + in the password
4. Configure mongodb_uri = mongodb://grayloguser:secret+@localhost:27017/graylog
5. Restart Graylog
6. Graylog's logs (server.log) say "Authentication failed on MongoDB error code 18"
7. Try to connect to MongoDB from bash with the same mongodb_uri and it works:
mongo mongodb://grayloguser:secret+@localhost:27017/graylog

MongoDB doc does not say the sign + needs to be encoded:
If the username or password includes the at sign @, colon :, slash /, or the percent sign % character, use percent encoding.
https://docs.mongodb.com/manual/reference/connection-string/

## Context
Puppet automatically chooses a password for the graylog MongoDB user. + was in the charset used.

## Your Environment
* Graylog Version: 2.4.6
* Elasticsearch Version: 5.6
* MongoDB Version: 3.6.6
* Operating System: CentOS 7
* Browser version: Firefox ESR

mariussturm:
Hi @frantz45,
thanks for the report! This is a bug in the mongodb library version we use for 3.0. The issue was already fixed upstream here: https://github.com/mongodb/mongo-java-driver/commit/4a44f05b2ae47d4771453d5874c910deba1dd61c
I have created a PR to get this into Graylog 3.0.1 https://github.com/Graylog2/graylog2-server/pull/5736

mariussturm:
@frantz45 unfortunately we can't jump easily to a more recent  mongo-java-driver version. For now the user needs to escape plus signs by hand with this string: `%2B`.

We will add this to the examples in the default configuration.

bernd:
Since we have a (documented https://github.com/Graylog2/graylog2-server/pull/5795) workaround now, I will close this ticket. The underlying issue is not fixed, but will be fixed once we update the MongoDB driver.

PR #5736, opened 2019-02-28
Update mongodb-driver version to 3.10.1

mariussturm:
Update mongodb client library to latest version

## Motivation and Context

This fixes https://github.com/Graylog2/graylog2-server/issues/5680

## How Has This Been Tested?

Regression test added

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/MongoDbConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/MongoDbConfiguration.java
@@ -21,6 +21,7 @@ import com.github.joschi.jadconfig.ValidationException;
 import com.github.joschi.jadconfig.ValidatorMethod;
 import com.github.joschi.jadconfig.validators.PositiveIntegerValidator;
 import com.github.joschi.jadconfig.validators.StringNotBlankValidator;
+import com.mongodb.ConnectionString;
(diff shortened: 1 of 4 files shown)


mariussturm:
Looks like fongo is not compatible with newer mongodb-driver versions :/
https://github.com/fakemongo/fongo/issues/357

bernd:
> Looks like fongo is not compatible with newer mongodb-driver versions :/
> [fakemongo/fongo#357](https://github.com/fakemongo/fongo/issues/357)

Okay, how about doing the `+` character replacement on our own as a workaround? That's what the [MongoDB Java driver](https://github.com/mongodb/mongo-java-driver/commit/4a44f05b2ae47d4771453d5874c910deba1dd61c) is doing:

```diff
-            userInfo = userAndHostInformation.substring(0, idx);
+            userInfo = userAndHostInformation.substring(0, idx).replace("+", "%2B");
```

It's not really nice but it would allow us to work around the issue until we can refactor/update our testing infrastructure.

@mariussturm What do you think?

mariussturm:
Well we need some partial parsing of the connection string to only apply the replacement on the username and password. So that we replicate the behaviour of the mongo-driver code. If this is not too much overhead we can do it like this.

PR #5738, opened 2019-02-28
Add GracefulShutdownService

bernd:
Services can register themselves with the GracefulShutdownService to
make sure they will be stopped gracefully on server shutdown.

This allows plugins to participate in a graceful server shutdown to
make sure the server is not terminated before the plugin shutdown
routine is done.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/system/shutdown/GracefulShutdownService.java
+++ b/graylog2-server/src/main/java/org/graylog2/system/shutdown/GracefulShutdownService.java
@@ -36,7 +36,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * A service that participates in the Graylog server graceful shutdown.
  * <p>
  * Services can implement {@link GracefulShutdownHook} and register themselves with this service to make sure they
- * get shut down properly on server shutdown.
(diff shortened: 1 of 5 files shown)


danotorrey:
@bernd @ceruleancee This tests successfully with the Forwarder.

bernd:
@danotorrey @ceruleancee I added some more changes, please test again. Thank you! :smiley:

graylog2-server/src/main/java/org/graylog2/system/shutdown/GracefulShutdownService.java
```diff
@@ -0,0 +1,139 @@
...
+                        LOG.error("Problem shutting down <{}>", hookName, e);
+                    } finally {
+                        latch.countDown();
+                    }
+                });
+            }
+
+            latch.await();
+            executor.shutdown();
+            executor.awaitTermination(1, TimeUnit.MINUTES);
```
kroepke: I don't think is necessary here, if the latch is open, all tasks have completed and we don't need to wait for anything.

bernd: Agreed, thanks! :+1:

PR #5853, opened 2019-04-08
Move web build and cache directory into target/

bernd:
The "build" and "cache" directories are currently located in the project
root of plugins. This change moves them to the "target/web" directory.

That way we don't need any custom configuration to remove these
directories on "mvn clean" and they should also not be indexed by
IntelliJ and consume resources. (most developers exclude the "build" and
"cache" directories in IntelliJ for performance reasons)

### DIFF
--- a/graylog-plugin-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/graylog-plugin-archetype/src/main/resources/archetype-resources/pom.xml
@@ -96,7 +96,7 @@
 
     <build>
         <resources>
-            <resource><directory>build</directory></resource>
+            <resource><directory>${web.build-dir}</directory></resource>
             <resource>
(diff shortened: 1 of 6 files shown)


bernd:
This also requires a change to all plugins:

```diff
diff --git pom.xml pom.xml
index 4bf05c4..188c31c 100644
--- pom.xml
+++ pom.xml
@@ -138,7 +138,7 @@
     <build>
         <resources>
             <resource>
-                <directory>build</directory>
+                <directory>${web.build-dir}</directory>
             </resource>
             <resource>
                 <directory>src/main/resources</directory>
```

bernd:
@edmundoa @kyleknighted @kmerz Any opinion on this? :smiley: Otherwise I will create PRs for the other repos and merge it.

Issue #5860, opened 2019-04-10
Add `name` attribute to inputs


kyleknighted:
Created Issue related to older PR #5449 

## Expected Behavior
For permit to extensions like password managers (enpass e.g.) to permit auto-login,
and for permit global companies configuration to provide an auth form based e.g Onelogin.
We need to have a name attribute on inputs in the login form to allow this kinds of services to login users automatically.

## Current Behavior
Inputs do not have `name` attributes that some password managers rely on.

## Possible Solution
Add a name props on Input component.
Add name props on Login page for username input and password input.

* Graylog Version: 3.0

kyleknighted:
Closed by https://github.com/Graylog2/graylog2-server/pull/5861

PR #5861, opened 2019-04-10
Add `name` prop to Input component

kyleknighted:
## Description
This adds the `name` attribute to the prop definitions on the Bootstrap Input component. 
It also defaults to `id` attribute, which is required, if the `name` attribute is not defined.

## Motivation and Context
Issue #5860 
Helps with auto-fill tools, like password managers
Helps with some a11y tools 

## How Has This Been Tested?
Validated in browser
Created jest test w/ snapshot

## Types of changes

## Checklist:


graylog2-web-interface/src/pages/LoginPage.jsx
```diff
@@ -67,7 +79,12 @@ const LoginPage = createReactClass({
       return (
         <div className="form-group">
           <Alert bsStyle="danger">
-            <a className="close" onClick={this.resetLastError}>×</a>{error}
+            <button className="close"
```
kyleknighted: eslint was complaining about this being an anchor without a `href`, so i swapped it to a button

PR #5869, opened 2019-04-15
Fixed shebang path.

jed-frey:
Currently doesn't work on FreeBSD as is. Not all distros use /bin/bash either.

/usr/bin/env bash should always point to the correct shell.

## Description

Changed ```/bin/bash``` to ```/usr/bin/env bash```.

https://en.wikipedia.org/wiki/Env

## Motivation and Context

Doesn't work on FreeBSD as tested.

## How Has This Been Tested?

```/usr/bin/env``` has been fairly tested. There are implementations on almost every major POSIX environment.

## Types of changes

## Checklist:
- [X ] My code follows the code style of this project.

### DIFF
--- a/bin/graylogctl
+++ b/bin/graylogctl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 CMD=$1
 NOHUP=${NOHUP:=$(which nohup)}


dennisoelkers:
Thanks for the contribution, @jed-frey! This is definitely an improvement. Considering that `bash` is not included in the base of Free/Net/OpenBSD, maybe a next step would be to check for any bashisms in the script and make it runnable by a plain bourne shell?

PR #5894, opened 2019-04-16
Checking health of production build assets during compilation.

dennisoelkers:
## Motivation and Context

Due to dependency updates or other changes, regressions impacting the build can sneak in easily.
For development, this will be noticed very quickly, but for production this could go unnoticed. Therefore, this change adds a small script which is utilizing puppeteer and headless chrome to bootstrap a small express instance, generate HTML boilerplate referencing assets extracted from the module manifest and navigating to it in the browser instance.

When no errors happen during page loading (including any logging to the console, which is considered to be an error as well), the script returns successfully. If not, errors are printed to the console and the build fails.

## Types of changes

## Checklist:


bernd:
@dennisoelkers What's the plan for testing the plugins? As far as I can see this PR doesn't address that, right?

dennisoelkers:
@bernd: No, this PR is for core only. The plan is to gather some experience in core and getting it to run reliably there, then tackling making it usable in plugins. The script itself can already take an arbitrary number of plugin manifests (the json files generated by a webpack run for plugins, listing the generated assets) and test them together with the core.

graylog2-web-interface/checkProductionBuild.js
```diff
@@ -0,0 +1,188 @@
...
+const cors = require('cors');
+
+const VENDORMODULE = 'vendor-module.json';
+const BUILDMODULE = 'module.json';
+
+const HEADLESS_SHELL = 'test/bin/headless_shell';
+
+const isExecutable = (filename) => {
+  try {
+    // eslint-disable-next-line no-bitwise
```
kyleknighted: Can we include a comment here as to why we need to disable bitwise operator linting? 
Should this just be a rule we change instead of disabling? 
Is there a way to code this that doesn't require disabling of linter?

kmerz: No I think this code needs to be written that way.
At least for me it is clear that way what is happening.
So I would go for the bit-wise OR since it is what is happening here.

Issue #2607, opened 2016-08-03
Provide ability to bulk import data with GELF API


vangheem:
## Expected Behavior
To be able to push more than one log entry at a time from the GELF API.

## Current Behavior
No able to do so.

## Possible Solution
Provide ability to bulk push entries with GELF endpoint.

## Context
Pushing data could be faster if bulk allowed.

joschi:
@vangheem What specifically do you mean with "GELF API"?

vangheem:
http://docs.graylog.org/en/2.0/pages/gelf.html#example-payload

kroepke:
We currently are not planning to update the GELF spec.

haizaar:
Hi guys,

Just to confirm - there is currently no way to send multiple log messages in a single [HTTP POST](http://docs.graylog.org/en/2.4/pages/sending_data.html#gelf-via-http), right?

joschi:
@haizaar Correct.

haizaar:
I see.

Did you ever consider introducing it? May be through a dedicated API endpiont, e.g.

```
POST /gelf/_bulk
{"short_message": "first_message"}
{"short_message": "second_message"}
{"short_message": "third_message"}
...
```

(Apologies if this is not the right place to discuss this)

bernd:
@haizaar Please open a separate feature request issue for this. Thank you!

PR #5924, opened 2019-05-02
#2607: allow bulk import / multiple messages input for http gelf

darashenka:
allow bulk import / multiple messages in request for http gelf input 

## Description
you can create json-array of well-known gelf input objects and send them to graylog within 1 request

## Motivation and Context
http gelf input is quite expensive and sending millions of log entries take quite a lot time. 
You can combine these messages into one array and send them within one request.

## How Has This Been Tested?
locally

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -36,6 +36,7 @@ import org.graylog2.plugin.inputs.annotations.ConfigClass;
 import org.graylog2.plugin.inputs.annotations.FactoryClass;
 import org.graylog2.plugin.inputs.codecs.AbstractCodec;
 import org.graylog2.plugin.inputs.codecs.CodecAggregator;
+import org.graylog2.plugin.inputs.codecs.MultiMessageCodec;
 import org.graylog2.plugin.inputs.transports.NettyTransport;
 import org.graylog2.plugin.journal.RawMessage;
 import org.joda.time.DateTime;
@@ -45,11 +46,13 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
 
 @Codec(name = "gelf", displayName = "GELF")
-public class GelfCodec extends AbstractCodec {
+public class GelfCodec extends AbstractCodec implements MultiMessageCodec {
     private static final Logger log = LoggerFactory.getLogger(GelfCodec.class);
     private static final String CK_DECOMPRESS_SIZE_LIMIT = "decompress_size_limit";
     private static final int DEFAULT_DECOMPRESS_SIZE_LIMIT = 8388608;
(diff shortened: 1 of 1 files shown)


danotorrey:
Thank you for your submission @darashenka. Understood. May I ask where the log messages are originating from? I would just like to make sure you are aware that we do already support both GELF TCP and UDP inputs, which allow messages to be streamed to Graylog with lower overhead than GELF HTTP. Are these an option for your use case?

Paging @kroepke @bernd for feedback. If this were to be supported in Graylog, I think an entirely new "GELF HTTP (batched)" input with associated codec would need to be created for backwards compatibility purposes.

bernd:
@danotorrey I just took a quick look at the change but it looks backwards compatible to me. (we need to verify that, though!)

But I see a different problem. With this change the GELF HTTP input will accept the following two payloads:

**Single Message**

```json
{"version": "1.1", "host": "..."}
```

**List of Messages**

```json
[
  {"version": "1.1", "host": "..."},
  {"version": "1.1", "host": "..."},
  {"version": "1.1", "host": "..."}
]
```

When sending a list of messages, the client and also the server need to load the payload into memory before sending/parsing it. (at least as it is implemented at the moment) This can lead to memory problems on the client and server side and thus to potential message loss if the payload is too large.

We could add size limits to the input to avoid memory issues, but it's always hard to find good defaults and also for the user to choose a value that makes sense. There will always be that one payload that is slightly too large and then the user is losing messages.

My suggestion is to introduce a different HTTP endpoint in the GELF HTTP input to accept a stream of newline delimited JSON messages. That way the client can stream these messages to the server without building one single large JSON message and the server can parse it without loading one single large JSON message into memory.

The current GELF HTTP input accepts messages at the `/gelf` endpoint. We could introduce a `/gelf/bulk` enpoint (or something similar) to handle the stream of JSON messages. This would also preserve backwards compatibility.

danotorrey:
Thanks @bernd! Makes sense. Are you thinking of short streaming bursts with a single HTTP 1.1 request, or using something else like HTTP v2? I think generally 1.1 request/responses are not supposed to be long-lived.

I think if we were to go the streaming route, we would want singular GELF JSON objects to be streamed in (as opposed to an array), so I don't think JSON size would be an issue any longer. Let me know if I am missing something from your last comment.

bernd:
@danotorrey 

> Thanks @bernd! Makes sense. Are you thinking of short streaming bursts with a single HTTP 1.1 request, or using something else like HTTP v2? I think generally 1.1 request/responses are not supposed to be long-lived.

Yes, single HTTP 1.1 requests that stream the payload. (batch requests) That way the client doesn't have to read all data into memory and the server can also stream the payload instead of having to read it into memory completely.

> I think if we were to go the streaming route, we would want singular GELF JSON objects to be streamed in (as opposed to an array), so I don't think JSON size would be an issue any longer. Let me know if I am missing something from your last comment.

If a single JSON message is **really** big, we can still run into memory issues. :wink: But I think that is okay for now.

jeremyjpj0916:
Was thinking to get this functionality in sooner rather than later, one way would be to provide a boolean enable/disable option on the GELF HTTP endpoint UI(defaulted to disabled) lets call it ```multi-message``` or something, when enabling it with a warning about with great power comes great responsibility etc. , basically a statement about we will support multi batch message JSON input BUT be responsible and don't try to send 20 MB of HTTP JSON data through a single POST 😄 and warning that doing so may lead to loss of logs . Would that be a PR that @bernd and Graylog team would allow?

cc @darashenka

Telavian:
Is there a timeframe that this will be merged?

SergeyKulgan:
> Thanks @bernd! Makes sense. Are you thinking of short streaming bursts with a single HTTP 1.1 request, or using something else like HTTP v2? I think generally 1.1 request/responses are not supposed to be long-lived.
> 
> I think if we were to go the streaming route, we would want singular GELF JSON objects to be streamed in (as opposed to an array), so I don't think JSON size would be an issue any longer. Let me know if I am missing something from your last comment.

So, guys, do we have now possibility to send list/array/multiple messages over Http?
I'm using Graylog Server 3.1.3.
I really need this feature due to performance issues. TCP or UDP is not applicable for us due to some security and other reasons.
So why you restricting us to do that?
Another question, about streaming over Http. Is it possible now, can I send multiple messages in same stream not closing connection? 
My goal is sending multiple messages (batching) over http in single request (keeping connection open). Our load demand is >500 messages per second and sending each in separate request with opening and closing connection every time - is absurd. 
We experiencing serious performance-related errors like SNAT port exhaustion and other connection-related errors.

ciastooo:
> Is there a timeframe that this will be merged?

Bump

a-reznic:
any news?

kroepke:
Hey, we cannot merge this without a signed CLA, I'm afraid.

ciastooo:
@darashenka could you do it? it would be really a life saver to have this merged

a-reznic:
maybe somebody make this? maybe @darashenka stoped write code or something else?

macaty:
pls add this feature

danotorrey:
Posting this here for visibility:

This was implemented in https://github.com/Graylog2/graylog2-server/pull/11529 and released in the recent Graylog 4.3 release. 

https://www.graylog.org/post/announcing-graylog-v4-3-graylog-operations-graylog-security

Usage docs are available here: 
https://docs.graylog.org/docs/ingest-gelf

jeremyjpj0916:
Glad to see this was eventually handled. Velocity a bit slow but cool to see stuff community asked for eventually get in there.

PR #5936, opened 2019-05-12
Prevent logging ReadTimeoutException as an error if KeepAlive is conf…

jesseczko:
…igured by the user.

It's reasonable to let users configure a timeout for keepalive connections but this event should not be logged as an error when the time expires.

## Description
The user can configure the GelfHTTPInput to timeout after a certain amount of time. The input also supports tcp keep alive. If both of these settings are enabled ( say timeout after 1 hour and keepalive ) then an error will be logged in the graylog server logs when the persistent connection is terminated by netty ( ``ReadTimeoutException``)

I don't think this timeout should log an error in the internal graylog server logs. 

## Motivation and Context
I saw the read timeout exception messages in the server logs and was worried messages were getting dropped. This change will change the log message to be a ``TRACE`` level instead of ``ERROR``

## How Has This Been Tested?
Manual testing
* I built Graylog locally and verified that the error message shows up before the change on timeout. After the fix the message only shows up at trace level

Automated testing
* The cost/value of unit testing seems to be low in this case..there would be a lot of mocking and don't see any examples in the code base of unit testing logger statements. Open to suggestions though

Impact to other code
* None: this code simply changes error logging to trace logging for a specific case

Testing Environment
* MacOSX Mojave

After the fix
image

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [] I have added tests to cover my changes.
* ran the existing tests tests. Only ones that failed were some IT tests that expect something listening on 19200 which I don't have setup.. everything else passed

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/netty/ExceptionLoggingChannelHandler.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/netty/ExceptionLoggingChannelHandler.java
@@ -18,16 +18,24 @@ package org.graylog2.inputs.transports.netty;
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.timeout.ReadTimeoutException;
 import org.graylog2.plugin.inputs.MessageInput;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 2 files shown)


mpfz0r:
Thank you for your contribution @jesseczko I've moved this into a new PR https://github.com/Graylog2/graylog2-server/pull/7819

Closing this one.

PR #5956, opened 2019-05-21
Fix auto-completion in several drop-down fields

### LINKED ISSUE #5659: Auto-complete on Lookup Table creation page is broken

danotorrey:
Backport of #5665 

In all these cases the matchProp needs to be set to "label" and not "value"
because the value contains the object ID or another identifier.

After this, the auto-completion is looking at same data that the user
is seeing.

Cherry-picked from fbb6923154b25d747298dcf7fee35edf6ad590d4 and resolved a few merge conflicts.


danotorrey:
@bernd I went ahead and created a backport for this last dropdown field auto-completion fix.

PR #5957, opened 2019-05-21
Create Pipeline Rules syntax highlighting mode for Ace

kyleknighted:
## Description
Created custom Mode for Ace Editor to introduce syntax highlighting for Pipeline Rules

## Motivation and Context
* User request
* Improve code readability 

**Before:**
image

**After:**
image

## Types of changes

## Checklist:


kyleknighted:
Errors currently in master

```
client.js:772 POST http://localhost:9000/api/system/pipelines/rule/parse 400 (Bad Request)

FetchProvider.js:19 There was an error fetching a resource: cannot POST http://localhost:9000/api/system/pipelines/rule/parse (400). Additional information: Not available
```

After adding in the syntax highlighting, an additional error is added to the mix, even though the mode actually loads fine and renders appropriately

```
index.js:3802 GET http://127.0.0.1:8080/system/pipelines/rules/mode-pipeline.js net::ERR_ABORTED 404 (Not Found)
```

I could use an additional set of eyes to help me figure out what's causing this 404 even though it actually works.

graylog2-web-interface/src/components/common/mode-pipeline.js
```diff
@@ -0,0 +1,103 @@
...
+  constructor() {
+    super();
+
+    const keywords = 'let|rule|match|pipeline';
+
+    const operators = 'and|or|not';
+
+    const builtinConstants = 'all|either|during|when|then|end';
+
+    const builtinFunctions = 'to_bool|to_double|to_long|to_string|to_map|is_bool|is_number|is_double|is_long|is_string|'
```
bernd: @kyleknighted What do you think about making this list dynamic? We could create an API endpoint that returns all builtin functions. That way we can't forget to add a function here.

And since pipeline functions are pluggable, using a hardcoded list would only work for the core functions. With a dynamic list, functions that are provided by plugins would also be properly colored.

kyleknighted: > We could create an API endpoint
Probably not necessary, we could use the same endpoint that populates the documentation on the pipeline rules page

> pipeline functions are pluggable
Great to know!

> making this list dynamic
I think it's a great idea, I'll toy with the possibilities of that... not sure how feasible it is considering Ace Editor needs that data when it loads.

graylog2-web-interface/src/components/common/mode-pipeline.js
```diff
@@ -0,0 +1,103 @@
+import URLUtils from 'util/URLUtils';
```
edmundoa: I think the file name could be more descriptive or even show an indication that this file is related to `SourceCodeEditor`. What do you think about renaming it or putting it in a folder where it's clearer what it is?

graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -15,6 +15,8 @@ import 'brace/mode/text';
 import 'brace/mode/yaml';
 import 'brace/theme/tomorrow';
 import 'brace/theme/monokai';
+
+import PipelineRulesMode from './mode-pipeline.js';
```
edmundoa: Running a production build with the change I am seeing this error on the console:

`Uncaught SyntaxError: Unexpected token <` on `mode-pipeline.js:1`

Looking a bit more, I am seeing a network request to `/system/pipelines/rules/mode-pipeline.js`, which does not match anything on the server, so it returns the web interface's HTML document.

I have the suspicion that this is caused because `SourceCodeEditor` is loaded through `loadable-components`, but I'm not 100% sure about it.

edmundoa: Anyway, the highlighting appears to work even with the console error, and makes it nicer to read the code! 🎉

Issue #4583, opened 2018-02-15
Allow dots in field names again


joschi:
## Expected Behavior

When using Elasticsearch 5.x or 6.x, Graylog should allow using dots in field names again.

## Current Behavior

Dots in field names are converted to underscores by Graylog.

## Possible Solution

Allow dots in field names again.

In order to not break existing extractors and pipeline rules the conversion should be optional.

The default for this setting should be *disabled*, because otherwise new users will miss to disable it and still have the dots in field names converted to underscores.

## Steps to Reproduce (for bugs)

1. Create a GELF UDP input.
2. `echo -n '{ "version": "1.1", "host": "example.org", "short_message": "A short message", "level": 5, "_with.dot": "foo" }' | nc -w0 -u graylog.example.com 12201`
3. Search query: `with.dot:foo`
4. No results.

## Context

https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/release-notes-5.0.0.html

Refs elastic/elasticsearch#19443
Refs #1667
Refs #1721
Refs #2939

loyd:
Any progress?

gimmic:
I think this should either be allowed via configuration override or as a detected method based on backend elasticsearch version.

joseraeiro:
Any progress on this? I really need this, badly.

patrickmann:
@joseraeiro 
> Any progress on this? I really need this, badly.
Please describe your use case and the impact of not being able to use dots in field names. It will help us to properly prioritize the issue.

gimmic:
> @joseraeiro
> 
> > Any progress on this? I really need this, badly.
> 
> Please describe your use case and the impact of not being able to use dots in field names. It will help us a lot in properly prioritizing the issue.

I've since moved on, but this is a good example of where dots would be incredibly useful: following elastic common schema.

https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html

joseraeiro:
> @joseraeiro
> 
> > Any progress on this? I really need this, badly.
> 
> Please describe your use case and the impact of not being able to use dots in field names. It will help us a lot in properly prioritizing the issue.

When trying to read the indexes on Wazuh's Kibana the fields are not present in the UI. And that's a big issue.

Bilal-SG:
[...]

Issue #668, opened 2014-08-11
GELF spec 1.2


bernd:
The GELF spec states the following about the `timestamp` field.

> UNIX microsecond timestamp; SHOULD be set by client library. Will be set to NOW by server if absent.

The GELF parser in graylog2-server parses the `timestamp` field of incoming messages like this:

``` java
double messageTimestamp = doubleValue(json, "timestamp");
DateTime timestamp;
if (messageTimestamp <= 0) {
    timestamp = Tools.iso8601(); // current time
} else {
    // we treat this as a unix timestamp
    timestamp = Tools.dateTimeFromDouble(messageTimestamp);
}

...

/*
 * The double representation of a UNIX timestamp with milliseconds is a strange, human readable format.
 *
 * This sucks and no format should use the double representation. Change GELF to use long. (zomg)
 */
public static DateTime dateTimeFromDouble(double x) {
    return new DateTime(Math.round(x*1000), DateTimeZone.UTC);
}
```

All of the client libraries we checked use something like this for the timestamp:

``` java
double timestamp = logEvent.getTimeStamp() / 1000.0;
```

So in reality the `timestamp` field is a UNIX timestamp in **seconds** with optional **milliseconds** after the decimal point.

That means that either the GELF 1.1 specification is wrong or everyone (including ourselves) implemented it incorrectly.

Two suggestions:
1. Adjust the `timestamp` field specification to match reality: "Seconds since UNIX epoch; SHOULD be set by client library. Will be set to the current system time by server if absent.".
2. Add an optional `timestamp_ms` field which holds a millisecond timestamp. If `timestamp_ms` exists in an incoming message, that one should be used as timestamp.

This ensures backwards compatibility with existing clients but also allows fine resolution timestamps to be sent.

Any comments? Any other changes for a version 1.2 of the GELF spec?

gbu-censhare:
Compression of GELF Messages does not work atm for TCP Connections.

Maybe a stream Compression would be more efficient here, i opened a pull request for this (closed it in the meantime, as i wanted to first check it against GELF-MessagePerMessage Compression, which didn't work :-( )
 
https://github.com/Graylog2/graylog2-server/pull/1364

gbu-censhare:
and: what about supporting Arrays and Objects?

i have a metric i'd like to send in

we have a function call, where different listeners are called. and i want to know which listener is taking how long.

so normaly i would do it this way:

[...]

PR #5983, opened 2019-05-31
Allow dots in field names.

iinuwa:
## Description

This removes the features that filter and rename field names on ingestion. 
## Motivation and Context

Now that Elasticsearch supports dots in field names to delimit nested objects and Graylog 3.0 has dropped support for Elasticsearch versions without this feature, this PR re-enables dots in field names. 

Fixes. #4583 
## How Has This Been Tested?
The tests should all pass. I have also done an informal test in my environment, and the nested object mappings are created in Elasticsearch as expected.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (Cf. graylog2/documentation@e60ad962)

I am not sure how the documentation should be modified, but I created a proposed file change in the documentation repo to be published if this is accepted in Graylog 3.1.0.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -45,7 +45,10 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 @Codec(name = "gelf", displayName = "GELF")
@@ -205,7 +208,15 @@ public class GelfCodec extends AbstractCodec {
             final JsonNode value = entry.getValue();
 
             final Object fieldValue;
-            if (value.isContainerNode()) {
+            if (value.isArray()) {
+                List<String> fieldArray = new ArrayList<>();
+                
+                for (final JsonNode objNode : value) {
+                    fieldArray.add(objNode.asText());
+                }
+                fieldValue = fieldArray;
+            }
+            else if (value.isContainerNode()) {
                 fieldValue = value.toString();
             } else if (value.isFloatingPointNumber()) {
                 fieldValue = value.asDouble();
--- a/graylog2-server/src/main/java/org/graylog2/inputs/extractors/JsonExtractor.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/extractors/JsonExtractor.java
@@ -85,7 +85,7 @@ public class JsonExtractor extends Extractor {
 
         this.flatten = firstNonNull((Boolean) extractorConfig.get(CK_FLATTEN), false);
         this.listSeparator = firstNonNull((String) extractorConfig.get(CK_LIST_SEPARATOR), ", ");
-        this.keySeparator = firstNonNull((String) extractorConfig.get(CK_KEY_SEPARATOR), "_");
+        this.keySeparator = firstNonNull((String) extractorConfig.get(CK_KEY_SEPARATOR), ".");
         this.kvSeparator = firstNonNull((String) extractorConfig.get(CK_KV_SEPARATOR), "=");
         this.replaceKeyWhitespace = firstNonNull((Boolean) extractorConfig.get(CK_REPLACE_KEY_WHITESPACE), false);
         this.keyWhitespaceReplacement = firstNonNull((String) extractorConfig.get(CK_KEY_WHITESPACE_REPLACEMENT), "_");
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -247,32 +247,7 @@ public class Message implements Messages {
             }
 
             final Object value = entry.getValue();
-            // Elasticsearch does not allow "." characters in keys since version 2.0.
-            // See: https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#_field_names_may_not_contain_dots
-            if (key.contains(".")) {
-                final String newKey = key.replace('.', KEY_REPLACEMENT_CHAR);
-
-                // If the message already contains the transformed key, we skip the field and emit a warning.
-                // This is still not optimal but better than implementing expensive logic with multiple replacement
-                // character options. Conflicts should be rare...
-                if (!obj.containsKey(newKey)) {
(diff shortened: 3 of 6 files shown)


kroepke:
We do plan on allowing nested objects in the future, however we cannot simply revert the existing behavior and change extractors because that will break existing users' setups. Pipeline functions might need changes as well.

In addition displaying nested objects opens up all kinds of issues, too, which aren't addressed here.

Data semantics are also an issue, for example to make sense of arrays in nested objects, we will need to support nested aggregations, which is not straightforward to implement because the query engine needs the exact mapping type or it will generated invalid queries.

I'm reluctant to close this PR because I support the goal, however I believe that nested object support is much more involved than just allowing indexing them.

I'll leave it open for now for the sake of discussion, but please don't be disappointed if this does not get merged any time soon.

iinuwa:
No, I completely understand; this PR was mostly to get the ball rolling on a feature that we need in my organization. Glad there's a plan for it!

Some thoughts I had:

We haven't implemented Graylog yet, so I'm happy to help with greenfield testing, but I imagine that the harder part will be supporting existing setups as you say. I wonder if it may be useful to publish a  new spec for GELF with support for arrays and objects (cf [#668](https://github.com/Graylog2/graylog2-server/issues/668#issuecomment-251902866)). That may provide a migration path by allowing users to configure GELF inputs to use a specific GELF spec. That way, the inputs defined with the 1.1 spec would still process as expected, and 1.2 inputs could use the new functionality.

IMO, displaying nested objects with dot-separated field names in the UI is pretty natural. I believe that is how Elasticsearch queries reference the fields and how Kibana displays them as well. That may cause issues, though, because there wouldn't be a visible difference between field names with dots in them and fields of nested objects.

I ran into the problem with defining the data structure up front pretty quickly while testing this. It would most likely have to be defined up front for aggregation.  I saw a proposal on IRC (I believe from @gimmic) to have a way to define the Elasticsearch index templates from the web app. That may help to make it easier to define the data structure for the indices up front so that the log messages are ingested correctly.

I guess we'd also have to come up with a syntax for doing advanced queries of nested objects.(Elasticsearch/Lucene doesn't even have that (cf. elasticsearch/elasticsearch#11322)

Thanks for opening it up for discussion!

gimmic:
I do think it is strange having graylog rename fields for backwards compatibility which is no longer required. The field rename has confused me with parsing rules now more than once until I look at the raw input/field value(because of the way the graylog web interface shows the renamed value). 

As backwards compatibility is dropped, it would make sense to sunset this transparent-renaming feature as well, maybe invert it with a flag, or provide a checkbox in the UI to enable/disable this feature. 

I also obviously support editing of templates via the web UI but that's an entire new feature request.

kroepke:
Sorry, this turned into an essay :/

Indeed, ECS support (or something similar to it) is one of the motivations for supporting them.
Let me explain the rationale (this will get philosophical quickly, please bear with me):
Initially Graylog didn't care what you sent, as long as it could be mapped to what GELF said, this predates our use of elasticsearch even (so pre 2012 IIRC).
Now, Graylog never supported you in querying or displaying those fields well.
When in the Elasticsearch 2.x time frame dots where disallowed and allowed again, we had to make an impossible choice: either mangle the field names or drop the data entirely. Neither is a good option, so we chose the former.

This was bad then and is bad now, IMHO, because it needs to be done just prior to sending it Elasticsearch due to backwards compatibility, however no one ever came up with a better way of doing it, so we never changed it again.

Now the philosophy part:
What do people even want to accomplish with (nested) objects in the context of log management?
AFAICS there are two things:
 1. Several fields are so closely related that they are (almost) always used or manipulated together, such as geo information, ip/port, results of certain lookups like whois results
 2. There's an unknown or repeating part in the event/message that one does not want to see as a top level field (or even worse a series of fields suffixed with an index), because since it's unknown it really is only useful for displaying it.

Nested objects as in Elasticsearch's mapping type is rarely what people really want, at least not in logging, as it has to create one document per array index and comes with a few restrictions.

That being said, scenarios where I've heard requests for nested field support are ECS support, generic JSON documents and messages that have fields containing a list of things, such as a stack trace.

ECS support falls into scenario 1 from above, it provides a logical grouping that enables treating certain attributes as one, which totally makes sense from a processing and display point of view.

Generic JSON support is something that sounds great on paper, but then it's rare that people can actually articulate how they intend to make sense of it. Querying is difficult, aggregations are hopelessly complex to express (do you need JSONPath to extract something, how about arrays, objects, what about data types like IP addresses that are represented as simple strings etc). The same goes for XML messages.

Things like stack traces are interesting, because they are a combination of "contains interesting information" and "I just need to display it properly for analysis".
It's probably the closest in spirit to log management use cases.

Now, where does this leave us?
We know that something like ECS support is a sensible request and could benefit us a lot.
Also various beats send some kind of nested data, that follow the same rationale (rtt fields in heartbeat for example).
Generic JSON/XML document support sounds nice, but in reality it's sort of an edge case, because not all searches that seem natural actually make sense (and don't really do what people think they do).
Lastly stacktraces that are delivered as a list of lines really are a string and should probably be treated as such. In general anything that relies on a specific order in an array is pretty much an anti-pattern.
Essentially in the last two cases what you really want is to store verbatim what you received and do some kind of preprocessing to extract the valuable data that you might be searching on. For stacktraces that's likely the exception that was the root cause and _maybe_ its line information. Certainly not the entire stack.

Hope that sheds some light on what we are thinking about this.

kroepke:
> I do think it is strange having graylog rename fields for backwards compatibility which is no longer required. The field rename has confused me with parsing rules now more than once until I look at the raw input/field value(because of the way the graylog web interface shows the renamed value).

The backwards compatibility concern really is that we cannot silently change behavior in running setups.
I fully understand that for new users this seems silly, but that's the nature of supporting decisions from the past.
Please note that I'm not trying to argue that we don't want to allow fields like these, but we need to make sure that we don't screw up existing users, which is very important to us.

gimmic:
Yeah, I get it. At some point you will simply have to introduce a breaking change. Thankfully, this is a fairly minor breaking change compared to some of the stunts Elastic have pulled over the years. I think given sufficient notification this should be something the community could generally fix or otherwise have an option to toggle. The old functionality could be emulated on upgraded instances.

iinuwa:
@kroepke, thanks for taking the time to explain your thoughts.

My eventual goal is to be able to send messages using ECS as well. I would be fine with just having dotted field names for now until we can get actual nested objects (fake it 'til you make it.)

For this PR, would it be enough to add an option to rename `.` to `_` in field names and set it to be on by default? That would preserve the current functionality while allowing new users to take the new functionality. Maybe later on, it could be marked as deprecated, and then removed if/when we have a clear way to migrate to use the new functionality. (I can also remove the commit for parsing array values as keywords—didn't mean to include that here.)

PR #5998, opened 2019-06-06
Wizard component style customization

edmundoa:
This change allows customers of the `Wizard` component to customize its appearance in two different ways:

- The `justified` prop allows the navigation to take the full width of its container
- The `navigationClassName` prop allows consumers to set a CSS class for the navigation, so it is possible to alter the way the navigation looks without changing its code

Screenshot:

preview

### DIFF
--- a/graylog2-web-interface/src/components/common/Wizard.jsx
+++ b/graylog2-web-interface/src/components/common/Wizard.jsx
@@ -55,36 +55,45 @@ class Wizard extends React.Component {
   }
 
   _wizardChanged = (eventKey) => {
-    this.props.onStepChange(eventKey);
+    const { onStepChange } = this.props;
+    onStepChange(eventKey);
     this.setState({ selectedStep: eventKey });
   };
 
   _disableButton = (direction) => {
-    const len = this.props.steps.length;
+    const { steps } = this.props;
+    const { selectedStep } = this.state;
+    const len = steps.length;
     const disabledPosition = direction === 'next' ? (len - 1) : 0;
-    const currentPosition = this.props.steps.findIndex(step => step.key === this.state.selectedStep);
+    const currentPosition = steps.findIndex(step => step.key === selectedStep);
     const otherPosition = direction === 'next' ? (currentPosition + 1) : (currentPosition - 1);
-    const otherStep = (this.props.steps[otherPosition] || {});
-    return this.props.steps[disabledPosition].key === this.state.selectedStep || otherStep.disabled;
+    const otherStep = (steps[otherPosition] || {});
+    return steps[disabledPosition].key === selectedStep || otherStep.disabled;
   };
 
   _onNext = () => {
-    this._wizardChanged(this.props.steps[this._getSelectedIndex() + 1].key);
+    const { steps } = this.props;
+    this._wizardChanged(steps[this._getSelectedIndex() + 1].key);
   };
 
   _onPrevious = () => {
-    this._wizardChanged(this.props.steps[this._getSelectedIndex() - 1].key);
+    const { steps } = this.props;
+    this._wizardChanged(steps[this._getSelectedIndex() - 1].key);
   };
 
   _getSelectedIndex = () => {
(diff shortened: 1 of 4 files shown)


kyleknighted:
Could we add a `activeStep`, or whatever we want to call it, as a prop?
That way i can just pass an integer and have it render the appropriate step.

edmundoa:
I think we could add that. So far the component behaves in an uncontrolled way https://reactjs.org/docs/uncontrolled-components.html, but we can add that some props to allow consumers to also use it in a controlled manner.

kyleknighted:
The problem with it being uncontrolled is that if a user wants to go back three steps to edit something,

they will have to click a "Previous" button multiple times, instead of just clicking the a nav item showing the step they want to direct to.
Alternatively, we will have to keep duplicative track of the current step in state so that we can loop through and automate the clicking of the buttons

Also, we need to be able to disable the buttons in this component and be able to have alternative buttons in the view be able to continue on to the next step.

IMO this should be a controlled component - or we should create an alternative version that is controlled, if we need this component to remain uncontrolled.

edmundoa:
From what I can see, the current component already supports jumping to a certain step when clicking in the `NavItem`, so the user does not have to click several times on "preview" or "next" to achieve that.

I agree that sometimes it makes more sense to leave the consumer of the `Wizard` to handle the state and decide which step should be rendered, so I'm working on that :)

graylog2-web-interface/src/components/common/Wizard.jsx
```diff
@@ -114,30 +147,40 @@ class Wizard extends React.Component {
...
           })}
         </Nav>
       </Col>
     );
   };
 
+  _getSelectedStep() {
+    const { activeStep } = this.props;
+    const { selectedStep } = this.state;
+    return lodash.defaultTo(activeStep, selectedStep);
```
kyleknighted: To avoid loading in lodash, could this just be `return activeStep || selectedStep;` ?

edmundoa: The only problem I see is if `activeStep` is a number, and for some reason needs to be 0. We can always restrict the key to only be a string, but for now I thought it was safer going with `defaultTo()`.

graylog2-web-interface/src/components/common/Wizard.jsx
```diff
@@ -114,30 +147,40 @@ class Wizard extends React.Component {
...
   render() {
-    const rightComponentCols = this.props.horizontal ? 5 : 3; // If horizontal, use more space for this component
+    const { steps, horizontal, containerClassName, children } = this.props;
+    const rightComponentCols = horizontal ? 5 : 3; // If horizontal, use more space for this component
     return (
-      <Row className={this.props.containerClassName}>
-        {this.props.horizontal ? this._renderHorizontalStepNav() : this._renderVerticalStepNav()}
+      <Row className={containerClassName}>
+        {horizontal ? this._renderHorizontalStepNav() : this._renderVerticalStepNav()}
         <Col md={7}>
```
kyleknighted: Not sure if this component should dictate content width with the `col-md-7` class that it generated.

image

edmundoa: The original purpose of that was to leave space on the right side to render a preview, which I think is fine. The page you are working on is also the first one using some centered content, making this problem more apparent than if we added some left-side content.

I think we can add a prop to decide if a preview should be rendered or not, and use the full container width if there's no preview. In that way consumers can decide if they want to render a preview on the right, even if that side of the screen is initially empty. Would that solve that issue for you?

kyleknighted: Well, `children` should be present if there is a preview, so we shouldn't really need a prop for it. Just need to do something similar to line 215 probably.

edmundoa: I was thinking in case the initial preview had no content, but you are right, having a preview without content doesn't make much sense. Will take care of that 👍

PR #6010, opened 2019-06-12
Implement events index template

### LINKED ISSUE #5884: Alert Server Changes - Index-Set Index Templates
Index sets currently use a message specific Elasticsearch index template but for events we need a different index template. We have to add support for different index templates to index sets. This doesn't need to be a full-blown index template editor, being able to select if an index set should use the "messages" or "events" template should be enough.

bernd:
This introduces index template types so we can install different index mapping templates for different index sets depending on the value of `index_template_type`.

This also adds an "events" index template that will be used for the events/alerting system.

Closes #5884

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/EventsIndexMapping.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/EventsIndexMapping.java
@@ -1,0 +1,223 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.indexer;
+
+import com.github.zafarkhaja.semver.Version;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import org.graylog2.indexer.indexset.IndexSetConfig;
+
+import java.util.Map;
+
+public class EventsIndexMapping implements IndexMappingTemplate {
+    private final Version elasticsearchVersion;
+
+    EventsIndexMapping(Version elasticsearchVersion) {
+        this.elasticsearchVersion = elasticsearchVersion;
+    }
+
+    @Override
+    public Map<String, Object> toTemplate(IndexSetConfig indexSetConfig, String indexPattern, int order) {
+        final String indexPatternsField;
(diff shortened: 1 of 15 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/EventsIndexMapping.java
```diff
@@ -0,0 +1,223 @@
...
+                                                .put("type", "keyword")
+                                                .build())
+                                        .put("creator_id", map()
+                                                .put("type", "keyword")
+                                                .build())
+                                        .put("timestamp", map()
+                                                .put("type", "date")
+                                                // Use the same format we use for the "message" mapping to make sure we
+                                                // can use the search.
+                                                .put("format", "yyyy-MM-dd HH:mm:ss.SSS")
```
dennisoelkers: As a sidenote: I am wondering if it would make sense to change this to include the timezone in the future. While it does not necessarily introduce bugs to leave it out and use UTC everywhere, it requires everyone indexing a message to adhere to this, while storing the timestamp with the timezone does not.

dennisoelkers: Should we reuse `Tools.ES_DATE_FORMAT` here?

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
```diff
@@ -117,6 +125,10 @@
     @NotBlank
     public abstract String indexTemplateName();
 
+    @JsonProperty("index_template_type")
+    @NotBlank
+    public abstract TemplateType indexTemplateType();
```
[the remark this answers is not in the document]
bernd: Good idea, will do. :+1:

graylog2-server/src/main/java/org/graylog2/indexer/EventsIndexMapping.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Override
+    public Map<String, Object> toTemplate(IndexSetConfig indexSetConfig, String indexPattern, int order) {
+        final String indexPatternsField;
+        final Object indexPatternsValue;
+
+        if (elasticsearchVersion.satisfies("^5.0.0")) {
+            indexPatternsField = "template";
+            indexPatternsValue = indexPattern;
+        } else if (elasticsearchVersion.satisfies("^6.0.0")) {
```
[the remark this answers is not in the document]
bernd: I think we should keep the strict version requirements even though we probably don't need to change anything for ES 7. We always have to be really careful we supporting a new ES version and having strict version requirements forces us to think about that.

graylog2-server/src/main/java/org/graylog2/indexer/EventsIndexMapping.java
bernd: Totally, yes. :+1:

bernd: I have been experimenting with several builtin date formats here but the problem is, that our searches won't work at the moment when we use a different date format in ES. :confused: 

So changing this will probably result in a lot of work. ES is storing all dates as UTC internally (it's basically a long) so not sure if it's worth chaning it. But I agree that dates without timezone always look fishy. :wink:

dennisoelkers: Maybe we can address this after replacing the existing search with the views, I think we have a few more options there.

PR #6011, opened 2019-06-13
Merging Views into Open Source Core.

dennisoelkers:
## Description

This PR is introducing the former enterprise-only feature called "Views" into the open source core. Views are a new way to search and analyze your data. It is used for interactive, reproducible, in-depth investigation of your log data, with search capabilities that go beyond the current search. Namely:

  * Being able to select multiple streams to search in
  * Create graphs containing an arbitrary number of metrics and row/column pivots to break down and visualize your data the way you want it
  * Full customization of your message tables by including only the fields you need, in the order you want
  * On-demand highlighting of values, allowing you track these through your data
  * All of this can be saved, shared and reused, without having to recreate it

We are open-sourcing all of this, to enable everybody to make better use of Graylog, to make the product valuable even for the smallest projects and to gather feedback from our beloved community as part of our constant effort to improve Graylog. Have fun!

**Important: Do not squash this PR when merging or history will be lost.**

## Types of changes

## Checklist:


dennisoelkers:
@bernd: It shows only if you have the enterprise plugin installed, but no valid license.

bernd:
> @bernd: It shows up only if you have the enterprise plugin installed, but no valid license.

Ah, makes sense. Thanks for the clarification!

bernd:
@dennisoelkers Good work! :+1: Merging it now.

PR #6034, opened 2019-06-24
Updating retrofit from 2.4.0 to 2.6.0.

dennisoelkers:
## Motivation and Context

This change is updating Retrofit from 2.4.0 to 2.6.0. The current version is fixing a XXE vulnerability, described here[[0]]. It most probably has no impact on the product, as we do not use XML in the places where Retrofit is used, but it is unclear if e.g. a MitM could force the Retrofit client to interpret XML and therefore trigger the vulnerability.

The changelog for 2.6.0 is found here[[1]].

[0]: https://nvd.nist.gov/vuln/detail/CVE-2018-1000844
[1]: https://github.com/square/retrofit/blob/parent-2.6.0/CHANGELOG.md

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #6035, opened 2019-06-24
Updating OkHttp from 3.11.0 to 3.14.2.

dennisoelkers:
## Motivation and Context

This PR is updating OkHttp to 3.14.2.

Any OkHttp version up to 3.12.0 include a vulnerability that allows an attacker doing MitM to bypass certificate pinning. It does require full control of the Graylog server before, as it requires the attacker to control the startup/process space of the Graylog process though.

Changelog: https://github.com/square/okhttp/blob/parent-3.14.2/CHANGELOG.md
Bug Report: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-20200

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/shared/bindings/providers/OkHttpClientProviderTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/shared/bindings/providers/OkHttpClientProviderTest.java
@@ -92,7 +92,7 @@ public class OkHttpClientProviderTest {
         assertThat(server.getRequestCount()).isEqualTo(1);
         final RecordedRequest recordedRequest = server.takeRequest();
         assertThat(recordedRequest.getMethod()).isEqualTo("GET");
-        assertThat(recordedRequest.getPath()).isEqualTo("http://www.example.com/");
+        assertThat(recordedRequest.getPath()).isEqualTo("/");
         assertThat(recordedRequest.getHeader(HttpHeaders.HOST)).isEqualTo("www.example.com");
     }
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog2/shared/bindings/providers/OkHttpClientProviderTest.java
```diff
@@ -92,7 +92,7 @@ public void testSuccessfulProxyConnectionWithoutAuthentication() throws IOExcept
         assertThat(server.getRequestCount()).isEqualTo(1);
         final RecordedRequest recordedRequest = server.takeRequest();
         assertThat(recordedRequest.getMethod()).isEqualTo("GET");
-        assertThat(recordedRequest.getPath()).isEqualTo("http://www.example.com/");
+        assertThat(recordedRequest.getPath()).isEqualTo("/");
```
danotorrey: @dennisoelkers What is the reason why these URLs/paths had to be changed?

dennisoelkers: This is related to the introduction of https://github.com/square/okhttp/pull/4235 in `3.12.0`, which is fixing path parsing for IPv6. Actually it makes sense, because `http://www.example.com` is not part of the request path.

This is only related to the tests, the actual OkHttp client code has not changed in this regard.

danotorrey:
@dennisoelkers from reviewing the OkHttp change log, TLS versions 1 and 1.1 are no longer supported without a explicitly enabling them again. This might present an issue for the HTTP Alarm Callback when it communicates with older HTTP servers. 
 >TLSv1 and TLSv1.1 are no longer enabled by default. Major web browsers are working towards removing these versions altogether in early 2020. If your servers aren't ready yet you can configure OkHttp 3.13 to allow TLSv1 and TLSv1.1 connections:
https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-3130

danotorrey:
@dennisoelkers Do you know if this change will cause any issues?

>Proxy authenticators are now asked for preemptive authentication. OkHttp will now request authentication credentials before creating TLS tunnels through HTTP proxies (HTTP CONNECT). Authenticators should identify preemptive authentications by the presence of a challenge whose scheme is "OkHttp-Preemptive".
https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-3120

dennisoelkers:
> @dennisoelkers from reviewing the OkHttp change log, TLS versions 1 and 1.1 are no longer supported without a explicitly enabling them again. This might present an issue for the HTTP Alarm Callback when it communicates with older HTTP servers. Should we enable TLS 1 and 1.1 in `OkHttpClientProvider`?

I don't think so. It is a problem for anyone connecting to a TLSv1.0/1 endpoint, but these are _very_ rare these days and inherently insecure and the protocol has been deprecated in 2008. The IETF actively [prohibits(!) fallback to TLSv1.0/1 in any application](https://tools.ietf.org/id/draft-moriarty-tls-oldversions-diediedie-00.html).

We could document this in the upgrade notes though.

dennisoelkers:
> @dennisoelkers Do you know if this change will cause any issues?
> 
> > Proxy authenticators are now asked for preemptive authentication.

From my understanding this is an additional feature and not a replacement of the current proxy authentication mechanism, so no.

danotorrey:
Thanks @dennisoelkers. Understood. I'll finish testing then I'll merge this and #6034 

// Including @bernd for reference since we talked about this.

PR #5391, opened 2018-12-10
Switch Elasticsearch IT tests to use testcontainers

bernd:
This PR switches all Elasticsearch `*IT.java` tests over to use the Docker based [testcontainers](https://www.testcontainers.org/) project.

Previously we have been using `elasticsearch-maven-plugin` and a custom fork of `nosqlunit-elasticsearch-http` to write and run the IT tests for Elasticsearch. That made it hard to run the tests in development because a developer needed to start the ES instance manually. The custom fork of the `nosqlunit-elasticsearch-http` library also doesn't support ES 6.x yet and would need an update.

One major change in how we execute the tests is, that we now start a new Elasticsearch instance for every test that is running. That means the tests became pretty slow compared to the previous way of running the tests. There are some options to improve that situation in the future (e.g. start a single instance for every test class) but I would like to do that later to avoid making this PR even bigger.

Summary of benefits from this PR:

- Easier to support new Elasticsearch versions because no changes to test libraries needed
- Getting rid of a custom library fork we would have to maintain
- Easier to run IT tests in development because no manual ES start needed
- Removal of lots of indirections through the nosqlunit code makes test classes and data loading easier to understand

Summary of drawbacks from this PR:

- Slower test runs (for now)
- Adds dependency on Docker

PR #6057, opened 2019-07-01
Use `embedded-elasticsearch` in Es-related/integration tests.

dennisoelkers:
## Motivation and Context

This is not supposed to be merged, but to be discussed in order to move integration testing forward.

While reviewing #6039 a quick way to test the before/after states was needed. Unfortunately, the current Elasticsearch testing infrastructure became quite heavy after embedding a node was prohibited by elastic.
This is an attempt to utilize [embedded-elasticsearch](http://github.com/allegro/embedded-elasticsearch) to embed Es-related testing into the test codebase, making it independent from external bootstrapping/configuration and allowing a lightweight way to test with a full-fledged Elasticsearch.

The benefits of it are:

  - tests are self-contained, no need to run them in a specific environment (e.g. having to start Es using maven/docker/both before)
  - the Elasticsearch configuration is part of/accessible from the test suite, so it is possible to test with specific settings for each individual test suite/case
  - as Elasticsearch is bootstrapped for each test class (unless done differently), it makes clean slate testing (and parallelizing it) easier (as each test starts with an empty instance)

The disadvantage towards externally-managed Elasticsearch instances:

  - Per default Elasticsearch is started/stopped for each test class, resulting in longer run times

Obviously, this test bakes in bootstrapping into the actual test suite.
This is for demonstration purposes to make the required code compact and laid out in one file. If we decide to employ the library for testing, it would need to be extracted in a base class or junit rule to consolidate the bootstrapping code and its configuration.

### DIFF
--- a/graylog-project-parent/pom.xml
+++ b/graylog-project-parent/pom.xml
@@ -668,6 +668,12 @@
                 <version>${pkts.version}</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>pl.allegro.tech</groupId>
+                <artifactId>embedded-elasticsearch</artifactId>
+                <version>${embedded-elasticsearch.version}</version>
+                <scope>test</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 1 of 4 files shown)


bernd:
Related to https://github.com/Graylog2/graylog2-server/pull/5391.

alex-konn:
Closing this, as we've decided to go with `testcontainers`.

PR #6070, opened 2019-07-03
Add "tags" field to event index mapping

bernd:
This is required to be able to filter and aggregate on tags.


bernd:
We came up with a different solution. I will mark this as ready to review again when ready.

bernd:
Replaced by https://github.com/Graylog2/graylog2-server/pull/6078

PR #6078, opened 2019-07-04
Rename and add fields in EventIndexMapping

bernd:
- Rename "creator_type" to "event_definition_type" and store the event
  definition type in it
- Rename "creator_id" to "event_definition_id"
- Add "alert" field
- Add and use "origin_context" field


bernd:
I need to add some more changes and will re-add the ready-for-review label once I am done.

bernd:
This one is ready for review again.

PR #6089, opened 2019-07-05
Add migration to create events and system-events index sets and streams

### LINKED ISSUE #6085: Create migration to add events and system-events streams and index sets
## Description

The event/alerting system needs a default events index set and a system-events index set. These should be created with a migration during server startup.

## Tasks

-  Create new migration to create
    - [ ] Default events stream
    - [ ] System events stream
    - [ ] Default events index set
    - [ ] System events index set
- [ ] The events and system-events index sets need to be of type `TemplateType.EVENTS`

## Open Questions

- [ ] tbd

bernd:
This adds:

- Default events stream + index set
- Default system events stream + index -set

Closes #6085

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
@@ -17,26 +17,19 @@
 package org.graylog2.configuration;
 
 import com.github.joschi.jadconfig.Parameter;
-import com.github.joschi.jadconfig.ValidationException;
-import com.github.joschi.jadconfig.ValidatorMethod;
-import com.github.joschi.jadconfig.converters.StringListConverter;
 import com.github.joschi.jadconfig.util.Duration;
-import com.github.joschi.jadconfig.validators.FilePathReadableValidator;
-import com.github.joschi.jadconfig.validators.InetPortValidator;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
```diff
@@ -17,26 +17,19 @@
...
-import javax.validation.constraints.NotNull;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.Collections;
-import java.util.List;
 import java.util.Locale;
 
 public class ElasticsearchConfiguration {
+    public static final String DEFAULT_EVENTS_INDEX_PREFIX = "default_events_index_prefix";
```
mpfz0r: Should we add these into our `graylog.conf`?

bernd: I am a bit on the fence on this. This should **only** be used by users if they already have an index prefix of `gl-events` configured for one of their index sets. Adding it to the example `graylog.conf` might motivate users to change it. :wink:

mpfz0r: fair enough :)

PR #6090, opened 2019-07-05
Dynamic pager in message lists

kmerz:
## Description
Only display the pager in message lists, when the amount of messages is bigger
than the page size.

## Motivation and Context
We right now always show a pager, even though we never need it, since the we only
ask for max 150 messages and the page size 150.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
+++ b/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
@@ -1,6 +1,5 @@
 // @flow strict
 import * as React from 'react';
-import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
 import * as Immutable from 'immutable';
 import connect from 'stores/connect';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/views/components/widgets/MessageList.jsx
```diff
@@ -118,13 +131,16 @@ const MessageList = createReactClass({
...
     return (
       <span>
-        <div className={styles.messageListPaginator}>
-          <MessageTablePaginator currentPage={Number(currentPage)}
-                                 onPageChange={newPage => this.setState({ currentPage: newPage })}
-                                 pageSize={pageSize}
-                                 position="top"
-                                 resultCount={messages.length} />
-        </div>
+        { messages.length > Messages.DEFAULT_LIMIT
```
dennisoelkers: I think this should be `messages.length > pageSize`?

PR #6092, opened 2019-07-05
Reduce processing_status_persist_interval default value to 1 second

bernd:
Using 5 seconds is not often enough with regard to alert job execution.
Alert job execution checks this data in the database to decide if an
alert job can be executed or not.


bernd:
@mpfz0r Good catch! Fixed it.

PR #6094, opened 2019-07-08
Frontend integration tests setup

dennisoelkers:
## Motivation and Context

This PR is adding the required infrastructure for writing integration tests. What it does is:

  * Adding the `@babel/plugin-transform-runtime` plugin to enable `async`/`await`
  * Adding `testing-library`, which allows us in contrast to e.g. `enzyme` to wait for the appearance of a specified element. With `enzyme` et al it was necessary to advance the correct number of ticks to be in the desired state of rendering within the test, which is naturally undesirable.
  * Creating a custom environment that bootstraps a very basic API the frontend can connect against. (6f603c5) Functions allowing to customize/add API responses from within the test will be supplied later. The environment is opt-in and can be activated by using a pragma in the header like this: 
```
/**
 * @jest-environment <rootDir>/test/integration-environment.js
 */
```

  * Fixing an issue in the `FieldList` component which unconditionally tries to extract sizing information from an element (which might not be available in tests/JSDom). (7d77852)
  * Fixing an import issue with  `jquery-ui`, where two files did not include the required depencies. This was encountered in the `Widget` test already, but it was not fixed in the originating files, but the test itself at that time. (61f4a80)

  * Adding two test cases for creating a new view/extended search. A test case (after boilerplate) to test creating a new view from the views list looks like this:

```
  it('Create a new view using Views Page', async () => {
    const { getByText, getAllByText } = render(<AppRouter />);
    history.push(Routes.VIEWS.LIST);

    const button = await waitForElement(() => getAllByText('Create new view')[0]);
    fireEvent.click(button);
    await waitForElement(() => getByText('New View'));
  });
```
  This allows us to test extensive user stories requiring a lot of infrastructure pretty light-weight.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
```diff
@@ -72,10 +72,13 @@ const FieldList = createReactClass({
...
 
-    const maxHeight = this.props.maximumHeight
-      - fieldsContainer.getBoundingClientRect().top;
+    const { maximumHeight } = this.props;
 
-    this.setState({ maxFieldsHeight: Math.max(isNaN(maxHeight) ? 0 : maxHeight, this.MINIMUM_FIELDS_HEIGHT) });
+    if (fieldsContainer && fieldsContainer.getBoundingClientRect) {
+      const maxHeight = maximumHeight - fieldsContainer.getBoundingClientRect().top;
+
+      this.setState({ maxFieldsHeight: Math.max(isNaN(maxHeight) ? 0 : maxHeight, this.MINIMUM_FIELDS_HEIGHT) });
```
kyleknighted: should we use `Number.isNan` here?

dennisoelkers: No. While the code is not good, type coercion of `isNaN` (which `Number.isNaN` does not perform) is actually desired, as it would require checks for `undefined` et al otherwise. I think we should replace it with `lodash.isFinite` instead though.

PR #6113, opened 2019-07-10
Using function instead of `arrayMove` from `react-sortable-hoc`.

dennisoelkers:
## Motivation and Context
Before this change, `arrayMove` was used from the `react-sortable-hoc`
package. It will not be exported from it in the future anymore, so this
change is now adding a simple function doing the same.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/aggregationbuilder/SortableSelect.jsx
```diff
@@ -2,14 +2,20 @@ import React from 'react';
...
 import { findIndex } from 'lodash';
 
 import Select from 'views/components/Select';
 import styles from './SortableSelect.css';
 
 const SortableValueList = SortableContainer(components.ValueContainer);
 
+const arrayMove = (array, from, to) => {
+  const result = array.slice();
+  result.splice(to < 0 ? array.length + to : to, 0, array.splice(from, 1)[0]);
```
kmerz: result.splice(to < 0 ? array.length + to : to, 0, result.splice(from, 1)[0]);

graylog2-web-interface/src/views/components/aggregationbuilder/SortableSelect.jsx
```diff
@@ -2,14 +2,20 @@ import React from 'react';
...
+  return array;
```
kmerz: return result;

graylog2-web-interface/src/views/components/aggregationbuilder/SortableSelect.jsx
kmerz: result.splice(to < 0 ? array.length + to : to, 0, result.splice(from, 1)[0]);

dennisoelkers: Seems like at least one test case would be nice! :D

PR #6147, opened 2019-07-16
Check for Plugin inputConfiguration before opening Inputs modal

kyleknighted:
## Description
Check for `inputConfiguration` in the PluginStore to provide a unique Input experience for specific pluggable integrations.

## Motivation and Context
Related PR: https://github.com/Graylog2/graylog-plugin-integrations/pull/142

Allows us to provide a unique experience for specific Inputs

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
+++ b/graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
@@ -9,7 +9,6 @@ import ActionsProvider from 'injection/ActionsProvider';
 
 import StoreProvider from 'injection/StoreProvider';
 import { PluginStore } from 'graylog-web-plugin/plugin';
-import history from 'util/History';
 
 import { InputForm } from 'components/inputs';
 
--- a/graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
+++ b/graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
@@ -8,6 +8,8 @@ import { ExternalLinkButton, Select } from 'components/common';
 import ActionsProvider from 'injection/ActionsProvider';
 
 import StoreProvider from 'injection/StoreProvider';
+import { PluginStore } from 'graylog-web-plugin/plugin';
+import history from 'util/History';
 
 import { InputForm } from 'components/inputs';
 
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
```diff
@@ -54,6 +56,14 @@ const CreateInputControl = createReactClass({
 
   _openModal(event) {
     event.preventDefault();
+    const { selectedInput } = this.state;
+
+    const customConfiguration = PluginStore.exports('inputConfiguration').filter(inputConfig => inputConfig.matches(selectedInput));
```
edmundoa: As far as I know by the usual unwritten convention, the plugin provides a `type` static field and the web interface does the matching. I think that's a bit easier for plugin authors (they don't have to write code to do the matching themselves), and that way we also avoid duplicating matches in different places. That is, unless there's some particular reason to do it in a different way here.

danotorrey:
Closes https://github.com/Graylog2/graylog-plugin-integrations/issues/130

graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
dennisoelkers: The idea was to allow a matcher, so we can match the beginning of the input's type string and use it for every input starting with `org.graylog.integrations.aws`. If we want to do that, we should also allow specifying a `type` key which the implementation falls back to if `matches` is not defined.

@kyleknighted: Do we really need to match a substring? Afaics we only have a single input class named `org.graylog.integrations.aws.inputs.AWSInput`, so we could also use a constant string comparison here?

graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
```diff
@@ -54,6 +56,14 @@ const CreateInputControl = createReactClass({
...
+
+    if (customConfiguration.length > 0) {
+      history.push(customConfiguration[0].path);
```
dennisoelkers: While this works for this specific use case, a static path will not work for updating an existing input, because we are missing the input details/the id of the input to retrieve them. So we need a function in any case, even if we just want to generate a url using it. And in this case, we could do it the same way we do it in [other places](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/views/components/actions/FieldActions.jsx#L55-L64) and allow specifying a React component that is rendered. That React component could - in the integrations use case - just push a new route to the history.

Issue #6082, opened 2019-07-04
Implement Lookup Table field provider UI


edmundoa:
Add form to make possible to use a Lookup Table as a field provider for Event fields.

PR #6151, opened 2019-07-17
Changes in Field Form step

### LINKED ISSUE #6133: Update field provider UI for template
We should update the help text for the template field value provider form to not talk about freemarker (we are using JMTE again) and maybe add some small help text to that form with and example. (e.g. `${source.foo}`)

image

edmundoa:
This PR makes some improvements in the Field Form step for Event Definitions:

- Make list of Fields more compact
- Display Field form in a different screen, helping us to reduce issues with field sorting, and duplicate field names
- Make keys more prominent, and add help popover to make it easier to understand what keys are
- Update Template Field Value Provider form, to display the right template used, and also to add the newly supported fields
- Implement Lookup Table Field Value form and summary

Fixes #6133 and #6082

Before:
Screenshot 2019-07-17 at 11 39 35

After:
Screenshot 2019-07-17 at 11 38 52


graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/LookupTableFieldValueProviderForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+                   checked={provider.key_context === 'event'}
+                   onChange={this.handleChange}>
+              Generated Event
+            </Radio>
+          </FormGroup>
+
+          <FormGroup controlId="lookup-provider-table">
+            <ControlLabel>Lookup Table Key Field</ControlLabel>
+            <Select name="lookup-provider-key"
+                    placeholder="Select Lookup Table"
```
bernd: This should probably be "Select Field" or something similar.

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/LookupTableFieldValueProviderForm.jsx
```diff
@@ -0,0 +1,130 @@
...
+                   checked={provider.key_context === 'source'}
+                   onChange={this.handleChange}>
+              Source log message
+            </Radio>
+            <Radio id="lookup-event-key-context"
+                   name="key_context"
+                   value="event"
+                   checked={provider.key_context === 'event'}
+                   onChange={this.handleChange}>
+              Generated Event
```
bernd: I think we cannot use fields from the generated event, unfortunately. Because the field spec is a map/object instead of a list of objects in the backend, there is no defined execution order of the field spec. That means if we take a field for the lookup table provider from the event, it might not be defined yet.

So I guess we can only use fields from the message at the moment. Otherwise we would have to change the backend structure so we have a defined field provider execution order.

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/TemplateFieldValueProviderForm.jsx
```diff
@@ -19,29 +15,49 @@ class TemplateFieldValueProviderForm extends React.Component {
...
     const { config } = this.props;
     const provider = config.providers.find(p => p.type === TemplateFieldValueProviderForm.type);
 
+    const helpText = (
+      <span>
+        Type a literal text to set to this Field or use{' '}
+        <ExternalLink href="https://cdn.rawgit.com/DJCordhose/jmte/master/doc/index.html">
+          JMTE syntax
+        </ExternalLink>
+        {' '}to add a dynamic Value.
```
bernd: What do you think about adding an example here as well? Otherwise uses don't really know that they have to use the `source.` prefix to access fields from the source message.

graylog2-web-interface/src/components/event-notifications/event-notification-types/EmailNotificationSummary.jsx
```diff
@@ -21,16 +21,32 @@ class EmailNotificationSummary extends React.Component {
...
+            <td>Email recipients</td>
+            <td>
+              {notification.config.email_recipients.join(', ') || 'No email addresses are configured to receive this notification.'}
+            </td>
+          </tr>
+          <tr>
+            <td>Email Body</td>
+            <td>
+              <pre className={`${styles.bodyPreview} pre-scrollable`}>
+                {notification.config.body_template || <em>Empty body</em>}
```
bernd: The email body looks a bit to wide. :smiley: 

image

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/LookupTableFieldValueProviderForm.jsx
edmundoa: All I did was reading the documentation we have for the provider and also the fields the API supports. I can remove that from the frontend, but we should do the same in the API if that is not going to work.

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/LookupTableFieldValueProviderForm.jsx
edmundoa: ✅

graylog2-web-interface/src/components/event-notifications/event-notification-types/EmailNotificationSummary.jsx
edmundoa: ✅

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/TemplateFieldValueProviderForm.jsx
edmundoa: I didn't add an example because I was going to add that into the preview space. Also, does the template always get the results in the `source` field? I think at some point that was different for aggregations and filters, wasn't it?

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/LookupTableFieldValueProviderForm.jsx
bernd: Yes, the idea was to support this but then I realized that it doesn't work because we use a map instead of a list of maps in the backends.

Yeah, I will prepare a PR to remove it from the API as well. It's a backwards incompatible change, but since we didn't have any UI for that yet to create a lookup-table field provider, no one should run into any issues. :smile:

bernd: The alternative is to redo the backend structure to make it a list of field providers. The we can define a processing order for field providers. But that would require lots of changes in the UI, I guess. The user then also needs to be able to reorder the field spec in the UI and has to actually think about the order to make sure referencing previously defined fields works. And then the backends needs to validate that as well...

So I guess it's easier to not allow using fields from the event. Easier for the user and for the implementation. :smile_cat:

graylog2-web-interface/src/components/event-definitions/event-definition-form/field-value-providers/TemplateFieldValueProviderForm.jsx
bernd: No, I think it was always `source` regardless of the event definition. I was using `source` because it references the source message/event that led to the current event that is created. Theoretically we could remove the `source.` prefix requirement. But that would be an issue should we add more information to the context at some point. So I think we should keep the prefix.

bernd: This is the PR to remove the `key_context` from the lookup table field value provider configuration.

https://github.com/Graylog2/graylog-plugin-events/pull/64

edmundoa: So, if `source` references to the source message/event that lead to the current event, what would be `source` in case of an aggregation? I think in that case there is no single event that lead to the event being generated, but a group of messages.

bernd: @edmundoa 

For the following aggregation:

```json
{
  "group_by" : ["username", "remote_ip"],
  "series" : [
    {
      "id" : "ec167af6-3fce-4b05-b197-5422c9e1e4ef",
      "function" : "avg",
      "field" : "bytes"
    },
    {
      "id" : "ec167af6-3fce-4b05-b197-5422c9e1e4ff",
      "function" : "count",
      "field" : null
    }
  ]
}
```

you will get the following source message:

```json
{
  "username": "bernd",
  "remote_ip": "127.0.0.1",
  "aggregation_key": "bernd|127.0.0.1",
  "aggregation_value_avg_bytes": 256,
  "aggregation_value_count": 1234
}
```

We include the group field values, the key value and for every series a `aggregation_value_<function>_<field>` value.

edmundoa: Ok, thank you for the details! I think this needs some screen space to be documented, and I think an inline example isn't enough in this case.

bernd: Agreed. I think at some point it would be nice to have an endpoint that returns a template or test result for a aggregation so we can pre-fill the field spec for the user.

bernd: Or show it as "available fields".

edmundoa: 👍

PR #6160, opened 2019-07-18
Fix "Routed into streams" element on show message page

bernd:
The problem is that the MessageResource returns a ResultMessage instead
of a ResultMessageSummary. Due to that, we serialize the regular Message
object instead of only the fields. This exposes internals of the Message
object and also breaks the "Routed into streams" element on the show
message page because the UI elements don't expect the Message structure.

This change is basically backwards incompatible change for the get
message API because the response gets changed.
We need to check if that's okay.


danotorrey:
LGTM too. Tests successfully. I agree with not backporting to `3.0`.

//cc @bernd This is the server-side change for the AWS CloudWatch integration.

danotorrey:
Oops, approved the wrong PR. Stand by for approval confirmation.

danotorrey:
I did not get to review this today. I'll review tomorrow.

danotorrey:
I also found and created issue #6174 along the way, which appears to be unrelated.

bernd:
@danotorrey Good catch, thanks! I will take a look at this again in a few days to fix the issues.

bernd:
Since we are working on replacing the old search pages in 3.2.0, I am closing this PR.

PR #6172, opened 2019-07-19
Fix closing of widget filter menu.

dennisoelkers:
## Motivation and Context

Before this change, closing the widget filter menu popover resulted in
it immediately showing again and logging an error to the console. This
was the result of not passing a function for `onChange` to the
`QueryInput` component that returns a Promise.

This change is fixing that by returning a promise from the `onChange`
handler.

Before:
WidgetFilterMenu-before

After:
WidgetFilterMenu-after

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/WidgetFilterMenu.jsx
+++ b/graylog2-web-interface/src/views/components/widgets/WidgetFilterMenu.jsx
@@ -2,9 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Button, OverlayTrigger, Popover } from 'react-bootstrap';
 
-import connect from 'stores/connect';
-
-import { SearchStore } from 'views/stores/SearchStore';
 import QueryInput from '../searchbar/AsyncQueryInput';
 
 import style from './WidgetFilterMenu.css';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/views/components/widgets/WidgetFilterMenu.jsx
```diff
@@ -2,9 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Button, OverlayTrigger, Popover } from 'react-bootstrap';
 
-import connect from 'stores/connect';
-
-import { SearchStore } from 'views/stores/SearchStore';
 import QueryInput from '../searchbar/AsyncQueryInput';
```
kyleknighted: Should this import as `AsyncQueryInput` ?

kyleknighted: And, conversely, should the `AsyncQueryInput.jsx` export the var named `AsyncQueryInput`?

dennisoelkers: This is a special case. There are two files, `QueryInput.jsx`, which is just the component, `AsyncQueryInput.jsx`, which is reexporting the same component, but in an async way using `loadable-components`, so dependencies like `react-ace` are splitted out and loaded asynchronously. So from my perspective the export is fine, importing it as `QueryInput` too, because it _is_ that component when rendering.

PR #6183, opened 2019-07-23
Updating react-select styles

kyleknighted:
## Description
- Updating the styles for `react-select` to better match the Input styles
- Adding `react-select` to `graylog-web-plugin/package.json`
- Allow `components` prop to be passed to common `Select` component

## Motivation and Context
[graylog-plugin-integrations](https://github.com/Graylog2/graylog-plugin-integrations) needs a way to create a custom component within Select and the styles should be consistent.

image
image

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.

### DIFF
--- a/graylog2-web-interface/packages/graylog-web-plugin/package.json
+++ b/graylog2-web-interface/packages/graylog-web-plugin/package.json
@@ -49,6 +49,7 @@
     "react-dom": "16.8.6",
     "react-router": "3.2.1",
     "react-router-bootstrap": "0.23.2",
+    "react-select": "2.4.3",
     "reflux": "0.2.13",
     "webpack": "4.27.0",
     "webpack-cleanup-plugin": "0.5.1",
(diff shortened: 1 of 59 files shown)


graylog2-web-interface/packages/graylog-web-plugin/package.json
```diff
@@ -49,6 +49,7 @@
     "react-dom": "16.8.6",
     "react-router": "3.2.1",
     "react-router-bootstrap": "0.23.2",
+    "react-select": "2.4.3",
```
edmundoa: Why is the reason for including `react-select` here? For what I could find, so far the package is only installed in `graylog2-web-interface` and plugins use `react-select` only through the `Select` component.

Adding `react-select` here would require it to be removed from `graylog2-web-interface/package.json`, otherwise we may run into issues when the package is updated in one place and not the other.

kyleknighted: Thanks for the tip, I will remove it from `graylog2-web-interface/package.json`.
It needed to be added here so that the graylog plugins are able to `import` from this package.

Otherwise each plugin will need to do their own installation of the dependency.

edmundoa: Sounds good to me, just be aware that we may need to update the `yarn.lock` files in the plugins as well.
Alternatively we can put it in the integrations' `package.json` for now and add it back here once more plugins depend on it.

PR #6193, opened 2019-07-25
Add backlog size config to Event UI

mpfz0r:


graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationSettingsForm.jsx
```diff
@@ -41,9 +43,17 @@ class NotificationSettingsForm extends React.Component {
     this.setState(stateUpdate);
   };
 
+  handleBacklogSizeChange = (e) => {
+    const { eventDefinition, onSettingsChange } = this.props;
+    const nextNotificationSettings = lodash.cloneDeep(eventDefinition.notification_settings);
+    nextNotificationSettings.backlog_size = e.target.value;
```
edmundoa: HTML `<input>`s always keep values as strings, and this is not converting the input into a number, which is what the API expects, so the request may fail or end up storing something wrong in the DB. Please use `FormsUtils.getValueFromInput()` to get a converted value from the input.

graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationSettingsForm.jsx
```diff
@@ -70,6 +81,15 @@ class NotificationSettingsForm extends React.Component {
               Events with keys have different grace periods for each key.
             </HelpBlock>
           </FormGroup>
+          <Input type="number"
+                 id="backlogsize"
+                 name="backlog_size"
+                 label="Message Backlog"
+                 required
+                 onChange={this.handleBacklogSizeChange}
+                 help="The maximum number of backlog messages to be included in notifications. Use 0 to disable."
```
edmundoa: I think we should disable the backlog with a checkbox, not using 0. We could use something like this, if we want to put everything in a single input: https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/components/forms/#forms-input-addons

graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationSettingsForm.jsx
```diff
@@ -19,6 +20,7 @@ class NotificationSettingsForm extends React.Component {
 
   state = {
     lastEnabledGracePeriod: undefined,
+    lastBacklogSize: undefined,
```
edmundoa: With the current implementation we don't need to keep the backlog in the state, since it always have the same value as the props. This may be needed, though, if we want to use a `checkbox` to disable the backlog, as I suggested. That way we can use `0` internally but still display the last value the field had.

Issue #6155, opened 2019-07-17
Events UI improvements


edmundoa:
- [x] Display helpful message and link to create Event Definition form, when a system has no Events and no Event Definitions
- [x] Improve layout or content to avoid "create" buttons to take a row on their own
- [x] Sort all the things
- [x] Allow clearing numeric inputs
- [x] Changing time unit when duration is 0 does not work
- [x] Make clearer how many events an Event Definition will generate
- [x] Fix or remove selecting Events as source in Filter & Aggregation step
- [x] Moving through steps in the Event Definition form resets some state, resulting in confusion if some configuration was persisted or not
- [x] Resolve Stream IDs in summary

Edit with some more improvements:

- [x] Improve some texts:
  - [x] Text in Events page header is too complex
  - [x] Example in Aggregation group by is not helpful
  - [x] Improve Field information: Why would you like to create a Field?
  - [x] Grace period description is not clear
- [x] Improve information on Condition types
- [ ] Use longer name in Aggregation functions
- [x] Use buttons instead of link to add Fields and Notifications, as they make clearer where one should click
- [ ] Wizard makes editing more complicated
- [ ] Previous/Next buttons in Wizard have to much space between them
- [x] Make clearer the difference between Alert and Event
- [x] When creating a new Event Definition, display title in all steps of the form
- [x] Make selection of Condition and its form closer together
- [x] When using inline checkbox to disable input, change color to make clearer that the input is disabled
- [x] Display confirmation dialog before leaving forms
- [x] Use consistent URL for `edit` pages

edmundoa:
Closing this meta-issue. Most issues have already been fixed, and the remaining will not be done for 3.1.0. Instead of keeping this open, I created three separate issues in #6289, #6290, and #6291.

PR #6198, opened 2019-07-26
Several UI improvements in Events

edmundoa:
This PR fixes some of the issues/improvements described in #6155, namely:

- Add link to create Event Definition from Event list page when user visits the page and no Event Definitions have been created
- Display title of new Event Definition in page header
- Move select of Event Definition type to the Condition step
- Add information about available Event Definition types
- Add information about number of Events created from Event Definition
- Add or improve information about Filter, Aggregation, and group by
- Only update Filter preview when query, streams, or search time range change
- Use buttons instead of links to add Notifications / Fields, making clearer where the action is from a quick look


graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterPreview.jsx
```diff
@@ -54,15 +55,30 @@ class FilterPreview extends React.Component {
...
+      <React.Fragment>
+        <HelpPanel collapsible
+                   title="How many Events will Filter & Aggregation create?">
+          <p>
+            The Filter & Aggregation Condition will generate different number of Events, depending on how it is
+            configured:
+          </p>
+          <ul>
+            <li><b>Filter:</b>&emsp;One Event per message matching the filter</li>
+            <li><b>Aggregation without groups:</b>&emsp;One Event every time the Event Definition is executed</li>
```
bernd: I think "one event every time the event definition is executed" is a bit misleading. It's more like: "one event every time the condition matches the result of the message aggregation" (same for the "per group" case)

edmundoa: I tried to make it more accurate now 👍

PR #6210, opened 2019-07-29
Add refresh button to Events page

### LINKED ISSUE #6154: Add control to refresh Events
Users should be able to refresh the list of Events at least manually, like we did with Alerts before.

edmundoa:
This PR adds the functionality to manually refresh the Alert/Event list.
It additionally refactor the `Events` component to make it more modular and improves the overall styling of the Event search controls.

Screenshot 2019-07-29 at 16 58 44

Fixes #6154

### DIFF
--- a/graylog2-web-interface/src/components/common/PaginatedList.jsx
+++ b/graylog2-web-interface/src/components/common/PaginatedList.jsx
@@ -46,8 +46,13 @@ class PaginatedList extends React.Component {
   };
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.activePage !== nextProps.activePage) {
+    const { pageSize, activePage } = this.props;
+
+    if (activePage !== nextProps.activePage) {
       this.setState({ currentPage: nextProps.activePage });
+    }
+    if (pageSize !== nextProps.pageSize) {
+      this.setState({ pageSize: nextProps.pageSize });
     }
   }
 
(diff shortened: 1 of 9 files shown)


mpfz0r:
eventbug

I cannot seem to show event details anymore

edmundoa:
Fixed in d02503a

graylog2-web-interface/src/components/common/PaginatedList.jsx
```diff
@@ -46,9 +46,14 @@ class PaginatedList extends React.Component {
...
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.activePage !== nextProps.activePage) {
+    const { pageSize, activePage } = this.props;
+
+    if (activePage !== nextProps.activePage) {
       this.setState({ currentPage: nextProps.activePage });
     }
+    if (pageSize !== nextProps.pageSize) {
+      this.setState({ pageSize: nextProps.pageSize });
```
bernd: I was wondering if calling `setState()` twice in this method will trigger a double re-render of components?
If that's the case, can we only do one `setState()`?

edmundoa: No, calling `setState()` only enqueues the changes but the re-render does not happen immediately.

From https://reactjs.org/docs/react-component.html#setstate:
> Think of `setState()` as a request rather than an immediate command to update the component. For better perceived performance, React may delay it, and then update several components in a single pass. React does not guarantee that the state changes are applied immediately.

bernd: Okay, got it. Thanks for the explanation. :+1:

PR #6213, opened 2019-07-29
Add "source_streams" to event and index mapping

### LINKED ISSUE #6211: Add "source_streams" to events
## Description

We need a `source_streams` field in events. This can be used to use permission filters on events based on stream permissions.

## Tasks

- [ ] Add `source_streams` field to event interface and implementations
- [ ] Add `source_streams` field to event index schema
- [ ] Adjust all event processors to fill the `source_streams` field

bernd:
This PR adds the `source_streams` field to the event index schema and the event structure. The aggregation event processor will also set this field depending on the aggregation processor mode. (filter or aggregation)

Closes #6211

## Testing

To be able to test this, you have to delete the "All events" and "All system events" stream and the "Graylog Events" and "Graylog System Events" index sets. After that the server needs to be restarted to make sure the index sets and streams will be created again.


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -168,6 +179,12 @@ private void filterSearch(EventFactory eventFactory, AggregationEventProcessorPa
                 final Message msg = resultMessage.getMessage();
                 final Event event = eventFactory.createEvent(eventDefinition, msg.getTimestamp(), eventDefinition.title());
                 event.setOriginContext(EventOriginContext.elasticsearchMessage(resultMessage.getIndex(), msg.getId()));
+
+                // We don't want source streams in the event which are unrelated to the event definition
+                msg.getStreamIds().stream()
+                    .filter(stream -> config.streams().contains(stream))
```
mpfz0r: Hmm, but shouldn't that use `streams` in case it was overloaded with parameters?

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -229,6 +251,7 @@ private boolean satisfiesConditions(AggregationKeyResult keyResult) {
             // TODO: Do we have to set any other event fields here?
             event.setTimerangeStart(parameters.timerange().getFrom());
             event.setTimerangeEnd(parameters.timerange().getTo());
+            sourceStreams.forEach(event::addSourceStream);
```
mpfz0r: wouldn't sth like `event.setSourceStreams(sourceStreams)` be more efficient?

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
bernd: Good catch, yes! :+1:

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
bernd: Yes, I think it would. The idea of not providing `set*` style methods for streams and source-streams was to avoid overriding all streams by accident.

PR #6214, opened 2019-07-29
Clean up completed OnceJobSchedule triggers after one day

mpfz0r:


graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
```diff
@@ -222,6 +223,21 @@ public boolean delete(String triggerId) {
         return db.remove(DBQuery.is(FIELD_ID, triggerId)).getN() > 0;
     }
 
+    /**
+     *
+     * @param olderThan triggers that are older than the current time - olderThan in ms will be removed
+     * @return the number of deleted triggers
+     */
+    public int deleteCompletedOnceSchedulesOlderThan(long olderThan) {
```
bernd: What do you think about using the following method signature? I think that would make it easier to use the API.

```suggestion
    public int deleteCompletedOnceSchedulesOlderThan(long olderThan, TimeUnit unit) {
```

We can use `unit.toMillis(olderThan)` to get the milliseconds.

PR #6223, opened 2019-07-31
Events system configuration

mariussturm:
## Description

Adds system configurations for the Events module. The three options are:

Search Timeout: ES queries should be interrupted at some point
Notification retry period: In case of an error notifications are retried
Notification default backlog size: The default value for backlogs in notifications is used as a fallback in case of a configuration error as well as the default value for new event definitions
 
Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/1063

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfiguration.java
@@ -21,8 +21,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
 
-import org.joda.time.Period;
-
 @JsonAutoDetect
 @JsonDeserialize(builder = AutoValue_EventsConfiguration.Builder.class)
 @AutoValue
@@ -31,23 +29,23 @@ public abstract class EventsConfiguration {
     private static final String FIELD_NOTIFICATIONS_RETRY_PERIOD = "events_notification_retry_period";
     private static final String FIELD_NOTIFICATIONS_DEFAULT_BACKLOG = "events_notification_default_backlog";
 
-    private static final Period DEFAULT_SEARCH_TIMEOUT = Period.seconds(60);
-    private static final Period DEFAULT_NOTIFICATIONS_RETRY_PERIOD = Period.minutes(5);
+    private static final long DEFAULT_SEARCH_TIMEOUT_MS = 60000;
+    private static final long DEFAULT_NOTIFICATIONS_RETRY_MS = 300000;
     private static final long DEFAULT_NOTIFICATIONS_BACKLOG = 50;
 
     @JsonProperty(FIELD_SEARCH_TIMEOUT)
-    public abstract Period eventsSearchTimeout();
+    public abstract long eventsSearchTimeout();
 
     @JsonProperty(FIELD_NOTIFICATIONS_RETRY_PERIOD)
-    public abstract Period eventNotificationsRetry();
+    public abstract long eventNotificationsRetry();
 
     @JsonProperty(FIELD_NOTIFICATIONS_DEFAULT_BACKLOG)
     public abstract long eventNotificationsBacklog();
 
     public static Builder builder() {
         return new AutoValue_EventsConfiguration.Builder()
-                .eventsSearchTimeout(DEFAULT_SEARCH_TIMEOUT)
-                .eventNotificationsRetry(DEFAULT_NOTIFICATIONS_RETRY_PERIOD)
+                .eventsSearchTimeout(DEFAULT_SEARCH_TIMEOUT_MS)
+                .eventNotificationsRetry(DEFAULT_NOTIFICATIONS_RETRY_MS)
                 .eventNotificationsBacklog(DEFAULT_NOTIFICATIONS_BACKLOG);
     }
 
@@ -56,10 +54,10 @@ public abstract class EventsConfiguration {
     @AutoValue.Builder
     public static abstract class Builder {
         @JsonProperty(FIELD_SEARCH_TIMEOUT)
-        public abstract Builder eventsSearchTimeout(Period searchTimeout);
+        public abstract Builder eventsSearchTimeout(long searchTimeout);
 
         @JsonProperty(FIELD_NOTIFICATIONS_RETRY_PERIOD)
-        public abstract Builder eventNotificationsRetry(Period notificationRetry);
+        public abstract Builder eventNotificationsRetry(long notificationRetry);
 
         @JsonProperty(FIELD_NOTIFICATIONS_DEFAULT_BACKLOG)
         public abstract Builder eventNotificationsBacklog(long defaultBacklog);
--- a/graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfiguration.java
@@ -1,0 +1,53 @@
+package org.graylog.events.configuration;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+
+import org.joda.time.Period;
+
+@JsonAutoDetect
+@JsonDeserialize(builder = AutoValue_EventsConfiguration.Builder.class)
+@AutoValue
+public abstract class EventsConfiguration {
+    private static final String FIELD_SEARCH_TIMEOUT = "events_search_timeout";
+    private static final String FIELD_NOTIFICATIONS_RETRY_PERIOD = "events_notification_retry_period";
+    private static final String FIELD_NOTIFICATIONS_DEFAULT_BACKLOG = "events_notification_default_backlog";
+
+    private static final Period DEFAULT_SEARCH_TIMEOUT = Period.seconds(60);
+    private static final Period DEFAULT_NOTIFICATIONS_RETRY_PERIOD = Period.minutes(5);
+    private static final long DEFAULT_NOTIFICATIONS_BACKLOG = 50;
+
+    @JsonProperty(FIELD_SEARCH_TIMEOUT)
+    public abstract Period eventsSearchTimeout();
+
+    @JsonProperty(FIELD_NOTIFICATIONS_RETRY_PERIOD)
+    public abstract Period eventNotificationsRetry();
+
+    @JsonProperty(FIELD_NOTIFICATIONS_DEFAULT_BACKLOG)
+    public abstract long eventNotificationsBacklog();
+
+    public static Builder builder() {
+        return new AutoValue_EventsConfiguration.Builder()
+                .eventsSearchTimeout(DEFAULT_SEARCH_TIMEOUT)
+                .eventNotificationsRetry(DEFAULT_NOTIFICATIONS_RETRY_PERIOD)
+                .eventNotificationsBacklog(DEFAULT_NOTIFICATIONS_BACKLOG);
+    }
+
+    public abstract Builder toBuilder();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+        @JsonProperty(FIELD_SEARCH_TIMEOUT)
+        public abstract Builder eventsSearchTimeout(Period searchTimeout);
+
+        @JsonProperty(FIELD_NOTIFICATIONS_RETRY_PERIOD)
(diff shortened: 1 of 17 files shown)


graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfigurationProvider.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public EventsConfiguration get() {
+        return getEventsConfig().orElse(DEFAULT_CONFIG);
+    }
+
+    @NotNull
+    public Optional<EventsConfiguration> getEventsConfig() {
+        final EventsConfiguration dto = clusterConfigService.get(EventsConfiguration.class);
+
+        return dto != null ? Optional.of(dto) : Optional.empty();
```
bernd: You can use `Optional.ofNullable` here.

graylog2-server/src/main/java/org/graylog2/migrations/V20190730000000_CreateDefaultEventsConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+import org.graylog.events.configuration.EventsConfigurationProvider;
+import org.graylog2.plugin.cluster.ClusterConfigService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import java.time.ZonedDateTime;
+import java.util.Optional;
+
+public class V20190730000000_CreateDefaultEventsConfiguration extends Migration {
```
bernd: I think we don't need a migration for this. The provider can just return a default instance of the config when there is no entry in the database. The first time a user updates the configuration it will be written to the database.

graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String FIELD_SEARCH_TIMEOUT = "events_search_timeout";
+    private static final String FIELD_NOTIFICATIONS_RETRY_PERIOD = "events_notification_retry_period";
+    private static final String FIELD_NOTIFICATIONS_DEFAULT_BACKLOG = "events_notification_default_backlog";
+
+    private static final Period DEFAULT_SEARCH_TIMEOUT = Period.seconds(60);
+    private static final Period DEFAULT_NOTIFICATIONS_RETRY_PERIOD = Period.minutes(5);
+    private static final long DEFAULT_NOTIFICATIONS_BACKLOG = 50;
+
+    @JsonProperty(FIELD_SEARCH_TIMEOUT)
+    public abstract Period eventsSearchTimeout();
```
bernd: What do you think about using a long representing milliseconds here? The Period objects are sometimes weird to work with because they take DST into account.
Also it's hard for users to use these ISO duration forms. With milliseconds we ca just use a UI form that offers milliseconds, seconds, ... in a drop-down.

graylog2-server/src/main/java/org/graylog/events/notifications/EventNotificationExecutionJob.java
```diff
@@ -130,12 +132,14 @@ public JobTriggerUpdate execute(JobExecutionContext ctx) throws JobExecutionExce
...
-                            retryTimeInMinutes,
+                            eventsConfiguration.eventNotificationsRetry().toStandardDuration().getStandardMinutes(),
                             notification.id(),
                             notification.title(),
                             notification.config().type()),
                     trigger,
-                    ctx.jobTriggerUpdates().retryIn(retryTimeInMinutes, TimeUnit.MINUTES),
+                    ctx.jobTriggerUpdates().retryIn(
+                        eventsConfiguration.eventNotificationsRetry().toStandardDuration().getStandardMinutes(),
+                        TimeUnit.MINUTES),
```
bernd: Using MINUTES here doesn't allow users to use a retry under 1 minute. I think we should use MILLISECONDS here to make it more flexible.

graylog2-server/src/main/java/org/graylog/events/notifications/EventNotificationExecutionJob.java
```diff
@@ -58,21 +60,21 @@
...
-                                         Map<String, EventNotification.Factory> eventNotificationFactories) {
+                                         Map<String, EventNotification.Factory> eventNotificationFactories,
+                                         EventsConfigurationProvider configurationProvider) {
         this.jobConfig = (Config) jobDefinition.config();
         this.notificationService = dbNotificationService;
         this.eventDefinitionService = eventDefinitionService;
         this.notificationGracePeriodService = notificationGracePeriodService;
         this.eventNotificationFactories = eventNotificationFactories;
-        this.retryTimeInMinutes = 1L; // TODO: Make retry time configurable in the plugin settings
+        this.eventsConfiguration = configurationProvider.get();
```
bernd: Using `.get()` on the provider in the constructor can be problematic. If it throws an exception, users will get an ugly injection exception. If we call it in the execute method, we can catch the exception and log a nicer error.

graylog2-server/src/main/java/org/graylog/events/notifications/EventNotificationService.java
```diff
@@ -28,10 +30,13 @@
...
 
     private final EventBacklogService eventBacklogService;
+    private final EventsConfiguration eventsConfiguration;
 
     @Inject
-    public EventNotificationService(EventBacklogService eventBacklogService) {
+    public EventNotificationService(EventBacklogService eventBacklogService,
+                                    EventsConfigurationProvider configurationProvider) {
         this.eventBacklogService = eventBacklogService;
+        this.eventsConfiguration = configurationProvider.get();
```
bernd: Same as above. Not calling .get() in the constructor.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
```diff
@@ -67,19 +69,21 @@
...
                                   SearchJobService searchJobService,
-                                  QueryEngine queryEngine) {
+                                  QueryEngine queryEngine,
+                                  EventsConfigurationProvider configProvider) {
         this.config = config;
         this.parameters = parameters;
         this.searchOwner = searchOwner;
         this.searchJobService = searchJobService;
         this.queryEngine = queryEngine;
+        this.eventsConfiguration = configProvider.get();
```
bernd: Same as above. Not calling .get() in the constructor.

graylog2-server/src/main/java/org/graylog2/migrations/V20190730000000_CreateDefaultEventsConfiguration.java
mariussturm: The cluster config resource is not providing any data if there is nothing in the database so the UI component will not load. I don't see at which point we could force the creation through an access?

bernd: Ah, I see. Okay, then we need the migration. :confused: Good catch!

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+  getInitialState() {
+    const { config } = this.props;
+    return {
+      config: ObjectUtils.clone(config),
+    };
+  },
+
+  componentWillReceiveProps(newProps) {
+    this.setState({ config: ObjectUtils.clone(newProps.config) });
```
edmundoa: In this case you are not modifying the `config` prop, so you don't need to clone it.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+        events_notification_retry_period: 300000,
+        events_notification_default_backlog: 50,
+      },
+    };
+  },
+
+  getInitialState() {
+    const { config } = this.props;
+    return {
+      config: ObjectUtils.clone(config),
```
edmundoa: In this case you are not modifying the `config` prop, so you don't need to clone it.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+        update[key] = FormUtils.getValueFromInput(value.target);
+      } else {
+        update[key] = value;
+      }
+      this.setState({ config: update });
+    };
+  },
+
+  _titleCase(str) {
+    return lodash.startCase(lodash.toLower(str));
```
edmundoa: Please use [`lodash.capitalize()`](https://lodash.com/docs/4.17.11#capitalize) instead.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+  _searchTimeoutValidator(milliseconds) {
+    return milliseconds >= 1000;
+  },
+
+  _notificationsRetryValidator(milliseconds) {
+    return milliseconds >= 0;
+  },
+
+  _onBacklogUpdate(key) {
```
edmundoa: This method seems to complex for what it does. Since it's only taking care of the backlog, I would rewrite it as:

```
  _onBacklogUpdate(event) {
    const value = FormUtils.getValueFromInput(event.target);
    this._propagateChanges('events_notification_default_backlog', value);
  },
```

In the component using it, you can pass a reference to the function instead of calling it directly.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+        <BootstrapModalForm ref={(modal) => { this.modal = modal; }}
+                            title="Update Events System Configuration"
+                            onSubmitForm={this._saveConfig}
+                            onModalClose={this._resetConfig}
+                            submitButtonText="Save">
+          <fieldset>
+            <FormGroup controlId="search-timeout-field">
+              <TimeUnitInput label="Search Timeout"
+                             update={this._onSearchTimeoutUpdate}
+                             defaultEnabled={eventsSearchTimeout.duration !== 0}
```
edmundoa: I think we don't need that. The component is required and cannot be disabled.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+                             units={TIME_UNITS}
+                             required />
+              <HelpBlock>
+                Amount of time after which an Elasticsearch query is interrupted.
+              </HelpBlock>
+            </FormGroup>
+            <FormGroup controlId="notifications-retry-field">
+              <TimeUnitInput label="Notifications retry period"
+                             update={this._onRetryPeriodUpdate}
+                             defaultEnabled={eventsNotificationRetryPeriod.duration !== 0}
```
edmundoa: I think we don't need that. The component is required and cannot be disabled.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+  _onRetryPeriodUpdate(nextValue, nextUnit, enabled) {
+    const durationInMs = enabled ? moment.duration(nextValue, nextUnit).asMilliseconds() : 0;
+    if (this._notificationsRetryValidator(durationInMs)) {
+      this._propagateChanges('events_notification_retry_period', durationInMs);
+    }
+  },
+
+  _searchTimeoutValidator(milliseconds) {
+    return milliseconds >= 1000;
```
edmundoa: I would document this in the `HelpBlock` of the input.

graylog2-web-interface/src/components/configurations/EventsConfig.jsx
```diff
@@ -0,0 +1,183 @@
...
+  },
+
+  _notificationsRetryValidator(milliseconds) {
+    return milliseconds >= 0;
```
edmundoa: I would document this in the `HelpBlock` of the input.

graylog2-server/src/main/java/org/graylog/events/configuration/EventsConfigurationProvider.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public EventsConfiguration get() {
+        return getEventsConfig().orElse(DEFAULT_CONFIG);
+    }
+
+    @NotNull
+    public Optional<EventsConfiguration> getEventsConfig() {
+        final EventsConfiguration dto = clusterConfigService.get(EventsConfiguration.class);
```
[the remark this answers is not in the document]
mariussturm: I am fine with this but just as a side note: This removes the defaults from to above comment.

PR #6224, opened 2019-07-31
Improve job scheduler responsiveness

bernd:
The loop used to sleep for a second when there were no free worker pool
slots or when there were no runnable triggers. This was done to avoid
busy spinning.

This was an issue when jobs need to catch up a big backlog and their
execution time is very short.

The job execution engine is now publishing an event when a job execution
has been completed. The scheduler service is subscribing to these events
and interrupts the sleep when receiving such an event to immediately
check for another job to execute.

We are now also waiting for the server to reach RUNNING status before we
start the scheduling loop.

Closes Graylog2/graylog-plugin-enterprise#1107
Closes Graylog2/graylog-plugin-enterprise#1067

This PR also makes the scheduler loop sleep duration configurable via the server config file.


graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerService.java
```diff
@@ -61,8 +90,13 @@ protected void run() throws Exception {
                 LOG.debug("Starting scheduler loop iteration");
                 try {
                     if (!jobExecutionEngine.execute() && isRunning()) {
-                        // We wait because there are either no free worker threads or no runnable triggers
-                        clock.sleepUninterruptibly(1, TimeUnit.SECONDS); // TODO: Duration should be configurable
+                        // When the execution engine returned false, there are either no free worker threads or no
+                        // runnable triggers. To avoid busy spinning we sleep for the configured duration or until
+                        // we receive a job completion event via the scheduler event bus.
+                        if (sleeper.sleep(loopSleepDuration.getQuantity(), loopSleepDuration.getUnit())) {
```
kroepke: This can throw `InterruptedException` which will cause an error to be logged below.
Since interruptions can also happen spuriously, those aren't actually an error.
Can we add another catch arm here and log on trace or debug for those?
All other Exceptions are fine (even if none should be thrown here) on LOG.error, afaics.

graylog2-server/src/main/java/org/graylog/scheduler/eventbus/JobSchedulerEventBus.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.common.eventbus.EventBus;
+
+import static com.codahale.metrics.MetricRegistry.name;
+
+/**
+ * Job scheduler specific event bus instance. This is a <em>synchronous</em> event bus.
+ * Subscribers must ensure that the callback method is fast or put expensive work into an executor or queue.
+ */
+public class JobSchedulerEventBus extends EventBus {
+    private static final String METRICS_PREFIX = "job-scheduler-eventbus";
```
mpfz0r: This won't be shown on the metrics page, because the default filter needs everything to start with `org.`

Why not use:

```
@@ -28,16 +28,14 @@ import static com.codahale.metrics.MetricRegistry.name;
  * Subscribers must ensure that the callback method is fast or put expensive work into an executor or queue.
  */
 public class JobSchedulerEventBus extends EventBus {
-    private static final String METRICS_PREFIX = "job-scheduler-eventbus";
-
     private final Counter registrationsCount;
     private final Timer postTimer;
 
     @SuppressWarnings("WeakerAccess")
     public JobSchedulerEventBus(String name, MetricRegistry metricRegistry) {
         super(name);
-        this.registrationsCount = metricRegistry.counter(name(METRICS_PREFIX, name, "registrations"));
-        this.postTimer = metricRegistry.timer(name(METRICS_PREFIX, name, "posts"));
+        this.registrationsCount = metricRegistry.counter(name(this.getClass(), name, "registrations"));
+        this.postTimer = metricRegistry.timer(name(this.getClass(), name, "posts"));
     }
 
     @Override
```

graylog2-server/src/main/java/org/graylog/scheduler/eventbus/JobSchedulerEventBus.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void unregister(Object object) {
+        registrationsCount.dec();
+        super.unregister(object);
+    }
+
+    @Override
+    public void post(Object event) {
+        try (final Timer.Context ignored = postTimer.time()) {
```
mpfz0r: not sure I understand. What's that supposed to time? :)

graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerService.java
```diff
@@ -75,9 +109,66 @@ protected void run() throws Exception {
...
+            // Always release so we can start with a clean slate when sleep is called again
+            semaphore.release();
+            return !acquired;
+        }
+
+        /**
+         * Interrupt a {@link #sleep(long, TimeUnit)} call so it unblocks.
+         */
+        public void interrupt() {
+            semaphore.release();
```
kroepke: This looks like it could "overrelease" permits because the `tryAcquire` and `release` calls are unbalanced: there are more jobs that complete than sleeps.

graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerService.java
kroepke: This can throw `InterruptedException` which will cause an error to be logged below.
Since interruptions can also happen spuriously, those aren't actually an error.
Can we add another catch arm here and log on trace or debug for those?
All other Exceptions are fine (even if none should be thrown here) on LOG.error, afaics.

graylog2-server/src/main/java/org/graylog/scheduler/eventbus/JobSchedulerEventBus.java
bernd: Good point! :+1: Will fix it.

graylog2-server/src/main/java/org/graylog/scheduler/eventbus/JobSchedulerEventBus.java
bernd: Posting the event to the event bus is synchronous. That way we can see if any subscriber methods are slow.

graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerService.java
bernd: Done. :+1:

graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerService.java
bernd: Good catch, thank you! :+1: We fixed it by draining all permits before the `tryAcquire(timeout)`.

kroepke:
I'm fine with my stuff

PR #6225, opened 2019-07-31
Cleanup stale job trigger on server startup

bernd:
This will only cleanup trigger that are owned by the calling server
node. It will fix stale job trigger after an unclean JVM or server
shutdown.

Closes Graylog2/graylog-plugin-enterprise#1139

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
+++ b/graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
@@ -46,6 +46,7 @@ public class JobExecutionEngine {
     private final JobWorkerPool workerPool;
 
     private final AtomicBoolean isRunning = new AtomicBoolean(true);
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
```diff
@@ -69,13 +70,29 @@ public void shutdown() {
         isRunning.set(false);
     }
 
+    private void cleanup() {
+        // The cleanup should only run once before starting job processing to avoid damaging any state.
+        if (shouldCleanup.getAndSet(false)) {
+            final int releasedTrigger = jobTriggerService.forceReleaseOwnedTriggers();
+            if (releasedTrigger > 0) {
+                LOG.warn("Force-released {} stale job trigger after an unclean job scheduler shutdown", releasedTrigger);
```
mpfz0r: trigger_s_  :bowtie:

PR #6230, opened 2019-08-01
Content Pack Integration for EventDefintions

kmerz:
##  Description
Event Definition do not have yet a content pack integration.
This PR will add ContentPackAble Interface to export a EventDefintion and its config
to a content pack entity. And add a NativeEntityConverter to export a EventDefintionEntity
to its NativeEntity (EventDefintion).
This PR, also, adds a EventDefintionFacade

## Motivation and Context
EventDefinition need also to be able to be exported to a content pack.

## How Has This Been Tested?
- Exported and imported a EventDefintion via Content Pack

## Known Issue
- Using parameters on long values is still a open issue.

The build failes becaus of:
Refs: https://github.com/Graylog2/graylog-plugin-enterprise/pull/1169

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/events/contentpack/facat/AggregationConditionsEntity.java
```diff
@@ -0,0 +1,50 @@
...
+import org.mongojack.Id;
+import org.mongojack.ObjectId;
+
+import javax.annotation.Nullable;
+
+@AutoValue
+@JsonDeserialize(builder = AggregationConditionsEntity.Builder.class)
+public abstract class AggregationConditionsEntity {
+
+    private static final String FIELD_EXPRESSION = "expression";
```
bernd: This seems to be unused. Are there no fields in this value object?

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
```diff
@@ -68,5 +69,10 @@ public String type() {
         public ValidationResult validate() {
             throw new UnsupportedOperationException();
         }
+
+        @Override
+        public Object toContentPackEntity() {
+            return null;
```
bernd: Shouldn't this return something?

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -136,4 +138,18 @@ public static Builder create() {
     public ValidationResult validate() {
         return new ValidationResult();
     }
+
+    @Override
+    public Object toContentPackEntity() {
+        return AggregationEventProcessorConfigEntity.builder()
+            .type(type())
+            .query(ValueReference.of(query()))
+            .streams(streams()) // TODO?
```
bernd: Is this `TODO` still relevant?

graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackAble.java
```diff
@@ -0,0 +1,21 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.contentpacks;
+
+public interface ContentPackAble<T> {
+    Object toContentPackEntity();
```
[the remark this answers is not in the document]
kmerz: The only case where it can be null was with the FallbackConfig, and now I throw there an `UnsupportedOperationException` which is what the other functions do as well.

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
kmerz: I throw there an UnsupportedOperationException which is what the other functions do as well. If you would rather have a `Optional.empty()` I can do that as well. But at some point I have to raise an exception, because I can't go to defined behavior with a `FallbackConfig`

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
kmerz: Yes, but I wrote it down with pen and paper :)

PR #6233, opened 2019-08-01
Fix error for missing legacy notification plugins in Event Definition form

edmundoa:
When selecting a notification from a removed legacy alarm callback plugin in the Event Definition form, going to the summary step results in an error.

This commit fixes the issue by checking if the configuration is defined before accessing it and displaying a message like in #6199.


mariussturm:
Looks good so far!
We could expand the notification details by default if there are problems on the summary page. So that's more obvious a non existing plugin was picked.

PR #6243, opened 2019-08-05
Add metrics support for Events, Notifications and scheduler

### LINKED ISSUE #6196: Add metrics to events/alerting system
The new events/alerting system doesn't implement any metrics, yet.

To be able to debug issues or monitor behavior we have to add metrics to several subsystems.

## Tasks

- Notification execution
  - [x] Meter for success, error and retries
- Event definition execution
  - [x] Counter for executions, success, error and retries (meter is probably not that useful with long intervals between executions)
  - [x] Number of created events per event definition
- Job scheduler
  - [x] Timer for job execution
  - [ ] Meter for exceptions and trigger locks and releases

## Open Questions

- [ ] tbd

mpfz0r:
Closes #6196


mpfz0r:
image

..to get an overview of the naming

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorExecutionMetrics.java
```diff
@@ -0,0 +1,103 @@
...
+    void recordException(EventProcessor eventProcessor, String definitionId) {
+        final String name = getNameforField(eventProcessor, definitionId, Fields.EXECUTION_EXCEPTION);
+        MetricUtils.getOrRegister(metricRegistry, name, new Counter()).inc();
+    }
+    void recordCreatedEvents(EventProcessor eventProcessor, String definitionId, int count) {
+        final String name = getNameforField(eventProcessor, definitionId, Fields.EVENTS_CREATED);
+        MetricUtils.getOrRegister(metricRegistry, name, new Meter()).mark(count);
+    }
+
+    private static String getNameforField(EventProcessor eventProcessor, String definitionId, Fields field) {
```
bernd: How about renaming this for camel-case consistency?

```suggestion
    private static String getNameForField(EventProcessor eventProcessor, String definitionId, Fields field) {
```

graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
```diff
@@ -47,6 +52,10 @@
     private final JobTriggerUpdates.Factory jobTriggerUpdatesFactory;
     private final Map<String, Job.Factory> jobFactory;
     private final JobWorkerPool workerPool;
+    private final MetricRegistry metricRegistry;
+    private Counter execution_successful;
+    private Counter execution_failed;
+    private Timer execution_time;
```
bernd: Can we use camel-case for these as well, please? :smiley:

graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
```diff
@@ -181,4 +196,10 @@ private void executeJob(JobTriggerDto trigger, JobDefinitionDto jobDefinition, J
             }
         }
     }
+
+    private void setupMetrics() {
```
bernd: If we inline this method and create the metric fields inside the constructor, we could get rid of the `metricRegistry` field. Calling methods inside a constructor can also be tricky because the object might not be completely initialized. If the `setupMetrics()` call would be moved above the `this.metricRegistry = metricRegistry`, the method call would fail.

As an alternative you could also pass in the `metricRegistry` into the `setupMetrics()` method like you did in `JobWorkerPool#registerMetrics`. I usually make these kind of methods that are used in constructors `static` so I cannot use `this.` by accident and thus avoid using a potentially uninitialized field.

graylog2-server/src/test/java/org/graylog/scheduler/job/EventProcessorExecutionJobTest.java
```diff
@@ -61,6 +62,7 @@
 
     private JobScheduleStrategies jobScheduleStrategies;
     private JobSchedulerTestClock clock;
+    private EventProcessorExecutionMetrics metrics;
```
bernd: This field is not in use and can be removed.

graylog2-server/src/main/java/org/graylog2/shared/metrics/MetricUtils.java
```diff
@@ -187,6 +187,20 @@ public static MetricFilter filterSingleMetric(String name) {
         }
     }
 
+    public static <T extends Metric> T getOrRegister(MetricRegistry metricRegistry, String name, T newMetric) {
```
bernd: Can you please extend the `MetricUtilsTest` class for this new method? Thank you! :smiley:

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorExecutionMetrics.java
```diff
@@ -0,0 +1,103 @@
...
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
```
bernd: Duplicate license.

bernd:
@mpfz0r asked me to do the fixes because he had to leave. :wink:

PR #6244, opened 2019-08-06
Check stream permissions in aggregation event definition

bernd:
Without this, any use with event definition create permission could
configure alerts for any stream.


bernd:
I just noticed an issue. Please wait with the review until I add the label back.

bernd:
Should be ready for review now.

PR #6250, opened 2019-08-06
Implement permission check for events search

### LINKED ISSUE #6238: Check permissions in events search
## Description

The `EventsResource` is currently not checking any stream permissions. It must take the allowed streams for the user into account and add a filter on the `source_streams` field of events.

## Tasks

- [ ] tbd

bernd:
The events search code is now checking if the user that executes the
search is allowed to see events based on the "source_streams" field.

Example:

If an event got created from a message in stream A, the event has stream
A in its "source_stream" field. If a user has read permissions on stream
A, the event is visible to the user in the events search.

Closes #6238


edmundoa:
The changes seem to work as expected when events have a single source stream. However, I noticed that events with several source streams can be viewed by users with permissions to read any of those streams. Is that intended?

bernd:
> The changes seem to work as expected when events have a single source stream. However, I noticed that events with several source streams can be viewed by users with permissions to read any of those streams. Is that intended?

@edmundoa Good point! The PR basically implements an `OR` query at the moment but it should actually implement an `AND` query for the streams in `source_streams`. Otherwise users *might* see data in events from other streams they don't have access to.

I will come up with a fix. Thanks again!

bernd:
@edmundoa Can you please check again? It should be fixed now.

PR #6256, opened 2019-08-07
Avoid using `Stream#peek` for counting items in paginated result set.

### LINKED ISSUE #6254: Pagination for views list always shows only single page
## Expected Behavior

When more than 10 views exist, the views list should show multiple pages in the pagination.

## Current Behavior

When more than 10 views exist, only a single page is shown.

1. Create more than 10 views
2. Go to views list
3. Try to access view no. 11

BrokenViewsListPagination

## Your Environment

* Graylog Version: 3.1.0-rc.2-SNAPSHOT
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

Before this change, in `PaginatedDbService#findPaginatedWithQueryFilterAndSort` a single stream was used to filter (based on the query), paginate (based on page size and page number) and determine the number of total items (before pagination). The latter is performed using `Stream#peek`, which has the following characteristics:

  1. The docs say: `This method exists mainly to support debugging, where you want to see the elements as they flow past a certain point in a pipeline`
  2. `peek` is an intermediate operation on a stream, which means that
    a) it's actual execution is driven by the terminal operation of the stream
    b) it has the benefit in this case that a single stream execution is able to perform multiple tasks

The code before this PR relies on `peek` being executed for each item on the stream before pagination (`skip`/`limit`) is performed. This is the case for OracleJDK up to and including 1.8.0_202.  

Unfortunately OpenJDK (tested with `1.8.0_222`) is performing an optimization, where the actual stream execution is determined by the terminal operation (see 2a) and `peek` only sees items after filtering _and_ pagination (`Streams are lazy; computation on the source data is only performed when the terminal operation is initiated, and source elements are consumed only as needed.`).

Therefore, this change is replacing the usage of `peek` with a simpler variant, which is using two separate stream instantiations for counting the total items and the actual pagination.

Fixes #6254.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
@@ -153,17 +153,22 @@ public abstract class PaginatedDbService<DTO> {
                                                                      int page,
                                                                      int perPage) {
         // Calculate the total amount of items matching the query/filter, but before pagination
-        final long total = streamQueryWithSort(query, sort).filter(filter).count();
-
-        // Then create another filtered stream and only collect the entries according to page and perPage
-        Stream<DTO> resultStream = streamQueryWithSort(query, sort).filter(filter);
-        if (perPage > 0) {
-            resultStream = resultStream.skip(perPage * Math.max(0, page - 1)).limit(perPage);
+        final long total;
+        try (final Stream<DTO> cursor = streamQueryWithSort(query, sort)) {
+            total = cursor.filter(filter).count();
         }
 
-        final long grandTotal = db.count();
+        // Then create another filtered stream and only collect the entries according to page and perPage
+        try (final Stream<DTO> resultStream = streamQueryWithSort(query, sort)) {
+            Stream<DTO> filteredResultStream = resultStream.filter(filter);
+            if (perPage > 0) {
+                filteredResultStream = filteredResultStream.skip(perPage * Math.max(0, page - 1)).limit(perPage);
+            }
 
-        return new PaginatedList<>(resultStream.collect(Collectors.toList()), Math.toIntExact(total), page, perPage, grandTotal);
+            final long grandTotal = db.count();
+
+            return new PaginatedList<>(filteredResultStream.collect(Collectors.toList()), Math.toIntExact(total), page, perPage, grandTotal);
+        }
 }
 
     /**
--- a/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
@@ -30,7 +30,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -153,25 +152,18 @@ public int delete(String id) {
                                                                      DBSort.SortBuilder sort,
                                                                      int page,
                                                                      int perPage) {
-        final AtomicInteger total = new AtomicInteger(0);
+        final long total = streamQueryWithSort(query, sort).filter(filter).count();
```
bernd: Please make sure to close the stream because it's backed by a MongoDB cursor. (as documented in the `streamQueryWithSort()` javadoc)

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -153,25 +152,18 @@ public int delete(String id) {
...
-                    .filter(filter)
-                    .peek(dto -> total.incrementAndGet());
-
-            // Then use that filtered stream and only collect the entries according to page and perPage
-            Stream<DTO> resultStream = Stream.of(dtoStream).flatMap(stream -> stream);
-            if (perPage > 0) {
-                resultStream = resultStream.skip(perPage * Math.max(0, page - 1)).limit(perPage);
-            }
+        // Then use that filtered stream and only collect the entries according to page and perPage
+        Stream<DTO> resultStream = streamQueryWithSort(query, sort).filter(filter);
```
bernd: Same as above, please make sure to close the stream.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
```diff
@@ -153,25 +152,18 @@ public int delete(String id) {
...
-            final Stream<DTO> dtoStream = cursor
-                    .filter(filter)
-                    .peek(dto -> total.incrementAndGet());
-
-            // Then use that filtered stream and only collect the entries according to page and perPage
-            Stream<DTO> resultStream = Stream.of(dtoStream).flatMap(stream -> stream);
-            if (perPage > 0) {
-                resultStream = resultStream.skip(perPage * Math.max(0, page - 1)).limit(perPage);
-            }
+        // Then use that filtered stream and only collect the entries according to page and perPage
```
bernd: I think the comment is outdated now because the first part is missing.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
dennisoelkers: Thanks for the reminder, I was actually wondering about the relevance of that comment, as it should be taken care of the `onClose` handler which is [added to the stream upon instantiation](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java#L213), or am I wrong?

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
dennisoelkers: I took the chance to rephrase this one/add a comment to also indicate what the total is for (count of items after filtering but before pagination).

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
bernd: The `onClose` handler is called when a stream is closed by calling `.close()` or using something like try-with-resources. At least to my knowledge. :smiley: 

The Java Stream docs say:

> Streams have a BaseStream.close() method and implement AutoCloseable, but nearly all stream instances do not actually need to be closed after use. **Generally, only streams whose source is an IO channel (such as those returned by Files.lines(Path, Charset)) will require closing.**

Source: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html

I think the last part in that excerpt is relevant to our use case since the returned stream is backed by a MongoDB cursor object.

bernd: That said, the MongoDB server will close the cursor automatically at some point:

> Instructs the server to close a cursor and free associated server resources. The server will automatically close cursors that have no remaining results, as well as cursors that have been idle for a period of time and lack the cursor.noCursorTimeout() option.

Source: https://docs.mongodb.com/manual/reference/method/cursor.close/

So while it's not a huge issue not closing the cursor, I think we should always try to close the cursor if we can to avoid having dangling cursors.

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
bernd: Thank you!

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog2/database/PaginatedDbService.java
dennisoelkers: ✅

PR #6270, opened 2019-08-09
Replacing dashboards list page with views-based page.

dennisoelkers:
## Description

The goal of this PR is to replace the current dashboards list page with one that lists existing views-based dashboards. For this, it does:

  * Add a REST resource that fetches and returns  all views with type `dashboard`
    * Currently this resource is mounted at `/views/dashboard`, but this should go to `/dashboards` when the legacy code is removed
  * Add a Reflux store that fetches, searches and lists all of these dashboards
  * Add a `DashboardsPage` based on the `ViewManagementPage` listing these

Screen Shot 2019-08-09 at 10 10 00

**Note:**

Requires #6265, #6271 to be merged before.


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/ViewsResource.java
```diff
@@ -14,6 +14,22 @@
  * You should have received a copy of the GNU General Public License
  * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
  */
+/**
```
kmerz: This is a bit redundant!

graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
```diff
@@ -14,6 +14,22 @@
...
+ * (at your option) any later version.
+ * <p>
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * <p>
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
kmerz: A bit redundant!

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/ViewsResource.java
dennisoelkers: Indeed! D:

PR #6272, opened 2019-08-09
Implement new getPackageName function for versionConstraints

kmerz:
Prior to this change, the correlation config did not add its enterprise
plugin constraint to the constraints list.

This change adds a new function to the ContentPackable constraint:
getPackageName.

This is necessary since the package name of CorrelationConfig is
graylog.plugins.events but we need graylog.plugin.enterprise since
it is the name of the constraint.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/contentpack/facade/EventDefinitionFacade.java
+++ b/graylog2-server/src/main/java/org/graylog/events/contentpack/facade/EventDefinitionFacade.java
@@ -19,6 +19,7 @@ package org.graylog.events.contentpack.facade;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.graph.Graph;
 import com.google.common.graph.GraphBuilder;
 import com.google.common.graph.ImmutableGraph;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog/events/contentpack/facade/EventDefinitionFacade.java
```diff
@@ -86,6 +94,15 @@ private Entity exportNativeEntity(EventDefinitionDto eventDefinition, EntityDesc
         return Optional.of(exportNativeEntity(eventDefinition.get(), entityDescriptorIds));
     }
 
+    private ImmutableSet<Constraint> versionConstraints(EventDefinitionDto eventDefinitionDto) {
+        // The report facade is in the enterprise plugin so we need to add a constraint for that
```
bernd: This looks like a leftover from reporting code?

graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackable.java
```diff
@@ -24,4 +25,9 @@
     default void resolveNativeEntity(EntityDescriptor entityDescriptor,
                                      MutableGraph<EntityDescriptor> mutableGraph) {
     }
+
+    @JsonIgnore
+    default String getPackageName() {
```
bernd: What do you think about using a more content pack specific name for this? As it its now, the method name is pretty generic. How about `getContentPackPluginPackage()` or something similar?

Issue #6163, opened 2019-07-18
Improve handling of Email Notifications with no recipients


edmundoa:
Assigning an Email Notification with no recipients to an Event Definition ends up throwing an exception in the Graylog logs:

```
2019-07-18 16:30:28,286 ERROR: org.graylog.events.scheduler.JobExecutionEngine - Job execution error - trigger=5d30824793a96e56fe7574c3 job=5d30683b93a96e4c75f16244
org.graylog.events.scheduler.JobExecutionException: Cannot send emails: empty recipient list.
    at org.graylog.events.notifications.jobs.EmailNotificationJob.execute(EmailNotificationJob.java:111) ~[classes/:?]
    at org.graylog.events.scheduler.JobExecutionEngine.executeJob(JobExecutionEngine.java:117) ~[classes/:?]
    at org.graylog.events.scheduler.JobExecutionEngine.handleTrigger(JobExecutionEngine.java:95) ~[classes/:?]
    at org.graylog.events.scheduler.JobExecutionEngine.lambda$execute$0(JobExecutionEngine.java:70) ~[classes/:?]
    at org.graylog.events.scheduler.worker.JobWorkerPool.lambda$execute$0(JobWorkerPool.java:93) ~[classes/:?]
    at com.codahale.metrics.InstrumentedExecutorService$InstrumentedRunnable.run(InstrumentedExecutorService.java:181) [metrics-core-4.0.3.jar:4.0.3]
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_191]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_191]
    at com.codahale.metrics.InstrumentedThreadFactory$InstrumentedRunnable.run(InstrumentedThreadFactory.java:66) [metrics-core-4.0.3.jar:4.0.3]
    at java.lang.Thread.run(Thread.java:748) [?:1.8.0_191]
Caused by: org.graylog.events.notifications.jobs.EmailSender$ConfigurationError: Cannot send emails: empty recipient list.
    at org.graylog.events.notifications.jobs.EmailSender.sendEmails(EmailSender.java:172) ~[classes/:?]
    at org.graylog.events.notifications.jobs.EmailNotificationJob.execute(EmailNotificationJob.java:106) ~[classes/:?]
    ... 9 more
```

[...]

PR #6286, opened 2019-08-13
Dont throw exception on misconfigured email notification

mariussturm:
This PR reduces exceptions in cases of an empty Email receiver list. The lists are already validated in the frontend and the API. So this case indicates a broken setup and we should not retry to send the notification.

Fixes: https://github.com/Graylog2/graylog2-server/issues/6163

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/notifications/types/EmailSender.java
+++ b/graylog2-server/src/main/java/org/graylog/events/notifications/types/EmailSender.java
@@ -180,9 +180,10 @@ public class EmailSender {
 
         if (emailRecipients.isEmpty()) {
             if (!emailConfig.isEnabled()) {
-                throw new ConfigurationError("Email transport is not enabled in server configuration file!");
+                LOG.debug("Email transport is not enabled in server configuration file!");
             }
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog/events/notifications/types/EmailSender.java
```diff
@@ -180,9 +180,10 @@ void sendEmails(EmailEventNotificationConfig notificationConfig, EventNotificati
 
         if (emailRecipients.isEmpty()) {
             if (!emailConfig.isEnabled()) {
```
mpfz0r: why is this only checked if the recipients are empty? looks like a bug.

mariussturm: It's the behaviour of the current EmailAlarmCallback, but I guess we can change it. Sounds also better to me to check it independently.  https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java#L91

PR #6294, opened 2019-08-13
Fix bytes formatting after numeral major update.

### LINKED ISSUE #6175: Index size display error
## Expected Behavior

When modifying the index settings for rotation, you can set the index size to trigger a rotation.

In previous versions, the index size got displayed human-friendly inside an input group on the right side of the input field as a multiplier of 1024. **Since 3.1 it seems to be 1000.**
Screenshot 2019-07-22 at 14 16 22
Screenshot 2019-07-22 at 14 16 28

## Your Environment

* Graylog Version: Graylog 3.1.0-beta.2
* Browser version: Brave Version 0.66.100 Chromium: 75.0.3770.142 (Official Build) (64-bit)

bernd:
Numeral 2.0 changed the "b" format modifier and we now have to use "ib"
to get the Graylog 3.0 (and earlier) behavior.

Relevant changelog entry:

> Breaking change / Feature: Bytes are now formatted as: b (base 1000) and ib (base 1024)

Switch TrafficGraph.jsx to use NumberUtils instead of using numeral
directly.

Fixes #6175

### DIFF
--- a/graylog2-web-interface/src/components/cluster/TrafficGraph.jsx
+++ b/graylog2-web-interface/src/components/cluster/TrafficGraph.jsx
@@ -5,7 +5,6 @@ import HistogramVisualization from 'components/visualizations/HistogramVisualiza
 import _ from 'lodash';
 import moment from 'moment';
 import crossfilter from 'crossfilter';
-import numeral from 'numeral';
 import DateTime from 'logic/datetimes/DateTime';
 import NumberUtils from 'util/NumberUtils';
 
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/cluster/TrafficGraph.jsx
```diff
@@ -7,6 +7,7 @@ import moment from 'moment';
 import crossfilter from 'crossfilter';
 import numeral from 'numeral';
```
mpfz0r: `numeral` is not used anymore

bernd: Good catch, thanks!

PR #6305, opened 2019-08-16
Don't block server shutdown in FakeHttpMessageInput

mpfz0r:
A long sleep time for the Random HTTP message generator
will prevent the server from shutting down.

Excplicitely interrupt the AbstractExecutionThreadService runThread
to abort the sleep.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/GeneratorTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/GeneratorTransport.java
@@ -47,6 +47,8 @@ public abstract class GeneratorTransport extends ThrottleableTransport {
     @Override
     public void doLaunch(final MessageInput input) throws MisfireException {
         generatorService = new AbstractExecutionThreadService() {
+            Thread runThread;
+
             @Override
             protected void run() throws Exception {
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/GeneratorTransport.java
```diff
@@ -60,6 +62,16 @@ protected void run() throws Exception {
...
             }
+
+            @Override
+            protected void startUp() throws Exception {
+                runThread = Thread.currentThread();
+            }
+
+            @Override
+            protected void triggerShutdown() {
+                runThread.interrupt();
```
bernd: Are there situations where `runThread` might be `null` because `triggerShutdown()` is called without calling `startUp` before`? If so, we should probably add a null check for `runThread` here.

mpfz0r: I think the service lifecycle guarantees prevent that from happening: https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/Service.html

PR #6379, opened 2019-09-03
Bookmarking for saved searches

kmerz:
# Description
Add 3 buttons for bookmarking searches in the search as view.
1. Button brings back a clean search
1. Button lets save a view to the server
1. Button lets load a view from server

## Motivation and Context
The old search had saved search, but they were only for saving the query and the timerange.
The new bookmarking feature also saves the widgets and there poisitions.

Peek 2019-09-03 14-38

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/searchbar/bookmark/BookmarkControls.jsx
```diff
@@ -0,0 +1,192 @@
...
+    const { viewStoreState } = this.props;
+    const { view } = viewStoreState;
+
+    if (!view.id) {
+      return;
+    }
+
+    const newView = view.toBuilder()
+      .title(newTitle)
+      .type(View.Type.Search)
```
dennisoelkers: I think this should be set upon creation of the search instead, that means in `NewSearchPage`. No need to change this though, as this will be on in #6390.

graylog2-web-interface/src/views/components/searchbar/bookmark/BookmarkControls.jsx
```diff
@@ -0,0 +1,192 @@
...
+      .then(this.toggleFormModal)
+      .then(() => UserNotification.success(`Saving view "${newView.title}" was successful!`, 'Success!'))
+      .catch(error => UserNotification.error(`Saving view failed: ${this._extractErrorMessage(error)}`, 'Error!'));
+  };
+
+  _extractErrorMessage = (error) => {
+    return (error
+      && error.additional
+      && error.additional.body
+      && error.additional.body.message) ? error.additional.body.message : error;
```
dennisoelkers: I was thinking about how to make this nicer. lodash offers `get`, but using a string for the path breaks type safety and refactorings. Maybe something like this:

```
const orDefault = (fn, dflt) => { try { return fn(); } catch (_) { return dflt; }; };
```

it would be

```
return orDefault(() => error.additional.body.message, error);
```
then.

graylog2-web-interface/src/views/components/searchbar/bookmark/BookmarkList.test.jsx
```diff
@@ -0,0 +1,150 @@
...
+    it('should handle with load', () => {
+      const onLoad = jest.fn(() => { return new Promise(() => {}); });
+      const views = {
+        pagination: {
+          total: 0,
+          page: 0,
+          perPage: 5,
+        },
+        list: [
+          View.builder()
```
dennisoelkers: I think the `View` fixture creation should be extracted into a constant or function to prevent repeating yourself (and to make it easier to apply changes like adding another mandatory property in the future).

graylog2-web-interface/src/views/components/searchbar/bookmark/BookmarkForm.test.jsx
```diff
@@ -0,0 +1,123 @@
...
+      const wrapper = mount(<BookmarkForm value="new Title"
+                                          onChangeTitle={() => {}}
+                                          saveAsSearch={() => {}}
+                                          disableCreateNew
+                                          toggleModal={onToggleModal}
+                                          isCreateNew={false}
+                                          target={() => {}}
+                                          saveSearch={() => {}} />);
+      wrapper.find('button[children="Cancel"]').simulate('click');
+      expect(onToggleModal.mock.calls.length).toBe(1);
```
dennisoelkers: Instead of checking semi-internal structures that might change, we could also do:
```suggestion
      expect(onToggleModal).toHaveBeenCalled();
```

or

```suggestion
      expect(onToggleModal).toHaveBeenCalledTimes(1);
```

PR #6380, opened 2019-09-03
Add IfSearch component

kmerz:
## Description
Add a component which only renders its children when the
context provides a ViewType of search

Also add unit test for IfDashboard

## Motivation and Context
We need a way to only render components based on which page we are.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/dashboard/IfDashboard.test.jsx
+++ b/graylog2-web-interface/src/views/components/dashboard/IfDashboard.test.jsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import 'helpers/mocking/react-dom_mock';
+
+import View from 'views/logic/views/View';
+import ViewTypeContext from 'views/components/contexts/ViewTypeContext';
+
+import IfDashboard from './IfDashboard';
+
+describe('IfDashboard', () => {
+  it('should render children with Dashboard context', () => {
+    const wrapper = renderer.create(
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/views/components/dashboard/IfDashboard.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+          <span>Fear is the mind-killer.</span>
+        </IfDashboard>
+      </ViewTypeContext.Provider>,
+    );
+
+    expect(wrapper).toMatchSnapshot();
+  });
+
+
+  it('should not render children without context', () => {
```
dennisoelkers: The description is incorrect. Children are rendered, because the context defaults to `View.Type.Dashboard`. This is also reflected in the snapshot.

graylog2-web-interface/src/views/components/dashboard/IfDashboard.test.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import 'helpers/mocking/react-dom_mock';
+
+import View from 'views/logic/views/View';
+import ViewTypeContext from 'views/components/contexts/ViewTypeContext';
+
+import IfDashboard from './IfDashboard';
+
+describe('IfDashboard', () => {
```
dennisoelkers: ❤️ for including a `IfDashboard` test!

PR #6386, opened 2019-09-04
Font Awesome Icon component abstraction

kyleknighted:
## Description

Migrate the use of `<i className="fa fa-foobar" />` to an actual component that we can manage globally `<Icon name="foobar" />`

## Motivation and Context

- Provide a centralized component for Icons that can be styled globally as necessary 
- Allows for updating or changing icon libraries with a single API

## Types of changes

## Checklist:


graylog2-web-interface/src/components/graylog/Icon.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import 'font-awesome/css/font-awesome.css';
+
+const Icon = ({ ...props }) => {
```
dennisoelkers: What about passing a `type` prop, that specifies the name of the icon? For example for `<Icon type="times" />` it generates `<i className="fa fa-times" />`. This way we could also generate a mapping translating from font awesome to something else, if we move away from it.

kyleknighted: That's the intention, just going through one step at a time for clearer commit history for easier review 😄

graylog2-web-interface/src/components/extractors/extractors_configuration/GrokExtractorConfiguration.jsx
```diff
@@ -133,7 +133,7 @@ class GrokExtractorConfiguration extends React.Component {
         <Row>
           <Col mdOffset={2} md={1}>
             <Button bsStyle="info" onClick={this._onTryClick} disabled={this._isTryButtonDisabled()}>
-              {trying ? <i className="fa fa-spin fa-spinner" /> : 'Try against example'}
+              {trying ? <Icon name="spinner" spin /> : 'Try against example'}
```
dennisoelkers: This and similar ones might also be replaceable with the `Spinner` component.

kyleknighted: Yea, i think this is a good idea, but will most likely do it within a follow-up PR

graylog2-web-interface/src/components/common/Icon.jsx
```diff
@@ -0,0 +1,77 @@
...
+  spin,
+  ...props
+}, ref) => {
+  const cleanIconName = name.replace(/^fa-/, ''); // remove "fa-" prefix if it exists
+
+  return (
+    <i className={`
+         fa
+         fa-${cleanIconName}
+         ${flip && `fa-flip-${flip}`}
```
dennisoelkers: All of these optional parameters do generate unwanted values (`undefined`/`false`) if they are not set because of these conditionals. Instead, we should build the class names string conditionally and only add classes if the properties are true, instead of adding the result of the conditionals.

You can see the effect of the current logic pretty well in the generated snapshots.

linuspahl:
We are using the the font awesome icons also in graylog-plugin-collector and graylog-plugin-enterprise. Is the plan to replace them with the Icon component there as well? Would we do this in another PR?

PR #6390, opened 2019-09-05
Providing type of actual view instead of hardcoding one in context provider.

dennisoelkers:
Before this change, creating a new search/dashboard or loading one was
explicitly providing the view type by calling the
`ViewTypeContext.Provider` with a static value. This was prone to errors
caused by forgetting to do this in other places, leading to the default
being used.

Instead of this, this PR is now consolidating the context providers to a
single place that takes the view type from the currently loaded view.


graylog2-web-interface/src/views/components/dashboard/__snapshots__/IfDashboard.test.jsx.snap
```diff
@@ -22,8 +22,5 @@ exports[`IfDashboard should render children without context since Dashboard is t
   <span>
```
kmerz: Titlte nees update now!

PR #6413, opened 2019-09-05
Remove Event indices from global search default

### LINKED ISSUE #6384: Remove Event indices from global search
We should not include the new alerting event indices in the default search.
First of all, events are special and are not really meant to be viewed with the search UI.
Second of all, they create unnecessary failures if users want to sort an fields that
are not present in the Event indices.

mpfz0r:
Fixes #6384

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchResource.java
@@ -84,6 +84,7 @@ import java.util.stream.Stream;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static java.util.stream.Collectors.toSet;
+import static org.graylog2.plugin.streams.Stream.DEFAULT_EVENT_STREAM_IDS;
 
 // TODO permission system
 @Api(value = "Enterprise/Search", description = "Searching")
@@ -203,6 +204,8 @@ public class SearchResource extends RestResource implements PluginRestResource {
                 throwStreamAccessForbiddenException();
             }
 
+            // Unless explicitly queried, exclude event indices by default
+            allAvailableStreamIds.removeAll(DEFAULT_EVENT_STREAM_IDS);
 
             final ImmutableSet<Query> newQueries = search.queries().stream().map(query -> {
                 if (query.usedStreamIds().isEmpty()) {
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchResource.java
@@ -205,6 +205,9 @@ public class SearchResource extends RestResource implements PluginRestResource {
             }
 
             // Unless explicitly queried, exclude event indices by default
+            // Having the event indices in every search, makes sorting almost impossible
+            // because it triggers https://github.com/Graylog2/graylog2-server/issues/6378
+            // TODO this can be removed, once we implement https://github.com/Graylog2/graylog2-server/issues/6490
             allAvailableStreamIds.removeAll(DEFAULT_EVENT_STREAM_IDS);
 
             final ImmutableSet<Query> newQueries = search.queries().stream().map(query -> {
--- a/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -52,6 +52,7 @@ import org.elasticsearch.search.aggregations.bucket.terms.Terms;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.elasticsearch.search.fetch.subphase.highlight.HighlightBuilder;
 import org.graylog2.Configuration;
+import org.graylog2.configuration.ElasticsearchConfiguration;
 import org.graylog2.database.NotFoundException;
 import org.graylog2.indexer.ElasticsearchException;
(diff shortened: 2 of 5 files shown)


lbogdan:
Can we please get this merged and cut a new release as soon as possible, as it's kind of a blocker since upgrading to 3.1 broke sorting? (it works if I specifically go to the "All messages" stream and search / sort from there, but it just feels like a workaround, not a proper workflow)

Thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchResource.java
```diff
@@ -203,6 +204,8 @@ public Response executeQuery(@ApiParam(name = "id") @PathParam("id") String id,
                 throwStreamAccessForbiddenException();
             }
 
+            // Unless explicitly queried, exclude event indices by default
+            allAvailableStreamIds.removeAll(DEFAULT_EVENT_STREAM_IDS);
```
bernd: Please extend the comment to describe why we are doing this and also add the link to https://github.com/Graylog2/graylog2-server/issues/6490 as a `TODO` so we know that it can be removed once that issue got implemented. Thanks!

graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
```diff
@@ -915,6 +919,12 @@ private QueryBuilder standardAggregationFilters(TimeRange range, String filter)
         for (IndexRange indexRange : indexRanges) {
             // if we aren't in a stream search, we look at all the ranges matching the time range.
             if (indexSet == null && filter == null) {
+                // Exclude Event indices by name
+                // TODO This will fail if the prefix settings are changed. But it's the only info we have in IndexRanges
+                if (indexRange.indexName().startsWith(elasticsearchConfiguration.getDefaultEventsIndexPrefix()) ||
```
bernd: I think we can check if the index is part of an event index set so we don't have to rely on the default events index prefixes.

Something like this could work (untested :smile:):

```diff
diff --git graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
index 0091396d1..d39f8962a 100644
--- graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
+++ graylog2-server/src/main/java/org/graylog2/indexer/searches/Searches.java
@@ -60,6 +60,7 @@ import org.graylog2.indexer.IndexMapping;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.IndexSetRegistry;
 import org.graylog2.indexer.cluster.jest.JestUtils;
+import org.graylog2.indexer.indexset.IndexSetConfig;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.indexer.ranges.IndexRange;
 import org.graylog2.indexer.ranges.IndexRangeService;
@@ -912,9 +913,21 @@ public class Searches {
 
         final ImmutableSortedSet.Builder<IndexRange> indices = ImmutableSortedSet.orderedBy(IndexRange.COMPARATOR);
         final SortedSet<IndexRange> indexRanges = indexRangeService.find(range.getFrom(), range.getTo());
+        final Set<String> affectedIndexNames = indexRanges.stream().map(IndexRange::indexName).collect(Collectors.toSet());
+        final Set<IndexSet> nonEventIndexSets = indexSetRegistry.getForIndices(affectedIndexNames).stream()
+                .filter(indexSet1 -> !IndexSetConfig.TemplateType.EVENTS.equals(indexSet1.getConfig().indexTemplateType().orElse(IndexSetConfig.TemplateType.MESSAGES)))
+                .collect(Collectors.toSet());
+
         for (IndexRange indexRange : indexRanges) {
             // if we aren't in a stream search, we look at all the ranges matching the time range.
             if (indexSet == null && filter == null) {
+                // Don't include the index range if it's for an event index set to avoid sorting issues.
+                // See the following issues for details:
+                // - https://github.com/Graylog2/graylog2-server/issues/6384
+                // - https://github.com/Graylog2/graylog2-server/issues/6490
+                if (nonEventIndexSets.stream().noneMatch(set -> set.isManagedIndex(indexRange.indexName()))) {
+                    continue;
+                }
                 indices.add(indexRange);
                 continue;
             }
```

mpfz0r: Hmm, I don't see how this is any different.
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/MongoIndexSet.java#L106

If the user changes the event index name, we will still fail to exclude the old indices that are created before the renaming took place.

mpfz0r: Ah, nevermind. We would still have the IndexSets with the old names then. :)

bernd: The difference is, that the user cannot change the index prefix for existing index sets. The prefix can only be set when creating an index set. The default prefix can be changed any time in the config file. I agree that this probably doesn't make a huge difference in practice at the moment. But once we allow users to create event index sets on their own, the code that checks the default event index set prefixes will break.

bernd:
@lbogdan We will cut a 3.1.3 release that includes this fix soon.

lbogdan:
@bernd Thank you! Sorry for the harshness, I was a bit annoyed that, out of the blue, sorting wasn't working. You folks are doing a wonderful job!

PR #6462, opened 2019-09-19
Do not create compound types if types differ by properties.

### LINKED ISSUE #6313: Can't aggregate "source" field because it is compound(string,string)
I'm trying to aggregate on the `source` field but I can't because it's reported as `compound(string,string)`, which doesn't seem to make much sense.

source

What could be the reason for this? Can we avoid it?

Graylog v3.1.0

dennisoelkers:
## Motivation and Context
Before this change, compound types were created also for these types:

  * `text` -> `string{properties=[full-text-search]}`
  * `keyword` -> `string{properties=[enumerable]}`

This is incorrect, as these are the same (mapped) types, but vary only
on capabilities.

This change is now merging these two types into a single non-compound
type, with a property set that is the intersection of the given
properties of all merged types.

Fixes #6313.

## Types of changes

## Checklist:


lennartkoopmann:
:pray: Thank you!

PR #6474, opened 2019-09-20
Conceptual Scratchpad Feature

kyleknighted:
## Types of changes

## Checklist:


dennisoelkers:
I like the feature. We could also potentially add actions e.g. in messages allowing us to push references/links to them to the scratch pad.

I do think though that storing the content in local storage is not ideal. It would mean that logging in as the same user from a different browser/machine would make contents previously worked on unavailable. Instead we should periodically sync it to the backend.

kyleknighted:
> I do think though that storing the content in local storage is not ideal.

I didn't plan on leaving this forever, but I haven't researched how to add things to the user preferences yet.

PR #6475, opened 2019-09-22
Link to the latest documentation and use HTTPS

supahgreg:
## Description
This PR switches links to [docs.graylog.org](https://docs.graylog.org/) from HTTP to HTTPS (now that it is available) and switches some links from specific versions to `latest`.

## Motivation and Context

The usual reasons to prefer HTTPS: integrity, privacy, etc.

Switching to `latest` ensures users see the most up-to-date documentation.

## How Has This Been Tested?

Accessed all updated links to ensure they still work and are relevant.

## Types of changes

## Checklist:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6,7 +6,7 @@
 
 Welcome! _Graylog_ is an open source log management platform.
 
-You can read more about the project on our [website](https://www.graylog.org/) and check out the [documentation](http://docs.graylog.org/) on the documentation site.
+You can read more about the project on our [website](https://www.graylog.org/) and check out the [documentation](https://docs.graylog.org/) on the documentation site.
 
 
 ## Issue Tracking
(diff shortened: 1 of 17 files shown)


jalogisch:
@supahgreg while having the documentation linked to https is really a good catch it is on purpose that the linked version of documentation is always the same as the server version. 

This way a user never get confused that something is not like in his version

supahgreg:
@jalogisch Okay, makes sense.  I've switched `UPGRADING.rst` back to `3.0`, but left some other files (which were referencing `2.x`) as `latest` since it seems they haven't been getting updated with each release.  Let me know if you'd prefer `3.1`, `3.2`, or something else in any of those cases.

supahgreg:
@jalogisch Although I understand your point, it seems links in `graylog.conf` haven't been kept in sync with the release (as the current links are to `2.3`).  What would you like to see happen?  Link to the current `2.3`, or switch to something like `3.1`, `3.2`, or `latest` (until it becomes incorrect)?

PR #6518, opened 2019-10-01
Add error handling to LookupTable Results / Add Caffeine based LookupCache

### LINKED ISSUE #6295: DNS lookup adapter fills cache with failed lookup results
If you misconfigure a dns lookup adapter (using a non-responding DNS server IP),
it seems to fill the cache with `null` results.

After fixing your error, the lookup will still return no results for a previously
queried and cached domain.
After purging the cache, everything is back to normal.

## Expected Behavior
Lookup failures should be treated differently than `NXDOMAIN` results.

## Current Behavior
I'm just guessing here, but it seems any error is treated as `NXDOMAIN`

## Possible Solution
Distinguish between those errors and don't insert lookup errors into the cache.

1. Configure DNS data adapter to use a wrong DNS server (e.g. `1.2.3.4`)
2. Purge cache
3. Wait until input processing looked up some messages (with empty results)
4. Configure a valid DNS server
5. Messages for previously looked up domains are still created with empty results

* Graylog Version: 3.1-rc2

mpfz0r:
This allows the users of LookupTables to distinguish
between empty results and error conditions.
LookupResults with errors will get a short TTL (5 seconds)
so they can be retried more aggressively.

Add a Caffeine based LookupCache
This allows a custom expiry time per key, so we can evict
LookupResults with errors faster from the cache.

Since the settings are identical to the GuavaLookupCache,
we can have both implementations share the same frontend components.

Fixes #6295


graylog2-server/src/main/java/org/graylog2/lookup/LookupTable.java
```diff
@@ -71,6 +71,9 @@ public String error() {
     public LookupResult lookup(@Nonnull Object key) {
         final LookupResult result = cache().get(LookupCacheKey.create(dataAdapter(), key), () -> dataAdapter().get(key));
 
+        if (result.hasError()) {
+            return result;
```
mpfz0r: Note to self: This introduces a change of behavior. Lookups with errors will not return a default value anymore

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
```diff
@@ -56,11 +68,13 @@
     @Nullable
     public abstract Map<Object, Object> multiValue();
 
+
+    @JsonProperty("has_error")
+    public abstract boolean hasError();
```
bernd: I was wondering if we might want to use an error message (or a list of error messages) instead of a simple flag. That way consumers of the lookup table API would be able to better communicate issues to the user. (e.g. showing the error message in the frontend for debugging)

What do you think?

graylog2-server/src/main/java/org/graylog2/lookup/caches/CaffeineLookupCache.java
```diff
@@ -0,0 +1,277 @@
...
+
+import javax.annotation.Nullable;
+import javax.inject.Named;
+import javax.validation.constraints.Min;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class CaffeineLookupCache extends LookupCache {
```
bernd: Would it be possible to replace the usage of the guava cache in `GuavaLookupCache` with caffeine? The caffeine cache can do everything that the guava cache can do, right? That way we don't have to explain to users which cache to use in which situations.

The internal naming in the code wouldn't be accurate anymore, but I guess that's not a huge issue. We can even rename it as long as we don't change the `NAME` value.

What do you think?

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
```diff
@@ -42,8 +44,18 @@
  */
 @AutoValue
 public abstract class LookupResult {
+    private static final long NO_TTL = Long.MAX_VALUE;
+
+    // Cache erroneous results with a shorter TTL
+    private static final long ERROR_CACHE_TTL = Duration.ofSeconds(5).toMillis();
```
bernd: I think we should make this TTL user configurable. There will be use cases where you want shorter or longer cache times of errors.

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
mpfz0r: Keeping it a boolean has the advantage that we can always return the same Object (`ERROR_LOOKUP_RESULT`) and thus save some memory.
But adding a list is okay for me as well.

graylog2-server/src/main/java/org/graylog2/lookup/caches/CaffeineLookupCache.java
mpfz0r: We could do that. Making Caffeine optional gives us the relieve that we don't break anything, though ;-)

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
mpfz0r: Globally? Or per DataAdapter?

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
bernd: Okay good point, let's keep the boolean flag then.

graylog2-server/src/main/java/org/graylog2/lookup/caches/CaffeineLookupCache.java
bernd: As discussed offline, let's migrate the existing cache over to caffeine. :smile:

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupResult.java
bernd: Per data adapter. Depending on the data you deal with, you might want to use a different TTL for errors.

mpfz0r:
I think it makes sense to merge this change (and not squash it)

graylog2-server/src/main/java/org/graylog2/lookup/adapters/HTTPJSONPathDataAdapter.java
```diff
@@ -171,14 +170,18 @@ protected LookupResult doGet(Object key) {
             if (!response.isSuccessful()) {
                 LOG.warn("HTTP request for key <{}> failed: {}", key, response);
                 httpRequestErrors.mark();
-                return LookupResult.empty();
+                return this.resultWithError;
             }
 
-            return parseBody(singleJsonPath, multiJsonPath, response.body().byteStream());
+            final LookupResult result = parseBody(singleJsonPath, multiJsonPath, response.body().byteStream());
+            if (result == null) {
```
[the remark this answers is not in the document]
mpfz0r: But `parseBody` is static and we don't have access to `resultWithError`

graylog2-server/src/main/java/org/graylog2/lookup/caches/CaffeineLookupCache.java
```diff
@@ -99,11 +121,16 @@ protected void doStop() throws Exception {
 
     @Override
     public LookupResult get(LookupCacheKey key, Callable<LookupResult> loader) {
+        final Function<LookupCacheKey, LookupResult> mapFunction = unused -> {
```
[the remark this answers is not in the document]
mpfz0r: How would I do that? It gets a different `loader` every time.

graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
```diff
@@ -432,6 +436,15 @@ protected void installLookupDataAdapter(String name,
         registerJacksonSubtype(configClass, name);
     }
 
+    protected void installLookupDataAdapter2(String name,
+                                            Class<? extends LookupDataAdapter> adapterClass,
+                                            Class<? extends LookupDataAdapter.Factory2> factoryClass,
+                                            Class<? extends LookupDataAdapterConfiguration> configClass) {
+        install(new FactoryModuleBuilder().implement(LookupDataAdapter.class, adapterClass).build(factoryClass));
+        lookupDataAdapterBinder2().addBinding(name).to(factoryClass);
+        registerJacksonSubtype(configClass, name); // TODO twice?
```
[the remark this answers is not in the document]
mpfz0r: oops, no :)

graylog2-server/src/main/java/org/graylog2/lookup/dto/DataAdapterDto.java
```diff
@@ -81,6 +101,15 @@ public static Builder builder() {
         @JsonProperty(FIELD_NAME)
         public abstract Builder name(String name);
 
+        @JsonProperty(FIELD_CUSTOM_ERROR_TTL_ENABLED)
+        public abstract Builder customErrorTTLEnabled(@Nullable Boolean enabled);
```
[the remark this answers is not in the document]
mpfz0r: Not having the option would store the default into each DataAdapter once it is edited by the user.
That means if we decide to change the default, some users might not benefit from our change, because we cannot tell, whether they wanted the default, or changed it on purpose.
It feels indeed like a lot of boilerplate, but since it's written already, I'd leave it as is :)

graylog2-server/src/main/java/org/graylog2/lookup/caches/CaffeineLookupCache.java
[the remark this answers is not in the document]
bernd: Yes, you are right. Sorry, I misread that.

graylog2-server/src/main/java/org/graylog2/lookup/dto/DataAdapterDto.java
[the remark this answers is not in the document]
bernd: Okay, fair enough. :+1:

graylog2-server/src/main/java/org/graylog2/lookup/adapters/HTTPJSONPathDataAdapter.java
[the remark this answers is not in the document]
bernd: True, missed that, sorry.

graylog2-server/src/main/java/org/graylog2/plugin/lookup/LookupDataAdapter.java
```diff
@@ -40,16 +41,26 @@
     private final LookupDataAdapterConfiguration config;
     private final Timer requestTimer;
     private final Timer refreshTimer;
+    protected LookupResult resultWithError;
```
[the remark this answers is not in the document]
mpfz0r: Oops, missed that. Fixed :)

PR #6534, opened 2019-10-04
Migrating App.jsx to functional component

kyleknighted:
## Description

Migrates from deprecated `createReactClass` library to functional component

## Motivation and Context

Bringing older components up to date with latest guidelines

## Types of changes

## Checklist:


graylog2-web-interface/src/routing/App.jsx
```diff
@@ -1,54 +1,60 @@
...
+    }),
+  }),
+  location: PropTypes.object.isRequired,
+};
+
+App.defaultProps = {
+  currentUser: undefined,
+};
+
+export default connect(App, { currentUser: CurrentUserStore });
```
dennisoelkers: We can flatten the structure of the `currentUser` prop by using a props mapper here. This would reduce the clutter a bit with the nested `currentUser.currentUser` prop needing destructuring inside the component and makes it easier to replace the `connect` with a context later. You can do it like this:

```suggestion
export default connect(App, { currentUser: CurrentUserStore }, ({ currentUser: { currentUser } = {} }) => ({ currentUser }));
```

linuspahl: I like the idea, one thought:

Would `currentUser: { currentUser } = {}` overwrite the `App.defaultProps`? In this case we would also need to modify the `if (!currentUser)` check.

dennisoelkers: The default assignment is for the case when `connect` calls `getInitialState` on the `CurrentUserStore` when it has no state currently (which should never happen™, because this means that no user is logged in), so it is just a safe coding measure. In this case, the `currentUser` prop which is passed to the `App` component would be mapped to `undefined`, which is effectively the default prop. Does this answer the question?

graylog2-web-interface/src/routing/App.jsx
```diff
@@ -1,54 +1,58 @@
...
-                    permissions={this.state.currentUser.permissions} />
-        <div id="scroll-to-hint" style={{ display: 'none' }} className="alpha80">
-          <i className="fa fa-arrow-up" />
-        </div>
-        <AppErrorBoundary>
-          {this.props.children}
-        </AppErrorBoundary>
-        <Footer />
+const App = ({ children, currentUser, location }) => {
+  if (Object.keys(currentUser).length === 0) {
```
linuspahl: Just to be sure, the `currentUser: { currentUser } = {}` is always overwriting the `App.defaultProps`, right?
I think we should still get rid of the default prop `currentUser: undefined` because it would throw an error when using it with `Object.keys(currentUser).length`

linuspahl: Ah, just saw dennis answer for the first part!

graylog2-web-interface/src/routing/App.jsx
linuspahl: Yes, thank you

graylog2-web-interface/webpack/vendor-module-ids.json
```diff
@@ -921,7 +921,30 @@
...
+      "node_modules/object-is/index.js": 92,
+      "node_modules/is-regex/index.js": 93,
+      "node_modules/has/src/index.js": 94,
+      "node_modules/function-bind/index.js": 95,
+      "node_modules/function-bind/implementation.js": 96,
+      "node_modules/regexp.prototype.flags/index.js": 97,
+      "node_modules/regexp.prototype.flags/shim.js": 98,
+      "node_modules/is-date-object/index.js": 99,
+      "node_modules/react-is/cjs/react-is.production.min.js": 100,
+      "node_modules/@babel/runtime/helpers/extends.js": 101
```
dennisoelkers: Why did this change become necessary? Usually the `vendor-module-ids.json` file should only change if either the `vendor.modules.js` or `package.json`/`yarn.lock` changes.

graylog2-web-interface/src/routing/App.jsx
linuspahl: @kyleknighted Just had a look together with Dennis and he mentioned, that in this context the above change is the only necessary one, because `({ currentUser: { currentUser } = {} }) => ({ currentUser }))` will not set the App prop currentUser to an empty object, it would still be undefined. I misread that one.

So the defaultProp for currentUser can still be undefined and the check can still be `if (!currentUser)` as well.
Sorry for the confusion.

graylog2-web-interface/webpack/vendor-module-ids.json
dennisoelkers: It might be related to doing a `yarn install` in a different branch before and not cleaning up before a `yarn build`. In any case, the files should not go into this PR, as it is unrelated to the change.

PR #6567, opened 2019-10-10
Introduce a cluster global API browser button

### LINKED ISSUE #5920: API browser behavior not as expected through load balancer
The Graylog server cluster is behind a load balancer with the **http_external_uri** configured per the manual. The **http_external_uri** corresponds to the LB VIP. The **http_publish_uri** must correspond to the local address of the Graylog server for cluster connectivity, otherwise unexpected behavior may result when communicating through the LB. 

If the **http_external_uri** is being used through the browser and the user clicks on the API browser button, then the **http_publish_uri** of the node is returned, resulting in broken browser connection. The node should instead return the **http_external_uri** to the browser. 

The current workaround is to avoid the API browser button for user education and instead send them to **http://$LB_VIP$:9000/api/api-browser** directly.

Steps to reproduce
1. Set up Graylog servers such that **http_publish_uri** is different than **http_external_uri**
2. Add NAT LB to graylog 
3. Go to System-->Nodes-->API Browser
4. Observe the location of the link returned, and how it fails mightily.

Context
Some users that do not control the environment will still need access to the API for features or metrics that are not easily available via the browser or searchable log. It is easier to have a predictable path to the API browser rather than requiring a bookmark for each Graylog server.

* Graylog Version: 3.0.1
* MongoDB Version: 4.0.2
* Operating System: Centos 7
* Browser version:

mpfz0r:
Usually, the link to the API browser is meant to talk to one specific node only.

In cases where Graylog is served behind a load balancer, the node
specific URL is not reachable.
We autodetect that ( `http_external_uri` != `http_publish_uri` )
and display a "cluster global API browser button".

Since the API browser behind this address might end up talking
to random nodes of the cluster, display a warning box, to make
users aware of this.

Fixes #5920

Refs #2360
Refs https://github.com/Graylog2/graylog2-server/pull/2587

### DIFF
--- a/graylog2-server/src/main/resources/swagger/css/screen.css
+++ b/graylog2-server/src/main/resources/swagger/css/screen.css
@@ -413,6 +413,21 @@ body #header {
     font-size: 14px;
 }
 
+.warning-box {
+    color: #9F6000;
+    background-color: #FEEFB3;
+    vertical-align:middle;
+    position: fixed;
+    top: 75px;
+    left: 20px;
+    width: 250px;
+    margin: 5px;
+    padding: 20px;
+    font-size: 14px;
+    border-radius:.5em;
+    box-shadow: 6px 6px 11px -4px rgba(0,0,0,0.43);
+}
+
 body #header a#logo {
     font-size: 1.5em;
     font-weight: bold;
--- a/graylog2-server/src/main/resources/swagger/index.html.template
+++ b/graylog2-server/src/main/resources/swagger/index.html.template
@@ -1,8 +1,9 @@
+
 <!DOCTYPE html>
 <html>
 <head>
   <title>Graylog REST API browser</title>
-  <base href="${baseUri}api-browser/">
+  <base href="${baseUri}api-browser/${globalModePath}">
   <link href='css/fonts.css' rel='stylesheet' type='text/css'/>
   <link href='css/highlight.default.css' media='screen' rel='stylesheet' type='text/css'/>
   <link href='css/screen.css' media='screen' rel='stylesheet' type='text/css'/>
@@ -22,7 +23,7 @@
     var apiTarget = "${baseUri}";
     $(function () {
       window.swaggerUi = new SwaggerUi({
-      url: apiTarget + "api-docs",
+      url: apiTarget + "api-docs" + "${globalUriMarker}",
       dom_id: "swagger-ui-container",
       supportedSubmitMethods: ['get', 'post', 'put', 'delete'],
       onComplete: function(swaggerApi, swaggerUi){
@@ -37,6 +38,10 @@
           console.log("Loaded SwaggerUI")
         }
         $('pre code').each(function(i, e) {hljs.highlightBlock(e)});
+
+        if ("${showWarning}") {
+          $('.warning-box').css('visibility', '');
+        }
       },
       onFailure: function(data) {
         if(console) {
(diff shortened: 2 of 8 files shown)


graylog2-web-interface/src/components/nodes/NodesList.jsx
```diff
@@ -9,20 +9,20 @@ import { Spinner, EntityList, Pluralize } from 'components/common';
...
     permissions: PropTypes.array.isRequired,
+    nodes: PropTypes.array.isRequired,
   },
 
-  mixins: [Reflux.connect(NodesStore), Reflux.connect(ClusterOverviewStore)],
+  mixins: [Reflux.connect(ClusterOverviewStore)],
 
   _isLoading() {
-    return !(this.state.nodes && this.state.clusterOverview);
+    return !(this.props.nodes && this.state.clusterOverview);
```
kyleknighted: props and state should be destructered

graylog2-server/src/main/resources/swagger/css/screen.css
```diff
@@ -413,6 +413,21 @@ body #header {
     font-size: 14px;
 }
 
+.warning-box {
```
kyleknighted: If we use the `Alert` component this will not be necessary

graylog2-server/src/main/resources/swagger/index.html.template
```diff
@@ -67,6 +72,13 @@
 </head>
 
 <body>
+<div class="warning-box" style="visibility:hidden;">
```
kyleknighted: We should consider moving this to a JSX file so we can use our component library

graylog2-web-interface/src/components/nodes/NodesList.jsx
```diff
@@ -9,20 +9,20 @@ import { Spinner, EntityList, Pluralize } from 'components/common';
...
 
-const NodesStore = StoreProvider.getStore('Nodes');
 const ClusterOverviewStore = StoreProvider.getStore('ClusterOverview');
 
 const NodesList = createReactClass({
   displayName: 'NodesList',
 
   propTypes: {
     permissions: PropTypes.array.isRequired,
+    nodes: PropTypes.array.isRequired,
```
kyleknighted: Do we know what this array should contain? Perhaps we can use `arrayOf` instead?

graylog2-server/src/main/resources/swagger/index.html.template
kyleknighted: Sorry, ignore this, thought it was a react template file, not swagger, probably the best way to accomplish this

graylog2-server/src/main/resources/swagger/css/screen.css
kyleknighted: Sorry, ignore this, thought it was a react css file, not swagger, probably the best way to accomplish this

graylog2-server/src/main/resources/swagger/swagger-ui.js
```diff
@@ -1435,7 +1435,7 @@ helpers = this.merge(helpers, Handlebars.helpers); data = data || {};
 
     ResourceView.prototype.render = function() {
       var operation, _i, _len, _ref4;
-      console.log(this.model.description);
+      //console.log(this.model.description);
```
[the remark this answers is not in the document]
kmerz: I agree with @kyleknighted here

graylog2-web-interface/src/components/nodes/NodesList.jsx
```diff
@@ -9,20 +9,20 @@ import { Spinner, EntityList, Pluralize } from 'components/common';
 import StoreProvider from 'injection/StoreProvider';
 import NodeListItem from './NodeListItem';
 
-const NodesStore = StoreProvider.getStore('Nodes');
 const ClusterOverviewStore = StoreProvider.getStore('ClusterOverview');
 
 const NodesList = createReactClass({
```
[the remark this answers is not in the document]
bernd: @kyleknighted @mpfz0r This PR needs to be backported to 3.1.x so we need to be careful to not change the components in a way that makes them only work with the current master. :smiley:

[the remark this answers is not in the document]
mpfz0r: I'm with bernd here. I just moved the code one level up. This bug already took longer to fix than I anticipated :)

PR #6560, opened 2019-10-09
Button component ref handling

kyleknighted:
## Description

Updating `Button` component to properly pass through ref prop

## Motivation and Context

Ref prop was not properly referring to the Bootstrap component.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #6569, opened 2019-10-10
Updating Button reference issue

kyleknighted:
## Description

- Added `useCallback` for styled-component function memoization (Thank you @kmerz for this solution!) 
- Added `forwardRef` to allow refs to be forwarded to styled button

This PR is a revision of #6560 as it had a unnecessary changes and unnecessary commits

## Motivation and Context

If you tried to use `ref` in a new themeable button, it would not properly pass the ref through and the button would rebuild in the DOM losing the context of the selector.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/bootstrap/__snapshots__/Input.test.jsx.snap
+++ b/graylog2-web-interface/src/components/bootstrap/__snapshots__/Input.test.jsx.snap
@@ -68,7 +68,7 @@ exports[`Input renders a button after the input if buttonAfter is passed 1`] = `
   addonAfter={null}
   bsStyle={null}
   buttonAfter={
-    <Button
+    <ForwardRef
       active={false}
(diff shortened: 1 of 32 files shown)


kyleknighted:
**For posterity, comment from #6560**

Would it help avoiding `useCallback` if we would move the `StyledButton` declaration to the top scope of the file and make it look like this:

```
const StyledButton = ({ active, ...rest }) => {
  const Component = styled(BootstrapButton)`${buttonStyles({ active })}`;
  return <Component {...rest} />;
};
```

Same for the following usages of `useCallback`.


> _@dennisoelkers_
---

We are unable to move the StyledButton creation outside of the component declaration because the theme stuff is all using hooks which "Hooks can only be called inside of the body of a function component."

> _@kyleknighted_

PR #6572, opened 2019-10-11
Implement function for transforming search view to dashboard view

kmerz:
## Motivation and Context
Prior this change we could not transform a serach view
into a dashboard view. But the user wants to be able
to take a search view and convert it to a dashboard view.

## Description
This change will add a new function which takes a search
view and returns a new dashboard view.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/logic/views/ViewTransformer.js
```diff
@@ -0,0 +1,32 @@
...
+
+const ViewTransformer = (searchView: View): View => {
+  const newViewStateMap: ViewStateMap = (searchView.state || Map.of()).map((viewState: ViewState, queryId: string) => {
+    const { timerange, query, filter = List.of() } = searchView.search.queries.find(q => q.id === queryId);
+
+    const widgets: List<Widget> = viewState.widgets.map((widget: Widget) => {
+      return widget.toBuilder()
+        .timerange(timerange)
+        .query(query)
+        .streams(filter.filter(value => value instanceof Map && value.get('type') === 'stream')
```
dennisoelkers: `instanceof` will not work across plugin boundaries. I would check if `value` is defined and assume it is an immutable map if it is. If you want to play real safe, you can also do a `new Map(value).get('type') === 'stream'`. 

P.S.: We should probably add an ESLint rule that forbids using `instanceof`.

graylog2-web-interface/src/views/logic/views/ViewTransformer.test.js
```diff
@@ -0,0 +1,129 @@
...
+        .search(search)
+        .build();
+      const dashboardView = viewTransformer(searchView);
+      expect(dashboardView.state.get('query-id').widgets.first().timerange).toBeUndefined();
+      expect(dashboardView.state.get('query-id').widgets.first().query).toBeUndefined();
+      expect(dashboardView.state.get('query-id').widgets.first().streams).toStrictEqual(['1234-abcd']);
+    });
+  });
+
+  describe('transform with the whole view', () => {
```
dennisoelkers: The test description is a bit too vague for me. What does the block actually do?

graylog2-web-interface/src/views/logic/views/ViewTransformer.js
```diff
@@ -0,0 +1,42 @@
...
+import { List, Map } from 'immutable';
+import Widget from 'views/logic/widgets/Widget';
+
+import type { ViewStateMap } from './View';
+import View from './View';
+import ViewState from './ViewState';
+
+const ViewTransformer = (searchView: View): View => {
+  const newViewStateMap: ViewStateMap = (searchView.state || Map.of()).map((viewState: ViewState, queryId: string) => {
+    const { timerange, query, filter = List.of() } = searchView.search.queries.find(q => q.id === queryId);
```
dennisoelkers: We could prepare a `Map<QueryId, Query>` one level up once, so we can do a cheap lookup here instead of iterating over the list.

PR #6584, opened 2019-10-14
Implement explanation how to add widgets to dashboard, if dashboard is empty.

linuspahl:
The title says it all. The explanation will work for every dashboard  query tab independently.

## How Has This Been Tested?
Added two simple tests, to test if the message will be displayed correctly.

## Types of changes

## Checklist:
- [] I have updated the documentation accordingly.

### DIFF
--- a/graylog2-web-interface/src/views/components/Query.jsx
+++ b/graylog2-web-interface/src/views/components/Query.jsx
@@ -1,13 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Immutable from 'immutable';
+import DocsHelper from 'util/DocsHelper';
 
+import { Col, Row, Jumbotron } from 'components/graylog';
+import { CurrentViewStateActions } from 'views/stores/CurrentViewStateStore';
 import { Spinner } from 'components/common';
-
 import { widgetDefinition } from 'views/logic/Widgets';
+import DocumentationLink from 'components/support/DocumentationLink';
 import WidgetGrid from 'views/components/WidgetGrid';
 import WidgetPosition from 'views/logic/widgets/WidgetPosition';
-import { CurrentViewStateActions } from 'views/stores/CurrentViewStateStore';
 import { PositionsMap, ImmutableWidgetsMap } from './widgets/WidgetPropTypes';
 
 const MAXIMUM_GRID_SIZE = 12;
@@ -56,7 +58,27 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
   );
 };
 
+const _emptyDashboardInfo = () => {
+  return (
+    <Row className="content" style={{ marginRight: 0, marginLeft: 0 }}>
+      <Col md={12}>
+        <Jumbotron style={{ marginBottom: 0 }}>
+          <h2>This dashboard has no widgets yet</h2>
+          <p>
+            You can create a new widget by selecting a wdget type in the left sidebar section &quot;Create&quot;.<br />
+            Have a look at the <DocumentationLink page={DocsHelper.PAGES.DASHBOARDS} text="documentation" /> to learn more about the widget creation.
+          </p>
+        </Jumbotron>
+      </Col>
+    </Row>
+  );
+};
+
 const Query = ({ allFields, fields, results, positions, widgetMapping, widgets, queryId }) => {
+  if (!widgets || widgets.size === 0) {
+    return _emptyDashboardInfo();
+  }
+
   if (results) {
     const content = _renderWidgetGrid(widgets, widgetMapping.toJS(), results, positions, queryId, fields, allFields);
     return (<span>{content}</span>);
--- a/graylog2-web-interface/src/views/components/Query.jsx
+++ b/graylog2-web-interface/src/views/components/Query.jsx
@@ -65,8 +65,8 @@ const _emptyDashboardInfo = () => {
         <Jumbotron style={{ marginBottom: 0 }}>
           <h2>This dashboard has no widgets yet</h2>
           <p>
-            You can create a new widget by selecting a wdget type in the left sidebar section &quot;Create&quot;.<br />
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/Query.jsx
```diff
@@ -56,7 +58,27 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
...
+            Have a look at the <DocumentationLink page={DocsHelper.PAGES.DASHBOARDS} text="documentation" />, to learn more about the widget creation.
+          </p>
+        </Jumbotron>
+      </Col>
+    </Row>
+  );
+};
+
 const Query = ({ allFields, fields, results, positions, widgetMapping, widgets, queryId }) => {
+  if (!widgets || widgets.size === 0) {
```
linuspahl: Is there a more common way to check if an Immutable Map is empty?

And I tried to use == for the comparison, because in my opinion it makes totally sense, but there is an eslint to prevent this. Does this rule makes sense in this case?

graylog2-web-interface/src/views/components/Query.jsx
```diff
@@ -56,7 +58,27 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
...
+  return (
+    <Row className="content" style={{ marginRight: 0, marginLeft: 0 }}>
+      <Col md={12}>
+        <Jumbotron style={{ marginBottom: 0 }}>
+          <h2>This dashboard has no widgets yet</h2>
+          <p>
+            You can create a new widget by selecting a widget type in the left sidebar section &quot;Create&quot;.<br />
+            Have a look at the <DocumentationLink page={DocsHelper.PAGES.DASHBOARDS} text="documentation" />, to learn more about the widget creation.
+          </p>
+        </Jumbotron>
```
linuspahl: I used the Jumbotron because I wanted to have the same font size, like the message when creating a new aggregation. The Jumbotron is responsible for the font-size. I have the feeling the Jumbotrone is the wrong component for especially the first case. Whats the preferred way to deal with different font-sizes. Shall I evaluate a possible solution, because I think we will want to use this font-size for many more cases, but without the Jumbotrons margin and padding.

graylog2-web-interface/src/views/components/Query.jsx
```diff
@@ -56,7 +58,27 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
   );
 };
 
+const _emptyDashboardInfo = () => {
```
linuspahl: I used the underscore for the name, because all other "sub" functions do, whats the preferred way? :)

graylog2-web-interface/src/views/components/Query.jsx
linuspahl: Another question I had, when do we use Immutable, is it a best practice for all "new data"

graylog2-web-interface/src/views/components/Query.jsx
linuspahl: Answer: undersore for private functions, in my case an own component makes more sense

graylog2-web-interface/src/views/components/Query.jsx
linuspahl: Answer: Yes in the mentioned cases we use the Jumbotron mainly for the font size, I will talk with @kyleknighted about a general font-size concept

graylog2-web-interface/src/views/components/Query.jsx
linuspahl: Answer: Regarding immutable, we try to use it, if possible. The empty check is okay.

Regarding the comparison: we are always using `===`

linuspahl:
Ah, I need to adjust the tests

graylog2-web-interface/src/views/components/Query.test.jsx
```diff
@@ -124,14 +121,74 @@ describe('Query', () => {
...
+             widgets={widgets}
+             onToggleMessages={() => {}}
+             queryId="someQueryId"
+             showMessages
+             allFields={Immutable.List()}
+             fields={Immutable.List()} />
+    ));
+
+    expect(wrapper.contains('You can create a new widget by selecting a widget type')).toEqual(false);
+  });
```
linuspahl: Little more context: I used the `mount` function for the first two tests, because a shallow would not render the message, when using the context provider.
I am still using the `shallow` function for the last test, because it is not necessary to use the context provider there.

graylog2-web-interface/src/views/components/Query.jsx
```diff
@@ -56,7 +60,45 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
...
+        </p>
+        <p>
+          A few tips for creating searches and dashboards
+        </p>
+        <ul>
+          <li><p>1. Start with a <b>question</b> you want to answer. Define the problem you want to solve.</p></li>
+          <li><p>2. <b>Limit</b> the data to only the data points you want to see.</p></li>
+          <li><p>3. <b>Visualize</b> the data. Does it answer your question?</p></li>
+          <li><p>4. <b>Share</b> the view with your colleagues. Prepare it for <b>reuse</b> by using parameters (contained in <a href="https://www.graylog.org/graylog-enterprise-edition" target="_blank" rel="noopener noreferrer">Graylog Enterprise</a>).</p></li>
+        </ul>
```
linuspahl: I had to write the list this way (for now), because of the described Jumbotron behaviour. (see other comments)

graylog2-web-interface/src/views/components/Query.test.jsx
dennisoelkers: ✅

graylog2-web-interface/src/views/components/Query.jsx
dennisoelkers: Good point! We can replace `view` with `dashboard` here and render the fourth point conditionally using `IfDashboard`.

PR #6585, opened 2019-10-14
Adding "Full Screen" action to dashboards configuring tv mode.

dennisoelkers:
## Motivation and Context

This change is adding a "Full Screen" action to the views actions on a
dashboard. It is supposed to generate and redirect a URL leading to a
page that shows a dashboard in a stripped down version (only widgets,
no controls) meant for passive consumption when shown on a big screen.
The action triggers a configuration modal that allows the user to set:

  * Refresh time: every n seconds the dashboard should refresh
  * Tabs: Select the tabs to be shown in rotation
  * Tab switch interval: every n seconds the dashboard cycles the tab
shown

When the user "Saves" the dialog, a redirection to a special page for
the big display mode is generated. It contains the id of the dashboard
and the aforementioned properties in url/query-encoding.

**TODO (Follow-up PR):** Handle unsaved changes in current view, warn user before leaving page to prevent losing them.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/dashboard/BigDisplayModeConfiguration.jsx
```diff
@@ -0,0 +1,127 @@
...
+const BigDisplayModeConfiguration = ({ view, open }: Props) => {
+  const [showConfigurationModal, setShowConfigurationModal] = useState(open);
+  const onCancel = useCallback(() => setShowConfigurationModal(false), [setShowConfigurationModal]);
+
+  const onSave = (config: Configuration) => redirectToBigDisplayMode(view, createQueryFromConfiguration(config));
+
+  return (
+    <React.Fragment>
+      {showConfigurationModal && <ConfigurationModal view={view} onCancel={onCancel} onSave={onSave} />}
+      <MenuItem onSelect={() => setShowConfigurationModal(true)}><i className="fa fa-desktop" /> Full Screen</MenuItem>
```
kmerz: disable={!view.id}

PR #6594, opened 2019-10-16
Allow bootstrap component style overwrite

linuspahl:
Currently we can't easily overwrite a bootstrap components style. This is mainly a problem when using styled-components in combination with bootstrap components, for example:
```
const StyledRow = styled(Row)`
  margin-left: 0; // had no effect
`
```
There are workarounds, like increasing the specificity, but with this PR we don't depend on workarounds and it works like expected.
In general we should try to not adjust the bootstrap style, but sometimes we have to.

## How Has This Been Tested?
I've clicked trough the frontend a lot and couldn't find any suspicious behavior.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/graylog/SplitButton.jsx
```diff
@@ -10,7 +10,7 @@ const SplitButton = forwardRef(({ active, bsStyle, ...props }, ref) => {
     ${buttonStyles({ active })};
 
     ~ .btn.dropdown-toggle {
-      ${buttonStyles({ active, specific: false })};
```
linuspahl: So far this is the only part where I am not totally sure what the effects are: I think we did not had to use the specificity here, because `~ . btn.dropdown-toggle` is increasing it anyway. In this case we can safely do this change.

graylog2-web-interface/src/components/graylog/styles/buttonStyles.js
```diff
@@ -10,15 +10,14 @@ const buttonStyles = ({ colors, active, utility }) => {
     const darken125 = utility.darken(color, 1.25);
 
     return css`
-    && {
```
kyleknighted: This whole file should be gone. Must've gotten mixed in with a merge somewhere. We can rely on `components/graylog/styles/button.jsx`

PR #6602, opened 2019-10-16
Showing dashboards in non-interactive big display/tv mode.

dennisoelkers:
## Motivation and Context

This PR is adding functionality to show dashboards in a non-interactive,
read-only mode suitable for big displays and unattended setups. It tries
to:

  * Maximize usage of screen estate by removing navigation and side bar
  * Removes interactive elements of widgets
  * Cycles through (configured) tabs of dashboard automatically
  * Refreshes the search in configured intervals

Fixes #6401.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #6607, opened 2019-10-17
Fixing root route in `AppRouter`.

dennisoelkers:
## Motivation and Context
In #6602 the `AppRouter`'s route structure was changed to allow supplying routes from plugins with an explicitly defined `null` value for the `parentComponent`. This enforces hoisting of these routes to the top level of the router so they are not contained within the `App` or other parent components.

Unfortunately the path of the global `AppErrorBoundary` was left in place, erroneously meant to fix a unrelated route matching issue, leading to a ` ` or `/` route to go nowhere.

This change is removing the path specification so ` ` and `/` routes are correctly routed to the `StartPage` component again.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/routing/AppRouter.test.jsx
+++ b/graylog2-web-interface/src/routing/AppRouter.test.jsx
@@ -25,7 +25,7 @@ jest.mock('injection/CombinedProvider', () => {
 });
 
 // To prevent exceptions from getting swallwoed
-jest.mock('./AppErrorBoundary',() => mockComponent('AppErrorBoundary'));
+jest.mock('./AppErrorBoundary', () => mockComponent('AppErrorBoundary'));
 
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/routing/AppRouter.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+import AppRouter from './AppRouter';
+
+jest.mock('pages', () => ({
+  StartPage: mockComponent('StartPage'),
+}));
+
+jest.mock('injection/CombinedProvider', () => {
+  const mockCurrentUserStoretore = MockStore('get', 'listen', ['getInitialState', () => ({
+    currentUser: {
+      full_name: 'Ares Vallis',
```
kmerz: :heart:

PR #6608, opened 2019-10-17
Make searchbar reponsive and add refresh controls to dashboard searchbar

linuspahl:
Initially I just wanted to add the RefreshControls to the dashboard searchbar. During the layout adjustment I couldn't resist and made the search and dashboard searchbar more responsive. 

To make the layout perfect I would have to rewrite a lot more things, but for now I fixed the general responsiveness and unified some component styles. I know the responsiveness is not that important in general, but I think it was worth the effort. I suspect it is a common scenario to have the search on one half of the screen and something else on the other.

## How Has This Been Tested?
Tested in Chrome, Safari and Firefox.

Before (on a very small screen to show most problems): 
image

After:
image 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/SearchBar.jsx
+++ b/graylog2-web-interface/src/views/components/SearchBar.jsx
@@ -72,7 +72,7 @@ const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = fal
               <form method="GET" onSubmit={submitForm}>
 
                 <Row className="no-bm extended-search-query-metadata">
-                  <Col md={3}>
+                  <Col md={4}>
                     <TimeRangeTypeSelector onSelect={newRangeType => QueriesActions.rangeType(id, newRangeType).then(performSearch)}
                                            value={rangeType} />
                     <TimeRangeInput onChange={(key, value) => QueriesActions.rangeParams(id, key, value).then(performSearch)}
@@ -85,7 +85,7 @@ const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = fal
                     <HorizontalSpacer className="mdHidden" />
                   </Col>
 
-                  <Col md={6} xs={8}>
+                  <Col md={5} xs={8}>
                     <StreamsFilter value={streams}
                                    streams={availableStreams}
                                    onChange={value => QueryFiltersActions.streams(id, value)} />
--- a/graylog2-web-interface/src/views/components/SearchBar.jsx
+++ b/graylog2-web-interface/src/views/components/SearchBar.jsx
@@ -17,6 +17,7 @@ import QueryInput from 'views/components/searchbar/AsyncQueryInput';
 import StreamsFilter from 'views/components/searchbar/StreamsFilter';
 import RefreshControls from 'views/components/searchbar/RefreshControls';
 import ScrollToHint from 'views/components/common/ScrollToHint';
+import HorizontalSpacer from 'views/components/horizontalspacer/HorizontalSpacer';
 import { QueriesActions } from 'views/stores/QueriesStore';
 import { CurrentQueryStore } from 'views/stores/CurrentQueryStore';
 import { StreamsStore } from 'views/stores/StreamsStore';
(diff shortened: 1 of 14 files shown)


graylog2-web-interface/src/views/components/SearchBar.jsx
```diff
@@ -80,18 +81,23 @@ const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = fal
                                     config={config} />
                   </Col>
 
-                  <Col md={6}>
+                  <Col mdHidden lgHidden>
+                    <HorizontalSpacer />
+                  </Col>
```
linuspahl: As you can see in the screenshots, one problem is the spacing between the Col's when they get the full width. I thought about adding a class to the specific column and controlling its bottom margin with a media query. The problem is, we can't access bootstraps media query points and would have to hard code them here and in every other comparable scenario.
It feels a bit strange to have a component only for some spacing. But this way we can use bootstraps col properties `mdHidden lgHidden` and in comparison it is more readable.

graylog2-web-interface/src/views/components/horizontalspacer/HorizontalSpacer.jsx
```diff
@@ -0,0 +1,38 @@
...
+  height: PropTypes.number,
+  className: PropTypes.string,
+};
+
+HorizontalSpacer.defaultProps = {
+  height: 10,
+  className: undefined,
+};
+
+export default HorizontalSpacer;
```
linuspahl: If you are fine with the mentioned reasons for why I added this component, I will add an example and a test.

graylog2-web-interface/src/views/components/searchbar/RefreshControls.jsx
```diff
@@ -67,27 +89,36 @@ class RefreshControls extends React.Component<Props, State> {
...
     const intervalOptions = RefreshControls.INTERVAL_OPTIONS.map(([label, interval]: [string, number]) => {
       return <MenuItem key={`RefreshControls-${label}`} onClick={() => this._onChange(interval)}>{label}</MenuItem>;
     });
-    const intervalDuration = moment.duration(this.props.refreshConfig.interval);
+    const intervalDuration = moment.duration(refreshConfig.interval);
     const naturalInterval = intervalDuration.asSeconds() < 60
       ? <span>{intervalDuration.asSeconds()} <Pluralize singular="second" plural="seconds" value={intervalDuration.asSeconds()} /></span>
       : <span>{intervalDuration.asMinutes()} <Pluralize singular="minute" plural="minutes" value={intervalDuration.asMinutes()} /></span>;
-    const buttonLabel = <span>Update every {naturalInterval}</span>;
+    const buttonLabel = this._buttonLabel(refreshConfig.enabled, naturalInterval);
```
linuspahl: I am not that happy with the naming of the var in combination with the function, but couldn't get up with a better one, what do you think?

graylog2-web-interface/src/views/components/searchbar/StreamsFilter.jsx
```diff
@@ -5,9 +5,10 @@ import Select from 'components/common/Select';
 
 const StreamsFilter = ({ disabled, value, streams, onChange }) => {
   const selectedStreams = value.join(',');
+  const placeholder = 'Select streams the search should include. Searches in all streams if empty.';
   return (
-    <div style={{ position: 'relative', zIndex: 10 }} data-testid="streams-filter">
-      <Select placeholder="Select streams the search should include. Searches in all streams if empty."
+    <div style={{ position: 'relative', zIndex: 10 }} data-testid="streams-filter" title={placeholder}>
+      <Select placeholder={placeholder}
```
linuspahl: I couldn't find a way to add the title prop to the select, so I just added it to the container. this will allow the user to read the title, even when the select will may not display the hole text due to `text-overflow: ellipsis`

graylog2-web-interface/src/views/components/horizontalspacer/HorizontalSpacer.jsx
```diff
@@ -0,0 +1,38 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+/**
+ * Component to display horisontal space.
+ * Mainly needed for responsive bootstrap grids.
+ * Can be used in combination with bootstrap Cols to show horizontal space only for specific viewports.
+ */
```
kmerz: Don't add comments please.

graylog2-web-interface/src/views/components/SearchBar.jsx
kmerz: I am cool with that solution. I like the idea! :+1:

graylog2-web-interface/src/views/components/searchbar/RefreshControls.jsx
kmerz: No sure what you are not happy about. But that's what functions are for :smile: :+1:

PR #6611, opened 2019-10-18
Add bash script to execute linting in the same way the CI does

linuspahl:
I just had the problem, that the linting for my PR fails because the CI is linting the all the changed files. I thought it is a good idea to have a yarn script for that. I think the CI script will not change in the near future

Possible cons:
- When the CI script changes, we need to adjust this script as well
- As far as I know you have to make the bash script executable by hand (I added a comment inside the script)
- There is maybe a better place than the dev dir. What do you think?

- [x] Make development easier

## Checklist:


alex-konn:
shouldn't it be possible to use the exact same script on the ci server?

PR #6617, opened 2019-10-18
Themed Alert

kyleknighted:
## Description

Updating `Alert` component to be themed

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/graylog/Alert.jsx
+++ b/graylog2-web-interface/src/components/graylog/Alert.jsx
@@ -1,6 +1,7 @@
 import React, { forwardRef, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
+// eslint-disable-next-line no-restricted-imports
 import { Alert as BootstrapAlert } from 'react-bootstrap';
 
 import { useTheme } from 'theme/GraylogThemeContext';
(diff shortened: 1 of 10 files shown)


linuspahl:
Looks good to me!
Just the linting is throwing an error.

graylog2-web-interface/src/components/graylog/Alert.jsx
```diff
@@ -0,0 +1,46 @@
+import React, { forwardRef, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import styled, { css } from 'styled-components';
+// eslint-disable-next-line no-restricted-imports
+import { Alert as BootstrapAlert } from 'react-bootstrap';
+
+import { useTheme } from 'theme/GraylogThemeContext';
+import bsStyleVariant from './variants/bsStyle';
+
+const Alert = forwardRef(({ bsStyle, ...props }, ref) => {
```
linuspahl: Just out of interest, I understand what `forwardRed` does, but why do we need it in this case?

kyleknighted: Because the ref doesn't get forwarded to the actual button if we don't, so if you try and use `<Button ref={...} />`it will give you React errors

linuspahl: alright!

PR #6621, opened 2019-10-21
Adjust query tab title editing

linuspahl:
This PR changes the way we edit a query tab title. Right now we have to double click on the current query tab title to edit it inline. There are two problems: 
- The user will may not see this option
- The react bootstrap key bindings are causing a strange behaviour when using e.g the arrow keys. The input loses the focus, a new tab will be created, the focus is still on another, not active, tab.

The new way is just another entry in the query tab context menu "Edit Title". Clicking on this entry will open a model containing the title input.

I did some refactoring of the QueryTitle and QueryTabs component in this PR (like adding flow strict). I hope you can still get an overview of the changes.

## How Has This Been Tested?
Added tests for all changes / new functionality

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/QueryBar.jsx
+++ b/graylog2-web-interface/src/views/components/QueryBar.jsx
@@ -44,7 +44,7 @@ const QueryBar = ({ children, queries, queryTitles, router, viewMetadata }) => {
   const selectQueryAndExecute = queryId => onSelectQuery(queryId);
   return (
     <QueryTabs queries={queries}
-               selectedQuery={activeQuery}
+               selectedQueryId={activeQuery}
                titles={queryTitles}
                onSelect={selectQueryAndExecute}
                onTitleChange={onTitleChange}
--- a/graylog2-web-interface/src/views/components/QueryBar.jsx
+++ b/graylog2-web-interface/src/views/components/QueryBar.jsx
@@ -1,7 +1,6 @@
 // @flow strict
 import React from 'react';
 import PropTypes from 'prop-types';
-// $FlowFixMe: imports from core need to be fixed in flow
 import { withRouter } from 'react-router';
 
 import connect from 'stores/connect';
--- a/graylog2-web-interface/src/views/components/QueryTabs.jsx
+++ b/graylog2-web-interface/src/views/components/QueryTabs.jsx
@@ -23,39 +23,24 @@ class QueryTabs extends React.Component {
     children: null,
   }
 
-  state = {
-    showTitleEditModal: false,
-    titleDraft: '',
+  openTitleEditModal = (activeQueryTitle) => {
+    if (this.queryTitleEditModal) {
+      this.queryTitleEditModal.open(activeQueryTitle);
+    }
   }
 
-  _toggleTitleEditModal = (currentTitle) => {
-    const { showTitleEditModal } = this.state;
-    // Toggle edit modal depending on current state
-    // We need to set the selected query title as the draft on open
-    // and reset the draft input on close
-    this.setState({
-      showTitleEditModal: !showTitleEditModal,
-      titleDraft: showTitleEditModal ? '' : currentTitle,
-    });
-  };
-
-  _onTitleDraftSave = () => {
-    const { titleDraft } = this.state;
-    const { onTitleChange, selectedQueryId } = this.props;
-    onTitleChange(selectedQueryId, titleDraft);
-    this._toggleTitleEditModal();
-  };
-
-  // eslint-disable-next-line no-undef
-  _onTitleDraftChange = (evt: SyntheticInputEvent<HTMLInputElement>) => {
-    evt.preventDefault();
-    evt.stopPropagation();
-    this.setState({ titleDraft: evt.target.value });
-  };
-
   render() {
-    const { children, onSelect, onRemove, queries, selectedQueryId, titles, onSaveView, onSaveAsView } = this.props;
-    const { showTitleEditModal, titleDraft } = this.state;
+    const {
+      children,
+      onRemove,
+      onSaveAsView,
+      onSaveView,
+      onSelect,
+      onTitleChange,
+      queries,
+      selectedQueryId,
+      titles,
+    } = this.props;
     const queryTitles = titles;
     const queryTabs = queries.map((id, index) => {
       const title = queryTitles.get(id, `Query#${index + 1}`);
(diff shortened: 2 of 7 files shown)


graylog2-web-interface/src/views/components/QueryBar.jsx
```diff
@@ -44,7 +43,7 @@ const QueryBar = ({ children, queries, queryTitles, router, viewMetadata }) => {
   const selectQueryAndExecute = queryId => onSelectQuery(queryId);
   return (
     <QueryTabs queries={queries}
-               selectedQuery={activeQuery}
+               selectedQueryId={activeQuery}
```
linuspahl: I renamed this prop, because I `selectedQuery` sounds like the attribute is holding the hole query. I think `selectedQueryId` is easier to understand. I would love to rename `activeQuery` to `activeQueryId`.. but that would be too much for this PR and may be you prefer the other name.

graylog2-web-interface/src/views/components/QueryTabs.jsx
```diff
@@ -1,64 +1,112 @@
...
+              onSelect={onSelect}>
+          {tabs}
+        </Tabs>
+        {/*
+          The title edit modal can't be part of the QueryTitle component,
+          due to the react bootstrap tabs keybindings.
+          The input would always lose the focus when using the arrow keys.
+        */}
+        <QueryTitleEditModal onTitleChange={(newTitle: string) => onTitleChange(selectedQueryId, newTitle)}
+                             ref={(queryTitleEditModal) => { this.queryTitleEditModal = queryTitleEditModal; }} />
```
linuspahl: Normally I would have placed this component inside the QueryTitle component. But due to the described problems I thought this is the best idea. Even now where the input is inside the modal the key binding was very strange. I was able to prevent it from creating a new query tab when using the arrow keys, but the input still lost the focus. That was very annoying when you want to e.g. select the full text

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.test.jsx
```diff
@@ -0,0 +1,79 @@
...
+    openModal(modalRef);
+    // Modal should be visible
+    expect(queryByText(modalHeadline)).not.toBeNull();
+  });
+
+  it('has correct initial input value', () => {
+    let modalRef;
+    const { getByDisplayValue } = render(
+      <QueryTitleEditModal ref={(ref) => { modalRef = ref; }}
+                           onTitleChange={() => Promise.resolve()} />,
```
linuspahl: I could not use the typical `{() => {}}` due to flow strict here. Is this the best way to deal with this case?

graylog2-web-interface/src/views/components/QueryTabs.jsx
```diff
@@ -1,64 +1,112 @@
...
+import Query from 'views/logic/queries/Query';
+import type { TitlesMap } from 'views/stores/TitleTypes';
+import View from 'views/logic/views/View';
+import ViewState from 'views/logic/views/ViewState';
 
-  const tabs = [queryTabs, newTab];
+type Props = {
+  children: React.Node,
+  onRemove: (queryId: string) => Promise<void> | Promise<ViewState>,
+  onSaveAsView: (view: View) => Promise<mixed>,
```
kmerz: why do you use `mixed` here?
Should the Promise not return the new `View`?

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+import type { TitlesMap } from 'views/stores/TitleTypes';
+
+/**
+ * Component that allows the user to update a QueryTab title.
+ * It takes the active query title as an argument on open and will use it as the draft.
+ * The open action is getting called outside by referencing this component.
+ * The state of this modal can't be handled inside the QueryTitle, due to the react bootstrap tabs key bindings.
+ * E.g. using the arrow keys inside the title input would result in losing the focus.
+ */
```
kmerz: I am not a big fan of that kind of comments, since they tend to be obsolete and out of date quite soon.

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.jsx
```diff
@@ -0,0 +1,91 @@
...
+
+type Props = {
+  /** Function to update the query tab tite on save */
```
kmerz: Please remove this comment as well, the function name should (and is in this case) enough.

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.test.jsx
kmerz: Yes that is the best we have.

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.jsx
linuspahl: I don't think we are going to remove the react bootstrap tabs soon, are we?
But I get your pint and will adjust it! Would you remove the full comment? I thought it is nice to have this main component comment for the UI frontend docs.

linuspahl: I would suggest removing the last two lines.

linuspahl:
@kmerz Thanks for the review! I like your idea with the max length for the title input. 
If no one has a different opinion I would add a max length with 40 (!?) chars.
I can't really tell if there are some users with longer titles who would not see it in full length, but I think thats okay.

graylog2-web-interface/src/views/components/queries/QueryTitleEditModal.jsx
```diff
@@ -0,0 +1,88 @@
...
+          <Modal.Title>Editing query title</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <Input autoFocus
+                 help="The title of the query tab."
+                 id="title"
+                 label="Title"
+                 name="title"
+                 onChange={this._onDraftChange}
+                 required
```
linuspahl: It just came trough my mind, that this required does nothing, because we are not in the context of a form. I will implement the manual is empty check together with the inout maxlength change, once we are sure we want this restriction.

kmerz:
> @kmerz Thanks for the review! I like your idea with the max length for the title input.
> If no one has a different opinion I would add a max length with 40 (!?) chars.
> I can't really tell if there are some users with longer titles who would not see it in full length, but I think thats okay.

@linuspahl: I would suggest to just add a html5 validation. If a user has already a long title he will get an error when he edit it. But I would say that noone has long titles till now, since the Query Titles did not support spaces till now.

PR #6624, opened 2019-10-21
Generate unique aggregation ids for multiple percentile metrics

kmerz:
## Motivation and Context
Prior to this change, the user was unable to create more than one
percentile metric for same field in aggregation. Since the aggregation
series name was matching on aggregation name and field.

## Description
This change will add the percentile to the aggregation name. That way
the user can add more then one percentile aggergation.

## How Has This Been Tested?
Add more than one percentile series

## Types of changes

## Checklist:


alex-konn:
@kmerz und @dennisoelkers, the only reason I haven't merged this right away was that I wasn't 100% sure that this is as trivial a change as it looks like at first glance. Is the name of the SeriesSpec persisted?

PR #6636, opened 2019-10-22
Introduce a cluster global API browser button (#6567)

### LINKED ISSUE #5920: API browser behavior not as expected through load balancer

bernd:
Usually, the link to the API browser is meant to talk to one specific node only.

In cases where Graylog is served behind a load balancer, the node
specific URL is not reachable.
We autodetect that ( `http_external_uri` != `http_publish_uri` )
and display a "cluster global API browser button".

Since the API browser behind this address might end up talking
to random nodes of the cluster, display a warning box, to make
users aware of this.

Fixes #5920

Refs #2360
Refs https://github.com/Graylog2/graylog2-server/pull/2587

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/documentation/DocumentationBrowserResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/documentation/DocumentationBrowserResource.java
@@ -65,6 +65,7 @@ public class DocumentationBrowserResource extends RestResource {
                 .build();
     }
 
+    // Serve Swagger for a specific node, using HttpPublishUri
     @GET
     @Produces(MediaType.TEXT_HTML)
     @Path("index.html")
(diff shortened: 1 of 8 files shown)


bernd:
@mpfz0r The build fails due to some UI infrastructure changes in master. I will fix it.

PR #6637, opened 2019-10-23
Change url after loading a bookmarked search

kmerz:
## Motivation and Context
Prior this change, after the user loaded a bookmarked search
the id was not shown in the url. So the user could not create
a browser bookmark for that search.

## Description
This change adds browserHistory.push at the points where a bookmarked
search will be loaded.

## How Has This Been Tested?
- Add a new bookmark and checked for the ID in the url bar of the browser
- Loaded a old bookmark and checked for the url
- Cleared the view and checked that I got back to `/search`

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/pages/ShowViewPage.jsx
+++ b/graylog2-web-interface/src/views/pages/ShowViewPage.jsx
@@ -6,11 +6,13 @@ import connect from 'stores/connect';
 import Spinner from 'components/common/Spinner';
 
 import View from 'views/logic/views/View';
+import ViewLoaderContext from 'views/logic/ViewLoaderContext';
 import { SearchExecutionStateStore } from 'views/stores/SearchExecutionStateStore';
 import withPluginEntities from 'views/logic/withPluginEntities';
 import type { ViewHook } from 'views/logic/hooks/ViewHook';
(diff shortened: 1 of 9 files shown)


graylog2-web-interface/src/views/pages/ShowViewPage.jsx
```diff
@@ -82,6 +84,8 @@ class ShowViewPage extends React.Component<Props, State> {
     ).then((results) => {
       this.setState({ loaded: true });
       return results;
+    }).then(() => {
```
dennisoelkers:
```suggestion
    }).then(() => SearchActions.executeWithCurrentState())
```

This is not only syntactic sugar, it also leads to the result of `executeWithCurrentState` (a promise) becoming part of the promise chain, so the following `catch` is also applied to it.

linuspahl:
Apart from the pipeline, this looks good to me!

One thing (unimportant, but interesting, in my opinion). When I am on a bookmarked search and delete a different bookmarked search, I stay on the current page, but loose the relation for my current bookmark / can't update the current bookmark (icon is no longer blue). The "relation" is getting restored, after a page reload.

PR #6648, opened 2019-10-23
Contrasting Color Generator Utility 

kyleknighted:
## Description

- Robust Button styles
- Proper WCAG level AA contrast on text in `Button`s & `Alert`

## Motivation and Context

As we generate colors for our buttons, alerts, and other components from a single defined theme color, we need to make sure that the rendered color is properly contrasted against the background, but for aesthetics did not want to just render stark black or white colors.

## How Has This Been Tested?

Added test file in `util` directory to make sure the output stays consistent

Alert

Button

SplitButton & DropdownButton

## Types of changes

## Checklist:


linuspahl:
I really like these changes!
I understand the logic behind the button colors, it just looks like we need a way to deal with the following cases:

Before:
image
After:
image

graylog2-web-interface/src/components/graylog/SplitButton.jsx
```diff
@@ -1,21 +1,23 @@
...
   const StyledSplitButton = useCallback(styled(BootstrapSplitButton)`
-    ${buttonStyles({ active })};
+    ${buttonStyles(props)};
 
     ~ .btn.dropdown-toggle {
-      ${buttonStyles({ active })};
+      ${buttonStyles(props)};
     }
-  `, [active]);
+  `, [active, bsStyle, disabled]);
```
dennisoelkers: If we pass in `props` to the `buttonStyles` function used in the callback, imo we also need to use the whole `props` object as a dependency, because otherwise if the `buttonStyles` changes and we suddenly also use a e.g. `bsSize` prop from it, the callback might get stale if only `bsSize` changes.

PR #6666, opened 2019-10-25
Publish GrokPatternsUpdatedEvent on pattern import

bernd:
This ensures the reload of running extractors to use the imported grok
patterns.


graylog2-server/src/main/java/org/graylog2/grok/MongoDbGrokPatternService.java
```diff
@@ -186,6 +186,8 @@ public GrokPattern update(GrokPattern pattern) throws ValidationException {
             patternNames.add(savedGrokPattern.name());
         }
 
+        clusterBus.post(patternNames.build());
```
mpfz0r:
```suggestion
        clusterBus.post(GrokPatternsUpdatedEvent.create(patternNames.build()));
```

mpfz0r: Fixes the test

Issue #4637, opened 2018-03-05
Server message processing lock when there are multiple index optimization jobs


2fa:
## Expected Behavior
When you have multiple index optimization jobs run simultaneously they shouldn't block server message processing.

## Current Behavior
Multiple index optimization jobs simultaneously running cause message processing lock on the master node

## Steps to Reproduce
1. Create multiple index sets with enabled index optimization after rotation
2. Simultaneously rotate 2 or more of them

## Context
We have graylog cluster configuration with 3 nodes and multiple index sets, some of them large (~17GB, 7-8 million messages per index), some not.

Index optimization for a large index set one takes a long time (almost an hour) and if other indexes start cycle job (and therefore index optimization) it's causes master server to lock until first job completes.

We've disable index optimization after rotation for all but one index set and it 'solved' our problem.

Also, when this happens there's no struggle for CPU or disk utilization on the server and elasticsearch cluster (and there's no messages related to this in ES logs).

Below i added chunk of graylog master log:

>2018-03-02T02:35:10.435+03:00 INFO  [AbstractRotationStrategy] Deflector index <Default> (index set <graylog_1916>) should be rotated, Pointing deflector to new index now!
2018-03-02T02:35:10.440+03:00 INFO  [MongoIndexSet] Cycling from <graylog_1916> to <graylog_1917>.
2018-03-02T02:35:10.440+03:00 INFO  [MongoIndexSet] Creating target index <graylog_1917>.
2018-03-02T02:35:10.594+03:00 INFO  [Indices] Successfully created index template graylog-internal
2018-03-02T02:35:13.505+03:00 INFO  [MongoIndexSet] Waiting for allocation of index <graylog_1917>.
2018-03-02T02:35:13.514+03:00 INFO  [MongoIndexSet] Index <graylog_1917> has been successfully allocated.
2018-03-02T02:35:13.514+03:00 INFO  [MongoIndexSet] Pointing index alias <graylog_deflector> to new index <graylog_1917>.
2018-03-02T02:35:14.270+03:00 INFO  [SystemJobManager] Submitted SystemJob <30f2d7e1-1da9-11e8-8999-00259061c4c5> [org.graylog2.indexer.indices.jobs.SetIndexReadOnlyAndCalculateRangeJob]
2018-03-02T02:35:14.270+03:00 INFO  [MongoIndexSet] Successfully pointed index alias <graylog_deflector> to index <graylog_1917>.
[...]

PR #6675, opened 2019-10-28
Increase ES client HTTP connection limits

bernd:
This change is bumping the "elasticsearch_max_total_connections" and
"elasticsearch_max_total_connections_per_route" default settings to
avoid message processing stalls.

The default limit of 2 connections per ES server is way too low and can
cause delayed or event blocked message processing.

Example:

With a limit of 2 connections per ES server, 2 index optimization jobs
that are running in parallel will cause message processing to stall until
the optimization jobs are done.

Refs #4637


bernd:
This PR needs to be backported into the 3.1 branch.

PR #6678, opened 2019-10-29
Implement heatmap aggregation

linuspahl:
This PR adds the heatmap visualization to the aggregation builder visualizations.

To give you some background info for the related changes:
My main goal was to rely as much as possible on the existing ChartData.js logic.
The main problem: For e.g. the line chart, the `chartData` function generates one series for each column pivot. The heatmap plotly chart expects only one series with another property z, wich contains all the series data in a specific format:
```
  [
    [1, 3, 2],
    [0, 2, 3]
  ]
```
The main array contains one array for each y coordinate. These arrays contain one value for each x coordinate. 
A requirement, we defined for the aggregation builder: The field, defined for the row input should configure the y coordinates and the field defined for the column input, should define the x coordinates.

During the development I had different approaches, to deal with these requirements. Currently, the `extractSeries` function creates an object `valuesBySeries`. This variable has a key for each x coordinate with an array for every related y coordinate. I've implemented a `formatSeries` function, which receives the variable `valuesBySeries`. In the default case it proceeds with the data as needed for all other charts.
I've just allowed overwriting this format function, when calling `chartData` inside the visualization component. In the heatmap component we just flip the x and y labels and use the `valuesBySeries` data for our z property. The final trick is adding `transpose: true` to the series,  this flips the matrix over its diagonal. And we don't have to do this step manually (Thanks plotly)

Some improvements, I did not implemented yet:
- When there is no data for a field (z=null), the field will have no color / display the white background.
For e.g. the count metric, it would be nice to display them in the same color, like the smallest number. Just adding zeros, does not work, it would distort the color scale (for example when most values are around 1000, all fields would be red). That would only work in combination with the property `zmid`. In this case we would have to add the zeros and calculate the average value of z, to define it manually.
It is a pragmatic approach, but I think the easiest solution, would be adding a default background color, depending on the metric type. What do you think?
- When there is no data for a field (z=null) the tooltip shows NaN. I though about adjusting the z value in the first place, but that does not work. As well as adjusting the hovertemplate.

## How Has This Been Tested?
I've added a snapshot test for the HeatmapVisualization. I think it is the easiest way, to check if the format function does it's job and the GenericPlot receives the expected data.
I also added a test for the chartData function, to check if the optional formatFunction is getting called correctly.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -10,64 +10,81 @@ export type ChartDefinition = {
...
 const _defaultKeyJoiner = keys => keys.join('-');
 
-export const extractSeries = (keyJoiner: KeyJoiner = _defaultKeyJoiner) => {
-  return (results: Rows): ExtractedSeries => {
-    // $FlowFixMe: Somehow flow is unable to infer that the result consists only of Leafs.
-    const leafs: Array<Leaf> = results.filter(row => (row.source === 'leaf'));
+const _defaultChartGenerator = (type, name, labels, values): ChartDefinition => ({ type, name, x: labels, y: values });
 
-    const x = leafs.map(({ key }) => key);
+const _flattenLeafs = (leafs: Array<Leaf>, showSummarySeries: boolean = true) => {
```
linuspahl: When we currently create a chart with the metric count, e.g. for a linechart, we always display a "line" for a summary of all counts. This would make the heatmap useless. This is the best approach that came through. Any ideas of a different approach? Or a different code style?

graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -10,64 +10,81 @@ export type ChartDefinition = {
...
+  return flatten(leafs.map(l => l.values.filter(v => filterCondition(v)).map(v => [l.key, v])));
+};
 
-    const y = flatten(leafs.map(l => l.values.filter(v => v.source.endsWith('leaf')).map(v => [l.key, v])));
+export const formatSeries = ({ valuesBySeries, xLabels }: {valuesBySeries: Object, xLabels: Array<any>}): ExtractedSeries => {
+  return Object.keys(valuesBySeries).map(value => [
+    value,
+    xLabels,
+    valuesBySeries[value],
+    [],
```
linuspahl: I had to add a default value here, I tried to make the z value optional, in the generateChart parameter type definition, but got a flow error, that this (passing null, undefined) is not possible for a value inside an array.

graylog2-web-interface/src/views/components/visualizations/__tests__/ChartData.test.js
```diff
@@ -86,6 +86,28 @@ describe('Chart helper functions', () => {
...
     });
+    it('should allow passing a format series function to modify the series structure', () => {
+      const input = readFixture('ChartData.test.oneColumOneRowPivot.json');
+      const generatorFunction = (type, name, x, y, z) => ({ type, name, x, y, z });
+      const formatSeriesCustom = ({ valuesBySeries, xLabels }) => {
+        // In this example we want to create only one series, with an z value, which contains all series data
+        const z: Array<any> = Object.values(valuesBySeries).map((series) => {
+          const newSeries = fill(Array(xLabels.length), null);
+          return merge(newSeries, series);
+        });
```
linuspahl: By default the function would return nothing for missing values in the `valuesBySeries` arrays. This is not problem when working with plotly, but it's a problem when saving the result as JSON. If just added nulls as a workaround

linuspahl:
@dennisoelkers We talked about a way to identify the selected metric type (in the context of count() and sum()). I could not find the related code, I just found the metric type encapsulated in the config object. Could you show me an example of the current implementation?

graylog2-web-interface/src/views/components/visualizations/ChartData.js
dennisoelkers: I would replace it with a matcher instead. The default would look like `value => value.source.endsWith('leaf')` while you would override it in your use case with `value => (value.source === 'row-leaf')`. This makes it way more flexible and in the function's scope the abstraction level is not left (extracting leafs, vs. summary series).

linuspahl:
Thanks for the review!

@kmerz That is a good idea. We can implement this, by using plotlys `hovertemplate` + `customdata` props.

graylog2-web-interface/src/views/components/visualizations/ChartData.js
linuspahl: Thanks, I really like this approach. I've implemented the matcher function and added a test.

graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -10,64 +10,72 @@ export type ChartDefinition = {
...
 
-export const extractSeries = (keyJoiner: KeyJoiner = _defaultKeyJoiner) => {
-  return (results: Rows): ExtractedSeries => {
-    // $FlowFixMe: Somehow flow is unable to infer that the result consists only of Leafs.
-    const leafs: Array<Leaf> = results.filter(row => (row.source === 'leaf'));
+const _defaultChartGenerator = (type, name, labels, values): ChartDefinition => ({ type, name, x: labels, y: values });
 
-    const x = leafs.map(({ key }) => key);
+const _flattenLeafs = (leafs: Array<Leaf>, matcher: string => boolean = source => source.endsWith('leaf')) => {
+  return flatten(leafs.map(l => l.values.filter(value => matcher(value.source)).map(v => [l.key, v])));
```
dennisoelkers: Wouldn't it be better if we would pass in `value` instead of `value.source` to the `matcher` function? It would offer better flexibility (maybe we want to filter based on key or values?) and does not make calling significantly longer (the matcher would be specified as `({ source }) => ...` instead of `source => ...`.

graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -78,9 +86,15 @@ export const chartData = (
   data: Rows,
   chartType: string,
   generator: Generator = _defaultChartGenerator,
-): Array<ChartDefinition> => flow([
-  transformKeys(rowPivots, columnPivots),
-  extractSeries(series.length === 1 ? doNotSuffixTraceForSingleSeries : undefined),
-  removeNulls(),
-  generateChart(chartType, generator),
-])(data);
+  formatSeriesCustom?: ({valuesBySeries: Object, xLabels: Array<any>}) => ExtractedSeries,
```
dennisoelkers: Maybe `customSeriesFormatter` is a better name? We could also define a default value here, removing the need for the conditional assignment below.

linuspahl: Good catch with the default value, I change the function name as well.

graylog2-web-interface/src/views/components/visualizations/ChartData.js
linuspahl: Now I get your point regarding the flexibility. I did not saw the benefits of using the full value in the first place.

PR #6687, opened 2019-10-30
Filter searches for only permitted ones when GET /views/search is requested

alex-konn:
This PR makes sure that the search backend only returns permitted searches for the current user when `GET views/search` is requested. 
It also introduces a new class to serve as an entry point into the `search` package domain logic.

## Description
Added new class `SearchDomain` in top level `search` package. This class should serve as an entry point into search-related domain.
Moved filtering of searches for permitted ones from `SearchDbService` into this class, because it's domain logic and not directly data access-related.
Extracted search permissions from views into new class `ViewPermissions` to encapsulate the relaled logic and three necessary dependencies.
Added two new exceptions `EntityNotFoundException` and `PermissionException` to avoid reusing exceptions from `database` (`NotFoundException`) and `jersey` (another `NotFoundException`, `ForbiddenException` packages.
Last but not least implemented filtering of searches list in `SearchDomain` reusing the permission logic for loading a single search.

## Motivation and Context
- Add missing feature of filtering down the searches list to only permitted ones
- Establish clearer boundaries between domain logic, data access, rest api
- Improve testability by encapsulating in stateless classes. `SearchDomain` and `ViewPermission` are easier to test than `SearchDbService`, because the have not dependencies to MongoDB.

## How Has This Been Tested?
- Unit tests for new filtering of searches list
- Unit tests for already existing permission check when loading a single search
- Manual api requests with different users

## Types of changes


graylog2-server/src/main/java/org/graylog/plugins/views/search/ViewPermissions.java
```diff
@@ -0,0 +1,65 @@
...
+
+        if (views.isEmpty())
+            return false;
+
+        return hasSharedView(user, views) || hasDirectReadPermissionForAny(views, viewReadPermission);
+    }
+
+    private boolean hasSharedView(User user, Collection<ViewDTO> views) {
+        return views.stream()
+                .anyMatch(view -> isSharedWith(view, user));
```
dennisoelkers: This results in n+1 queries, maybe it makes sense to introduce a `ViewSharingService#forViews` that takes a set of views and returns a map keyed with the view (id) that could be iterated over here?

PR #6688, opened 2019-10-30
Hide root user when the RootAccountRealm is deactivated

thll:
## Description
The root user will not show up in the UI or in the API anymore if the "root-user" authentication provider is disabled.

## Motivation and Context

Even when "root-user" was listed under `deactivated_builtin_authentication_providers` in the configuration file, the root user would still be shown on the frontend and used at some places in the code. Having the root-user linger even though it can't really be used is confusing.

The root user will now be hidden in the frontend and it has been made clearer in the API that the root user might not exist (by making the return value optional).

Setting the root user's password in the config is not required anymore when the user is deactivated.

## How Has This Been Tested?

- Added "root-user" to the `deactivated_builtin_authentication_providers` config option. Opened the UI and verified that the local admin user wasn't visible anymore.
- Performed API request to `/api/users` to verify that the `local:admin` user was not present anymore
- Started Graylog server with an empty MongoDB to verify that the touched migration didn't cause any trouble

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/migrations/V20161125142400_EmailAlarmCallbackMigration.java
```diff
@@ -117,6 +117,11 @@ public void upgrade() {
     }
 
     private Optional<String> migrateStream(Stream stream) {
+        if (localAdminUser == null) {
+            throw new IllegalStateException("The root user is disabled. This migration will run only with the root " +
```
bernd: This will throw an exception on every startup of the server. I think we can default to `"admin"` if the root user is disabled. We only pass the user ID (the name) to the alarm callback config and that alarm callback code is deprecated and will go away in the future.

thll: If I didn't get it wrong, it wouldn't throw on every startup but only if:
- the migration has never completed successfully before
- **and** there actually are streams which need migration
- **and** the root user is disabled

I wouldn't expect many installations getting into this situation so I thought it was tolerable to throw an exception in that case.
But looking into it again following your suggestion, the ID of the local admin user is fixed to "local:admin" regardless of the username configured in the config. So as the alarm callback code is going away anyways I'm now simply using this fixed string for the migration. Regardless of the activation/deactivation state of the root user.

bernd: Thanks! :+1:

PR #6707, opened 2019-11-01
Pluggable login form

edmundoa:
Extracts login form into its own component, independent from `LoginPage`, and lets one plugin to display an alternative login page.

I decided not to migrate `LoginPage` to a functional component to not pollute the PR with that, but I can also make the change if you consider it appropriate.

---
Update to include how to use a plugin to log into Graylog. I don't see a good place in the code to add this information, so I will leave it here for now.

The plugin needs to register a form component in `loginProviderType` taking care of displaying a form, or button, the user can use to log into Graylog. Only one plugin can do that at the time, so far we don't need to support any number of custom login forms.

The form may talk to an authenticator provided by the plugin API and/or an external system to handle the login. Once the user is authenticated, we need to ensure the session information is stored in the browser, like we do in the regular log in. This can be accomplished by calling `SessionActions.login.completed()` with the session ID and usernames returned from the server:

```
const session = { sessionId: session_id, username: username };
SessionActions.login.completed(session);
```

That will also take care of propagating the changes to other stores and components listening to it.

### DIFF
--- a/graylog2-web-interface/src/components/login/LoginForm.jsx
+++ b/graylog2-web-interface/src/components/login/LoginForm.jsx
@@ -1,0 +1,64 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { Button, FormGroup } from 'components/graylog';
+
+import { Input } from 'components/bootstrap';
+import CombinedProvider from 'injection/CombinedProvider';
+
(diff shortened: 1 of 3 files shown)


edmundoa:
@dennisoelkers I updated the original comment to document the process, let me know if this still needs more information.

PR #6711, opened 2019-11-03
Make messages with identical timestamps sortable by ULID

### LINKED ISSUE #2741: Overcoming log ordering issues with millisecond precision
Currently, if logs are timestamped the same (to the millisecond), we can get a somewhat random ordering when searching Graylog. This is fundamentally an issue with Elasticsearch not providing enough precision, only going as close to a millisecond (https://github.com/elastic/elasticsearch/issues/10005). In this issue it is suggested that if messages are timestamped the same, ordering will be by oldest-first or first-written. This is a huge problem if your input doesn't guarantee order, such as Kafka. It can still be an issue though even if you use a reliable ordering input when you start to cluster.

If Graylog is supposed to be a log management system, it needs to have a strategy for dealing with this issue. This strategy could be functionality of Graylog, or it could be a documented external solution imposed on users. 
## Possible Solution

My team has come up with a couple possible solutions:
- **Solution 1: Save a secondary timestamp in nano's**

By having a secondary "timestamp" field in Graylog, we could use it for a finer way to search. The issue becomes relying on frameworks that provide API's for nanosecond timestamps, and also the issue of indexing these huge numbers (increasing document size and search time).
- **Solution 2: Save a log sequence count in the logs**

The nature of our logging requires we have a request ID field attached to each log message which correlates all messages which have been created as a result of an original user request. We could easily store a sequence count which increments with each logged message on that request. The issue then becomes that Graylog only allows you to sort by **one** field at a time. In order for this to work we need to be able to search by timestamp, then sequence count.
## Context

This is really a big issue for developers trying to solve issues, especially in a microservice environment where they might not be familiar with the codebase of another service, log ordering is crucial.

This isn't just an issue for distributed systems though. (As mentioned earlier) Even if you have a monolith application which uses a reliable order input such as TCP GELF, as soon as you cluster Graylog, log messages aren't guaranteed to be written in order to Elasticsearch. 
## Your Environment
- Graylog Version: 2.0.3
- Elasticsearch Version: 2.3.2
- MongoDB Version: 3.2.9

mpfz0r:
The ULID format in the `gl2_message_id` field is composed of a 48 bit timestamp followed by 80 bits
of randomness.
Use the first 16 bits of the random field to embed a sequence number
for each message.
If a batch of messages was received with identical timestamps
(the same millisecond), the original receive order is kept by the
encoded sequence number which directly follows the timestamp.

This allows us to sort messages by `gl2_message_id` which should have the correct original
order in most cases.

## CAVEATS
This is a best effort approach to a complicated problem. It's not a silver bullet.
Here are reasons why the `gl2_message_id` sort order might not always be correct:

- The sequence number is generated per node and input. 
  This means that sorting will not work if an input is load balanced over multiple nodes.

 - There is only space for `64535` messages with the same timestamp and input.

- Also, there is a small chance that, if too many batches of messages with the same timestamp and input get processed
  in parallel, the sort order might be wrong. 

## Performance Impact

Running a benchmark, which ingests 8 million messages.
Four parallel curl loops send batches of `5000` messages with identical timestamps.

Without this change, this takes `3m 19s`
With this change: `3m 25s` 

Which means approx 40k msg/sec and no measurable performance impact.

Fixes #2741


graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+    public MessageULIDGenerator(ULID ulid, Journal journal) {
+        this.ulid = ulid;
+
+        sequenceNrCache = Caffeine.newBuilder()
+                .maximumSize(100)
+                .initialCapacity(100)
+                .build();
+    }
+
+    public Message addULID(Message message) {
```
bernd: What do you think about making the API of this class a bit more testable? Instead of modifying the Message object, I think it's nicer to just return the generated ID, too.

```diff
diff --git graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
index 9abc3898e..0fade5802 100644
--- graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
+++ graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
@@ -18,9 +18,9 @@ package org.graylog2.shared.buffers.processors;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
+import com.google.common.annotations.VisibleForTesting;
 import de.huxhorn.sulky.ulid.ULID;
 import org.graylog2.plugin.Message;
-import org.graylog2.shared.journal.Journal;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,7 +37,7 @@ public class MessageULIDGenerator {
     private static final int OFFSET_GAP = 500;
 
     @Inject
-    public MessageULIDGenerator(ULID ulid, Journal journal) {
+    public MessageULIDGenerator(ULID ulid) {
         this.ulid = ulid;
 
         sequenceNrCache = Caffeine.newBuilder()
@@ -46,7 +46,12 @@ public class MessageULIDGenerator {
                 .build();
     }
 
-    public Message addULID(Message message) {
+    public String createULID(Message message) {
+        return createULID(message.getSourceInputId(), message.getTimestamp().getMillis(), message.getSequenceNr());
+    }
+
+    @VisibleForTesting
+    String createULID(String inputId, long timestamp, int sequenceNr) {
         // Fill the first 16 bits of the ULIDs random section with a sequence number that reflects the order
         // in which messages were received by an input.
         //
@@ -60,13 +65,10 @@ public class MessageULIDGenerator {
         // This means that the first recorded sequence number can be higher than the one of later messages.
         // To account for this, we simply add a constant (OFFSET_GAP) to prevent negative messageSequenceNrs.
 
-        final long messageTimestamp = message.getTimestamp().getMillis();
-        final int sequenceNr = message.getSequenceNr();
-
-        final String key = message.getSourceInputId() + messageTimestamp;
+        final String key = inputId + timestamp;
         final Integer subtrahend = sequenceNrCache.get(key, k -> sequenceNr);
 
-        final ULID.Value nextUlid = ulid.nextValue(messageTimestamp);
+        final ULID.Value nextUlid = ulid.nextValue(timestamp);
         final long mostSignificantBits = nextUlid.getMostSignificantBits();
         final long leastSignificantBits = nextUlid.getLeastSignificantBits();
 
@@ -77,7 +79,6 @@ public class MessageULIDGenerator {
             messageSequenceNr %= RANDOM_MSB_MASK;
         }
         final ULID.Value sequencedUlid = new ULID.Value(msbWithZeroedRandom + messageSequenceNr, leastSignificantBits);
-        message.addField(Message.FIELD_GL2_MESSAGE_ID, sequencedUlid.toString());
-        return message;
+        return sequencedUlid.toString();
     }
 }
diff --git graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
index 9190d0074..27b589612 100644
--- graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
+++ graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
@@ -135,7 +135,7 @@ public class ProcessBufferProcessor implements WorkHandler<MessageEvent> {
             // Don't overwrite already existing message IDs.
             // They might already be set if we are receiving messages from a Graylog forwarder
             if (message.getField(Message.FIELD_GL2_MESSAGE_ID) == null) {
-                messageULIDGenerator.addULID(message);
+                message.addField(Message.FIELD_GL2_MESSAGE_ID, messageULIDGenerator.createULID(message));
             }
 
             // The processing time should only be set once all message processors have finished
diff --git graylog2-server/src/test/java/org/graylog2/shared/buffers/processors/MessageULIDGeneratorTest.java graylog2-server/src/test/java/org/graylog2/shared/buffers/processors/MessageULIDGeneratorTest.java
new file mode 100644
index 000000000..eaa8efc77
--- /dev/null
+++ graylog2-server/src/test/java/org/graylog2/shared/buffers/processors/MessageULIDGeneratorTest.java
@@ -0,0 +1,18 @@
+package org.graylog2.shared.buffers.processors;
+
+import de.huxhorn.sulky.ulid.ULID;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class MessageULIDGeneratorTest {
+    @Test
+    public void generate() {
+        final MessageULIDGenerator generator = new MessageULIDGenerator(new ULID());
+
+        assertThat(generator.createULID("input1", 0, 0)).startsWith("00000000000001");
+        assertThat(generator.createULID("input1", 0, 65536)).startsWith("00000000000001");
+
+        // TODO: Implement more tests :)
+    }
+}

```

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+public class MessageULIDGenerator {
+    private static final Logger LOG = LoggerFactory.getLogger(MessageULIDGenerator.class);
+
+    private final Cache<String, Integer> sequenceNrCache;
+    private final ULID ulid;
+    private static final long RANDOM_MSB_MASK = 0xFFFFL;
+    private static final int OFFSET_GAP = 500;
+
+    @Inject
+    public MessageULIDGenerator(ULID ulid, Journal journal) {
```
bernd: Unused `journal` argument.

graylog2-server/src/main/resources/org/graylog2/plugin/journal/raw_message.proto
```diff
@@ -25,6 +27,8 @@ message JournalMessage {
     optional RemoteAddress remote = 7;
 
     optional bytes payload = 8;
+
+    optional uint32 sequence_nr = 9;
```
bernd: This means the sequence number is an integer. Wouldn't it be safer to use a long? What happens if the atomic integer wraps to `Integer.MIN_VALUE` on `.incrementAndGet()`? I guess we don't want negative sequence numbers.

graylog2-server/src/main/resources/org/graylog2/plugin/journal/raw_message.proto
```diff
@@ -1,8 +1,10 @@
+syntax = "proto2";
+
 package org.graylog2.plugin.journal;
 
-// to regenerate:
-// install protobuf compiler (this was generated with 2.5.0)
```
bernd: Let's please generate the `JournalMessages` with protobuf compiler version 2.5.0 to avoid the big diff.

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
```diff
@@ -0,0 +1,83 @@
...
+        //
+        // Since our message processing is multi threaded, messages can pass each other during processing.
+        // This means that the first recorded sequence number can be higher than the one of later messages.
+        // To account for this, we simply add a constant (OFFSET_GAP) to prevent negative messageSequenceNrs.
+
+        final long messageTimestamp = message.getTimestamp().getMillis();
+        final int sequenceNr = message.getSequenceNr();
+
+        final String key = message.getSourceInputId() + messageTimestamp;
+        final Integer subtrahend = sequenceNrCache.get(key, k -> sequenceNr);
```
bernd: Do we need to handle `.get()` errors? IntelliJ shows a warning in line 74 for the `subtrahend` variable.("Unboxing of 'subtrahend' may produce 'NullPointerException' ") The method can also throw exceptions.

graylog2-server/src/main/java/org/graylog2/plugin/inputs/MessageInput.java
```diff
@@ -345,6 +348,7 @@ public void processRawMessage(RawMessage rawMessage) {
         rawMessage.setCodecName(codec.getName());
         rawMessage.setCodecConfig(codecConfig);
         rawMessage.addSourceNode(getId(), serverStatus.getNodeId());
+        rawMessage.setSequenceNr(sequenceNr.incrementAndGet());
```
bernd: Using `.getAndIncrement()` would give us one more sequence number because we would start at `0`. :wink:

bernd (dismissed):
A few comments:

graylog2-server/src/main/resources/org/graylog2/plugin/journal/raw_message.proto
mpfz0r: Good catch. I didn't' use long on purpose to save some space in the journal.
I've added code to avoid wrapping into negative values.

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
mpfz0r: I don't see how this could ever happen, but I've added some precaution to fallback on a a timestamp only ULID in case we hit one.

bernd:
Checking for an existing "gl2_message_id" and using the message timestamp is done in https://github.com/Graylog2/graylog2-server/pull/7290.

bernd:
@mpfz0r Does this implementation still work when a non-local message journal implementation (e.g., Kafka) is used? This just crossed my mind.

mpfz0r:
> @mpfz0r Does this implementation still work when a non-local message journal implementation (e.g., Kafka) is used? This just crossed my mind.

It does. I had the same thought at first, but only my first iteration depended on the journal :sweat_smile:

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
```diff
@@ -0,0 +1,96 @@
...
+        // setting the messageSequenceNr accordingly.
+        if (messageSequenceNr < 0) {
+            LOG.warn("Message sequence number wrapped ({} -> {}). Sort order might be wrong.", subtrahend, sequenceNr);
+            messageSequenceNr = OFFSET_GAP;
+            sequenceNrCache.put(key, sequenceNr);
+        } else if (messageSequenceNr >= RANDOM_MSB_MASK) {
+            LOG.warn("Message sequence number does not fit into ULID ({} >= 65535). Sort order might be wrong.", messageSequenceNr);
+            messageSequenceNr %= RANDOM_MSB_MASK;
+        }
+        final ULID.Value sequencedUlid = new ULID.Value(msbWithZeroedRandom + messageSequenceNr, leastSignificantBits);
```
patrickmann, line 93: Since we're operating at the byte level I think this is easier to read and a tiny bit faster too:
```suggestion
        final ULID.Value sequencedUlid = new ULID.Value(msbWithZeroedRandom | messageSequenceNr, leastSignificantBits);
```

patrickmann (approved):
LGTM
Tested with syslog input - sorting by `gl2_message_id` works as expected

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
mpfz0r: makes sense :+1:

thll (changes requested):
Very nice improvement! 👍

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/MessageULIDGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+        // the messageSequenceNr can become negative.
+        // This can also happen if the sequenceNr counter in a MessageInput wraps.
+        // We handle this by updating the sequenceNrCache and setting the messageSequenceNr accordingly.
+        if (messageSequenceNr < 0) {
+            LOG.warn("Got negative message sequence number ({} -> {}). Sort order might be wrong.", subtrahend, sequenceNr);
+            messageSequenceNr = REORDERING_GAP;
+            sequenceNrCache.put(cacheKey, sequenceNr);
+        // If we receive more than 60535 messages with the same timestamp and input, they will exhaust the 16 bit of space in the ULID.
+        } else if (messageSequenceNr >= ULID_RANDOM_MSB_MASK) {
+            LOG.warn("Message sequence number <{}> input <{}> timestamp <{}> does not fit into ULID ({} >= 65535). Sort order might be wrong.",
```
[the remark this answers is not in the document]
mpfz0r: Yeah, we can do that. Reminds me that I wanted to test 
https://github.com/Swrve/rate-limited-logger
at some point.

thll (approved):

Issue #6696, opened 2019-10-30
Pipeline Rule allows invocation of function "set_field()" without setting a value for the new field.


willman42:
The documentation for Pipeline Rule function "set_field()" states that the second argument "value" is a required parameter for the function. However, Graylog allows me to invoke the function without specifying that argument.

## Expected Behavior
When writing the Pipeline Rule, an error message should appear and Graylog should forbid the user from saving the Rule until the issue is resolved, as is the behavior for other functions if not invoked correctly.

## Current Behavior
I am able to save the Pipeline Rule without supplying an argument for the "value" parameter. However the Rule does not appear to work, as the new field that is supposed to be created does not show up in logs.

## Possible Solution
Force the user to fix the error in the Pipeline Rule by supplying an argument for the "value" parameter in the "set_field()" function.

## Steps to Reproduce (for bugs)
1. Create Pipeline Rule like the following example:
```
rule "test set_field()"
when
   true
then
   set_field("foo");
end
```
2. Save the Rule, connect it to a Stream receiving logs.
3. Go to Stream, generate new logs.
4. Note that new logs do not have the field "foo" at all.

## Context
Since Pipeline Rules have no "elif" or "else" clauses, this makes setting binary fields useless, as instead you can just test for the existence of the field itself. Take the a modified version of the above Pipeline Rule for example:
```
rule "If true, create field foo and set to true"
when
   // List of conditions:
   $message.condition1 == "x"
then
   set_field("foo", true)
end
```

If want to avoid acting upon all logs that don't match the supplied condition, there's no reason for me to write a separate Pipeline Rule to set field "foo" to false, as it is implied by not matching against the above Rule.

Therefore, why must I set "foo" to any value at all? If the supplied conditions match, "foo" will equal Boolean True, and if conditions don't match, the field won't be set at all. So it would be more efficient to simply test for the existence of the field "foo", as its mere existence implies its value equal to Boolean True.

Honestly, instead of forcing the user to set a value for the new field in the set_field() function, why not make it officially optional? Until Pipeline Rules can extend beyond strictly "if/then" scenarios and include "elif" and "else" clauses, there is no reason for setting Boolean values at all, as demonstrated above.

In addition, currently the only way to test the existence of a field is by employing the "is_null()" and "is_not_null()" functions, which is rather indirect. Could a new function be written strictly for checking the existence of a field, like the "_exists_" search term in Graylog Search?

## Your Environment

* Graylog Version: 3.0.2
[...]

PR #6716, opened 2019-11-04
Parse error for missing required positional parameters with optional parameters present

kroepke:
## Description
Check for missing required parameters when being passed as positional parameters.

## Motivation and Context
The original check missed a branch when there were optional parameters following the required ones.
The rule never executed because it would fail during runtime, but the parser missed this and allowed creating the rule.

Fix issue #6696 

## How Has This Been Tested?
Added unit test for the scenario.

n/a

## Types of changes

## Checklist:


kroepke:
Good point! Will fix later today.

Thanks!

PR #6718, opened 2019-11-04
Heatmap visualization hovertemplate

linuspahl:
This PR contains two changes regarding the heatmap:

1. We've changed the colors for the color scale and implemented a default background color. @dennisoelkers and I already talked about, when to display a default color, for undefined values depending on the metric type. Currently it looks like it makes sense for every metric type. What do you think?

2. Konrad also mentioned, it would be great to see the title of the y and x axis in the heatmap info box, which appears when hovering a field. 

Plotly provides a property hovertemplate (template string), but:
- it is hard to read custom data like the axis title inside the string.
Setting and reading the axis title works, but would display the titles.
- when we use a hovertemplate and z is not a number, we receive a console error, every time we hover over one of these fields.

A possible solution:
- With this PR I am just using a js template string to display the x and y axis and also the name of series.
- plotly provides also a customdata property. It needs the same structure like z and allows to display a different value for each z value. This way, we can change the text for the z related `NaN` message in the info box as well. It is the only way we can get rid of this message. It is also the only way we can use the hovertemplate without throwing errors, if z is not a number.

"Disadvantages":
- we can't use plotly's transpose function and need to do it manually, because we need the matrix in the correct format when using the customdata attribute.
- we need to do an extra round to create an info like `None` for every undefined z value.
- we need to pass the config to the generator function.

image

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/visualizations/heatmap/HeatmapVisualization.jsx
```diff
@@ -10,10 +11,66 @@ import type { ChartDefinition, ExtractedSeries } from '../ChartData';
...
+  [0.85, '#95d840'],
+  [0.90, '#b8de29'],
+  [0.95, '#dce319'],
+  [1.00, '#fde725'],
+];
+
+const _generateSeries = (type, name, x, y, z, idx, total, config): ChartDefinition => {
+  const xAxisTitle = get(config, ['_value', 'rowPivots', idx, 'field']);
+  const yAxisTitle = get(config, ['_value', 'columnPivots', idx, 'field']);
+  const seriesTitle = get(config, ['_value', 'series', idx, '_value', 'function']);
```
linuspahl: One thing, that bothers me, what should we display, when multiple pivots / series are defined. Something like: `dayOfTheWeek / weekday: 2` would be too huge for the info box. Regarding the series, we can't display the correct metric, when multiple are defined. Should we may just show `value:` instead?

kmerz:
I am not read the code yet, simply testing it, but here is a small problem I stumbled over:
Graylog (26)
 project: shows dayOfTheWeek value and vice versa.

linuspahl:
@kmerz Thanks for your feedback!! That just slipped in again.. I've made the necessary changes and resolved the conflicts.

linuspahl:
@kmerz Because we talked about the series specific hover template labels: While testing the discussed changes, I've realised, the are scenarios, where we don't have this static series order. It's an abstract example, but something like:
image

I had to create the array for the series titles in a dynamic way, does this implementation suit you?

PR #6719, opened 2019-11-04
Themed MenuItem component

kyleknighted:
## Description

- Themed `MenuItem`
- Added `NavDropdown` to wrap navigation `MenuItem`s

image
image
image

## Types of changes

## Checklist:


graylog2-web-interface/src/components/graylog/DropdownButton.jsx
```diff
@@ -1,24 +1,15 @@
...
+import menuItemStyles from './styles/menuItem';
 import buttonStyles from './styles/button';
-import { propTypes, defaultProps } from './props/button';
 
-const DropdownButton = forwardRef((props, ref) => {
-  const { bsStyle, active, disabled } = props;
-  const StyledDropdownButton = useCallback(styled(BootstrapDropdownButton)`
-    ${buttonStyles(props)};
-  `, [bsStyle, active, disabled]);
+const DropdownButton = React.memo(styled(BootstrapDropdownButton)`
```
linuspahl: You are using React.memo here to avoid unnecessary rendering of the Bootstrap component, right? Out of interest, how do you decide, when to use React.memo?

graylog2-web-interface/src/components/graylog/styles/menuItem.js
```diff
@@ -0,0 +1,44 @@
...
+
+      :hover {
+        color: ${contrastingColor(teinte.secondary.due, 'AA')};
+      }
+    }
+  }
+`;
+};
+
+export default menuItemStyles;
```
linuspahl: What is your strategy for the styles defined in `components/graylog/styles/`? I always try to define the component related styles alongside the component. You are only extracting the color specific styles?

kyleknighted: It is being used on multiple components, so instead of repeating the code I am abstracting and importing it into each other place using the same styles

graylog2-web-interface/src/components/graylog/DropdownButton.jsx
kyleknighted: It was added in concern of too having to generate styles and slowing down the render.
Removed it until we notice places of concern

PR #6733, opened 2019-11-06
Copy widget from search to dashboard

kmerz:
##  Description
Add new action to widget drop down so a widget can be send from Search to Dashboard.
The time-range, query and streams will be taken from the top level of the search and
stored to the dashboard.

## Motivation and Context
The user might use the Search as a workbench for widgets. After finishing up a Widget
he wants to send it to a dashboard.

## How Has This Been Tested?
Add a new widget in the Search and copied to a Dashboard.

Graylog (25)

## Types of changes

## Checklist:


graylog2-web-interface/src/views/logic/views/UpdateSearchForWidgets.js
```diff
@@ -0,0 +1,29 @@
+// @flow strict
+import { get } from 'lodash';
+import View from 'views/logic/views/View';
+import SearchTypesGenerator from 'views/logic/searchtypes/SearchTypesGenerator';
+
+
+const UpdateSearchForWidgets = (view: View): View => {
+  const states = view.state;
```
kyleknighted: Should this be destructed? 
```suggestion
  const { state: states } = view;
```

And if we have to change the name of the constant, should we just update the name of the key in the object to be plural as well?

graylog2-web-interface/src/views/pages/ShowViewPage.jsx
```diff
@@ -63,6 +72,7 @@ class ShowViewPage extends React.Component<Props, State> {
 
   loadView = (viewId: string): Promise<?View> => {
     const { location, loadingViewHooks, executingViewHooks, viewLoader } = this.props;
+    // eslint-disable-next-line react/prop-types
```
kyleknighted: Instead of disabling eslint, could this destruct be part of line 74 destruct?

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+type Props = {
+  onCancel: () => void,
+  onSubmit: (string, string) => void,
+  widgetId: string,
+  dashboards: DashboardsStoreState,
+};
+
+const CopyToDashboardForm = (props: Props) => {
+  const { widgetId, onCancel, dashboards: { list, pagination }, onSubmit } = props;
```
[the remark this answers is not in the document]
kmerz: Ah yes. that was a artifact from debugging! Thanks!

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const [selectedDashboard, setSelectedDashbaord] = useState('');
+  const [paginationState, setPaginationState] = useState({ query: '', page: 1, perPage: 5 });
+
+  const execSearch = useCallback((query, page, perPage) => {
+    DashboardsActions.search(query, page, perPage);
+  });
```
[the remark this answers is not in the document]
kmerz: Yeah. Probably. Need to check again. But I think you are right!

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,90 @@
...
+      query,
+      ...paginationState,
+    });
+    setSelectedDashbaord('');
+    execSearch(query, paginationState.page, paginationState.perPage);
+  }, []);
+
+  const handleSearchReset = useCallback(() => handleSearch(''), []);
+
+  const handlePageChange = useCallback((page: number, perPage: number) => {
```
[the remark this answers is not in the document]
kmerz: No.. and yes. From what I learnt from my workshop. Is that if it is used as a callback, then it is most certainly wiser to do so.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,90 @@
...
+      page,
+      perPage,
+      ...paginationState,
+    });
+    setSelectedDashbaord('');
+    execSearch(paginationState.query, page, perPage);
+  }, []);
+
+  useEffect(() => {
+    execSearch(paginationState.query, paginationState.page, paginationState.perPage);
```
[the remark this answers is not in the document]
kmerz: Yeah. That could make sense. If a long running requests will return to a unmounted widget.. I will investigate that!

graylog2-web-interface/src/views/pages/ShowViewPage.jsx
kmerz: I will look at that!

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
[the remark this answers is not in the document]
kmerz: No we don't need to since it is triggering an action which does not alter our state.

graylog2-web-interface/src/views/logic/views/UpdateSearchForWidgets.js
kmerz: The reason why it is called states here, is to clarify that it is something to iterate over. But to have it as state is also valid. So no I would not rename it in the object.

kyleknighted: I agree with plural being a an easy identifier that it is iterable, but I just don't understand why we are renaming the variable in the middle.
If `view.state` is going to be defined as an object or array, then I don't see why we just don't call it `view.states` from the get-go instead of renaming it.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,89 @@
...
+  }, []);
+
+  const dashboardList = (list || []).map((dashboard) => {
```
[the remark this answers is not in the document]
kmerz: I tried that before but did not understand the flow error. Now I did :) And fixed it!

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,89 @@
...
+      perPage,
+      ...paginationState,
+    });
+    setSelectedDashbaord('');
+    execSearch(paginationState.query, page, perPage);
+  }, []);
+
+  useEffect(() => {
+    execSearch(paginationState.query, paginationState.page, paginationState.perPage);
+  }, []);
```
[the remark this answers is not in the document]
kmerz: I thought that a good idea, but it seems it does not work that way. I could not figure out completely why, but when I add `paginationState` here and remove `execSearch` from where `setPaginationState` is called the search will not executed. If I leave `execSearch` it will be called 2 times. One with the new state, and one with the old state. So no.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const handleSearch = useCallback((query) => {
+    setPaginationState({
+      query,
+      ...paginationState,
+    });
+    setSelectedDashbaord('');
+    execSearch(query, paginationState.page, paginationState.perPage);
+  }, []);
```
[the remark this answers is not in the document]
kmerz: Fixed Typo :+1:
Add the dependencies but couldn't see any difference to before.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,89 @@
...
+  onSubmit: (string, string) => void,
+  widgetId: string,
+  dashboards: DashboardsStoreState,
+};
+
+const CopyToDashboardForm = ({ widgetId, onCancel, dashboards: { list, pagination }, onSubmit }: Props) => {
+  const [selectedDashboard, setSelectedDashbaord] = useState('');
+  const [paginationState, setPaginationState] = useState({ query: '', page: 1, perPage: 5 });
+
+  const execSearch = useCallback((query, page, perPage) => {
```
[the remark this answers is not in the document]
kmerz: True! :+1:

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,85 @@
...
+  dashboards: DashboardsStoreState,
+};
+
+const CopyToDashboardForm = ({ widgetId, onCancel, dashboards: { list = [], pagination }, onSubmit }: Props) => {
+  const [selectedDashboard, setSelectedDashboard] = useState<?string>(null);
+  const [paginationState, setPaginationState] = useState({ query: '', page: 1, perPage: 5 });
+
+  const handleSearch = useCallback((query) => {
+    setPaginationState({
+      query,
```
dennisoelkers: This order leads to `paginationState.query` always overwriting `query`.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,85 @@
...
+    });
+    setSelectedDashboard(null);
+    DashboardsActions.search(query, paginationState.page, paginationState.perPage);
+  }, [paginationState]);
+
+  const handleSearchReset = useCallback(() => handleSearch(''), []);
+
+  const handlePageChange = useCallback((page: number, perPage: number) => {
+    setPaginationState({
+      page,
```
dennisoelkers: This order leads to `paginationState` always overwriting `page`/`perPage`.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,85 @@
...
+  const handleSearch = useCallback((query) => {
+    setPaginationState({
+      query,
+      ...paginationState,
+    });
+    setSelectedDashboard(null);
+    DashboardsActions.search(query, paginationState.page, paginationState.perPage);
+  }, [paginationState]);
+
+  const handleSearchReset = useCallback(() => handleSearch(''), []);
```
dennisoelkers: This needs `handleSearch` as dependency as well.

graylog2-web-interface/src/views/components/widgets/CopyToDashboardForm.jsx
```diff
@@ -0,0 +1,85 @@
...
+  const [paginationState, setPaginationState] = useState({ query: '', page: 1, perPage: 5 });
+
+  const handleSearch = useCallback((query) => {
+    setPaginationState({
+      query,
+      ...paginationState,
+    });
+    setSelectedDashboard(null);
+    DashboardsActions.search(query, paginationState.page, paginationState.perPage);
+  }, [paginationState]);
```
dennisoelkers: We should make these consistent with the dependencies of `handlePageChange`, I think.

graylog2-web-interface/src/views/components/widgets/Widget.jsx
```diff
@@ -49,11 +60,21 @@ type State = {
   configChanged?: boolean,
   editing: boolean,
   oldConfig?: WidgetConfig,
+  showCopyToDashboard: boolean,
 };
 
 class Widget extends React.Component<Props, State> {
+  static _visualizationForType(type) {
```
dennisoelkers: This is unrelated to the PR, but while you are touching it: These can be moved out of the class and be just plain functions defined in the top scope of this file.

graylog2-web-interface/src/views/components/widgets/Widget.test.jsx
```diff
@@ -106,55 +170,100 @@ describe('<Widget />', () => {
...
+      fireEvent.click(copyToDashboard);
+      const view1ListItem = getByText('view 1');
+      fireEvent.click(view1ListItem);
+      const selectBtn = getByText('Select');
+      fireEvent.click(selectBtn);
+    };
+
+    it('should get dashboard from backend', async () => {
+      renderAndClick();
+      await wait(() => expect(ViewManagementActions.get).toHaveBeenCalledTimes(1));
```
dennisoelkers: We should also check the parameters of the action call here and the following tests.

PR #6737, opened 2019-11-06
Show circle markers on worldmap, independently of a defined metric

linuspahl:
When a user selects the worldmap visualization and defines a pivot, but no metric, the backend already returns a list of locations, but we are not displaying them on the map. They are currently only visible, if a metric is defined. With the following changes, we show the available location, without any defined metric.

## Tests
I've added two tests, to check if the new `_createSeriesWithoutMetric` function and the existing `const series =` part in the worldmap component behave correctly. It would be great to know if these tests are appropriate. 

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/visualizations/worldmap/MapVisualization.jsx
```diff
@@ -94,8 +96,14 @@ const MapVisualization = createReactClass({
...
-            <dt>Value:</dt>
-            <dd>{value}</dd>
+            {value
+              && (
+              <React.Fragment>
+                <dt>Value:</dt>
+                <dd>{value}</dd>
+              </React.Fragment>
+              )
+            }
```
linuspahl: Is it fine to display the value conditionally or is there any case where we want to display a value of 0?

dennisoelkers: No, that is fine. We actually have no value here.

graylog2-web-interface/src/views/components/visualizations/worldmap/WorldMapVisualization.jsx
```diff
@@ -1,37 +1,48 @@
...
+  const leafs = getLeafsFromRows(rows);
+  const xLabels = getXLabelsFromLeafs(leafs);
+  return { valuesBySeries: { 'No metric defined': xLabels.map(() => null) }, xLabels };
+};
 
 const WorldMapVisualization: VisualizationComponent = ({ config, data, editing, onChange, width, ...rest }: VisualizationComponentProps) => {
-  const onRenderComplete = useContext(RenderCompletionCallback);
   const { rowPivots } = config;
+  const onRenderComplete = useContext(RenderCompletionCallback);
+  const hasMetric = !isEmpty(config._value.series);
```
dennisoelkers: `_value` is an internal field of the `AggregationWidgetConfig` class. please use `config.series` instead.

graylog2-web-interface/src/views/components/visualizations/worldmap/WorldMapVisualization.jsx
```diff
@@ -1,37 +1,48 @@
...
+    seriesExtractor,
     formatSeries,
-    results => results.map(arrayToMap),
+    results => results.map(_arrayToMap),
   ]);
+
   const series = pipeline(data)
     .map(({ name, x, y }) => {
-      const newX = x.map(lastKey);
+      const newX = x.map(_lastKey);
```
dennisoelkers: Optional: Not an issue with your code, but while you are working on it, I think it is a good idea to extract these two `.map(...)` calls into separate functions and them to the `flow([...])` statement, as they are part of the pipeline.

graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -37,14 +37,19 @@ export const formatSeries = ({ valuesBySeries, xLabels }: {valuesBySeries: Objec
   ]);
 };
 
+export const getLeafsFromRows = (rows: Rows): Array<Leaf> => {
+  // $FlowFixMe: Somehow flow is unable to infer that the result consists only of Leafs.
+  return rows.filter(row => (row.source === 'leaf'));
+};
+
+export const getXLabelsFromLeafs = (leafs: Array<Leaf>): Array<any> => leafs.map(({ key }) => key);
```
dennisoelkers: The return type of this should actually be `Array<Array<Key>>` instead of `Array<any>`. Although `Key` is currently `*`, we might be able/want to narrow it down in the future. Another alternative is to specify the type of `map` and do something like this:

```suggestion
export const getXLabelsFromLeafs = (leafs: Array<Leaf>) => leafs.map<Array<Key>>(({ key }) => key);
```

graylog2-web-interface/src/views/components/visualizations/worldmap/WorldMapVisualization.jsx
linuspahl: Good idea. I've created another function and also summarized the two `.map`s in one `.map`.

PR #6745, opened 2019-11-07
Themed Nav & Navbar components

kyleknighted:
## Description

- Themed `Nav`
- Themed `Navbar`

image
image

## Types of changes

## Checklist:


graylog2-web-interface/src/components/graylog/Navbar.jsx
```diff
@@ -0,0 +1,260 @@
...
+      > .open > a {
+        &,
+        &:hover,
+        &:focus {
+          color: ${darken(0.15, teinte.tertiary.uno)};
+          background-color: ${darken(0.065, teinte.secondary.due)};
+        }
+      }
+
+      @media (max-width: @grid-float-breakpoint-max) {
```
linuspahl: I wish that would work, but as far as I know, this needs to be a fix width.

graylog2-web-interface/src/components/graylog/Navbar.jsx
```diff
@@ -0,0 +1,260 @@
...
+      > .open > a {
+        &,
+        &:hover,
+        &:focus {
+          color: ${teinte.primary.due};
+          background-color: ${lighten(0.20, teinte.primary.tre)};
+        }
+      }
+
+      @media (max-width: 767px) {
```
linuspahl: I think this is the right time to store the breakpoints in e.g. the theme as well. What do you think? We would still need to unify them with the bootstrap variables, but it's an improvement, compared to hard coded values in the styles.

graylog2-web-interface/src/components/graylog/Navbar.jsx
kyleknighted: It was just a missed variable replacement, thanks for spotting it

graylog2-web-interface/src/components/graylog/Navbar.jsx
kyleknighted: Sounds good

graylog2-web-interface/src/components/graylog/Nav.jsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  &.nav-tabs.nav-justified {
+    > .active > a,
+    > .active > a:hover,
+    > .active > a:focus {
+      border-color: ${teinte.tertiary.quattro};
+    }
+
+    @media (min-width: 768px) {
```
linuspahl: You add the them breakpoint here as well. I really appreciate this implementation!

graylog2-web-interface/src/theme/breakpoints.js
```diff
@@ -0,0 +1,19 @@
...
+  sm: '768px',
+  md: '992px',
+  lg: '1200px',
+};
+
+const max = {};
+
+Object.keys(min).forEach((bp) => {
+  max[bp] = min[bp] - 1;
+});
```
linuspahl: Looks like this will set every value in max to NaN

PR #6803, opened 2019-11-14
Refactor search sidebar layout

### LINKED ISSUE #6725: Saved Search shows "No Description"
## Current Behavior

If you save a search, it doesn't allow for a Description, but when you load a search, it shows that Description was provided.

Either remove the description from the Load page or add Description to the Save dialog

image

image

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

linuspahl:
This PR
- unifies the sidebar sections layout
- changes the structure of the create section. Instead of list of buttons, we are displaying two named sections.
- implements a spacer for the toggle arrow / close button. Looks a bit unfamiliar first, but I think it is important to distinguished these buttons from the section icons.
- removes the "cache" for the last clicked section. In my opinion it feels more intuitive now.
- adds the option to expand sections to the right side. So far implemented for the field list. The field list is still not using it's full with. It is not so easy to do this in a dynamic way. For now we could increase the width manually.
- shows the loading spinner for each section, instead of the full sidebar.
- it converts the sidebar css classes to styled components. I started with the container initially, because it fixes a strange content flashing on page load. I continued with all classes, because
  - it was  very helpful for some use cases
  - it felt strange to mix these two ways
  - it improves the components render part readability
- fixes #6725

All of the above points are just suggestions, your feedback is very appreciated!

## How Has This Been Tested?
Adjusted the existing tests

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
```diff
@@ -1,65 +1,130 @@
...
+  padding: 9px 10px;
+  display: flex;
+  align-items: center;
+  cursor: pointer;
+  position: relative;
+
+  ${props => (props.isSelected ? css`
+    color: #FF3633;
+    background: #393939;
+  ` : '')}
```
linuspahl: I could not figure out a way to use something like
```
${props => (props.isSelected && css`
  color: #FF3633;
  background: #393939;
`)}
```


It throws the following flow error:
Cannot assign template string to `Title` because  `InterpolatableComponent` [1] is incompatible with  `React.ComponentType`

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
```diff
@@ -1,18 +1,77 @@
...
+  padding: 0 10px;
+  margin: 5px 0 10px 0;
+  
+  &::after {
+    content: ' ';
+    display: block;
+    width: 100%;
+    border-bottom: 1px solid currentColor;
+  }
+`;
```
linuspahl: This discussion is maybe something for the frontend chat:

What do you think about creating a separate file for these styled-components as well, when they are to many? In my opinion the component looks not that clear / readable with 60 lines of styling.

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
```diff
@@ -98,85 +152,60 @@ class SideBar extends React.Component<Props, State> {
...
-                        {children}
-                      </NavItem>
-                    </React.Fragment>
-                  )
-              }
-            </div>
-          </div>
-        </div>
-      </div>
+      <Container ref={(node) => { this.wrapperRef = node; }} open={open}>
```
linuspahl: I still get the following floe error here:

Cannot assign `node` to `this.wrapperRef` because  mixed [1] is incompatible with  `HTMLDivElement`

It works, when I use the old way with `innerRef` but it throws many deprecated warnings, when running the tests.

Looks like this is related to styled-components, I could not find an elegant way.. the only "solution" which worked for me is adding the type `mixed` for the ref class attribute.

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
dennisoelkers: I don't think there is much to discuss here. If a component grows too big, it is a good idea to move it to a separate file.

linuspahl: Great! I've created a new file for the styles. We should just define / discuss about a convention for the file name.
I don't like the idea (anymore) to have a file like `./styles.jsx` inside each component directory.
So far, I've named the file `Sidebar.styles.jsx`.

graylog2-web-interface/src/views/components/sidebar/AddWidgetButton.jsx
```diff
@@ -2,11 +2,30 @@
...
 
-import { ButtonGroup, Button } from 'components/graylog';
+import { Button } from 'components/graylog';
 import { ViewStore } from 'views/stores/ViewStore';
 import View from 'views/logic/views/View';
 
+
+const Group = styled.div`
+  margin-bottom: 20px;
+  &:last-child {
```
kyleknighted: the `&` is unnecessary unless it helps you read the code better.
Also, what do you think about having a blank line between blocks of styles?

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
kyleknighted: You could try something like
```
color: ${({ isSelected }) => isSelected ? '#FF3633' : 'currentcolor';
background-color: ${({ isSelected }) => isSelected ? '#393939' : 'currentcolor';
```

or maybe without the `css` function
```
${props => (props.isSelected && `
  color: #FF3633;
  background: #393939;
`)}
```

kyleknighted: You could also do something like

```
export const Title: React.ComponentType<StyleProps> = styled.div(props => `
/* ... */
color: ${props.isSelected ? '#FF3633' : 'currentcolor'};
/* ... */
`);
```

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
```diff
@@ -1,65 +1,130 @@
...
+
+  ${props => (props.isSelected && props.expandRight ? css`
```
kyleknighted:
```suggestion
  ${props => ((props.isSelected && props.expandRight) && css`
```

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
```diff
@@ -1,65 +1,130 @@
...
+      display: block;
+      position: absolute;
+      right: 0;
+      width: 0;
+      height: 0;
+      border-top: 15px solid transparent;
+      border-right: 15px solid white;
+      border-bottom: 15px solid transparent;
+    }
+  ` : '')}
```
kyleknighted:
```suggestion
  `)}
```

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
```diff
@@ -1,65 +1,130 @@
...
-    : styles.contentClosed;
-
   return (
-    <div>
-      <div role="presentation" onClick={onClick} className={`${styles.sidebarNav} ${selectedColor}`}>
-        <div className={styles.sidebarIcon}>{icon}</div>
-        {(isOpen && <div className={styles.sidebarNavFont}>{text}</div>)}
-      </div>
+    <React.Fragment>
+      <Title role="presentation" onClick={onClick} isSelected={isSelected} expandRight={expandRight}>
```
kyleknighted: @Graylog2/frontend what do we think about adding a lint rule that requires line breaks after `N` props/attributes? Maybe at 3 or 4 we require breaks for ease of readability?

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
```diff
@@ -2,17 +2,18 @@
...
 import PropTypes from 'prop-types';
 
-import { AddWidgetButton, SearchResultOverview } from 'views/components/sidebar';
+
+import { AddWidgetButton } from 'views/components/sidebar';
 import { Icon, Spinner } from 'components/common';
 import type { ViewMetaData } from 'views/stores/ViewMetadataStore';
 
-import styles from './SideBar.css';
+import { Container, ContentOverlay, SidebarHeader, Headline, ToggleIcon, HorizontalRuler } from './Sidebar.styles.jsx';
```
kyleknighted: @Graylog2/frontend Same thought on imports - do we want to have a max number of imports before we want to start line breaking?

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
```diff
@@ -2,17 +2,18 @@
...
 import CustomPropTypes from '../CustomPropTypes';
 import HighlightingRules from './highlighting/HighlightingRules';
 import NavItem from './NavItem';
+import ViewDescription from './ViewDescription';
 
-const defaultNewViewTitle = 'New View';
-const defaultNewViewSummary = 'No summary.';
+const defaultNewViewTitle = 'New Search';
```
kyleknighted: If this is a file constant, what do you think of using an `ALL_CAPS` variable name? Or should those only be used on app wide constants?

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
linuspahl: Thanks for the hints! I did not thought about `styled.div({ isSelected } => css`. It allows the usage of 
```
${(isSelected && css`
    color: #FF3633;
    background: #393939;
`)}
```
This will be very helpful in the future! But in this case I will use something like your first example.

linuspahl: Just saw I still get a different Flow error. I am trying to find a solution.

kmerz: In doubt I would check if our flow typed definitions are uptodate?

graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
```diff
@@ -0,0 +1,47 @@
...
+  viewMetadata: ViewMetaData,
+};
+
+const ViewDescription = ({ results, viewMetadata }: Props) => {
+  const formatViewDescription = (view: ViewMetaData) => {
+    const { description } = view;
+    if (description) {
+      return <span>{description}</span>;
+    }
+    return <i>No view description.</i>;
```
kmerz: A search does not have a `description` and will not get any.. so you could make a `<IfDashboard />` arround it somehow and fix: #6725

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
```diff
@@ -89,7 +107,7 @@ const ViewAdditionalContextProvider = connect(AdditionalContext.Provider, { view
 const ExtendedSearchPage = ({ route, searchRefreshHooks }: Props) => {
   const refreshIfNotUndeclared = view => _refreshIfNotUndeclared(searchRefreshHooks, SearchExecutionStateStore.getInitialState(), view);
 
-  useEffect(() => {
+  React.useEffect(() => {
```
kmerz: I like `useEffect` directly imported more. What is the reason for changing  it?

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
kmerz: I would say the second one. Only for app wide constants. But I am open for arguments :smiley:

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
kmerz: I do not care to much. But I try to not go over 120 chars.

graylog2-web-interface/src/views/components/sidebar/AddWidgetButton.jsx
linuspahl: Good to know. I like the idea of a blank line between blocks of styles.

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
linuspahl: +1
I am a fan of this rule.

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
linuspahl: @konrad I changed it, because I need to use `React.ComponentType`
I also prefer the direct import. Are you fine with:
```
import React, { useEffect } from 'react';
import type { ComponentType } from 'react';
```
Feels a bit strange to import from react multiple times, but I could deal with it.

graylog2-web-interface/src/views/components/sidebar/SideBar.jsx
linuspahl: Not a problem anymore.

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
kmerz: Thats how we do it on other places! So I am fine with it. :+1:

graylog2-web-interface/src/views/components/sidebar/NavItem.jsx
dennisoelkers: IMO there are two ways to solve this:

- implicitly by the `max-len` rule (which is currently off), enforcing line breaks after N characters
- force switching to a multi-line declaration (with a single prop per line) after 3-4 props

I am open for both options.

linuspahl:
@dennisoelkers Thanks for the review and feedback. I renamed the sidebar section.

PR #6819, opened 2019-11-16
Add testcontainers based MongoDBBaseTest class

bernd:
This class can be used for integration tests that need a MongoDB
database. It is using the testcontainers library to start a database
instance for each test class.

It supports our standard database fixture files by using the
`MongoDBFixtures` annotation on test methods.


bernd:
Closing this because I have a better idea to make this more flexible. (without having to inherint from a base class)

Issue #6823, opened 2019-11-18
Cluster Global API Browser is showing wrong url


mummy1027:
I have 3.1 running on a host with nginx reverse proxy.

It is okay to access example.io/graylog/api/api-browser/global/index.html

But Cluster Global API Browser is showing example.io/graylog/api/graylog/api-browser/global/index.html.

More details could be found on 
https://community.graylog.org/t/cluster-global-api-browser-and-api-browser-seems-not-right/12798

* Graylog Version: 3.1.3
* Elasticsearch Version:6.8.3
* MongoDB Version:4.0.10
* Operating System:Ubuntu 18.04
* Browser version:Chrome

mpfz0r:
Thank you for your report. That's indeed a bug.

PR #6826, opened 2019-11-19
Fix global API browser link

mpfz0r:
Using Routes, will run 'qualifyUrl()' twice.

### DIFF
--- a/graylog2-web-interface/src/pages/NodesPage.jsx
+++ b/graylog2-web-interface/src/pages/NodesPage.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import createReactClass from 'create-react-class';
 import Reflux from 'reflux';
-import Routes from 'routing/Routes';
 import URLUtils from 'util/URLUtils';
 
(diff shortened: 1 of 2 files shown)


jalogisch:
I guess this FIX: https://github.com/Graylog2/graylog2-server/issues/6823 ?

graylog2-web-interface/src/pages/NodesPage.jsx
```diff
@@ -28,7 +28,7 @@ const NodesPage = createReactClass({
     }
     if (this._hasExternalURI()) {
       return (
-        <ExternalLinkButton bsStyle="info" href={URLUtils.qualifyUrl(Routes.GLOBAL_API_BROWSER)}>
+        <ExternalLinkButton bsStyle="info" href={URLUtils.qualifyUrl('/api-browser/global/index.html')}>
```
dennisoelkers: Can you please extract it into a constant in the same file? That makes it way more visible.

PR #6836, opened 2019-11-20
Fix adding new converters to a extractor

edmundoa:
This seems to have been broken forever, at least I couldn't find the cause on a quick look at the file.

Before the change, adding an extractor converter didn't add anything to the list of converters, as you can see in the gif:
converter

### DIFF
--- a/graylog2-web-interface/src/components/extractors/EditExtractorConverters.jsx
+++ b/graylog2-web-interface/src/components/extractors/EditExtractorConverters.jsx
@@ -45,7 +45,9 @@ class EditExtractorConverters extends React.Component {
   };
 
   _onConverterAdd = () => {
-    this.setState({ selectedConverter: undefined });
+    const { displayedConverters, selectedConverter } = this.state;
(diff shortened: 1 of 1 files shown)


edmundoa:
Looking a bit more on this after talking to @dennisoelkers, it looks like this was working as expected on both 2.4.6 and 3.1.3. I will investigate a bit more when exactly this broke, even if to know the cause of it.

edmundoa:
So looking a bit more into the code, it seems that the code stopped working after a51beb16dd2f36dc10a03fc8bcb94dda29484b2a. Looking at the `_onConverterAdd` in that revision:

```
 _onConverterAdd = () => {
    const newDisplayedConverters = this.state.displayedConverters;
    newDisplayedConverters.push(this.state.selectedConverter);
    this.setState({ selectedConverter: undefined, converters: newDisplayedConverters });
  };
```

we stored the updated `this.state.displayedConverters` into the wrong state key, but since we updated the array with `push()` the array was mutating and the changes affected the `displayedConverters` key as well.

PR #6847, opened 2019-11-22
Fix "COMMONAPACHELOG" grok pattern.

### LINKED ISSUE #1647: GROK extractor COMMONAPACHELOG overrides timestamp with incompatible value
GROK `${COMMONAPACHELOG}` overrides `timestamp` field with badly formatted value resulting in an error.

`Graylog 2.0.0-20151217162713`

``` json
{"request":"/","MONTH":"Dec","agent":"curl/7.38.0","COMMONAPACHELOG":"172.17.0.6 - - [22/Dec/2015:14:51:51 +0000] \"GET / HTTP/1.1\" 403 381","docker_name":"/clever_blackwell","HOUR":"14","auth":"-","gl2_remote_ip":"10.42.19.191","ident":"-","gl2_remote_port":36239,"BASE10NUM":"381","source":"2e9bc6ef2102","gl2_source_input":"5679495ce4b04887915b4780","INT":"+0000","YEAR":"2015","clientip":"172.17.0.6","COMBINEDAPACHELOG":"172.17.0.6 - - [22/Dec/2015:14:51:51 +0000] \"GET / HTTP/1.1\" 403 381 \"-\" \"curl/7.38.0\"","USERNAME":"-","gl2_source_node":"f65bd405-369c-47f2-9432-0b86cb740c05","MINUTE":"51",
"timestamp":"22/Dec/2015:14:51:51 +0000",
"IP":"172.17.0.6","IPV4":"172.17.0.6","streams":[],"verb":"GET","TIME":"14:51:51","message":"172.17.0.6 - - [22/Dec/2015:14:51:51 +0000] \"GET / HTTP/1.1\" 403 381 \"-\" \"curl/7.38.0\"","USER":"-","version":"1.1","referrer":"-","docker_container":"0621eb14e4bd3e3e2e76080436baf63a6a2cfccffccc302f0c72e8114d4f0c3d","bytes":"381","response":"403","SECOND":"51","httpversion":"1.1","QUOTEDSTRING":"curl/7.38.0","MONTHDAY":"22","docker_image":"php:5.6-apache"}
```

`ES 2.1`

```
failed to parse [timestamp]]; nested: IllegalArgumentException[Invalid format: "22/Dec/2015:14:51:51 +0000" is malformed at "/Dec/2015:14:51:51 +0000"
```

thll:
## Description

The COMMONAPACHELOG grok pattern uses the original date string found in the source to propagate the "timestamp" message field. When the extractor tries to use this value to set the field on the message, it will be ignored, as it is not a proper date.

We adjust the pattern so that it creates a date object which the message accepts as a valid value for the "timestamp" field.

The pattern is fixed in the content pack which is installed automatically for a new graylog setup. For existing setups, a migration is provided which fixes the pattern if it is still in the original, broken state. 

## Motivation and Context

Fixes #1647

## How Has This Been Tested?

Started a fresh graylog setup. Started an existing graylog setup with the original content pack already installed. Both setups ended up with the fixed pattern.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/V20191121145100_FixDefaultGrokPatterns.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/V20191121145100_FixDefaultGrokPatterns.java
@@ -1,0 +1,150 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.migrations;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import org.graylog.autovalue.WithBeanGetter;
+import org.graylog2.grok.GrokPattern;
+import org.graylog2.grok.GrokPatternService;
+import org.graylog2.plugin.cluster.ClusterConfigService;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/migrations/V20191121145100_FixDefaultGrokPatterns.java
```diff
@@ -0,0 +1,150 @@
...
+    public ZonedDateTime createdAt() {
+        return ZonedDateTime.parse("2019-11-21T14:51:00Z");
+    }
+
+    @Override
+    public void upgrade() {
+        final MigrationCompleted migrationCompleted = configService.get(MigrationCompleted.class);
+
+        final Set<String> patternNames = patternsToMigrate.stream()
+                .map(PatternToMigrate::name)
```
mpfz0r: One small nitpick: If we want to write a second migration for the same pattern, we cannot use the name here, because it would be skipped. But a second identifier could still be introduced when need it :)

Issue #5965, opened 2019-05-23
filter UI element resets when actions are performed


gimmic:
When filtering lists (such as pipeline rules) using the filter function, the filter is 'reset' when you perform an action against a selected object.

## Expected Behavior
The filter should remain in place even after performing actions on items in the filtered list.

## Current Behavior
The filter still shows the selected option, but the list resets to an unfiltered state after performing an action. 

![filter-reset](https://user-images.githubusercontent.com/568957/58265831-0f562f80-7d46-11e9-8441-d03f28d9b3f5.gif)

## Possible Solution
Filters should remain active when remaining inside the same UI view. 

## Steps to Reproduce (for bugs)
Create pipeline rules. Filter rules for objects, delete one of the objects. Filtered view is reset.

## Context
You have to re-apply the filter while working on a filtered set. This can be annoying in workflow and runs counter-intuitive to a filtered view. 

* Graylog Version: 3.0.1

PR #6852, opened 2019-11-22
Fix DataTable filter to reapply filter after data updates

ousmaneo:
Fix DataTable filter to reapply filter after data updates
## Description

Before this change when filtering lists (such as pipeline rules) using the filter function, the filter is 'reset' when you perform an action against a selected object. Now the filter is re-applied if data is changed on component update.

Fix (#5965)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/DataTable.jsx
+++ b/graylog2-web-interface/src/components/common/DataTable.jsx
@@ -10,8 +10,6 @@ import DataTableElement from './DataTableElement';
  * input to the data table by using the the `TypeAheadDataFilter` component.
  */
 class DataTable extends React.Component {
-  typeAheadRef = React.createRef();
-
   static propTypes = {
     /** Adds a custom children element next to the data filter input. */
     children: PropTypes.node,
@@ -60,7 +58,7 @@ class DataTable extends React.Component {
      */
     useResponsiveTable: PropTypes.bool,
   };
- 
+
   static defaultProps = {
     children: undefined,
     className: '',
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/common/TypeAheadDataFilter.jsx
```diff
@@ -131,59 +162,63 @@ class TypeAheadDataFilter extends React.Component {
...
     }
 
-    const filteredData = this.props.data.filter((datum) => {
+    const filteredData = data.filter((datum) => {
       return this._matchFilters(datum) && this._matchStringSearch(datum);
     }, this);
 
-    this.props.onDataFiltered(filteredData);
+    onDataFiltered(filteredData);
+    return true;
```
edmundoa: I don't understand why we return `true` in this function, could you please explain that?

graylog2-web-interface/src/components/common/DataTable.jsx
```diff
@@ -121,36 +132,59 @@ class DataTable extends React.Component {
     this.setState({ filteredRows });
   };
 
+  _updateState() {
```
edmundoa: Since `headers` and `rows` always mirror the values in `props`, I think it would be good to remove them from the state. This method should only reset the `filteredRows` if the original `rows` changed. We should also rename the method to make clearer what it does.

graylog2-web-interface/src/components/common/DataTable.jsx
```diff
@@ -71,25 +74,32 @@ class DataTable extends React.Component {
...
-  };
+  constructor(props) {
+    super(props);
+    const { headers, rows } = this.props;
+    this.state = {
+      headers: headers,
+      rows: rows,
+      filteredRows: rows,
+    };
+    this.typeAheadRef = React.createRef();
```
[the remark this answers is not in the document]
ousmaneo: Ok I missed that :)

graylog2-web-interface/src/components/common/TypeAheadDataFilter.jsx
ousmaneo: This is to fix linter issue : https://eslint.org/docs/rules/consistent-return

edmundoa: Okay. Looking at where this function is used, I think in this case we only use a return before in order to not execute the rest of the function when a `filterData` function is given. I think it would be clearer to return with no value in the if case and remove this line.

graylog2-web-interface/src/components/common/DataTable.jsx
```diff
@@ -71,25 +73,29 @@ class DataTable extends React.Component {
...
+      this._updateState();
+    }
   }
 
   getFormattedHeaders = () => {
     let i = 0;
-    const formattedHeaders = this.state.headers.map((header) => {
-      const el = <DataTableElement key={`header-${i}`} element={header} index={i} formatter={this.props.headerCellFormatter} />;
-      i++;
+    // const { headers } = this.props;
```
edmundoa: We should remove this line now :)

PR #6859, opened 2019-11-25
Adding `eslint-plugin-flowtype` for proper flow support in ESLint.

dennisoelkers:
Prior to this change, ESLint threw errors for undefined variables in some cases when flow utility types have been used. These required disabling the `no-undef` rule selectively when this was the case.

This PR adds the [`eslint-plugin-flowtype`](https://github.com/gajus/eslint-plugin-flowtype) ESLint plugin, which helps avoiding this. Right now it is using the recommended configuration, further fine-tuning according to our desires can be performed later.


kyleknighted:
Will this PR be followed-up with something removing the disabled `no-undef` rules?

dennisoelkers:
@kyleknighted: Sure, even in the same PR! :)

PR #6868, opened 2019-11-27
Add delay for displaying the loading spinner

linuspahl:
Adding a 200ms delay for displaying the loading spinner is a simple change to improve the look and feel.

The idea:
Everything that loads faster than 200ms does not need a loading spinner. When something takes longer than 200ms to load, the user will not really see the delay, because it is so short.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Spinner.jsx
+++ b/graylog2-web-interface/src/components/common/Spinner.jsx
@@ -1,22 +1,52 @@
-import React from 'react';
+// @flow
+import React, { useState, useEffect } from 'react';
+import type { ComponentType } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 
 import Icon from './Icon';
+
+const Wrapper: ComponentType<{ visible: boolean }> = styled.span`
+  visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
+`;
+
+type Props = {
+  delay?: number,
+  name?: string,
+  text?: string,
+}
 
 /**
  * Simple spinner to use while waiting for something to load.
  */
-const Spinner = ({ name, text, ...props }) => <span><Icon name={name} spin {...props} /> {text}</span>;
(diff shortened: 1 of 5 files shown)


linuspahl:
I just saw, there is still an error `Error: Uncaught [Invariant Violation: Invalid hook call.` when running the tests. I will have another look at this during the day.

graylog2-web-interface/src/components/common/Spinner.jsx
```diff
@@ -1,22 +1,52 @@
...
+// @flow
+import React, { useState, useEffect } from 'react';
+import type { ComponentType } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 
 import Icon from './Icon';
 
+const Wrapper: ComponentType<{ visible: boolean }> = styled.span`
+  visibility: ${({ visible }) => (visible ? 'visible' : 'hidden')};
```
kyleknighted: Should this be `display: none`? `visibility` keeps the space in the layout allocated while display removes the rendering completely

graylog2-web-interface/src/components/common/Spinner.jsx
```diff
@@ -1,22 +1,52 @@
...
+
+  useEffect(() => {
+    const delayTimeout = window.setTimeout(() => {
+      setDelayFinished(true);
+    }, delay);
+
+    return () => clearTimeout(delayTimeout);
+  }, []);
+
+  return <Wrapper visible={delayFinished}><Icon name={name} spin {...rest} /> {text}</Wrapper>;
```
kyleknighted: maybe we just `return null` if the Spinner should not be visible so that we're not drawing invisible elements?

kyleknighted:
@linuspahl Looks like you'll need to create an Enterprise snapshot update as well

linuspahl:
I am closing this PR, because I've created a new branch for these changes, based on the current master: https://github.com/Graylog2/graylog2-server/pull/7539

Issue #5616, opened 2019-01-29
Installing Content Pack may create new All messages Stream


edmundoa:
## Expected Behavior
The "All messages" default Stream is somehow special, and installing a Content Pack that has it in the entity list should not create a new Stream for it, but use the existing one.

## Current Behavior
Installing a Content Pack that has the "All messages" Stream as dependency creates a new "All messages" Stream, that does not behave in the same way as the original one.

That affects Pipelines that may need to be applied into all messages, and specially Dashboard Widgets created against the Stream. The latter case is specially bad, because we don't let users alter the Stream after the Widget was created.

Example of content pack having this issue: 
[v2.txt](https://github.com/Graylog2/graylog2-server/files/2808711/v2.txt)

## Steps to Reproduce (for bugs)
1. Upload given Content Pack
2. Install the Content Pack
3. Go to the Streams page
4. See that two "All messages" Streams exist

## Your Environment
* Graylog Version: Graylog 3.0.0-rc.1-SNAPSHOT (790939a)

bernd:
Refs https://github.com/Graylog2/graylog2-server/issues/5439

pschichtel:
Opened the pull request #6873.

PR #6873, opened 2019-11-27
Allow pipelines connected to "All messages" in content packs

### LINKED ISSUE #5439: Default stream must not added as entity to new content pack
When a pipeline is attached to the default stream the content pack will contain it as an entity, which then prevents installing it.

## Expected Behavior
When creating content pack with a pipeline and rule, it should be installable.

## Current Behavior
If the stream the pipeline is bound to already exists in the system, which is always the case for the default stream, then it cannot be installed at all.

1. Create a pipeline, attach it to the default stream.
2. Create a content pack with that pipeline (it will add the default stream into the entity list).
3. Install the newly created content pack.
4. Note that it cannot be installed: an error is shown.

## Your Environment
* Graylog Version: 3.0.0-alpha.5

pschichtel:
## Description

This pull request prevents the content pack exporter from exporting clones of default streams when including pipelines that are connected to e.g. "All messages".

This closes #5439 and #5616.

## Motivation and Context

In the project I'm working we intend to bootstrap a graylog container from scratch without any manual configuration for consistent production deployments and automated updates. This requires us to include out "entrypoint pipeline" on "All messages" for initial routing. Currently this is not possible, as the import will fail.

## How Has This Been Tested?

As of now the changes have been tested in a local test environment.

## Types of changes

## Checklist:

There is a test failure, at least on my system, that has been there before I changed anything and still failes. It would be great of someone else could verify. I updated the PipelineFacade test, but I haven't added specific tests for the changes.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java
@@ -361,7 +361,6 @@ public class ContentPackService {
 
         final ImmutableSet.Builder<Entity> entities = ImmutableSet.builder();
         for (EntityDescriptor entityDescriptor : resolvedEntities) {
-            // don't export default streams
             if (EntityDescriptorIds.isDefaultStreamDescriptor(entityDescriptor)) {
                 continue;
             }
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java
@@ -55,7 +55,6 @@ import org.graylog2.contentpacks.model.entities.NativeEntityDescriptor;
 import org.graylog2.contentpacks.model.entities.references.ValueReference;
 import org.graylog2.contentpacks.model.entities.references.ValueType;
 import org.graylog2.contentpacks.model.parameters.Parameter;
-import org.graylog2.plugin.streams.Stream;
 import org.graylog2.utilities.Graphs;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


pschichtel:
Btw this also cleanly applied to 3.1

graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java
```diff
@@ -361,6 +361,10 @@ private ContentPackUninstallation uninstallContentPack(ContentPackInstallation i
 
         final ImmutableSet.Builder<Entity> entities = ImmutableSet.builder();
         for (EntityDescriptor entityDescriptor : resolvedEntities) {
+            // don't export default streams
```
kmerz: Please remove the comment. Your code is self-explaining.

graylog2-server/src/main/java/org/graylog2/contentpacks/EntityDescriptorIds.java
```diff
@@ -44,11 +46,22 @@ public static EntityDescriptorIds of(EntityDescriptor... entityDescriptors) {
 
     public static EntityDescriptorIds of(Collection<EntityDescriptor> entityDescriptors) {
         final ImmutableMap<EntityDescriptor, String> descriptorIds = entityDescriptors.stream()
-                .collect(ImmutableMap.toImmutableMap(Function.identity(), d -> UUID.randomUUID().toString()));
+                .collect(ImmutableMap.toImmutableMap(Function.identity(), d -> {
+                    // use the hardcoded IDs for default streams
```
kmerz: Please remove unnecessary comment.

graylog2-server/src/main/java/org/graylog2/contentpacks/facades/PipelineFacade.java
```diff
@@ -156,8 +160,20 @@ Entity exportNativeEntity(PipelineDao pipelineDao, EntityDescriptorIds entityDes
             final Object stream = nativeEntities.get(descriptor);
             if (stream instanceof Stream) {
                 streams.add((Stream) stream);
-            } else {
-                throw new MissingNativeEntityException(descriptor);
+            }
+            else {
```
kmerz: please do not put else in a new line.

graylog2-server/src/main/java/org/graylog2/plugin/streams/Stream.java
```diff
@@ -100,4 +100,8 @@ public static MatchingType valueOfOrDefault(String name) {
     String getIndexSetId();
 
     void setIndexSetId(String indexSetId);
+
+    static boolean isDefaultStream(String id) {
```
kmerz: Since we give a `id` to check perhaps `isDefaultStreamId` is better?

pschichtel:
@kmerz Yep, I agree with suggestions. Just pushed a commit to the branch.

pschichtel:
The test is still missing though

pschichtel:
I added tests based on the existing createNativeEntity and exportNativeEntity tests

PR #6877, opened 2019-11-27
Ensure that aggregation events always have a "source_streams" value

### LINKED ISSUE #6876: Invalid "source_streams" for events created for an empty search result
## Expected Behavior

Events created by aggregations should always have a non-empty `source_streams` field because this is used for permission checks.

## Current Behavior

When users create an event definition that creates events when the aggregation doesn't return any results (e.g. create event when `count() < 1`), the resulting events don't have any streams set in the `source_streams` field.

1. Create new event definition
2. Configure search query that will not return a result
3. Use condition `count() < 1`
4. Wait until events are created
5. Check that created events have an empty `source_streams` field

## Your Environment

* Graylog Version: 3.1.3

bernd:
Fixes #6876

This needs to be backported to 3.1.


mpfz0r:
Works :+1:

Issue #6353, opened 2019-08-26
Support multiple functions and conditions in aggregation event defintions


bernd:
## Description

The "Filter & Aggregation" event definition is currently only supporting a single function and condition:

![image](https://user-images.githubusercontent.com/461/63704568-9fb37d00-c82b-11e9-885c-4a89cf7601bb.png)

To be able to allow more advanced use cases, we should allow multiple aggregation functions and conditions.

PR #6879, opened 2019-11-27
Support multiple conditions in aggregation event definitions

edmundoa:
Adds support for multiple conditions in aggregation event definitions. It also adds the ability to create groups of expressions in the UI, and a corresponding `group` expression in the backend.

Screenshot 2019-12-18 at 11 57 42

Implements https://github.com/Graylog2/graylog2-server/issues/6353


graylog2-web-interface/src/components/event-definitions/event-definition-types/AggregationConditionExpressions/NumberRefExpression.jsx
```diff
@@ -0,0 +1,122 @@
...
+  const createSeries = () => {
+    return { id: expression.ref };
+  };
+
+  const getOrCreateSeries = (seriesId) => {
+    return getSeries(seriesId) || createSeries();
+  };
+
+  const getSeriesId = (currentSeries, func, field) => {
+    return `${lodash.defaultTo(func, currentSeries.function)}-${lodash.defaultTo(field, currentSeries.field || '')}`;
```
mpfz0r: Hmm, this seems to be broken.
If I add a new condition to an already existing count() I get this json:
```
 "series" : [ 
            {
                "id" : "31bf89e7-33d7-49bd-8e16-ad5160867b9a",
                "function" : "count",
                "field" : null
            }, 
            {
                "id" : "count-",
                "function" : "count",
                "field" : null
            }
```
The new Id is called `count-` which also leads to the series being added twice.

edmundoa: This change is to avoid exactly that problem to happen. I don't think the frontend is the right place to normalize existing series, but we can think of adding a migration in the server for that.

edmundoa: So, to clarify, you get 2 series because one of them was already added and has a unique ID. Modifying that series in the frontend will change the ID to `<function>-<field>`, and since that's already there, there will only be one series.

mpfz0r: Ok, so that's on purpose.
So without the migration the backend would need to compute the series twice?

edmundoa: It may be, depends on what the user does. Computing the series twice doesn't work in the backend at the moment and I don't know how to fix it. I think it should also not be done, since that will only return the same results twice.

edmundoa: So from my point of view we should either do a migration or document this and ask people to edit the original condition if they run in this issue. I think the first option is a bit better, but I also didn't include it in the PR since I'm unsure how big of a problem this will be.

edmundoa:
I updated the PR with some comments from @mariussturm.
The series migration (https://github.com/Graylog2/graylog2-server/pull/6879#discussion_r360379898) and/or fix in the backend to de-duplicate series is not included, though.

mpfz0r:
I took the liberty and added some tests for the grouped expression.

graylog2-server/src/main/java/org/graylog/events/conditions/Expr.java
```diff
@@ -314,4 +316,32 @@ public Double accept(ExpressionVisitor visitor) {
...
     }
+
+    @AutoValue
+    public static abstract class Group implements Expression<Boolean> {
+        static final String EXPR = "group";
+
+        @JsonProperty(CHILD)
+        public abstract Expression<Boolean> child();
+
+        @JsonProperty(OPERATOR)
```
mpfz0r: maybe we could add a comment, that the operator isn't actually used for evaluating expressions?

graylog2-web-interface/src/components/event-definitions/event-definition-types/AggregationConditionExpressions/NumberRefExpression.jsx
mpfz0r: I've added a migration to this PR.

mpfz0r:
I've only skimmed over the frontend code, but it works fine. So if you're happy with my changes, go ahead and merge it. :+1:

edmundoa:
The changes @mpfz0r did look fine to me, waiting for @bernd to have another look.

PR #6880, opened 2019-11-27
Add testcontainers based MongoDBInstance rule

bernd:
This rule can be used for integration tests that need a MongoDB
database. It is using the testcontainers library to start a database
instance.

It supports our standard database fixture files by using the
MongoDBFixtures annotation on test methods.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
@@ -18,7 +18,6 @@ package org.graylog2.database;
 
 import com.mongodb.DB;
 import com.mongodb.Mongo;
-import com.mongodb.MongoClient;
 import com.mongodb.client.MongoDatabase;
 
 public interface MongoConnection {
@@ -39,11 +38,4 @@ public interface MongoConnection {
      * @return The configured MongoDB database.
      */
     MongoDatabase getMongoDatabase();
-
-    /**
-     * Get the client for the configured MongoDB database.
-     *
-     * @return the MongoDB client
-     */
-    MongoClient getMongoClient();
 }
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
```diff
@@ -38,4 +39,11 @@
      * @return The configured MongoDB database.
      */
     MongoDatabase getMongoDatabase();
+
+    /**
+     * Get the client for the configured MongoDB database.
+     *
+     * @return the MongoDB client
+     */
+    MongoClient getMongoClient();
```
alex-konn: So far we're not using this anywhere. Are you sure we're going to need it?

graylog2-server/src/test/java/org/graylog/testing/mongodb/MongoDBFixtureImporter.java
```diff
@@ -0,0 +1,243 @@
...
+
+            if (indexDoc.containsKey(FIELD_INDEX_OPTIONS)) {
+                collection.createIndex(indexFields, createIndexOptions(indexDoc.get(FIELD_INDEX_OPTIONS, Document.class)));
+            } else {
+                collection.createIndex(indexFields);
+            }
+        }
+    }
+
+    private IndexOptions createIndexOptions(Document indexOptionsDoc) {
```
alex-konn: I assume you have left out [`collation`](https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#collation) on purpose, because we don't need it, right?

bernd: Yes, I think we can add that once we need it. :smiley:

graylog2-server/src/main/java/org/graylog2/database/MongoConnection.java
bernd: Good catch! It might be a leftover from a previous iteration of the PR. I will check.

bernd: Aye, it's not needed anymore and I removed it. Thank you! :+1:

PR #6890, opened 2019-11-28
Render events in timebased bar, line and scatter charts

kmerz:
## Description
A new configuration option is add to the aggregation controls when a timestamp is
selected for rows and the graph is either a bar, line or scatter graph.
When enabled the widget will add a new search type ('events') to the pivot
search type to retrieve all events to the timerange matching the streams the user
is able to see or selected.
The user then sees vertical lines for a timestamp where alerts occurred.
The user can choose a color for the alert annotation.

## Motivation and Context
The user wants to see when an alert occurred and wants to compare it to other data
to draw conclusions from it.

## How Has This Been Tested?
Add a event definition which will produce alerts and configure the rendering of events
on a timebased bar char.

Graylog (30)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
@@ -1,0 +1,31 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { Checkbox, FormGroup } from 'components/graylog';
+
+type Props = {
+  enabled: boolean,
+  onChange: () => void,
+};
+
+const EventListConfiguration = ({ enabled, onChange }: Props) => {
+  return (
+    <form>
+      <FormGroup>
+        <Checkbox onChange={onChange} value={enabled}>Enable Event Annotation</Checkbox>
+      </FormGroup>
+    </form>
+  );
+};
+
+EventListConfiguration.propTypes = {
+  enabled: PropTypes.bool,
+  onChange: PropTypes.func,
+};
+
+EventListConfiguration.defaultProps = {
+  enabled: false,
+  onChange: () => {},
+};
+
+export default EventListConfiguration;
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
@@ -12,8 +12,8 @@ const EventListConfiguration = ({ enabled, onChange }: Props) => {
   return (
     <form>
       <FormGroup>
-        {/* eslint-disable-next-line no-undef */ /* $FlowFixMe: checked is part of target */}
-        <Checkbox onChange={(event: SyntheticEvent<HTMLInputElement>) => onChange(event.target.checked)} checked={enabled}>
+        {/* eslint-disable-next-line no-undef */}
+        <Checkbox onChange={(event: SyntheticInputEvent<HTMLInputElement>) => onChange(event.target.checked)} checked={enabled}>
           Enable Event Annotation
         </Checkbox>
       </FormGroup>
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
@@ -5,14 +5,17 @@ import { Checkbox, FormGroup } from 'components/graylog';
 
 type Props = {
   enabled: boolean,
-  onChange: () => void,
+  onChange: (value: boolean) => void,
 };
 
 const EventListConfiguration = ({ enabled, onChange }: Props) => {
   return (
     <form>
       <FormGroup>
-        <Checkbox onChange={onChange} value={enabled}>Enable Event Annotation</Checkbox>
+        {/* eslint-disable-next-line no-undef */ /* $FlowFixMe: checked is part of target */}
+        <Checkbox onChange={(event: SyntheticEvent<HTMLInputElement>) => onChange(event.target.checked)} checked={enabled}>
+          Enable Event Annotation
+        </Checkbox>
       </FormGroup>
     </form>
   );
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.test.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.test.jsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { render, fireEvent, cleanup } from '@testing-library/react';
+
+import EventListConfiguration from './EventListConfiguration';
+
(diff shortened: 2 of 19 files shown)


graylog2-web-interface/src/views/components/visualizations/bar/BarVisualization.jsx
```diff
@@ -42,10 +43,16 @@ const BarVisualization: VisualizationComponent = ({ config, data, effectiveTimer
   const _seriesGenerator = (type, name, labels, values): ChartDefinition => ({ type, name, x: labels, y: values, opacity });
 
   const rows = data.chart || Object.values(data)[0];
+  const chartDataResult = chartData(config, rows, 'bar', _seriesGenerator);
+  if (config.eventAnnotation && data.events) {
+    const { eventChartData, shapes } = EventHandler.toVisualizationData(data.events, config.formattingSettings);
+    chartDataResult.push(eventChartData);
+    layout.shapes = shapes;
+  }
```
linuspahl: I don't know the best place for it right now, but it would be great to extract this logic to a function, which we can use in all three charts.

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
```diff
@@ -0,0 +1,34 @@
...
+  enabled: boolean,
+  onChange: (value: boolean) => void,
+};
+
+const EventListConfiguration = ({ enabled, onChange }: Props) => {
+  return (
+    <form>
+      <FormGroup>
+        {/* eslint-disable-next-line no-undef */ /* $FlowFixMe: checked is part of target */}
+        <Checkbox onChange={(event: SyntheticEvent<HTMLInputElement>) => onChange(event.target.checked)} checked={enabled}>
```
linuspahl: Is there a reason, you are using `SyntheticEvent` here? With `SyntheticInputEvent`, we don't need the $FlowFixMe comment

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
```diff
@@ -0,0 +1,34 @@
...
+        </Checkbox>
+      </FormGroup>
+    </form>
+  );
+};
+
+EventListConfiguration.propTypes = {
+  enabled: PropTypes.bool,
+  onChange: PropTypes.func,
+};
```
linuspahl: Mainly because we are describing these attributes as required in the flow type definition, I would do the same for the PropType definition.

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.test.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { render, fireEvent, cleanup } from '@testing-library/react';
+
+import EventListConfiguration from './EventListConfiguration';
+
+describe('EventListConfiguration', () => {
+  afterEach(cleanup);
+
+  it('should render minimal', () => {
+    const { container } = render(<EventListConfiguration />);
+    expect(container).toMatchSnapshot();
```
linuspahl: This is more a general discussion. What is the reason you are using a snapshot here? Couldn't we test the same case, by just checking if the text `Enable Event Annotation` gets rendered?

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
kmerz: I did not know :D

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.jsx
kmerz: Since both values have defaults, they should never be optional for flow.

graylog2-web-interface/src/views/components/aggregationbuilder/EventListConfiguration.test.jsx
kmerz: I feel like this represents the widget better. But I am open for suggestions. Checking on the name though, I don't think it is enough.

graylog2-web-interface/src/views/components/visualizations/bar/BarVisualization.jsx
kmerz: `toVisualizationData` is already all that I could extract. If you have any better idea I am open for suggestions.

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+};
+
+export type Shapes = Array<Shape>;
+const eventsDisplayName = 'Alerts';
+
+export default {
+  convert(events: Array<Event>) {
+    return events;
+  },
```
linuspahl: Could you explain in which case we need the convert function?

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
```diff
@@ -0,0 +1,105 @@
...
+      marker: {
+        size: 5,
+        color: '#d3d3d3',
+      },
+    };
+  },
+
+  toShapeData(timestamps: Array<string>, formattingSettings: WidgetFormattingSettings = WidgetFormattingSettings.create({})): Shapes {
+    const { chartColors } = formattingSettings;
+    const shapeColor = chartColors[eventsDisplayName] || '#d3d3d3';
```
linuspahl: Just a minimal improvement, but we could store the default color in variable and use it in `toChartData` as well.

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
```diff
@@ -0,0 +1,105 @@
...
+      mode: 'markers',
+      name: eventsDisplayName,
+      type: 'scatter',
+      opacity: 0.5,
+      x: xValues,
+      y: yValues,
+      text: textValues,
+      marker: {
+        size: 5,
+        color: '#d3d3d3',
```
linuspahl: When the user selects a custom color, the alerts legend color is still grey. We could just pass the `formattingSettings` to this function, to use the custom color here as well.

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
kmerz: So this code, does not break: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/views/logic/QueryResult.js#L12

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
kmerz: Graylog (33)
It is already that way. There is also the alpha value applied and with that it is not a strong color.

PR #6894, opened 2019-11-28
Implement error boundary for widgets.

dennisoelkers:
## Motivation and Context

Prior to this change, a malfunctioning widget implementation (throwing
an error intentionally or accidentally) would lead to the general error
page and the search/dashboard would not be usable.

This change implements an error boundary in the `Widget` component, that
catches all errors thrown by components below it in the component tree
and displays them. This allows the user to operate on the
search/dashboard to fix the issue and continues showing other, working
widgets.

Before:
Screen Shot 2019-11-28 at 17 24 53

After:
Screen Shot 2019-11-28 at 17 24 05
Screen Shot 2019-11-28 at 17 24 16

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/widgets/WidgetFailed.jsx
```diff
@@ -0,0 +1,23 @@
...
+const WidgetFailed = ({ error }: Props) => (
+  <div className={styles.spinnerContainer}>
+    <Icon name="exclamation-triangle" size="3x" className={styles.iconMargin} />
+    <span>
+      <strong>While rendering this widget, the following error occurred:</strong>
+
+      <p>{error.toString()}</p>
+    </span>
+  </div>
+);
```
linuspahl: We could use the `ErrorWidget` component for the Icon and text style here. It has the same layout, but the component would need some adjustments, like a prop to provide a custom headline.

PR #6897, opened 2019-11-28
Add support for lookup table backed parameters

mpfz0r:
Using lookup table parameters in event definition queries
allows us to create events in case a condition matches a list
that is fetched via lookup tables.

 - Split Parameters into two subtypes:
   ValueParameter and LookupTableParameter.

 - Allow Parameters without a Binding, since LookupTableParameters don't require one.

 - Add queryParameters to AggregationEventProcessorConfig

 - If the LUT does return an empty list and there is no default parameter value defined,
   just skip running the query and assume that we received no results.
   This avoids running queries with an invalid ES query string.

 - Add frontend code to create parameters for an event defintion

cf https://github.com/Graylog2/graylog-plugin-enterprise/issues/1195

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/search/MoreSearch.java
+++ b/graylog2-server/src/main/java/org/graylog/events/search/MoreSearch.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
+import com.google.common.collect.ImmutableSet;
 import io.searchbox.client.JestClient;
 import io.searchbox.core.Search;
 import io.searchbox.core.SearchResult;
@@ -31,7 +32,14 @@ import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.graylog.events.event.EventDto;
 import org.graylog.events.processor.EventProcessorException;
+import org.graylog.plugins.views.search.Parameter;
+import org.graylog.plugins.views.search.Query;
+import org.graylog.plugins.views.search.SearchJob;
+import org.graylog.plugins.views.search.elasticsearch.ESQueryDecorators;
+import org.graylog.plugins.views.search.elasticsearch.ElasticsearchQueryString;
 import org.graylog.plugins.views.search.elasticsearch.IndexRangeContainsOneOfStreams;
+import org.graylog.plugins.views.search.errors.EmptyParameterError;
+import org.graylog.plugins.views.search.errors.SearchException;
 import org.graylog2.Configuration;
 import org.graylog2.database.NotFoundException;
 import org.graylog2.indexer.IndexHelper;
@@ -85,6 +93,7 @@ public class MoreSearch extends Searches {
     private final ScrollResult.Factory scrollResultFactory;
     private final JestClient jestClient;
     private final boolean allowLeadingWildcardSearches;
+    private final ESQueryDecorators esQueryDecorators;
 
     @Inject
     public MoreSearch(StreamService streamService,
@@ -94,13 +103,15 @@ public class MoreSearch extends Searches {
                       MetricRegistry metricRegistry,
                       ScrollResult.Factory scrollResultFactory,
                       JestClient jestClient,
-                      Configuration configuration) {
+                      Configuration configuration,
+                      ESQueryDecorators esQueryDecorators) {
         super(configuration, indexRangeService, metricRegistry, streamService, indices, indexSetRegistry, jestClient, scrollResultFactory);
         this.streamService = streamService;
         this.indexRangeService = indexRangeService;
         this.scrollResultFactory = scrollResultFactory;
         this.jestClient = jestClient;
         this.allowLeadingWildcardSearches = configuration.isAllowLeadingWildcardSearches();
+        this.esQueryDecorators = esQueryDecorators;
     }
 
     /**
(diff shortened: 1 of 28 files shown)


graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
```diff
@@ -41,6 +53,45 @@ class FilterForm extends React.Component {
...
+        offset: 0,
+      }]);
+
+    const query = queryBuilder.build();
+
+    const search = Search.create().toBuilder()
+      .queries([query])
+      .build();
+
+    SearchMetadataStore.parseSearch(search).then((res) => {
```
mpfz0r: use action instead of calling store directly

graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
```diff
@@ -85,6 +196,38 @@ class FilterForm extends React.Component {
     };
   };
 
+  renderQueryParameters = () => {
+    const { eventDefinition, onChange, lookupTables, validation } = this.props;
+    const { query_parameters: queryParameters } = eventDefinition.config;
+    const debug = (key, val) => {
```
mpfz0r: DEBUG

graylog2-server/src/main/java/org/graylog/events/search/MoreSearch.java
```diff
@@ -298,8 +319,21 @@ public void scrollQuery(String queryString, Set<String> streams, TimeRange timeR
         return streams;
     }
 
+    private String decorateQuery(Set<Parameter> queryParameters, TimeRange timeRange, String queryString) {
+        org.graylog.plugins.views.search.Search search = org.graylog.plugins.views.search.Search.builder()
```
bernd: Can you please document why we have to use search job and query dummy objects here and that we probably want to refactor the ES query decorator do allow easier access to parameter substitution. Thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/ViewsBindings.java
```diff
@@ -187,13 +189,16 @@ private void registerViewSharingSubtypes() {
...
+    }
+
     private void registerSharingStrategies() {
         registerSharingStrategy(AllUsersOfInstance.TYPE, AllUsersOfInstanceStrategy.class);
         registerSharingStrategy(SpecificRoles.TYPE, SpecificRolesStrategy.class);
         registerSharingStrategy(SpecificUsers.TYPE, SpecificUsersStrategy.class);
     }
-
-    private void registerExceptionMappers() {
+private void registerExceptionMappers() {
```
bernd: This looks like a formatting issue.

graylog2-server/src/main/java/org/graylog/plugins/views/search/ValueParameter.java
```diff
@@ -0,0 +1,126 @@
...
+@JsonTypeName(ValueParameter.TYPE_NAME)
+@JsonDeserialize(builder = ValueParameter.Builder.class)
+public abstract class ValueParameter implements Parameter {
+    public static final String TYPE_NAME = "value-parameter-v1";
+
+    @Override
+    public abstract String type();
+
+    @JsonProperty
+    public abstract String name();
```
bernd: You don't have to include the inherited fields from `Parameter` again.

graylog2-server/src/main/java/org/graylog/plugins/views/search/ValueParameter.java
```diff
@@ -0,0 +1,126 @@
...
+
+        final Binding binding = objectMapper.convertValue(bindingState, Binding.class);
+
+        return toBuilder().binding(binding).build();
+    }
+
+    @AutoValue.Builder
+    public abstract static class Builder implements Parameter.Builder<Builder> {
+        @JsonProperty
+        public abstract Builder name(String name);
```
bernd: You can move shared builder methods to `Parameter.Builder` to avoid having to include them in every parameter auto-value object.

graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/ParameterExpansionError.java
```diff
@@ -0,0 +1,62 @@
...
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * This file is part of Graylog.
```
bernd: Duplicate license.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/lookup/LookupTableResource.java
```diff
@@ -153,12 +155,14 @@ public LookupTableResource(DBLookupTableService dbTableService,
                                DBCacheService dbCacheService,
                                Map<String, LookupCache.Factory> cacheTypes,
                                Map<String, LookupDataAdapter.Factory> dataAdapterTypes,
+                               Map<String, LookupDataAdapter.Factory2> dataAdapterTypes2,
```
bernd: Can you please rebase or merge master? This change is in master already.

graylog2-server/src/main/java/org/graylog2/shared/utilities/ExceptionUtils.java
```diff
@@ -50,7 +50,7 @@ public static String formatMessageCause(Throwable t) {
             causeMessage.append(message);
         }
 
-        if (message != null && !message.endsWith(".")) {
+        if (message != null && !message.endsWith(".") && !message.endsWith("!")) {
```
bernd: This has been fixed in https://github.com/Graylog2/graylog2-server/pull/6958/files already. Please rebase or merge master. Thanks!

bernd:
@mpfz0r I just approved the backend changes. Please add the warning about the 1024 items limit to the UI before merging. Thanks!

PR #6903, opened 2019-11-29
Implement MessageList pagination

linuspahl:
Some details for the implementation. We are:
-  replacing the `MessageTablePaginator` with the common `PaginatedList` component inside the MessageList component. When we are removing the old search, we can remove the `MessageTablePaginator` completely.
- refactoring the PaginatedList component
- creating a separate component (`MessageTable`) to abstract some logic from the `MessageList` component.
- implementing another SearchStore execute function (`reexecuteSearchTypes`) to reexecute the search for some specific search types. The implementation is able to reexecute multiple search types, but we are only doing it for one search type, when using the message list pagination. The only thing that does not work so far is providing an effective time ranges for each search type.
- sending the elastic search result window limit to the frontend to display a useful message for the user. By default elasticsearch has a limit of 10000 messages, when the user access page 500 with a page size of 150 messages, he will see a related error message the error message also contains an info which pages are working for his configuration.
- adjusting the general widget layout. It's now more dynamic and e.g. the header does not need a fixed height.

With the next changes I will create smaller PR's by creating separate ones for e.g. the `PageList` refactoring.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/SearchTypeError.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/SearchTypeError.java
@@ -18,27 +18,54 @@ package org.graylog.plugins.views.search.errors;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.graylog.plugins.views.search.Query;
-import org.graylog.plugins.views.search.Query;
 
 import javax.annotation.Nonnull;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class SearchTypeError extends QueryError {
     @Nonnull
     private final String searchTypeId;
 
+    private final Integer resultWindowLimit;
+
     public SearchTypeError(@Nonnull Query query, @Nonnull String searchTypeId, Throwable throwable) {
         super(query, throwable);
+
+        this.resultWindowLimit = parseResultLimit(throwable);
+
         this.searchTypeId = searchTypeId;
     }
 
     public SearchTypeError(@Nonnull Query query, @Nonnull String searchTypeId, String description) {
         super(query, description);
+
+        this.resultWindowLimit = parseResultLimit(description);
+
         this.searchTypeId = searchTypeId;
+    }
+
+    private Integer parseResultLimit(Throwable throwable) {
+        return parseResultLimit(throwable.getMessage());
+    }
+
+    private Integer parseResultLimit(String description) {
+        if (description.toLowerCase().contains("result window is too large")) {
+            final Matcher matcher = Pattern.compile("[0-9]+").matcher(description);
+            if (matcher.find())
+                return Integer.parseInt(matcher.group(0));
+        }
+        return null;
     }
 
     @Nonnull
     @JsonProperty("search_type_id")
     public String searchTypeId() {
         return searchTypeId;
+    }
+
+    @JsonProperty("result_window_limit")
+    public Integer getResultWindowLimit() {
+        return resultWindowLimit;
     }
 }
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/SearchTypeError.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/SearchTypeError.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.graylog.plugins.views.search.Query;
 
 import javax.annotation.Nonnull;
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -50,7 +51,7 @@ public class SearchTypeError extends QueryError {
     }
 
     private Integer parseResultLimit(String description) {
-        if (description.toLowerCase().contains("result window is too large")) {
+        if (description.toLowerCase(Locale.US).contains("result window is too large")) {
             final Matcher matcher = Pattern.compile("[0-9]+").matcher(description);
             if (matcher.find())
                 return Integer.parseInt(matcher.group(0));
--- a/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
+++ b/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
@@ -89,8 +89,15 @@ class MessageList extends React.Component<Props, State> {
   componentDidMount() {
     const onRenderComplete = this.context;
     InputsActions.list().then(() => (onRenderComplete && onRenderComplete()));
+    SearchActions.execute.completed.listen(this._resetPagination);
   }
 
+  _resetPagination = () => {
+    const { currentPage } = this.state;
+    if (currentPage !== 1) {
+      this.setState({ currentPage: 1 });
+    }
+  }
 
   _resultWindowLimitMessage = (errors = []) => {
     const { pageSize } = this.props;
--- a/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
+++ b/graylog2-web-interface/src/views/components/widgets/MessageList.jsx
@@ -148,7 +148,6 @@ class MessageList extends React.Component<Props, State> {
     const totalAmount = (data && data.total) || 0;
     const { currentPage, expandedMessages } = this.state;
     const messageSlice = messages
-      .slice((currentPage - 1) * pageSize, currentPage * pageSize)
       .map((m) => {
         return {
           fields: m.message,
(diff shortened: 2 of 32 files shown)


graylog2-web-interface/src/views/components/widgets/MessageTable.jsx
```diff
@@ -0,0 +1,216 @@
...
+  border-collapse: collapse;
+  padding-left: 13px;
+  width: 100%;
+  word-break: break-all;
+
+  table.messages td, table.messages th {
+    position: relative;
+    left: 13px;
+  }
+
```
linuspahl: All these styles are part of the `graylog2.less` file and children of the selector `.messages`. The next step is to implement the `tr` styles to the MassageTableEntry component as styled components. I don't want to blow up this PR even more, but will gladly do these changes afterwards.

graylog2-server/src/main/java/org/graylog/plugins/views/search/errors/SearchTypeError.java
```diff
@@ -18,27 +18,54 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.graylog.plugins.views.search.Query;
-import org.graylog.plugins.views.search.Query;
 
 import javax.annotation.Nonnull;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
```
alex-konn: Note to the reviewer(s): I was torn between adding the special case to the existing `SearchTypeError` and adding a new kind of error. i went with adding it in, because I did not want to complicate `ElasticsearchBackend` with detecting the special case and constructing an instance of an alternative error class.

graylog2-web-interface/src/views/logic/search/SearchExecutionState.js
```diff
@@ -6,6 +6,13 @@ import type { QueryString, TimeRange } from '../queries/Query';
 export type GlobalOverride = {
   timerange?: TimeRange,
   query?: QueryString,
+  keep_search_types?: string[],
+  search_types?: {
+    [searchTypeId: string]: {
+      limit: number;
+      offset: number;
+    };
+  }
```
linuspahl: Two questions regarding this adjustment of the GlobalOverride type:

- Currently limit and offset can only be used for a message search type. What are the best practices for this case? Should I create a union type?
- I would prefer camelCase for the attribute names. We already talked about he implementation `.toJSON`. So far this only works for top level attributes. How do we deal with these encapsulated attributes? Should I adjust them inside the default `.toJSON()` method as well?

graylog2-web-interface/src/views/components/widgets/MessageList.jsx
```diff
@@ -194,4 +159,11 @@ export default connect(MessageList,
   {
     selectedFields: SelectedFieldsStore,
     currentView: ViewStore,
-  });
+    searches: SearchStore,
+  }, props => Object.assign(
+    {},
+    props,
+    {
+      effectiveTimerange: props.searches.result.results[props.currentView.activeQuery].effectiveTimerange,
```
linuspahl: We need the effective timerang range to "freeze" the search when using the pagination.
I would like to implement a cleaner solution for the above row. I thought about extracting the effective timerange in the ExtendedSearchPage, because we are already connecting the SearchStore there. But I am not a fan of passing the attribute through all the components in between.

I also thought about passing the full results objects to the MessageList component to extract the attribute there in a safer way. but I prefer to pass only the needed attributes to the component.

Anyone has some thoughts on this?

graylog2-web-interface/src/views/logic/SearchResult.js
```diff
@@ -1,18 +1,43 @@
...
+    this.results = mapValues(result.results, queryResult => new QueryResult(queryResult));
+    this.errors = get(result, 'errors', []).map(error => new SearchError(error));
   }
 
   forId(queryId) {
     return this.results[queryId];
   }
+
+  updateSearchTypes(searchTypeResults) {
+    const updatedResult = new SearchResult(this.result).result;
```
linuspahl: In a previous version https://github.com/Graylog2/graylog2-server/pull/6903/commits/04fce92df75f52884f94d9efa92694d645638ab0
we just used `this.result` for `updatedResult`.
Because we are adjusting the object in line 21, the old implementation would mutate the "current version" of this.result, right?
It feels a bit strange to run `new SearchResult` twice.

graylog2-web-interface/src/views/logic/SearchResult.js
```diff
@@ -1,18 +1,43 @@
...
+    });
+    return new SearchResult(updatedResult);
+  }
+
+  getSearchTypesFromResponse(searchTypeIds) {
+    const searchTypes = searchTypeIds.map((searchTypeId) => {
+      const relatedQuery = this._getQueryBySearchTypeId(searchTypeId);
+      return SearchResult._getSearchTypeFromQuery(relatedQuery, searchTypeId);
+    });
+    return compact(searchTypes);
```
linuspahl: When there is an error due to the elastic search result window limit, the array of search types, created in line 27 will contain an `undefined`, because the api response does not contain the search type anymore, but the id is still defined in the `searchTypeIds` props. Right now we are just removing the undefined entires, to prevent any further errors.

In them message list we are showing an error message, for this case. My question is how to deal with this case properly. Is it okay to "ignore" the error here, because we are handling this case in the MessageList?

graylog2-web-interface/src/views/components/widgets/MessageList.jsx
```diff
@@ -51,141 +63,94 @@ class MessageList extends React.Component<Props, State> {
...
+    SearchActions.reexecuteSearchTypes(searchTypePayload, effectiveTimerange).then((response) => {
+      let errors = [...response.result.errors];
+      if (!isEmpty(errors)) {
+        const validPagesInfo = this._resultWindowLimitMessage(response.result.errors);
+        if (validPagesInfo) {
+          errors = [
+            validPagesInfo,
+            ...errors,
+          ];
+        }
```
linuspahl: Is it okay to just add a custom error, which has only a description? We are only doing this to inform the user about the working / valid pages.

graylog2-web-interface/src/views/logic/SearchError.js
```diff
@@ -21,3 +21,30 @@ export default class SearchError {
 
   get type() { return this._state.type; }
 }
+
+export class ResultWindowLimitError extends SearchError {
```
linuspahl: This is the approach to move the error logic from MessageList component to the `SearchError` class. The goal is extending the description if the error is a result window limit error. I was not so sure about, where to create the different errors (currently in the SearchResult constructor) and how to deal with the prop result. Should it be a part of the class? Right now it is not a part of the class and I am only accessing it when constructing the class. That's why I am using static classes to extend the description.

graylog2-web-interface/src/views/logic/SearchResult.js
```diff
@@ -1,17 +1,60 @@
...
 class SearchResult {
   constructor(result) {
-    this.result = result;
+    this._result = fromJS(result);
 
-    this.results = _.mapValues(result.results, queryResult => new QueryResult(queryResult));
-    this.errors = _.get(result, 'errors', []).map(error => new SearchError(error));
+    this._results = fromJS(mapValues(result.results, queryResult => new QueryResult(queryResult)));
+    this._errors = fromJS(get(result, 'errors', []).map((error) => {
+      if (error.result_window_limit) {
```
kmerz: To distinguish a class on an attribute (other then type) is usually not a good idea.

Is there a reason why `ResultWindowLimitError` is not its own `SearchError` implementation?

But I do not have to strong feelings here.

graylog2-web-interface/src/views/logic/SearchResult.js
kmerz: Create here a function `filterFailedSearchTypes` where you do the `compat` then it is clear what you do. The place seems right to me.

graylog2-web-interface/src/views/logic/SearchResult.js
linuspahl: I would also prefer a different approach here. The only idea I have so far is adding a create method for the `SearchError` class to decide which error to create. But I would still just check the attribute inside this method.

linuspahl: We've added the `result_window_limit` type for the related error class in the backend, to identify which class to use in the frontend. After talking with @dennisoelkers I've implemented a static create method for the `SearchError` class, to decide which error class we should use.

But I received an error due to the dependency cycle. I can't import the `SearchError` class inside the `ResultWindowLimit` file, while importing the `ResultWIndowLimit` class inside the `SearchError` file. So far I've just used the type attribute to identify the related class inside the above map.

PR #6925, opened 2019-12-04
Migrating Saved Searches to Views.

### LINKED ISSUE #6395: Migrate existing saved searches to Views

dennisoelkers:
## Description

This PR is adding a migration that takes all existing saved searches and
migrates them to views-based searches. Those do include the query,
timerange, title and field selection of the migrated saved search. A
default set of widgets is added to the new search: a message count histogram
and a message list with the field selection of the saved search.

Fixes #6395.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20191203120602_MigrateSavedSearchesToViewsSupport/search/Time.java
```diff
@@ -0,0 +1,38 @@
...
+package org.graylog.plugins.views.migrations.V20191203120602_MigrateSavedSearchesToViewsSupport.search;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class Time {
+    public static final String NAME = "time";
+
+    public String type() { return NAME; }
```
alex-konn: Missing `@JsonProperty`. Without the `type`property the migrated `Search` can't be executed.

graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20191203120602_MigrateSavedSearchesToViewsSupport/V20191203120602_MigrateSavedSearchesToViewsTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Test
+    public void writesMigrationCompletedAfterSuccess() {
+        this.migration.upgrade();
+
+        final MigrationCompleted migrationCompleted = captureMigrationCompleted();
+        assertThat(migrationCompleted.savedSearchIds()).isEmpty();
+    }
+
+    @Test
```
alex-konn: I would prefer writing different tests for the different kinds of saved searches instead of one big test. I had to test all three timerange variants manually too.

dennisoelkers: ✅

alex-konn: Nice! Is there a special reason for not deduplicating the code by e.g. extracting a method that takes the string identifiers for fixtures and entities as parameters?
One parameterized test might also work, if we have an alternative way to load the fixture than the `MongoDBFixtures` annotation.

dennisoelkers: There is no special reason. Deduplicating the code would make sense if we can expect to change the code afterwards, which I do not for a one-off migration.

alex-konn: Ok, fair enough.

PR #6955, opened 2019-12-09
Improve API browser for several resources

mpfz0r:
Most event resources were either missing `@ApiParam` annotations
or their models were not annotated with `@JsonAutoDetect`.

Both are needed to show meaningful model data in swagger.

Also add the word `legacy` to the old alarm callback resources, so API users are less confused.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/notifications/NotificationDto.java
+++ b/graylog2-server/src/main/java/org/graylog/events/notifications/NotificationDto.java
@@ -16,6 +16,7 @@
  */
 package org.graylog.events.notifications;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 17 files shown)


bernd:
> Most event resources were either missing `@ApiParam` annotations
> or their models were not annotated with `@JsonAutoDetect`.

@mpfz0r We have to check every resource that didn't have the `@JsonAutoDetect` annotation before for changes. This might result in added fields in the JSON response for fields/methods that haven't been annotated with `@JsonProperty` before.

mpfz0r:
> > Most event resources were either missing `@ApiParam` annotations
> > or their models were not annotated with `@JsonAutoDetect`.
> 
> @mpfz0r We have to check every resource (and also database collection DTO) that didn't have the `@JsonAutoDetect` annotation before for changes. This might result in added fields in the JSON response for fields/methods that haven't been annotated with `@JsonProperty` before. (and add `@JsonIgnore` annotations for those)

Thanks for the heads up. I'll check that.

mpfz0r:
> > > Most event resources were either missing `@ApiParam` annotations
> > > or their models were not annotated with `@JsonAutoDetect`.
> > 
> > 
> > @mpfz0r We have to check every resource (and also database collection DTO) that didn't have the `@JsonAutoDetect` annotation before for changes. This might result in added fields in the JSON response for fields/methods that haven't been annotated with `@JsonProperty` before. (and add `@JsonIgnore` annotations for those)
> 
> Thanks for the heads up. I'll check that.

Checked all models, don't see a difference

PR #6964, opened 2019-12-10
Migrate search page field charts

linuspahl:
These changes will display an alert for the search page, to allow the user to migrate the field charts, created for the old search page version.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/bindings.jsx
+++ b/graylog2-web-interface/src/views/bindings.jsx
@@ -1,4 +1,5 @@
 // @flow strict
+import React from 'react';
 import Routes from 'routing/Routes';
 import * as Permissions from 'views/Permissions';
 
@@ -20,6 +21,7 @@ import PieVisualization from 'views/components/visualizations/pie/PieVisualizati
 import ScatterVisualization from 'views/components/visualizations/scatter/ScatterVisualization';
 import WorldMapVisualization from 'views/components/visualizations/worldmap/WorldMapVisualization';
 import HeatmapVisualization from 'views/components/visualizations/heatmap/HeatmapVisualization';
+import MigrateFieldCharts from 'views/components/MigrateFieldCharts';
 
 import PivotConfigGenerator from 'views/logic/searchtypes/aggregation/PivotConfigGenerator';
 import PivotHandler from 'views/logic/searchtypes/pivot/PivotHandler';
@@ -45,6 +47,7 @@ import RemoveFromTableActionHandler from 'views/logic/fieldactions/RemoveFromTab
 import RemoveFromAllTablesActionHandler from 'views/logic/fieldactions/RemoveFromAllTablesActionHandler';
 import CreateCustomAggregation from 'views/logic/creatoractions/CreateCustomAggregation';
 import SelectExtractorType from 'views/logic/valueactions/SelectExtractorType';
+import Store from 'logic/local-storage/Store';
 
 import VisualizationConfig from 'views/logic/aggregationbuilder/visualizations/VisualizationConfig';
 import WorldMapVisualizationConfig from 'views/logic/aggregationbuilder/visualizations/WorldMapVisualizationConfig';
@@ -124,6 +127,8 @@ const searchPages = enableNewSearch
     search: { component: NewSearchPage },
   }
   : {};
+
+const hasLegacyFieldCharts = () => !Store.get('pinned-field-charts-migrated') && Store.get('pinned-field-charts');
 
 export default {
   pages: {
@@ -357,4 +362,7 @@ export default {
   'views.hooks.loadingView': [
     requirementsProvided,
   ],
+  'views.elements.header': [() => hasLegacyFieldCharts() && (
+    <MigrateFieldCharts />
+  )],
 };
--- a/graylog2-web-interface/src/views/bindings.jsx
+++ b/graylog2-web-interface/src/views/bindings.jsx
@@ -128,7 +128,7 @@ const searchPages = enableNewSearch
   }
   : {};
 
-const hasLegacyFieldCharts = () => !Store.get('pinned-field-charts-migrated') && Store.get('pinned-field-charts');
+const hasLegacyFieldCharts = () => !Store.get('pinned-field-charts-migrated') && !!Store.get('pinned-field-charts');
 
 export default {
   pages: {
--- a/graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
+++ b/graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
@@ -22,6 +22,10 @@ import type { InterpolationMode } from 'views/logic/aggregationbuilder/visualiza
 import { Alert, Button, Row, Col } from 'components/graylog';
 import Spinner from 'components/common/Spinner';
 
+// localStorage keys
+const FIELD_CHARTS_KEY = 'pinned-field-charts';
+const FIELD_CHARTS_MIGRATED_KEY = 'pinned-field-charts-migrated';
+
 type LegacySeries = 'mean' | 'max' | 'min' | 'total' | 'count' | 'cardinality';
 type LegacyInterpolation = 'linear' | 'step-after' | 'basis' | 'bundle' | 'cardinal' | 'monotone';
 type LegacyInterval = 'minute' | 'hour' | 'day' | 'week' | 'month' | 'quarter' | 'year'
@@ -152,7 +156,7 @@ const _onMigrate = (legacyCharts: Array<LegacyFieldChart>, setMigrating: boolean
 
   ViewStatesActions.update(currentView.activeQuery, newViewState).then(
     () => SearchActions.executeWithCurrentState().then(() => {
-      Store.set('pinned-field-charts-migrated', true);
+      Store.set(FIELD_CHARTS_MIGRATED_KEY, true);
       setMigrating(false);
       setMigrationFinished(true);
     }),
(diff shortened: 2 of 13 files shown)


graylog2-web-interface/src/views/bindings.jsx
```diff
@@ -344,4 +349,7 @@ export default {
   'views.hooks.loadingView': [
     requirementsProvided,
   ],
+  'views.elements.header': [() => hasLegacyFieldCharts() && (
+    <MigrateFieldCharts />
+  )],
```
linuspahl: I want to display the `<MigrateFieldCharts />` inside the `<HeaderElements />` component and thought it is enough to add this section, but it never gets executed. Did I forgot something?

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,189 @@
...
+    const widgetRow = lastWidgetPosition.row + lastWidgetPosition.height + index + widgetDef.defaultHeight;
+    newWidgetPositions[newWidget.id] = new WidgetPosition(1, widgetRow, widgetDef.defaultHeight, Infinity);
+    return newWidget;
+  });
+
+  const newViewState = currentView.state
+    .toBuilder()
+    .widgets(Immutable.List([...currentView.state.widgets, ...newWidgets]))
+    .widgetPositions(newWidgetPositions)
+    .build();
```
linuspahl: When writing the tests I mocked `CurrentViewStateStore.getInitialState` to return an object with widgets and widget positions. This way I can't use currentView.state.toBuilder() because it is not a function. It is also not possible to use `ViewState.create` inside the mock. Does this mean I have to create the current view + its state manually before running the tests? Or can you think of a better way?

graylog2-web-interface/src/views/components/visualizations/XYPlot.jsx
```diff
@@ -38,7 +38,7 @@ const XYPlot = ({
   chartData,
   currentQuery,
   timezone,
-  effectiveTimerange,
+  effectiveTimerange = {},
```
linuspahl: With the migration we are adding the new widgets to the current view and fetching the search result afterwards.
Because the widgets are visible, before the search got executed, they are throwing a few errors, because of the missing result data object. I thought about passing a function to the `MigrateFieldCharts` component to hide all widgets before finishing the search execution. Another solution is to define the related attributes as optional in the `XYPlot.jsx` an `TransformKeys.js`. Which way do you prefer?

linuspahl: Answer: I will try to show the loading spinner for the `SearchResult` component, if it's not a suitable solution, I will try the second way.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
linuspahl: Answer: With `useMock` we can return the `ViewState` inside the `CurrentViewStateStore.getInitialState` mock and don't have to create the full View initally.

graylog2-web-interface/src/views/bindings.jsx
linuspahl: Thanks to @kmerz I figured out the `HeaderElements` component gets called before we are registering the `ViewBindings` inside the `index.jsx`, it looks like it's caused by the script import order. I am searching for a solution.

linuspahl: Answer: Not caused by script order, but the way we were importing the ExtendedSearchPage

graylog2-web-interface/src/views/components/__tests__/MigrateFieldCharts.fixtures.js
```diff
@@ -0,0 +1,53 @@
+import Immutable from 'immutable';
+
+import ViewState from 'views/logic/views/ViewState';
+import AggregationWidget from 'views/logic/aggregationbuilder/AggregationWidget';
+import AggregationWidgetConfig from 'views/logic/aggregationbuilder/AggregationWidgetConfig';
+import WidgetPosition from 'views/logic/widgets/WidgetPosition';
+
+// local storage mock for pinned-field-charts
+const mockFieldCharts = ({
```
linuspahl: Maybe this file is not the right place for this mock, because it is not fix?

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+  interval: LegacyInterval,
+}
+
+const Actions = styled.div`
+  margin-top: 10px;
+`;
+
+const mapSeries = (legacySeries: LegacySeries, field: string) => {
+  let seriesName;
+  switch (legacySeries) {
```
dennisoelkers: I would suggest extracting the `switch`-statement into a separate `mapFunction` fn. The benefit is that you can do a simple `return` in each `case`-branch and do not have to assign a variable, making the whole construct easier to understand. E.g. for `total`:
```
  case 'total':
    return 'sum';
```

`mapSeries` would then just look like this: `(legacySeries, field) => `${mapFunction(legacySeries)}(${field});`

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+    case 'quarter':
+      return { unit: 'months', value: 3 };
+    default:
+      return { unit: `${legacyTime}s`, value: 1 };
+  }
+};
+
+const createVisualizationConfig = (legacyInterpolation: LegacyInterpolation, visualization: string) => {
+  let interpolation: InterpolationMode;
+  switch (legacyInterpolation) {
```
dennisoelkers: Same here, you could extract a `mapInterpolation` function here.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+      interpolation = 'linear';
+  }
+
+  switch (visualization) {
+    case 'line':
+      return new LineVisualizationConfig(interpolation);
+    case 'area':
+      return new AreaVisualizationConfig(interpolation);
+    default:
+      return undefined;
```
dennisoelkers: Maybe you should `throw` an error here, clearly something is contained in the legacy field chart that we do not expect.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+  switch (legacyInterpolation) {
+    case 'basis':
+    case 'bundle':
+    case 'cardinal':
+    case 'monotone':
+      interpolation = 'spline';
+      break;
+    case 'linear':
+    case 'step-after':
+      interpolation = legacyInterpolation;
```
dennisoelkers: After extracting `mapInterpolation` you could just return `legacyInterpolation` for both cases here.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+      break;
+    default:
+      interpolation = 'linear';
```
dennisoelkers: In this case it might make sense to `throw` an error instead.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+  }
+};
+
+const _initialRowOffset = (widgetPositions) => {
+  const lastWidgetPosition = maxBy(values(widgetPositions), (position: WidgetPosition): number => position.row);
+  return lastWidgetPosition ? (lastWidgetPosition.row + lastWidgetPosition.height) : 0;
+};
+
+const _onMigrate = (legacyCharts: Array<LegacyFieldChart>, setMigrating: boolean => void, setMigrationFinished: boolean => void) => {
+  setMigrating(true);
```
dennisoelkers: We could make the API simpler by just returning a `Promise` that wraps the migration code and is chained with `ViewStatesActions.update`, instead of passing `setMigrating`/`setMigrationFinished`.

The `onClick` handler would be separated from the actual migration code. It would call `setMigrating` when starting and `setMigrationFinished` after the promise was resolved, but the migration code itself would not have to take care of updating the UI.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const newViewState = currentView.state
+    .toBuilder()
+    .widgets(Immutable.List([...currentView.state.widgets, ...newWidgets]))
+    .widgetPositions(newWidgetPositions)
+    .build();
+
+  ViewStatesActions.update(currentView.activeQuery, newViewState).then(
+    () => SearchActions.executeWithCurrentState().then(() => {
+      Store.set('pinned-field-charts-migrated', true);
```
dennisoelkers: Field names like `pinned-field-charts` & `pinned-field-charts-migrated` should be extracted to constants in this module.

graylog2-web-interface/src/views/components/__tests__/MigrateFieldCharts.fixtures.js
dennisoelkers: I think this is fine. If you do not feel it is the right place, I would be fine with moving it to the test file itself as well.

graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
linuspahl: I've added this check for `mapVisualization()`, the function will throw an error for unsupported visualizations. In the the context of the visualization config, we need to return undefined, for every visualization, which is not a line or bar chart.

PR #6995, opened 2019-12-13
Support existing search URLs in new search page

linuspahl:
This PR adds support for the existing search URL's in the new search page. We are using the parameters e.g. when clicking on "Show all Messages" in the inputs overview. Besides the query (parameter `q`) we support the time range, e.g.:
`?rangetype=absolute&from=[date]&to=[date]`
`?rangetype=keyword&keyword=[keyword]`
`?rangetype=relative&relative=0`
or just
`?relative=0`

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.js
+++ b/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.js
@@ -15,24 +15,28 @@ const _getTimerange = (query = {}) => {
   let timerange;
   const rangeType = query.rangetype || DEFAULT_RANGE_TYPE;
 
-  if (query.relative || query.absolute || query.keyword) {
-    switch (rangeType) {
-      case 'relative':
+  switch (rangeType) {
+    case 'relative':
+      if (query.relative) {
         timerange = { type: rangeType, range: query.relative };
-        break;
-      case 'absolute':
+      }
+      break;
+    case 'absolute':
+      if (query.from || query.to) {
         timerange = {
           type: rangeType,
           from: query.from,
           to: query.to,
         };
-        break;
-      case 'keyword':
-        timerange = { type: rangeType, keyword: query.relative };
-        break;
-      default:
-        throw new Error(`Unsupported range type ${rangeType}`);
-    }
+      }
+      break;
+    case 'keyword':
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.js
```diff
@@ -0,0 +1,69 @@
...
+    return Promise.resolve();
+  }
+  return QueriesActions.timerange(queryId, timeRangeParam);
+};
+
+const bindSearchParamsFromQuery: ViewHook = ({ query, view }: Arguments) => {
+  if (view.type !== 'SEARCH') {
+    return Promise.resolve(true);
+  }
+  const { id: queryId } = CurrentQueryStore.getInitialState();
```
linuspahl: I could also extract the id from the view instead of using `.getInitialState()`. What is the preferred way?

graylog2-web-interface/src/views/pages/NewSearchPage.jsx
```diff
@@ -44,7 +45,18 @@ class NewSearchPage extends React.Component<Props, State> {
...
+    const { query } = location;
+    processHooks(
+      ViewActions.create(View.Type.Search).then(({ view }) => view),
+      loadingViewHooks,
+      executingViewHooks,
+      query,
+    ).then(
+      () => this.setState({ loaded: true }),
+    ).catch(
+      error => UserNotification.error(`Executing search failed with error: ${error}`, 'Could not execute search'),
```
linuspahl: When the url params range type is `keyword` and the provided keyword is invalid, the search page does not render. On the old page we displayed an inline error message, but I think it's also okay to display this type of error message?

dennisoelkers: For now, yes.

linuspahl:
@dennisoelkers  we talked about implementing the `HookComponent` render part in the `StreamSearchPage` as well. While having another look, I saw we are dealing with this case differently.  E.g. `NewSearchPage` vs. `NewDashboardPage` In the context of the `StreamSearchPage` we should prefer the implementation used in `NewSearchPage` to set the hook component?

PR #6996, opened 2019-12-14
Scratchpad browser resize

kyleknighted:
## Description

- If the user resizes the browser, there is still a chance Scratchpad gets lost, this watches for resize and moves Scratchpad if necessary
- It remembers the opened or closed state now, so if you reload your Scratchpad comes back up too

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/InteractableModal.jsx
```diff
@@ -115,13 +113,26 @@ const InteractableModal = ({
...
-      y: newY,
+      x: newX < 0 ? 0 : newX,
+      y: newY < 0 ? 0 : newY,
     };
     setDragPosition(setPosition);
     onDrag(setPosition);
   };
 
+  const handleBrowserResize = debounce((event) => {
+    handleDragStop(event, { node: { style: { ...resizeSize } }, ...dragPosition });
```
edmundoa: `resizeSize` and `dragPosition` seem to be using stale state values here and may lead to some weird behaviour:

Kapture 2019-12-17 at 16 41 43

graylog2-web-interface/src/components/common/InteractableModal.jsx
```diff
@@ -90,56 +90,56 @@ const InteractableModal = ({
...
-  const handleBrowserResize = debounce((event) => {
-    handleDragStop(event, { node: { style: { ...resizeSize } }, ...dragPosition });
+    handleDragStop(null, { x: nextX, y: nextY });
   }, 150);
 
   useEffect(() => {
     setDragHandleClassName(dragHandleRef.current.classList[0]);
+  }, []);
+
+  useEffect(() => {
```
kyleknighted: Might not technically need two `useEffects`. I don't think re-running `setDragHandleClassName` will cause much performance issue. Not sure which direction you think makes the most sense

graylog2-web-interface/src/components/common/InteractableModal.jsx
```diff
@@ -10,8 +10,8 @@ import teinte from 'theme/teinte';
 import { debounce } from 'lodash';
 
 const DEFAULT_SIZE = { width: 450, height: 400 };
-const halfWidth = Math.ceil((document.body.offsetWidth / 2) - (DEFAULT_SIZE.width / 2));
-const halfHeight = Math.ceil((document.body.offsetHeight / 2) - (DEFAULT_SIZE.height / 2));
+const halfWidth = Math.ceil((window.innerWidth / 2) - (DEFAULT_SIZE.width / 2));
```
kyleknighted: helps using the correct width/height params :D

graylog2-web-interface/src/components/common/InteractableModal.jsx
edmundoa: For me it makes it clearer having both in different `useEffect`. Also that way is clearer why the dependency on `dragPosition`.

Issue #7009, opened 2019-12-18
Pinned Field Chart Migration should not be possible in Dashboard View


kmerz:
## Expected Behavior
When opening a Dashboard graylog should not ask me to migrate my
pinned-field-chart into the Dashboard.

## Current Behavior
![Graylog (34)](https://user-images.githubusercontent.com/448763/71069654-753e5480-2179-11ea-9518-cc61f23b0a63.png)

## Possible Solution
Check if current view is a dashboard.

## Steps to Reproduce (for bugs)
[...]

PR #7011, opened 2019-12-18
Only show field chart migration info on search page

linuspahl:
The field chart migration info should only be visible on the search page. It should not be visible on the dashboards.

Fixes https://github.com/Graylog2/graylog2-server/issues/7009

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/bindings.jsx
+++ b/graylog2-web-interface/src/views/bindings.jsx
@@ -22,6 +22,7 @@ import ScatterVisualization from 'views/components/visualizations/scatter/Scatte
 import WorldMapVisualization from 'views/components/visualizations/worldmap/WorldMapVisualization';
 import HeatmapVisualization from 'views/components/visualizations/heatmap/HeatmapVisualization';
 import MigrateFieldCharts from 'views/components/MigrateFieldCharts';
+import IfSearch from 'views/components/search/IfSearch';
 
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/components/MigrateFieldCharts.jsx
```diff
@@ -186,11 +188,15 @@ const _onCancel = (setMigrationFinished) => {
 };
 
 const MigrateFieldCharts = () => {
+  const viewType = useContext(ViewTypeContext);
+  if (viewType !== View.Type.Search) {
+    return null;
+  }
```
linuspahl: I thought about using the `<IfSearch>` component, but I want to avoid calling the localstorage on every dashboard page.

graylog2-web-interface/src/views/bindings.jsx
```diff
@@ -108,8 +108,6 @@ ViewSharing.registerSubtype(SpecificUsers.Type, SpecificUsers);
 Parameter.registerSubtype(ValueParameter.type, ValueParameter);
 Parameter.registerSubtype(LookupTableParameter.type, LookupTableParameter);
 
-const hasLegacyFieldCharts = () => !Store.get('pinned-field-charts-migrated') && !!Store.get('pinned-field-charts');
```
linuspahl: I removed this check, because we are doing the same check inside the component.

PR #7024, opened 2019-12-19
Add content pack to install Source Page as Dashboard

kmerz:
## Description
Add a content pack which will be installed via migration. When the migration is through
we set a marker cluster config to prevent new installations.

## Motivation and Context
The old Sources Page was not more then the old Search Page with predefined widgets.
We now provide a dashboard which has the predefined Widgets. This dashboard
will be installed via a content pack on first start.

## How Has This Been Tested?
- Run the migration and look that the Dashboard is installed.
- Run the migration again and look that not another Content Pack was installed.

Graylog (40)

Fixes: #6904

## Types of changes

## Checklist:


graylog2-server/src/test/java/org/graylog2/migrations/V20191219090834_AddSourcesPageTest.java
```diff
@@ -0,0 +1,145 @@
...
+                .comment("Comment")
+                .createdAt(Instant.now())
+                .createdBy("admin")
+                .build();
+        when(configService.get(V20191219090834_AddSourcesPage.MigrationCompleted.class)).thenReturn(null);
+        when(objectMapper.readValue(contentPackURL, ContentPack.class)).thenReturn(contentPack);
+        when(contentPackPersistenceService.insert(contentPack)).thenReturn(Optional.empty());
+        boolean exceptionCalled = false;
+        try {
+            migration.upgrade();
```
dennisoelkers: You can do `assertThatExceptionOfType(ContentPackException.class).isThrownBy(() -> migration.upgrade());` instead.

graylog2-server/src/main/java/org/graylog2/migrations/V20191219090834_AddSourcesPage.java
```diff
@@ -0,0 +1,104 @@
...
+            final ContentPack contentPack = this.objectMapper.readValue(contentPackURL, ContentPack.class);
+            final ContentPack pack = this.contentPackPersistenceService.insert(contentPack)
+                    .orElseThrow(() -> {
+                        configService.write(V20191219090834_AddSourcesPage.MigrationCompleted.create(contentPack.id().toString()));
+                        return new ContentPackException("Content pack " + contentPack.id() + " with this revision " + contentPack.revision() + " already found!");
+                    });
+
+            try {
+                contentPackService.installContentPack(pack, Collections.emptyMap(), "Add Sources Page", "admin");
+            } catch(ContentPackException e) {
```
dennisoelkers: Catching the exception _here_ would mean that the migration is marked as completed and could not be repeated after fixing the underlying error. Instead, we could catch it only level higher and rethrow a more helpful exception (together with the `IOException`) by adding a catch block for the `Exception` class, preventing the migration to be marked as completed.

Issue #7021, opened 2019-12-18
Field Action `Add To Table` is missing


kmerz:
## Expected Behavior
When open Field Action on a Message Table the Field Action should show `Add To Table` as action which should add the field to this table.

## Current Behavior
`Add To Table` is missing when clicking on a Field Action of a Message Table.

![Graylog (39)](https://user-images.githubusercontent.com/448763/71095129-60c77f80-21ac-11ea-94a4-9dc562918872.png)

## Your Environment

* Graylog Version: 3.2 dev (c9c2ee7393467652238811423a415206e969f9da)

PR #7030, opened 2019-12-19
Enable 'Add to Table' action for message list widget

linuspahl:
Fixes https://github.com/Graylog2/graylog2-server/issues/7021

The main problem was the (I guess) missing data structure change for the `isHidden` method.
Another problem was the implemented `instanceof` check. It works for the dashboards, but not for the search page. I could not identify the difference. Please let me know if the new type check is fine or if I should spend more time finding the mentioned difference.

This PR also fixed the same problems for the `RemoveFromTableActionHandler`.

## Types of changes

## Checklist:


dennisoelkers:
`instanceof` is generally unreliable, especially when it is used across asset/bundle boundaries, so replacing it with a different check is generally fine.

graylog2-web-interface/src/views/logic/fieldactions/AddToTableActionHandler.js
```diff
@@ -13,15 +10,18 @@ const AddToTableActionHandler: FieldActionHandler = ({ field, contexts: { widget
   return WidgetActions.updateConfig(widget.id, newConfig);
 };
 
-AddToTableActionHandler.isEnabled = ({ contexts: { widget }, field }) => {
-  if (widget instanceof MessagesWidget && widget.config instanceof MessagesWidgetConfig) {
+const isEnabled: FieldActionHandlerCondition = ({ contexts: { widget }, field }) => {
+  if (widget.type === 'messages' && widget.config) {
```
dennisoelkers: We could improve this in two ways:

  1.) Extract the check into a `isMessageWidget(widget)` function in a separate module and reuse it.
  2.) Use the `MessagesWidget.type` constant

The check generally is okay. The only concern that I have is that we _might_ have subclasses in the future with a different type forcing us to take care of refining the check. That is why I am suggesting to extract the function, to make it more visible.

linuspahl: Good point, I implemented this check inside `MessagesWidget` is it fine or do you have a better place in mind?

PR #7068, opened 2019-12-30
Update AppFacade to functional component

kyleknighted:
## Description

Updating deprecated createReactClass to functional React component 

## Motivation and Context

Every update to AppFacade causes a lint failure because of using `createReactClass`, this solves that

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/routing/AppFacade.jsx
+++ b/graylog2-web-interface/src/routing/AppFacade.jsx
@@ -1,8 +1,11 @@
-import React from 'react';
-import createReactClass from 'create-react-class';
-import Reflux from 'reflux';
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
 
 import loadAsync from 'routing/loadAsync';
+import ServerUnavailablePage from 'pages/ServerUnavailablePage';
+import StoreProvider from 'injection/StoreProvider';
+import connect from 'stores/connect';
+import GraylogThemeProvider from 'theme/GraylogThemeContext';
 
 import 'bootstrap/less/bootstrap.less';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/routing/AppFacade.jsx
```diff
@@ -24,38 +22,58 @@ const LoginPage = loadAsync(() => import(/* webpackChunkName: "LoginPage" */ 'pa
...
-      </GraylogThemeProvider>
-    );
-  },
-});
-
-export default AppFacade;
+const SERVER_PING_TIMEOUT = 20000;
+
+export const AppFacade = ({ currentUser, server, sessionId }) => {
+  let interval;
```
dennisoelkers: There is no reason to declare `interval` in this scope, we can limit it to the scope of the anonymous function passed to `useEffect` and declare it in line 32 instead.

graylog2-web-interface/src/routing/AppFacade.jsx
```diff
@@ -24,38 +22,58 @@ const LoginPage = loadAsync(() => import(/* webpackChunkName: "LoginPage" */ 'pa
...
+  let Page;
+
+  useEffect(() => {
+    interval = setInterval(ServerAvailabilityStore.ping, SERVER_PING_TIMEOUT);
+    return () => {
```
dennisoelkers: We can shorten this to

```suggestion
    return () => clearInterval(interval);
```

a) we know that `interval` has been assigned before, b) even if `setInterval` returns something invalid because of an error, `clearInterval` does not raise an error.

PR #7071, opened 2020-01-02
Handle Request Entity Too Large errors in ElasticSearchOutput

### LINKED ISSUE #5091: Graylog doesn't handle Elasticsearch "413 Request Entity Too Large" reply
One graylog node lost the connection with elastic cluster.
It seems sometimes, somehow the graylog make bigger batch (based on output_batch_size), then the elasticsearch can allow (ES bulk_max_body_size default 100mb). The graylog can't recognize the received error from ES.
After a graylog service restart can send the logs out to elasticsearch.

After time the GL lost the connection with all ES nodes, and can't store logs. Journal start to grow.
Issue details: https://community.graylog.org/t/elastic-cluster-connection-lost/6189/8

## Expected Behavior
GL create smaller size of batch then the ES limit, or handle the error, and refregment the batch if iit receive "413 Request Entity Too Large" error

## Current Behavior
GL makes a batch, try to send to ES, ES denies it, ES send TCP reset, GL try another ES with the same batch.

GL: log: 
```
2018-09-10T11:49:04.401+02:00 ERROR [Messages] Caught exception during bulk indexing: java.net.SocketException: Connection reset, retrying (attempt #28).
```

slice of TCPdump
```
12:05:45.204959 IP SUBNET.41.57818 > SUBNET.36.9200: Flags [P.], seq 18272:18503, ack 489, win 269, options [nop,nop,TS val 1985608753 ecr 1984697849], length 231
.......PV..d....E...p1@.@...dN.)dN.$..#.g.9..x@'...........
vY.1vL..POST /_bulk HTTP/1.1
Content-Length: 230363373
Content-Type: application/json; charset=UTF-8
Host: SUBNET.36:9200
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.4 (Java/1.8.0_181)
Accept-Encoding: gzip,deflate

12:05:45.205301 IP SUBNET.36.9200 > SUBNET.41.57818: Flags [P.], seq 489:549, ack 31535, win 752, options [nop,nop,TS val 1984716154 ecr 1985608753], length 60
.......PV.......E..p..@.@...dN.$dN.)#....x@'g.m.....md.....
vL]zvY.1HTTP/1.1 413 Request Entity Too Large
content-length: 0

12:05:45.205709 IP SUBNET.36.9200 > SUBNET.41.57818: Flags [R.], seq 848, ack 90903, win 1272, options [nop,nop,TS val 0 ecr 1985608754], length 0
.......PV.......E..4..@.@...dN.$dN.)#....xA.g.U{....bK.....
....vY.2{"index":{"_id":

12:05:45.967516 IP SUBNET.41.34322 > SUBNET.38.9200: Flags [P.], seq 95405:95636, ack 1034, win 245, options [nop,nop,TS val 1985609516 ecr 1983368835], length 231
.......PV..d....E.....@.@..TdN.)dN.&..#.....OW`............
vY.,v7..POST /_bulk HTTP/1.1
Content-Length: 230363373
Content-Type: application/json; charset=UTF-8
Host: SUBNET.38:9200
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.4 (Java/1.8.0_181)
Accept-Encoding: gzip,deflate

12:05:45.967917 IP SUBNET.38.9200 > SUBNET.41.34322: Flags [P.], seq 1034:1094, ack 108668, win 1432, options [nop,nop,TS val 1983383888 ecr 1985609516], length 60
.......PV.BS....E..p..@.@...dN.&dN.)#...OW`..........&.....
v8      PvY.,HTTP/1.1 413 Request Entity Too Large
content-length: 0

12:05:45.968792 IP SUBNET.38.9200 > SUBNET.41.34322: Flags [R], seq 1331126799, win 0, length 0
.......PV.BS....E..(..@.@.V3dN.&dN.)#...OWb.....P...{..........,v7............

12:05:46.708613 IP SUBNET.41.47776 > SUBNET.78.9200: Flags [P.], seq 33578:33809, ack 505, win 458, options [nop,nop,TS val 1985610257 ecr 1984247522], length 231
.......PV..d....E...AU@.@..udN.)dN.N..#..V.`ubS...... .....
vZ..vE6.POST /_bulk HTTP/1.1
Content-Length: 230363373
Content-Type: application/json; charset=UTF-8
Host: SUBNET.78:9200
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.4 (Java/1.8.0_181)
Accept-Encoding: gzip,deflate

12:05:46.709076 IP SUBNET.78.9200 > SUBNET.41.47776: Flags [P.], seq 505:565, ack 46841, win 6634, options [nop,nop,TS val 1984261327 ecr 1985610257], length 60
.......PV.3{....E..p..@.@...dN.NdN.)#...ubS..W./....eH.....
vEl.vZ..HTTP/1.1 413 Request Entity Too Large
content-length: 0
```
// it contains the GL uncommited log entries, sample every 5min
kep

## Possible Solution
Recreate batch if ES denies with error 413 Request Entity Too Large
Chech the bulk_max_body_size before communication and don't create bigger batch then the bulk_max_body_size
Or make a new variable in server.conf to set the max size of batch size (in KB or MB)

1. Set GL output_batch_size to higher (eg. 10000, def 1000)
2. send lot of log under one sec
log_number*log_size > ES bulk_max_body_size && log_number < GL output_batch_size
eg. 1000 piece*1M log

## Your Environment

* Graylog Version: v2.4.6+ceaa7e4
* Elasticsearch Version: 5.6.10
* MongoDB Version: NA
* Operating System: CentOS 7, 3.10.0-862.11.6.el7.x86_64
* Browser version: NA

### LINKED ISSUE #6965: Server message processing blocked in some cases when index optimization occurs
## Expected Behavior
Message processing should not be blocked when indices are optimized or if processing is temporarily blocked processing should recover when the reason for block is resolved.

## Current Behavior
When index optimization starts in some cases it causes message processing to stop and currently only known way to resolve this is to restart graylog service. 

When this bug is triggered it seems that Graylog has issues opening network connections to Elasticsearch. Normally there should always be some number of established connections to Elasticsearch but when this issue is active there is zero established connections from Graylog to Elasticsearch.

Screenshot 2019-12-10 at 14 42 01 copy

As you can see from the screenshot, there is only one active system job which is started on the Master node, but it is different node which can't output any messages.

When bug is active following command returns 0.
`sudo netstat -antpu | grep 9200 | grep ESTABLISHED | wc -l`

After service restart there are always some number of established connections even in our smallest test envinronment.

The outgoing message metrics from the applicable node.

Screenshot 2019-12-10 at 15 30 17

About at the same time there is following event in the logs
`2019-12-10T14:13:21.729+02:00 ERROR [Messages] Caught exception during bulk indexing: java.net.SocketException: Broken pipe (Write failed), retrying (attempt #88).`

Currently I don't have steps for reproducing this. For example today this has happened in our largest cluster two times in four hours, both at different nodes. This does not happen every time and this is not related to the Graylog Master node or multiple index optimization tasks as per #4637 . We have also updated the connection limits, but it did not help, because this issue is not caused by the limit because Graylog doesn't open any connections.

## Context
Currently Graylog is unreliable because message processing gets jammed multiple times per day and requires service restart. We have implemented monitoring for this so we notice can react to it immediately.

As you can see from the screenshot, we have quite large volumes incoming - constantly over 10.000 msg/sec and over 1 terabyte of logs per day. 

## Your Environment

* Graylog Version: 3.0.2
* Elasticsearch Version: 6.8.4
* MongoDB Version: 4.0.13
* Operating System: CentOS 7.7

mpfz0r:
If we try to bulk index a batch of messages that exceeds the
elastic search `http.max_content_length` setting. (default 100MB)
Elastic will respond with an HTTP 413 Entity Too Large error.

In this case we retry the request by splitting the message batch
in half.

When responding with an HTTP 413 error, the server is allowed to close the connection
immediately. This means that our HTTP client (Jest) will simply report
an IOException (Broken pipe) instead of the actual error.
This can be avoided by sending the request with an Expect-Continue
header, which also avoids sending data that will be discarded later on.

Fixes #5091
Fixes #6965

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -27,14 +27,18 @@ import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.WaitStrategies;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
 import io.searchbox.client.JestClient;
 import io.searchbox.client.JestResult;
+import io.searchbox.client.http.JestHttpClient;
 import io.searchbox.core.Bulk;
 import io.searchbox.core.BulkResult;
 import io.searchbox.core.DocumentResult;
 import io.searchbox.core.Get;
 import io.searchbox.core.Index;
 import io.searchbox.indices.Analyze;
+import org.apache.http.client.config.RequestConfig;
+import org.graylog2.indexer.ElasticsearchException;
 import org.graylog2.indexer.IndexFailure;
 import org.graylog2.indexer.IndexFailureImpl;
 import org.graylog2.indexer.IndexMapping;
@@ -52,6 +56,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
@@ -66,6 +71,7 @@ import static com.codahale.metrics.MetricRegistry.name;
 public class Messages {
     private static final Logger LOG = LoggerFactory.getLogger(Messages.class);
     private static final Duration MAX_WAIT_TIME = Duration.seconds(30L);
+    private static final int MAX_SPLIT_TRIES = 3;
     private static final Retryer<BulkResult> BULK_REQUEST_RETRYER = RetryerBuilder.<BulkResult>newBuilder()
             .retryIfException(t -> t instanceof IOException)
             .withWaitStrategy(WaitStrategies.exponentialWait(MAX_WAIT_TIME.getQuantity(), MAX_WAIT_TIME.getUnit()))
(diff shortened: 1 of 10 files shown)


mpfz0r:
> The build failed:
> 
> ```
> [ERROR] Forbidden method invocation: org.joda.time.DateTime#now() [Constructing a DateTime without a time zone is dangerous]
> [ERROR]   in org.graylog2.indexer.messages.MessagesIT (MessagesIT.java:120)
> ```

I wonder why it doesn't fail here: (where I copied it from)

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/test/java/org/graylog2/indexer/indices/IndicesIT.java#L80

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -185,7 +238,9 @@ private void recordTimestamp(List<Map.Entry<IndexSet, Message>> messageList, Set
 
     private BulkResult runBulkRequest(final Bulk request, int count) {
         try {
-            return BULK_REQUEST_RETRYER.call(() -> client.execute(request));
+            // Enable Expect-Continue to catch 413 errors before we send the actual data
+            final RequestConfig requestConfig = RequestConfig.custom().setExpectContinueEnabled(true).build();
```
bernd: I noticed the following in the documentation for `RequestConfig#isExpectContinueEnabled()`:

> 'Expect: 100-continue' handshake should be used with caution, as it may cause problems with HTTP servers and proxies that do not support.

So I wonder if we should make this configurable in `graylog.conf`? (default: enabled) Otherwise we cannot do anything if a user has a broken proxy in front of Elasticsearch.

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -137,40 +142,88 @@ public ResultMessage get(String messageId, String index) throws DocumentNotFound
...
         }
     }
 
+    private List<BulkResult.BulkResultItem> bulkIndexChunked(final List<Map.Entry<IndexSet, Message>> messageList, boolean isSystemTraffic, int chunkSize) throws EntityTooLargeException {
+        chunkSize = Math.min(messageList.size(), chunkSize);
+
+        final List<BulkResult.BulkResultItem> failedItems = new ArrayList<>();
+        final Iterable<List<Map.Entry<IndexSet, Message>>> partition = Iterables.partition(messageList, chunkSize);
+        int partitionCount = 1;
+        for (List<Map.Entry<IndexSet, Message>> subMessageList: partition) {
```
bernd: I think this implementation assumes that the messages inside `messageList` are uniformly sized.

Let's say we have a body limit of 10 MB and we have a `messageList` with message sizes of `[1 MB, 1 MB, 1 MB, 10 MB]`. If we run this, we will get a `413` for the first try and split the message list into two partitions. The first partition (two messages) will be successfully written to ES but for the second partition we will get a `413` response from ES again. Then we throw an `EntityTooLargeException` and start over indexing **all** messages again.

This will result in duplicated messages.

One possible solution could be to store the offset of the successfully written messages in the `EntityTooLargeException` and then use `List#subList()` to only pass in the remaining messages into `bulkIndexChunked()`.
This would still not handle the issue that `failedItems` of successful bulk requests would be lost. (we could include them in the exception as well...)

There might be a better way.

graylog2-server/src/main/java/org/graylog2/outputs/ElasticSearchOutput.java
```diff
@@ -112,6 +112,7 @@ public void writeMessageEntries(List<Map.Entry<IndexSet, Message>> messageList)
         }
         failures.mark(failedMessageIds.size());
 
+        // TODO why doesn't this exclude the failedMessageIds?
```
bernd: I think we don't exclude the failed IDs because we don't know if we can retry the failed messages. If we have a message that couldn't be written to ES due to a data type error and we don't commit its offset, then it will be retried over and over again. Okay, in practice it will most probably not because that would require the message to be the last one.

But then it also doesn't make sense to exclude the offset either because if the faulty message is in the middle of a batch and we commit the last offset, it's gone anyway.

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
mpfz0r: Good catch! I've made some changes to handle this.

graylog2-server/src/main/java/org/graylog2/outputs/ElasticSearchOutput.java
mpfz0r: fair enough. should I add that to the comment?

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
mpfz0r: good idea. I've added an option

Issue #5908, opened 2019-04-26
Errorpage "Cannot read property 'name' of null" in Authentication - Roles Creation


iT3han:
## Expected Behavior
If you leave the "Create new role" page by clicking in the left menu on "Roles" again, you should not get the error.

## Current Behavior
If you leave the "Create new role" page by clicking in the left menu on "Roles" again, you get the "Something went wrong." page with the error message displayed below.

## Possible Solution
-

## Steps to Reproduce (for bugs)
1. Go to Authentication
2. Go to Roles
3. Click "Add new role"
4. Leave page by clicking on "Roles" again

![image](https://user-images.githubusercontent.com/43262157/56784375-6a9d0c80-67f0-11e9-9f54-7ac809f4f081.png)
![image](https://user-images.githubusercontent.com/43262157/56784422-a59f4000-67f0-11e9-97ab-782df502c971.png)

## Context
Not critical. If you leave the page by clicking on any other menu or submenu (i tried most of them) or by clicking the cancel button, everything works. 

## Your Environment

* Graylog Version: Graylog 3.0.1+de74b68 
* Elasticsearch Version: 6.7.1
* MongoDB Version: 3.6.3
* Operating System: Ubuntu 18.04.2 LTS
* Browser version: Chrome 73.0.3683.103 (64-Bit)

edmundoa:
This will be fixed in Graylog 3.2.0.

PR #7073, opened 2020-01-02
Fix issue when clicking on Roles menu after clicking "Create role" button

ousmaneo:
## Description

Before this change clicking on the roles menu on the Authentication page after initially loading the create role from resulted in Errorpage "Cannot read property 'name' of null".
Now we use default state to ensure role object is always set. 

## Motivation and Context

#5908

## Types of changes

## Checklist:


graylog2-web-interface/src/components/users/EditRole.jsx
```diff
@@ -1,85 +1,77 @@
...
     };
-  },
+  }
 
-  componentWillReceiveProps(newProps) {
-    this.setState({ role: newProps.initialRole, initialName: this._safeRoleName(newProps.initialRole) });
-  },
+  static getDerivedStateFromProps(nextProps, prevState) {
+    const {
+      role = {
```
edmundoa: Since we use `getDerivedStateFromProps` to initialize an "empty" role, and `getDerivedStateFromProps` is called before the first render and before any subsequent update is rendered, I think the constructor can be simplified and doesn't need to handle the default case.

So I think we can simplify this logic doing the following:
1. In the constructor, we set the role to the `initialRole`, since we are also using that to get the `initialName`
2. In `getDerivedStateFromProps` we check if `prevState.role` is set, and if it is not set we initialize it to the "empty" role

ousmaneo: Ok for that, added the update

edmundoa:
LGTM, thank you! 👍

PR #7081, opened 2020-01-06
Fix reference errors in swagger.js

jrunu:
Bug: If an API endpoint returns an error the api browser becomes unfunctional with `ReferenceError: error is not defined`

## Description
If one or more API endpoints return error the API browser does not show any documentation for any of the endpoints or that an error occured with said api endpoints. This is because the JavaScript stops processing due to the reference error.

## How Has This Been Tested?
This is untested. However for the change at line 308 was indirectly verfied by breaking to it in a debugger and examining the objects.  As there is no statusText for the object in line 308 I'm making the assumption there's none for the response object in line 62 either.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/resources/swagger/lib/swagger.js
+++ b/graylog2-server/src/main/resources/swagger/lib/swagger.js
@@ -59,7 +59,7 @@
             } else if (response.status === 404) {
               return _this.fail('Can\'t read swagger JSON from ' + _this.url);
             } else {
-              return _this.fail(response.status + ' : ' + response.statusText + ' ' + _this.url);
+              return _this.fail('Server return ' + response.status + ' for ' + _this.url);
             }
           },
           response: function(rawResponse) {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/resources/swagger/lib/swagger.js
```diff
@@ -305,7 +305,7 @@
           headers: {},
           on: {
             error: function(response) {
-              return _this.api.fail("Unable to read api '" + _this.name + "' from path " + _this.url + " (server returned " + error.statusText + ")");
+              return _this.api.fail("Unable to read api '" + _this.name + "' from path " + _this.url + " (server returned " + error.status + ")");
```
mpfz0r: this should be `response.status`

jrunu: Well, that's a facepalm. I corrected it, probably not in the workflow github envisioned for this, but its there anway.

PR #7087, opened 2020-01-06
Add Listfield option in ConfigurationRequest

### LINKED ISSUE #4653: Cannot make a ListField not optional
## Expected Behavior
Make ListField not optional in conditional alert plugin

## Possible Solution
In the check function in class configurationRequest  there is no case for the type ListField 

https://github.com/Graylog2/graylog2-server/blob/ce9b5f268bda917288331d2055614c52abb14f5a/graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java#L98-L125

* Graylog Version: 2.4.3

ceruleancee:
## Description

ConfigurationRequest class was missing Listfield as a field type

## Motivation and Context

Resolves #4653

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java
```diff
@@ -150,6 +152,8 @@ public Configuration filter(Configuration config) {
                     }
                     break;
                 case TextField.FIELD_TYPE:
+                case ListField.FIELD_TYPE:
+                    break;
```
bernd: This will break the `TextField.LIST_TYPE` case, because it now also returns on the `break`. We need to implement a separate case for `ListField.FIELD_TYPE` that also sets the list value like the `DropdownField.FIELD_TYPE` does. (just for lists) This probably needs a `listIsSet(String)` method implementation in `org.graylog2.plugin.configuration.Configuration`.

graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java
```diff
@@ -150,6 +152,8 @@ public Configuration filter(Configuration config) {
                     }
                     break;
                 case TextField.FIELD_TYPE:
+                case ListField.FIELD_TYPE:
```
bernd: We also need to add a case for `ListField.FIELD_TYPE` in the `check()` method in this class.

graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java
ceruleancee: oh okay. will add method.

graylog2-server/src/main/java/org/graylog2/plugin/configuration/Configuration.java
```diff
@@ -138,6 +138,10 @@ public void setBoolean(String key, boolean value) {
         return firstNonNull(lists.get(key), defaultValue);
     }
 
+    public boolean listIsSet(String key) {
+        return !lists.isEmpty();
```
bernd: This needs to check if the list for the given `key` is set. The current code is checking if the map of lists is empty or not. So this needs to be something along the lines of:

```suggestion
        return lists.containsKey(key);
```

graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java
```diff
@@ -113,6 +115,12 @@ public void check(Configuration configuration) throws ConfigurationException {
                         }
                         break;
                     case TextField.FIELD_TYPE:
+                        break;
```
bernd: This breaks the `TextField.FIELD_TYPE`. Before this PR, the `TextField` and `DropdownField` used the same implementation. (by relying on the fallthrough property of the switch case) This needs to stay the way it was before.

graylog2-server/src/main/java/org/graylog2/plugin/configuration/ConfigurationRequest.java
```diff
@@ -150,6 +158,12 @@ public Configuration filter(Configuration config) {
                     }
                     break;
                 case TextField.FIELD_TYPE:
+                    break;
```
bernd: Same as above, the fallthrough for the `DropdownField` implementation needs to stay for the `TextField`.

PR #7088, opened 2020-01-06
Upgrade to fsevents v1.2.11 to support node 12

edmundoa:
fsevents does not provide binaries for node 12 and the package also failed
to build even if the dependencies for node-gyp appeared to be installed.
Fortunately they released a new 1.2.11 version simplifying the module's
compilation.

This change only affects the development server under macOS, so it should be safe to merge.

### DIFF
--- a/graylog2-web-interface/yarn.lock
+++ b/graylog2-web-interface/yarn.lock
@@ -2798,6 +2798,13 @@ binary-search-bounds@^2.0.0, binary-search-bounds@^2.0.3, binary-search-bounds@^
   resolved "https://registry.yarnpkg.com/binary-search-bounds/-/binary-search-bounds-2.0.4.tgz#eea0e4081da93baa851c7d851a7e636c3d51307f"
   integrity sha512-2hg5kgdKql5ClF2ErBcSx0U5bnl5hgS4v7wMnLFodyR47yMtj2w+UAZB+0CiqyHct2q543i7Bi4/aMIegorCCg==
 
+bindings@^1.5.0:
+  version "1.5.0"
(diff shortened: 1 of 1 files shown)


kyleknighted:
@edmundoa this PR is working for me as is, but was curious on your thoughts of making the needed version more explicit.

What do you think of adding `"fsevents": "^1.2.11"` to the `resolutions` object in the `package.json`?

It will end up generating something similar to
```
fsevents@^1.2.7:
  version "1.2.11"
  resolved "https://registry.yarnpkg.com/fsevents/-/fsevents-1.2.11.tgz#67bf57f4758f02ede88fb2a1712fef4d15358be3"
  integrity sha512-+ux3lx6peh0BpvY0JebGyZoiR4D+oYzdPZMKJwkZ+sFkNJzpL7tXc/wehS49gUAxg3tmMHPHZkA8JU2rhhgDHw==
  dependencies:
    bindings "^1.5.0"
    nan "^2.12.1"
```

edmundoa:
@kyleknighted that was my first approach but it has the problem that `fsevents` is an optional package only used in macOS platforms. When adding it to resolutions yarn will try to install it in all platforms, ending in an error. Adding `fsevents` to both `resolutions` and `optional` packages also made no difference. I couldn't find any other working solution so I decided to not use more time for this and bump versions in `yarn.lock`.

edmundoa:
@kyleknighted I rebased all PRs and updated all `yarn.lock` files, since there were some changes since last time. It would be cool if you could take another look soon before I have to redo it all that over again 😄

PR #6915, opened 2019-12-02
Implement ui for url white lists configuration

ousmaneo:
## Description

This changes add UI interface to configuration page that make it possible to:
* Show list of white-listed url on the configuration page 
* Add form to remove, add or update Url to the white list.
(#6853)
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #6854, opened 2019-11-25
URL whitelist service

thll:
## Description

There are data adapters and event notifications which access URLs. These URLs will now have to be whitelisted in order for the requests to succeed.
This pull request adds a whitelist service, a corresponding REST API and a migration to provide this functionality.

## Motivation and Context

Fixes #6853

## How Has This Been Tested?

Whitelist created and updated via REST API. Migration executed.

Further extensive testing required once we are happy with the approach.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #6853, opened 2019-11-22
Implement URL white list service


bernd:
## Description

When non-admin users have permissions to configure a lookup data adapter, they can use the `HTTPJSONPath` adapter to query potentially sensitive information. (e.g. EC2 instance metadata in AWS) The `HTTPJSONPath` data adapter executes a HTTP request for an URL that is defined by the user that configures the data adapter.

This is **only** an issue if an admin user is giving lookup-table permissions to a regular user. This is not the default.

To mitigate this issue, we want to implement an URL whitelist service. This service can be used by other subsystems that deal with user defined URLs to check if an URL is allowed to be used. The whitelist service requires a separate set of permissions to make sure users can be given permission to create/modify lookup tables but are not allowed to modify the URL whitelist.

## Tasks

- [x] Must be enabled by default
- [x] Support different match types
  - Regexp
  - Literal string
- [x] Separate permissions for the whitelist service
- [x] Check URLs at runtime
- [x] Configuration UI on "System/Configuration" page
- ~~[UI form component for URL whitelist that shows good error message and option to add new URL to whitelist (if user has permission).~~ This was moved to #7169
- [x] Use white list service in lookup data adapters (e.g. `HTTPJSONPath`, `DSVHTTP`) and event notifications
- [x] Create system notification when a runtime URL check failed
- [x] Create migration to add URL from known entities to the white list (e.g. lookup adapters, event notifications)
- [x] Backport to 3.1.x
- [x] Content pack support
- [x] Documentation

thll:
We'll enable the whitelist by default. Should we add an option on the whitelist configuration page to disable the whitelist altogether if users don't want to be bothered with whitelisting?
Even without an explicit option the user can still easily create a catch-all whitelist by adding a regex like `.*`. But that might not be convenient enough.

thll:
Reopening as there are some tasks left to be addressed.

bernd:
@thll Can you please create a new issue for the remaining tasks so we can close this issue for 3.1.4? Thank you! :smiley:

thll:
> @thll Can you please create a new issue for the remaining tasks so we can close this issue for 3.1.4? Thank you! 😃

Sure, I created  #7169 for this.

bernd:
Done! Thank you!

PR #6766, opened 2019-11-12
Migrate non-Reflux stores to Reflux

edmundoa:
Replace old Typescript stores fetching data to Reflux stores. This does
not add actions or any advanced Reflux feature, but leaves the stores
simple to avoid creating a larger changeset.

Using Reflux ensures stores' methods are bound to the object and will
execute in the right execution context, avoiding issues as the one
described in #6704.

Fixes #6704

PR #7110, opened 2020-01-08
Add URL whitelist service (#6854 and #6915)

thll:
Backport of backend (#6854) and frontend (#6915) of the URL Whitelist feature #6853


danotorrey:
This looks good to me. Code looks complete. Waiting on backport for UI (https://github.com/Graylog2/graylog2-server/pull/6915) from @ousmaneo before doing the final review.

One thing I noticed: I think we also need a backport for the Enterprise audit event.

graylog2-web-interface/src/stores/tools/ToolsStore.js
```diff
@@ -0,0 +1,166 @@
+// @flow strict
+import Reflux from 'reflux';
+
+import fetch from 'logic/rest/FetchProvider';
+import ApiRoutes from 'routing/ApiRoutes';
+import URLUtils from 'util/URLUtils';
+import UserNotification from 'util/UserNotification';
+
+const ToolsStore = Reflux.createStore({
```
edmundoa: Adding this store may be problematic, since 3.1. still had the old Typescript `ToolStore.ts` (see https://github.com/Graylog2/graylog2-server/pull/6766 for the PR doing that change).

I think it should be possible to keep this new JS version of the store and remove the old Typescript store, but we should ensure that only one of them is there, otherwise I'm not sure which one the import is going to get.

edmundoa:
After the last changes this looks good to me. @danotorrey I re-requested your review, in case you want to double-check.

danotorrey:
@ousmaneo Can you take a quick look at the Linter check error?

edmundoa:
@danotorrey the linter complains about a file touched in the last fix @ousmaneo did. Since the linting errors are not directly related to the changes, I think fixing them is not worth the risk of introducing new bugs in the backport.

ousmaneo:
@danotorrey  from Edmundo's advice linter error that are not from my change, should be left like that as they might introduce more bugs or break things.

Issue #7109, opened 2020-01-08
Data Table with more fields than horizontal space is not scrollable inside the AggregationBuilder


linuspahl:
This issue is related to #7091. The fix for #7091 reenables the horizontal scrolling for the data table. The scroll bar also exists when editing the data table with the aggregation builder, but is not visible, due to some height calculation for the aggregation builder layout.

![image](https://user-images.githubusercontent.com/46300478/71969820-bf11dc00-3207-11ea-9c11-e4348edb09ec.png)

When adjusting the aggregation builder layout, the new layout should be flexible enough to fix #7105 as well.

## Steps to Reproduce (for bugs)
1. Open the aggregation builder
2. Create a data table with many table columns
3. Try to scroll horizontically

## Your Environment
* Graylog Version: 3.2 beta2

dennisoelkers:
Fixed in #7114.

Issue #7105, opened 2020-01-08
Left Side of Widget Edit Mode in Aggregation Builder should be scrollable


dennisoelkers:
Steps to reproduce:

1. Edit Aggregation widget
2. Zoom in just enough so the sidebar to the left of the graph is not shown completely
3. Try to scroll it to reach hidden controls

![Screen Shot 2020-01-08 at 10 45 06](https://user-images.githubusercontent.com/41929/71967875-3f364280-3204-11ea-8a55-aa326081dc28.png)

linuspahl:
Fixed with https://github.com/Graylog2/graylog2-server/pull/7114

PR #7114, opened 2020-01-08
Make configuration section layout in AggregationBuilder more responsive

linuspahl:
This PR fixes the horizontal scrolling / overflow problem described in #7109 and also the vertical scrolling problem described in #7105 regarding the AggregationBuilder.

It looks like we could also get rid of `MeasureDimensions`, used in `Widget.jsx`, it is just passing the container height to its children, but I don't see any benefit. I am not doing this refactoring inside this PR, because I want to avoid a bigger layout change at this moment.

## How Has This Been Tested?
Tested in Chrome, Firefox and Safari

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/DescriptionBox.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/DescriptionBox.jsx
@@ -12,12 +12,16 @@ import HoverForHelp from './HoverForHelp';
 export default class DescriptionBox extends React.Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
-    description: PropTypes.string.isRequired,
     configurableOptions: PropTypes.node,
+    description: PropTypes.string.isRequired,
+    help: PropTypes.string,
+    style: PropTypes.object,
   };
 
   static defaultProps = {
     configurableOptions: undefined,
+    style: undefined,
+    help: undefined,
   };
 
   constructor(props) {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/views/components/aggregationbuilder/DescriptionBox.jsx
```diff
@@ -12,12 +12,16 @@ import HoverForHelp from './HoverForHelp';
 export default class DescriptionBox extends React.Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
-    description: PropTypes.string.isRequired,
     configurableOptions: PropTypes.node,
+    description: PropTypes.string.isRequired,
+    help: PropTypes.string,
+    style: PropTypes.object,
```
linuspahl: It's just a small implementation detail, but let's talk about the style prop. All I need to do is overwriting the top margin of one DescriptionBox inside the Aggregation Builder. I had two options:
- The current way
- Adding a styled-component inside the AggregationBuilder which extends the DescriptionBox
Requires the new component + adding a className prop + an && workaround for the css, to be more specific

It's the second time we are passing the style attribute as a prop to the inner component. Is it okay to do it this way?

dennisoelkers: Considering the drawbacks of the `styled-component` approach, passing a `style` prop for small & isolated cases is fine to me.

PR #7122, opened 2020-01-09
Use Urlwhitelist API instead of clusterConfig API

ousmaneo:
Before this change we were using ClusterConfig API to update
Urlwhitelist, result of that audit log shows cluster config event
instead of Urlwhitelist event.

* Add updateWhitelist method to store
* Update ConfigurationsPage to use updateWhitelist when updating
urlwhitelist

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
+++ b/graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
@@ -122,11 +122,11 @@ const ConfigurationsStore = Reflux.createStore({
       () => {
         this.configuration[configType] = config;
         this.propagateChanges();
-        UserNotification.success('Configuration updated successfully');
+        UserNotification.success('Url Whitelist Configuration updated successfully');
         return config;
       },
       (error) => {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
```diff
@@ -115,6 +115,24 @@ const ConfigurationsStore = Reflux.createStore({
...
+
+    promise.then(
+      () => {
+        this.configuration[configType] = config;
+        this.propagateChanges();
+        UserNotification.success('Configuration updated successfully');
+        return config;
+      },
+      (error) => {
+        UserNotification.error(`Search config update failed: ${error}`, `Could not update search config: ${configType}`);
```
thll: We should adjust the wording here to refer to the URL whitelist instead of "Search".

ousmaneo: sure ! Thank you for the catch.

PR #7125, opened 2020-01-09
Complete migration to React Select v2

### LINKED ISSUE #6946: Update React-Select props
After the upgrade to React-Select v2, it seems that we missed some props and didn't update them. These are the ones that got removed or renamed, according to the [upgrade guide](https://react-select.com/upgrade-guide#prop-update-guide)

```
clearable -> isClearable
disabled -> isDisabled
displayKey - Removed
inputProps - Removed, use components API
matchProp - Removed, use createFilter()
multi -> isMulti
onBlurResetsInput - Removed
onClose -> onMenuClose
onCloseResetsInput - Removed
optionRenderer - Removed, use components API
required - Removed, may be implemented later
style - Removed, use styles API
valueKey - Removed
valueRenderer - Removed, use components API
```

## Current Behavior
This affects to some of the selects across the product in different ways. Most of them will work as expected though.

## Your Environment
* Graylog Version: Graylog-3.2.0-SNAPSHOT, Graylog-3.1.3

edmundoa:
During the React Select migration to v2 we missed to replicate the functionality of some props that got removed or renamed, so some features of selects we and plugins relied on were not working as expected in some cases.

This PR fixes all missing functionality I found, and adds some tests to ensure that is currently working. Here is some information on the props adapted in this PR:

- `clearable` was renamed to ` isClearable`. Since the default value for this property in v1 was `true`, I used the same as default
- `disabled` inside the `options` object is no longer supported, we now use `isOptionDisabled` to check if that property is set and disable the `option`
- `inputProps` was removed, we use a custom component to achieve that. All usages of `inputProps` were replaced by `inputId`, since setting the inner input ID is all what we were using that for
- `optionRenderer` was removed, we use a custom component to achieve that functionality
- `valueRenderer` was removed, we use a custom component to achieve that functionality
- `matchProp` was removed, we use `createFilter` to create a custom filter
- `required` was removed but may be implemented later. I don't see an easy way to replicate that behaviour, so I vote for leaving the props as they are and adding an issue that targets that specifically
- `style` was removed and I don't see an easy way to adapt that in a generic way, since the API for that changed dramatically. We only used that `prop` in two components and as far as I could see was not really necessary, so I removed it from them
- `onBlurResetsInput` and `onCloseResetsInput` were also removed. They were only used in `SeriesSelect` as far as I could see, and after asking the Search team we decided to remove them

For reference, here are the [v2 upgrade guide](https://react-select.com/upgrade-guide#prop-update-guide) and the [v1 props](https://github.com/JedWatson/react-select/tree/v1.x#select-props).

Fixes #6946

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -156,7 +156,7 @@ type Option = { [string]: any };
 type Props = {
   onChange: (string) => void,
   placeholder: string,
-  clearable: boolean,
+  clearable?: boolean,
   displayKey?: string,
   valueKey?: string,
   delimiter?: string,
@@ -214,6 +214,7 @@ class Select extends React.Component<Props, State> {
      */
     components: PropTypes.arrayOf(PropTypes.node),
     disabled: PropTypes.bool,
+    clearable: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -230,6 +231,7 @@ class Select extends React.Component<Props, State> {
     addLabelText: undefined,
     onReactSelectChange: undefined,
     components: null,
+    clearable: true,
   };
 
   constructor(props: Props) {
@@ -313,6 +315,7 @@ class Select extends React.Component<Props, State> {
     const {
       multi: isMulti = false,
       disabled: isDisabled = false,
+      clearable: isClearable,
       ...rest
     } = this.props;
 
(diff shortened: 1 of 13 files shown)


edmundoa:
The linter build fails in components where some `Select` prop changed. I think fixing those linting errors falls out of the scope of this PR.

kyleknighted:
Lint errors are unrelated to this change, agree that they can wait.

PR #7130, opened 2020-01-09
Enforce system wide TLS protocol settings

### LINKED ISSUE #6444: Graylog 3.1 is not honoring java security options for collector inputs
I hava Java security options set to disable TLS 1.0/1.1 for the beats inputs only.
jdk.tls.disabledAlgorithms=SSLv2Hello, SSLv3, TLSv1, TLSv1.1, EC keySize < 160, RSA keySize < 2048, DSA keySize < 2048
jdk.certpath.disabledAlgorithms=MD2, MD4, MD5, EC keySize < 160, RSA keySize < 2048, DSA keySize < 2048

Graylog 3.1 instance is not honoring the settings, the Graylog 2.5.2 instance allows TLS 1.1/1.2 to be disabled.

mpfz0r:
With the introduction of netty 4 in Graylog 3.0, we changed the default
SslProvider to OpenSSL.
While this gives a better performance, it lacks the ability to disable
TLS protocols that are considered flawed.
Changes to the java security properties are only respected by jdk based
SslProviders (jersey / grizzly in our case).

Introduce a system wide default setting for supported TLS protocols,
with a sane default to TLSv1.2 (and TLSv1.3 if the jdk supports it).
These protocol settings are applied to netty based inputs and
to the webserver.

Changes can be made with a new server setting:
`#enabled_tls_protocols = TLSv1.2,TLSv1.3`

Fixes #6444

PS: This can be tested by using 
`nmap --script ssl-enum-ciphers -p <inputs-port> 127.0.0.1`

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -27,6 +27,7 @@ import com.github.joschi.jadconfig.validators.PositiveDurationValidator;
 import com.github.joschi.jadconfig.validators.PositiveIntegerValidator;
 import com.github.joschi.jadconfig.validators.PositiveLongValidator;
 import com.github.joschi.jadconfig.validators.StringNotBlankValidator;
+import com.google.common.collect.ImmutableSet;
 import org.graylog2.plugin.BaseConfiguration;
 import org.graylog2.security.realm.RootAccountRealm;
 import org.graylog2.utilities.IPSubnetConverter;
@@ -157,6 +158,11 @@ public class Configuration extends BaseConfiguration {
 
     @Parameter(value = "deactivated_builtin_authentication_providers", converter = StringSetConverter.class)
     private Set<String> deactivatedBuiltinAuthenticationProviders = Collections.emptySet();
+
+    // Defaults to TLS protocols that are currently considered secure
+    @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
+    private Set<String> enabledTlsProtocols = ImmutableSet.of("TLSv1.2", "TLSv1.3");
+
 
     public boolean isMaster() {
         return isMaster;
@@ -311,6 +317,10 @@ public class Configuration extends BaseConfiguration {
 
     public Set<String> getDeactivatedBuiltinAuthenticationProviders() {
         return deactivatedBuiltinAuthenticationProviders;
+    }
+
+    public Set<String> getEnabledTlsProtocols() {
+        return enabledTlsProtocols;
     }
 
     @ValidatorMethod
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -158,6 +159,11 @@
     @Parameter(value = "deactivated_builtin_authentication_providers", converter = StringSetConverter.class)
     private Set<String> deactivatedBuiltinAuthenticationProviders = Collections.emptySet();
 
+    // Defaults to TLS protocols that are currently considered secure
+    @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
```
bernd: We should add an entry to `misc/graylog.conf` including a comment that describes the available options.

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -158,6 +159,11 @@
...
+    private Set<String> enabledTlsProtocols = ImmutableSet.of("TLSv1.2", "TLSv1.3");
```
bernd: I noticed that the default we use in 3.1 and earlier is `TLSv1.0`, `TLSv1.1` and `TLSv1.2`. Changing this to `TLSv1.2` and `TLSv1.3` only would be a breaking change and might result in broken user and customer setups. So I think we need to enable all previous defaults to avoid any issues.

We should probably document that people should switch to more secure ones. With the next major release (v4) we can change the default.

mpfz0r: I disagree. TLS 1.0 and 1.1 are known to have security issues. We should be shipping with a secure default.
TLS 1.2 is supported in major browsers since ~2013 and is the de-facto standard. 94% global usage in 2018. https://www.ixiacom.com/company/blog/current-state-tls  

I think the fallout will be very small. We can make the change (and workaround) prominent in the 3.2 release notes.
At the very least, we should disable TLS 1.0 (since that one is not even PCI DSS compliant anymore)
https://blog.pcisecuritystandards.org/are-you-ready-for-30-june-2018-sayin-goodbye-to-ssl-early-tls

mpfz0r: cf https://portswigger.net/daily-swig/the-end-is-nigh-browser-makers-ditch-support-for-aging-tls-1-0-1-1-protocols

bernd: Okay, we talked about this offline. I assumed this changes should be backported to 3.1, but that isn't the case because there is a workaround. (will be documented in #6444)

mpfz0r:
> As discussed offline, please add some documentation to the `UPGRADING.rst` file for 3.2. Thank you!

I did that in a separate PR: https://github.com/Graylog2/graylog2-server/pull/7161

bernd:
> > As discussed offline, please add some documentation to the `UPGRADING.rst` file for 3.2. Thank you!
> 
> I did that in a separate PR: #7161

Great! Thanks!

PR #7131, opened 2020-01-09
Display "Create extractor" action only in the context of a message

### LINKED ISSUE #7111: "Create extractor" action should be available only in the context of a message
Steps to reproduce:

1. Create aggregation
2. Click on any value
3. Select "Create extractor"

This raises an error due to the absence of a `message` object in the action's context. Using the "Create extractor" action requires loading a sample message, so it only makes sense when triggered in a message context.

linuspahl:
As described in #7111 the action "Create extractor" should only be clickable in the context of a message and e.g. not in the context of a data table.

Fixes #7111

## Types of changes

## Checklist:


graylog2-web-interface/src/views/bindings.jsx
```diff
@@ -249,7 +249,7 @@ export default {
     {
       type: 'create-extractor',
       title: 'Create extractor',
-      isEnabled: (({ type }) => (type.type === 'string' && !type.isDecorated()): ActionHandlerCondition),
+      isEnabled: (({ type, contexts }) => (!!contexts.message && type.type === 'string' && !type.isDecorated()): ActionHandlerCondition),
```
linuspahl: Interesting learning: I had to implement the boolean casting, otherwise the function would not return `false` but `undefined`. I thought it is enough to do the chaining with `&&`.

graylog2-web-interface/src/views/bindings.test.jsx
```diff
@@ -155,5 +156,51 @@ describe('Views bindings', () => {
           .toEqual(false);
       });
     });
+    describe('CreateExtractor', () => {
```
dennisoelkers: This is now part of the `field actions` `describe`-block, although it is a value action. I think this is a good time to either introduce a new `value actions` block, or split the file into two, `bindings.fieldActions.test.jsx` & `bindings.valueActions.test.jsx`.

graylog2-web-interface/src/views/bindings.jsx
dennisoelkers: Returning `undefined` here is okay, as falsyness is considered sufficient.

graylog2-web-interface/src/views/bindings.test.jsx
linuspahl: Good catch! Because we are planing to split up the bindings in the future, I've create two separate test files. When we are adding even more separate test files, it would be okay, to move them inside a `__tests__` dir, right?

dennisoelkers: Sure!

PR #7133, opened 2020-01-09
Fixing effective time range in pivot search type results.

### LINKED ISSUE #7128: Timeline chart uses start/end timestamps from documents, not time range
A timeline chart (any line/area/scatter/bar visualization with a `timestamp` or any other `date` field as row pivot) uses `min(timestamp)`/`max(timestamp)` of the document set instead of the time range as start/end times of the x axis.

Steps to reproduce:

1. Search for anything that occurs within a time range, but not at its boundaries
2. Create an aggregation with a `count()` metric and a `timestamp` row pivot

Screen Shot 2020-01-09 at 16 06 50

dennisoelkers:
## Motivation and Context

Prior to this change, the effective time range of the pivot was used for
the search type's result only if the result contains no documents
(causing the min/max aggregations to return `null`). This led to the
effective time range for the pivot search type being the
smallest/largest timestamp of its documents, instead of the
beginning/end of the supplied time range.

This PR is now changing this behavior to do one of these:

  - If the effective time range of the search type is set to "All
messages" and the search returns documents, the effective time range of
the result is the smallest/largest timestamp of the result set
  - If the effective time range of the search type is set to "All
messages" and the search returns no documents, the effective time range
of the result is set to unix epoch 0/now
  - In all other cases, the time range of the result is set to the
effective time range of the search type

Fixes #7128.

## Types of changes

## Checklist:


kmerz:
Done! :+1:

PR #7055, opened 2019-12-23
Convert week/month bucket interval to days if value is bigger than 1.

dennisoelkers:
## Motivation and Context

Due to limitations in Elasticsearch, it does not support time unit
bucket intervals with values bigger than 1 for units bigger than days.
Therefore, this PR transparently converts these to days (with multiples
of 7 resp. 30) if the value is bigger than 1, leaving the original unit
otherwise.

Fixes #7022.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #7140, opened 2020-01-10
Convert week/month bucket interval to days if value is bigger than 1.

### LINKED ISSUE #7022: Can't select more then 1 week in time bucket Interval
## Expected Behavior
When disabling automatic bucket interval the user can choose between different bucket sizes (minute, hour... weeks, months). When selecting weeks the user expects that he can choose more
then 1 week.

## Current Behavior
When choosing `weeks` and selecting more then one elasictsearch is raising an error.

## Possible Solution
Convert to days?

Peek 2019-12-18 16-28

## Your Environment
* Graylog Version: 3.2 dev (c9c2ee7393467652238811423a415206e969f9da)

dennisoelkers:
## Motivation and Context

Due to limitations in Elasticsearch, it does not support time unit
bucket intervals with values bigger than 1 for units bigger than days.
Therefore, this PR transparently converts these to days (with multiples
of 7 resp. 30) if the value is bigger than 1, leaving the original unit
otherwise.

In contrast to the initial attempt in #7055, this PR is performing the
change in the backend, transparent to the caller and relieving the
caller from having to do the conversion before.

Fixes #7022.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
@@ -42,6 +42,11 @@ public abstract class TimeUnitInterval implements Interval {
     @JsonProperty
     public abstract String timeunit();
 
+    @Override
+    public DateHistogramInterval toDateHistogramInterval(TimeRange timerange) {
+        return new DateHistogramInterval(adjustUnitsLongerThanDays(timeunit()));
+    }
+
     private String adjustUnitsLongerThanDays(String timeunit) {
         final Matcher matcher = TIMEUNIT_PATTERN.matcher(timeunit());
         checkArgument(matcher.matches(),
@@ -58,11 +63,6 @@ public abstract class TimeUnitInterval implements Interval {
             case "M": return quantity == 1 ? timeunit : (30 * quantity) + "d";
             default: throw new RuntimeException("Invalid time unit: " + timeunit);
         }
-    }
-
-    @Override
-    public DateHistogramInterval toDateHistogramInterval(TimeRange timerange) {
-        return new DateHistogramInterval(adjustUnitsLongerThanDays(timeunit()));
     }
 
     @AutoValue.Builder
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
@@ -51,7 +51,7 @@ public abstract class TimeUnitInterval implements Interval {
         final Matcher matcher = TIMEUNIT_PATTERN.matcher(timeunit());
         checkArgument(matcher.matches(),
                 "Time unit must be {quantity}{unit}, where quantity is a positive number and unit [smhdwM].");
-        final int quantity = Integer.parseInt(matcher.group("quantity"), 10);
+        final int quantity = Integer.parseInt(matcher.group("quantity"));
         final String unit = matcher.group("unit");
 
         switch (unit) {
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
@@ -24,11 +24,17 @@ import com.google.auto.value.AutoValue;
 import org.elasticsearch.search.aggregations.bucket.histogram.DateHistogramInterval;
 import org.graylog2.plugin.indexer.searches.timeranges.TimeRange;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
 @AutoValue
 @JsonTypeName(TimeUnitInterval.type)
 @JsonDeserialize(builder = TimeUnitInterval.Builder.class)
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
```diff
@@ -24,21 +24,45 @@
...
     public abstract String type();
 
     @JsonProperty
     public abstract String timeunit();
 
+    private String adjustUnitsLongerThanDays(String timeunit) {
+        final Matcher matcher = TIMEUNIT_PATTERN.matcher(timeunit());
+        checkArgument(matcher.matches(),
+                "Time unit must be {quantity}{unit}, where quantity is a positive number and unit [smhdwM].");
+        final int quantity = Integer.parseInt(matcher.group("quantity"), 10);
```
alex-konn: What's the benefit of specifying a radix of 10 here explicitly? I would find it clearer, if we omitted it.

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
```diff
@@ -24,21 +24,45 @@
...
     public static final String type = "timeunit";
+    public static final Pattern TIMEUNIT_PATTERN = Pattern.compile("(?<quantity>\\d+)(?<unit>[smhdwM])");
 
     @JsonProperty
     public abstract String type();
 
     @JsonProperty
     public abstract String timeunit();
 
+    private String adjustUnitsLongerThanDays(String timeunit) {
```
alex-konn: Can we move the method below `toDateHistogramInterval` where it's first used?

graylog2-server/src/test/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitIntervalTest.java
```diff
@@ -0,0 +1,83 @@
...
+public class TimeUnitIntervalTest {
+    private TimeUnitInterval.Builder builder() {
+        return TimeUnitInterval.Builder.builder();
+    }
+
+    @Test
+    public void doesNotAllowInvalidTimeUnit() {
+        assertThatExceptionOfType(RuntimeException.class)
+                .isThrownBy(() -> builder().timeunit("foobar").build())
+                .withMessage("Time unit must be {quantity}{unit}, where quantity is a positive number and unit [smhdwM].");
```
alex-konn: :+1: for validating the exception message! I usually put the message in a static field on the class and reference it in the test to avoid having it to change it in two places when the time comes.

graylog2-server/src/test/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitIntervalTest.java
```diff
@@ -0,0 +1,83 @@
...
+package org.graylog.plugins.views.search.searchtypes.pivot.buckets;
+
+import org.elasticsearch.search.aggregations.bucket.histogram.DateHistogramInterval;
+import org.graylog2.plugin.indexer.searches.timeranges.InvalidRangeParametersException;
+import org.graylog2.plugin.indexer.searches.timeranges.RelativeRange;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
```
alex-konn: Nice and clean tests! I think they could be further improved by using parameterized tests. That would make them even more compact and facilitate adding more cases. Totally optional though. The tests are fine as they are now.

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitInterval.java
dennisoelkers: No other reason besides being as explicit as possible. If it's too explicit, I will remove it :)

graylog2-server/src/test/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitIntervalTest.java
dennisoelkers: I do understand that, but I usually prefer to have literals in tests, because it is harder for me to overlook mismatching expectations because I _think_ I know what the variable is.

graylog2-server/src/test/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/TimeUnitIntervalTest.java
dennisoelkers: I was also thinking about that, but I hesitated doing it, because afaik it requires all methods of a test class to be parameterized, so the test class would have to be split up into one for the validation tests and one for the valid inputs. I can still do it though, I just preferred to have a single test class right now, because it makes it easier to spot corner cases which are not covered.

dennisoelkers: Well no, I can just wrap those in an inner class. Will do!

alex-konn: Right. Forgot that we're on JUnit 4. JUnit 5 has much better support for parameterized tests. For the short term we should consider using https://github.com/Pragmatists/JUnitParams then.

PR #7141, opened 2020-01-10
Remove widget position from view state, on widget delete

### LINKED ISSUE #6985: Widget positions persist in view state when removing a widget.
This issue does not result in a visible bug, our widget grid is still able to display the widgets correctly. But it looks like we should remove the related widget position in the view state, when removing a widget.

Screenshot of the view state after removing a widget. This state persists when e.g. saving a dashboard.
image

linuspahl:
As described in #6985 the widget position persists in the view state when deleting a widget. When saving a dashboard the widget position still persists in the database as well.

This PR just removes all positions without a related widget, when updating the widgets.

Fixes #6985

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
+++ b/graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
@@ -86,13 +86,15 @@ describe('CurrentViewStateStore', () => {
   });
 
   it('should remove widget positions for deleted widgets', async () => {
+    const widgetOne = MessagesWidget.builder().id('widget-one').build();
+    const widgetOnePos = new WidgetPosition(1, 1, 5, 6);
     const existingViewState = viewState.toBuilder()
       .widgetPositions({
-        'widget-one': new WidgetPosition(1, 1, 5, 6),
+        'widget-one': widgetOnePos,
         'widget-two': new WidgetPosition(1, 6, 5, 6),
       })
       .widgets([
-        MessagesWidget.builder().id('widget-one').build(),
+        widgetOne,
         MessagesWidget.builder().id('widget-two').build(),
       ])
       .build();
@@ -100,9 +102,9 @@ describe('CurrentViewStateStore', () => {
     viewStateMap[viewId] = existingViewState;
     const sMap = Immutable.Map(viewStateMap);
 
-    const expectedWidgets = [existingViewState.widgets[0]];
+    const expectedWidgets = [widgetOne];
     const expectedViewState = viewState.toBuilder()
-      .widgetPositions({ 'widget-one': existingViewState.widgetPositions['widget-one'] })
+      .widgetPositions({ 'widget-one': widgetOnePos })
       .widgets(expectedWidgets)
       .build();
     const updateFn = mockAction(jest.fn(() => Promise.resolve(expectedViewState)));
--- a/graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
+++ b/graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
@@ -3,7 +3,6 @@ import * as Immutable from 'immutable';
 import { PluginStore } from 'graylog-web-plugin/plugin';
 
 import mockAction from 'helpers/mocking/MockAction';
-import asMock from 'helpers/mocking/AsMock';
 import MessagesWidget from 'views/logic/widgets/MessagesWidget';
 import WidgetPosition from 'views/logic/widgets/WidgetPosition';
 import ViewState from 'views/logic/views/ViewState';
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
```diff
@@ -103,4 +103,39 @@ describe('CurrentViewStateStore', () => {
...
+    const updateFn = mockAction(jest.fn((id, newViewState) => {
+      expect(id).toEqual(viewId);
+      expect(newViewState).toEqual(expectedViewState);
+      return Promise.resolve(expectedViewState);
+    }));
+
+    ViewStatesActions.update = updateFn;
+    CurrentViewStateStore.onViewStoreChange({ activeQuery: viewId, view: existingViewState });
+    CurrentViewStateStore.onViewStatesStoreChange(sMap);
+    CurrentViewStateStore.widgets(expectedWidgets);
```
dennisoelkers: I think we should use `await` or `then` here to make sure that the promise returned from `ViewStatesActions.update` and its handlers are resolved before checking anything. I think that is also the case in the other test cases in this file.

linuspahl: I get your point, but could not find a similar approach for the other test cases. I've implemented `async/await` for all cases.

graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
```diff
@@ -100,7 +100,42 @@ describe('CurrentViewStateStore', () => {
...
+      expect(id).toEqual(viewId);
+      expect(newViewState).toEqual(expectedViewState);
+      return Promise.resolve(expectedViewState);
+    }));
+
+    ViewStatesActions.update = updateFn;
+    CurrentViewStateStore.onViewStoreChange({ activeQuery: viewId, view: existingViewState });
+    CurrentViewStateStore.onViewStatesStoreChange(sMap);
+    await CurrentViewStateStore.widgets(expectedWidgets);
     expect(updateFn).toHaveBeenCalledTimes(1);
```
dennisoelkers: We can also change this to:
```
expect(updateFn).toHaveBeenCalledWith(viewId, expectedViewState);
```
and just turn the definition of `updateFn` into:
```
const updateFn = jest.fn(() => Promise.resolve(expectedViewState));
```

Which makes the actual expectation easier to read because the test case is way more linear (not having to jump from the `expect` back to the definition of `updateFn`.

graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
```diff
@@ -100,7 +100,42 @@ describe('CurrentViewStateStore', () => {
     ViewStatesActions.update = updateFn;
     CurrentViewStateStore.onViewStoreChange({ activeQuery: viewId, view: oldViewState });
     CurrentViewStateStore.onViewStatesStoreChange(sMap);
-    CurrentViewStateStore.widgets(expectedWidgets);
+    await CurrentViewStateStore.widgets(expectedWidgets);
+    expect(updateFn).toHaveBeenCalledTimes(1);
+  });
+
+  it('should remove widget positions for deleted widgets', async () => {
+    const widgetOnePos = new WidgetPosition(1, 1, 5, 6);
```
dennisoelkers: Imo the following variable declarations could be changed into one big inline declaration of the view state. We do not reuse these values, so we do not really need to capture them into variables.

graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
```diff
@@ -100,7 +100,42 @@ describe('CurrentViewStateStore', () => {
...
+      .build();
+
+    const updateFn = mockAction(jest.fn((id, newViewState) => {
+      expect(id).toEqual(viewId);
+      expect(newViewState).toEqual(expectedViewState);
+      return Promise.resolve(expectedViewState);
+    }));
+
+    ViewStatesActions.update = updateFn;
+    CurrentViewStateStore.onViewStoreChange({ activeQuery: viewId, view: existingViewState });
```
dennisoelkers: Setup of the test case and the actual action(s) which are tested should be clearer, by adding new lines in between. Imo assigning `ViewStatesActions.update` is still part of the setup, I am not sure about this line. I think the actual action under test is `CurrentViewStateStore.widgets`, so it should be visually separated from the setup and checking the expectations afterwards.

linuspahl:
@dennisoelkers Thanks for the good feedback. I've implemented the suggestions for the new and existing test cases.

graylog2-web-interface/src/views/stores/CurrentViewStateStore.test.js
linuspahl: I have to reuse the widget and widget position for the expected view state. I still prefer the inline declaration and just used `existingViewState.widgets[0]` for the expected state. is this implementation okay or an anti pattern? I can imagine not reusing the "initial state" for the "expected state", to be a best practice.

PR #7145, opened 2020-01-10
Zoom global override when zooming in timeline chart on dashboard.

dennisoelkers:
## Motivation and Context

Prior to this change, zooming (dragging a range) on a timeline chart on
a dashboard did nothing, because it set the query's time range which is
not considered on dashboards.

This change is now conditionally setting the global override's time
range if on a dashboard and the query's time range if not. It does this
by checking the view's type in the zoom handler.

Ideally we should instead provide a `DrilldownContext` for all actions,
that provides commonly used actions/states (setting/getting current time
range, query, stream selection), so actions do not determine on their
own which one to pick.

## Types of changes

## Checklist:


dennisoelkers:
Good catch, fixed that. Thanks!

PR #7159, opened 2020-01-13
Implement overlay for select menu

linuspahl:
The current dropdown menu of the `Select` component sometimes gets cut off and some options are no longer selectable. Here is an example of the aggregation builder:

image

This PR implements the `Overlay` from `react-overlay` to display the menu independently from its parent containers.

image

The new menu wrapper also defines a minimum width of 200px, until now a dropdown could be very small, depending on the parent container.
By default the menu has still the width of the related select input.

With these changes we are also breaking long option names.
Before (master):
image
The user has to scroll horizontally to read the full name, the style is broken, even the list items have a width of 100%. 

After:
image

In my opinion the second solution it easier to read. Even when long option names, separated by underscores are not getting perfectly wrapped. What do you think?

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/Select.jsx
+++ b/graylog2-web-interface/src/views/components/Select.jsx
@@ -1,7 +1,10 @@
 // @flow strict
-import * as React from 'react';
+import React, { useRef, useMemo } from 'react';
+import type { Node, ElementType } from 'react';
 import PropTypes from 'prop-types';
+import { get } from 'lodash';
 import ReactSelect, { components as Components } from 'react-select';
+import { Overlay } from 'react-overlays';
 
 const MultiValueRemove = (props) => {
   return (
@@ -9,6 +12,20 @@ const MultiValueRemove = (props) => {
       ×
     </Components.MultiValueRemove>
   );
+};
+
+const OverlayInner = ({ children, style }: {children: Node, style?: Object}) => React.Children.map(children,
+  child => React.cloneElement(child, { style: Object.assign({}, style, child.props.style) }));
+
+const menu = (selectRef) => {
+  const defaultMinWidth = 200;
+  const containerWidth = get(selectRef, 'current.select.controlRef.offsetWidth');
+  const width = containerWidth > defaultMinWidth ? containerWidth : defaultMinWidth;
+  return base => ({
+    ...base,
+    position: 'relative',
+    width,
+  });
 };
 
 const valueContainer = base => ({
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/views/components/Select.jsx
```diff
@@ -62,9 +99,10 @@ const Select = ({ components, styles, ...rest }: Props) => {
     multiValueLabel,
     multiValueRemove,
     valueContainer,
+    menu: menu(selectRef),
```
linuspahl: Would it make sense to implement `useMemo` for this function call as well?

graylog2-web-interface/src/views/components/Select.jsx
```diff
@@ -11,6 +14,20 @@ const MultiValueRemove = (props) => {
   );
 };
 
+const OverlayInner = ({ children, style }: {children: Node, style?: Object}) => React.Children.map(children,
+  child => React.cloneElement(child, { style: Object.assign({}, style, child.props.style) }));
+
+const menu = (selectRef) => {
+  const defaultMinWidth = 200;
```
linuspahl: Should I make this min width optional?
Right now I could not find a use case where the min width of `200px` is not suitable.

graylog2-web-interface/src/views/components/Select.jsx
```diff
@@ -11,6 +14,20 @@ const MultiValueRemove = (props) => {
...
+  const containerWidth = get(selectRef, 'current.select.controlRef.offsetWidth');
+  const width = containerWidth > defaultMinWidth ? containerWidth : defaultMinWidth;
+  return base => ({
+    ...base,
+    position: 'relative',
+    width,
```
linuspahl: With this implementation we are defining the width without a unit. The `px` info gets added automatically. Do you prefer adding `px` manually with a template string, to be more explicit?

kyleknighted: I think being explicit with the width is a good thing `50` does not tell me what width it will be... is that `px`, `vw`, `em`, etc

kyleknighted:
image

This doesn't look great...

kyleknighted:
image


Text is cut off on this Select

linuspahl:
@kyleknighted Thanks for having a look. My main focus was on the dropdown.

Regarding your screenshot of the series select, I've implemented this layout on Purpose, due to the following problem. 
https://github.com/Graylog2/graylog2-server/pull/6748
I could not find a better solution, do you have any suggestions?

Regarding the placeholder overflow, this bothered me for a long time as well, I've created another PR to fix this issue:
https://github.com/Graylog2/graylog2-server/pull/7170

graylog2-web-interface/src/views/components/Select.jsx
kmerz: I think it is fine that way. I would wait for the need.

PR #7177, opened 2020-01-14
Synchronize timerange & query with URL on search page.

### LINKED ISSUE #7144: Search state is not reflected in history/URL
## Expected Behavior
When changing the query text, reload interval or time frame, I expect the URL and browser history to reflect that state so that both reload and back/forward navigation works.
This was also the behavior of the previous search pages.

## Current Behavior
Neither URL nor history state reflects any changes made on the page, which causes using the browser navigation or reload to lose the user's state.

## Possible Solution
Add a limited amount of state to the URL hash and use it to populate the store's state when loading the page initially. This would also allow quick copy/paste sharing of simple searches (OTOH actually sharing it is obviously out of scope here).

1. Navigate to "System" menu.
2. Navigate to "Search".
3. Enter query string "source:example".
4. Use browser back button.
This then takes you back to the system menu instead of the empty search.

Other changes that demonstrate this behavior include time range selection and stream selection.

## Context
Making simple changes about the search is one of the tasks users perform the most and to have no means of easily navigating back and forth makes it very difficult to use.
Additionally, this is a direct breaking change to how the search page behaved before, which means a lot of people will struggle with it.

## Your Environment

* Graylog Version: 3.2.0-beta.2

dennisoelkers:
## Motivation and Context

The current search has a feature, where the current time range and query
string is synchronized with the URL in the browser. This means that:

  - current time range and query string are part of the URL and copying
and replaying the URL allows restoring the state for these
  - a change to time range or query string creates a new state in the
history stack, allowing the user to go back and forward between
different states
  - going back or forward in the browser history restores the time range
& query string of this specific state

This PR is recreating this functionality for the new search.

Fixes #7144.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
+++ b/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
@@ -29,7 +29,11 @@ const extractTimerangeParams = (timerange) => {
 export const syncWithQueryParameters = (query: string) => {
   const { view } = ViewStore.getInitialState() || {};
   if (view && view.type === View.Type.Search) {
-    const firstQuery = view.search.queries.first();
+    const { queries } = view.search;
+    if (queries.size !== 1) {
+      throw new Error('Searches must only have a single query!');
+    }
+    const firstQuery = queries.first();
     if (firstQuery) {
       const { query: { query_string: queryString }, timerange } = firstQuery;
       const baseUri = new URI(query).setSearch('q', queryString)
--- a/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
+++ b/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
@@ -27,8 +27,8 @@ const extractTimerangeParams = (timerange) => {
 };
 
 export const syncWithQueryParameters = (query: string) => {
-  const { view } = ViewStore.getInitialState();
-  if (view.type === View.Type.Search) {
+  const { view } = ViewStore.getInitialState() || {};
+  if (view && view.type === View.Type.Search) {
     const firstQuery = view.search.queries.first();
     if (firstQuery) {
       const { query: { query_string: queryString }, timerange } = firstQuery;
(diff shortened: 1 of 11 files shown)


graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
```diff
@@ -0,0 +1,53 @@
...
+    case 'absolute': return Object.entries({ ...result, from: timerange.from, to: timerange.to });
+    default: return Object.entries(result);
+  }
+};
+
+export const syncWithQueryParameters = (query: string) => {
+  const { view } = ViewStore.getInitialState() || {};
+  if (view && view.type === View.Type.Search) {
+    const firstQuery = view.search.queries.first();
+    if (firstQuery) {
```
linuspahl: Out of interest, is there a reason you are checking for the first query here, but take it for granted in `BindSearchParamsFromQuery.js`?

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
```diff
@@ -112,14 +122,30 @@ const DashboardSearchBarWithStatus = WithSearchStatus(DashboardSearchBar);
...
+  const { pathname, search } = router.getCurrentLocation();
+  const query = `${pathname}${search}`;
   const refreshIfNotUndeclared = () => _refreshIfNotUndeclared(searchRefreshHooks, SearchExecutionStateStore.getInitialState());
 
   useEffect(() => {
-    style.use();
+    const { view } = ViewStore.getInitialState();
 
+    bindSearchParamsFromQuery({ view, query: location.query, retry: () => Promise.resolve() });
+  }, [query]);
```
linuspahl: One thing I realised while testing these changes:

Changing e.g. the relative time range, results in a new URL due to `useSyncWithQueryParameters`. because of the new URL, the query variable is different and `bindSearchParamsFromQuery ` gets called again.

`bindSearchParamsFromQuery` will unnecessarily update the current query and also trigger another, but unnecessary search execution.

graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
dennisoelkers: Not really. Added the check to `SyncWithQueryParameters` too.

PR #7178, opened 2020-01-14
Clean up naming of views endpoints in api browser

### LINKED ISSUE #7150: Clean up naming of Views endpoints in API browser
The api browser still names the Views endpoints "Enterprise/Search" and "Enterprise/Views".

## Expected Behavior
Views endpoints should not have the Enterprise prefix anymore.

## Possible Solution
Remove the "Enterprise" prefix from Views endpoints
Add "Legacy" prefix to old search endpoints.

* Graylog Version: 3.2.1.beta3

alex-konn:
Removed "Enterprise/" prefix for new views endpoints and added "Legacy/" for old ones.

Fixes #7150

Left `DecoratorResource` untouched: https://github.com/Graylog2/graylog2-server/blob/c02f86de0e283dc53aa4ff2ef9f14a833b5cbe43/graylog2-server/src/main/java/org/graylog2/rest/resources/search/DecoratorResource.java#L52

It technically is an old resource but views depend on it and we should move it to the `views` package imo.

## Motivation and Context
Views is not an Enterprise feature anymore and we want to clearly mark the old endpoints as legacy to discourage users from depending on them.

## How Has This Been Tested?
Manual review in local dev environment

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/ViewSharingResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/ViewSharingResource.java
@@ -45,7 +45,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-@Api(value = "Views")
+@Api(value = "Views/Sharing")
 @Path("/views/{id}/share")
 @Produces(MediaType.APPLICATION_JSON)
 @RequiresAuthentication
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/ViewSharingResource.java
```diff
@@ -45,7 +45,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
-@Api(value = "Enterprise/Views", description = "View Sharing management")
+@Api(value = "Views")
```
dennisoelkers: I think this should be more precise, maybe "Views Sharing"?

alex-konn: Good point! Will do shortly.

alex-konn: Changed it to "Views/Sharing".

PR #7187, opened 2020-01-15
Add "gl2_accounted_message_size" field to indexed messages

### LINKED ISSUE #6838: Track message size as system message field
## What?
Graylog uses a certain algorithm to keep track of the size of a message, which is the basis for traffic accounting (https://docs.graylog.org/en/3.1/pages/enterprise/setup.html#details-on-license-verification and https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/plugin/Message.java#L569-L571).
This value should be tracked in Graylog as a system level field.

## Why?
In order to find out which sources send how much data to Graylog (for internal accounting, debugging or purely informational purposes), it makes sense to use Graylog's own features to build searches and dashboards instead of implementing analysis functions tailored to only message size accounting.

## Tasks

- [ ] Add `gl2_accounted_message_size` field as `long` to index mapping (so it's not getting auto-created as `keyword`/`string`)
- [ ] Serialize message size into `gl2_accounted_message_size` field when writing to ES
- [ ] Write upgrade notes to add `gl2_accounted_message_size` field to active write index mappings (must be done before upgrading to 3.2)

## Your Environment

* Graylog Version: 3.2

bernd:
That field contains the message size that will be accouted for
Enterprise license checks. It can be used to analyze traffic per stream
or similar use cases.

This change adds the new field to all message index templates and writes
the field to indexed messages.

Closes #6838

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -43,6 +43,10 @@ For each of the index names returned by the previous command, the following comm
 
   curl -s -X PUT --data '{"properties":{"gl2_accounted_message_size":{"type": "long"}}}' -H Content-Type:application/json localhost:9200/<active-write-index-name>/_mapping/message
 
+The two steps could also be combined::
+
+  for index in `curl -s localhost:9200/_cat/aliases/*_deflector?h=index`; do curl -s -X PUT --data '{"properties":{"gl2_accounted_message_size":{"type": "long"}}}' -H Content-Type:application/json localhost:9200/$index/_mapping/message ; done'
+
 The Graylog servers can now be restarted with the 3.2.0 version.
 
 Known Bugs and Limitations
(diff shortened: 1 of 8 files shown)


UPGRADING.rst
```diff
@@ -26,6 +26,25 @@ Indexing Requests use HTTP Expect: 100-Continue Header
...
+All of the following commands need to be executed against one of the Elasticsearch servers in the cluster.
+
+First, a list of all active write indices is needed::
+
+  curl -s localhost:9200/_cat/aliases/*_deflector?h=index
+
+For each of the index names returned by the previous command, the following command needs to be executed (``<active-write-index-name>`` in the URL needs to be replaced with the actual index name)::
+
+  curl -s -X PUT --data '{"properties":{"gl2_accounted_message_size":{"type": "long"}}}' -H Content-Type:application/json localhost:9200/<active-write-index-name>/_mapping/message
+
```
mpfz0r: FWIW, I used:
 ```for index in `curl -s localhost:9200/_cat/aliases/*_deflector?h=index`; do    curl -s -X PUT --data '{"properties":{"gl2_accounted_message_size":{"type": "long"}}}' -H Content-Type:application/json localhost:9200/$index/_mapping/message ; done```

Should we add that?

bernd: I added it. :+1:

PR #7189, opened 2020-01-15
 Add pipeline functions to modify lookup tables

mpfz0r:
This adds the following pipeline functions:

- `lookup_set_value("table-name", "key", "value")`
- `lookup_set_string_list("table-name", "key", ["value1", "value2"])`
- `lookup_add_string_list("table-name", "key", ["value1", "value2"])`
- `lookup_remove_string_list("table-name", "key", ["remove1", "remove2"])`
- `lookup_clear_key("table-name", "key")`

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupAddStringList.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupAddStringList.java
@@ -38,7 +38,7 @@ public class LookupAddStringList extends AbstractFunction<Object> {
     private final ParameterDescriptor<Object, Object> keyParam;
     @SuppressWarnings("rawtypes")
     private final ParameterDescriptor<List, List> valueParam;
-    private final ParameterDescriptor<Boolean, Boolean> appendOption;
+    private final ParameterDescriptor<Boolean, Boolean> keepDuplicates;
 
     @Inject
     public LookupAddStringList(LookupTableService lookupTableService) {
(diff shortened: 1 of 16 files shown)


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupAddStringList.java
```diff
@@ -0,0 +1,89 @@
...
+                .description("The existing lookup table to use to add the given list")
+                .transform(tableName -> lookupTableService.newBuilder().lookupTable(tableName).build())
+                .build();
+        keyParam = object("key")
+                .description("The key to add in the lookup table")
+                .build();
+        valueParam = ParameterDescriptor.type("list_value", List.class)
+                .description("The list value that should be added into the lookup table")
+                .build();
+        appendOption = bool("append_values")
```
bernd: What do you think about naming this (and also in other parts of the backend code) `keep_duplicates`? From reading the parameter name, I wasn't really sure what it meant.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupAddStringList.java
```diff
@@ -0,0 +1,89 @@
...
+                .optional()
+                .description("Append values instead of merging them with existing ones into a set. Defaults to false (merge)")
+                .build();
+    }
+
+    @Override
+    public Object evaluate(FunctionArgs args, EvaluationContext context) {
+        Object key = keyParam.required(args, context);
+        if (key == null) {
+            return null;
```
bernd: What do you think about returning an empty `LookupResult` instead of `null`s here and below? Since we don't have conditionals in the rule language, returning `null` might result in exceptions if the return value will be used in the rule.

Returning an empty or even an error `LookupResult` should avoid that because we always return a value.

mpfz0r: Actually, the pipeline simulator reports NPEs rather nicely.
I've just copied existing behaviour:
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupValue.java#L58

Should I change this here as well?

bernd: Hmm, okay. Then let's keep the `null` return values for consistency. :+1:

mpfz0r:
lookup_clear_key changes LGTM :+1:

PR #7199, opened 2020-01-16
Add support to dump the current ProcessBuffer state

mpfz0r:
Introduce a new API call that returns the messages
that are currently processed in each ProcessBufferProcessor.

This is useful to debug cases where message processing got stuck.
The most common case is the use of an inefficient regular expression
in extractors or pipeline rules.
Understanding which regular expression caused the problem can be
difficult, without knowing which message triggered the problem.

Example:

- Set up pipeline rule with an evil backtracking regex like `/(x+x+)+y/`
- Send message that spends hours in processing.

- GET https://172.16.1.1:9000/api/system/processbufferdump

```
{
  "processbuffer_dump": {
    "ProcessBufferProcessor #4": "idle",
    "ProcessBufferProcessor #2": "MessageEvent{raw=null, message=source: t480 | message: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx { application_name: mpf | level: 5 | gl2_remote_ip: 127.0.0.1 | timestamp: 2020-01-16T11:00:48.296+01:00 } [...] }",
    "ProcessBufferProcessor #3": "idle",
    "ProcessBufferProcessor #0": "idle",
    "ProcessBufferProcessor #1": "idle"
  }
}
```


graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
```diff
@@ -56,6 +57,7 @@
     private final ULID ulid;
     private final DecodingProcessor decodingProcessor;
     private final Provider<Stream> defaultStreamProvider;
+    private MessageEvent currentEvent;
```
bernd: I think we should make this variable `volatile` to ensure cross-thread visibility.

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
```diff
@@ -75,11 +77,13 @@ public ProcessBufferProcessor(MetricRegistry metricRegistry,
         incomingMessages = metricRegistry.meter(name(ProcessBufferProcessor.class, "incomingMessages"));
         outgoingMessages = metricRegistry.meter(name(ProcessBufferProcessor.class, "outgoingMessages"));
         processTime = metricRegistry.timer(name(ProcessBufferProcessor.class, "processTime"));
+        currentEvent = null;
     }
 
     @Override
     public void onEvent(MessageEvent event) throws Exception {
         try {
+            currentEvent = event;
```
bernd: How about setting (and clearing) this in `#dispatchMessage()` instead? That way we only get the single message. The `MessageEvent` could contain one or more messages.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemResource.java
```diff
@@ -116,5 +117,26 @@ public SystemThreadDumpResponse threadDump(@ApiParam(name = "nodeId", value = "T
             throw new WebApplicationException(response.message(), BAD_GATEWAY);
         }
     }
+
+    @GET
+    @Timed
+    @ApiOperation(value = "Get a process buffer dump of the given node")
+    @RequiresPermissions(RestPermissions.PROCESSBUFFER_DUMP)
+    @Path("{nodeId}/processbufferdump")
```
bernd: What do you think about adding another endpoint that returns the dumps for all nodes? That would make it way easier to use. We have `ProxiedResource#getForAllNodes()` for that. Example:

https://github.com/Graylog2/graylog2-server/blob/cb33786987e845aed42956bed67a335c2b91471c/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java#L74

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
mpfz0r: Good catch :+1:

graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java
mpfz0r: I agree. The downside is that we wouldn't see problems within a `DecodingProcessor` but that's less unlikely.

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemResource.java
mpfz0r: Nice. Thanks. BTW, I think we should also add this to get all metrics (`org.`) from the cluster

lennartkoopmann:
This is currently immensely helpful with a customer issue. It would be even better if we had the full message in the dump. We are calling `toString()` on the message objects, which truncates the `message` field after 250 characters. To debug a message against a possibly faulty regular expression, we need the complete message though.

See also:
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/plugin/Message.java#L437-L441

PR #7206, opened 2020-01-16
Add mapping of stream ids in view content packs

### LINKED ISSUE #7180: ViewFacade does not map stream ids to native entities
## Expected Behavior
When installing a content pack with views referencing to streams, the content pack installation
should map the stream ids of the content pack view with the new installed streams from the content pack.

## Current Behavior
The facade does not do the mapping. This leads to an error when installing dashboards referting
to streams.

## Possible Solution
`toNativeEntity` must implement the mapping.

## Your Environment
* Graylog Version: Graylog 3.2.beta.2

kmerz:
## Description
    This change adds this mapping to the Widget and WidgetEntit
    for the DashboardWidget.
    For the search types there are now Entities so we can easily
    map the streams during creation and installation of the
    content pack.
    For the query we needed to add a generic Builder access
    for the Filter so we could change the Stream id for
    a StreamFilter and rebuild the Filter refered in the
    QueryEntity and Query.

## Motivation and Context
    Prior this change, when creating a content pack containing
    views with streams the stream was collected as an dependency
    into the content pack but the StreamEntity got a new id.
    This id needs to be used in all references of the stream.
    Therefore also in the ViewEntity where a streamid is persisted.
    Also when installing this content pack the installed
    stream has a new ID yet again and the view needs to know
    this as well.

Fixes #7180 

## How Has This Been Tested?
- Created a search and dashboard view with refering streams content pack.
- Removed Stream and View.
- Installed Content Pack

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/Query.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/Query.java
@@ -32,6 +32,10 @@ import org.graylog.plugins.views.search.engine.BackendQuery;
 import org.graylog.plugins.views.search.engine.EmptyTimeRange;
 import org.graylog.plugins.views.search.filter.AndFilter;
 import org.graylog.plugins.views.search.filter.StreamFilter;
+import org.graylog2.contentpacks.ContentPackable;
+import org.graylog2.contentpacks.EntityDescriptorIds;
+import org.graylog2.contentpacks.model.ModelTypes;
+import org.graylog2.contentpacks.model.entities.QueryEntity;
 import org.graylog2.plugin.indexer.searches.timeranges.TimeRange;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -43,6 +47,7 @@ import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
@@ -53,7 +58,7 @@ import static java.util.stream.Collectors.toSet;
 @JsonAutoDetect
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonDeserialize(builder = Query.Builder.class)
-public abstract class Query {
+public abstract class Query implements ContentPackable<QueryEntity> {
     private static final Logger LOG = LoggerFactory.getLogger(Query.class);
 
     @JsonProperty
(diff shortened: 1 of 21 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/Query.java
```diff
@@ -242,4 +247,36 @@ public Query build() {
...
         }
     }
+
+    // This code assumes that we only add streams via gui shallow.
+    private Filter shallowMappedFilter(EntityDescriptorIds entityDescriptorIds) {
+        return Optional.ofNullable(filter())
+                .map(optFilter -> {
+                    Set<Filter> newFilters = optFilter.filters().stream()
+                            .map(filter -> {
+                                if (filter.type().matches(StreamFilter.NAME)) {
```
mpfz0r: why not `.equals(StreamFilter.NAME)`?

graylog2-server/src/test/java/org/graylog2/contentpacks/facades/ViewFacadeTest.java
```diff
@@ -62,15 +70,18 @@
...
 import org.graylog2.shared.bindings.providers.ObjectMapperProvider;
+import org.graylog2.streams.StreamImpl;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 
+import javax.swing.text.html.Option;
```
mpfz0r: This isn't used

graylog2-server/src/main/java/org/graylog/plugins/views/search/Query.java
```diff
@@ -242,4 +247,36 @@ public Query build() {
             return autoBuild();
         }
     }
+
+    // This code assumes that we only add streams via gui shallow.
```
mpfz0r: I'd make this a TODO:
Suggestion: 
// TODO This code assumes that we only use shallow filters for streams. If this ever changes, we need to implement a mapper that can handle filter trees.

graylog2-server/src/main/java/org/graylog2/contentpacks/model/entities/QueryEntity.java
```diff
@@ -0,0 +1,168 @@
...
+        static Builder createWithDefaults() {
+            return new AutoValue_QueryEntity.Builder().searchTypes(of());
+        }
+
+        public QueryEntity build() {
+            return autoBuild();
+        }
+    }
+
+    // This code assumes that we only add streams via gui shallow.
```
mpfz0r: Same as above

PR #7214, opened 2020-01-17
Improving search controls by adding text and dropdown, fixing reset.

### LINKED ISSUE #7162: "Empty search" button does nothing
The "Empty search" button is doing nothing for transient searches. This is confusing.

## Expected Behavior
I think it should actually reset to an empty transient search.

## Current Behavior
If no saved search is opened, the "Empty search" button actually does nothing although it's clickable.
empty-search-button

After opening a saved search it actually "unloads" the saved search and resets the screen to an empty default search. That's a useful feature, but "Empty search" to me implies modifying the saved search instead.

## Possible Solution
- Change the label (maybe to "Restore Default Search" or something, not sure)
- Implement restoring the default also for a modified transient search

1. Navigate to "Search"
2. Enter a query
3. Click "Empty search"
4. Stare into void

* Graylog Version: 3.2.beta2
* Browser version: Firefox 72.0.1

### LINKED ISSUE #7168: Replace Search glyph controls with mixed glyphs/dropdown with text
Currently the search controls (for saving/loading/clearing/export a search) consist of a button group with icons. These are unclear, so most users have to wait for tooltips to appear. 

Instead, we should use a mixed glyph/text variant, which renders clear icons for the most common actions (saving, loading the current search, using a star/folder glyph and "save" / "load" text) and a "More" button opening a dropdown showing text items for other actions (clearing search, exporting to dashboard, ...)

dennisoelkers:
## Motivation and Context

Prior to this change, the search controls consisted of icons only. This
required the user to remember their meaning or waiting for the tooltips
to appear.

This change is aiming to improve this by:

  - Adding text to the load/save buttons
  - Putting other options into a dropdown and adding text
  - Renaming "Empty search" to "Reset search"
  - Fixing the "Reset search" functionality

Fixes #7168.
Fixes #7162.

Screen Shot 2020-01-16 at 17 23 16

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/logic/queries/Query.js
+++ b/graylog2-web-interface/src/views/logic/queries/Query.js
@@ -32,6 +32,22 @@ export type ElasticsearchQueryString = {
 
 export const createElasticsearchQueryString = (query: string = ''): ElasticsearchQueryString => ({ type: 'elasticsearch', query_string: query });
 
+const _streamFilters = (selectedStreams: Array<string>): Array<{ type: string, id: string }> => {
+  return selectedStreams.map(stream => ({ type: 'stream', id: stream }));
+};
+
+export const filtersForQuery = (streams: ?Array<string>) => {
+  if (!streams || streams.length === 0) {
+    return null;
+  }
+  const streamFilters = _streamFilters(streams);
+  return {
+    type: 'or',
+    filters: streamFilters,
+  };
+};
+
+
 export type QueryString = ElasticsearchQueryString;
 
 export type TimeRangeTypes = 'relative' | 'absolute' | 'keyword';
--- a/graylog2-web-interface/src/views/logic/queries/Query.js
+++ b/graylog2-web-interface/src/views/logic/queries/Query.js
@@ -47,7 +47,6 @@ export const filtersForQuery = (streams: ?Array<string>) => {
   };
 };
 
-
 export type QueryString = ElasticsearchQueryString;
 
 export type TimeRangeTypes = 'relative' | 'absolute' | 'keyword';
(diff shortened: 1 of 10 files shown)


linuspahl:
The functionality of the reset button was broken after merging https://github.com/Graylog2/graylog2-server/pull/7177
@kmerz and I had a look at possible solutions. We found one for most use cases, but still need to have another look at some related topics:
- The `ViewLoaderContext` is needed to load a saved search. By default it loads the new view and changes the url. For all pages (excluding the `ShowViewPage`) it is enough to do a redirect, because the `ShowViewPage` will load the view.
- Using the browser navigation works, but not perfectly. E.g. clicking on the search page will create an entry in the history for `/search` and `/search?q=&rangetype=relative&relative=300`
This is also the case when loading a saved search or clicking on a stream. We should find a way to update the URL query params, without creating an entry in the browser history, while still creating an entry, when e.g. just changing the time range of the current search.

linuspahl:
@kyleknighted The reset button is now disabled, until the query options have changed. 

Regarding the refresh, this happens only in the context of saved searches (as you mentioned also on load / create), I see the benefit of these improvements, but would like to do them in another PR.

graylog2-web-interface/src/views/logic/queries/Query.js
```diff
@@ -32,6 +32,22 @@ export type ElasticsearchQueryString = {
 
 export const createElasticsearchQueryString = (query: string = ''): ElasticsearchQueryString => ({ type: 'elasticsearch', query_string: query });
 
+const _streamFilters = (selectedStreams: Array<string>): Array<{ type: string, id: string }> => {
+  return selectedStreams.map(stream => ({ type: 'stream', id: stream }));
+};
+
+export const filtersForQuery = (streams: ?Array<string>) => {
```
linuspahl: I am not so sure where the best place for this helper function is. I can't use the `QueriesFilterStore`, due to a dependency cycle.

PR #7215, opened 2020-01-17
Improve form validation for dashboard full screen config modal

### LINKED ISSUE #7211: Dashboard full screen form allows negative integers for refresh and tab cycle interval
When setting up the full screen mode for a dashboard, a user can enter negative integers for the refresh and cycle interval. The full screen mode still kind of works, but the first time I tried this, the server was in a loop where it never stopped calling the elastic search instance. Unfortunately I could not reproduce this behaviour afterwards.

image

Another little thing I noticed while testing this form, adding an invalid value, e.g. by just entering a minus after the number results in a console error:
```
Warning: Received NaN for the `value` attribute. If this is expected, cast the value to a string.
    in input (created by FormControl)
    in FormControl (created by Context.Consumer)
    in StyledComponent (created by FormControl)
```

## Expected Behavior
Negative values should not be possible. Ideally the values should also be validated when changing the full screen mode url params.

## Steps to Reproduce
1. Create a dashboard
2. Start the full screen mode (Actions -> Full Screen)
3. Enter negative values

## Your Environment

linuspahl:
This PR fixes the behaviour described in #7211 (allowing negative numbers for refresh and cycle interval) regarding the full screen dashboard config modal, mainly by adding a HTML form element.

I had to adjust the form submit inside the related tests, because submitting the form by triggering a click on the save button results in the following error:
`Not implemented: HTMLFormElement.prototype.submit`.

Fixes #7211

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/dashboard/BigDisplayModeConfiguration.jsx
+++ b/graylog2-web-interface/src/views/components/dashboard/BigDisplayModeConfiguration.jsx
@@ -47,50 +47,56 @@ const ConfigurationModal = ({ onSave, onCancel, view }: ConfigurationModalProps)
 
   return (
     <Modal show bsSize="large" onHide={onCancel}>
-      <Modal.Header closeButton>
-        <Modal.Title>Configuring Full Screen</Modal.Title>
-      </Modal.Header>
-      <Modal.Body>
-        <Input id="refresh-interval"
-               type="number"
-               name="refresh-interval"
-               label="Refresh Interval"
-               help="After how many seconds should the dashboard refresh?"
-               onChange={({ target: { value } }) => setRefreshInterval(Number.parseInt(value, 10))}
-               required
-               value={refreshInterval} />
+      <form onSubmit={_onSave}>
+        <Modal.Header closeButton>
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/dashboard/BigDisplayModeConfiguration.jsx
```diff
@@ -47,50 +47,56 @@ const ConfigurationModal = ({ onSave, onCancel, view }: ConfigurationModalProps)
...
+          <Modal.Title>Configuring Full Screen</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <Input id="refresh-interval"
+                 type="number"
+                 min="1"
+                 name="refresh-interval"
+                 label="Refresh Interval"
+                 help="After how many seconds should the dashboard refresh?"
+                 onChange={({ target: { value } }) => setRefreshInterval(value ? Number.parseInt(value, 10) : value)}
```
kmerz: When the value falsy is you set the refreshInterval to that? Why not to 0 or a other default value?

PR #7220, opened 2020-01-17
Fix Alerting Lookup Table Parameters UI

edmundoa:
This PR makes some improvements in the UI around support for Lookup Table Parameters in Filter & Aggregation Alerts, and also removes two warnings from the developer console:
- `Warning: Failed prop type: The prop `children` is marked as required in `<<anonymous>>`, but its value is ``null``.`
- `Warning: validateDOMNesting(...): <h5> cannot appear as a descendant of <p>.`

Before:
Screenshot 2020-01-17 at 15 11 05

After:
Screenshot 2020-01-20 at 12 26 10
Screenshot 2020-01-20 at 12 36 41


edmundoa:
It looks like this still needs some more work.

PR #7222, opened 2020-01-17
Reset time in SearchesCleanUpJobWithDBServicesTest teardown method

bernd:
This adds a missing `DateTimeUtils.setCurrentMillisSystem()` to avoid
a sporadic test failure in `DBJobTriggerServiceTest`.

The `SearchesCleanUpJobWithDBServicesTest` is using
`DateTimeUtils.setCurrentMillisFixed()` to set the time to a specific
value. This requires the usage of `DateTimeUtils.setCurrentMillisSystem()`
to reset the time to the current time. Otherwise tests running in the
same JVM using the current time might fail.

I noticed that the `DBJobTriggerServiceTest#deleteCompleted` test method
failed from time to time on a new Jenkins worker machine. This is
probably due to a changed execution order because the machine is faster
than the older worker machine. Still, strange that it didn't show up
earlier.

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/plugins/views/search/QueryEffectiveTimeRangeTest.java
+++ b/graylog2-server/src/test/java/org/graylog/plugins/views/search/QueryEffectiveTimeRangeTest.java
@@ -23,6 +23,7 @@ import org.graylog2.plugin.indexer.searches.timeranges.InvalidRangeParametersExc
 import org.graylog2.plugin.indexer.searches.timeranges.RelativeRange;
 import org.graylog2.plugin.indexer.searches.timeranges.TimeRange;
 import org.joda.time.DateTimeUtils;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


bernd:
> While this looks good, I found a few more tests that don't reset this properly.
> `org.graylog.plugins.views.search.elasticsearch.searchtypes.pivot.ESPivotTest`
> `org.graylog.plugins.views.search.elasticsearch.ElasticsearchBackendUsingCorrectIndicesTest`
> `org.graylog.plugins.views.search.QueryEffectiveTimeRangeTest`
> 
> Shouldn't we fix those as well?

@mpfz0r I checked the tree already and didn't find any tests that don't reset the time. The ones you mentioned all reset the time via `DateTimeUtils.setCurrentMillisSystem();` at the end of the test.

mpfz0r:
> @mpfz0r I checked the tree already and didn't find any tests that don't reset the time. The ones you mentioned all reset the time via `DateTimeUtils.setCurrentMillisSystem();` at the end of the test.

But what happens if their assertion fails?

bernd:
> But what happens if their assertion fails?

Yeah, good point. I will update the PR to move this into an `@After` block. Good catch! :+1:

PR #7228, opened 2020-01-20
Hide Pagination in PaginatedList when context is not interactive

### LINKED ISSUE #7210: Message list pagination is not working in dashboard full screen mode
The message list pagination is not working, when a dashboard is in full screen mode.
Clicking on a page results in an empty list.

## Possible Solution
Disabling the pagination is in my opinion the best solution. Mainly because no widget is interactive in the full screen mode.

1. Create a dashboard
2. Create a message list widget (which displays pagination for the current search)
3. Save the dashboard
4. Start the full screen mode, by clicking on Actions -> Full Screen
5. Use the message list pagination

## Your Environment

linuspahl:
As described in #7210, the MessageList pagination does not work in the dashboard full screen mode. This PR hides the pagination inside the PaginatedList, when the `InteractiveContext` defines the context as not interactive. We can safely add this to the shared component because the context is by default interactive.

I would prefer to disable the pagination, but the bootstrap component does not support this functionality and we would have to implement all the styles and logic on our own.

Fixes #7210

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/PaginatedList.jsx
+++ b/graylog2-web-interface/src/components/common/PaginatedList.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import * as React from 'react';
 import { Pagination } from 'components/graylog';
 import { Input } from 'components/bootstrap';
+import InteractiveContext from 'views/components/contexts/InteractiveContext';
 
 const defaultPageSizes = [10, 50, 100];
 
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/common/PaginatedList.jsx
```diff
@@ -114,23 +115,29 @@ class PaginatedList extends React.Component<Props, State> {
...
-                      maxButtons={10}
-                      activePage={currentPage}
-                      onSelect={this._onChangePage}
-                      prev
-                      next
-                      first
-                      last />
-        </div>
-      </>
+      <InteractiveContext.Consumer>
```
kmerz: @dennisoelkers has made a shortcut for that. `<IfInteractive>` 

Take a look here: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/views/components/widgets/Widget.jsx#L291

PR #7234, opened 2020-01-21
Bump set-value to 2.0.1

edmundoa:
Fixes security issue in set-value:
https://github.com/advisories/GHSA-4g88-fppr-53pp


edmundoa:
The upgrade needs to also happen on plugins, will open PRs for that shortly.

PR #7259, opened 2020-01-23
Introduce MongoDBUpsertRetryer

### LINKED ISSUE #7258: MongoDB upserts need to be retried on failure
MongoDB upserts are not happen atomically and should be retried on failures.
This is described here https://jira.mongodb.org/browse/SERVER-14322
and only fixed in mongodb version >= 4.2  with https://jira.mongodb.org/browse/SERVER-37124

mpfz0r:
This allows us to wrap upsert calls, which will
be retried once more if they produce a DuplicateKeyException.

Fixes #7258

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/MongoDBUpsertRetryer.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/MongoDBUpsertRetryer.java
@@ -17,31 +17,49 @@
 
 package org.graylog2.database;
 
+import com.github.rholder.retry.Attempt;
 import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.RetryListener;
 import com.github.rholder.retry.Retryer;
 import com.github.rholder.retry.RetryerBuilder;
 import com.github.rholder.retry.StopStrategies;
 import com.mongodb.DuplicateKeyException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
+/**
+ * MongoDB upsert requests can fail when they are creating a new entry concurrently.
+ * https://jira.mongodb.org/browse/SERVER-14322
+ * This helper can be used to retry upserts if they throw a {@link DuplicateKeyException}
+ */
 public class MongoDBUpsertRetryer {
+    private static final Logger LOG = LoggerFactory.getLogger(MongoDBUpsertRetryer.class);
+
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/database/MongoDBUpsertRetryer.java
```diff
@@ -0,0 +1,47 @@
...
+import com.github.rholder.retry.RetryException;
+import com.github.rholder.retry.Retryer;
+import com.github.rholder.retry.RetryerBuilder;
+import com.github.rholder.retry.StopStrategies;
+import com.mongodb.DuplicateKeyException;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+
+public class MongoDBUpsertRetryer {
```
bernd: Can you please add a comment to this class that explains why we built this? (including relevant links) Thank you!

graylog2-server/src/main/java/org/graylog2/database/MongoDBUpsertRetryer.java
```diff
@@ -0,0 +1,47 @@
...
+                .build();
+        try {
+            return retryer.call(c);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
+        } catch (RetryException e) {
+            if (e.getCause() instanceof DuplicateKeyException) {
+                throw (DuplicateKeyException) e.getCause();
+            }
+            throw new RuntimeException(e);
```
bernd: Should we use `e.getCause()` here as well? Otherwise we will put the `RetryException` as cause into the `RuntimeException`.

graylog2-server/src/main/java/org/graylog2/database/MongoDBUpsertRetryer.java
```diff
@@ -0,0 +1,47 @@
...
+public class MongoDBUpsertRetryer {
+    public static <T> T run(Callable<T> c) {
+        final Retryer<T> retryer = RetryerBuilder.<T>newBuilder()
+                .retryIfException(t -> t instanceof DuplicateKeyException && ((DuplicateKeyException) t).getErrorCode() == 11000)
+                .withStopStrategy(StopStrategies.stopAfterAttempt(1))
+                .build();
+        try {
+            return retryer.call(c);
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e);
```
bernd: Should we use `e.getCause()` here to make sure we don't put the `ExecutionException` as cause into the `RuntimeException`?

PR #7275, opened 2020-01-24
UrlWhitelist Contextual UI

### LINKED ISSUE #7169: Provide means to directly modify URL Whitelist when configuring URLs
## What?
Instead of a simple text input field for URLs that need whitelisting, create a UI form component which will allow the user to directly create or modify a Whitelist entry for the URL being entered. Ideally the user would also be provided with some means of testing the entered URL against the Whitelist, while an entry is modified.

The option to modify the whitelist should only be displayed if the user has the respective permissions.

This task was originally a part of #6853 

## Why?
When a Graylog entity is created/modified which requires a URL to be entered in the UI and that URL is not whitelisted, the user gets a form error. There is no convenient way to then whitelist that URL and make the error go away. The User will have to navigate to the Configuration page and modify the Whitelist accordingly.
It would help a lot if the Whitelist could be modified directly, e.g. by using a button next to the URL input field which will bring up the Whitelist modal.
Also, to make it easier for the user to create regular expression based Whitelist entries, it would be nice to have a "test" button or some other means to check a URL against the regular expression.

ousmaneo:
This PR add a functionality that wrap a normal component to allow having a button to show UrlWhiteListForm in modal on validation error.

Before this change to add a URL to the whitelist, the user had to go back to the configuration page.
With these change a user will be able to add a URL to the whitelist on the context of the form that has a URL field that needs to be validated.

* Implement URLWhitelistInput
* Implement URLWhitelistFormModal

Fixes #7169 

## Types of changes

## Checklist:


bernd:
@thll @ousmaneo 

> Ideally we would always propose a whitelist entry which makes most sense for each context.

Ack, I think this could be an optional callback function in the UI component where we can customize the whitelist type and value. For example: we pass in the URL string and return an object with the suggested entry type and (in the case we need to) and adjusted entry value. (e.g. replacing the `${key}` part in the URL with `.*?`)

ousmaneo:
@bernd, Thanks for the suggestion 
As discussed with @thll, we are working on the suggested. I'm adjusting the component to adapt to the user input and context (e.g. replace the ${key} and define the entry type automatically)

graylog2-web-interface/src/components/configurations/URLWhitelistFormModal.jsx
```diff
@@ -0,0 +1,157 @@
...
+      const { newUrlEntry, formType } = this.props;
+      const initialConfig = { entries: [...urlwhitelistConfig.entries, { id: uuid(), title: '', value: newUrlEntry, type: formType || 'literal' }], disabled: urlwhitelistConfig.disabled };
+      const { entries, disabled } = initialConfig;
+      const { isValid } = this.state;
+      return (
+        <>
+          <IfPermitted permissions="urlwhitelist:write">
+            <Button bsStyle="info" bsSize="xs" onClick={this._openModal}>Add to URL Whitelist</Button>
+          </IfPermitted>
+          <BootstrapModalForm ref={(configModal) => { this.configModal = configModal; }}
```
kyleknighted: Should we use [`createRef`](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) here?

graylog2-web-interface/src/components/event-notifications/event-notification-types/HttpNotificationForm.jsx
```diff
@@ -12,9 +11,7 @@ class HttpNotificationForm extends React.Component {
     onChange: PropTypes.func.isRequired,
   };
 
-  static defaultConfig = {
-    url: '',
-  };
+  ;
```
kyleknighted: Appears to be a floating semicolon

graylog2-web-interface/src/components/configurations/URLWhitelistFormModal.jsx
ousmaneo: I agree with that. will change that.

graylog2-web-interface/src/components/event-notifications/event-notification-types/HttpNotificationForm.jsx
ousmaneo: fixed

ousmaneo:
Hi @thll, I made some updates from our discussions and also fixed the comments.

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,122 @@
...
+  handleFormEvent: PropTypes.func.isRequired,
+  validationState: PropTypes.string,
+  validationMessage: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.string,
+  ]),
+  url: PropTypes.string,
+  labelClassName: PropTypes.string,
+  wrapperClassName: PropTypes.string,
+  formType: PropTypes.string,
```
edmundoa: I think we should restrict the actual values that are accepted for this prop. I also think that the name could be more descriptive, something like `urlType` is clearer to me than `formType`.

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,122 @@
...
+  const handleChange = (event: SyntheticInputEvent<EventTarget>) => {
+    handleFormEvent(event);
+  };
+
+  useEffect(() => {
+    checkIsWhitelisted();
+  }, [url, validationState]);
+
+
+  const addButton = isError() && !isWhitelisted ? <URLWhitelistFormModal newUrlEntry={suggestedUrl} onUpdate={onUpdate} formType={formType} /> : '';
```
edmundoa: The inline button there looks odd to me, maybe @kyleknighted has an idea to improve that a bit? Also I think we should treat differently the case when a URL is not in the whitelist, and when the URL is not valid.
Screenshot 2020-02-19 at 11 01 14

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,122 @@
...
+  label: string,
+  handleFormEvent: (event: SyntheticInputEvent<EventTarget>) => void,
+  validationMessage: string,
+  validationState: string,
+  url: string,
+  labelClassName: string,
+  wrapperClassName: string,
+  formType: string
+}
+const URLWhitelistInput = ({ label, handleFormEvent, validationMessage, validationState, url, labelClassName, wrapperClassName, formType }: Props) => {
```
edmundoa: Since the `URLWhitelistInput` will be used in many different forms across the product, I think it should live somewhere else. Since some shared inputs are already in `components/common`, I would suggest moving it there.

graylog2-web-interface/src/components/configurations/UrlWhiteListForm.jsx
```diff
@@ -132,7 +134,7 @@ const UrlWhiteListForm = ({ urls, onUpdate, disabled }: Props) => {
           <td>
             <Input type="text"
                    id={`title-input${idx}`}
-                   ref={(elem) => { inputs[`title${idx}`] = elem; }}
+                   ref={(elem) => { inputs[`title${idx}`] = elem; return true; }}
```
edmundoa: I'm not sure why we need to return `true` in here, could you please explain it? :)

graylog2-web-interface/src/components/configurations/UrlWhiteListForm.jsx
```diff
@@ -143,7 +145,7 @@ const UrlWhiteListForm = ({ urls, onUpdate, disabled }: Props) => {
           <td>
             <Input type="text"
                    id={`value-input${idx}`}
-                   ref={(elem) => { inputs[`value${idx}`] = elem; }}
+                   ref={(elem) => { inputs[`value${idx}`] = elem; return true; }}
```
edmundoa: Same as above.

graylog2-web-interface/src/components/configurations/URLWhitelistFormModal.jsx
```diff
@@ -0,0 +1,158 @@
...
+type Props = {
+  newUrlEntry: string,
+  onUpdate: () => void,
+  configuration: {},
+  currentUser: {permissions: Array<string>},
+  formType: string,
+};
+
+
+class URLWhitelistFormModal extends React.Component<Props, State> {
```
edmundoa: Most of the code in this component should be very similar to parts of what we have in `UrlWhiteListConfig.jsx`, I think we should try and make a single component doing this, so we can reuse it better whenever is needed.

graylog2-web-interface/src/components/lookup-tables/adapters/DSVHTTPAdapterFieldSet.jsx
```diff
@@ -107,4 +103,13 @@ const DSVHTTPAdapterFieldSet = ({ handleFormEvent, validationState, validationMe
   );
 };
 
+DSVHTTPAdapterFieldSet.propTypes = {
+  config: PropTypes.object.isRequired,
+  // eslint-disable-next-line react/no-unused-prop-types
+  updateConfig: PropTypes.func.isRequired,
```
edmundoa: This prop is marked as unused, do we really use it in the component or can we delete it?

graylog2-web-interface/src/components/lookup-tables/adapters/HTTPJSONPathAdapterFieldSet.jsx
```diff
@@ -17,46 +18,43 @@ class HTTPJSONPathAdapterFieldSet extends React.Component {
...
-               label="Lookup URL"
-               autoFocus
-               required
-               onChange={this.props.handleFormEvent}
-               help={this.props.validationMessage('url', 'The URL for the lookup. (this is a template - see documentation)')}
-               bsStyle={this.props.validationState('url')}
-               value={config.url}
-               labelClassName="col-sm-3"
-               wrapperClassName="col-sm-9" />
+        <URLWhitelistInput label="Lookup URL"
```
edmundoa: When creating a new Data Adapter the whitelist check is returning errors before the field is set, but I think that should not happen.
Kapture 2020-02-19 at 11 26 51

ousmaneo: This is caused by the normal data adapters form validation not the whitelist check.
Data adapter validation happen on OnChange, on any of the field that need to be sent to the backend. Sending the data without any value in the other fields result in some params missing in the data that are sent causing this.

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  };
+
+
+  const onUpdate = () => {
+    UIUtils.triggerInput(ref.current);
+    checkIsWhitelisted();
+  };
+
+  const handleChange = (event: SyntheticInputEvent<EventTarget>) => {
```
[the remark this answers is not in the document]
ousmaneo: Agree, this is some code from the previous version where I was doing another thing in the handleChange. Fixed!

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,122 @@
...
+  const [suggestedUrl, setSuggestedUrl] = useState(url);
+  const isError = () => currentValidationState === 'error';
+  const ref = useRef();
+
+  const handleCheckIsWhitelisted = () => {
+    const promise = ToolsStore.urlWhiteListCheck(url);
+    promise.then((result) => {
+      if (!result.is_whitelisted && validationState === null) {
+        setCurrentValidationState('error');
+        setOwnValidationMessage(`URL ${suggestedUrl} is not whitelisted or not valid URL.`);
```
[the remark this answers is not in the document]
ousmaneo: Fixed

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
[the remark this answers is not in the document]
ousmaneo: Changed !

graylog2-web-interface/src/components/lookup-tables/adapters/DSVHTTPAdapterFieldSet.jsx
ousmaneo: Fixed

graylog2-web-interface/src/components/lookup-tables/adapters/HTTPJSONPathAdapterFieldSet.jsx
ousmaneo: I will create an issue for this

graylog2-web-interface/src/util/URLUtils.js
```diff
@@ -41,17 +41,29 @@ const URLUtils = {
...
   areCredentialsInURLSupported() {
     const browser = this.parser.getBrowser();
     return browser.name !== 'IE' && browser.name !== 'Edge';
   },
-};
+  isValidURL(str) {
+    let isValid = true;
+    try {
+      // eslint-disable-next-line no-unused-vars
+      const test = new URL(str);
```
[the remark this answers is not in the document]
ousmaneo: ok

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
kyleknighted: @edmundoa What is the recommended way for validating a URL?

In my server host file I can add `127.0.0.1 foobar` and then this is a valid URL, even though there is no `dot com`.

`wikipediа.org` and `wikipedia.org` are both different thanks to Cyrillic and Latin homographs. So we can't really strip away everything that isn't a Latin letter.

I'm not sure if we have any specific way to validate what a "real" URL is.

As for the button placement, I would probably need more insight on the entire flow & design of this feature to provide much feedback there.

graylog2-web-interface/src/components/configurations/UrlWhiteListForm.jsx
kyleknighted: And is it possible to use the `createRef` that I had mentioned in a previous comment?

ousmaneo: Hi @kyleknighted , these are dynamically created refs for the url whitelist from that's why I use callback ref.

graylog2-web-interface/src/components/configurations/URLWhitelistInput.jsx
edmundoa: @kyleknighted this is not about ensuring a given URL is "real" or not, for what I have seen in the code, we consider a URL valid if `new URL()` does not throw an exception. So, for instance, a missing or wrong schema in the URL would do that.

The biggest concern I have is that users will see the message and try to put the URL in the whitelist. Then the form will complain that is not a valid URL, so it is kind of misleading when we could do that check in first place.

Regarding the button, I think the alignment and spacing doesn't look right, and also the colours look kind of odd to me. Other than that I think it's fine.

ousmaneo: Hi @edmundoa I added a fix to separate Url and Whitelist validation.

graylog2-web-interface/src/components/common/URLWhitelistFormModal.jsx
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  _update = (config, isValid) => {
+    const updatedState = { config, isValid };
+    this.setState(updatedState);
+  }
+
+
+  _resetConfig = () => {
```
edmundoa: I guess we can remove this empty method.

graylog2-web-interface/src/components/common/URLWhitelistFormModal.jsx
```diff
@@ -0,0 +1,158 @@
...
+      newUrlEntry: '',
+      onUpdate: () => {},
+      configuration: {},
+      urlType: '',
+    }
+
+    constructor(props) {
+      super(props);
+      this.state = {
+        config: { entries: [], disabled: false },
```
edmundoa: For what I could understand about this component, `URLWhitelistFormModal` is holding the current state of the whitelist entries, but those are generated from props on render time and they only come into the state when `URLWhitelistForm`'s `onUpdate` is called.

That works, but I think it goes against _the React way_ of using a one-way data flow in your component hierarchy. I think initializing and keeping an up-to-date `config` state here would help us not relying on a child component updating to get the right state in this component.

graylog2-web-interface/src/components/common/URLWhitelistInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+  label: string,
+  onChange: (event: SyntheticInputEvent<EventTarget>) => void,
+  validationMessage: string,
+  validationState: string,
+  url: string,
+  labelClassName: string,
+  wrapperClassName: string,
+  urlType: string
+}
+const URLWhitelistInput = ({ label, onChange, validationMessage, validationState, url, labelClassName, wrapperClassName, urlType }: Props) => {
```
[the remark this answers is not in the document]
ousmaneo: Done !

graylog2-web-interface/src/components/common/URLWhitelistFormModal.jsx
```diff
@@ -59,7 +59,22 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
...
-  _getConfig = (configType: string) => {
+    componentDidUpdate() {
+      const { config: { entries } } = this.state;
+      const urlwhitelistConfig = this._getConfig(URL_WHITELIST_CONFIG);
+      if (urlwhitelistConfig && entries.length === 0) {
+        this._setDefaultWhiteListState(urlwhitelistConfig);
+      }
+    }
+
+  _setDefaultWhiteListState =(urlwhitelistConfig) => {
```
edmundoa: I still think this function (or a strip down version of it) should be called from the constructor, so the initial state comes from the given props and it's not empty.

edmundoa:
I also saw a missing margin between the first "Add URL" button and the table in  [`UrlWhiteListForm`](https://github.com/Graylog2/graylog2-server/blob/afa81c140e4302a47c37c605be4509bc0c019e4d/graylog2-web-interface/src/components/configurations/UrlWhiteListForm.jsx#L191):

Screenshot 2020-03-04 at 12 14 24

That is not a problem in the configuration page, because it is using the [`top-margin` CSS class](https://github.com/Graylog2/graylog2-server/blob/cf8b5afae886841aab30c66a6680864b07ebae99/graylog2-web-interface/src/components/configurations/ConfigurationStyles.css#L14) that is applied there. However, we should probably change that with a styled component to ensure we have the same styling when using the form in other pages.

graylog2-web-interface/src/components/common/URLWhiteListInput.jsx
```diff
@@ -0,0 +1,124 @@
...
+        if (typeof suggestion === 'object') {
+          suggestion.then((result) => {
+            setSuggestedUrl(result.regex);
+          });
+        } else {
+          setSuggestedUrl(url);
+        }
+      }
+    };
+    const timer = setTimeout(() => checkSuggestion(), 500);
```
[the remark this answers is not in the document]
ousmaneo: I changed the timeout to 250 milliseconds, But as for the lodash debounce it was my first choice but it does not seems to work in React functional component, wrapping it with useCallback hook also does not help too.

graylog2-web-interface/src/components/common/URLWhiteListFormModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    componentDidMount() {
+      const { currentUser: { permissions } } = this.props;
+      if (PermissionsMixin.isPermitted(permissions, ['urlwhitelist:read'])) {
+        ConfigurationsActions.listWhiteListConfig(URL_WHITELIST_CONFIG);
+      }
+    }
+
+
```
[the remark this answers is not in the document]
ousmaneo: +1

graylog2-web-interface/src/components/configurations/UrlWhiteListForm.jsx
```diff
@@ -4,7 +4,8 @@ import PropTypes from 'prop-types';
 import uuid from 'uuid/v4';
 import { cloneDeep, debounce } from 'lodash';
 import Input from 'components/bootstrap/Input';
-import { Select, Icon } from 'components/common';
+import Select from 'components/common/Select';
+import Icon from 'components/common/Icon';
```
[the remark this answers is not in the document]
ousmaneo: That's what I was using but it resulted into eslint [no circular dependency](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md) and I fixed it by having these self references explicit.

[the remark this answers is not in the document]
edmundoa: Alright, let's add a comment before them to clarify that.

graylog2-web-interface/src/components/common/URLWhiteListFormModal.jsx
```diff
@@ -0,0 +1,172 @@
...
+import PropTypes from 'prop-types';
+import uuid from 'uuid/v4';
+import connect from 'stores/connect';
+import { Button } from 'components/graylog';
+import BootstrapModalForm from 'components/bootstrap/BootstrapModalForm';
+import UrlWhiteListForm from 'components/configurations/UrlWhiteListForm';
+import CombinedProvider from 'injection/CombinedProvider';
+import PermissionsMixin from 'util/PermissionsMixin';
+import type { WhiteListConfig } from 'stores/configurations/ConfigurationsStore';
+import IfPermitted from 'components/common/IfPermitted';
```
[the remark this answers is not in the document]
edmundoa: As in `UrlWhiteListForm`, let's add a comment about the import style.

graylog2-web-interface/src/components/common/URLWhitelistFormModal.jsx
edmundoa: As we discussed, the data is fetched in `componentDidMount()` and setting a default state until then is fine.

graylog2-web-interface/src/components/common/URLWhiteListInput.jsx
```diff
@@ -0,0 +1,125 @@
...
+  const isWhitelistError = () => currentValidationState === 'error' && isValidURL(url);
+  const ref = useRef();
+
+  const checkIsWhitelisted = () => {
+    if (url) {
+      const promise = ToolsStore.urlWhiteListCheck(url);
+      promise.then((result) => {
+        if (!result.is_whitelisted && validationState === null) {
+          setCurrentValidationState('error');
+          const message = isValidURL(url) ? `URL ${url} is not whitelisted` : `URL ${url} is not valid URL.`;
```
edmundoa: Nitpick:
```suggestion
          const message = isValidURL(url) ? `URL ${url} is not whitelisted` : `URL ${url} is not a valid URL.`;
```

PR #7302, opened 2020-01-27
Fix default selection for stream output assignment

bernd:
When loading the stream output page, the dropdown for "Assign existing
Output" shows the first existing output as selected. When pressing the
button, an error had been generated because there wasn't an actual
selection.

This change is fixing the AssignOutputDropdown to show the placeholder
by default and disable the "Assign existing Output" button until a
selection has been made.

### DIFF
--- a/graylog2-web-interface/src/components/outputs/AssignOutputDropdown.jsx
+++ b/graylog2-web-interface/src/components/outputs/AssignOutputDropdown.jsx
@@ -4,6 +4,8 @@ import React from 'react';
 import { Button } from 'components/graylog';
 
 class AssignOutputDropdown extends React.Component {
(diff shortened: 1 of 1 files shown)


bernd:
Example showing the issue:

Peek 2020-01-27 18-27

bernd:
> This is fixing the issue with selection but also leaving some stale code behind.
> The reason why the `selectedOutput` wasn't set in the state is because `PLACEHOLDER` was initialized only after it was used in the class. Moving the declaration of `PLACEHOLDER` to the top of the class should fix the issue.

Heh, good catch. I missed that. :smile: I updated the PR to only change the initialization of the `PLACEHOLDER` variable.

PR #7305, opened 2020-01-27
Dashboard widget action dropdown position

kyleknighted:
## Description

Dynamically place the dropdown on the opposite side if not enough room

## Motivation and Context

Dropdown on right side can disappear off of edge
Peek 2020-01-27 16-54

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/OverlayDropdown.jsx
+++ b/graylog2-web-interface/src/views/components/OverlayDropdown.jsx
@@ -1,77 +1,70 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { debounce } from 'lodash';
 
-import { Overlay } from 'react-overlays';
+import { Overlay, Transition } from 'react-overlays';
 
 import styles from './OverlayDropdown.css';
+
+const StyledList = styled.ul(({ show, minWidth }) => `
+  padding-left: 5px;
+  padding-right: 5px;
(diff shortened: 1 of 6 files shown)


linuspahl:
I like the idea! One thing I noticed, when the value of e.g. the message field is very long and you click at the beginning of the message, the dropdown will still be displayed at the window border and is not fully visible (due to `event.clientX`). What worked for me is using the caret as the reference and `toggleTarget.current.getBoundingClientRect().x` instead of `event.clientX`

kyleknighted:
@linuspahl I looked into it and `toggleTarget.current.getBoundingClientRect().x` is the text in the table cell and not the width of the dropdown.

Going to work on getting a good ref to the dropdown and see if I can get the math to work better because I am seeing similar results with longer fields as I test more

linuspahl:
@kyleknighted My idea was to set the reference for the caret, instead of the value text, so `toggleTarget.current.getBoundingClientRect().x` would return the distance of the caret to the left browser window border. But your new implementation is way more flexible!

I tried out different placements and different use cases. The only case where the dropdown is not being displayed perfectly, is when my browser window has half of the screen size (the window  has an effective width of 1280px) and I am working with very long field names in the field list:
image

Do you think this is an edge case or worth a more complex calculation / placement adjustment?

kyleknighted:
@linuspahl I think it's considerably better. We could introduce a max-width to the element and just have insanely long titles ellipsis off or wrap or something. I'm not sure if the development time is worthwhile at this point though.

PR #7306, opened 2020-01-28
Change color chart from Map to List of Value Class 

### LINKED ISSUE #7235: Color assignment in graphs doesn't work for values with dots
## Expected Behavior

It should be possible to assign a color in a graph for values that contain a dot (`.`).

## Current Behavior

When trying to assign a color for a value in a graph that contains a dot (e.g. `org.graylog.plugins.map.geoip.GeoIpResolverEngine`), saving the search fails:

Peek 2020-01-21 15-32

This is the server response:

```json
{
  "type": "ApiError",
  "message": "Invalid BSON field name org.graylog2.migrations.MigrationHelpers"
}
```
## Possible Solution

I haven't checked the code, but I guess we are storing the value that should use a specific color in as a key in MongoDB. (e.g. `{"the.dotted.value": "#0000ff"}`) MongoDB [doesn't support dots in key names](https://docs.mongodb.com/manual/core/document/#field-names) (or more specific: the database itself supports it since version 3.6, but the drivers don't support it and it's generally not recommended)

That means we cannot really use user generated values as keys in MongoDB. (there are other restrictions, too) As an alternative, we could use a list of objects to do the mapping.

1. Create aggregation on values that contain dots (`.`)
2. Change widget to a graph 
3. Save aggregation
4. Click on the color to set a new one
5. Save the search
6. Check error

## Your Environment

* Graylog Version: 3.2.0-beta.4-SNAPSHOT (rev ee3c87c)

kmerz:
## Description, Motivation and Context
Prior to this change, the data structure to store
WidgetFormattingSettings did use the "user input" for storing a map.
But mongodb does not allow to store keys with dot in the map.

This change will change the data structure from a map to a list
of objects. That way mongodb does not need to store the keys.

Also add migration to fix existing color charts in mongodbs.

Fixes #7235 

## How Has This Been Tested?
1. Add dashboard with color chart containing a non IP keyname with dots and see that it
does not get saved. 
1. Remove dots from name
1. Run migration
1. Check that chart color still exists and add the dots back to the name which should now successfully be saved.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/widgets/aggregation/ChartColorSetting.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/widgets/aggregation/ChartColorSetting.java
@@ -1,0 +1,35 @@
+package org.graylog.plugins.views.search.views.widgets.aggregation;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+@JsonDeserialize(builder = ChartColorSetting.Builder.class)
+public abstract class ChartColorSetting {
+   private static final String FIELD_NAME = "field_name";
+    private static final String FIELD_CHART_COLOR = "chart_color";
+
+    @JsonProperty(FIELD_NAME)
+    public abstract String fieldName();
+
+    @JsonProperty(FIELD_CHART_COLOR)
+    public abstract ChartColor chartColor();
+
+    @AutoValue.Builder
+    public static abstract class Builder {
+        @JsonProperty(FIELD_NAME)
+        public abstract Builder fieldName(String widgetId);
+
+        @JsonProperty(FIELD_CHART_COLOR)
+        public abstract Builder chartColor(ChartColor chartColor);
+
+        public abstract ChartColorSetting build();
+
+        @JsonCreator
+        static Builder builder() {
+            return new AutoValue_ChartColorSetting.Builder();
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/views/widgets/aggregation/WidgetFormattingSettings.java
```diff
@@ -22,27 +22,27 @@
...
+import java.util.List;
 
 @AutoValue
 @JsonDeserialize(builder = WidgetFormattingSettings.Builder.class)
 public abstract class WidgetFormattingSettings {
     private static final String FIELD_CHART_COLORS = "chart_colors";
 
     @JsonProperty(FIELD_CHART_COLORS)
-    public abstract Map<String, ChartColor> chartColors();
+    public abstract List<ChartColorSetting> chartColorSettings();
```
dennisoelkers: IMO the previous naming (`chartColors`) was slightly better.

graylog2-server/src/main/java/org/graylog/plugins/views/search/views/widgets/aggregation/ChartColorSetting.java
```diff
@@ -0,0 +1,35 @@
+package org.graylog.plugins.views.search.views.widgets.aggregation;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+@JsonDeserialize(builder = ChartColorSetting.Builder.class)
+public abstract class ChartColorSetting {
```
dennisoelkers: Maybe `ChartColorMapping` captures the purpose of this class better?

PR #7320, opened 2020-01-29
Increasing jest timeout for integration test.

dennisoelkers:
## Motivation and Context

The `CreateNewDashboard` integration test has proven to be unreliable
under certain conditions (most probably multiple `jest` runs happening
on CI hardware in parallel), leading to false negatives.

This change is increasing the jest timeout for this specific test suite,
hoping to reduce the number of false negatives when run under load.

### DIFF
--- a/graylog2-web-interface/src/views/spec/CreateNewDashboard.test.jsx
+++ b/graylog2-web-interface/src/views/spec/CreateNewDashboard.test.jsx
@@ -43,6 +43,10 @@ describe('Create a new dashboard', () => {
     PluginStore.register(new PluginManifest({}, viewsBindings));
   });
 
+  beforeEach(() => {
+    jest.setTimeout(30000);
+  });
+
   afterEach(cleanup);
 
   it('using Dashboards Page', async () => {


graylog2-web-interface/src/views/spec/CreateNewDashboard.test.jsx
```diff
@@ -43,6 +43,10 @@ describe('Create a new dashboard', () => {
     PluginStore.register(new PluginManifest({}, viewsBindings));
   });
 
+  beforeEach(() => {
+    jest.setTimeout(30000);
+  });
```
linuspahl: Is this fixing the problem locally for you, when you run `yarn test` multiple times? I am still receiving the same error.

`jest.setTimeout` will also work for the hole file without anything like `beforeEach()`. Do you use `beforeEach` to keep it in the scope of the `describe()`?

dennisoelkers: It is fixing the problem up to a certain concurrency level. Naturally, we cannot fix it for any amount of parallelism, as the runtime will just increase.

I wrapped it in `beforeEach` to limit it to the smallest scope available.

PR #7325, opened 2020-01-29
Fix blocking issue in DNS lookup data adapter

### LINKED ISSUE #5782: Changing DNS lookup adapter configuration completely halts message processing
When changing configuration of a DNS lookup adapter, the adapter seems to stop but never restart and leave the system halted (no more messages are being processed) until a service restart.

I can reproduce this behavior and tried it with a change to the timeout value as well as the DNS server address.

## Your Environment

* Graylog Version: Graylog 3.0.0+db6cf59

bernd:
We cannot use "Future#sync()" to wait for the result of the DNS
resolver. This can block forever (until the server gets restarted)
when the DnsNameResolver and the netty event loop gets shut down while a
DNS request is running.

This happened on busy systems that do a lot of DNS lookups and the DNS
lookup data adapter gets restarted while DNS requests are running. (e.g.
by updating the data adapter's configuration)

It can get even worse when multiple processing threads try to lookup the
same hostname or IP address at the same time, using the same lookup
table. One thread is waiting for a response but never returns.
All other processing threads are blocking on the cache lock, waiting on
the other thread to return.

The solution is to not use "Future#sync()" and use "Future#get()" with a
timeout so the thread doesn't wait for the future completion forever.

By default the timeout for waiting on the future is based on the query
timeout plus 100 ms to avoid canceling the future before a potential
DNS query timeout kicks in.

This change also ensures that the DnsNameResolver will be closed before
shutting down the event loop.

Fixes #5782

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/DnsLookupDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/DnsLookupDataAdapter.java
@@ -192,7 +192,7 @@ public class DnsLookupDataAdapter extends LookupDataAdapter {
         } catch (UnknownHostException e) {
             return LookupResult.empty(); // UnknownHostException is a valid case when the DNS record does not exist. Do not log an error.
         } catch (Exception e) {
-            LOG.error("Could not resolve [{}] records for hostname [{}]. Cause [{}]", A_RECORD_LABEL, key, e.toString());
+            LOG.error("Could not resolve [{}] records for hostname [{}]. Cause [{}]", A_RECORD_LABEL, key, ExceptionUtils.getRootCauseOrMessage(e));
             errorCounter.inc();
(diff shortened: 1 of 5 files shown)


bernd:
@mpfz0r @danotorrey This should most probably be backported to the 3.2 branch. (maybe even 3.1)

bernd:
@mpfz0r @danotorrey To reproduce the issue (without this fix), you can do the following:

1. Setup a DNS lookup table adapter that is using a DNS server IP address which doesn't work (to trigger the request timeout for every message)
1. Configure a pipeline rule to do a DNS lookup for every message
1. Send a lot of messages to make sure all processing threads are busy
1. Reconfigure the DNS lookup data dapter (e.g. change the DNS server IP address to a working one)
1. Check the JVM thread dump of the server to see all processbuffer threads being blocked

PR #7336, opened 2020-01-31
Handling `null` attributes when migrating quickvalues widgets.

### LINKED ISSUE #7334: RE- Dashboards page will not load after upgrade to 3.2
#7287  

@kmerz , the problem is still existing for me :(
I tried to upgrade from 3.1.4 to 3.2 RC1 but I had only the default source Dashboard and a view which was migrated to Dashboard.

[http://pastebin.fr/60590](url)

dennisoelkers:
## Motivation and Context

Under some circumstances, the `show_data_table` & `show_pie_chart`
attributes of legacy quick values widgets can be `null`, leading to NPEs
when migrating or installing legacy dashboards.

This PR is now handling the absence of these attributes. If they are
missing, they are handled the following way:

  - `show_data_table` is `true` if `show_pie_chart` is missing or
`false` and `false` otherwise, ensuring that at least one of both is created
  - `show_pie_charts` is `false` if missing

This logic is identical to the rendering logic of the legacy dashboard
code.

Fixes #7334.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20191125144500_MigrateDashboardsToViewsSupport/dashboardwidgets/QuickValuesConfig.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20191125144500_MigrateDashboardsToViewsSupport/dashboardwidgets/QuickValuesConfig.java
@@ -47,6 +47,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
+
 @AutoValue
 @JsonAutoDetect
 @JsonIgnoreProperties(ignoreUnknown = true)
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20191125144500_MigrateDashboardsToViewsSupport/dashboardwidgets/QuickValuesConfig.java
```diff
@@ -186,8 +188,10 @@ static QuickValuesConfig create(
                 query,
                 Optional.ofNullable(streamId),
                 field,
-                showDataTable,
-                showPieChart,
+                (showDataTable == null || !showDataTable) && (showPieChart == null || !showPieChart)
```
alex-konn: I think a few simple unit tests would be in order here to test all possible combinations.

PR #7342, opened 2020-01-31
Add more tests for search page components

linuspahl:
This PR adds more tests for the `NewSearchPage`, `NewDashboardPage`, `ShowDashboardInBigDisplayMode`, `StreamSearchPage` and `SearchResult`

It's mainly in draft mode to clarify some remaining questions.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/dashboard/BigDisplayModeHeader.jsx
```diff
@@ -36,8 +36,14 @@ const BigDisplayModeHeader = ({ view: { activeQuery, view } = {} }: Props) => {
...
-  }).isRequired,
+    view: PropTypes.shape({
+      title: PropTypes.string.isRequired,
+    }),
+  }),
+};
+
+BigDisplayModeHeader.defaultProps = {
+  view: undefined,
 };
```
linuspahl: This throwed a PropType error. I removed the `isRequired` and added `undefined` as a default props, because it's the state, before the store got connected.

graylog2-web-interface/src/views/pages/NewDashboardPage.jsx
```diff
@@ -31,7 +32,7 @@ const NewDashboardPage = ({ route, location, loadingViewHooks, executingViewHook
     const { state = {} } = location;
     const { view: searchView } = state;
     if (searchView && searchView.search) {
-      const query = location;
+      const { query } = location;
```
linuspahl: When I go it right `processHooks` expects the query object instead of the full location.

graylog2-web-interface/src/views/pages/ShowDashboardInBigDisplayMode.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+    await wait(() => expect(RefreshActions.enable).toHaveBeenCalledTimes(1));
+  });
+
+  // it('should display view title', async () => {
+  //   const { getByText } = render(<ShowDashboardInBigDisplayMode route={{}}
+  //                                                               params={{ viewId: mockView.id }}
+  //                                                               location={mockLocation} />);
+
+  //   await waitForElement(() => getByText('view title'));
+  // });
```
linuspahl: Question: At first I was not aware of the `connect` mock we are using for several other tests. The above tests are working without the `connect` mock, but at some point I needed to add the `cleanup` method in `afterEach`.
Implementing the `cleanup` method without the `conntect` mock throws the following error: `TypeError: desub is not a function`

But with the `connect` mock, the view returned from `ViewStore.getInitialState` is no longer available in e.g. the `BigDisplayModeHeader`

My question is, how to deal with this scenario?

graylog2-web-interface/src/views/pages/NewSearchPage.test.jsx
```diff
@@ -0,0 +1,194 @@
...
+  },
+}));
+jest.mock('views/logic/views/ViewLoader', () => {
+  const originalModule = jest.requireActual('views/logic/views/ViewLoader');
+  return {
+    __esModule: true,
+    ...originalModule,
+    processHooks: jest.fn((promise, loadHooks, executeHooks, query, onSuccess) => Promise.resolve().then(onSuccess)),
+  };
+});
```
linuspahl: Question: So far we are not using `requireActual`. it looked like the best solution for me, in the context of the following case:
This is the related import `import ViewLoader, { processHooks } from 'views/logic/views/ViewLoader';`
We need to mock `processHooks`, to test if it got called correctly. But I don't want to mock the ViewLoader, because I need to trigger e.g. the onError function we are passing to the `ViewLoader`. This is the main reason for the above implementation. 

Having a look at e.g. the the `ShowViewPage`, I saw a different usage of the `ViewLoader` regarding the import. The `ViewLoader` and `processHooks` are getting imported the same way, but we are not using the `ViewLoader` directly. We are defining it as a default prop. 

My question is, should we unify the usage of the `ViewLoader` for `NewDashoardPage`, `NewSearchPage`, `ShowViewPage` and `StreamSearchPage`?
The Implementation with the default prop would simplify the described test case.

graylog2-web-interface/src/views/pages/NewSearchPage.test.jsx
```diff
@@ -0,0 +1,194 @@
...
+
+    //   const { getByText } = render(<NewSearchPage route={{}}
+    //                                               router={mockRouter}
+    //                                               location={{}} />);
+    //   const viewLoadButton = await waitForElement(() => getByText('Load view'));
+    //   fireEvent.click(viewLoadButton);
+
+    //   await waitForElement(() => getByText('The hook component'));
+    // });
+  });
```
linuspahl: Question: Having a look at the implementation of `processHooks`, I realized we should unify the usage, especially in the context of the error case. Sometimes we are catching the error and creating a `UserNotification`:
```
 return processHooks(
      ...
 ).catch(
     error => UserNotification.error(`Executing search failed with error: ${error}`, 'Could not execute search'),
);
```
sometimes we are using the `processHooks` `onError` argument, to set the `hookComponent` state 
```
return viewLoader(
      ...
      (e) => {
        if (e instanceof Error) {
          throw e;
        }
        this.setState({ hookComponent: e });
      },
```
and sometimes we are doing the same, but without checking the error instance.

My main question is related to the `hookComponent`. It looks like we are defining a `hookComponent` when `processHooks` errors. When the `hookComponent` is defined we'll display it instead of the `ExtendedSearchPage`. I want to test tis case, but looking at the code I don't understand when the `hookComponent` (e.g. in the last code snippet) gets defined. because when it is an error, `setState` will not get called. When it is not an error, the `onError` function passed to `processHooks` will not getCalled.

graylog2-web-interface/src/views/pages/ShowDashboardInBigDisplayMode.jsx
```diff
@@ -64,4 +62,4 @@ const ShowDashboardInBigDisplayMode = ({ location, params, route, view: { view,
   );
 };
 
-export default withRouter(connect(ShowDashboardInBigDisplayMode, { view: ViewStore, viewMetadata: ViewMetadataStore, query: CurrentQueryStore }));
```
linuspahl: I removed  the `ViewMetadataStore` and `CurrentQueryStore ` usage, because it looks like we are not using `viewMetadata` and `query`.

PR #7350, opened 2020-02-03
Suppress system notification for whitelist error from test alert

thll:
## Description

When a HTTP event notification fails due to an URL Whitelist error, we now check if the failed event notification was a test-notification. If that's the case, the global system notification showing the whitelist error will be suppressed.

## Motivation and Context

When an HTTP event notification is tested and the whitelist entered was not whitelisted, a sticky system notification would show up, indicating the whitelist error. This system notification was however only meant to be triggered during the execution of "real" alerts.

## How Has This Been Tested?

Quick manual test via the frontend.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/notifications/NotificationTestData.java
+++ b/graylog2-server/src/main/java/org/graylog/events/notifications/NotificationTestData.java
@@ -30,7 +30,9 @@ import org.graylog2.plugin.Tools;
 import org.graylog2.plugin.rest.ValidationResult;
 import org.graylog2.plugin.streams.Stream;
 
-class NotificationTestData {
+public class NotificationTestData {
+    public static final String TEST_NOTIFICATION_ID = "1234";
+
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog/events/notifications/NotificationTestData.java
```diff
@@ -30,7 +30,9 @@
 import org.graylog2.plugin.rest.ValidationResult;
 import org.graylog2.plugin.streams.Stream;
 
-class NotificationTestData {
+public class NotificationTestData {
+    public static final String TEST_NOTIFICATION_ID = "1234";
```
mpfz0r: maybe we could make that id a little more explicit. like `testing-testing-123` :)

thll: As this is not a valid ObjectId anyway I didn't think it's necessary to make it more obvious. But sure, I can do this ;)

PR #7378, opened 2020-02-04
Use upsert to create admin role

mpfz0r:
This avoids running into conflicts when multiple nodes
are bootstrapped simultaneously.


mariussturm:
Looks good to me! Couldn't trigger a duplicate key error with this in a clustered setup any more.

mariussturm:
Any news on this? Can we get this merged for the next release?

mpfz0r:
I think this is fine as it is :-)

PR #7401, opened 2020-02-05
Add API to generate a whitelist regex for a template URL

thll:
## Description

Exposes the regex generation logic previously only used in a migration. It can be used by the frontend to propose a regex for a URL template.

## Motivation and Context

This is required for #7169 to be implemented.
#7275 will make use of the new API.

## How Has This Been Tested?

Manual test by POSTing
```
{
  "url_template": "https://localhost:8080/get?key=${key}",
  "placeholder": "${key}"
}
```
to ``http://localhost:9000/api/system/urlwhitelist/generate_regex``
## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/system/UrlWhitelistResource.java
```diff
@@ -84,9 +89,27 @@ public Response put(@ApiParam(name = "whitelist", required = true) @NotNull fina
...
+    @POST
+    @Path("/generate_regex")
+    @Timed
+    @ApiOperation(value = "Generates a regex that can be used as a value for a whitelist entry.")
+    @NoAuditEvent("Utility function only.")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public WhitelistRegexGenerationResponse generateRegex(@ApiParam(name = "JSON body", required = true)
+    @Valid @NotNull final WhitelistRegexGenerationRequest generationRequest) {
+        final String regex;
+        if (generationRequest.placeholder() == null) {
```
bernd: Should we check here if the placeholder is empty as well? (e.g. by using `isNullOrEmpty(generationRequest.placeholder)`)

thll: Yep, makes sense. Fixed that.

PR #7444, opened 2020-02-12
fix the link to ipfix docs.

lingpri:
The details of the fix,is described in the issue below:
[354](https://github.com/Graylog2/graylog-plugin-integrations/issues/354)

## Description
The ipfix link is broken while navigating the links, system -> nodes -> details -> Available Inputs -> documentation.

## Motivation and Context
fix the broken documentation link for IPFIX.

## How Has This Been Tested?
-  tested this locally and is not working in my local, as the version is pointing to 3.3, if I manually
  change the version to 3.2 it works.
- needs to be backported to 3.2
 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/DocsHelper.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/DocsHelper.java
@@ -18,6 +18,7 @@ package org.graylog2.plugin;
 
 public enum DocsHelper {
     PAGE_SENDING_JSONPATH("sending_data.html#json-path-from-http-api-input"),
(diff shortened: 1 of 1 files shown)


lingpri:
@bernd - is there a possibility, that we can have stable and latest as version numbers ? 

read_the_docs

bernd:
> @bernd - is there a possibility, that we can have stable and latest as version numbers , just so the doc
> link is not broken.
> 
> read_the_docs

@lingpri With the `DocHelper` class, we always use the version number of the server to make sure the user gets the correct documentation.

PR #7446, opened 2020-02-12
Fix tests by using JSONAssert equals

### LINKED ISSUE #7445: Test can fail if serialization result is in a different order
Similar to the change you already merged for `ValueReferenceTest.java`:
test `shouldDeserializeSerializedContentPack` in `org.graylog2.contentpacks.model.ContentPackTest` uses `jackson` to serialize objects into JSON and assert the serialized results with JSON string. However, the order of serialized JSON is not guaranteed so the test may fail if the order is different.

## Expected Behavior

Tests should always pass if no changes made to source code

## Current Behavior

Tests may fail or pass with no changes made to the source code.

## Possible Solution

Instead using string comparison, more sophisticated comparison should be used here. For example, uses library like: https://github.com/skyscreamer/JSONassert

## Context

According to the specification, `An object is an unordered collection of zero or more name/value pairs, where a name is a string and a value is a string, number, boolean, null, object, or array.`

testfixer:
This PR fixes #7445
## Description

It uses `JSONAssert` to make the test is more stable.

## Motivation and Context

The test method uses `jackson` to serialize objects into JSON and assert the serialized results with JSON string. However, the order of serialized JSON is not guaranteed so test 

This PR proposes to use JSONAssert and modify the corresponding JSON test assertions so that the test is more stable.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/contentpacks/model/ContentPackTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/contentpacks/model/ContentPackTest.java
@@ -37,8 +37,10 @@ import org.graylog2.contentpacks.model.parameters.StringParameter;
 import org.graylog2.jackson.AutoValueSubtypeResolver;
 import org.graylog2.shared.bindings.providers.ObjectMapperProvider;
 import org.joda.time.DateTime;
+import org.json.JSONException;
 import org.junit.Before;
 import org.junit.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
 
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


dennisoelkers:
This looks good and might be an improvement to test stability, but due to using the non-strict mode of `JSONAssert#assertEquals`, it is less strict, not only considering reordered properties, but also considering that the actual value is allowed to be a superset of the expected value (see [here](http://jsonassert.skyscreamer.org/apidocs/index.html)).

dennisoelkers:
@testfixer: Is this an automated account? If not (or anyone is reading this), could you address the comment above?

PR #7482, opened 2020-02-17
Display full height of data tables in reports

### LINKED ISSUE #7349: Enterprise reports don't contain all table rows
When generating reports with the Graylog Enterprise plugin the data tables do not contain all table rows but seem to be limited to only a few rows.

## Expected Behavior
The report should contain all table rows.

## Current Behavior
Not all rows are included.

1. Create a dashboard with a huge table
2. Generate a report from this table

## Your Environment

* Graylog Version: Graylog 3.2.0+4edc81f on (Oracle Corporation 1.8.0_242 on Linux 5.3.0-28-generic)
* Graylog Enterprise Version: 3.2.0
* Operating System: Ubuntu 18.04 (Docker)
* Browser version: Safari (Version 13.0.5 (15608.5.11))

### Old reports (Graylog version 6.1.x):
Screenshot 2020-02-02 at 17 10 37

### New reports (Graylog version 6.2.0):
Screenshot 2020-02-02 at 17 10 20

linuspahl:
As described in https://github.com/Graylog2/graylog2-server/issues/7349 the data table is not fully visible in reports.

The reason is the visualization config in `bindings` returns `true` for the prop `needsControlledHeight`. With this PR we are checking the widget type to return `false` if the widget is a data table.

Please have a look at the related PR in the enterprise repo. After merging these changes, we need to create a backport PR for the next bugfix release.

Fixes #7349 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/bindings.jsx
+++ b/graylog2-web-interface/src/views/bindings.jsx
@@ -2,6 +2,7 @@
 import React from 'react';
 import Routes from 'routing/Routes';
 import * as Permissions from 'views/Permissions';
+import { get } from 'lodash';
 
 import { MessageListHandler } from 'views/logic/searchtypes';
 import { MessageList } from 'views/components/widgets';
@@ -131,7 +132,7 @@ export default {
       defaultWidth: 6,
       visualizationComponent: MessageList,
       editComponent: EditMessageList,
-      needsControlledHeight: false,
+      needsControlledHeight: () => false,
       searchResultTransformer: (data: Array<*>) => data[0],
       searchTypes: MessageConfigGenerator,
       titleGenerator: () => 'Untitled Message Table',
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/bindings.jsx
```diff
@@ -143,7 +144,13 @@ export default {
       defaultWidth: 4,
       visualizationComponent: AggregationBuilder,
       editComponent: AggregationControls,
-      needsControlledHeight: true,
+      needsControlledHeight: (widget: Widget) => {
+        const widgetVisualization = get(widget, 'config.visualization');
+        const flexibleHeightWidgets = [
```
linuspahl: Do you know a better place for this collection of widgets types which do not need a controlled height?

dennisoelkers: For now it's fine, because the plugin definitions contain the required knowledge. If we want to be able to delegate it, we should switch over to having a `needsControlledHeight` method on the `Widget` class. In that case we could delegate it further to the visualizations, in case of the `aggregation` widget.

Issue #5819, opened 2019-03-28
Update Kafka input plugin to modern kafka API


gimmic:
The current mechanisms in use by the kafka input plugin are deprecated in the current kafka release. The new system uses a different API for things like offset records.

https://kafka.apache.org/documentation/#upgrade
https://kafka.apache.org/protocol.html#The_Messages_OffsetCommit

I struggled with queue management for a while using the Kafka Input plugin until I downgraded our kafka instance to 1.1.1 (current is 2.2).

## Expected Behavior
Kafka Input plugin should register with the client offset API to manage the offsets, and also create a consumergroup so kafka 2.x is more aware of the state of the input consumers.

## Current Behavior
Graylog input uses an outdated mechanism to consume from kafka partitions

## Possible Solution
update kafka plugin to modern codebase

## Context
Currently in Kafka 2.2, the graylog kafka input "works" but the features it is using are in a degraded state in Kafka itself. queue monitoring and management is no longer possible in Kafka 2.x.

* Graylog Version: 3.0

Issue #5778, opened 2019-03-15
Direct connection to Apache Kafka


mommys-little-hacker:
## Expected Behavior

When creating a Kafka input the user is asked to input *broker* hosts and consumer group

## Current Behavior

When creating any Kafka input, GrayLog requires to input ZooKeeper host address. Essentially, ZooKeeper is not needed to discover and establish connection to Kafka brokers and generally it is acceptable to have it sealed from outside. For example, in my setup ZooKeeper runs inside Kubernetes and it's service is not published, while Kafka's is available from outside by an aliased security protocol. GrayLog approach does not work in this case.

## Possible Solution

Reworking Kafka connection procedure.

## Steps to Reproduce (for bugs)

Consider the following configuration snippet:
```
"advertised.listeners": "EXTERNAL://kafka-broker.example.com:31090"
"listener.security.protocol.map": "PLAINTEXT:PLAINTEXT,EXTERNAL:PLAINTEXT"
```

If I publish the ZooKeeper service and attempt to connect to Kafka, connection will fail since the underlying library does not know about this protocol:
```
2019-03-15 15:48:56,231 INFO    [ZookeeperConsumerConnector] - [graylog2_logs-graylog-0.logs-graylog.tiller.svc.cluster.local-1552664935968-9faee038], exception during rebalance  - {}
kafka.common.KafkaException: Failed to parse the broker info from zookeeper: {"listener_security_protocol_map":{"PLAINTEXT":"PLAINTEXT","EXTERNAL":"PLAINTEXT"},"endpoints":["PLAINTEXT://brok
er-kafka-0.broker-kafka-headless.tiller:9092","EXTERNAL://broker-0.kafka.example.com:31090"],"jmx_port":5555,"host":"broker-kafka-0.broker-kafka-headless.tiller","timestamp":"155265732
5894","port":9092,"version":4}
        at kafka.cluster.Broker$.createBroker(Broker.scala:85) ~[graylog.jar:?]
        at kafka.utils.ZkUtils$$anonfun$getCluster$1.apply(ZkUtils.scala:563) ~[graylog.jar:?]
        at kafka.utils.ZkUtils$$anonfun$getCluster$1.apply(ZkUtils.scala:561) ~[graylog.jar:?]
        at scala.collection.Iterator$class.foreach(Iterator.scala:893) ~[graylog.jar:?]
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1336) ~[graylog.jar:?]
[...]

PR #7504, opened 2020-02-19
Add support for high level Kafka consumer mode

mpfz0r:
Newer Kafka versions are supported by using the new high level consumer API.
Lucky for us, support for the new API is already included in our currently used Kafka
client library. This means we can avoid updating it, and still be
compatible with our on-disk journal.

Instead of creating a new Transport class, use a legacy switch to keep
supporting older Kafka versions.

Most of this was spearheaded by Muralidhar Basani in
 #4770
Thank you very much! :-)

To configure the new Kafka client, a list of one or more Kafka brokers needs to be provided.
The legacy one needs to connect to Zookeeper.

Introduce a custom properties field, so users can tweak advanced settings or setup TLS.
Refs https://github.com/Graylog2/graylog2-server/pull/5088
Refs https://github.com/Graylog2/graylog2-server/issues/3960

Fixes https://github.com/Graylog2/graylog2-server/issues/5819
Fixes https://github.com/Graylog2/graylog2-server/issues/5778
Fixes https://github.com/Graylog2/graylog2-server/issues/5001
Fixes https://github.com/Graylog2/graylog2-server/issues/4481
Fixes https://github.com/Graylog2/graylog2-server/issues/3730
Fixes https://github.com/Graylog2/graylog2-server/issues/2780


graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -173,17 +187,124 @@ public void setMessageAggregator(CodecAggregator ignored) {
...
+                } catch (Exception e) {
+                    LOG.warn("Could not create KafkaConsumer", e);
+                    throw e;
+                }
+
+                while (!stopped) {
+                    final ConsumerRecords<byte[], byte[]> consumerRecords = consumer.poll(1000);
+                    final Iterator<ConsumerRecord<byte[], byte[]>> consumerIterator = consumerRecords.iterator();
+                    try {
+                        // we have to use hasNext() here instead foreach, because next() marks the message as processed immediately
```
bernd: Is that comment still valid? I checked the code briefly and I cannot see where `next()` would mark the message as processed.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -173,17 +187,124 @@ public void setMessageAggregator(CodecAggregator ignored) {
...
+                    consumer = new KafkaConsumer<>(nprops);
+                    //noinspection ConstantConditions
+                    consumer.subscribe(Pattern.compile(configuration.getString(CK_TOPIC_FILTER)), new NoOpConsumerRebalanceListener());
+                } catch (Exception e) {
+                    LOG.warn("Could not create KafkaConsumer", e);
+                    throw e;
+                }
+
+                while (!stopped) {
+                    final ConsumerRecords<byte[], byte[]> consumerRecords = consumer.poll(1000);
```
bernd: The `consumer.poll()` method can throw a few exceptions which we never catch anywhere. This would basically abort the thread **without** closing any resources. (no `consumer.close()` call)

I think we should handle these errors and check what we can do to avoid having a dead input.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -173,17 +191,127 @@ public void setMessageAggregator(CodecAggregator ignored) {
...
+                            if (stopped) {
+                                break;
+                            }
+                            if (isThrottled()) {
+                                blockUntilUnthrottled();
+                            }
+
+                            // process the message, this will immediately mark the message as having been processed. this gets tricky
+                            // if we get an exception about processing it down below.
+                            // final MessageAndMetadata<byte[], byte[]> message = consumerIterator.next();
```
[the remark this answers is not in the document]
mpfz0r: the first one, i think yes. but the code is a leftover, deleting it

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
mpfz0r: I guess that depends. The default is `auto.commit.enable` = true. So I'd say it does.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
mpfz0r: good catch. I found a couple of issues while fixing this.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -173,17 +197,158 @@ public void setMessageAggregator(CodecAggregator ignored) {
...
+                    consumerIterator = tryPoll();
+                    if (! consumerIterator.isPresent()) {
+                        LOG.error("Caught recoverable exception. Retrying");
+                        Thread.sleep(2000);
+                        continue;
+                    }
+                } catch (KafkaException | InterruptedException e) {
+                    LOG.error("Caught unrecoverable exception in poll. Restarting input", e);
+                    // (Ab)use serverEventBus to properly restart the entire input.
+                    serverEventBus.post(InputCreated.create(input.getId()));
```
mpfz0r: This is ugly. Maybe we should just not do this at all.
I might be misinterpreting the `KafkaException` being not retryable.
Furthermore, I don't know how to trigger one, so this isn't a "real" problem.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
mpfz0r: Nevermind. You're right :)

graylog2-server/src/main/java/org/graylog2/plugin/configuration/fields/ConfigurationField.java
```diff
@@ -42,4 +42,8 @@
     List<String> getAttributes();
 
     Map<String, Map<String, String>> getAdditionalInformation();
+
+    default int getPosition() {
+        return 100;
```
dennisoelkers: I think we should return `DEFAULT_POSITION` here instead.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -389,8 +575,18 @@ public ConfigurationRequest getRequestedConfiguration() {
                     DEFAULT_GROUP_ID,
                     "Name of the consumer group the Kafka input belongs to",
                     ConfigurationField.Optional.OPTIONAL));
+            cr.addField(new TextField(
+                    CK_CUSTOM_PROPERTIES,
+                    "Custom Kafka properties",
+                    "",
+                    "A newline separated list of Kafka properties. (e.g.: \"ssl.keystore.location=/etc/graylog/server/kafka.keystore.jks\").",
+                    ConfigurationField.Optional.OPTIONAL,
+                    110,
```
dennisoelkers: Instead of a magic number we should use a constant introduced in `ConfigurationField` here, maybe something like `BELOW_DEFAULT`?

graylog2-server/src/main/java/org/graylog2/plugin/configuration/fields/ConfigurationField.java
mpfz0r: Thanks! that was a sloppy oversight :/

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
mpfz0r: ack!

muralibasani:
Good to see this finally. Thank you @dennisoelkers

marcobelmonte:
image

PR #7513, opened 2020-02-20
Display all LDAP mapped groups

### LINKED ISSUE #7502: Make clearer which group mappings are being used
When LDAP/AD group mappings are set, changing the default group search pattern hides the mappings not matching the new pattern from the UI. However, those mappings are still persisted and will still be used when a user logs in.

## Expected Behavior
I think one of this two options would suffice:
- Only apply mappings matching the group search pattern, similarly to how the UI handles it
- Display all available mappings in the UI, letting users know about them and delete them

## Current Behavior
LDAP/AD group to role mappings done before a group search pattern change are not displayed in the UI, but still applied to users logging in.

1. Configure LDAP/AD, and add group mappings
2. Change the group search pattern so it doesn't match all configured groups
3. See that the group mapping page only displays groups matching the new group search pattern
4. Log in with a user having groups that matched previously and see that the roles are still applied

## Your Environment
* Graylog Version: Graylog-3.3.0-SNAPSHOT (62084a471cc92b7bd3eebb9782064900357bd422)

edmundoa:
In the LDAP group mapping page, display groups having a mapping configured, regardless if they match the current group configuration or not. This let users remove previously configured mappings if they ever changed their group configuration.

Additionally, add a reminder in the LDAP configuration page to review group mapping after changes in the group configuration are made.

Fixes #7502

### DIFF
--- a/graylog2-web-interface/src/components/ldap/LdapComponent.jsx
+++ b/graylog2-web-interface/src/components/ldap/LdapComponent.jsx
@@ -18,6 +18,16 @@ import LdapComponentStyle from './LdapComponent.css';
 const { RolesStore } = CombinedProvider.get('Roles');
 const { LdapActions } = CombinedProvider.get('Ldap');
 
+const GroupMappingLink = ({ text, onClick }) => {
+  return (
+    <Button bsStyle="link" bsSize="xs" className="btn-text" onClick={onClick}>{text}</Button>
+  );
+};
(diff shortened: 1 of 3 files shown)


edmundoa:
The linter fails due to errors in `TestLdapLogin`, but I consider that doing a big refactoring of the component is out of scope of this PR.

PR #7520, opened 2020-02-21
Fixing flow types for `immutable.js` by adding custom typedef file.

dennisoelkers:
This PR is the first (and biggest) in a series of changes aiming to
restore proper flow typing for external dependencies. We are
unfortunately ignoring `node_modules` completely for flow, due to these
reasons:

  * type checking all files in `node_modules` costs time and memory
  * type checking of `node_modules` might result in errors in 3rd party
deps which we do not control, due to them being written for an older,
less strict version of flow

At the same time, we want to have proper type checking for _our usages_
of external dependencies, `immutable.js` being the biggest and most
important piece being used in flow-typed code so far. Therefore, the
proposition is to remove `node_modules` from the `[ignore]` part of
`.flowconfig` and move it into a `[declarations]` block, which leads to
it being used for declarations, but not being type-checked itself.

Doing this results in >260 flow errors being thrown that need to be
addressed, which would result in a gigantic PR. Therefore, I have
decided to perform the conversion in two steps:

  1. Not changing `.flowconfig`, but adding a proper typedef file for
`immutable.js` to `flow-typed` and fixing all resulting errors
  2. Moving `node_modules` from `[ignore]` to `[declarations]` and
fixing the remaining errors

This PR is performing 1.), adding a custom(!) typedef file based on the
one included in `3.8.2` to `flow-typed` and fixing the errors found by
flow. While fixing these, the following order was used to determine the
kind of change to ease reviewing:

  1. If an error is fixable by adding an explicit type annotation, this
is preferred
  2. If a code change is necessary, but it can be done either in the
test or the SUT, the first is preferred
  3. Doing a code change in the SUT/module itself

This order is used to keep the required code changes in productive code
to a minimum and therefore make reviewing this PR easier. Nevertheless,
it leads to the changes not being perfect, there are a couple of areas
where flow types could be improved, but this should be performed in
future, smaller PRs. This PR is solely targeted to add flow types and
eliminate resulting flow errors with minimal productive code changes.

The custom flow typedef file for `immutable.js` is required, because the
one included in the package is missing e.g. statics (for instantiation)
for `OrderedMap` & `OrderedSet` which we use quite often. In addition,
upgrading to `immutable.js` 4.0 was not an option right now, as it has
semantic changes that require rewriting parts of the code.

### DIFF
--- a/graylog2-web-interface/src/views/components/datatable/DataTableEntry.test.jsx
+++ b/graylog2-web-interface/src/views/components/datatable/DataTableEntry.test.jsx
@@ -1,6 +1,6 @@
 // @flow strict
 import * as React from 'react';
-import * as Immutable from 'immutable';
+import { List, OrderedSet } from 'immutable';
 import { mount } from 'wrappedEnzyme';
 
 import mockComponent from 'helpers/mocking/MockComponent';
@@ -8,13 +8,13 @@ import mockComponent from 'helpers/mocking/MockComponent';
 import FieldTypeMapping from 'views/logic/fieldtypes/FieldTypeMapping';
 import { FieldTypes } from 'views/logic/fieldtypes/FieldType';
 import Series from 'views/logic/aggregationbuilder/Series';
+import SeriesConfig from 'views/logic/aggregationbuilder/SeriesConfig';
 import DataTableEntry from './DataTableEntry';
-import SeriesConfig from '../../logic/aggregationbuilder/SeriesConfig';
 import EmptyValue from '../EmptyValue';
 
 jest.mock('views/components/common/UserTimezoneTimestamp', () => mockComponent('UserTimezoneTimestamp'));
 
-const fields = Immutable.OrderedSet(['nf_dst_address', 'count()', 'max(timestamp)', 'card(timestamp)']);
+const fields = OrderedSet(['nf_dst_address', 'count()', 'max(timestamp)', 'card(timestamp)']);
 const item = {
   nf_dst_address: '192.168.1.24',
   nf_proto_name: {
@@ -57,7 +57,7 @@ describe('DataTableEntry', () => {
                         fields={fields}
                         item={item}
                         series={series}
-                        types={[]}
+                        types={List([])}
                         valuePath={valuePath} />
       </table>
     ));
@@ -80,7 +80,7 @@ describe('DataTableEntry', () => {
                         fields={fields}
                         item={item}
                         series={series}
-                        types={types}
+                        types={List(types)}
                         valuePath={valuePath} />
       </table>
     ));
@@ -101,7 +101,7 @@ describe('DataTableEntry', () => {
                         fields={fields}
                         item={item}
                         series={series}
-                        types={[]}
+                        types={List([])}
                         valuePath={valuePath} />
       </table>
     ));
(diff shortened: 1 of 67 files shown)


graylog2-web-interface/src/views/components/datatable/DataTableEntry.test.jsx
```diff
@@ -1,20 +1,20 @@
 // @flow strict
 import * as React from 'react';
-import * as Immutable from 'immutable';
+import { List, OrderedSet } from 'immutable';
```
kyleknighted: Everywhere else we use 'immutable' we use `import * as Immutable from 'immutable';`

Should we not destruct here to keep consistent with the rest of the codebase?

dennisoelkers: We do not use one form consistently across the codebase (see `ViewState`, `ViewTransformer`, `CopyWidgetToDashboard` for examples of importing individual parts. Both ways have pros and cons (one being explicit, the other being shorted and allowing tree shaking), so I would not vote for enforcing one style.

graylog2-web-interface/src/views/logic/searchtypes/aggregation/PivotConfigGenerator.js
```diff
@@ -84,7 +84,7 @@ export default ({ config }: { config: AggregationWidgetConfig }) => {
 };
 
 class ConfigBuilder {
-  value: Set;
+  value: Set<any>;
```
linuspahl: Instead of any, we can use an object here with `{  id: string, name: string, type: string, config?: {}}`. I could not find a different use case.

graylog2-web-interface/src/views/logic/aggregationbuilder/visualizations/BarVisualizationConfig.js
```diff
@@ -49,10 +49,11 @@ export default class BarVisualizationConfig extends VisualizationConfig {
   }
 }
 
+type InternalBuilderState = Immutable.Map<string, any>;
```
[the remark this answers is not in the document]
dennisoelkers: That would mean that every value in the state of the builder would be of type `BarVisualizationConfigType`, which is not the case. Unfortunately we cannot really type the builder state stricter, because we have different types for the values of different keys.

I have stumbled across this too and was thinking a bit about it. We could use either JS plain objects or [Immutable Records](https://immutable-js.github.io/immutable-js/docs/#/Record) for this, but it requires changing the code and is therefore out of scope for this PR.

graylog2-web-interface/src/views/logic/searchtypes/aggregation/PivotConfigGenerator.js
dennisoelkers: Added that, thanks!

graylog2-web-interface/src/views/logic/aggregationbuilder/visualizations/BarVisualizationConfig.js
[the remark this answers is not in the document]
linuspahl: Sounds reasonable, thanks for the clarification.

PR #7526, opened 2020-02-22
Don't expose KafkaJournalConfiguration via the HTTP API

### LINKED ISSUE #7521: Disk journal and other node metrics do not load (3.2.2) with nginx as a reverse proxy
## Expected Behavior

There should not be an error when viewing the page for a node.

## Current Behavior

On two systems using Nginx as a reverse proxy (one using a subpath and the other using a subdomain), upon going to the page there's a red box with an error 400. "Getting journal information on node \*\* failed. 

image

Additionally there is no sign of activity in the Buffers area. 

image

Messages are still being processed.

The behaviour is the same whether using a subpath or not. I do not know if this is a problem when not using a reverse proxy.

1. Install graylog behind a reverse proxy.
2. Try to go to a node's page 
3.
4.

For example:
journalinfo

## Your Environment

* Graylog Version: Graylog 3.2.2 (docker (tag graylog:3.2) and on another system I'm using the Debian package (3.2.2-1))
* Elasticsearch Version: docker tag elasticsearch-oss:6.8.2
* MongoDB Version: 3.6.17 (docker tag mongo:3)
* Operating System: Debian unstable
* Browser version: Firefox 73 & Chrome 80.0.3987.116

bernd:
This fixes the "/api/cluster/<node-id>/journal" API endpoint to not
throw an error anymore.

In #7359 we introduced the PathConfiguration#getNativeLibDir() method
and this broke deserialization of the JournalSummaryResponse because
"#getNativeLibDir" was automatically serialized but couldn't be
deserialized.

This change introduces a new KafkaJournalConfigurationSummary value that
we can serialize and deserialize in the API. Using this also avoids
exposing new fields from Pathconfiguration or KafkaJournalConfiguration
automatically.
This also removes the constructor and jackson annotations in
KafkaJournalConfiguration beacuse they are not needed anymore.

Fixes #7521

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/KafkaJournalConfigurationSummary.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/KafkaJournalConfigurationSummary.java
@@ -1,0 +1,86 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.rest.resources.system.responses;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import org.graylog.autovalue.WithBeanGetter;
+import org.graylog2.plugin.KafkaJournalConfiguration;
+import org.joda.time.Duration;
+
+import java.nio.file.Path;
+
+@AutoValue
+@WithBeanGetter
+@JsonAutoDetect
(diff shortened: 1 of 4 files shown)


bernd:
@mpfz0r This needs to be backported for the 3.2.3 release. We can either backport the complete PR or just add a `@JsonIgnore` annotation to `PathConfiguration#getNativeLibDir()` for a minimal fix.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/KafkaJournalConfigurationSummary.java
```diff
@@ -0,0 +1,86 @@
...
+@WithBeanGetter
+@JsonAutoDetect
+public abstract class KafkaJournalConfigurationSummary {
+    private static final String FIELD_DIRECTORY = "directory";
+    private static final String FIELD_SEGMENT_SIZE = "segment_size";
+    private static final String FIELD_SEGMENT_AGE = "segment_age";
+    private static final String FIELD_MAX_SIZE = "max_size";
+    private static final String FIELD_MAX_AGE = "max_age";
+    private static final String FIELD_FLUSH_INTERVAL = "flush_interval";
+    private static final String FIELD_FLUSH_AGE = "flush_age";
```
mpfz0r: I wonder if we should include the formerly exposed fields `bin_dir`, `data_dir` and `plugin_dir` so we don't change the API?

bernd: I think it's okay to change the API for this. The path data was never supposed to be part of the response and also doesn't make much sense in that context.

mpfz0r: Fine with me, I added the api change label, just in case

mpfz0r:
> @mpfz0r This needs to be backported for the 3.2.3 release. We can either backport the complete PR or just add a `@JsonIgnore` annotation to `PathConfiguration#getNativeLibDir()` for a minimal fix.

I'm going for the `@JsonIgnore` approach

PR #7384, opened 2020-02-04
Remove disconnected.less

kyleknighted:
## Description

Continuation of #7382
Removing another LESS file

## Motivation and Context

Remove LESS as a build dependency 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #7539, opened 2020-02-24
Add 200ms delay for displaying loading spinner

linuspahl:
This PR adds a 200ms delay for displaying the loading spinner to improve the look and feel.

The idea:
Everything that loads faster than 200ms does not need a loading spinner. When something takes longer than 200ms to load, the user will not really see the delay, because it is so short.

I've already created a PR for this change in the past (https://github.com/Graylog2/graylog2-server/pull/6868), but did not managed to merge it before the release testing phase. I've decided to create a new, fresh PR, due to the amount and type of changes we've made since I've created the first PR.

Please have a look at the related PR in the enterprise repo, we need to merge the enterprise PR first.


graylog2-web-interface/src/components/common/Spinner.test.jsx
```diff
@@ -1,17 +1,21 @@
 import React from 'react';
-import { mount } from 'wrappedEnzyme';
+import { render, cleanup } from 'wrappedTestingLibrary';
```
kyleknighted: Should we have a test for the `delay`?

linuspahl: I added a test, it only tests styling, but I agree it makes sense.

graylog2-web-interface/src/components/common/Spinner.jsx
```diff
@@ -1,22 +1,62 @@
...
+  fixedWidth?: boolean,
+  name?: string,
+  pulse: boolean,
+  text?: string,
+}
+
 /**
  * Simple spinner to use while waiting for something to load.
  */
-const Spinner = ({ name, text, ...props }) => <span><Icon name={name} spin {...props} /> {text}</span>;
```
linuspahl: I had a problem with the usage of {...props} or in my case {...rest} and replaced it with some specific props and prop definitions. i hope I found every use case, so far I only found `pulse` and `fixedWidth`

image

linuspahl: I saw the usage of `{...rest}` in other components (e.g. `views/components/Select.jsx`) where we are using flow as well. I don't really understand why we do not see the same error there.

edmundoa: I was testing locally, and I can see the error when the spread happens at the end of the list of props.

This throws the error:
```
<Icon name={name} spin {...props} /> {text}
```

This seems to be fine:
```
<Icon {...props} name={name} spin /> {text}
```

I guess that makes sense, since the rest props could contain a `spin` property that overrides the value, when the spread happens after the prop is set.

linuspahl: Thanks for the clarification, this totally makes sense! I will reimplement the `{...props}` because I like the flexibility.

linuspahl: I just realized the update of `eslint-sonfig-airbnb` https://github.com/Graylog2/graylog2-server/pull/7586 forbids the props spreading. We need to discuss If we really want this, but the explanation contains some good points. I will not reimplement the props spreading for now.

edmundoa: Sounds good, I think for this component is also fine having them explicitly set.

linuspahl:
@edmundoa Thanks for the good feedback, I've disabled the spinner delay for the loading page. Depending on which PR gets merged first, this can result in a conflict with: https://github.com/Graylog2/graylog2-server/pull/7384
But it would be an easy one.

PR #7559, opened 2020-02-26
Remove not needed @testing-library/jest-dom imports

linuspahl:
The import of `@testing-library/jest-dom` and `@testing-library/jest-dom/extend-expect` is no longer required for a specific test, because the import is now part of `wrappedTestingLibrary.jsx`

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/AggregationControls.test.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/AggregationControls.test.jsx
@@ -3,7 +3,6 @@ import * as React from 'react';
 import * as Immutable from 'immutable';
 import { mount } from 'wrappedEnzyme';
 import { render } from 'wrappedTestingLibrary';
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/components/aggregationbuilder/AggregationControls.test.jsx
```diff
@@ -3,7 +3,6 @@ import * as React from 'react';
 import * as Immutable from 'immutable';
 import { mount } from 'wrappedEnzyme';
 import { render } from 'wrappedTestingLibrary';
-import '@testing-library/jest-dom/extend-expect';
```
kyleknighted: Don't we need this for [`toHaveTextContent`](https://github.com/testing-library/jest-dom#tohavetextcontent) on line 38?

kyleknighted: Oh wait, that's right, it was moved to wrappedTestingLibrary.

PR #7569, opened 2020-02-27
Using field type of sort field for unmapped type in message list.

### LINKED ISSUE #6490: Use "unmapped_type" option when sorting on custom fields in searches
When users run searches that sort the messages on a custom field, the search query will throw an error if the field that is being sorted on doesn't exist in all index mappings.

Example error message:
```
No mapping found for [event_id] in order to sort on
```

The error above can happen if a user is searching in `stream-a` and `stream-b`, sorting on field `event_id` and only indices in the index set for `stream-a` have the `event_id` field in their index mapping.

In the old search this mostly happened for admin users because they were the only ones being able to search in all streams at once. With views in <=3.1 and the upcoming search in 3.2 this will probably happen more often because regular users are able to search in multiple streams at once.

These queries can be executed without error if we use the [unmapped_type](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#_ignoring_unmapped_fields) option for the search request. This requires us to specify a data type value (e.g. `long`) for the `unmapped_type` option.

## References

The following issues are related to this problem:

- https://github.com/Graylog2/graylog2-server/issues/4367
- https://github.com/Graylog2/graylog2-server/issues/6378
- https://github.com/Graylog2/graylog2-server/issues/6384

A workaround for the event index sets in older Graylog versions is getting implemented in https://github.com/Graylog2/graylog2-server/pull/6413.

dennisoelkers:
## Motivation and Context

**Note: Requires backport to `3.2`.**

Before this change, no `unmapped_type` option was passed to Elasticsearch, when a field was used for sorting retrieved messages. This has the effect that when a field that is being sorted on is not present in all indices, an exception is thrown from ES.

This change is doing these things to address this:

  * Providing a map of field types to all search types through the `ESGeneratedQueryContext`, scoped to the effective streams of this search type
  * Using field types to look up the type of all fields used sorting
  * Adding the `unmapped_type` option to a sort in the `ESMessageList` class, using the retrieved field type for the respective field

Fixes #6490.

## Types of changes

## Checklist:


graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageListTest.java
```diff
@@ -88,6 +92,47 @@ public void appliesDecoratorsToQueryStringIfHighlightingActivated() {
         JsonPathAssert.assertThat(doc).jsonPathAsString("$.highlight.highlight_query.query_string.query").isEqualTo("Foobar!");
     }
 
+    @Test
+    public void passesTypeOfSortingFieldAsUnmappedType() {
+        final MessageList messageList = someMessageListWithSorting();
+        final ESGeneratedQueryContext context = mockQueryContext(messageList);
+        when(context.fieldTypes(Collections.singleton("stream1"))).thenReturn(Collections.singletonMap("somefield", Collections.singleton("long")));
```
alex-konn: It's unclear at first glance where "stream1" and "someField" come from.
I'd suggest either declaring them in variables at the start of test and passing them down or extracting them from the messageList object.

graylog2-server/src/main/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageList.java
```diff
@@ -83,8 +85,22 @@ public void doGenerateQueryPart(SearchJob job, Query query, MessageList messageL
...
-        sorts.forEach(sort -> searchSourceBuilder.sort(sort.field(), sort.order()));
+        sorts.forEach(sort -> searchSourceBuilder.sort(
+                SortBuilders.fieldSort(sort.field())
+                        .order(sort.order())
+                        .unmappedType(unmappedTypeForSort(sort, queryContext, effectiveStreamIds))
+        ));
+    }
+
+    private String unmappedTypeForSort(Sort sort, ESGeneratedQueryContext queryContext, Set<String> effectiveStreamIds) {
+        final Map<String, Set<String>> allFieldTypes = firstNonNull(queryContext.fieldTypes(effectiveStreamIds), Collections.emptyMap());
```
alex-konn: I'd suggest pushing the null check down to `ESGeneratedQueryContext` or even `FieldTypesLookup`.

I even think all of the logic in this method could be pushed down to `FieldTypesLookup`. Instead of asking for all field types for these streams, ask explicitly for the type of the field, i.e., `FieldTypesLookup#getType(String fieldName, Set<String> streamIds)`. That would encapsulate all this stuff in `FieldTypesLookup` and this method could be replaced with: `queryContext.getFieldType(sort.field(), effectiveStreamIds)`

graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageListTest.java
```diff
@@ -88,6 +92,47 @@ public void appliesDecoratorsToQueryStringIfHighlightingActivated() {
...
+
+        final ESGeneratedQueryContext queryContext = generateQueryPartWithContextFor(messageList, true, Collections.emptySet(), context);
+
+        final DocumentContext doc = JsonPath.parse(queryContext.searchSourceBuilder(messageList).toString());
+        JsonPathAssert.assertThat(doc).jsonPathAsString("$.sort[0].somefield.unmapped_type").isEqualTo("long");
+    }
+
+    @Test
```
alex-konn: This and the next test would become obsolete, if you followed my suggestion above to move that logic into `FieldTypeLookup`. I would add a few simpler unit tests for `FieldTypesLookup` instead.

graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageListTest.java
dennisoelkers: I decided to pass them to the method generating the message list, so they are close to the assertion.

graylog2-server/src/main/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageList.java
```diff
@@ -83,8 +85,22 @@ public void doGenerateQueryPart(SearchJob job, Query query, MessageList messageL
...
         final List<Sort> sorts = firstNonNull(messageList.sort(), Collections.singletonList(Sort.create(Message.FIELD_TIMESTAMP, SortOrder.DESC)));
-        sorts.forEach(sort -> searchSourceBuilder.sort(sort.field(), sort.order()));
+        sorts.forEach(sort -> searchSourceBuilder.sort(
+                SortBuilders.fieldSort(sort.field())
+                        .order(sort.order())
+                        .unmappedType(unmappedTypeForSort(sort, queryContext, effectiveStreamIds))
+        ));
+    }
+
+    private String unmappedTypeForSort(Sort sort, ESGeneratedQueryContext queryContext, Set<String> effectiveStreamIds) {
```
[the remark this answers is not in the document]
dennisoelkers: ✅

graylog2-server/src/main/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageList.java
dennisoelkers: Good idea! ✅

graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/searchtypes/ESMessageListTest.java
dennisoelkers: ✅

PR #7586, opened 2020-02-28
Update eslint-config-airbnb from 17.1.1 to 18.1.0

linuspahl:
Please have a look at the changelog for a detailed overview of the important changes: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/CHANGELOG.md

There are a few rules we may want to discuss:
* prop spreading will no longer be allowed. (I disabled this rule, have a look at the PR comments for a detailed explanation)
* `Object.assign` will be converted to a destructuring assignment. So far I had no problems.

Other changes are:
* `.find(variable => {})` will be `.find((variable) => {})`. Currently `const foo = bar => {}` will already be transformed to `const foo = (bar) => {}`
* `<React.Fragments>` will be `<>`
* PropTypes need to be declared outside the class body
* State initialization needs to be part of the constructor

### DIFF
--- a/graylog2-web-interface/packages/eslint-config-graylog/package.json
+++ b/graylog2-web-interface/packages/eslint-config-graylog/package.json
@@ -14,7 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "eslint": "5.16.0",
-    "eslint-config-airbnb": "18.0.1",
+    "eslint-config-airbnb": "18.1.0",
     "eslint-import-resolver-webpack": "0.12.1",
     "eslint-plugin-flowtype": "^4.5.2",
     "eslint-plugin-import": "2.20.1",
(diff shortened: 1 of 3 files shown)


dennisoelkers:
> prop spreading will no longer be allowed.

This is actually a benefit for typing things with flow. You can also use something like `<Foo {...props} />` with flow, but there is the potential that it leads to complex errors when you are doing e.g. `<Foo {...props} {...extraProps} />`, because the branches for different types get large pretty quick.

> `Object.assign` will be converted to a destructuring assignment. So far I had no problems.

This is fine, imo. I could not think of a scenario where it makes things more complex/less readable. At the same time, I do not think this rule provides significant value, so I would also be fine disabling it if others dislike it.

linuspahl:
@dennisoelkers Regarding props spreading: Good point, I will disable this rule under these circumstances.

graylog2-web-interface/packages/eslint-config-graylog/package.json
```diff
@@ -14,7 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "eslint": "5.16.0",
-    "eslint-config-airbnb": "17.1.1",
+    "eslint-config-airbnb": "^18.0.1",
```
kyleknighted: Should we remove the `^` so we stay pinned at the specific version?

PR #7605, opened 2020-03-02
Show pivots with a value of zero in chart widgets

linuspahl:
With this PR, chart widgets display pivots with a value of zero. It looks like the one line I've adjusted is the only needed change.

I had a look at all different chart types with different compositions and could not find / think of a case were this could result in an unintended visualization.

Only when all values of e.g. a line chart are zero, the line is. hardly visible:
image

I thought about starting the y-axis with a negative value, but I would not adjust the current behaviour only for this case. What do you think?

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/visualizations/ChartData.js
+++ b/graylog2-web-interface/src/views/components/visualizations/ChartData.js
@@ -53,7 +53,7 @@ export const extractSeries = (keyJoiner: KeyJoiner = _defaultKeyJoiner, leafValu
     flatLeafs.forEach(([key, value]) => {
       const joinedKey = keyJoiner(value.key);
       const targetIdx = xLabels.findIndex(l => isEqual(l, key));
-      if (value.value) {
+      if (value.value === 0 || value.value) {
         set(valuesBySeries, [joinedKey, targetIdx], value.value);
       }
     });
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/components/visualizations/ChartData.js
```diff
@@ -53,7 +53,7 @@ export const extractSeries = (keyJoiner: KeyJoiner = _defaultKeyJoiner, leafValu
     flatLeafs.forEach(([key, value]) => {
       const joinedKey = keyJoiner(value.key);
       const targetIdx = xLabels.findIndex(l => isEqual(l, key));
-      if (value.value) {
+      if (value.value === 0 || value.value) {
```
dennisoelkers: What about testing that it's neither `undefined` nor `null` at this point? Otherwise we would need to check for other falsy values as well.

```suggestion
      if (value.value !== null && value.value !== undefined) {
```

linuspahl: What would be the benefit of including falsy values like an empty string or NaN?

dennisoelkers: `extractSeries` is meant to be reused, so we should take care of its single purpose of extracting values. `null` or `undefined` would mean the absence of a value, while `''` `NaN` or `false` do not.

linuspahl: Alright!

kmerz:
Before:
Graylog (2)
After:
Graylog (1)

PR #7607, opened 2020-03-02
Specify ES search timeout to match the HTTP request timeout

### LINKED ISSUE #4614: Elasticsearch getting stuck on long running queries that are not canceled
Recently we notice our Graylog instance getting stuck regularly. Stuck means: Process buffer 100%, Input and Output buffer empty, Disk Journal piling up, no / few messages written to Elasticsearch. 

Further investigation shows several long running queries in ES - usually with empty filter, time frame from 1-1-1970 until now, all indices (we have several hundreds of them). Fix is cancelling all those queries from ES, the ES starts consuming messages again immediately after cancel operation.

Listing queries: curl -XGET 'localhost:9200/_tasks?pretty&detailed' 
Canceling queries:  curl -XPOST 'localhost:9200/_tasks/_cancel?actions=*search

The search in the graylog WebUI times out as specified in the server.conf using elasticsearch_request_timeout.

## Expected Behavior

Query in ES should by canceled /  timeout as specified in elasticsearch_request_timeout

## Current Behavior

WebUI shows error message, but search query is still running  in ES. Several of those queries leave ES stuck.

## Possible Solution

Cancel query in ES or set timeout on ES query.

1. set search time frame to "Search in all messages"
2. leave query field empty
3. start search
4. check running queries in ES using ES API

## Your Environment

1 Graylog server
1 ES master node
2 ES data nodes

* Graylog Version: 2.4.3
* Elasticsearch Version: 5.6.8 (and previously on 2.5.x)
* MongoDB Version:  3.6.2
* Operating System: Centos 7
* Browser version: Chrome / Firefox

cbarton:
## Description
The `elasticsearch_request_timeout` configuration value will set the
HTTP request timeout of ES operations but it does not tell ES to cancel
the search request with that timeout. This means the request will
continue to run in ES to completion while the client has timed out. In
turn, this could cause the user to resubmit the request which would
issue another long running query to stack against and overwhelm the
ES cluster.

## Motivation and Context
Fixes #4614

## How Has This Been Tested?

Tested by monitoring `_tasks` for a search that times out in a few seconds and noticed that the task goes away when the request times out.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/search/MoreSearch.java
+++ b/graylog2-server/src/main/java/org/graylog/events/search/MoreSearch.java
@@ -17,6 +17,7 @@
 package org.graylog.events.search;
 
 import com.codahale.metrics.MetricRegistry;
+import com.github.joschi.jadconfig.util.Duration;
 import com.google.auto.value.AutoValue;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
@@ -26,6 +27,7 @@ import io.searchbox.core.Search;
 import io.searchbox.core.SearchResult;
 import io.searchbox.core.search.sort.Sort;
 import io.searchbox.params.Parameters;
+import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.common.xcontent.ToXContent;
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
@@ -61,6 +63,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Collections;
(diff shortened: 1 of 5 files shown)


cbarton:
@florianvolle any update on this?

mpfz0r:
Thank you for the PR @cbarton 

I've done some research on the search request timeout, and there seem to be quite a few cases
where the timeout is just ignored by ES:
see "WARNING" under:
https://www.elastic.co/guide/en/elasticsearch/guide/current/_search_options.html#_timeout_2

It seems they timeout is only relevant for Lucene itself, expensive long running aggregations
done by ES won't be cancelled. https://github.com/elastic/elasticsearch/issues/9156

I also didn't manage to create an expensive query where the timeout was obeyed.

Therefore I think we should fix this either via the tasks API, or maybe `searchAsync` once we support ES7.

Furthermore, the `elasticsearch_request_timeout` configuration value isn't used anymore since 3.0.
The matching configuration option would be `elasticsearch_socket_timeout`

mpfz0r:
Thanks again for your work @cbarton 
As already mentioned, we are aiming for a better solution instead. Closing this.

PR #7618, opened 2020-03-04
Load `react-ace` asynchronously in `SourceCodeEditor`.

dennisoelkers:
## Motivation and Context

The `SourceCodeEditor` component is reused in plugins. It is referencing
`react-ace-builds`, which is pulling in ~700kb for every plugin (+ core)
using the component.

This PR is loading `react-ace-builds` asynchronously, by splitting out
its code into a separate asset. Overall assets size is not reduced, but
the entry bundles are ~0.7mb smaller (unminified).

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/SourceCodeEditor.jsx
```diff
@@ -11,6 +10,7 @@ import fetch from 'logic/rest/FetchProvider';
 import { Button, ButtonGroup, ButtonToolbar, OverlayTrigger, Tooltip } from 'components/graylog';
 import PipelineRulesMode from 'components/rules/mode-pipeline';
 
+import AceEditor from './AsyncAceEditor';
```
kyleknighted: Could we just use `React.lazy` ?

```suggestion
const AceEditor = React.lazy(() => import('react-ace-builds'));
```

kyleknighted: https://reactjs.org/docs/code-splitting.html#reactlazy

> This will automatically load the bundle containing the `react-ace-builds` when this component is first rendered.

dennisoelkers: Sure, I tried it a couple of weeks ago and it works. Not in this PR though. 

The benefit is close to zero though, as `loadable-component` is doing a couple of other things (loading indicator, error/fallback component) already and its size is pretty small. Nevertheless we can do it at some point in another PR, just to get rid of a dependency.

dennisoelkers:
This PR is unnecessary. I oversaw that `react-ace-builds` is already in separate chunks in plugins, because an async component is exported from `components/common/index.jsx`. Closing.

PR #7625, opened 2020-03-04
Expand empty streams to default streams for Aggregation Alerting 3.2

### LINKED ISSUE #7619: Aggregation Event Definition does not work if no streams are configured
An Aggregation Event Definition does not yield any results if it doesn't have any streams configured.

The empty stream list is not expanded to a default set of streams, leading
to an ES query against no specific indices.
This query never returns any results.

Furthermore, this leads to errors in the elastic search log, because it might query indices that are
not under the control of Graylog and don't conform to our mapping.

* Graylog Version: 3.1, 3.2, 3.3-snapshot
* Elasticsearch Version: 6.8.6

kmerz:
We cannot run a query with an empty stream list, because it will
be executed against all known indices in elastic search
and never yield a result.

Thus, like we already do in MoreSearch: If no streams are provided,
use all streams instead.

Fixes #7619

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
@@ -39,6 +39,7 @@ import org.graylog.plugins.views.search.errors.QueryError;
 import org.graylog.plugins.views.search.errors.SearchError;
 import org.graylog.plugins.views.search.filter.OrFilter;
 import org.graylog.plugins.views.search.filter.StreamFilter;
+import org.graylog.plugins.views.search.rest.PermittedStreams;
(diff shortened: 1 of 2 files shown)


mpfz0r:
build fails because jenkins is hosed. it checks out master for all other repos:

```Repository: git@github.com:Graylog2/graylog-plugin-collector.git
Cloning git@github.com:Graylog2/graylog-plugin-collector.git into /var/lib/jenkins/workspace/graylog-project-pr-snapshot@3/.repos/graylog-plugin-collector (shallow clone)
    git clone --depth=1 --no-single-branch git@github.com:Graylog2/graylog-plugin-collector.git /var/lib/jenkins/workspace/graylog-project-pr-snapshot@3/.repos/graylog-plugin-collector
      Cloning into '/var/lib/jenkins/workspace/graylog-project-pr-snapshot@3/.repos/graylog-plugin-collector'...
Checkout revision: master```

PR #7626, opened 2020-03-04
Cleaning up old message components.

dennisoelkers:
## Motivation and Context

During the search cleanup, some message components were left but
components they are referencing through the `components/search` package
index were removed, as rewriting the components was out of scope for
that PR.

This PR is now cleaning up those components and reducing them to their
specific use case. They are being used in two places only:

  * When loading a message for configuring an extractor
  * When running a pipeline against a message in the pipeline simulator

As these are specific use cases that require customization, these
components are kept and the message components from the new search are
not reused.

During the course of cleaning up and stripping down these components,
178 linter hints were fixed.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/search/MessageFieldExtractorActions.jsx
```diff
@@ -34,15 +34,16 @@ class MessageFieldExtractorActions extends React.Component {
   };
 
   render() {
-    const messageField = this.props.message.fields[this.props.fieldName];
```
linuspahl: Not really part of this PR, but could you move the `key` prop in the above `_formatExtractorMenuItem` method from the `MenuItem` to the `LinkContainer`? This will fix a warning. Otherwise I can do this in a separate PR.

dennisoelkers: ✅

graylog2-web-interface/src/components/messageloaders/LoaderTabs.jsx
```diff
@@ -106,22 +82,25 @@ const LoaderTabs = createReactClass({
...
-  },
+  };
 
-  _formatMessageLoaders() {
+  _formatMessageLoaders = () => {
     const messageLoaders = [];
 
     if (this._isTabVisible('recent')) {
+      const { inputs } = this.props;
+      const { selectedInputId } = this.props;
```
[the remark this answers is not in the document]
dennisoelkers: ✅

dennisoelkers:
Good catches, @linuspahl! I have fixed them and rebased the PR.

graylog2-server/src/main/java/org/graylog2/inputs/transports/KafkaTransport.java
```diff
@@ -187,7 +186,9 @@ public void run() {
         final Properties props = new Properties();
 
         props.put("group.id", configuration.getString(CK_GROUP_ID, DEFAULT_GROUP_ID));
-        props.put("client.id", "gl2-" + nodeId + "-" + input.getId());
+        props.put("client.id", "gl2-" + nodeId + "-" + input.getId(
+
+        ));
```
linuspahl: @dennisoelkers looks like this slipped in

dennisoelkers: Thanks! Fixed.

PR #7642, opened 2020-03-05
Http polling with links

lingpri:
the url for the `Request.Builder` in `HttpPollTransport`, changes dynamically, based on the link tag returned from the server response.

## Description
* added a nextlink attribute, which is a dynamic url used for paginating the next batch of records.
* useful for okta http polling.

## Motivation and Context
[okta](https://github.com/Graylog2/graylog-plugin-enterprise-integrations/issues/123)

## How Has This Been Tested?
* unit testing for sunny day scenarios
* we could do much better to add more scenario and concurrency testing 

http_poll_transport

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
@@ -53,6 +53,7 @@ import java.util.Map;
 public class GelfCodec extends AbstractCodec {
     private static final Logger log = LoggerFactory.getLogger(GelfCodec.class);
     private static final String CK_DECOMPRESS_SIZE_LIMIT = "decompress_size_limit";
+    public static final String NAME = "gelf";
     private static final int DEFAULT_DECOMPRESS_SIZE_LIMIT = 8388608;
 
     private final GelfChunkAggregator aggregator;
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpPollTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpPollTransport.java
@@ -1,16 +1,16 @@
 /**
  * This file is part of Graylog.
- * <p>
+ *
  * Graylog is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * <p>
+ *
  * Graylog is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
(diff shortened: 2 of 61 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
```diff
@@ -53,6 +53,7 @@
 public class GelfCodec extends AbstractCodec {
     private static final Logger log = LoggerFactory.getLogger(GelfCodec.class);
     private static final String CK_DECOMPRESS_SIZE_LIMIT = "decompress_size_limit";
+    public static final String NAME = "gelf";
```
danotorrey: Is this needed for anything?

graylog2-server/src/main/java/org/graylog2/inputs/transports/HttpPollTransport.java
```diff
@@ -30,10 +30,7 @@
 import org.graylog2.plugin.ServerStatus;
 import org.graylog2.plugin.configuration.Configuration;
 import org.graylog2.plugin.configuration.ConfigurationRequest;
-import org.graylog2.plugin.configuration.fields.ConfigurationField;
-import org.graylog2.plugin.configuration.fields.DropdownField;
-import org.graylog2.plugin.configuration.fields.NumberField;
-import org.graylog2.plugin.configuration.fields.TextField;
+import org.graylog2.plugin.configuration.fields.*;
```
danotorrey: `*` imports are generally not used in Graylog projects. As a general standard, we always import specifically only the needed classes. This can be done by setting the 'Class count to use import with '*'' to a very high value as described in this article:

https://intellij-support.jetbrains.com/hc/en-us/community/posts/206203659-Turn-off-Wildcard-imports-

graylog2-server/src/main/java/org/graylog2/inputs/codecs/GelfCodec.java
lingpri: the signature for adding a codec into a plugin module is `protected void addCodec(String name, Class<? extends Codec> codecClass)` . I `NAME` attribute was missing in the GelfCodec and hence I added it, part of this PR. 
After speaking with @bernd , I think we will need a new codec called `OktaCodec` for the OktaInput we are developing. Not sure, if we can add any Codec into a plugin module, without a `name` attribute . Let me know what you think ?

danotorrey: You are correct about the `NAME` constant for plugins. Thanks for clarifying :) It will be needed. I registration in the plugin module will also be needed like this:

https://github.com/Graylog2/graylog-plugin-integrations/blob/6c53151852e0332026d2075d9591fef52f18fccf/src/main/java/org/graylog/integrations/IntegrationsModule.java#L99

lingpri:
@danotorrey - I'm closing this PR. I will plan to address the GelfCodec name change in an another PR.

PR #7645, opened 2020-03-06
Implement sorting options for message table widget

### LINKED ISSUE #7119: Missing message table sorting options
## Expected Behavior

The search page in 3.1 and earlier offered the possibility to sort the message table based on fields in the table. (ascending and descending)

image

## Current Behavior

The message list widget in 3.2 doesn't have this.

## Your Environment

* Graylog Version: 3.2.0-beta.3-SNAPSHOT

linuspahl:
As described in #7119 the sorting option is not part of the new message list widget.
This PR implements the option for the new implementation of the widget.

For now the sorting can only be changed inside the message list edit modal. The functionality would also work outside the edit modal, but the changed sort would be saved with e.g. the dashboard. We need to discuss if we appreciate this behaviour or if we want to reset the sort, when the user changes it outside the edit modal.

I've added a few best practice questions as comments to the PR, but a few general points:
* I've used some classes like the `Direction` or `SortConfig`, they are currently part of the `views/logic/aggregationbuilder` directory. Would it make sense to move them to `views/logic/widgets`?
* When using the message list sort icons, we are showing a loading indicator beside the message widget title, but it will be displayed with a delay (not the spinner default spinner delay). I could not really find a way to improve the the timing, but we could e.g. directly show a loading spinner or a different kind of feedback for the clicked sort icon. 
* There is a margin problem with the loading spinner and the widget title, I will fix this in another PR.

Fixes #7119

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/widgets/FieldSortSelect.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+    const { getByText } = render(<FieldSortSelect fields={fields} onChange={() => {}} sort={sort} />);
+    expect(getByText('http_method')).not.toBeNull();
+  });
+
+  it('should open menu when focused', async () => {
+    const { getByText, container } = render(<FieldSortSelect fields={fields} onChange={() => {}} sort={sort} />);
+    fireEvent.focus(container.getElementsByTagName('input')[0]);
+    await waitForElement(() => getByText(/2 results available./));
+  });
+});
```
linuspahl: I've tried to add some meaningful tests for the `FieldSortSelect`, but I could not really find a way to test e.g. the select menu and the selection of an option. It looks like this was also a problem in the `SeriesSelect` test.

graylog2-web-interface/src/views/components/widgets/FieldSortIcon.jsx
```diff
@@ -0,0 +1,101 @@
...
+  if (fieldSortDirection && fieldSortDirection === Direction.Ascending.direction) {
+    return 'sort-amount-asc';
+  }
+  return 'sort-amount-desc';
+};
+
+
+const FieldSortIcon = ({ fieldName, config, onConfigChange, setLoadingState }: Props) => {
+  const sortDirectionName = _sortDirectionName(config, fieldName);
+  const newSortDescription = _newSortDescription(fieldName, sortDirectionName);
```
linuspahl: Would you implement `useMemo` for these two functions or are the not complex enough?

graylog2-web-interface/src/views/components/widgets/FieldSortIcon.jsx
```diff
@@ -0,0 +1,101 @@
...
+    background: transparent;
+    color: ${color};
+
+    padding: 5px;
+
+    cursor: pointer;
+  `;
+});
+
+const _newSort = (config, fieldName, currentSortDirectioName) => {
```
linuspahl: I've had some inconsistent behaviour when I tried to compare e.g. `Direction.Ascending` with the direction of a widget config sort and decided to use the direction name instead of the class.

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConifg.js
```diff
@@ -0,0 +1,36 @@
+// @flow strict
+import Direction from 'views/logic/aggregationbuilder/Direction';
+import SortConfig, { Builder } from 'views/logic/aggregationbuilder/SortConfig';
+
+export type MessageSortConifgJson = {
+  field: string,
+  order: 'ASC' | 'DESC',
+};
+
+export default class MessageSortConifg {
```
linuspahl: @dennisoelkers we talked about the usage of an extra class to transform the message sort configs when calling the api. I've tried to create a class which is as simple as possible, but it would be great to hear your opinion on how to define the relation with the `SortConfig` class. It feels a bit wrong to use e.g. the `SortConfig` Builder for the `fromJSON` method.

graylog2-web-interface/src/views/components/widgets/MessageTable.jsx
```diff
@@ -220,6 +229,12 @@ class MessageTable extends React.Component<Props, State> {
                     <Field type={this._fieldTypeFor(selectedFieldName, fields)}
                            name={selectedFieldName}
                            queryId={activeQueryId} />
+                    {editing && (
```
dennisoelkers: I guess we need to wrap that in an `<IfInteractive>...</IfInteractive>` block?

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConifg.js
```diff
@@ -0,0 +1,36 @@
+// @flow strict
+import Direction from 'views/logic/aggregationbuilder/Direction';
+import SortConfig, { Builder } from 'views/logic/aggregationbuilder/SortConfig';
+
+export type MessageSortConifgJson = {
```
dennisoelkers: There is a typo in the class, file and type names: `MessageSortConifg` -> `MessageSortConfig`

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConifg.js
dennisoelkers: I think we should extract a common base class that is used by both `MessageSortConfig` and `SortConfig`. Maybe it would make sense to rename `SortConfig` to `AggregateSortConfig` and use `SortConfig` as the name for the base class. Shared code would go in there and the two specialized classes would use it internally to hold their state, but customize construction & (de-)serialization.

graylog2-web-interface/src/views/components/widgets/FieldSortIcon.jsx
linuspahl: Answer: These functions are not complex

graylog2-web-interface/src/views/components/widgets/MessageTable.jsx
linuspahl: We are using `<IfInteractive>` only to differentiate between the big display mode. I could create a wrapper like `<IfEditing>` but we are using the edit check just a few times.

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConifg.js
linuspahl: Good catch, thanks.

linuspahl:
@dennisoelkers While testing the previous changes I had the same problem we saw last week. The sort in `AutoValue_MesageList.java` is always empty. Even when I send the requests with Postman. It is strange, because I changed nothing in the backend and the request payload is also the same. I rebased the master branch, but it would be surprising if that cause this behaviour.
Can you test if the sorting still works for you when having a look at previous changes?

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConfig.js
```diff
@@ -0,0 +1,36 @@
...
+export type MessageSortConfigJson = {
+  field: string,
+  order: 'ASC' | 'DESC',
+};
+
+export default class MessageSortConfig {
+  _value: SortConfig
+
+  constructor(type: string, field: string, direction: Direction) {
+    this._value = new SortConfig(type, field, direction);
```
dennisoelkers: I think there is no benefit reusing `SortConfig` here. Instead, we could just store the `field`/`direction` values in internal fields (we do not even need getters for them) and just define the JSON (de-)serialization format in here.

graylog2-web-interface/src/views/logic/searchtypes/messages/MessageSortConfig.js
```diff
@@ -0,0 +1,36 @@
...
+      field,
+      order: direction === Direction.Ascending ? 'ASC' : 'DESC',
+    };
+  }
+
+  static fromJSON(value: MessageSortConfigJson) {
+    const { field, order } = value;
+
+    // eslint-disable-next-line no-use-before-define
+    return new Builder()
```
dennisoelkers: This actually returns a `SortConfig` instead of a `MessageSortConfig`.

graylog2-web-interface/src/views/components/widgets/MessageList.test.jsx
```diff
@@ -108,6 +108,7 @@ describe('MessageList', () => {
     const wrapper1 = mount(<MessageList editing
                                         data={data}
                                         config={config}
+                                        onConfigChange={() => {}}
```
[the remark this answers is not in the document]
linuspahl: I've implemented this change, but out of interest: When we define a default prop for `onConfigChange` we need to remove the `.isRequired` `PropType` definition. And is it worth it? Without `onConfigChange` the component would not work properly and I would expect a `PropType` error, when using the component without all necessary attributes.

graylog2-web-interface/src/views/components/widgets/MessageList.jsx
```diff
@@ -42,35 +42,39 @@ type State = {
...
+      activeQuery: string,
+      view: {
+        id: number,
+      },
+    },
   data: { messages: Array<Object>, total: number, id: string },
-  selectedFields?: Immutable.Set<string>,
+  editing: boolean,
+  fields: FieldTypeMappingsList,
+  onConfigChange: (MessagesWidgetConfig) => Promise<void>,
```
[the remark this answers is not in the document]
linuspahl: I've created a `onSortChange` method inside the `MessageList`, so `FieldSortIcon` does not need to know how to change the config. But I did not changed `onConfigChange` in `Widget.jsx` because all widgets are receiving this function and I thought onSortChange` would be too specific.

graylog2-web-interface/src/views/components/widgets/MessageList.test.jsx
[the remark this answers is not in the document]
dennisoelkers: I think it does still work properly. You just cannot change its config, turning it into a read-only component. 

Generally: If we consider it to be a bigger risk that we are forgetting to add the new prop in a place and therefore want to be warned about it (either by flow or through a prop types warning), then we should not add a default prop. In this case the number of places where we are reusing the component is very small, so chances are small that we are forgetting it.

Also, coming back to the original issue of having to add new props in lots of places in a test suite: Another pattern might be to define a central place in the test suite where the component is rendered, something like this:

```
const renderMessageList = (props} => mount(<MessageList editing data={data} .... {...props />);
```

This way you can have a central set of default props where you are adding a new one, but you can still override individual ones in the individual tests.

PR #7671, opened 2020-03-12
Implementing `shouldComponentUpdate` for `connect`.

dennisoelkers:
## Motivation and Context

In the web interface we are using a `connect` HOC which can be used to wrap another component and specify a map of Reflux stores, whose states are synchronized with the props of the component. This allows us to use Reflux stores in functional and class-based components. This works by generating a component that gets the initial state of all stores and listens to them and update the props of its child component on every store update.

Before this change, a component wrapped with the `connect` HOC was rerendered whenever a store updated, even if the state of the store remained the same. E.g.:

  - Component `Foo` uses `connect` to synchronize with `FooStore`, but is only interested in the key `bar` of its state: `connect(FooComponent, { foo: FooStore }, ({ foo: { bar } }) => ({ bar }))`
  - `FooStore` triggers an update, because the `foo` key is updated: `this.trigger({ foo: 42 })`
  -  `FooComponent` is rerendered although `bar` is not updated

To prevent this, this PR implements a `shouldComponentUpdate` method for each generated wrapper component and returns the result of a deep equality check for the _generated_ props (after applying the props mapper function) of its child component.
Although deep equality checks are generally discouraged, this still improves rendering times for e.g. the search by ~40%.

**Note:**

While doing this, it was noticed that the `RefreshControls` did trigger starting/stopping the auto refresh of a search/dashboard, but they did not render the correct state. This was traced back to the `RefreshStore` mutating its state inline, which is generally an anti-pattern and a source for subtle bugs. In this case it lead to `shouldComponentUpdate` always lead to false, because the state of the store before and after the update appeared to be identical (due to inline mutation). This was fixed in this PR as well. **There might be other cases like this hidden in the source code.** A structural search in the code base did not reveal other cases like this which have an effect, but this should therefore not be backported to `3.2` and instead tested well in `master` for `3.3`.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/connect.jsx
+++ b/graylog2-web-interface/src/stores/connect.jsx
@@ -101,7 +101,7 @@ function connect<Stores: Object, Props, ComponentType: React.ComponentType<Props
 
     render() {
       const nextProps = this._genProps(this.state);
-      const { key, ref, ...componentProps } = this.props;
+      const { ref, ...componentProps } = this.props;
 
       return <Component {...nextProps} {...componentProps} />;
     }
--- a/graylog2-web-interface/src/stores/connect.jsx
+++ b/graylog2-web-interface/src/stores/connect.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { ListenerMethods } from 'reflux';
 import { isEqualWith, isFunction } from 'lodash';
 import * as Immutable from 'immutable';
+import isDeepEqual from './isDeepEqual';
 
 const _isEqual = (first, second) => {
   if (first && first.equals && isFunction(first.equals)) {
@@ -84,7 +85,7 @@ export default (Component, stores, mapProps = props => props) => {
       const thisChildProps = this._genProps(this.state);
       const nextChildProps = this._genProps(nextState);
 
-      return !(isEqualWith(thisChildProps, nextChildProps, _isEqual) && isEqualWith(this.props, nextProps, _isEqual));
+      return !(isDeepEqual(thisChildProps, nextChildProps) && isDeepEqual(this.props, nextProps, _isEqual));
     }
 
     _genProps = (state) => {
--- a/graylog2-web-interface/src/stores/connect.jsx
+++ b/graylog2-web-interface/src/stores/connect.jsx
@@ -9,6 +9,7 @@ type StoreType<State> = {
 };
 
 type ExtractStoreState = <V, Store: StoreType<V>>(Store) => V;
+type ExtractComponentProps = <Props>(React.ComponentType<Props>) => Props;
 
 type ResultType<Stores> = $ObjMap<Stores, ExtractStoreState>;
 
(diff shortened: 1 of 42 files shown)


graylog2-web-interface/src/stores/connect.jsx
```diff
@@ -27,57 +38,77 @@ import { isFunction } from 'lodash';
...
+function connect<Stores: Object, Props, ComponentType: React.ComponentType<Props>, MappedProps>(
+  Component: ComponentType,
+  stores: Stores,
+  mapProps: (ResultType<Stores>) => MappedProps = props => props,
+): React.ComponentType<$Diff<$Call<ExtractComponentProps, ComponentType>, MappedProps>> {
   const wrappedComponentName = Component.displayName || Component.name || 'Unknown/Anonymous';
-  class ConnectStoresWrapper extends React.Component {
-    constructor(props) {
-      super(props);
+  class ConnectStoresWrapper extends React.Component<$Diff<$Call<ExtractComponentProps, ComponentType>, MappedProps>> {
```
linuspahl: I am currently seeing the following warning, when I open any search page:
```
ConnectStoresWrapper[Widget] stores=view: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.
```

graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
```diff
@@ -37,12 +37,19 @@ export type VisualizationComponentProps = {|
...
   toggleEdit: () => void,
-|};
+};
 
 export type VisualizationComponent =
-  { type?: string, propTypes?: any }
+  { type: string, propTypes?: any }
   & React.ComponentType<VisualizationComponentProps>;
 
+export const makeVisualization = (component: React.ComponentType<VisualizationComponentProps>, type: string): VisualizationComponent => {
```
linuspahl: Due to the existing prop type warning we are receiving for some visualizations, I found out the component name gets lost here. Currently the warning looks like:
```
Warning: Failed prop type: Invalid prop `data` of type `object` supplied to `<<anonymous>>`, expected an array.
```

Which means we need to add something like `LineVisualization.displayName = 'LineVisualization'` to every visualization, or we add an argument to this function `makeVisualization(component, displayName, type)` and add it here with `visualizationComponent.displayName = displayName`.

The first solution looks more familiar because the `displayName` clearly belongs to the component, but I prefer the second approach, because we only need to add the `displayName` when we use `makeVisualization()` and flow will remind us, whenever we could forget about adding the attribute.

graylog2-web-interface/src/views/stores/isEqualForSearch.js
```diff
@@ -0,0 +1,24 @@
...
+    return first.equalsForSearch(second);
+  }
+  if (isImmutable(first) && isImmutable(second)) {
+    return isEqualWith(first.toJS(), second.toJS(), _isEqual);
+  }
+  if (hasEquals(first)) {
+    return first.equals(second);
+  }
+  return undefined;
+};
```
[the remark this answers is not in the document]
linuspahl: Just saw we can't reuse some of the logic in `isDeepEqual.js`, please ignore this comment.

graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
dennisoelkers: I think this happens for the `BarVisualization` component? In that case, I think the display name never got lost, because it was simply never present as it is a functional component. We could use the logic from `connect.jsx`:

```
  const wrappedComponentName = Component.displayName || Component.name || 'Unknown/Anonymous';
```
and use that as `displayName`, but we would not gain anything.

Long story short: I think we should just set `displayName` explicitly, when creating a functional component. E.g. adding: `BarVisualization.displayName = 'BarVisualization';` before exporting it. This is beyond the scope of this PR though.

graylog2-web-interface/src/stores/connect.jsx
dennisoelkers: ✅

graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
linuspahl: You are right, it never gets lost, I thought so because the components name was part of the warning before these changes.
I agree adding the `displayName` for functional components is a good idea in general, but out of the scope. Looks like a goos case for an ESLint rule, i will have a look at a possible solution.

PR #7674, opened 2020-03-12
Split input modal into transport and codec

mpfz0r:
Otherwise users can't see which config field is meant for transport
and which one for codec settings.

before:
image

after:
image

### DIFF
--- a/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
+++ b/graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import PropTypes from 'prop-types';
 import React from 'react';
+import lodash from 'lodash';
 
 import BootstrapModalForm from 'components/bootstrap/BootstrapModalForm';
 import { ConfigurationFormField, TitleField } from 'components/configurationforms';
@@ -18,6 +19,7 @@ class ConfigurationForm extends React.Component {
     // eslint-disable-next-line react/no-unused-prop-types
     values: PropTypes.object,
     wrapperComponent: PropTypes.elementType,
+    isInputForm: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -30,6 +32,7 @@ class ConfigurationForm extends React.Component {
     typeName: undefined,
     values: {},
     wrapperComponent: BootstrapModalForm,
+    isInputForm: false,
   };
 
   constructor(props) {
@@ -100,6 +103,22 @@ class ConfigurationForm extends React.Component {
     return diff;
   };
 
+  _renderConfigFields = (configFields, autoFocus) => {
+    let shouldAutoFocus = autoFocus;
+    const configFieldKeys = $.map(configFields, (field, name) => name)
+      .map((name, pos) => ({ name: name, pos: pos }))
+      .sort(this._sortByPosOrOptionality);
+
+    const renderedConfigFields = configFieldKeys.map((key) => {
+      const configField = this._renderConfigField(configFields[key.name], key.name, shouldAutoFocus);
+      if (shouldAutoFocus) {
+        shouldAutoFocus = false;
+      }
+      return configField;
+    });
+    return renderedConfigFields;
+  };
+
   _save = () => {
     const data = this.getValue();
 
@@ -151,7 +170,7 @@ class ConfigurationForm extends React.Component {
   };
 
   render() {
-    const { typeName, title, helpBlock, wrapperComponent: WrapperComponent = BootstrapModalForm, includeTitleField, children } = this.props;
+    const { typeName, isInputForm, title, helpBlock, wrapperComponent: WrapperComponent = BootstrapModalForm, includeTitleField, children } = this.props;
 
     let shouldAutoFocus = true;
     let titleElement;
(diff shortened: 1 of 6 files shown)


edmundoa:
I know nobody asked for my review here, but I'm wondering what is the reasoning behind this, something more than the short description above.

In my opinion we should go into abstracting our backend concepts and avoiding complexity for end users, and I don't see this aligning into that view. Technical users may be fine seeing concepts like transport and codec, but I can see many other people looking through our documentation to figure out what those are.

mpfz0r:
@edmundoa 
While setting up a Syslog/Kafka input, I couldn't make out which fields of the configuration are related to kafka and which are for syslog.
I agree that not every user would understand the terms "transport" and "codec", but I don't see any harm having them there either :man_shrugging: 

By "documentation" you mean this? https://docs.graylog.org/en/3.2/pages/sending_data.html#using-apache-kafka-as-transport-queue

edmundoa:
@mpfz0r I can understand the issue you had and we should improve how to separate those to make them clearer. However, inputs forms are sometimes very long and may have many "obscure" options, so I think we should be careful when adding more cognitive load to them.

By documentation I mean our whole product documentation, not just using the Kafka input. Look for "codec" or "input" in our documentation and you won't find any meaningful definition or explanation of what they are.

kyleknighted:
I agree with both you :) 

Maybe instead of "Transport", we could use `Network Settings` and instead of "Codec" maybe we do `Input Overrides` or something generic?

Although, ideally, we would move away from this modal version and create a better experience for this setup.

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -100,6 +103,22 @@ class ConfigurationForm extends React.Component {
     return diff;
   };
 
+  _renderConfigFields = (configFields, autoFocus) => {
+    let shouldAutoFocus = autoFocus;
+    const configFieldKeys = $.map(configFields, (field, name) => name)
```
kyleknighted, line 108: Let's use JS [`Array.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) instead of jQuery

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -1,6 +1,7 @@
 import $ from 'jquery';
 import PropTypes from 'prop-types';
 import React from 'react';
+import lodash from 'lodash';
```
kyleknighted, line 4: for treeshaking 

```suggestion
import { pickBy } from 'lodash';
```

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -168,17 +187,27 @@ class ConfigurationForm extends React.Component {
...
-      const configField = this._renderConfigField(configFields[key.name], key.name, shouldAutoFocus);
-      if (shouldAutoFocus) {
-        shouldAutoFocus = false;
-      }
-      return configField;
-    });
+    let renderedConfigFields;
+    if (isInputForm) {
+      // lodash partition
+      const transportFields = lodash.pickBy(configFields, field => field.context === 'transport');
```
kyleknighted, line 194:
```suggestion
      const transportFields = pickBy(configFields, field => field.context === 'transport');
```

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -168,17 +187,27 @@ class ConfigurationForm extends React.Component {
...
+      const codecFields = lodash.pickBy(configFields, field => field.context === 'codec');
```
kyleknighted, line 195:
```suggestion
      const codecFields = pickBy(configFields, field => field.context === 'codec');
```

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -168,17 +187,27 @@ class ConfigurationForm extends React.Component {
...
+      renderedConfigFields = (
+        <React.Fragment>
```
kyleknighted, line 197:
```suggestion
        <>
```

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.jsx
```diff
@@ -168,17 +187,27 @@ class ConfigurationForm extends React.Component {
...
+          <hr />
+          <h3>Transport Settings</h3>
+          <br />
+          { this._renderConfigFields(transportFields, shouldAutoFocus) }
+          <hr />
+          <h3>Codec Settings</h3>
+          <br />
+          { this._renderConfigFields(codecFields, false) }
+        </React.Fragment>
```
kyleknighted, line 206:
```suggestion
        </>
```

mpfz0r:
> I agree with both you :)
> 
> Maybe instead of "Transport", we could use `Network Settings` and instead of "Codec" maybe we do `Input Overrides` or something generic?
> 
> Although, ideally, we would move away from this modal version and create a better experience for this setup.

Thanks for your input @kyleknighted 
@edmundoa what do you think?
As an alternative, we could also just use the horizontal lines, without the headlines.

mpfz0r:
Nevermind. Dropping this for another reason.

Issue #7508, opened 2020-02-20
New Search should respect **UI analysis disabled for fields**


kmerz:
## Expected Behavior
The user can disable fields which take to long to analyze, so the admin disabled them in the **UI analysis disabled for fields** section of System Configuration

## Current Behavior
A user can click on `aggregate` on such a field and this violates the setting in System Configuration.

## Possible Solution
New Search should respect the System Configuration setting: **UI analysis disabled for fields**

## Context
[z#3152]

## Your Environment
* Graylog Version: 3.2.0

PR #7685, opened 2020-03-13
Disable "Show top value" and "Statistics" field actions when field analysis is disabled

linuspahl:
**Note: This PR requires a backport for 3.2**

As described in https://github.com/Graylog2/graylog2-server/issues/7508 field actions like "Show top values / aggregate" are enabled, even when the field is part of the list "UI analysis disabled for fields", which can be configured in System/Configurations.

With these changes we are simply checking if a field is part of the "analysis disabled fields" list, when determining if a field action should be enabled.

Please note, beside the mentioned "Show top values / aggregate" field action I've also added the check for the "Statistics" field action.

We need to merge the related enterprise PR first to avoid the failing test.

Fixes: #7508

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/bindings.jsx
+++ b/graylog2-web-interface/src/views/bindings.jsx
@@ -109,7 +109,7 @@ ViewSharing.registerSubtype(SpecificUsers.Type, SpecificUsers);
 Parameter.registerSubtype(ValueParameter.type, ValueParameter);
 Parameter.registerSubtype(LookupTableParameter.type, LookupTableParameter);
 
-const isAnalysisDisabled = (field: string, analysisDisabledFields: string[] = []) => !!analysisDisabledFields.includes(field);
+const isAnalysisDisabled = (field: string, analysisDisabledFields: string[] = []) => analysisDisabledFields.includes(field);
 
 export default {
   pages: {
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/bindings.jsx
```diff
@@ -109,6 +109,8 @@ ViewSharing.registerSubtype(SpecificUsers.Type, SpecificUsers);
 Parameter.registerSubtype(ValueParameter.type, ValueParameter);
 Parameter.registerSubtype(LookupTableParameter.type, LookupTableParameter);
 
+const isAnalysisDisabled = (field: string, analysisDisabledFields: string[] = []) => analysisDisabledFields.includes(field);
```
linuspahl: This function is very simple, but I would still prefer an own function to do the "analysis disabled" check. I would like to avoid, adding helper functions to the `bindings` file. Were would you add this function? Maybe as a static method of `views/logic/fieldtypes/FieldType.js`?

linuspahl: Regarding the default value for  `analysisDisabledFields`. Currently the value can't be `null` or `undefined`, but I saw no disadvantage in adding the default value. It also makes testing of related code easier.

linuspahl:
@kmerz Good catch! I just moved the `ViewAdditionalContextProvider` to the highest reasonable level.
https://github.com/Graylog2/graylog2-server/pull/7685/commits/a3fa79a8aa6953d15c522221b172e05ccc024e3c

PR #7712, opened 2020-03-18
LoginPage as functional component & canceled promise

kyleknighted:
## Description

- Rewrites `createReactClass` as functional component
- Adds promise `.cancel()` on unmount

## Motivation and Context

Component was using older ES5 code and also was causing warnings of state being set on an unmounted component.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/LoginPage.jsx
+++ b/graylog2-web-interface/src/pages/LoginPage.jsx
@@ -1,5 +1,4 @@
-import React, { useEffect, useState } from 'react';
-import { connect } from 'reflux';
+import React, { useState } from 'react';
 import { PluginStore } from 'graylog-web-plugin/plugin';
 
 import { DocumentTitle, Icon } from 'components/common';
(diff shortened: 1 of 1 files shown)


edmundoa:
The production build seems to be broken after the change. I get this error when loading the page:
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`
```

graylog2-web-interface/src/pages/LoginPage.jsx
```diff
@@ -14,80 +11,73 @@ import AuthThemeStyles from 'theme/styles/authStyles';
...
 
-const { SessionStore, SessionActions } = CombinedProvider.get('Session');
+const { SessionActions } = CombinedProvider.get('Session');
 
-const LoginPage = createReactClass({
-  displayName: 'LoginPage',
-  mixins: [Reflux.connect(SessionStore), Reflux.ListenerMethods],
+const LoginPage = () => {
+  const [didValidateSession, setDidValidateSession] = useState(false);
+  const [lastError, setLastError] = useState(false);
```
edmundoa: I think using `false` as default `lastError` is not the best thing, we should probably consider using `undefined` or `''`.

kyleknighted: I had this fixed at one point, guess i overwrote myself on a rebase

PR #7725, opened 2020-03-19
Extract IndexLookup from ElasticsearchBackend

alex-konn:
Previously the lookup of indices for streams and time range was done
in `ElasticsearchBackend` by loading indices, streams, and then finding
matches. It was impossible to reuse this logic elsewhere.
Furthermore this made `ElasticsearchBackend` harder to test, because
a lot of setup was required to make sure that proper index ranges and
streams are loaded during the test.

The corresponding change in the enterprise plugin can be found here: https://github.com/Graylog2/graylog-plugin-enterprise/pull/1498
Note that the Jenkins builds for both PRs will be red, because of this dependency. Both PRs should be merged at the same time.

## Description
- Extracts that logic into a separate class `IndexLookup`
- Simplifies tests for `ElasticsearchBackend` accordingly
- Adds tests for `IndexLookup`

## Motivation and Context
- reuse the index lookup logic for #7709
- improve testability of `ElasticsearchBackend`

## How Has This Been Tested?
Local unit and integration tests

## Types of changes

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/IndexLookupTest.java
+++ b/graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/IndexLookupTest.java
@@ -72,14 +72,6 @@ class IndexLookupTest {
         assertThat(result).containsExactly(matchingIndexRange.indexName());
     }
 
-    private RelativeRange someTimeRange() {
-        try {
-            return RelativeRange.create(1);
-        } catch (InvalidRangeParametersException e) {
-            throw new RuntimeException(e);
-        }
-    }
-
     @Test
     void returnsEmptySetForEmptyStreamIds() {
         Set<String> result = sut.indexNamesForStreamsInTimeRange(emptySet(), someTimeRange());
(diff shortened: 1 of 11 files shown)


graylog2-server/src/test/java/org/graylog/plugins/views/search/elasticsearch/IndexLookupTest.java
```diff
@@ -0,0 +1,140 @@
...
+        IndexRange matchingIndexRange = indexRanges.get(0);
+
+        sut.indexRangeContainsOneOfStreams = (i, s) -> i.equals(matchingIndexRange);
+
+        Set<String> result = sut.indexNamesForStreamsInTimeRange(streamIds, someTimeRange());
+
+        assertThat(result).containsExactly(matchingIndexRange.indexName());
+    }
+
+    private RelativeRange someTimeRange() {
```
kmerz: just a nitpick, but wouldn't it be more readable if the private function comes after the test?

alex-konn: :+1: Yup, moved it down. I placed it after the first usage in the first test, but I agree that it should go with the other helpers to the bottom of the file.

Issue #7732, opened 2020-03-23
Search query string should only be attached to widgets when exporting a search to a dashboard.


linuspahl:
When exporting a search to a dashboard, we are attaching the search query to every widget, but we are currently not removing it from the "query" itself. A query will be represented as a dashboard tab, but the query string is not visible anywhere. The mentioned query string is still getting involved in the search execution.

How to reproduce this behaviour:

1. on the main search page, search for e.g.: `http_method:GET`
2. export the search to a dashboard
4. remove the query string for the message table, click on save and execute a new search.

The message table still only shows the the results for `http_method:GET` because it still exists for the “query / dashboard tab” but we are not showing it anywhere

## Your Environment

* Graylog Version: 3.2.4

Issue #7701, opened 2020-03-18
Strange Dashboard Behavior in 3.2.x


yoke88:
After upgraded from graylog 3.1 to 3.2, the UI changed a lot ,especially the dashboard behivor.

## Expected Behavior

I have created a dashboard to overview all windows event logs to troubleshooting. so i created four tabs (view) and using different query to get a summary.
- All Errors and warnings with query `winlogbeat_level:(错误 OR Error OR 警告 OR Warning)`
- All Errors  with query `winlogbeat_level:(错误 OR Error)`
- All Warnings with query `winlogbeat_level:(警告 OR Warning)`
- NetLog Failed with query `winlogbeat_event_id:(5719 OR 4266)`

![捕获](https://user-images.githubusercontent.com/3146226/76920987-d35d0e00-6907-11ea-9b35-9978ab10d9b6.JPG)

when in graylog 3.1 every tab (the view show the search bar for every tab, and show the query i used, and i can change the query and using Drill-down to quick change the query ),after 3.2 ,the query bar for each tab disappeared and the search become global filter,the global filter will replace the search query for each widget in the dashboard.

for example:
The widget with query `winlogbeat_level:(错误 OR Error OR 警告 OR Warning)` , if i want to drill-down a host with name xxxx ,i using drill-down to add a quick query `winlogbeat_beat_name:xxxx` and i want to combine the search like this `winlogbeat_level:(错误 OR Error OR 警告 OR Warning) AND winlogbeat_beat_name:xxxx` , but the behaivor in graylog 3.2 will replace all the query with `winlogbeat_beat_name:xxxx`

so now in graylog 3.2 i have two options:
1. rewrite the whole query , but i need to know the original query,but i can not see the query of the widget.( I can see the query from chrom development tools ,but can not see it in the GUI)

> you will see in the screenshot below ,the widget shows the correct query result (with only error events) but does not show the query in the search bar
![Annotation 2020-03-18 111525](https://user-images.githubusercontent.com/3146226/76922004-64cd7f80-690a-11ea-87c9-9c81d46b9058.jpg)

2. do not using drill-down and the global-filter, boring life ,and i want to migrate to kibana

 
## Current Behavior
- widget query does not show in the GUI ,but i can capture it using chrome development tools
- dashboard global filter will replace all the widget query .which i want is "global filter is actually a filter but not a query search", or "global filter will combine the widget query"

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

[...]

PR #7734, opened 2020-03-23
Remove query_string from queries when exporting search to dashboard

linuspahl:
**Please note: This PR needs a backport for 3.2**

As described in #7732 we are currently not removing the `query_string` from the search queries when exporting a search to a dashboard. This can result in a not visible but invoked search query.
This is maybe also the reason for the following issue: https://github.com/Graylog2/graylog2-server/issues/7701

We still need to write a migration to remove the `query_string` from already exported searches. 

Fixes: #7732

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/logic/views/ViewTransformer.js
+++ b/graylog2-web-interface/src/views/logic/views/ViewTransformer.js
@@ -30,11 +30,14 @@ const ViewTransformer = (searchView: View): View => {
       .widgets(widgets)
       .build();
   });
+  const newQueries = searchView.search.queries.map(query => query.toBuilder().query({ ...query.query, query_string: '' }).build());
+  const newSearch = searchView.search.toBuilder().queries(newQueries).build();
 
   return searchView.toBuilder()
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/logic/views/ViewTransformer.js
```diff
@@ -30,11 +30,17 @@ const ViewTransformer = (searchView: View): View => {
       .widgets(widgets)
       .build();
   });
+  // Remove query string attached to the existing search query
+  const newQueries = searchView.search.queries.map(
+    query => query.toBuilder().query({ ...query.query, query_string: '' }).build(),
+  );
+  const newSearch = searchView.search.toBuilder().queries(newQueries).build();
```
linuspahl: With these changes the `search -> query` equals the query which gets created, on a dashboard creation. Which means it still has its own timerange. I thought about removing the timerange as well, but it is a required attribute and because every widget has its own timerange, this will not result in any problems.

PR #7743, opened 2020-03-24
Improve widget legend positioning

linuspahl:
There are a few cases where a chart legend is overlapping the x-axis labels.

You can sometimes see this behaviour on the main search page:
image

I had an extensive look at the available plotly layout settings and could not find a perfectly dynamic solution. In the end I've added a small manual calculation, to create more dynamic layout and reduce the problematic cases to a minimum. This will e.g. fix the problem, visible in the above screenshot.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/visualizations/GenericPlot.jsx
+++ b/graylog2-web-interface/src/views/components/visualizations/GenericPlot.jsx
@@ -124,24 +124,28 @@ class GenericPlot extends React.Component<Props, State> {
       autosize: true,
       showlegend: true,
       margin: {
-        autoexpand: true,
         t: 10,
         l: 40,
         r: 10,
-        b: 40,
+        b: 0,
         pad: 0,
       },
       legend: {
         orientation: 'h',
-        y: -0.14,
       },
       hoverlabel: {
         namelength: -1,
       },
+      yaxis: {
+        automargin: true,
+      },
+      xaxis: {
+        automargin: true,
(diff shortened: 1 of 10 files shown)


graylog2-web-interface/src/views/components/visualizations/GenericPlot.jsx
```diff
@@ -124,24 +124,28 @@ class GenericPlot extends React.Component<Props, State> {
...
       },
       hoverlabel: {
         namelength: -1,
       },
+      yaxis: {
+        automargin: true,
+      },
+      xaxis: {
+        automargin: true,
+      },
```
linuspahl: As far as I could see it the `automargin` attribute is currently only helpful when using the Heatmap. I still prefer adding them to the default layout, because they could be helpful for all charts in theory,

graylog2-web-interface/src/views/components/visualizations/GenericPlot.jsx
```diff
@@ -124,24 +124,28 @@ class GenericPlot extends React.Component<Props, State> {
       autosize: true,
       showlegend: true,
       margin: {
-        autoexpand: true,
         t: 10,
         l: 40,
         r: 10,
-        b: 40,
+        b: 0,
```
linuspahl: I removed the margin, to provide more space for the chart. The widget padding already displays enough "margin".

PR #7762, opened 2020-03-26
Use spread object creation to copy configurations before update

### LINKED ISSUE #7760: Updating configuration does not propagate change to overview component
On the configuration page, when you save a configuration update, it is not reflected on the configurations page overview components.

1. Goto to System / Configuration 
2. Click and update a configuration (ex. Sidecar System)
3. Save configuration
4.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

ousmaneo:
- create setConfigurationState and use cloneDeep to update
- Update methods to use setConfigurationState before propagateChanges

## Description

Before this change saving a configuration update on configuration page wasn't updating the overview on the same page because the different configuration were held in the same object inside the configurationStore. Resulting in the the connect function did not see change when a key of that object were updated. 
With these change we make sure to create a new object cloned from the current configuration and and set it back as configuration object after updating.

fixes #7760 
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
+++ b/graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
@@ -1,7 +1,6 @@
 // @flow
 import Reflux from 'reflux';
 import { qualifyUrl } from 'util/URLUtils';
-import { cloneDeep } from 'lodash';
 import fetch from 'logic/rest/FetchProvider';
 import UserNotification from 'util/UserNotification';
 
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/stores/configurations/ConfigurationsStore.js
```diff
@@ -49,7 +56,7 @@ const ConfigurationsStore = Reflux.createStore({
   list(configType) {
     const promise = fetch('GET', this._url(`/${configType}`));
     promise.then((response) => {
-      this.configuration[configType] = response;
+      this.setConfigurationState(configType, response);
```
dennisoelkers: I think we do not deep copying here. What we could do is creating a new object for `configuration` instead of mutating the existing one:

```suggestion
      this.configuration = { ...this.configuration, [configType]: response };
```

ousmaneo: Great ! much cleaner 👍

PR #7782, opened 2020-03-30
Show helpful message on dashboard page for missing stream permission

### LINKED ISSUE #7505: Show helpful message instead of 404 if user has access to view/search but not all streams
When a user opens a search/dashboard that is accessible through a sharing, but the user does not have access to all referenced streams, a 404 is thrown and the party gorilla appears in the web if.

Instead, if view/search are accessible, but not all streams are, a different http response could be returned, which allows the web interface to show a more helpful message to the user.

linuspahl:
As described in https://github.com/Graylog2/graylog2-server/issues/7505, opening a dashboard without permissions for the related streams results in a redirect to the "not found page".

With these changes we are showing a helpful error message which includes the stream ids the user needs permissions to.

Fixes #7505

## Types of changes

## Checklist:


graylog2-web-interface/src/views/pages/ExtendedSearchPage.test.jsx
```diff
@@ -261,4 +262,17 @@ describe('ExtendedSearchPage', () => {
...
+    let wrapper;
+    await act(async () => {
+      wrapper = mount(<SimpleExtendedSearchPage />);
+    });
+
+    if (wrapper) {
+      wrapper.update();
+      expect(wrapper.contains('The error message.')).not.toBeNull();
+    }
+  });
```
linuspahl: I received an `act()` warning because of the update triggered by throwing the error. How do we deal with this case normally when using enzyme? There has to be a more elegant way.

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
```diff
@@ -154,14 +155,22 @@ const ExtendedSearchPage = ({ route, location = { query: {} }, router, searchRef
         .push(SearchActions.refresh.listen(refreshIfNotUndeclared))
         .push(ViewActions.search.completed.listen(refreshIfNotUndeclared));
       return null;
-    });
+    }).catch((error) => setErrors([error]));
```
linuspahl: I wrote the `SearchExectuionErrors` component in a way it can handle all types of errors. initially I've tried to add the `.catch()` to `SearchActions.execute(executionState)` but I prefer catching and informing the user about all possible errors (which are not part of a 403 / 404 request).

I've also tried to add the `.catch()` before the `.finally()`. but this results in some test failures I couldn't fix easily.

graylog2-web-interface/src/logic/rest/FetchProvider.js
```diff
@@ -152,29 +149,29 @@ function queuePromiseIfNotLoggedin(promise) {
...
 }
 
-export function fetchPeriodically(method, url, body) {
+export function fetchPeriodically(method, url, body, handleForbidden, handleUnauthorized) {
   const promise = () => new Builder(method, url)
     .authenticated()
     .noSessionExtension()
-    .json(body)
+    .json(body, handleForbidden, handleUnauthorized)
     .build();
```
linuspahl, line 176: @dennisoelkers A few questions regarding these changes:
Would you add e.g. `logoutIfUnauthorized ` as a Builder method? And would you allow a customisation of the whole part or only the condition check? (in the context of `logoutIfUnauthorized` and `redirectIfForbidden `)

graylog2-web-interface/src/views/pages/ExtendedSearchPage.test.jsx
linuspahl: This comment is no longer relevant, because we've adjusted the error handling.

graylog2-server/src/test/java/org/graylog/plugins/views/search/SearchExecutionGuardTest.java
```diff
@@ -60,7 +61,7 @@ public void setUp() throws Exception {
     public void failsForNonPermittedStreams() {
         final Search search = searchWithStreamIds("ok", "not-ok");
 
-        assertThatExceptionOfType(ForbiddenException.class)
+        assertThatExceptionOfType(MissingStreamPermissionException.class)
```
dennisoelkers, line 64: We should also check if the exception contains the missing streams.

graylog2-web-interface/src/logic/rest/FetchProvider.js
```diff
@@ -54,7 +74,7 @@ export class Builder {
     return this;
   }
 
-  json(body) {
+  json(body, handleForbidden = redirectIfForbidden, handleUnauthorized = logoutIfUnauthorized) {
```
dennisoelkers, line 77: Instead of having to pass `handleForbidden`/`handleUnauthorized` to every method, we should introduce two builder methods for these, which allows consumers to set these. Creating a new `FetchBuilder` would default to `redirectIfForbidden` respectively `logoutIfUnauthorized`.

graylog2-web-interface/src/views/pages/ExtendedSearchPage.jsx
```diff
@@ -86,6 +87,7 @@ const ConnectedFieldList = connect(FieldList, { fieldTypes: FieldTypesStore, vie
   }));
 
 type Props = {
+  errors: Object[],
```
dennisoelkers, line 90: I think this should be either `Array<Error>` or `Array<mixed>`. `Object` is not recommended for use, ever:

> Object is an alias to any and will be deprecated and removed in a future version of Flow.

graylog2-web-interface/src/logic/errors/ReportedErrors.js
```diff
@@ -23,8 +28,20 @@ export const createReactError = (error: $PropertyType<ReactError, 'error'>, info
...
+  type: StreamPermissionErrorType,
+});
+
+export const createFromFetchError = (error: FetchError) => {
+  switch (error.status) {
+    case 403:
+      return error?.additional?.body?.type === 'MissingStreamPermission' ? createStreamPermissionError(error) : createUnauthorizedError(error);
+    default:
+      throw Error(`Provided FetchError is not a valid ReportedError because status code ${error.status} is not supported`);
+  }
```
linuspahl, line 46: This part is also the base to display the error message for 404 api requests, instead of redirecting to `/notfound`.
E.g. for the dashboard and saved search page, we just need to adjust the `ViewLoader`, once we have a custom `ReportedError`.

graylog2-web-interface/src/logic/rest/FetchProvider.js
```diff
@@ -78,15 +80,15 @@ export class Builder {
 
         // Redirect to the start page if a user is logged in but not allowed to access a certain HTTP API.
         if (SessionStore.isLoggedIn() && error.status === 403) {
-          ErrorsActions.report(createUnauthorizedError(error));
+          ErrorsActions.report(createFromFetchError(fetchError));
```
linuspahl, line 83: While having another look, I realised we were not providing a real `FetchError` when creating a `ReportedError`.
I've decided to create a `FetchError` instead of using the `SuperagentPromiseError`, because the checks we are doing when we create a `FetchError` are still useful and it is easier to work with our own interface.

linuspahl: One side note regarding the `FetchProvider`, during the development we've discussed the unification of redundant code. I did not implemented these changes inside this PR, because they are also part of the CSV export PR https://github.com/Graylog2/graylog2-server/pull/7709

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+type Props = {
+  error: FetchError,
+};
+
+const StreamPermissionErrorPage = ({ error }: Props) => {
+  const description = (
+    <>
+      <p>This resource includes streams you do not have permissions to.</p>
+      <p>Please contact your administrator and provide the error details which indlude a list of streams you need access to.</p>
```
dennisoelkers, line 18: Typo: `indlude` -> `include`

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
```diff
@@ -0,0 +1,34 @@
...
+import UnauthorizedErrorPage from './UnauthorizedErrorPage';
+
+type Props = {
+  error: FetchError,
+};
+
+const StreamPermissionErrorPage = ({ error }: Props) => {
+  const description = (
+    <>
+      <p>This resource includes streams you do not have permissions to.</p>
```
dennisoelkers, line 17: Not 100% sure, but I think it should be `you do not have permissions _for_`

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
```diff
@@ -0,0 +1,34 @@
...
+      <p>Please contact your administrator and provide the error details which indlude a list of streams you need access to.</p>
+    </>
+  );
+  const streamIds = error?.additional.body?.streams;
```
dennisoelkers, line 21: Is `additional` always present if `error` is, or do we want to make it optional too?

graylog2-web-interface/src/pages/UnauthorizedErrorPage.jsx
```diff
@@ -8,37 +8,56 @@ import { FetchError } from 'logic/rest/FetchProvider';
...
+  const defaultDescription = (
     <>
       <p>You do not have the required permissions to view this resource.</p>
       <p>Please contact your administrator and provide the error details.</p>
     </>
   );
+  const errorMessageString = createErrorMessageString(errorDetails, pageDetails, errorMessage);
   return (
-    <ErrorPage title="Missing Permissions" description={description}>
+    <ErrorPage title={title} description={description || defaultDescription}>
```
dennisoelkers, line 40: Whenever we do not want the result of a boolean expression, we should now use `??` instead of `||`. (As long as in this case we do not want to support setting the `description` prop to `null` to explicitly disable showing a description)

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
linuspahl: I thought about this as well and found examples for both cases, but having a look at the oxford dictionary it is "permissions for somebody/something" / "permissions to do something".

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
linuspahl: Good catch, the `FetchError` should always have `additional`, but we are handling errors with an unknown format e.g. in the `UauthorizedPage` as well, just in case,  so I added the another check.

graylog2-web-interface/src/pages/StreamPermissionErrorPage.jsx
dennisoelkers: But then, it is `permissions for (streams)`, right?

linuspahl: It is and I've changed it accordingly.

graylog2-web-interface/src/components/errors/ReportedError.jsx
```diff
@@ -4,10 +4,11 @@ import { useState, useEffect } from 'react';
...
 
 import ErrorsActions from 'actions/errors/ErrorsActions';
-import { type ReportedError as ReportedErrorType, ReactErrorType, UnauthorizedErrorType } from 'logic/errors/ReportedErrors';
+import { type ReportedError as ReportedErrorType, ReactErrorType, UnauthorizedErrorType, StreamPermissionErrorType } from 'logic/errors/ReportedErrors';
 
 import RuntimeErrorPage from 'pages/RuntimeErrorPage';
 import UnauthorizedErrorPage from 'pages/UnauthorizedErrorPage';
+import StreamPermissionErrorPage from 'pages/StreamPermissionErrorPage';
 
 const ReportedError = ({ children, router }) => {
```
dennisoelkers, line 13: This component needs a better name, imo. Maybe something like `ReportedErrorBoundary`? It is not a react error boundary, nevertheless that's what it is doing: Rendering its children _or_ an error if present. Renaming the component also removes the need to rename the `ReportedError` type import.

linuspahl: I agree, my only problem was I connect "ErrorBoundary" in React with `componentDidCatch`, but of course it is a more meaningful name.

graylog2-web-interface/src/logic/errors/ReportedErrors.js
```diff
@@ -3,18 +3,23 @@
 import { FetchError } from 'logic/rest/FetchProvider';
 
 export const ReactErrorType = 'ReactError';
+export const NotFoundErrorType = 'NotFoundError';
```
dennisoelkers: This constant is not used, from what I can see. I guess we will introduce it later, but let's remove it for now if it's not used yet.

Issue #7730, opened 2020-03-20
User with read permission on dashboard also sees delete menu option


edmundoa:
A reader user with permissions to read a Dashboard can also find a menu entry to delete that dashboard. This is only a cosmetic issue, since clicking the option does not work.

<img width="1673" alt="Screenshot 2020-03-20 at 17 23 49" src="https://user-images.githubusercontent.com/716185/77184175-ba06be00-6acf-11ea-883c-2531fde2d39f.png">

Permissions for the user:
```
"streams:read:<id>",
"users:tokencreate:reader",
"users:tokenremove:reader",
"dashboards:read:<id>",
"users:passwordchange:reader",
"users:edit:reader",
"users:tokenlist:reader",
"clusterconfigentry:read",
"indexercluster:read",
"messagecount:read",
"journal:read",
"messages:analyze",
"inputs:read",
"metrics:read",
"savedsearches:edit",
"fieldnames:read",
"buffers:read",
"system:read",
"savedsearches:create",
"jvmstats:read",
"decorators:read",
"throughput:read",
"savedsearches:read",
"messages:read"
```

## Expected Behavior
User with read permissions on a Dashboard should not see an option to delete it, or at least that option should be disabled.

## Current Behavior
User with read permissions on a Dashboard can find an option to delete the dashboard.

## Steps to Reproduce (for bugs)
1. Create a user with permissions to read a Dashboard
2. Login with that user
3. Go to the Dashboards page and see that there is a menu option to delete the Dashboard

## Your Environment
* Graylog Version: Graylog 3.2.5-SNAPSHOT, Graylog 3.3.0-SNAPSHOT

PR #7793, opened 2020-03-31
Only show dashboard delete button for users with correct permissions

linuspahl:
**We need to merge https://github.com/Graylog2/graylog2-server/pull/7817 first, it changes the backend permissions for view deletions. Otherwise the frontend permission check will not match.**

As described in https://github.com/Graylog2/graylog2-server/issues/7730 a user without the permissions to delete a dashboard still sees the `Delete` action in the dashboard action menu.

Fixes: https://github.com/Graylog2/graylog2-server/issues/7730

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/views/ViewList.jsx
+++ b/graylog2-web-interface/src/views/components/views/ViewList.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import createReactClass from 'create-react-class';
 import { ButtonToolbar, DropdownButton, MenuItem } from 'components/graylog';
+import IfPermitted from 'components/common/IfPermitted';
 
 import { PaginatedList, SearchForm, Spinner, EntityList } from 'components/common';
 import View from './View';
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/views/ViewList.jsx
```diff
@@ -1,126 +1,110 @@
...
+
+  const onViewDelete = (view) => {
     return () => {
-      this.props.handleViewDelete(view).then(() => {
-        this.setState({ page: 1 }, () => {
-          this.execSearch();
-        });
+      handleViewDelete(view).then(() => {
+        setPage(1);
+        execSearch();
```
linuspahl: We need to execute the search, when `page`, `perPage` or `query` changes, this works perfectly in combination with `useEffect`, except for this case (deleting a view). As you can see in the diff, we are running `execSearch ` after `setState`.  With the current implementation, `setPage(1)` will not trigger `executeSearch`, when the current page is 1.

The above ` execSearch();` in line 55 has no effect. also `setPage(1).then()` is not a possible solution.

Do you have any idea how to deal with this case?

PR #7821, opened 2020-04-02
Migrating InputStatusService

waab76:
Migrating the InputStatusService from the enterprise plugins repo to the main repo.

The InputStatusService was created to support Inputs that need to persist state for the duration of their lifecycle.  It was created to allow the Okta input to checkpoint on seen messages in order to avoid duplicate message processing when the input is restarted.

Original PR: https://github.com/Graylog2/graylog-plugin-enterprise-integrations/issues/136

## Description
Migrating the InputStatusService from the enterprise plugins repo to the main repo.

The InputStatusService was created to support Inputs that need to persist state for the duration of their lifecycle.  It stores Input state in the new `input_status` MongoDB collection. It was created to allow the Okta input to checkpoint on seen messages in order to avoid duplicate message processing when the input is restarted.

## Motivation and Context
Migration to the main graylog2-server repo is needed to allow non-enterprise Inputs to use this service

## How Has This Been Tested?
This feature has been tested locally by multiple Graylog developers.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/persistence/InputStateData.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/persistence/InputStateData.java
@@ -1,0 +1,71 @@
+package org.graylog2.inputs.persistence;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.google.common.collect.Maps;
+
+import java.util.Map;
+import java.util.Objects;
+
+@JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.EXISTING_PROPERTY,
+        property = InputStateData.TYPE_FIELD,
+        visible = true,
+        defaultImpl = InputStateData.Fallback.class
+)
+public interface InputStateData {
+    String TYPE_FIELD = "type";
+
+    @JsonProperty
+    String type();
+
+    public interface Builder<SELF> {
+        @JsonProperty(TYPE_FIELD)
+        SELF type(String type);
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/persistence/InputStatusService.java
```diff
@@ -0,0 +1,93 @@
...
+    private final JacksonDBCollection<InputStatusRecord, ObjectId> statusCollection;
+
+    @Inject
+    public InputStatusService(MongoConnection mongoConnection,
+                              MongoJackObjectMapperProvider objectMapperProvider,
+                              EventBus eventBus) {
+        DB mongoDatabase = mongoConnection.getDatabase();
+        DBCollection collection = mongoDatabase.getCollection(INPUT_STATUS_COLLECTION);
+
+        eventBus.register(this);
```
bernd, line 38: I think we should bind this service in a singleton scope (`bind(InputStatusService.class).asEagerSingleton()` and add `@Singleton` annotation to this class for documentation purposes) in the guice bindings to avoid multiple instances of this class. The problem is that every instance would register with the `EventBus` and since we don't have lifecycle support for this object where we could call `unregister` on shutdown, we would basically leak references into the event bus. We have that issue in other services as well and mostly solve it by binding the service as a singleton.

graylog2-server/src/main/java/org/graylog2/inputs/persistence/InputStateData.java
```diff
@@ -0,0 +1,71 @@
...
+        private String type;
+
+        private Map<String, Object> props = Maps.newHashMap();
+
+        @Override
+        public String type() {
+            return type;
+        }
+
+        @JsonAnySetter
```
bernd, line 43: Great idea using any-getter/setter! :+1:

graylog2-server/src/main/java/org/graylog2/inputs/persistence/InputStatusService.java
waab76: Will do

waab76:
License added to 4 files.

PR #7829, opened 2020-04-03
Improve triggering search executions from search bar.

### LINKED ISSUE #7572: Auto query / Update Search Results is too aggressive and queries at inopportune times.
The auto query to refresh the results triggers at inopportune times while building a query. This results in unnecessary user waits and excessive load on the graylog/elasticsearch clusters while doing large queries.

## Expected Behavior
When doing multiple filter settings in a query, there should be no background search until the user indicates it is time to execute a query. 

## Current Behavior
Start a query, then change to absolute time. Selecting a "start" time triggers an auto-query, before you're even able to set a "stop" time. 

## Possible Solution
Set a user preference option to disable auto query execution. That way if you're doing a lot of banal queries and you want this feature enabled, you can use it. Those of us who want to set multiple query filters before the data is "refreshed" can do so by disabling this feature.  

1. Try to rapidly change multiple aspects of any query. Especially bad with time window selection.
searchannoyance

### LINKED ISSUE #7596: Absolute time range input should not auto refresh search upon blur
As the absolute time range input consists of more than one input, it should not trigger a search refresh when blurring a single input. Instead, it should do when the form is submitted by pressing enter in one of the inputs or when another input (stream selection, query input) triggers a refresh.

Refs #7403.
Refs #7572.

### LINKED ISSUE #7724: Search URL query params are not being updated correctly when streams get removed quickly.
Selecting and removing a stream on a search page will update the URL query parameters.
Doing this very fast while e.g. the search is loading can result in an inconsistency between the URL query parameters and the actually selected streams

1. Open the search page
2. Select a few streams
3. Remove them very fast

## Context
* Graylog Version: 3.2.4

dennisoelkers:
## Motivation and Context

This PR is incorporating user feedback into the product to improve the behaviour of the search controls (time range, streams, query string) for the search & dashboard pages as well was when editing a widget on a dashboard.

The changes it introduces are:

  * Changing the time range type (relative, absolute, keyword) does not
    immediately trigger a search refresh
  * Changing one of the values for the absolute time range (from, to)
    does not trigger a search refresh, unless the user presses enter
  * Changing the keyword value does not trigger a search refresh (upon
    change or blur) unless the user presses enter
  * If validation for one of the fields fails (namely invalid timestamp
    for absolute from/to or invalid keyword), the search button is disabled
  * Blurring the query string input does not trigger a search refresh
    (pressing enter inside still does)
  * Changing the set of streams does not trigger a search refresh
  * When the user has changed a value and not submitted a search (the
    time range/query string), a yellow bubble will show up at the search
    button, indicating that there are uncommitted changes

There are couple of specialties in this PR:
  - To simplify form handling, the [formik](https://jaredpalmer.com/formik/docs/overview) library is introduced, helping a lot to make creation of a proper form with consistent validation, dirty flag & submission way easier.
  - This PR is updating `ace-builds` to v1.4.9 through a version resolution in `package.json`. The update is needed because of an issue (ajaxorg/ace-builds#171) in v1.4.8.
  - The change to `package.json` which modifies the `moduleNameMapper` section is needed, because of the import of `ace-builds/webpack-resolver`, which utilizes the `file-loader` to import themes/modes, but uses a file-loader setting that was previously not mached in jest's module name mapper which translates these statements into simple imports.

Fixes #7572.
Fixes #7596.
Fixes #7724.

## Types of changes

## Checklist:


alex-konn:
Very nice! :+1: 

I have played around with this a bit and find that it works really well. Much nicer to quickly edit the paramers and then hit enter than always having to wait for the automiatic refresh to finish.

The only thing that bugs me a bit is the exception for relative time ranges. If I change from another range type to relative, the button is marked dirty and not executed. Changing the range length when the type was already relative still auto executes the search. 

I would find it more consistent, if relative ranges would behave like all the other inputs and just marked the button as dirty.

graylog2-web-interface/src/views/components/searchbar/TimeRangeInput.jsx
```diff
@@ -1,31 +1,50 @@
...
+  },
+  keyword: {
+    component: KeywordTimeRangeSelector,
+  },
+};
+
+const timerangeStrategy = (type) => {
+  if (type === undefined) {
+    return { component: DisabledTimeRangeSelector };
   }
```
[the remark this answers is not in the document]
linuspahl: We can type the param `type` here as well.

graylog2-web-interface/src/views/components/TimerangeForForm.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export const migrateTimeRangeToNewType = (oldTimerange: ?TimeRange, type: string): ?TimeRange => {
+  const oldType = oldTimerange ? oldTimerange.type : 'disabled';
+
+  if (type === oldType) {
+    return oldTimerange;
+  }
+
+  return migrationStrategies[type](oldTimerange);
```
linuspahl, line 65: We checking if the provided type exists in comparable cases, let's check it here as well.

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
...
+
+  ::after {
+    position: absolute;
+    content: '';
+    height: 16px;
+    width: 16px;
+    top: -5px;
+    right: -6px;
+    border-radius: 50%;
+    background-color: #ffc107;
```
linuspahl, line 22: Something for a follow up PR: We should add this color to our theme.

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
...
+const SearchButton = ({ running, disabled, glyph, dirty }: Props) => {
+  const ButtonComponent = dirty ? DirtyButton : Button;
+  const title = dirty ? 'Perform search (changes were made after last search execution)' : 'Perform search';
+  return (
+    <ButtonComponent type="submit"
+                     bsStyle={running ? 'warning' : 'success'}
+                     disabled={disabled}
+                     title={title}
+                     className="pull-left search-button-execute">
+      {running ? <Spinner fixedWidth pulse text="" /> : <Icon name={glyph} />}
```
linuspahl, line 42: Not related to this PR, but we are never defining the `running` prop. Are we not using the button loading state, because of to the global loading indicator?

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
...
 import { Button } from 'components/graylog';
 import { Spinner, Icon } from 'components/common';
+import type { ThemeInterface } from 'theme/index';
 
-const SearchButton = ({ running, disabled, glyph }) => (
-  <Button type="submit" bsStyle={running ? 'warning' : 'success'} disabled={disabled} className="pull-left search-button-execute">
-    {running ? <Spinner fixedWidth pulse text="" /> : <Icon name={glyph} /> }
-  </Button>
-);
+const DirtyButton: StyledComponent<{}, ThemeInterface, any> = styled(Button)`
```
linuspahl, line 11: There is no immediate benefit, but `any` can be replaces with `HTMLButtonElement`

dennisoelkers: Actually there is, because if we make `wrappedTestingLibrary` return a properly typed function, we will need explicit element types in tests.

graylog2-web-interface/src/views/components/TimerangeForForm.js
dennisoelkers: ✅

graylog2-web-interface/src/views/components/searchbar/TimeRangeOverrideTypeSelector.jsx
```diff
@@ -1,42 +1,48 @@
...
+          Relative
+        </MenuItem>
+        <MenuItem eventKey="absolute"
+                  active={type === 'absolute'}>
+          Absolute
+        </MenuItem>
+        <MenuItem eventKey="keyword"
+                  active={type === 'keyword'}>
+          Keyword
+        </MenuItem>
```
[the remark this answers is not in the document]
dennisoelkers: ✅

graylog2-web-interface/src/views/components/searchbar/TimeRangeOverrideTypeSelector.jsx
```diff
@@ -1,42 +1,48 @@
...
+      </DropdownButton>
+    </ButtonToolbar>
+  );
 };
 
-TimeRangeOverrideTypeSelector.defaultProps = {
-  value: undefined,
-};
+TimeRangeOverrideTypeSelector.propTypes = {};
```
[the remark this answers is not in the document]
dennisoelkers: ✅

graylog2-web-interface/src/views/components/searchbar/TimeRangeInput.jsx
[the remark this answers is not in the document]
dennisoelkers: ✅

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
dennisoelkers: We never really used it, so I removed it.

Issue #7773, opened 2020-03-27
Sharing dialog for Saved Searches refers to dashboards


alex-konn:
...instead of Saved Searches.

![Screenshot from 2020-03-27 16-13-51](https://user-images.githubusercontent.com/45592524/77772964-3c9bfa00-7049-11ea-855c-79c0dcd37d56.png)

* Graylog Version: 3.3.0-SNAPSHOT

PR #7844, opened 2020-04-07
Show view type specific description for dashboard and saved searches share modal.

linuspahl:
As described in https://github.com/Graylog2/graylog2-server/issues/7773 the saved search modal displays a dashboard specific description. This PR adds `typeLabel` method for the view class, which can be used wherever we need to display the name of the view type.

Fixes: https://github.com/Graylog2/graylog2-server/issues/7773

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/views/ShareViewModal.jsx
+++ b/graylog2-web-interface/src/views/components/views/ShareViewModal.jsx
@@ -6,6 +6,7 @@ import { FormGroup, HelpBlock, Radio } from 'components/graylog';
 import BootstrapModalConfirm from 'components/bootstrap/BootstrapModalConfirm';
 import Select from 'views/components/Select';
 import Spinner from 'components/common/Spinner';
+import ViewTypeLabel from 'views/components/ViewTypeLabel';
 import StoreProvider from 'injection/StoreProvider';
 import connect from 'stores/connect';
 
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/views/components/views/ShareViewModal.jsx
```diff
@@ -135,7 +138,7 @@ class ShareViewModal extends React.Component<Props, State> {
           Any user of this Graylog
         </Radio>{' '}
         <Additional>
-          <HelpBlock>Anyone with an account can access the dashboard.</HelpBlock>
+          <HelpBlock>Anyone with an account can access the {view.typeLabel}.</HelpBlock>
```
dennisoelkers, line 141: We could also replace this with a `<ViewTypeLabel type={view.type} />` that renders the view type appropriately. This would also remove the need for a getter in the data class (and solve a presentation issue in the presentation layer).

linuspahl: I like the idea and implemented the approach. What bothers me a little, the component is just a function which returns a string. The main reason I am using the `<ViewTypeLabel type={view.type} />` syntax is the import from `views/components`. While the syntax is useful in most cases a string would be easier to handle in this case, e.g. for the modal title.

PR #7851, opened 2020-04-08
Fasterizing the Button render

kyleknighted:
## Description

- This memoizes the results of the button style generation so that we cache the results of the different variants based off of `bsStyle`

- Removes the use `styled-theming` as we phase that out for just swapping a color object in the ThemeProvider

## Motivation and Context

Some pages with lots of `Button` components were taking a long time to render due to the complex nature of the rendered buttons. Removed the dynamic and complex `styled-theming` and reduced the CSS builder down to just the few variants we need.

## Types of changes

## Checklist:


kyleknighted:
> when will the cache be expired

What do we need to expire the cache of the colors for? Can you provide an example?

> how to check for equality of the arguments so we can match them to a result

What do you need to check? There are tests for contrastingColors, I can create tests for the others. 

>  so I always consider it as a last resort.

Then why was it literally your first suggestion?

> we could also calculate those once in the theme provider and include a map variant -> style in the provided theme

So we should pass all the styles for Buttons, Alerts, Badges, Labels, ListGroupItems, Panels, ProgressBars, and probably more... to every single component from the Provider? Doesn't make sense to me.

kyleknighted:
I created https://github.com/Graylog2/graylog2-server/commit/e2170a563b81a9e38e38c6d0a6fb1ed50df692be that moves the button styles to the ThemeProvider and passes it as part of the context.
It is not any faster, welcome to feedback

kyleknighted:
Paired with @dennisoelkers and think we got to a good solution - ready for re-review

PR #7855, opened 2020-04-09
String Interpolation for Pipeline Rules

rmanibus:
Implements string  interpolation for pipeline rules
## Motivation and Context

avoid formulation like concat( "str1" , concat( "str2", ...) in pipeline rules

https://github.com/Graylog2/graylog-plugin-pipeline-processor/issues/95

## How Has This Been Tested?

Added a unit test which execute the following pipeline rule:
````
rule "stringConcat"
when true
then
    let result = "aa" + "bb" + "cc";
    set_field("result", result);
end
````
Check that the field result is equal to "aabbcc"

## Types of changes

## Checklist:

documentation: 

https://github.com/Graylog2/documentation/pull/851

### CODE REVIEW

**Comment by rmanibus** [comment]
Hi, Do you have some guidelines about running a test environment ?

**Comment by rmanibus** [comment]
Tests seems to be broken in master.
Also should I recreate an issue in this project to reference this ?

**Comment by rmanibus** [comment]
Fixes #7863

File: graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/ast/expressions/AdditionExpression.java [hunk 1]
```diff
@@ -92,6 +91,13 @@ public Object evaluateUnsafe(EvaluationContext context) {
             } else {
                 return new Duration(right, left);
             }
+        }else if(String.class.equals(type) ){
```
kroepke, line 94: Please format whitespace here according to the rest of the file (space before and after keywords, between `)` and `{` and no inner space in `()` expressions.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/ast/expressions/AdditionExpression.java
```diff
@@ -92,6 +91,13 @@ public Object evaluateUnsafe(EvaluationContext context) {
...
+            if(!isPlus){
+                return null;
+            }
+            final String left = (String) leftValue;
```
kroepke, line 98: Both this and the next line warn about `Casting 'leftValue' to 'String' may produce 'ClassCastException'`.
While technically the code cannot throw that exception (`leftValue` is a string because we checked and `rightValue` has to be a string, too, because of the parser/typechecker checks, we try to avoid warnings in code when possible.

My suggestion would be to rewrite this as `String.valueOf(leftValue)` to be future proof.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/ast/expressions/AdditionExpression.java
```diff
@@ -92,6 +91,13 @@ public Object evaluateUnsafe(EvaluationContext context) {
...
+            final String right = (String) rightValue;
```
kroepke, line 99: same as above

graylog2-server/src/test/java/org/graylog/plugins/pipelineprocessor/functions/FunctionsSnippetsTest.java
```diff
@@ -362,6 +362,15 @@ public static void registerFunctions() {
         functionRegistry = new FunctionRegistry(functions);
     }
 
+    @Test
+    public void string_concat(){
```
kroepke, line 366: Please change the name to `stringConcat` to look the same as the other tests.
That also requires the file `string_concat.txt` to be `stringConcat.txt` below.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/ast/expressions/AdditionExpression.java
```diff
@@ -92,6 +91,13 @@ public Object evaluateUnsafe(EvaluationContext context) {
             } else {
                 return new Duration(right, left);
             }
+        }else if(String.class.equals(type) ){
+            if(!isPlus){
+                return null;
```
kroepke, line 96: That is the correct return value, but I think the type checker should raise an error to the user when it sees this.

The corresponding code is at https://github.com/Graylog2/graylog2-server/blob/eb392d9180e552723a1b9805c6a06f270a16a18d/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/parser/PipelineRuleParser.java#L777-L780

In that if/else I think it should have a check for `leftType`/`rightType` being `String`, and then in an extra if check for `!isPlus()` which then raises a properly described `InvalidOperation` error. That way the user doesn't get surprised at runtime.

rmanibus: Thanks I missed that, I think It's ok now !

PR #7861, opened 2020-04-10
Provide global context for current user preferences.

dennisoelkers:
## Motivation and Context

This PR is providing a context containing user preferences (right now
untyped key/value pairs) that can be consumed by react components. A
default provider taking preferences from the currently logged in user
was created and employed in the `LoggedInPage` component so it is
available everywhere after a user's login.

In addition, the `PreferencesStore` is turned into using proper Reflux
actions and the `CurrentUserStore` listens to the action to save
preferences in order to reload its state after it has completed. This
makes changing preferences of a user effective in the `CurrentUserStore`
immediately.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/contexts/CurrentUserPreferencesProvider.jsx
+++ b/graylog2-web-interface/src/contexts/CurrentUserPreferencesProvider.jsx
@@ -1,0 +1,27 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { get } from 'lodash';
+
+import { useStore } from 'stores/connect';
+import CombinedProvider from 'injection/CombinedProvider';
+import UserPreferencesContext from './UserPreferencesContext';
+
+const { CurrentUserStore } = CombinedProvider.get('CurrentUser');
+
+const CurrentUserPreferencesProvider = ({ children }: { children: React.Node }) => {
+  const preferences = useStore(CurrentUserStore, (state) => get(state, 'currentUser.preferences'));
+  return preferences
+    ? (
+      <UserPreferencesContext.Provider value={preferences}>
+        {children}
+      </UserPreferencesContext.Provider>
+    )
+    : children;
+};
+
+CurrentUserPreferencesProvider.propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default CurrentUserPreferencesProvider;
--- a/graylog2-web-interface/src/pages/LoggedInPage.jsx
+++ b/graylog2-web-interface/src/pages/LoggedInPage.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import AppRouter from 'routing/AppRouter';
+import CurrentUserPreferencesProvider from '../contexts/CurrentUserPreferencesProvider';
 
-class LoggedInPage extends React.Component {
-  render() {
-    return <AppRouter />;
-  }
-}
+const LoggedInPage = () => (
+  <CurrentUserPreferencesProvider>
+    <AppRouter />
+  </CurrentUserPreferencesProvider>
+);
 
 export default LoggedInPage;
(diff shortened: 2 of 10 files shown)


graylog2-web-interface/src/contexts/CurrentUserPreferencesProvider.jsx
```diff
@@ -0,0 +1,27 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import { get } from 'lodash';
+
+import { useStore } from 'stores/connect';
+import CombinedProvider from 'injection/CombinedProvider';
+import UserPreferencesContext from './UserPreferencesContext';
+
+const { CurrentUserStore } = CombinedProvider.get('CurrentUser');
```
linuspahl, line 10: What is the reason to use `CombinedProvider.get(` here? Compared with `import CurrentUserStore from 'stores/users/CurrentUserStore';` the only difference is that using the `CombinedProvider`, `preferences` is not `undefined` initially.

graylog2-web-interface/src/pages/LoggedInPage.jsx
```diff
@@ -1,10 +1,11 @@
...
-class LoggedInPage extends React.Component {
-  render() {
-    return <AppRouter />;
-  }
-}
+const LoggedInPage = () => (
+  <CurrentUserPreferencesProvider>
+    <AppRouter />
+  </CurrentUserPreferencesProvider>
+);
```
linuspahl, line 9: One thought regarding the placement, which does not require a change right now: At some point it will be very useful to wrap the `<GraylogThemeProvider>` with the `<CurrentUserPreferencesProvider>` to show a specific theme based on the user settings. But we can still adjust this when we have the need.

graylog2-web-interface/src/stores/users/CurrentUserStore.jsx
```diff
@@ -1,13 +1,14 @@
...
-import URLUtils from 'util/URLUtils';
+import { qualifyUrl } from 'util/URLUtils';
 import fetch from 'logic/rest/FetchProvider';
 import ApiRoutes from 'routing/ApiRoutes';
 
 import CombinedProvider from 'injection/CombinedProvider';
 
 const { SessionStore, SessionActions } = CombinedProvider.get('Session');
 const { StartpageStore } = CombinedProvider.get('Startpage');
+const { PreferencesActions } = CombinedProvider.get('Preferences');
```
linuspahl, line 11: Out of interest, why are you using the `CombinedProvider` here as well? I thought it is the recommended to use the direct import, when possible.

graylog2-web-interface/src/pages/LoggedInPage.jsx
dennisoelkers: Good point!

graylog2-web-interface/src/stores/users/CurrentUserStore.jsx
dennisoelkers: It is recommended (and factually the only way possible) for stores/actions which are not exported using `singleton`. For all other stores we need to go through the `CombinedProvider` to force core and plugins getting the same instance of the store (and actions).

graylog2-web-interface/test/helpers/mocking/index.js
```diff
@@ -1,4 +1,7 @@
 export { default as CombinedProviderMock } from './CombinedProviderMock';
+export { default as MockCombinedProvider } from './CombinedProviderMock';
 export { default as StoreMock } from './StoreMock';
+export { default as MockStore } from './StoreMock';
 export { default as StoreProviderMock } from './StoreProviderMock';
+export { default as MockStoreProvider } from './StoreProviderMock';
 export { default as asMock } from './AsMock';
```
linuspahl, line 7: I guess it is a possible improvement for a follow up PR to replace the usage of `StoreProviderMock` and `StoreMock` in all our tests?

dennisoelkers: Yes, that was the plan. Good catch!

PR #7867, opened 2020-04-13
Removing n+1's in `StreamService#loadAll` by consolidating external loads

dennisoelkers:
## Motivation and Context

Before this change, calling `StreamService#loadAll` triggered loading of associated outputs and index sets, _both_ resulting in n+1 each.

This change is now placing fetching of outputs and index sets outside the loop iterating over each stream, leaving only simple lookups to be done inside the loop, effectively turning this into a call with a _constant_ number of mongodb queries.

Simple tests on my setup showed an improvement of ~270ms -> ~110ms.  While this is already impressive, I expect a much bigger improvement for setups with a high number of outputs.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamServiceImpl.java
@@ -236,8 +236,8 @@ public class StreamServiceImpl extends PersistedServiceImpl implements StreamSer
         return objectIds == null ? Collections.emptyList() : objectIds;
     }
 
-    private Map<String, IndexSet> indexSetsForStreams(List<DBObject> streamsQuery) {
-        final Set<String> indexSetIds = streamsQuery.stream()
+    private Map<String, IndexSet> indexSetsForStreams(List<DBObject> streams) {
+        final Set<String> indexSetIds = streams.stream()
                 .map(stream -> (String)stream.get(StreamImpl.FIELD_INDEX_SET_ID))
                 .filter(s -> !isNullOrEmpty(s))
                 .collect(Collectors.toSet());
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/streams/StreamServiceImpl.java
```diff
@@ -194,23 +195,57 @@ public Stream load(String id) throws NotFoundException {
...
+    private List<ObjectId> outputIdsForRawStream(DBObject o) {
+        final List<ObjectId> objectIds = (List<ObjectId>) o.get(StreamImpl.FIELD_OUTPUTS);
+        return objectIds == null ? Collections.emptyList() : objectIds;
+    }
+
+    private Map<String, IndexSet> indexSetsForStreams(List<DBObject> streamsQuery) {
+        final Set<String> indexSetIds = streamsQuery.stream()
+                .map(stream -> (String)stream.get(StreamImpl.FIELD_INDEX_SET_ID))
+                .filter(s -> !isNullOrEmpty(s))
+                .collect(Collectors.toSet());
```
mpfz0r, line 243: Maybe we shouldn't call this `streamsQuery`? 

```suggestion
    private Map<String, IndexSet> indexSetsForStreams(List<DBObject> streams) {
        final Set<String> indexSetIds = streams.stream()
                .map(stream -> (String)stream.get(StreamImpl.FIELD_INDEX_SET_ID))
                .filter(s -> !isNullOrEmpty(s))
                .collect(Collectors.toSet());
```

dennisoelkers: Absolutely! ✅

PR #7885, opened 2020-04-15
Enable/disable auto-completion in query input based on user preference.

### LINKED ISSUE #7809: Disabling enableSmartSearch doesn't disable fields autocompletion
## Description
When I define `enableSmartSearch=false` under user preferences, Graylog doesn’t seem to react and continues to show me fields autocompletion suggestions under the search query box and other boxes with autocompletion feature.

## Expected Behavior
Graylog stops showing fields autocompletion suggestions under the search query box.

## Current Behavior
Graylog keeps showing fields autocompletion suggestions under the search query box.

1. Disable `enableSmartSearch` under users preferences
2. Navigate back to the search menu, start writing search query

## Context
We are experiencing performance issues similar to https://github.com/Graylog2/graylog2-server/issues/7753. There are many unnecessary queries generated for getting those couple thousand fields. We are living with this problem for quite some weeks now, getting the same result using different browsers under users with different privileges.

## Your Environment

* Graylog Version: Graylog 3.2.4+a407287 under Docker
* Elasticsearch Version: 6.8.7
* MongoDB Version: 4.0.16
* Operating System: Linux/Ubuntu 19.10
* Browser version: Chromium 80.0.3987.149, Firefox 74.0

dennisoelkers:
## Motivation and Context

**Note: This needs #7829 to be merged before.**

This PR is adding back functionality which enables/disables the field
name autocompletion in the query input on the search/dashboards pages if
the user has a user preference for `enableSmartSearch` set which is
`false`.

Fixes #7809.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/SearchBar.jsx
+++ b/graylog2-web-interface/src/views/components/SearchBar.jsx
@@ -47,7 +47,12 @@ const defaultOnSubmit = ({ timerange, streams, queryString }, currentQuery: Quer
   return QueriesActions.update(newQuery.id, newQuery);
 };
 
-const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = false, queryFilters, onSubmit = defaultOnSubmit }: Props) => {
+const defaultProps = {
+  disableSearch: false,
+  onSubmit: defaultOnSubmit,
+};
+
+const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = defaultProps.disableSearch, queryFilters, onSubmit = defaultProps.onSubmit }: Props) => {
   if (!currentQuery || !config) {
     return <Spinner />;
   }
@@ -131,10 +136,7 @@ SearchBar.propTypes = {
   onSubmit: PropTypes.func,
 };
 
-SearchBar.defaultProps = {
-  disableSearch: false,
-  onSubmit: defaultOnSubmit,
-};
+SearchBar.defaultProps = defaultProps;
 
 export default connect(
   SearchBar,
--- a/graylog2-web-interface/src/views/components/__snapshots__/WidgetQueryControls.test.jsx.snap
+++ b/graylog2-web-interface/src/views/components/__snapshots__/WidgetQueryControls.test.jsx.snap
@@ -21,21 +21,19 @@ exports[`WidgetQueryControls should do something 1`] = `
             <button
               aria-expanded="false"
               aria-haspopup="true"
-              class="DropdownButton__StyledDropdownButton-sc-1343dcx-0 ZNXPm dropdown-toggle btn btn-info"
-              id="timerange-type"
+              class="DropdownButton__StyledDropdownButton-sc-1343dcx-0 hNEGNs dropdown-toggle btn btn-info"
               role="button"
               type="button"
             >
               <svg
                 class="svg-inline--fa fa-clock"
               />
-               
+
               <span
                 class="caret"
               />
             </button>
             <ul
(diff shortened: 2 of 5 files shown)


graylog2-web-interface/src/views/components/SearchBar.jsx
```diff
@@ -21,90 +23,100 @@ import HorizontalSpacer from 'views/components/horizontalspacer/HorizontalSpacer
...
+    .timerange(timerange)
+    .filter(filtersForQuery(streams))
+    .query(createElasticsearchQueryString(queryString))
+    .build();
+
+  return QueriesActions.update(newQuery.id, newQuery);
 };
 
-const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = false, onExecute: performSearch, queryFilters }: Props) => {
+const SearchBar = ({ availableStreams, config, currentQuery, disableSearch = false, queryFilters, onSubmit = defaultOnSubmit }: Props) => {
```
kyleknighted, line 50: Should probably keep these to the `defaultProps` instead of potentially resetting what the default is here.

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
...
 import { Button } from 'components/graylog';
 import { Spinner, Icon } from 'components/common';
+import type { ThemeInterface } from 'theme/index';
 
-const SearchButton = ({ running, disabled, glyph }) => (
-  <Button type="submit" bsStyle={running ? 'warning' : 'success'} disabled={disabled} className="pull-left search-button-execute">
-    {running ? <Spinner fixedWidth pulse text="" /> : <Icon name={glyph} /> }
-  </Button>
-);
+const DirtyButton: StyledComponent<{}, ThemeInterface, any> = styled(Button)`
```
kyleknighted, line 11: Since you're not using the Theme here, you can just pass `void` instead of `ThemeInterface`

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
+// @flow strict
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import type { StyledComponent } from 'styled-components';
```
kyleknighted, line 5:
```suggestion
import styled, { type StyledComponent } from 'styled-components';
```

graylog2-web-interface/src/views/components/searchbar/SearchButton.jsx
```diff
@@ -1,24 +1,60 @@
...
 
 import { Button } from 'components/graylog';
 import { Spinner, Icon } from 'components/common';
+import type { ThemeInterface } from 'theme/index';
```
kyleknighted, line 9:
```suggestion
```

graylog2-web-interface/src/views/components/SearchBar.jsx
dennisoelkers: The benefit of using default values is that flow is doing proper type checks, while the `PropTypes` type specs only do check very basic things (in this case if `onSubmit` is a function or not). Otoh, having runtime checks potentially helps us debugging issues caused by untested integration code.

The remaining risk is that someone could change one, but not the other, so they deviate. To address this, I created a `defaultProps` object which we are referencing in both places. I am not sure if it improves things or makes it more verbose.

PR #7898, opened 2020-04-16
Add Jenkinsfile for new Jenkins server

juju2112:
Add logic for a Jenkins pipeline.

## Description
Add a Jenkinsfile for CI pipelines. 

## Motivation and Context
Will be able to see fancy stage view UI in Jenkins and take advantage of Pipeline scripting in builds. 

## How Has This Been Tested?
The build runs and succeeds in the new Jenkins environment. We will also leave the old Jenkins build enabled this week so builds can be compared between the two environments. 

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [X ] Refactoring (non-breaking change)

## Checklist:


juju2112:
Please do not merge this yet. Would like to continue testing builds.

graylog2-server/src/main/java/org/graylog2/plugin/indexer/searches/timeranges/AbsoluteRange.java
```diff
@@ -115,14 +114,13 @@ private DateTime parseDateTime(String s) {
                 throw new IllegalArgumentException("Null or empty string");
             }
 
-            final DateTimeFormatter formatter;
```
bernd, line 118: I think this change to `AbsoluteRange.java` should not be in there. We need to do some more careful research & testing to ensure we can do this.

juju2112: You're right, I forgot to remove that. It's reverted now.

PR #7902, opened 2020-04-17
Add option to enable/disable event definitions

### LINKED ISSUE #7497: Add Disable/Stop Button on Event Definition
## What?
Add a button to disable a Event Definition

## Why?
1. First of all if a event is about to happen and the admin wants to switch of the alert to prevent emails beeing send out for no reason. So he does not have to delete the alter but can instead just pause it.
2. For debugging reasons so you can pause a event definition to easier find the culprit.

## Your Environment
* Graylog Version: 3.2.2 (dev to 3.3.0)

bernd:
This PR introduces the possibility to enable and disable event definitions. This can be done by using the actions menu for an event definition in the overview or by editing an event definition and using the "Enable scheduling" checkbox.

This PR also adds some detailed scheduling information for each event definition in the overview.

Fixes #7497


bernd:
> The info about whether an event definition is enabled should also be visible on the configuration overview page.

@mpfz0r Good point. Done!

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.inject.Inject;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class EventDefinitionContextService {
```
mpfz0r, line 41: A short javadoc about what this class does would be nice

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
```diff
@@ -0,0 +1,146 @@
...
+    private final DBJobDefinitionService jobDefinitionService;
+    private final DBJobTriggerService jobTriggerService;
+
+    @Inject
+    public EventDefinitionContextService(DBJobDefinitionService jobDefinitionService, DBJobTriggerService jobTriggerService) {
+        this.jobDefinitionService = jobDefinitionService;
+        this.jobTriggerService = jobTriggerService;
+    }
+
+    public ImmutableMap<String, Object> contextFor(List<EventDefinitionDto> eventDefinitions) {
```
mpfz0r, line 51: Just an idea. If we use Serializable instead, it would be clearer that this is meant to
be sent out as a response..

```suggestion
    public ImmutableMap<String, Serializable> contextFor(List<EventDefinitionDto> eventDefinitions) {
```

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public ImmutableMap<String, Object> contextFor(EventDefinitionDto eventDefinition) {
+        final ImmutableMap<String, SchedulerCtx> schedulerContext = schedulerContext(Collections.singletonList(eventDefinition));
+        return ImmutableMap.of("scheduler", schedulerContext.get(eventDefinition.id()));
+    }
+
+    private Map<String, List<JobDefinitionDto>> getJobDefinitions(List<EventDefinitionDto> eventDefinitions) {
+        final Set<String> eventDefinitionIds = eventDefinitions.stream().map(EventDefinitionDto::id).collect(Collectors.toSet());
+        return jobDefinitionService.getAllByConfigField(EventProcessorExecutionJob.Config.FIELD_EVENT_DEFINITION_ID, eventDefinitionIds);
```
mpfz0r, line 62: If we want to be extra correct, we would also query for the type being `EventProcessorExecutionJob.TYPE_NAME`

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
bernd: Aye :+1:

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
bernd: I would keep it as it is. We use `Map<String, Object>` for these types of return values all over the place. :smile:

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
bernd: We this in several other places as well without the type. I think it wouldn't be possible that we have two event definitions with the same `_id` but different types. Since the `_id` has a unique index.

bernd:
@mpfz0r I worked on your review comments.

graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
```diff
@@ -295,6 +295,13 @@ class FilterForm extends React.Component {
             <HelpBlock>{lodash.get(validation, 'errors.execute_every_ms[0]')}</HelpBlock>
           )}
         </FormGroup>
+        <Input id="schedule-checkbox"
+               type="checkbox"
+               name="_is_scheduled"
+               label="Enable scheduling"
+               help="Whether scheduling for this event definition should be enabled?"
```
mpfz0r, line 302: How about:
"Should this event definition be executed?"
Or maybe don't use the word `scheduling` at all, and just call it enable/disable?

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
mpfz0r: Fair enough

graylog2-server/src/main/java/org/graylog/events/context/EventDefinitionContextService.java
mpfz0r: My concern was not duplicates, but the fact that we are the config document from other types doesn't have the field we are querying.
But I guess it's fine.

bernd: Hmm. But then we would just get no result from the query, no?

mpfz0r: yeah, unless someone creates a plugin that uses the same field name in its config, which is quite unlikely :-D

graylog2-server/src/main/java/org/graylog/scheduler/DBJobDefinitionService.java
```diff
@@ -50,4 +58,52 @@ public DBJobDefinitionService(MongoConnection mongoConnection, MongoJackObjectMa
...
+        // Example aggregation output:
+        //   {
+        //     "config-field-value-1": [
+        //       {JobDefinitionDto}, {JobDefinitionDto}
+        //      ],
+        //     "config-field-value-2": [
+        //       {JobDefinitionDto}, {JobDefinitionDto}
+        //      ]
+        //   }
+        final Aggregation.Pipeline<Void> pipeline = Aggregation.match(DBQuery.in(field, values))
```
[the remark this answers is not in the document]
bernd: No, unfortunately not. The "config" structure in `JobDefinitionDto` is pluggable so there is no method that I could use for the grouping key. (without converting the object to a map or JsonNode and access the key again)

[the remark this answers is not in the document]
bernd: Okay, I tested this. Using the object mapper to deserialize the value and pick the config field isn't that hard. Less code than the aggregation.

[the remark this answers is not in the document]
bernd: @mpfz0r I implemented it with StreamEx. Good catch!

graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
bernd: Changed.

PR #7905, opened 2020-04-17
Moving All Hex Colors breakdown 1/6

kyleknighted:
## Description

Replacing hardcoded HEX color values to use theme values.

## Motivation and Context

So all parts of the app are themeable

## Types of changes

## Checklist:


graylog2-web-interface/src/components/content-packs/ContentPackConstraints.jsx
```diff
@@ -21,25 +25,23 @@ class ContentPackConstraints extends React.Component {
   };
 
   _rowFormatter = (item) => {
+    const { isFulfilled = item.fulfilled } = this.props;
```
ousmaneo, line 28: Hi @kyleknighted thank you for working on this.
A small thing I found is that on ContentPacks details page the constraints Badge always shows the unfulfilled version because we are using global isfulfilled value (which default to false) and not constraint.isfulfilled for each item.
And on ContentPack preview we pass isfulfilled=true which always show green checked badge.
I think we should use the constraint.isfulfilled to show the badges.

kyleknighted: The previous version was `const fulfilledIcon = (item.fulfilled || this.props.isFulfilled)`

We are not currently using `constraint`, are you saying that it is incorrectly rendering in master as well?

My version uses the `props` but has a fallback to `item.fulfilled` as its destructed if the prop is false.

kyleknighted: @ousmaneo Updated to use constraint with fallback

PR #7908, opened 2020-04-17
Show meaningful error message on missing permissions instead of redirecting to not found page

linuspahl:
We are currently redirecting the user to `/notfound` when an api response has the status code `403`.
With this PR we are no longer changing the url and providing a proper error message instead of the not found page.

While this PR is focused on unauthorized and react runtime errors, we can also use the error reporting to display the 404 Page (while staying on the same page) when a user tries to access a view which does not exist, instead of redirecting the user to `/notfound`

This PR adds:
- implements the `ErrorsAction` `report` which is being used to report unauthorized and react runtime errors
- implements two error boundaries, one app for unauthorized and react runtime errors and one for the routing which also catches runtime errors inside e.g. the navigation
- unifies the error pages layout
- bundles the error related components inside an own component directory.

## Types of changes

## Checklist:


graylog2-web-interface/src/pages/UnauthorizedErrorPage.jsx
```diff
@@ -0,0 +1,62 @@
...
+type Props = {
+  error: FetchError,
+  location: {
+    pathname: string
+  }
+}
+
+const UnauthorizedErrorPage = ({ error, location: { pathname } }: Props) => {
+  const errorMessage = error?.message || JSON.stringify(error);
+  const pageDetails = `The permissions check for the following request failed,\nwhile tying to access ${pathname}.`;
```
dennisoelkers, line 20: Typo: `tying` -> `trying`

graylog2-web-interface/src/pages/UnauthorizedErrorPage.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const description = (
+    <>
+      <p>You do not have the required permissions to view this ressource.</p>
```
dennisoelkers, line 23: Typo: `ressource` -> `resource`

graylog2-web-interface/src/components/errors/ReportedError.test.jsx
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  it('displays runtime error page when react error got reported', () => {
+    suppressConsole(async () => {
+      const { getByText } = render(<ReportedError router={router}>Hello World!</ReportedError>);
+
+      ErrorsActions.report(createReactError(new Error('The error message'), { componentStack: 'The component stack' }));
+
+      await waitForElementToBeRemoved(() => getByText('Hello World!'));
+      await waitForElement(() => getByText('Something went wrong.'));
```
dennisoelkers, line 58: `waitForElement` is [deprecated](https://testing-library.com/docs/dom-testing-library/api-async#waitforelement-deprecated-use-find-queries-or-waitfor) in the next version of the react testing-library and `waitFor` is a plugin replacement, so I think we should already start using it in new code.

graylog2-web-interface/src/components/errors/ReportedError.test.jsx
```diff
@@ -0,0 +1,95 @@
...
+
+
+  it('displays child component if there is no error', () => {
+    const { getByText } = render(<ReportedError router={router}>Hello World!</ReportedError>);
+
+    expect(getByText('Hello World!')).not.toBeNull();
+  });
+
+  it('displays runtime error page when react error got reported', () => {
+    suppressConsole(async () => {
```
dennisoelkers, line 52: `suppressConsole` should wrap the smallest possible scope to prevent suppressing valid console output, so maybe we can narrow it down? Why do we need it here anyway, I think there is not exception raised in here?

graylog2-web-interface/src/logic/rest/FetchProvider.js
```diff
@@ -76,7 +78,7 @@ export class Builder {
 
         // Redirect to the start page if a user is logged in but not allowed to access a certain HTTP API.
         if (SessionStore.isLoggedIn() && error.status === 403) {
-          history.replace(Routes.NOTFOUND);
+          ErrorsActions.report((createUnauthorizedError(error)));
```
dennisoelkers, line 81: Maybe we could create a factory (`createFromFetchError`?) in the `ReportedErrors` module which just takes the error and returns a proper subtype? It could look at the `status` field to find out which type to return. No need to do it in this PR though.

Optional: too many parentheses.

graylog2-web-interface/src/components/errors/ReportedError.test.jsx
linuspahl: Good point. I removed the not needed `suppressConsole`s and adjusted the scope of the remaining ones, which revealed a broken test.

graylog2-web-interface/src/logic/rest/FetchProvider.js
linuspahl: For now i removed the parentheses. I will add the factory when I take care of the `MissingStreamPermissions` case.

PR #7916, opened 2020-04-18
Add firstNonNull function to pipeline processor

### LINKED ISSUE #7895: Add pipeline firstNonNull function
## What?

Add a pipeline function to select a value from the first non-null value in an array of possible values. 

## Why?

When normalizing data it is common to define a value from a list of values, and set an implicit value if none exists. That process currently requires a pipeline rule for each possible value, plus an additional pipeline rule to set an implicit value if none of the source values exist.

The function:
`set_field ( field: "target_field", value: firstNonNull ( [ $message.value1, $message.value2, $message.value3, "default value" ] ) )`
... would replace 4 pipeline rules to do the same with the current functionality.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

radykal-com:
## Description
Adds a new function `first_non_null` to pipeline processor that accepts a list as unique parameter and returns the first non null value found in that list. Returns an empty string if none found.

## Motivation and Context
Fixes #7895 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
@@ -26,35 +26,31 @@ import org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor;
 import java.util.List;
 import java.util.Objects;
 
-import static org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor.object;
+public class FirstNonNull extends AbstractFunction<Object> {
+    public static final String NAME = "first_non_null";
 
-public class FirstNonNull extends AbstractFunction<String> {
-    public static final String NAME = "firstNonNull";
-
-    private final ParameterDescriptor<Object, Object> valueParam;
+    private final ParameterDescriptor<List, List> valueParam;
 
     public FirstNonNull() {
-        valueParam = object("value").description("The list of fields to find first non null value").build();
+        valueParam = ParameterDescriptor.type("value", List.class, List.class)
+                .description("The list of fields to find first non null value")
+                .build();
     }
 
     @Override
-    public String evaluate(FunctionArgs args, EvaluationContext context) {
-        Object elements = valueParam.required(args, context);
-        if (elements instanceof List) {
-            List elementsList = (List) elements;
-            return elementsList.stream().filter(Objects::nonNull).findFirst().orElse("").toString();
-        }
-        return "";
+    public Object evaluate(FunctionArgs args, EvaluationContext context) {
+        List elements = valueParam.required(args, context);
+        return elements.stream().filter(Objects::nonNull).findFirst().orElse(null);
     }
 
     @Override
-    public FunctionDescriptor<String> descriptor() {
-        return FunctionDescriptor.<String>builder()
+    public FunctionDescriptor<Object> descriptor() {
+        return FunctionDescriptor.builder()
                 .name(NAME)
                 .pure(false)
-                .returnType(String.class)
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import static org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor.object;
+
+public class FirstNonNull extends AbstractFunction<String> {
+    public static final String NAME = "firstNonNull";
+
+    private final ParameterDescriptor<Object, Object> valueParam;
```
kroepke, line 34: I believe `<List, List>` is the right signature here: we only accept lists anyway, so we should communicate that to the system, so it can generate meaningful error messages.

Right now passing anything that's not a `List` silently returns `""`, which can be confusing.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public FirstNonNull() {
+        valueParam = object("value").description("The list of fields to find first non null value").build();
```
kroepke, line 37: This would become `type("value", List.class)`

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public String evaluate(FunctionArgs args, EvaluationContext context) {
+        Object elements = valueParam.required(args, context);
+        if (elements instanceof List) {
+            List elementsList = (List) elements;
+            return elementsList.stream().filter(Objects::nonNull).findFirst().orElse("").toString();
+        }
+        return "";
```
kroepke, line 47: I think the default should be `null`.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
```diff
@@ -0,0 +1,60 @@
...
+    public FirstNonNull() {
+        valueParam = object("value").description("The list of fields to find first non null value").build();
+    }
+
+    @Override
+    public String evaluate(FunctionArgs args, EvaluationContext context) {
+        Object elements = valueParam.required(args, context);
+        if (elements instanceof List) {
+            List elementsList = (List) elements;
+            return elementsList.stream().filter(Objects::nonNull).findFirst().orElse("").toString();
```
kroepke, line 45: default would become `null` and no `.toString()`

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/FirstNonNull.java
kroepke: E.g. if the function accepts `ParameterDescriptor<List, List>` as its value, then the error message becomes:

```
org.graylog.plugins.pipelineprocessor.parser.ParseException: Errors:
 Expected type List for argument value but found String in call to function firstNonNull in line 4 pos 21
```

radykal-com:
I've made all requested changes. Thanks you for the review!

kroepke:
Many thanks!

PR #7929, opened 2020-04-21
Implement field types context for search pages

linuspahl:
This PR implements the `FieldTypesContext` for all search related cases where we are connecting the `FieldTypesStore`.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
```diff
@@ -175,10 +179,29 @@ const FieldList = createReactClass({
...
+FieldListWithContext.propTypes = {
+  viewMetadata: PropTypes.object.isRequired,
+  listHeight: PropTypes.number,
+};
+
+FieldListWithContext.defaultProps = {
+  listHeight: 50,
+};
+
+export default connect(FieldListWithContext, { viewMetadata: ViewMetadataStore });
```
linuspahl, line 207: Because we need `viewMetadata` for this component independently of the fields, I removed  the `ViewMetadataStore` connect from the `ExtendedSearchPage` `FieldList` and implemented it as a part of the component.

graylog2-web-interface/src/views/components/contexts/FieldTypesProvider.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    fieldTypes
+      ? (
+        <FieldTypesContext.Provider value={fieldTypes}>
+          {children}
+        </FieldTypesContext.Provider>
+      )
+      : children
+  );
+};
```
linuspahl, line 20: We had this discussion for e.g. the `CurrentUserProvider`: Right now I am expecting `fieldTypes` to be `undefined` initially, which means we need to test / check the `undefined` case in every component where we are using the `FieldTypesContext`.

graylog2-web-interface/src/views/components/SearchResult.jsx
```diff
@@ -1,12 +1,11 @@
 // @flow strict
-import * as React from 'react';
+import React, { useContext } from 'react';
```
dennisoelkers, line 2: We should split this up into:

```suggestion
import * as React from 'react';
import { useContext } from 'react';
```

This way we really do have `React` containing all exports and not just `default`.

graylog2-web-interface/src/views/components/contexts/FieldTypesContext.jsx
```diff
@@ -0,0 +1,8 @@
+// @flow strict
+import * as React from 'react';
+
+import { singleton } from 'views/logic/singleton';
+import type { FieldTypesStoreState } from 'views/stores/FieldTypesStore';
+
+const FieldTypesContext = React.createContext<?FieldTypesStoreState>();
```
dennisoelkers, line 7: Instead of reusing the store state, we should create an own type in the context that we are populating. If it is structurally identical to `FieldTypesStoreState`, we do not even need a conversion. This helps us decouple the context from any component that is providing it.

graylog2-web-interface/src/views/components/contexts/FieldTypesProvider.jsx
```diff
@@ -0,0 +1,26 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+import { useStore } from 'stores/connect';
+import { FieldTypesStore } from 'views/stores/FieldTypesStore';
+import FieldTypesContext from './FieldTypesContext';
+
+const FieldTypesProvider = ({ children }: { children: React.Node }) => {
```
dennisoelkers, line 9: We should name this either `DefaultFieldTypesProvider` or `StoreBackedFieldTypesProvider` or something like this, so it's clear that it is actually a specific implementation for the provider and not just the "interface".

graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
```diff
@@ -18,16 +20,16 @@ const isReservedField = (fieldName) => MessageFieldsFilter.FILTERED_FIELDS.inclu
 
 const FieldList = createReactClass({
```
dennisoelkers, line 21: Now that the mixins are gone, can we turn this into a class-based or functional component?

graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
dennisoelkers: The `ViewMetadataStore` is probably the next target that could be provided with a context instead.

graylog2-web-interface/src/views/components/SearchResult.jsx
linuspahl: So far I've used `* as React` only when needed, but I see the benefit in always exporting not just `default`

graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
linuspahl: I thought about it as well, but I prefer doing this in another PR, otherwise reviewing the `FieldList` changes would have been more complicated.

PR #7933, opened 2020-04-21
Remove waitForElementToBeRemoved from ReportedError test

linuspahl:
The tests are currently failing on the `master` branch, because `waitForElementToBeRemoved`  is not part of `wrappedTestingLibrary `.

This PR fixes the issue by using `wait` instead of `waitForElementToBeRemoved` for the related test cases.

### DIFF
--- a/graylog2-web-interface/src/components/errors/ReportedError.test.jsx
+++ b/graylog2-web-interface/src/components/errors/ReportedError.test.jsx
@@ -1,6 +1,6 @@
 // @flow strict
 import React from 'react';
-import { render, cleanup, wait, waitForElementToBeRemoved } from 'wrappedTestingLibrary';
+import { render, cleanup, wait } from 'wrappedTestingLibrary';
 import { act } from 'react-dom/test-utils';
 
 import suppressConsole from 'helpers/suppressConsole';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/errors/ReportedError.test.jsx
```diff
@@ -50,25 +50,24 @@ describe('ReportedError', () => {
...
   it('displays unauthorized error page when unauthorized error got reported', async () => {
-    const { getByText } = render(<ReportedError router={router}>Hello World!</ReportedError>);
+    const { getByText, queryByText } = render(<ReportedError router={router}>Hello World!</ReportedError>);
 
     suppressConsole(() => {
       ErrorsActions.report(createUnauthorizedError(new FetchError('The request error message', new Error('The request error message'))));
     });
 
-    await waitForElementToBeRemoved(() => getByText('Hello World!'));
+    await wait(() => expect(queryByText('Hello World!')).toBeNull());
```
kyleknighted, line 70: After talking with @linuspahl, he chose `queryByText` because 
> I had to use queryByText  because getByText  would throw “Unable to find an element with the text:”

After looking into [`queryBy` and `getBy`](https://testing-library.com/docs/react-testing-library/cheatsheet) it appears to me we should be using `queryBy` in these other tests as well since `getBy` will never return null, so the test will always succeed in `not.toBeNull()`

PR #7952, opened 2020-04-23
Post events from export to be used for audit logging

alex-konn:
This PR adds audit log events for the CSV export. It should be rebased after the main export PR is merged #7709.


graylog2-server/src/main/java/org/graylog/plugins/views/search/events/MessagesExportEvent.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public abstract String auditType();
+
+    public abstract DateTime timestamp();
+
+    public abstract AbsoluteRange timeRange();
+
+    public abstract String queryString();
+
+    public abstract Set<String> streams();
```
kmerz, line 68: I would add the `limit` for a rough estimation how many messages were exported. Or perhaps the total message count in the succeed case?

PR #7957, opened 2020-04-23
Create a common list of field names for the Graylog Schema.

waab76:
## Description
The Graylog Security team has created a common schema to use as the basis for future official content packs.  This change will make it easier for new and existing Inputs to re-use the fields defined for the Graylog Schema.

## How Has This Been Tested?
Fields have been tested with the new Okta enterprise plugin.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/inputs/GraylogSchemaFields.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/GraylogSchemaFields.java
@@ -1,3 +1,19 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/inputs/GraylogSchemaFields.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.inputs;
```
bernd, line 17: What do you think about moving this into the `org.graylog.schema` package? (doesn't exist yet) I think we will have more related code in there in the future and so it isn't buried in the inputs package. :smiley:

waab76: Great idea.  That will promote visibility and re-use.  I'll make the change.

Issue #6493, opened 2019-09-25
Allow POST in HTTP/JsonPath data adapter


kroepke:
## Expected Behavior
The HTTP/JsonPath data adapter should support the POST method and allow specifying a body template that gets evaluated with the key that is being looked up.

## Current Behavior
The current implementation of the HTTP/JsonPath data adapter only supports the GET method to retrieve data from the remote location.
For third party systems that model their searches as POSTs, such as MISP or even Graylog itself, only support GET is not enough.

## Possible Solution
Add a radio button or drop down for the HTTP method (radio if all we want to support is GET and POST, I think the others don't make sense right now).
Allow pasting a body template for POST and evaluate it the same way we do it for the URL right now.

## Context
When trying to read data from external systems like MISP, you are faced with the problem that it requires POSTing a search, which either forces you to write middleware or implement an entirely new plugin just to be able to send a POST request, which seems excessive.

## Your Environment

* Graylog Version: 3.1
* Elasticsearch Version: not applicable
* MongoDB Version: not applicable
* Operating System: not applicable
* Browser version: not applicable

solvaholic:
When this issue was raised, [Graylog 3.x had only `lookup()` and `lookup_value()` functions](https://docs.graylog.org/en/3.1/pages/pipelines/functions.html#lookup). Now Graylog 3.3 and 4.0 have `lookup_` functions for adding, removing, and setting values as well:

https://docs.graylog.org/en/4.0/pages/pipelines/functions.html#lookup
https://docs.graylog.org/en/3.3/pages/pipelines/functions.html#lookup

Making the HTTP request method part of the lookup table data adapter configuration, as requested here and in the issues @radykal-com linked last year, would require configuring separate lookup table data adapters for reading and writing data.

How would it be to, instead, extend the current HTTPJSONPath lookup table data adapter to select whichever [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) is appropriate for the `lookup_` function called in the pipeline rule?

I think having a general-purpose HTTP API lookup table data adapter like that would create opportunities for the Graylog community to create and share lookup table data adapter back-ends to meet a wide variety of needs.

PR #7972, opened 2020-04-24
Allow POST in HTTP/JsonPath input

radykal-com:
## Description
Adds new METHOD and BODY configuration options to the HTTP/JsonPath input to allow POST requests with body

## How Has This Been Tested?
Bring up a docker container like https://github.com/mendhak/docker-http-https-echo

Configure a HTTP/JsonPath input to send POST requests with any body to the http echo server and configure any of the request body fields (if using a json body) as JsonPath to parse.

Validate the server receives the expected request and also the mapping works extracting the field from the response, the extracted data should match the same as in the request body.

## Types of changes

## Checklist:


bernd:
@radykal-com Thank you!

Unfortunately we worded the description #6493 in an ambiguous way. :disappointed: It was supposed to describe a feature for the "HTTPJSONPath" lookup table data adapter and not the "JSON path from HTTP API" input. I will update the description in #6493.

We will review if this PR, of course. I might take some time though, because there are quite a few changes to the configuration fields system and we are already in beta for the 3.3. release.

radykal-com:
Sorry, I think it was my fault, as in the issue description the "input" was never there. Anyway, the implementation may have sense. I'll check the issue again and implement what is supposed to be implemented

PR #7989, opened 2020-04-28
Update snapshot which is failing the build

kmerz:
Update failing snapshot tests.


kyleknighted:
`continuous-integration/jenkins/branch` build is mistaken and @juju2112 is looking into the concerns. Merging now as it all passes locally.

PR #7992, opened 2020-04-28
Allow free license request from the Enterprise overview page

bernd:


graylog2-web-interface/src/components/enterprise/EnterpriseFreeLicenseForm.jsx
```diff
@@ -0,0 +1,127 @@
...
+    this.clearValues();
+  };
+
+  render() {
+    const { formFields: { firstName, lastName, company, email } } = this.state;
+
+    return (
+      <form onSubmit={this.submitForm}>
+        <Row>
+          <Col md={5}>
```
kroepke, line 78: This leads to a truncated form:
image

Can we make it a bit wider?

graylog2-web-interface/src/pages/EnterprisePage.jsx
```diff
@@ -1,48 +1,153 @@
...
-          <GraylogClusterOverview />
-          <PluginList />
+          <GraylogClusterOverview layout="compact">
+            <PluginList />
+          </GraylogClusterOverview>
+          <IfPermitted permissions="freelicenses:create">
+            <Row className="content">
+              <Col md={6}>
+                <BiggerFontSize>
+                  <p>Extend Graylog’s Open Source capabilities with a free trial of Graylog Enterprise (up to 5 GB/Day).
```
kroepke, line 111: I would add a break after the first sentence to make it stand out a little more.

graylog2-web-interface/src/pages/EnterprisePage.jsx
```diff
@@ -1,48 +1,153 @@
...
+                    Graylog Enterprise introduces productivity and compliance features designed to help organizations
+                    reduce risk while encouraging collaboration across a large number of users.</p>
+
+                  <p>Graylog Enterprise includes:</p>
+
+                  <EnterpriseFeatureList>
+                    <li>Automated <Link to={Routes.pluginRoute('SYSTEM_ARCHIVES')}><strong>archiving</strong></Link> and
```
kroepke, line 118: If the enterprise plugin isn't installed the internal links will end up in 404s, which isn't pretty.
Should we instead always link to the website?

graylog2-web-interface/src/components/enterprise/EnterpriseFreeLicenseForm.jsx
bernd: Good catch. Yes! :+1:

graylog2-web-interface/src/pages/EnterprisePage.jsx
bernd: Good catch! :+1:

Issue #7812, opened 2020-04-01
Need separate events for input stopped and input deleted


waab76:
## Expected Behavior
When an input is stopped either by the user or due to an error (deleted from the `InputRegistry`), an `InputStopped` event should be propagated by the EventBus.

When an input is deleted (deleted from the `InputStateService`), an `InputDeleted` event should be propagated by the EventBus

## Current Behavior
Currently, an `InputDeleted` event is propagated by the EventBus in both cases (input stopped and input deleted).  This makes it impossible for components to react to one event type but not the other.

## Context
For a new Input with persistent state stored in MongoDB, I would like to delete the MongoDB record when the Input is deleted, but retain the MongoDB record when the input is stopped.  This is currently impossible because both state transitions emit the same event type.

waab76:
@Graylog2/architecture Has any more thought been given to this issue? It now seems to be at least part of the root cause of a customer issue described here: https://github.com/Graylog2/support/issues/148

PR #7993, opened 2020-04-28
Add workaround for input event problems in MongoInputStatusService

bernd:
This is a workaround for the problem described in #7812.

When we receive an "InputDeleted" event, check if the input still exists
in the database. If so, the input has only been stopped and we don't
want to delete our state.

Refs #7812

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/inputs/persistence/MongoInputStatusServiceTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/inputs/persistence/MongoInputStatusServiceTest.java
@@ -22,6 +22,8 @@ import org.bson.types.ObjectId;
 import org.graylog.testing.mongodb.MongoDBFixtures;
 import org.graylog.testing.mongodb.MongoDBInstance;
 import org.graylog2.bindings.providers.MongoJackObjectMapperProvider;
+import org.graylog2.database.NotFoundException;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog2/inputs/persistence/MongoInputStatusServiceTest.java
```diff
@@ -182,24 +188,25 @@ public void delete_ReturnsZero_WhenDeletingNonExistantRecord() {
...
-        assertThat(optDbRecord, notNullValue());
-        assertThat(optDbRecord.isPresent(), is(true));
+        // Simulate that the input has actually been deleted
+        // TODO: This will change once we fix https://github.com/Graylog2/graylog2-server/issues/7812
+        when(inputService.find(deletedInput)).thenThrow(new NotFoundException());
+
+        cut.handleInputDeleted(inputDeletedEvent);
+        // The record should be removed from the DB
+        assertThat(cut.get(deletedInput).isPresent(), is(false));
     }
```
waab76, line 211: This should be split into two separate tests to make it easier to see if a change breaks one but not the other.

bernd: Good point! :+1:

PR #8009, opened 2020-04-30
Provide a different IndexHtmlGenerator for development

bernd:
The production "index.html" generator is throwing an exception when the
production assets cannot be found on server startup. That means
developers need to run a full maven build including the production web
interface before the development server can start successfully in the IDE.

With this PR we can switch to a different IndexHtmlGenerator
implementation that doesn't require the production web assets to be
present on server startup. (it serves a dummy "index.html" page)

Switching to this implementation requires the "DEVELOPMENT" environment
variable to be set.


bernd:
I found an issue, please wait with review.

bernd:
Nevermind, it was just a missing "mvn clean". :smile:

PR #8031, opened 2020-05-04
Mask passwords in input config returned from /system/inputstates

bernd:
This PR masks all passwords in input configuration values that are returned from the `/system/inputstates` endpoint.

Please double check if the inputstates endpoint consumers somehow rely on the passwords being returned in plain-text. I can't think of any reason why they should.


mpfz0r:
Refs: https://github.com/Graylog2/graylog2-server/pull/5432

PR #8041, opened 2020-05-05
Define less strict configuration for ESlint rule `jsx-a11y/label-has-associated-control`

linuspahl:
The ESLint rule [`jsx-a11y/label-has-associated-control`](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md) requires the following structure for input labels:
```
<label htmlFor="input-id">
  The input label
  <input id="input-id" />
</label>
```

We are usually not using this structure in our project, more common is:
```
<label htmlFor="input-id">The input label</label>
<input id="input-id" />
```

This PR is adjusting the ESLint rule, to allow either the nested format or the sibling structure.

### DIFF
--- a/graylog2-web-interface/packages/eslint-config-graylog/index.js
+++ b/graylog2-web-interface/packages/eslint-config-graylog/index.js
@@ -17,7 +17,7 @@ module.exports = {
     'import/extensions': 'off',
     'import/no-extraneous-dependencies': 'off',
     'import/no-unresolved': 'off',
-    'jsx-a11y/label-has-associated-control': [2, { assert: 'either' }],
+    'jsx-a11y/label-has-associated-control': ['error', { assert: 'either' }],
     'max-classes-per-file': 'off',
     'max-len': 'off',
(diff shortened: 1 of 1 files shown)


kyleknighted:
What's your thought on [`labelComponents`](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md#rule-details) referencing `src/components/graylog/Label.jsx` and `controlComponents` referencing `src/components/bootstrap/Input.jsx`?

Not sure how detailed we want to get this rule so that we can still be reasonably sure we're writing consistent code.

linuspahl:
@kyleknighted Right now I don't see a direct benefit, but I am not against adding them. We could e.g. use 
```
<label>
  The label
  <Input />
</label>
```
without the `htmlFor` attribute, but we are not currently not doing this. And when we start adding `labelComponents` and `controlComponents ` we should add many other components to the list as well, like all custom `Select` components.

PR #8050, opened 2020-05-06
Remove elasticsearch_request_timeout config option

mpfz0r:
This option was mostly unused since we started using Jest
to communicate with ES.
Migrate the only remaining usage in the Cluster health checker
to to `elasticsearch_socket_timeout`, which is the effective timeout
used for each Jest request.


mpfz0r:
I think we can merge this after 3.3 is branched

PR #8051, opened 2020-05-06
Extract method to generate new salt for encryption.

dennisoelkers:
## Motivation and Context

This change is adding a new static method to our `AESTools` utility
class, which generates a new random salt. This method is reused in
`LdapLdapSettingsImpl`. It will also be reused in future code changes
where the generation of a new salt for encryption is desired.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/AESTools.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/AESTools.java
@@ -25,6 +25,7 @@ import javax.annotation.Nullable;
 import javax.crypto.Cipher;
 import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
+import java.security.SecureRandom;
 import java.util.Arrays;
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/security/AESTools.java
```diff
@@ -110,6 +111,19 @@ public static String decryptSiv(String cipherText, byte[] encryptionKey) {
         return null;
     }
 
+    /**
+     * Generates a new random salt
+     *
+     * @return the generated random salt as a
```
thll, line 117:
```suggestion
     * @return the generated random salt as a string of hexadecimal digits.
```

graylog2-server/src/main/java/org/graylog2/security/AESTools.java
```diff
@@ -110,6 +111,19 @@ public static String decryptSiv(String cipherText, byte[] encryptionKey) {
...
+     */
+
```
thll, line 119:
```suggestion
```

PR #8052, opened 2020-05-06
Give search input focus when saved search list is opening

kmerz:
## Motivation
Prior to this change, a user needed to click on the search
input after he opened the saved search modal.

## Description
This change will give the search input of the saved search
modal focus, so the user needs one click less.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/SearchForm.jsx
+++ b/graylog2-web-interface/src/components/common/SearchForm.jsx
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import * as React from 'react';
 import Promise from 'bluebird';
 
 import { Button } from 'components/graylog';
@@ -74,6 +74,7 @@ class SearchForm extends React.Component {
       PropTypes.arrayOf(PropTypes.element),
       PropTypes.element,
     ]),
+    focusAfterMount: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -93,12 +94,20 @@ class SearchForm extends React.Component {
     loadingLabel: 'Loading...',
     queryHelpComponent: null,
     children: null,
+    focusAfterMount: false,
   };
 
   state = {
     query: this.props.query,
     isLoading: false,
   };
+
+  componentDidMount(): void {
+    const { focusAfterMount } = this.props;
+    if (focusAfterMount && this.searchInput) {
+      this.searchInput.focus();
+    }
+  }
 
   componentWillReceiveProps(nextProps) {
     // The query might get reset outside of this component so we have to adjust the internal state
@@ -162,6 +171,7 @@ class SearchForm extends React.Component {
           <div className="form-group has-feedback">
             {this.props.label && <label htmlFor="common-search-form-query-input" className="control-label">{this.props.label}</label>}
             <input id="common-search-form-query-input"
+                   ref={(ref) => { this.searchInput = ref; }}
                    onChange={this.handleQueryChange}
                    value={this.state.query}
                    placeholder={this.props.placeholder}
--- a/graylog2-web-interface/src/components/common/SearchForm.jsx
+++ b/graylog2-web-interface/src/components/common/SearchForm.jsx
@@ -102,13 +102,6 @@ class SearchForm extends React.Component {
     isLoading: false,
   };
 
-  componentDidMount(): void {
-    const { focusAfterMount } = this.props;
-    if (focusAfterMount && this.searchInput) {
-      this.searchInput.focus();
-    }
-  }
-
   componentWillReceiveProps(nextProps) {
     // The query might get reset outside of this component so we have to adjust the internal state
     if (this.props.query !== nextProps.query) {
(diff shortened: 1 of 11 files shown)


linuspahl:
Did you considered using the `autoFocus` input prop? Looks like we have an ESlint rule `eslintjsx-a11y/no-autofocus` (The autoFocus prop should not be used, as it can reduce usability and accessibility for users.) to avoid this, but is is technically doing the same.

kmerz:
> Did you considered using the `autoFocus` input prop? Looks like we have an ESlint rule `eslintjsx-a11y/no-autofocus` (The autoFocus prop should not be used, as it can reduce usability and accessibility for users.) to avoid this, but it is technically doing the same.

Didn't know that! That is much better! Thanks

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import * as React from 'react';
```
kyleknighted, line 2: Don't we only need to do `* as` when it's Flowtyped?

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -162,6 +171,7 @@ class SearchForm extends React.Component {
           <div className="form-group has-feedback">
             {this.props.label && <label htmlFor="common-search-form-query-input" className="control-label">{this.props.label}</label>}
             <input id="common-search-form-query-input"
+                   ref={(ref) => { this.searchInput = ref; }}
```
kyleknighted, line 174: What's your feeling on using React's [`createRef`](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs)?

graylog2-web-interface/src/components/common/SearchForm.jsx
```diff
@@ -93,13 +94,21 @@ class SearchForm extends React.Component {
...
     children: null,
+    focusAfterMount: false,
   };
 
   state = {
     query: this.props.query,
     isLoading: false,
   };
 
+  componentDidMount(): void {
```
kyleknighted, line 105: What is the `: void` for?

graylog2-web-interface/src/components/common/SearchForm.jsx
linuspahl: It is not required to add `* as` when using flow, but for a weeks I also prefer importing `React` this way, because we can be sure `React` contains all imports and not only `default`.

linuspahl:
@kmerz Looks good to me, but there are failing snapshot tests.

Because we talked today about general improvements today:
We could load a saved search directly when clicking on it (or on enter when an entry is focused).
In addition we need to add the "Delete" button e.g. as a trash can for each list item.

This would reduce the amount of clicks a user needs to load a saved search and we would support the keyboard navigation.

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchList.jsx
```diff
@@ -2,6 +2,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { browserHistory } from 'react-router';
+// $FlowFixMe should be there
```
linuspahl, line 5: Looks like this comment is not required

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchList.jsx
```diff
@@ -39,6 +40,12 @@ const NoSavedSearches: StyledComponent<{}, ThemeInterface, *> = styled(Alert)`
   align-items: center;
 `;
 
+const DeleteButton: StyledComponent<{}, ThemeInterface, *> = styled.span`
```
linuspahl, line 43: I see that we are using the `*` for other component type definitions in this file, but I would prefer `HTMLSpanElement`.

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchList.jsx
```diff
@@ -117,11 +127,19 @@ class SavedSearchList extends React.Component<Props, State> {
...
-        </ListGroupItem>
+        <ViewLoaderContext.Consumer key={savedSearch.id}>
+          {(loaderFunc) => (
+            <ListGroupItem active={selectedSavedSearch === savedSearch.id}
+                           onClick={() => this.onLoad(savedSearch.id, loaderFunc)}>
+              {savedSearch.title}
+              <span>
+                <DeleteButton bsSize="xsmall" bsStyle="danger" onClick={(e) => this.onDelete(e, savedSearch.id)}>
+                  <Icon name="trash" ttile="Delete" data-testid={`delete-${savedSearch.id}`} />
+                </DeleteButton>
```
linuspahl, line 138: The `DeleteButton` can't be accessed with the keyboard right now. Maybe we can use the `IconButton` in the future instead. https://github.com/Graylog2/graylog2-server/pull/8064

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchList.jsx
kmerz: Well my flow does say something else :shrug:

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchList.jsx
kmerz: When the future is ready, so am I ;)

PR #8069, opened 2020-05-08
Create flow type definition for search types

linuspahl:
This PR adds a flow type definition for query search types. There are a few more files were we could use this type definition, but these files are currently not using `// @flow`

## Motivation and Context
I want to remove the `searchType: any` declaration in `views/logic/queries/Query` and the `CSVExportModal`.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/logic/queries/SearchType.js
+++ b/graylog2-web-interface/src/views/logic/queries/SearchType.js
@@ -10,7 +10,7 @@ type SearchTypePivot = {
   field: string,
   limit?: number,
   interval?: Interval,
-  };
+};
 
 type SearchTypeBase = {
   filter: ?string,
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/views/logic/queries/SearchType.js
```diff
@@ -0,0 +1,40 @@
...
+import type { Decorator } from 'views/components/messagelist/decorators/Types';
+import type { Interval } from 'views/components/aggregationbuilder/pivottypes/Interval';
+import type { ElasticsearchQueryString, TimeRange } from './Query';
+
+type SearchTypePivot = {
+  type: string,
+  field: string,
+  limit?: number,
+  interval?: Interval,
+};
```
linuspahl, line 13: I thought about using a union type here as well, but it would be good to know if the already have a type definition for pivots? I found some, but they were not suitable for a search type pivot.

kmerz:
\rebase

PR #8081, opened 2020-05-11
Allow arbitrary key sizes for AESTools.encrypt/decrypt.

dennisoelkers:
## Motivation and Context

In the current state, the `AESTools.encrypt/decrypt` methods rely on key
sizes being 16, 24 or 32 bytes. This makes functionality using it with
user-supplied keys fail if their lengths are not exactly that size.

In order to allow arbitrary key sizes, this change is cutting keys if
they are longer than 32 bytes or pads them if they are shorter than the
desired key size. Also, it uses the first of the three key sizes which
can hold as most of the supplied key as possible.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/AESTools.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/AESTools.java
@@ -37,8 +37,21 @@ public class AESTools {
 
     private static final SivMode SIV_MODE = new SivMode();
 
+    /**
+     * Encrypt the given plain text value with the given encryption key and salt using AES CBC.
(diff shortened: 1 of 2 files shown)


dennisoelkers:
Good catch, @thll!

I have addressed this by converting the string to a bytes array early on, so the length of strings is handled properly. The downside is that if a multi-byte codepoint overlaps the desired key boundary, it will be truncated.

In general, the implementation would benefit from being able to (optionally) specify the secret in hexadecimal notation. This would also extend the space the key is chosen from beyond printable characters.

PR #8083, opened 2020-05-11
Improve rendering performance of field selects in event definition UI.

### LINKED ISSUE #7752: Slow field sorting with lots of fields in the UI
## Description
I noticed that the "Filter & Aggregation" tab in my event definition editor took around 7 seconds to fully load the form:

Peek 2020-03-25 15-16

The first part of the form loads fast, but the bottom part takes a while.

In the chrome profiler I noticed that the `naturalSort` function seems to be the issue. (from the "javascript-natural-sort" library)

image

I pinned it down to the `AggregationForm` component.

https://github.com/Graylog2/graylog2-server/blob/729bf09540231c9517a204629607b542e2fd25a3/graylog2-web-interface/src/components/event-definitions/event-definition-types/AggregationForm.jsx#L18-L25

In that component we use `naturalSortIgnoreCase()` (via `SortUtils` which then uses `naturalSort()`) to sort the fields for the field drop-down menus (group by, function field)

My setup has 2605 fields.

## Possible Solution

I extracted the code from our codebase and wrote a quick benchmark with nodejs and my exported fields list.

The result for the same sort function was around 5 seconds. I did a quick search and found the [fast-sort](https://www.npmjs.com/package/fast-sort) JavaScript library. Using that in my benchmark script I got the following results:

```
$ node sort.js fields.json 
number of fields: 2605
--[measurements]----------------
javascript-natural-sort (naturalSortIgnoreCase): 4831.59 ms
javascript-natural-sort (naturalSortCase): 476.23 ms
javascript-natural-sort (naturalSort): 487.63 ms
fast-sort: 1.09 ms
fast-sort (natural sort): 2.74 ms
fast-sort (natural sort, lower-case): 1.04 ms
```

I guess we should switch to that library. :smile: (or a different one if there is an even better one out there)

**UPDATE:** @dennisoelkers did some tests with the [JavaScript native sort function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) and that's even faster than the fast-sort library. That means we don't need a 3rd party library to improve the performance. Replacing "javascript-natural-sort" with the native sort function that's using a "natural sort" comparator instance should be sufficient.

We import the `javascript-natural-sort` library in 34 files in Graylog server (didn't check plugins) and the `naturalSortIgnoreCase()` function from `SortUtils` in 10 files. So we should probably replace the `javascript-natural-sort` library in our code base by something better to avoid massive slowdowns.

Here is my benchmark script. The `fields.json` I exported from my Graylog server via the `/api/views/fields` HTTP endpoint. (I can make this file available privately)

<details>
  <summary>sort.js</summary>

```js
const moment = require('moment');
const naturalSort = require('javascript-natural-sort');
const lodash = require('lodash');
const numeral = require('numeral');

const sort = require('fast-sort'); // https://www.npmjs.com/package/fast-sort

const fs = require('fs');
const path = require('path');
const process = require('process');
const { performance } = require('perf_hooks')

const filename = process.argv.slice(2)[0];


if (!filename) {
  console.log(`Usage: ${path.basename(process.argv[1])} /path/to/fields.json`);
  process.exit(1);
}


// Taken from SortUtils.js
function naturalSortIgnoreCase(s1, s2, sortOrder) {
  const effectiveSortOrder = sortOrder || 'asc';
  return (effectiveSortOrder === 'asc' ? naturalSort(s1.toLowerCase(), s2.toLowerCase()) : naturalSort(s2.toLowerCase(), s1.toLowerCase()));
}

function naturalSortCase(s1, s2, sortOrder) {
  const effectiveSortOrder = sortOrder || 'asc';
  return (effectiveSortOrder === 'asc' ? naturalSort(s1, s2) : naturalSort(s2, s1));
}

// Measure function execution
const measure = (title, f) => {
  const start = performance.now();
  f();
  console.log(title + ':', numeral(performance.now() - start).format('0.00'), 'ms');
};

// Reusable fast-sort instance with natural sort enabled
const fnSort = sort.createNewInstance({
  comparer: new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' }).compare,
});


const fields = JSON.parse(fs.readFileSync(filename));

console.log('number of fields:', fields.length)
console.log('--[measurements]----------------')

measure('javascript-natural-sort (naturalSortIgnoreCase)', () => {
  // Taken from AggregationForm.jsx -- this is what we currently use
  fields.sort((ftA, ftB) => naturalSortIgnoreCase(ftA.name, ftB.name));
});

measure('javascript-natural-sort (naturalSortCase)', () => {
  fields.sort((ftA, ftB) => naturalSortCase(ftA.name, ftB.name));
});

measure('javascript-natural-sort (naturalSort)', () => {
  fields.sort((ftA, ftB) => naturalSort(ftA.name, ftB.name));
});

measure('fast-sort', () => {
  sort(fields).asc(f => f.name);
});

measure('fast-sort (natural sort)', () => {
  fnSort(fields).asc(f => f.name);
});

measure('fast-sort (natural sort, lower-case)', () => {
  fnSort(fields).asc(f => f.name.toLowerCase());
});
```

</details>

1. Create lots of message fields
2. Create and save event definition ("Filter & Aggregation")
3. Edit event definition again and switch to the "Filter & Aggregation" tab
4. Observe slow loading

## Your Environment

* Graylog Version: 3.3.0-SNAPSHOT (rev 0ab8db)

dennisoelkers:
## Motivation and Context

This PR is doing two things to improve rendering for the event
definition UI:

  1. Use `defaultCompare` instead of `naturalSort`
  2. Add and use option for `Select` component to not strip
     accents/diacritics

Fixes #7752.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -187,6 +187,7 @@ type Props = {
   delimiter?: string,
   disabled?: boolean,
   displayKey?: string,
+  ignoreAccents?: boolean,
   inputProps?: { [string]: any },
   matchProp?: 'any' | 'label' | 'value',
   multi?: boolean,
@@ -292,7 +293,7 @@ class Select extends React.Component<Props, State> {
     };
   }
 
-  componentWillReceiveProps = (nextProps: Props) => {
+  componentWillReceivePropsProps = (nextProps: Props) => {
     const { inputProps, optionRenderer, value, valueRenderer } = this.props;
     if (value !== nextProps.value) {
       this.setState({ value: nextProps.value });
@@ -362,7 +363,7 @@ class Select extends React.Component<Props, State> {
     });
   };
 
-  createCustomFilter = (stringify) => {
+  createCustomFilter = (stringify: (any) => string) => {
     const { matchProp, ignoreAccents } = this.props;
     const options = { ignoreAccents };
 
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -293,7 +293,7 @@ class Select extends React.Component<Props, State> {
     };
   }
 
-  componentWillReceivePropsProps = (nextProps: Props) => {
+  componentWillReceiveProps = (nextProps: Props) => {
     const { inputProps, optionRenderer, value, valueRenderer } = this.props;
     if (value !== nextProps.value) {
       this.setState({ value: nextProps.value });
--- a/graylog2-web-interface/src/components/common/Select.jsx
+++ b/graylog2-web-interface/src/components/common/Select.jsx
@@ -222,6 +222,8 @@ class Select extends React.Component<Props, State> {
     disabled: PropTypes.bool,
     /** Indicates which option object key contains the text to display in the select input. Same as react-select's `labelKey` prop. */
     displayKey: PropTypes.string,
+    /** Indicates whether the auto-completion should return results including accents/diacritics when searching for their non-accent counterpart */
+    ignoreAccents: PropTypes.bool,
     /**
      * @deprecated Use `inputId` or custom components with the `components` prop instead.
      * Custom attributes for the input (inside the Select).
@@ -269,6 +271,7 @@ class Select extends React.Component<Props, State> {
     delimiter: ',',
     disabled: false,
     displayKey: 'label',
+    ignoreAccents: true,
     inputProps: undefined,
     matchProp: 'any',
     multi: false,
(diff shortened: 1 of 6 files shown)


mpfz0r:
@bernd do you mind taking that for a quick spin? I don't have that many fields in my setup that it would make a big difference

bernd:
> @bernd do you mind taking that for a quick spin? I don't have that many fields in my setup that it would make a big difference

Will do! :+1:

graylog2-web-interface/src/components/common/Select.jsx
```diff
@@ -289,7 +293,7 @@ class Select extends React.Component<Props, State> {
     };
   }
 
-  componentWillReceiveProps = (nextProps: Props) => {
+  componentWillReceivePropsProps = (nextProps: Props) => {
```
edmundoa, line 296: I was just looking around and this didn't look right :)

dennisoelkers: Hah, good catch! :)

dennisoelkers: ✅

PR #8084, opened 2020-05-11
Support editing title, description and summary of a saved search

linuspahl:
This PR implements the "Edit metadata" action for the saved searches search bar menu, which opens a modal to edit the title, description and summary.

It also:
* renames the "Edit" dashboard search bar action to "Edit metadata"
I think "Edit" is too generic. From a user's perspective: What is going to happen, will a click on the button display the search / widgets in an edit mode?
* adjusts the sidebar view description.
For an ad hoc search, an info will be displayed to save the search as a saved search / dashboard, to add a description / summary. On a dashboard / saved search the view description will include a view type specific placeholder, if a summary or description is not defined.

Regarding the tests for `SavedSearchControls`, I've tried to test if the edit modal opens, when clicking on "Edit Metadata", but I could not easily test this case with `enzyme`. Out of Interest I would like to have another look with someone else.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
+++ b/graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
@@ -43,10 +43,10 @@ const ViewDescription = ({ results, viewMetadata }: Props) => {
 
   return (
     <>
+      {resultsSection}
       <Section>
         {viewMetadata.summary || <i>No {viewTypeLabel} summary.</i>}
       </Section>
-      {resultsSection}
       <Section>
         {viewMetadata.description || <i>No {viewTypeLabel} description.</i>}
       </Section>
--- a/graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
+++ b/graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
@@ -1,46 +1,55 @@
 // @flow strict
-import React from 'react';
+import * as React from 'react';
+import { useContext } from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 
 import type { ViewMetaData } from 'views/stores/ViewMetadataStore';
+import QueryResult from 'views/logic/QueryResult';
 
-import IfDashboard from 'views/components/dashboard/IfDashboard';
+import ViewTypeLabel from 'views/components/ViewTypeLabel';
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/views/components/views/ViewPropertiesModal.jsx
```diff
@@ -23,7 +24,8 @@ export default class ViewPropertiesModal extends React.Component {
     };
   }
 
-  componentWillReceiveProps(nextProps) {
+  // eslint-disable-next-line camelcase
```
kmerz, line 27: I does not have to be in this PR, but wouldn't it be wiser to make a functional component of this component, instead of adding a UNSAFE_

graylog2-web-interface/src/views/components/views/ViewPropertiesModal.jsx
```diff
@@ -23,7 +24,8 @@ export default class ViewPropertiesModal extends React.Component {
...
+  UNSAFE_componentWillReceiveProps(nextProps) {
     const { title } = this.props;
     const { view } = this.state;
     if (title !== nextProps.title || !isEqual(view, nextProps.view)) {
```
kmerz, line 31: When we turn this into a functional component, we should do something better then comparing the whole view.
1. it is not very efficient
2. we do not really care about the whole view if I am not wrong

graylog2-web-interface/src/views/components/sidebar/ViewDescription.jsx
```diff
@@ -1,46 +1,55 @@
...
+  return (
+    <>
+      <Section>
+        {viewMetadata.summary || <i>No {viewTypeLabel} summary.</i>}
+      </Section>
+      {resultsSection}
       <Section>
-        <SearchResultOverview results={results} />
+        {viewMetadata.description || <i>No {viewTypeLabel} description.</i>}
       </Section>
```
kmerz, line 52: The order of Sections does not make much sense to me:
Graylog

I would rather have the description and summary together and the result in the beginning or the end.

linuspahl:
@kmerz I like the suggestion with the changed order and moved the search execution info above the summary and description.

I am also a fan of transforming `ViewPropertiesModal` to a functional component and added this task to my refactoring list.

PR #8096, opened 2020-05-12
Don't throw error in event wizard when query has no result

linuspahl:
There is currently a bug when the search query in the event wizard does not return any result.

event-bug

This PR is fixing the `cannot read property .length of undefined` error, but I am not sure why this error did not occurred before. This bug does not exist in 3.3.0-beta.1.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterPreview.jsx
+++ b/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterPreview.jsx
@@ -33,8 +33,8 @@ class FilterPreview extends React.Component {
     });
   };
 
-  renderSearchResult = (searchResult) => {
-    if (searchResult.messages.length === 0) {
+  renderSearchResult = (searchResult = {}) => {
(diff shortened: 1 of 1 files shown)


edmundoa:
@linuspahl Playing with it, it looks like this situation happens when the search execution returned an error (in this case, the query uses a colon but no value follows it). Using the same query string in 3.2 works and displays an error.
Screenshot 2020-05-12 at 11 56 38

Do you know if anything changed in how the search backend is reporting errors? This is how we detected query errors before: https://github.com/Graylog2/graylog2-server/blob/d2715cce9cc2e786a6ffa4a1eb6e8d2b4a5b3e6f/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterPreviewContainer.jsx#L94

linuspahl:
@edmundoa The error message changed, but the structure should be the same, with this bugfix I am seeing the following error message:
image

edmundoa:
@linuspahl alright, will take a look in a bit :)

PR #8112, opened 2020-05-13
Add source streams to the default email notification template

mpfz0r:
This might be helpful for some users


bernd:
@mpfz0r Can you please merge in master? The build fails due to missing version bumps.

PR #8120, opened 2020-05-14
use JsonUnit to ignore the order of fields in expected JSON string

### LINKED ISSUE #7445: Test can fail if serialization result is in a different order

testfixer:
This PR fixes #7445
## Description

It uses `JsonUnit` to make the test is more stable.
## Motivation and Context

The test method uses `jackson`to serialize objects into JSON and assert the serialized results with JSON string. However, the order of serialized JSON is not guaranteed so test

This PR proposes to use `JsonUnit` to compare JSON string without taking the order into consideration.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -667,6 +667,12 @@
             <version>1.5.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.javacrumbs.json-unit</groupId>
+            <artifactId>json-unit-assertj</artifactId>
+            <version>2.6.1</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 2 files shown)


testfixer:
@dennisoelkers  As you stated in #7446, JSONAssert#assertEquals is less strict in its non-strict mode. 
Instead, I found JsonUnit (https://github.com/lukas-krecan/JsonUnit) that is better for comparing JSON strings by ignoring the order, especially when already using AssertJ.  The test will generate a deterministic result now.

dennisoelkers:
Hey @testfixer!

Thanks for submitting this. I will take a look soon.

dennisoelkers:
Hey @testfixer,

right now there is no need to be tolerant towards the order of elements in the test, so we would like to avoid introducing another dependency. Thanks for your contribution anyway!

PR #8124, opened 2020-05-14
Introduce pluggable authorization realms

mpfz0r:
This gives us the possibility to extend the
authorization realms used by Shiro.

Refactor DefaultSecurityManagerProvider to avoid unneeded
dependcies. We can get the RootAccountRealm from the
OrderedAuthenticatingRealms directly.


mpfz0r:
This is for after 3.3. Removed "ready for review" label

PR #7836, opened 2020-04-07
Create separate context for current user

linuspahl:
This PR creates a context for the current user. Its value is being provided by the `CurrentUserProvider` which connects to the `CurrentUserStore`.

The `CurrentUserProvider` is being used in the `LoggedInPage` (wrapper for all pages which are accessible when being logged in), this way we don't need to connect the `CurrentUserStore` in `App.jsx`, `ViewActionsMenu.jsx`, `UserTimezoneTimestamp.jsx`, `SavedSearchControls.jsx`, `SearchResultOverview.jsx`, `ShareViewModal.jsx`, `View.jsx`, `XYPlot.jsx` 

This PR also implements the timezone configured in the `AppConfig` as a fallback, if the user timezone is not defined.

## Motivation and Context
We should reduce the times we call `connect` -> `render` and `connect` -> `shouldComponentUpdate`. Currently `render` is being called over 1000 times and `shouldComponentUpdate` over 600 times on the default search page.

Just removing `connect` from `UserTimezoneTimestamp` and `CustomHighlighting` (https://github.com/Graylog2/graylog2-server/pull/8276) reduces the calls of `connect` -> `render` to ~50 and `connect` -> `shouldComponentUpdate` to ~ 20 times.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #8125, opened 2020-05-14
Render event annotations with timezone setting of the user

kmerz:
## Motivation
Prior to this change, the events were not annotated in the histogram
if the user had a different timezone than UTC. Because the anotations
would be set with UTC and not the user set timezone.

## Description
This change will convert the chartdata timestamps to the users setting.

## How Has This Been Tested?
- Changed timezone of the user to CET
- Caused some alerts and saw them rendered correctly with adjusted timezone.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
```diff
@@ -31,17 +35,25 @@ export type Shapes = Array<Shape>;
 const eventsDisplayName = 'Alerts';
 const defaultColor = '#d3d3d3';
 
+const formatTimestamp = (timestamp, tz = 'UTC'): string => {
+  // the `true` parameter prevents returning the iso string in UTC (http://momentjs.com/docs/#/displaying/as-iso-string/)
+  return moment(timestamp).tz(tz || 'UTC').toISOString(true);
```
linuspahl, line 40: You can use `tz ?? 'UTC'` here.

Just to let you know, I started replacing 'UTC'  in another PR (https://github.com/Graylog2/graylog2-server/pull/7836) with `AppConfig.rootTimeZone();`. I thought about replacing it here as well, but we did not merged the mentioned PR and use `UTC` in comparable cases.

graylog2-web-interface/src/views/logic/searchtypes/events/EventHandler.js
```diff
@@ -31,17 +35,25 @@ export type Shapes = Array<Shape>;
...
+};
```
linuspahl, line 41: You can use `tz ?? 'UTC'` here.

Just to let you know, I started replacing `'UTC'` in another PR (#7836) with `AppConfig.rootTimeZone();`. I thought about replacing it here as well, but we did not merged the mentioned PR and use UTC in comparable cases.

kmerz: I would suggest to replace all occurrences of `'UTC'` together to prevent having different timezone again.

PR #8135, opened 2020-05-15
Fix search widget grid width when its content results in a scrollbar

linuspahl:
There is currently a visual bug, when the widget grid content results in a scrollbar:
image

This bug is appearing, because the widget grid changes its initially calculated width, when its content results in a scrollbar.

This PR is fixing the problem by reimplementing the `SizeMe` component for the `WidgetGrid`, which got removed with https://github.com/Graylog2/graylog2-server/pull/7833
We removed the `SizeMe` container because it fixed a bug with the worldmap widget viewport, but  we've implemented another solution for this problem in the meantime.

While testing this change I found the following edge case. When opening a dashboard page with a worldmap widget and no scrollbar after being on a page with a scrollbar, the widget has a bigger or smaller (depending on the bowser) white padding on the right side.

## How Has This Been Tested?
Tested with all major browsers, IE11 and different widgets

## Types of changes

## Checklist:


kmerz:
\rebase

PR #8136, opened 2020-05-15
Add support for more verbose authentication failure

thll:
## Description

With this change, an authenticating realm may throw an
`AuthenticationServiceUnavailableException` which will be retained if
none of the realms can successfully authenticate the user. This will
then be indicated to the API client with a non-successful status code
and response message.

## Motivation and Context

When an authenticating realm is using an external service, there was no
way to indicate that authentication failed in case the external service
was unavailable at the moment of checking the credentials. It could only
indicate that authentication was unsuccessful in general.
Therefore, in the unavailability case, the user would probably assume
that the credentials were wrong.

## How Has This Been Tested?

Testing this requires implementing an authenticating realm which throws the appropriate exception.

## Types of changes

## Checklist:


thll:
This is a draft only because it should only be merged after `3.3` was released.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -124,11 +126,15 @@ public JsonNode newSession(@Context ContainerRequestContext requestContext,
...
-            throw new NotAuthorizedException("Invalid credentials.", "Basic realm=\"Graylog Server session\"");
+        try {
+            Optional<Session> session = sessionCreator.create(sessionId, host, authToken);
+            if (session.isPresent()) {
+                return sessionResponseFactory.forSession(session.get());
+            } else {
+                throw new NotAuthorizedException("Invalid credentials.", "Basic realm=\"Graylog Server session\"");
+            }
+        } catch (AuthenticationServiceUnavailableException e) {
+            throw new ServiceUnavailableException(e.getMessage());
```
mpfz0r, line 137: I think we shouldn't expose too many details to unauthenticated users.
They shouldn't be able to discover which auth backends we are using.
How about
```suggestion
            throw new ServiceUnavailableException("Authentication service unavailable");
```

thll:
Thanks for the review @mpfz0r!

PR #8145, opened 2020-05-18
Add input selection for StreamRuleForm

kmerz:
## Description
Add a new StreamRuleType which matches Messages based on the input they are coming from.
The user can now select a input in the stream rule. The new stream rule match will then look if
the message has a `gl2_source_input_id` and will compare it to the user selected input id.

## Motivation and Context
A often recurring use case for me is, that I create a new stream for a new input, right now I needed
to make a `match exactly` rule, select the `gl2_source_input_id` and copy the `input id` into the value. This is error prone and take to long.

## How Has This Been Tested?
- Created new input and stream
- selected new input for the stream rule
- checked that new messages were routed to the new stream coming from that input

Graylog - Rules of Stream raw time

## TBD
- I tested it with events as well but a second glance would be welcoming. The problem is, that
  events do not have a `gl2_source_input_id`.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -135,10 +135,10 @@ public class StreamRouterEngine {
         this.rulesList.addAll(alwaysMatchRules);
         this.rulesList.addAll(presenceRules);
         this.rulesList.addAll(exactRules);
+        this.rulesList.addAll(matchInputRules);
         this.rulesList.addAll(greaterRules);
         this.rulesList.addAll(smallerRules);
         this.rulesList.addAll(containsRules);
-        this.rulesList.addAll(matchInputRules);
         this.rulesList.addAll(regexRules);
     }
 
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -54,7 +54,7 @@ import java.util.concurrent.TimeUnit;
 public class StreamRouterEngine {
     private static final Logger LOG = LoggerFactory.getLogger(StreamRouterEngine.class);
 
-    private final EnumSet<StreamRuleType> ruleTypesNotNeedingFieldPresence = EnumSet.of(StreamRuleType.PRESENCE, StreamRuleType.EXACT, StreamRuleType.REGEX, StreamRuleType.ALWAYS_MATCH, StreamRuleType.CONTAINS);
+    private final EnumSet<StreamRuleType> ruleTypesNotNeedingFieldPresence = EnumSet.of(StreamRuleType.PRESENCE, StreamRuleType.EXACT, StreamRuleType.REGEX, StreamRuleType.ALWAYS_MATCH, StreamRuleType.CONTAINS, StreamRuleType.MATCH_INPUT);
     private final List<Stream> streams;
     private final StreamFaultManager streamFaultManager;
     private final StreamMetrics streamMetrics;
@@ -90,6 +90,7 @@ public class StreamRouterEngine {
         final List<Rule> smallerRules = Lists.newArrayList();
         final List<Rule> regexRules = Lists.newArrayList();
         final List<Rule> containsRules = Lists.newArrayList();
+        final List<Rule> matchInputRules = Lists.newArrayList();
 
         for (Stream stream : streams) {
             for (StreamRule streamRule : stream.getStreamRules()) {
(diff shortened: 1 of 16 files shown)


graylog2-server/src/main/java/org/graylog2/streams/matchers/MatchInput.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class MatchInput implements StreamRuleMatcher {
+
+    @Override
+    public boolean match(Message msg, StreamRule rule) {
+       if(msg.getField(Message.FIELD_GL2_SOURCE_INPUT) == null) {
+           return rule.getInverted();
+       }
+        final String value = msg.getField(Message.FIELD_GL2_SOURCE_INPUT).toString();
+        return rule.getInverted() ^ value.trim().equals(rule.getValue());
```
dennisoelkers, line 30: IMO we can do a case-insensitive check in here, as afaics input IDs should be unique regarding the case.

graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
```diff
@@ -135,6 +139,7 @@ public StreamRouterEngine(@Assisted List<Stream> streams,
         this.rulesList.addAll(smallerRules);
         this.rulesList.addAll(containsRules);
         this.rulesList.addAll(regexRules);
+        this.rulesList.addAll(matchInputRules);
```
dennisoelkers, line 142: The rules list should be ordered by matching complexity of the rules (ascending), so the match input rules should be before or after `exactRules`.

kmerz: I moved it below `contains` rule. Since it is just that.

graylog2-server/src/main/java/org/graylog2/streams/matchers/MatchInput.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.streams.matchers;
+
+import org.graylog2.plugin.Message;
+import org.graylog2.plugin.streams.StreamRule;
+
+public class MatchInput implements StreamRuleMatcher {
```
dennisoelkers, line 22: The naming scheme for matchers is `<Type>Matcher`, so it should be `InputMatcher` for this one.

graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
dennisoelkers: It should be after the `exactRules`, as it does not do a `contains` check, but an `equals` check.

graylog2-server/src/test/java/org/graylog2/streams/matchers/MatchInputTest.java
```diff
@@ -0,0 +1,110 @@
...
+        msg.addField(Message.FIELD_GL2_SOURCE_INPUT, "message-id-beef");
+
+        StreamRuleMatcher matcher = getMatcher(rule);
+        assertFalse(matcher.match(msg, rule));
+    }
+
+    @Test
+    public void testUnsuccessfulMatchInverted() {
+        StreamRule rule = getSampleRule();
+        rule.setValue("message-id-dead");
```
dennisoelkers, line 80: Instead of `message-id-...` we should use `input-id-...` in the tests to make the intention clearer.

graylog2-web-interface/src/components/streamrules/HumanReadableStreamRule.jsx
```diff
@@ -23,7 +20,21 @@ class HumanReadableStreamRule extends React.Component {
     return undefined;
   };
 
+  _inputName = (inputId) => {
```
dennisoelkers, line 23: The function name is giving the wrong impression of what it's doing. It should be something like `findInput`.

graylog2-web-interface/src/components/streamrules/HumanReadableStreamRule.jsx
```diff
@@ -23,7 +20,21 @@ class HumanReadableStreamRule extends React.Component {
...
+    return inputs.find((input) => input.id === inputId);
+  }
+
   _formatRuleValue = (streamRule) => {
+    if (String(streamRule.type) === String(this.MATCH_INPUT)) {
+      const input = this._inputName(streamRule.value);
+
+      if (input) {
+        return `${input.title} (${input.name}: ${input.id})`;
+      }
```
dennisoelkers, line 35: Right now, if the input does not exist/is not started, only the id is shown. IMO we should also show an indicator, something like `<deleted input>`, like we do in message details.

graylog2-web-interface/src/components/streamrules/StreamRulesEditor.jsx
```diff
@@ -46,7 +46,13 @@ class StreamRulesEditor extends React.Component {
     index: '',
   }
 
-  state = {};
+  constructor() {
+    super();
```
kyleknighted, line 50: I think `constructor` needs to pass `props` to `super` or else it could lead to [bugs](https://reactjs.org/docs/react-component.html#constructor)
```suggestion
  constructor(props) {
    super(props);
```

graylog2-web-interface/src/components/streams/Stream.jsx
```diff
@@ -78,10 +78,19 @@ class Stream extends React.Component {
...
+    this.setState({ showStreamRuleForm: false });
+  };
+
+  _openStreamRuleForm = () => {
+    this.setState({ showStreamRuleForm: true });
+  };
+
+  _onDelete= (stream) => {
     // eslint-disable-next-line no-alert
     if (window.confirm('Do you really want to remove this stream?')) {
```
kyleknighted, line 95: Could we use `components/bootstrap/BootstrapModalConfirm.jsx` instead of `window.confirm`?

graylog2-web-interface/src/components/streamrules/HumanReadableStreamRule.jsx
kmerz: Best idea ever! Now you can remove the rules with deleted input!

PR #8168, opened 2020-05-20
Allow integration tests for multiple Elasticsearch versions

alex-konn:
## Description
This PR allows adding integration tests that iterate over multiple Elasticsearch versions with duplicating tests.
It leverages JUnit 5 [`@TestTemplate`s](https://junit.org/junit5/docs/snapshot/user-guide/#extensions-test-templates). The idea is that version-specific dependencies can be injected into the test template. It's not yet clear, what our version-independent interfaces, which we will introduce for 4.0, will look like. So for now only a version string and an instance of a `Client` that is configured with an `ElasticsearchInstance` of the requested version are injected. We should adapt `VersionInvocationContext` as we create new interfaces so that instances can be injected into tests.

## Motivation and Context
1. We are introducing capabilities for Graylog to support multiple Elasticsearch versions and want to cover these capabilities with tests without duplicating too much test code
2. We want to avoid breaking version-dependent functionality in the future. It's too much effort to manually test new features and fixes with multiple Elasticsearch versions. 

## How Has This Been Tested?
Local execution of new tests

## Types of changes

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/testing/completebackend/GraylogBackendExtension.java
+++ b/graylog2-server/src/test/java/org/graylog/testing/completebackend/GraylogBackendExtension.java
@@ -33,6 +33,7 @@ import org.slf4j.LoggerFactory;
 import java.util.concurrent.TimeUnit;
 
 import static io.restassured.http.ContentType.JSON;
+import static org.graylog.testing.junit5utils.Annotations.annotationFrom;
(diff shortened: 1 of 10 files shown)


dennisoelkers:
We are working around this right now in #9288 by adding base/sub classes. This PR/issue will be blocked until after 4.0, when we can continue working on this.

dennisoelkers:
Superseded by #11011.

PR #8236, opened 2020-05-29
Move widget to tab

### LINKED ISSUE #8152: Copy/move tabs between dashboards
## Expected Behavior
It should be possible to to copy or move a tab from one existing dashboard to another.

## Current Behavior
Not possible. Tabs and widgets have to be recreated.

## Context
A community user came up with the idea here:
https://community.graylog.org/t/combine-dashboards/15491

## Your Environment

* Graylog Version: 3.2.4

kmerz:
## Description
Add new menu item to widget menu: `Move to Page`. When clicking the user will see a modal
where he can select the target page. Optionally he can decide to keep a copy of the widget
on the current page. After selecting the target page he clicks on `Select`. The widget will
be moved to the new page and the active page will be set to the target query.

## Motivation and Context
A user was able to copy a widget from a search to the first page of a dashboard. But the widget
was stuck now there. There was no possibility to move the widget to a different page of the dashboard.

## How Has This Been Tested?
- Create a dashboard
- create a widget on the first page
- move the widget to a different page

Fixes #8152 

Graylog (3)
Graylog (2)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/MoveWidgetToTab.jsx
+++ b/graylog2-web-interface/src/views/components/widgets/MoveWidgetToTab.jsx
@@ -1,0 +1,56 @@
+// @flow strict
+import React, { useState } from 'react';
+import { Map } from 'immutable';
+import { Button, ListGroup, ListGroupItem, Modal } from 'components/graylog';
+
+import View from 'views/logic/views/View';
+
+type Props = {
+  view: View,
+  widgetId: string,
+  onCancel: () => void,
+  onSubmit: (widgetId: string, selectedTab: ?string) => void,
+};
+
+type TabEntry = { id: string, name: string };
+
+const _tabList = (view: View): Array<TabEntry> => {
+  const queryIds = Object.keys(view.state.toObject());
+  return queryIds.map((queryId, index) => {
+    const tabTitle = view.state.get(queryId).titles.get('tab', Map({ title: `Page#${index + 1}` }));
+    return ({ id: queryId, name: tabTitle.get('title') });
+  });
+};
+
+const MoveWidgetToTab = ({ view, onCancel, onSubmit, widgetId }: Props) => {
+  const [selectedTab, setSelectedTab] = useState(null);
+  const list = _tabList(view);
+
+  const tabList = list.map(({ id, name }) => (
+    <ListGroupItem header={name}
+                   onClick={() => setSelectedTab(id)}
+                   active={id === selectedTab}
+                   key={id} />
+  ));
+  const renderResult = list && list.length > 0
+    ? <ListGroup>{tabList}</ListGroup>
+    : <span>No dashboards found</span>;
+
+  return (
+    <Modal show>
+      <Modal.Body>
+        {renderResult}
+      </Modal.Body>
+      <Modal.Footer>
+        <Button bsStyle="primary"
+                disabled={selectedTab === null}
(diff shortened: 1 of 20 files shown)


graylog2-web-interface/src/views/components/widgets/MoveWidgetToTab.jsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+type TabEntry = { id: string, name: string };
+
+const _tabList = (view: View): Array<TabEntry> => {
+  const queryIds = Object.keys(view.state.toObject());
+  return queryIds.map((queryId, index) => {
+    const tabTitle = view.state.get(queryId).titles.get('tab', Map({ title: `Page#${index + 1}` }));
+    return ({ id: queryId, name: tabTitle.get('title') });
+  });
```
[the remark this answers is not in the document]
kmerz: > We should either sort the list alphabetically or use the order defined by the dashboard

As you see I need to "make up" the names of the page titles on the fly. That means it is in order defined by the dashboard.

graylog2-web-interface/src/views/components/widgets/MoveWidgetToTab.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const [keepCopy, setKeepCopy] = useState(false);
+
+  const list = _tabList(view);
+
+  const tabList = list.map(({ id, name }) => (
+    <ListGroupItem header={name}
+                   onClick={() => setSelectedTab(id)}
+                   active={id === selectedTab}
+                   key={id} />
+  ));
```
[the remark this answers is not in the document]
kmerz: I now used the `children` way since it looks much nicer than the header way. Thanks for pointing that out.

graylog2-web-interface/src/views/components/widgets/MoveWidgetToTabModal.jsx
```diff
@@ -0,0 +1,64 @@
...
+    return ({ id: queryId, name: tabTitle });
+  });
+};
+
+const MoveWidgetToTabModal = ({ view, onCancel, onSubmit, widgetId }: Props) => {
+  const [selectedTab, setSelectedTab] = useState(null);
+  const [keepCopy, setKeepCopy] = useState(false);
+  const { id: activeQuery } = useStore(CurrentQueryStore);
+
+  const list = _tabList(view).filter(({ id }) => id !== activeQuery);
```
linuspahl, line 34: Regarding the sort, it does not look correct to me, compared with the dashboard:
image

One solution is using the order / values provided by the `QueryIdsStore`.

linuspahl: With the last review I suggested to remove the current tab from the list, but after having another look I prefer to just disable the active tab. It has the benefits, that the user always sees the "same" list no matter on which tab he opens the modal and it is easier to find the surrounding tabs.

kmerz: Regarding the sort: https://github.com/Graylog2/graylog2-server/issues/8284 there is no order :( Which we need to fix before this PR I think now. :(

Regarding disable: I think you were right before, the user does not expect to find the current tab here. Therefore I would leave it out to reduce the visual noise. Let's discuss it later in or after the meeting.

PR #8294, opened 2020-06-05
Add option to display the search sidebar inline

linuspahl:
This PR adds an option (layer icon next to the search title) to display the sidebar inline next to the search results:

image

To unify both sidebar versions (inline and overlay) we had to adjust the sidebar layout. This PR also adds more content for the sidebar sections, to make use of the available space.

## Types of changes


graylog2-web-interface/src/views/components/sidebar/SectionInfo.jsx
```diff
@@ -0,0 +1,9 @@
+// @flow strict
+import styled, { type StyledComponent } from 'styled-components';
+import { type ThemeInterface } from 'theme';
+
+const Title: StyledComponent<{}, ThemeInterface, HTMLParagraphElement> = styled.p(({ theme }) => `
+  color: ${theme.colors.gray[60]};
```
dennisoelkers, line 6: IMO this is too light for anyone who is even slightly reading impaired. We should not use more than 50 for text, in this case 40 is even better.

graylog2-web-interface/src/views/components/sidebar/ContentColumn.jsx
```diff
@@ -0,0 +1,162 @@
...
+            <div>
+              <Header title={title}>
+                <CenterVertical>
+                  <Title onClick={closeSidebar}>{title}</Title>
+                </CenterVertical>
+                <CenterVertical>
+                  <OverlayToggle sidebarIsInline={sidebarIsInline}>
+                    <IconButton onClick={() => toggleSidebarPinning(searchPageLayout)}
+                                title={`Display sidebar ${sidebarIsInline ? 'as overlay' : 'inline'}`}
+                                name="layer-group" />
```
dennisoelkers, line 145: The icon is not ideal, it does not really give away what it does. IMO the pin icon was better.

graylog2-web-interface/src/views/components/sidebar/ContentColumn.jsx
```diff
@@ -0,0 +1,162 @@
...
+  const sidebarIsInline = config?.sidebar.isInline;
+  const newLayoutConfig = {
+    ...config,
+    sidebar: { isInline: !sidebarIsInline },
+  };
+  setConfig(newLayoutConfig);
+};
+
+const sidebarTitle = (viewMetadata: ViewMetadata, viewType: ?ViewType) => {
+  if (!viewMetadata.id) {
```
dennisoelkers, line 121: We should use `isNew` from `ViewStore` instead.

graylog2-web-interface/src/views/components/sidebar/ContentColumn.jsx
```diff
@@ -0,0 +1,162 @@
...
+    return 'Ad Hoc Search';
```
dennisoelkers, line 122: `Ad Hoc Search` is not a common way to describe that the search is new/untitled. I would prefer something more common. like `New Search` or `Untitled Search`. (I do remember why we changed from the first to the second option, but I am not convinced that someone could still mistake the heading for an action, the way it looks now)

graylog2-web-interface/src/views/components/sidebar/highlighting/HighlightingRules.jsx
```diff
@@ -1,23 +1,24 @@
...
-  margin-bottom: 10px;
-`;
-
 const HighlightingRules = () => {
   const rules = useContext(HighlightingRulesContext) ?? [];
   return (
     <>
-      <Headline>Highlighting</Headline>
+      <SectionInfo>
+        Search terms and field values can be highlighted. The search term highlighting can be configured in the graylog server config.
```
dennisoelkers, line 18:
```suggestion
        Search terms and field values can be highlighted. Highlighting your search query in the results can be enabled/disabled in the graylog server config.
```

graylog2-web-interface/src/views/components/sidebar/highlighting/HighlightingRules.jsx
```diff
@@ -1,23 +1,24 @@
...
+        A field value can be highlighted by clicking on &quot;Highlight this value&quot; in its context menu.
```
dennisoelkers, line 19:
```suggestion
        Any field value can be highlighted by clicking on the value and selecting &quot;Highlight this value&quot;.
```

graylog2-web-interface/src/views/components/sidebar/SectionInfo.jsx
linuspahl: That is true, I am now using the default text color.

graylog2-web-interface/src/views/components/sidebar/ContentColumn.jsx
linuspahl: Using `isNew` here would lead to a slightly different behaviour. We would never display the title `Untitled Dashboard`. Right now this title is being display when a user creates a new dashboard (`/dashboards/new`).

Do you prefer to display the title for the default search in this case as well?

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.jsx
```diff
@@ -0,0 +1,27 @@
+// @flow strict
+import * as React from 'react';
+import { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import SearchPageLayoutContext from './SearchPageLayoutContext';
+
+const defaultLayoutConfig = {
+  sidebar: {
+    isInline: false,
```
kyleknighted, line 10: Could we load the default from localStorage or user preferences?

linuspahl: Yes, the plan is to save this information as a user preference in another iteration.

graylog2-web-interface/src/views/components/sidebar/ContentColumn.jsx
dennisoelkers: Imo these are two separate concerns. One is to find out if we have a new, unsaved search/dashboard, the other is to format the title. In my opinion, we should use `isNew` to determine the first and use `viewType` to format a title according to the type.

linuspahl: Thanks for the helpful feedback, I've updated the code. I also change the default title for unsaved views to `Unsaved ...` in the same step.

PR #8298, opened 2020-06-08
Update font awesome icon names regarding font awesome v5

linuspahl:
With https://github.com/Graylog2/graylog2-server/pull/7746 we've updated font awesome and implemented a fallback for old icon names. The fallback displays a deprecation warning in the console when using a deprecated name.

This PR updates the existing Icon names, based on https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/components/common/icon-fallback.js

There are many linter warnings, which can't be fixed easily, due to the amount of changes files.

## Motivation and Context
Reducing the amount of warnings in the console.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchControls.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchControls.jsx
@@ -226,7 +226,6 @@ class SavedSearchControls extends React.Component<Props, State> {
     );
 
     const loaded = (view && view.id);
-    const savedSearchStyle = loaded ? 'star' : 'star-o';
     let savedSearchColor: string = '';
     if (loaded) {
       savedSearchColor = dirty ? theme.colors.variant.warning : theme.colors.variant.info;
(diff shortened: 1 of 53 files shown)


graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchControls.jsx
```diff
@@ -258,21 +257,21 @@ class SavedSearchControls extends React.Component<Props, State> {
             <ButtonGroup>
               <>
                 <Button title={title} ref={(elem) => { this.formTarget = elem; }} onClick={this.toggleFormModal}>
-                  <Icon style={{ color: savedSearchColor }} name={savedSearchStyle} /> Save
+                  <Icon style={{ color: savedSearchColor }} name="star" /> Save
```
kyleknighted, line 260: Do we need both a solid and outlined `star` here?

https://fontawesome.com/icons/star?style=regular
https://fontawesome.com/icons/star?style=solid

linuspahl: Thanks again for the hint, I've added the ability to outline icons and updated the icons which had `-o` in their name.

I've changed the `Icon` props interface in the same step, I think it is now easier to understand, but it would be great to hear your opinion. There are no changes needed in our plugins.

Issue #7303, opened 2020-01-27
Directly loading New Dashboard outputs propType error


kyleknighted:
## Current Behavior
Navigating directly to http://127.0.0.1:8080/dashboards/new outputs the follow console log

```
Warning: Failed prop type: Invalid prop `view` of type `View` supplied to `ViewActionsMenu`, expected instance of `View`.
    in ViewActionsMenu (created by withRouter(ViewActionsMenu))
    in withRouter(ViewActionsMenu) (created by ConnectStoresWrapper[withRouter(ViewActionsMenu)] stores=metadata,view,currentUser)
    in ConnectStoresWrapper[withRouter(ViewActionsMenu)] stores=metadata,view,currentUser (created by DashboardSearchBar)
    in div (created by DashboardSearchBar)
    in div (created by Col)
    in Col (created by DashboardSearchBar)
    in div (created by Row)
    in Row (created by DashboardSearchBar)
    in form (created by DashboardSearchBar)
    in div (created by Col)
    in Col (created by DashboardSearchBar)
    in div (created by Row)
    in Row (created by DashboardSearchBar)
    in span (created by ScrollToHint)
    in ScrollToHint (created by DashboardSearchBar)
    in DashboardSearchBar (created by ConnectStoresWrapper[DashboardSearchBar] stores=currentQuery)
    in ConnectStoresWrapper[DashboardSearchBar] stores=currentQuery
    in Unknown (created by ConnectStoresWrapper[Unknown/Anonymous] stores=searchMetadata,executionState,configurations)
    in ConnectStoresWrapper[Unknown/Anonymous] stores=searchMetadata,executionState,configurations (created by Context.Consumer)
    in IfDashboard (created by Context.Consumer)
    in IfInteractive (created by Context.Consumer)
    in div (created by Grid)
    in Grid (created by Context.Consumer)
    in StyledComponent (created by ExtendedSearchPage__SearchGrid)
    in ExtendedSearchPage__SearchGrid (created by Context.Consumer)
    in div (created by Context.Consumer)
    in StyledComponent (created by ExtendedSearchPage__SearchArea)
    in ExtendedSearchPage__SearchArea (created by Context.Consumer)
    in div (created by Context.Consumer)
    in StyledComponent (created by ExtendedSearchPage__GridContainer)
    in ExtendedSearchPage__GridContainer (created by Context.Consumer)
    in Unknown (created by ConnectStoresWrapper[Unknown/Anonymous] stores=view)
[...]

PR #8300, opened 2020-06-08
Fixing `ViewActionsMenu` and `EntityListItem` prop type errors

linuspahl:
This PR is fixing the prop type errors for the `EntityListItem` prop `contentRow`. It also avoids throwing a prop type error for the `ViewActionsMenu` `view` prop, by removing the `Proptypes.instanceOf` check.

Fixes: #7303

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/ViewActionsMenu.jsx
+++ b/graylog2-web-interface/src/views/components/ViewActionsMenu.jsx
@@ -106,7 +106,7 @@ ViewActionsMenu.propTypes = {
   metadata: PropTypes.shape({
     undeclared: ImmutablePropTypes.Set,
   }).isRequired,
-  view: PropTypes.instanceOf(View).isRequired,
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/ViewActionsMenu.jsx
```diff
@@ -106,7 +106,7 @@ ViewActionsMenu.propTypes = {
   metadata: PropTypes.shape({
     undeclared: ImmutablePropTypes.Set,
   }).isRequired,
-  view: PropTypes.instanceOf(View).isRequired,
+  view: PropTypes.object.isRequired,
```
linuspahl, line 109: We talked about the unreliable `instanceof` check in the past, it would be great to find a reliable solution, but for now I've just typed the prop as an object, like we do it in many other places.

PR #8302, opened 2020-06-08
Log the rejected URL within RestrictToMasterFilter

mpfz0r:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/RestrictToMasterFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/RestrictToMasterFilter.java
@@ -32,7 +32,7 @@ public class RestrictToMasterFilter implements ContainerRequestFilter {
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/shared/security/RestrictToMasterFilter.java
```diff
@@ -32,7 +32,7 @@
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
-        LOG.warn("Rejected request that is only allowed against master nodes. Returning HTTP 403.");
+        LOG.warn("Rejected request to <{}> which is only allowed against master nodes.", requestContext.getUriInfo().getPath());
```
dennisoelkers, line 35: Can `requestContext.getUriInfo()` be `null` under any circumstances?

mpfz0r: This gets initialized in jersey's `ConterRequest` constructor. So I'd say this never happens.

PR #8310, opened 2020-06-09
Extract interface from `MoreIndices`/`ElasticsearchProbe` classes to encapsulate ES-specifics

dennisoelkers:
## Motivation and Context

In order to move all ES-specific implementation details to a separate module, this change is extracting an interface for these from the `MoreIndices` & `ElasticsearchProbe` classes.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/system/stats/elasticsearch/ElasticsearchProbe.java
```diff
@@ -16,107 +16,48 @@
...
 
     @Inject
-    public ElasticsearchProbe(JestClient jestClient, IndexSetRegistry indexSetRegistry) {
-        this.jestClient = jestClient;
+    public ElasticsearchProbe(IndexSetRegistry indexSetRegistry, ClusterAdapter clusterAdapter) {
         this.indexSetRegistry = indexSetRegistry;
+        this.clusterAdapter = clusterAdapter;
     }
 
     public ElasticsearchStats elasticsearchStats() {
```
alex-konn, line 40: I think it make sense to move this method into `ClusterAdapter` (or create a separate interface which `ClusterAdapterES6` implements?) and get rid of `ElasticsearchProbe` completely. Right now `ElasticsearchProbe#elasticsearchStats` calls 4 different methods from `ClusterAdapter`, 3 of which are only used here. Those could become private, if the method were moved, thus reducing coupling.

The moved method should get the indices as a parameter to avoid a dependency on `IndexSetRegistry`, i.e. something like this: `ClusterAdapter#statsForIndices(List<String> indices)`.
It would imply that `org.graylog2.system.stats.ClusterStatsService` would need the dependency on `IndexSetRegistry`, but that seems acceptable to me.

graylog2-server/src/main/java/org/graylog/events/indices/MoreIndices.java
```diff
@@ -16,50 +16,36 @@
...
 
-import static java.util.Objects.requireNonNull;
-import static org.graylog2.plugin.Tools.buildElasticSearchTimeFormat;
-import static org.joda.time.DateTimeZone.UTC;
-
 /**
  * This class contains indices helper for the events system.
  */
 @Singleton
 public class MoreIndices {
```
alex-konn, line 39: Wouldn't it make sense to rename this to `EventIndices` while we're at it? I find `MoreIndices` too vague for it's very specific purpose. `EventWriter` or `EventIndexer` would be even more specific. Same would go for `MoreIndicesAdapter` and `MoreIndicesAdapterES6`.

The method name `bulkIndex` on the other hand is unnecessarily specific. `write` would be a fine alternative imo.

graylog2-server/src/main/java/org/graylog/events/indices/MoreIndices.java
```diff
@@ -74,62 +60,26 @@ public void bulkIndex(List<EventWithContext> eventsWithContext) {
...
+                                if (index == null) {
+                                    LOG.warn("Couldn't find index set of stream <{}> for event <{}> (definition: {}/{})", streamId,
+                                            event.getId(), event.getEventDefinitionType(), event.getEventDefinitionId());
+                                }
+                                return index;
+                            })
+                            .filter(Objects::nonNull)
+                            .collect(Collectors.toSet());
+                    return indices.stream()
+                            .map(index -> new AbstractMap.SimpleEntry<>(index, event));
```
alex-konn, line 80: I would love it, if you could unpack this by extracting lambda blocks to private methods. Ideally we'd only have one line lambdas, which are much easier to read than these nested blocks.

PR #8311, opened 2020-06-09
Extracting interface from `IndexFieldTypePoller` to encapsulate ES-specifics.

dennisoelkers:
## Motivation and Context

In order to move all ES-specific implementation details to a separate module, this change is extracting an interface for these from the `IndexFieldTypePoller` class.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerAdapter.java
@@ -1,0 +1,10 @@
+package org.graylog2.indexer.fieldtypes;
+
+import com.codahale.metrics.Timer;
+
+import java.util.Map;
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerAdapter.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.graylog2.indexer.fieldtypes;
+
+import com.codahale.metrics.Timer;
+
+import java.util.Map;
+import java.util.Optional;
+
+public interface IndexFieldTypePollerAdapter {
+    Optional<Map<String, String>> pollIndex(String indexName, Timer pollTimer);
```
alex-konn, line 25: Why not use `IndexFieldTypesDTO` here instead of just `Map<String, String>`? I think it would communicate the method's purpose more clearly.

PR #8312, opened 2020-06-09
Retry messages if target index is read-only

bernd:
* Retry messages if target index is read-only (#8245)

* Draft solution for retrying index failures due to flood stage reached (WIP)

- Added additional retry loop in `Messages`. While it's not ideal to complicate the indexing code further, this error case can't be included in the existing retry logic, because the we have to modify the next request to only include items that have failed due to this particular error
- Added a simple test to `MockedMessagesTest`

* Add test verifying that other errors occurring in same chunk as index block errors are not retried

* Treat retried index block errors that fail differently as persistent failures

* Reuse `com.github.rholder.retry.WaitStrategy` for index block retries

* add license headers

* Keep using milliseconds-based waiting for `BULK_REQUEST_RETRYER`

...but go with seconds-based for index block retries.

- extracted a seconds-based wait strategy to a field in Messages
- add explanatory comment for why a multiplier of 500 works, which is unintuitive
- added separate unit tests to document how the waiting strategies actually work

* Cleaned up code based on @dennisoelkers' feedback

- Mark variables as final where possible
- Rename method from `retryOnlyIndexBlockItemsForever` to `retryIfIndexBlocksPresent`
- Log warning, if index blocks present
- Log info after retries have been successful
- Removed `org.jetbrains.annotations.NotNull` annotation

* Renamed private method `blockExecutionForAttempt` to `waitBeforeRetrying`

* Renamed private method `retryIfIndexBlocksPresent` back to `retryOnlyIndexBlockItemsForever`.

It does indicate more clearly that only index blocks are retied here and that retries will go on "forever", if the blocks are not resolved.

(cherry picked from commit e7045448dccfede6f1d7710ab8d29093f1096efe)

* fix formatting glitch

(cherry picked from commit ed8a80ae118b63fb247ab4722c83457e5da06538)

* Fix unconditional warning

This was an oversight in pr #8245, which led to tons of misleading
warnings.

(cherry picked from commit 4295704d47a23195e45b257b359577ecda787f24)
(cherry picked from commit 76dddea85eaa6d441a63c75ba739449048a11aa1)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/IndexBlockRetryAttempt.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/IndexBlockRetryAttempt.java
@@ -1,0 +1,69 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


bernd:
@dennisoelkers @alex-konn This is a backport to 3.2.

PR #8333, opened 2020-06-12
feat: split multiple gelf messages on new line

alekc:
## Description
See https://github.com/Graylog2/graylog2-server/pull/5924

## Motivation and Context
Being able to transmit multiple messages over http gelf input (with them being split by the newline). 
In this way it should be possible to connect a lot of different log emitters (i.e. fluentd) without major issues. 

## How Has This Been Tested?
Running it in a patched version, so far no issues has arised. 
The most notable behaviour is that now GelfMessage returns an arraylist of messages.
Tests are running fine. 

This should not affect preexisting customers since as far as I know gelf message (even if coming from udp/tcp) doesn't contain any newline in it. If thats not the case, potentially we could add a checkbox in input type configuration I suppose (not sure how the config part works in Graylog inputs).

## Types of changes

## Checklist:


alekc:
Upon further reflection, this approach doesn't give enough flexibility, closing it in favour of dedicated input/codec

PR #8336, opened 2020-06-12
Extracting interface from `ScrollResult` to encapsulate ES-specifics.

dennisoelkers:
## Motivation and Context

In order to move all ES-specific implementation details to a separate module, this change is extracting an interface for these from the ScrollResult class.

The `ScrollResultIT` integration test was moved to the storage module completely, as it currently tests only a detail specific to this implementation (validating that no jest-specific metadata fields are present in the result). If we add further test cases which test cross-implementation functionality, we can introduce an abstract base class.

## Types of changes

## Checklist:


graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/ScrollResultES6.java
```diff
@@ -0,0 +1,164 @@
...
+            search = initialResult;
+            hits = initialResult.getHits(Map.class, false).stream()
+                .map(hit -> ResultMessage.parseFromSource(hit.id, hit.index, (Map<String, Object>)hit.source))
+                .collect(Collectors.toList());
+            this.initialResult = null;
+        }
+
+        if (hits.size() == 0) {
+            // scroll exhausted
+            LOG.debug("[{}] Reached end of scroll results.", queryHash, getOriginalQuery());
```
alex-konn, line 105: I know this bug was in here before, but there's a placeholder missing in the format string. I think we should fix that.

PR #8340, opened 2020-06-15
Fix `ActionDropdown` toggle on child select, if child is wrapped in another element

linuspahl:
The `ActionDropdown` manipulates the `onSelect` prop of its children, to close the dropdown when `onSelect` is being executed. This is currently not working, when a child is wrapped in an element like `IfDashboard`.
With this PR we also apply this manipulation to the children of the `ActionDropdown` children, if the main child does not have an `onSelect` prop.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/common/ActionDropdown.test.tsx
+++ b/graylog2-web-interface/src/views/components/common/ActionDropdown.test.tsx
@@ -75,6 +75,27 @@ describe('ActionDropdown', () => {
     expect(wrapper).not.toContainMatchingElement('ul.dropdown-menu');
   });
 
+  it('closes menu when MenuItem with a parent element is clicked', () => {
+    const onSelect = jest.fn();
+    const wrapper = mount((
+      <ActionDropdown element={<div>Trigger!</div>}>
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/common/ActionDropdown.jsx
```diff
@@ -97,22 +97,41 @@ class ActionDropdown extends React.Component<ActionDropdownProps, ActionDropdown
     this.setState(({ show }) => ({ show: !show }));
   };
 
-  render() {
-    const { children, container, element } = this.props;
-    const { show } = this.state;
+  adjustOptionOnSelectProp = (option: React.Element<*>, updateDepth: number) => {
```
kmerz, line 100: please rename `option` to `child` or `node`, it is in no case an option and it took me 3 minutes to understand that

linuspahl: Thanks for having a look. I've adjusted the naming.

kmerz (approved):
LGTM :+1: 

PR #8344, opened 2020-06-15
Refactor field list component

linuspahl:
This PR refactors the field list component by:
* removing usage of createReactClass by using a functional component
* splitting up the component in multiple components
* using styled component instead of css file
* adding flow types
* removing usage of not needed ViewMetadata store

## Motivation and Context
I always wanted to refactor this component and it helps avoiding linter warnings in open PRs.

## Types of changes


graylog2-web-interface/src/views/components/sidebar/fields/FieldsOverview.jsx
```diff
@@ -0,0 +1,125 @@
...
+      return { ...state, filter: undefined };
+    case 'changeMode':
+      return mode ? { ...state, currentMode: mode } : state;
+    default:
+      return state;
+  }
+};
+
+const FieldsOverview = ({ allFields, fields, viewMetadata, listHeight }: Props) => {
+  const [{ filter, currentMode }, dispatch] = useReducer(reducer, { filter: undefined, currentMode: 'current' });
```
dennisoelkers, line 61: Using a reducer here overcomplicates things, as the field category and the filter are not related. They could be handled by using two separate, independent states.

graylog2-web-interface/src/views/components/sidebar/fields/ChangeMode.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Props = {
+  changeMode: (newMode: string) => void,
+  currentMode: string,
+  mode: string,
+  text: string,
+  title: string,
+};
+
+const ChangeMode = ({ changeMode, currentMode, mode, text, title }: Props) => {
```
dennisoelkers, line 12: `ChangeMode` is not an ideal name, it is not really giving away what it is doing. Maybe something like `FieldGroup`?
I think we can also align props naming with React standards. Instead of `changeMode` using `onSelect` or `onClick` would give away what it's doing, instead of `currentMode`, we can use a boolean `selected` prop, allowing the user of the component to determine if it should be shown as selected or not.

graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
```diff
@@ -1,184 +0,0 @@
-import React from 'react';
```
dennisoelkers, line 1: It might make sense to move a file instead of deleting it (especially in this case, where we are still keeping lots from the component, just under a new name), so we retain its history.

linuspahl: That is a good point I did not thought about. I named the component `FieldsOverview` because I wanted to avoid confusion with the `List` component in the same directory.

graylog2-web-interface/src/views/components/sidebar/fields/ChangeMode.jsx
linuspahl: Alright, I only see one disadvantage with the select prop, we need to define the group (previously mode) twice for each `FieldGroup`. For the `group` and the `select` prop.

graylog2-web-interface/src/views/components/sidebar/FieldList.jsx
linuspahl: It looks like I have to create a new PR to tell git I moved the file and changed the name. At least this seems to be the easiest way

linuspahl: I created a new branch and used `git mv` to move the and rename the file. This seems to work when looking at the commit: https://github.com/Graylog2/graylog2-server/commit/a74d2e435f218ffef7fac8fab212bbfe2ff7b6b0
But not when looking at the PR..
https://github.com/Graylog2/graylog2-server/compare/refactor-field-list-v2?expand=1

graylog2-web-interface/src/views/components/sidebar/fields/ChangeMode.jsx
dennisoelkers: Agreed, it's not ideal.

PR #8354, opened 2020-06-16
 Define bar width if bar chart contains only data for a single date

### LINKED ISSUE #7167: Time based bar charts with a single value do not render correctly
## Expected Behavior
When rendering a time based bar chart with a single value, the user expects a broad bar
which is as broad as the time bucket.

## Current Behavior
Since we do not tell plotly how wide a time bucket is, it tries to calculated on its own, but since
it does not have a second data point to calculate the bucket size it assumes it to be really small.

## Possible Solution
We have the bucket size with which we aggregated the data points, we can tell plotly how wide
we want the bucktes (bars) to be.
https://plot.ly/javascript/bar-charts/#customizing-individual-bar-widths

1. Create a time based (row pivot is a timestamp) bar chart widget.
2. Search for a time range with a single data point.

Graylog (43)

## Your Environment
* Graylog Version: Graylog beta 3 (e5b97c8795774bdc3a66753f5c0ba4f337aee3ed)

linuspahl:
As described in #7167 a bar chart with a time range and only one search result, results in one very small bar. This PR is fixing the problem by defining a width for the bar. The width can't be defined in px, it has to be the unit of the x axis (in this case ms).

Fixes #7167

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
@@ -6,7 +6,7 @@ import AggregationWidgetConfig from 'views/logic/aggregationbuilder/AggregationW
 import type { FieldTypeMappingsList } from 'views/stores/FieldTypesStore';
 import type { Rows } from 'views/logic/searchtypes/pivot/PivotHandler';
 import type { Events } from 'views/logic/searchtypes/events/EventHandler';
-import type { TimeRange } from 'views/logic/queries/Query';
+import type { AbsoluteTimeRange } from 'views/logic/queries/Query';
 
 import EmptyAggregationContent from './EmptyAggregationContent';
 import FullSizeContainer from './FullSizeContainer';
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/views/components/aggregationbuilder/AggregationBuilder.jsx
```diff
@@ -31,7 +31,7 @@ export type VisualizationComponentProps = {
   config: AggregationWidgetConfig,
   data: { [string]: Rows, events?: Events },
   editing?: boolean,
-  effectiveTimerange: TimeRange,
+  effectiveTimerange: AbsoluteTimeRange,
```
linuspahl, line 34: As far as I know the effective time range is always an `AbsoluteTimeRange`. Or is there a case were it can be another type?

graylog2-web-interface/src/views/logic/aggregationbuilder/Pivot.js
```diff
@@ -2,6 +2,9 @@
 
 type ConfigType = { [string]: mixed };
 
+export const DateType = 'time';
+export const ValuesType = 'values';
```
linuspahl, line 6: I will create a follow up PR to replace all related hard coded values.

PR #8369, opened 2020-06-17
Theming Select component

kyleknighted:
## Description

Theming `Select` component
- Adding `theme` as prop to use in react-select configurations

## Motivation and Context

Theme All The Things!

## How Has This Been Tested?

snapshots, flow, and lint

image

## Types of changes

## Checklist:


graylog2-web-interface/src/theme/index.js
```diff
@@ -14,12 +16,20 @@ const theme: ThemeInterface = {
 
 const themeModes: Array<string> = ['teinte'];
 
+const themePropTypes = PropTypes.shape({
+  breakpoints: PropTypes.object,
+  colors: PropTypes.object,
+  fonts: PropTypes.object,
+  utils: PropTypes.object,
+});
```
linuspahl, line 24: Maybe something for a separate PR, but we can also use this prop type in `GenericPlot`, `SavedSearchControls` and `ColorLabel`

graylog2-web-interface/src/components/common/Select.jsx
```diff
@@ -72,28 +73,24 @@ const clearIndicator = (base) => ({
...
+  border: `1px solid ${theme.colors.variant.lighter.info}`,
 });
 
-const multiValueLabel = (base) => ({
+const multiValueLabel = ({ theme }) => (base) => ({
   ...base,
-  color: 'unset',
   paddingLeft: '5px',
   paddingRight: '5px',
+  fontSize: theme.fonts.size.body,
```
linuspahl, line 85: I like the new, larger font size, but we should adjust the select height (part of `controlNormal`) in the same step, right now the height of the select changes, when selecting e.g. a stream. 

Changing the font size is maybe something for another PR, since we should update the height of the time range selects in the same step as well.

kyleknighted: Switched this to use `size.small` so it'll render around 12px and then in the future we can determine if we want to modify the heights and font sizes of this along with the time range selector... as well as maybe move the time range selector over to using this Select component

graylog2-web-interface/src/theme/index.js
kyleknighted: I can update them in this PR if you think it's useful to do now. Otherwise it can be a follow-up PR.

linuspahl: Lets do this in  follow-up PR.

graylog2-web-interface/src/components/common/Select.jsx
linuspahl: sounds good to me!

PR #8371, opened 2020-06-18
Display error message for search bar time range inputs

### LINKED ISSUE #7191: Show info when absolute search time range "from" is greater than "to"
When a user tries to define an absolute time range, with a greater "from" than "to" value, he should see a warning. I can't imagine a use case where this combination is useful.

image

## Your Environment

linuspahl:
With this PR we are displaying an error message when the validation of a time range input in the search bar fails. The error message is being displayed as a tooltip above the input.

It also adds a validation for the absolute time range selector, to throw an error if the date "from" is after the date "to".

Fixes #7191

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
@@ -38,7 +38,7 @@ const _isValidDateString = (dateString: string) => {
 
   return DateTime.isValidDateString(dateString)
     ? undefined
-    : `Format must be: ${DateTime.Formats.TIMESTAMP}`;
+    : 'Format must be: YYYY-MM-DD [HH:mm:ss[.SSS]]';
 };
 
 const AbsoluteTimeRangeSelector = ({ disabled }: Props) => {
--- a/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
@@ -38,7 +38,7 @@ const _isValidDateString = (dateString: string) => {
 
   return DateTime.isValidDateString(dateString)
     ? undefined
-    : `Format must be: ${DateTime.Formats.DATETIME}`;
+    : `Format must be: ${DateTime.Formats.TIMESTAMP}`;
 };
 
 const AbsoluteTimeRangeSelector = ({ disabled }: Props) => {
--- a/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
@@ -38,7 +38,7 @@ const _isValidDateString = (dateString: string) => {
 
   return DateTime.isValidDateString(dateString)
     ? undefined
-    : `Invalid date: ${dateString}`;
+    : `Format must be: ${DateTime.Formats.DATETIME}`;
 };
 
 const AbsoluteTimeRangeSelector = ({ disabled }: Props) => {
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/views/components/searchbar/SearchBarForm.jsx
```diff
@@ -21,6 +22,16 @@ type Props = {
   children: ((props: FormikProps<Values>) => React$Node) | React$Node,
 };
 
+const validate = (values) => {
+  const errors = {};
+  if (values.timerange.type === 'absolute' && DateTime.isValidDateString(values.timerange.from) && values.timerange.from > values.timerange.to) {
```
linuspahl, line 27: The biggest problem I faced while implementing these changes:

I had to check here if the "from" date is a valid date string, which is also part of the `AbsoluteTimeRangeSelector` validation . Imagine the following scenario:
Start date: 2020-06-18 11:04:34.82 (invalid date because last number is missing)
End date: 2020-06-10 11:04:34.823

In this case the error message would be "Start date must be before end date" instead of "Format must be: YYYY-MM-DD HH:mm:ss", without the mentioned date string check in the form validation.

At one point I've created a `SearchBarFormValidation` file, but it was way more complex than the current implementation and I really prefer the component based validation approach. Unfortunately I could not find a way to give the field validation a higher priority than the form validation.

graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
```diff
@@ -36,7 +36,7 @@ const _isValidDateString = (dateString: string) => {
   }
   return DateTime.isValidDateString(dateString)
     ? undefined
-    : `Invalid date: ${dateString}`;
+    : `Format must be: ${DateTime.Formats.DATETIME}`;
```
dennisoelkers, line 39: During testing, I noticed that we have a slight inconsistency. The message says `Format must be: YYYY-MM-DD HH:mm:ss`, but we do allow specifying milliseconds. So the message should be `Format must be: YYYY-MM-DD HH:mm:ss[.SSS]`. This would mean not being able to reuse the `DateTime.Formats.DATETIME` constant, but this is okay, as we should support a couple of formats and use a combined format for the error message.

A suggestion for a future improvement: I think we should be a bit more relaxed when parsing the date. Right now we expect to specify all units going down to seconds. Imo seconds should be optional, maybe it should even be possible to make hours & minutes optional as well.

linuspahl: Good point, I've updated the error message.

graylog2-web-interface/src/views/components/searchbar/AbsoluteTimeRangeSelector.jsx
```diff
@@ -36,7 +36,7 @@ const _isValidDateString = (dateString: string) => {
   }
   return DateTime.isValidDateString(dateString)
     ? undefined
-    : `Invalid date: ${dateString}`;
+    : `Format must be: ${DateTime.Formats.TIMESTAMP}`;
```
dennisoelkers, line 39: This is now `YYYY-MM-DD HH:mm:ss.SSS`. It should be `YYYY-MM-DD HH:mm:ss[.SSS]` to indicate that milliseconds are optional. As mentioned in a [previous comment](https://github.com/Graylog2/graylog2-server/pull/8371#discussion_r442686678), I think we cannot reuse an existing date format string, as we do combine multiple of them (two at the moment, more if we also make e.g. seconds optional).

linuspahl: I thought the parentheses mark the difference, but I like the idea.
But when we mark a part as optional, shouldn't we display a message like `YYYY-MM-DD [HH:mm:ss[.SSS]]` since you can enter `2020-06-23` in the same way like `2020-06-23 10:10:37`?

dennisoelkers: Yes, absolutely. I was not aware of being able to leave out the time, I only tested leaving out seconds, which unfortunately does not work yet :/

PR #8381, opened 2020-06-19
Use an openjdk docker image with a fixed debian version

mpfz0r:
Otherwise, the test build will fail on systems that still
have an older version of the openjdk image around.
(debian stretch in my case)

Our official docker image has been using buster since the 3.1
release.

For reference:
```
$ docker image ls |grep jdk
openjdk                                         8-jre-slim-buster   778bc46b8d12        4 weeks ago         184MB
openjdk                                         8-jre-slim          54567f06e0e8        17 months ago       204MB
```

### DIFF
--- a/graylog2-server/src/test/resources/org/graylog/testing/graylognode/Dockerfile
+++ b/graylog2-server/src/test/resources/org/graylog/testing/graylognode/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre-slim
+FROM openjdk:8-jre-slim-buster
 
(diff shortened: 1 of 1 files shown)


alex-konn:
> Our official docker image has been using buster since the 3.1
> release.

You mean 3.3.1, right?
https://github.com/Graylog2/graylog-docker/commit/ca38ecc9b7200a9e35de2487f579c05d30d3e93c

But fair enough. Ideally we'd use the exact same docker image, I think. But I've been hesitant to use our official image, because I'm worried about our tests breaking due to changes in a completely different repo.

PR #8382, opened 2020-06-19
Update `/rebase` action.

dennisoelkers:
This change is doing the following:

  - updating the `/rebase` action to v1.3.1, to fix a couple of bugs (e.g. returning the error if the rebase failed)
  - adding an inline condition to skip the action if no `/rebase` was specified


alex-konn:
Worked. Woohoo!

PR #8451, opened 2020-06-30
Implement `MessagesAdapter` for ES7.

dennisoelkers:
## Motivation and Context

This PR is implementing the `MessagesAdapter` for ES7. It also performs quite extensive refactoring of the `Messages` <-> `MessagesAdapter` relationship. 
Before this PR, retrying logic was buried in the `MessagesAdapter#bulkIndex` call. This change is pulling it upwards into the `Messages` class, which is now retrying `bulkIndex` calls for thrown `IOException`s as well as indexing errors (e.g. index blocks due to target indices being read-only or the node being in flood stage). This is implemented by adding an abstraction for indexing errors (`Messages.IndexingError`) which is instantiated by the adapters but evaluated by the `Messages` class which decides how to proceed based on the error type.

Chunk splitting due to excessive payload sizes is still considered to be a concern of the adapter implementation, as it is related to the way of transport (HTTP). Nevertheless, parts of it are extracted into a shared class (`ChunkedBulkIndexer`) which can be employed by any adapter implementation to provide a strategy that splits chunks into halves until bulk indexing does not fail because of a 413 anymore.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/testing/FixtureImporterES6.java
+++ b/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/testing/FixtureImporterES6.java
@@ -20,10 +20,12 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.MoreObjects;
 import io.searchbox.client.JestClient;
+import io.searchbox.client.JestResult;
 import io.searchbox.core.Bulk;
 import io.searchbox.core.BulkResult;
 import io.searchbox.core.Index;
 import io.searchbox.indices.CreateIndex;
+import io.searchbox.indices.IndicesExists;
 import org.graylog.storage.elasticsearch6.jest.JestUtils;
 import org.graylog.testing.elasticsearch.FixtureImporter;
 import org.graylog2.jackson.TypeReferences;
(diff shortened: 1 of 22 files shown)


graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/testing/FixtureImporterES6.java
```diff
@@ -144,4 +159,16 @@ private void importNode(JsonNode root) throws IOException {
             throw new IllegalStateException(sb.toString());
         }
     }
+
+    private boolean indexExists(String indexName) {
+        final IndicesExists.Builder request = new IndicesExists.Builder(indexName);
+        final JestResult result = JestUtils.execute(jestClient, request.build(), () -> "Unable to check if index exists: " + indexName);
+
+        return result.isSucceeded();
```
alex-konn, line 167: `JestUtils#execute` will always throw an exception, if `result.isSucceeded() == false`

```suggestion
         try {
            final JestResult result = jestClient.execute(request.build());
            return result.isSucceeded();
        } catch (IOException e) {
            throw new RuntimeException("Unable to check if index exists: " + indexName, e);
        }
```

dennisoelkers: :heavy_check_mark:

PR #8456, opened 2020-07-01
Register ownership for newly created dashboards / eventdefinitions

mpfz0r:
Register entity ownership in the db service layer.
This way, we can call that from RestResources and also when entities are created from content packs.

Introduce a UserContext which can be used in RetResource methods.


graylog2-server/src/main/java/org/graylog/security/BuiltinCapabilities.java
```diff
@@ -53,7 +53,9 @@ public BuiltinCapabilities() {
                                 RestPermissions.STREAMS_EDIT,
                                 RestPermissions.STREAMS_CHANGESTATE,
                                 RestPermissions.DASHBOARDS_READ,
-                                RestPermissions.DASHBOARDS_EDIT
+                                RestPermissions.DASHBOARDS_EDIT,
+                                RestPermissions.EVENT_DEFINITIONS_READ,
```
bernd, line 57: Shouldn't we add the read permissions to the VIEW capability as well?

graylog2-server/src/main/java/org/graylog/security/UserContextBinder.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+import org.glassfish.jersey.process.internal.RequestScoped;
+
+public class UserContextBinder extends AbstractBinder {
+    @Override
+    protected void configure() {
+        bindFactory(UserContextFactory.class)
+                .to(UserContext.class)
+//                .proxy(true)
```
bernd, line 27: How about removing the commented code? Or at least add a comment why this is commented and should stay.

graylog2-server/src/main/java/org/graylog/security/entities/EntityOwnershipService.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Inject
+    public EntityOwnershipService(DBGrantService dbGrantService,
+                                  GRNRegistry grnRegistry) {
+        this.dbGrantService = dbGrantService;
+        this.grnRegistry = grnRegistry;
+    }
+
+    public void registerNewEventDefinition(String id, User user) {
+        final GRN grn = grnRegistry.newGRN(ModelTypes.EVENT_DEFINITION_V1.name(), id);
```
bernd, line 48: I think we should use `GRNTypes.EVENT_DEFINITION.type()` here instead of using the `ModelTypes`.

graylog2-server/src/main/java/org/graylog/security/entities/EntityOwnershipService.java
```diff
@@ -0,0 +1,70 @@
...
+        registerNewEntity(grn, user);
+    }
+
+    public void registerNewView(String id, User user) {
+        final GRN grn = grnRegistry.newGRN(ModelTypes.DASHBOARD_V2.name(), id);
```
bernd, line 53: Same as above, let's use `GRNTypes.DASHBOARD.type()` here.

graylog2-server/src/main/java/org/graylog/security/entities/EntityOwnershipService.java
```diff
@@ -0,0 +1,70 @@
...
+import org.graylog2.contentpacks.model.ModelTypes;
+import org.graylog2.plugin.database.users.User;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
+public class EntityOwnershipService {
```
bernd, line 33: How about writing a test for this class? :smile: Thanks!

mpfz0r:
Tests only fail on frontend stuff, we better ignore that, if we want to keep our sanity ;-)

mpfz0r:
> Tests are currently failing.

the enterprise plugin build failure is expected, because I've changed the `DBEventDefinitionService`. This needs another change in enterprise.

And for the ESLint stuff, what should I do about it? :shrug:

PR #8472, opened 2020-07-02
Rename "role" to "capability"

bernd:
We updated our concepts for roles and capabilities.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
+++ b/graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
@@ -103,7 +103,7 @@ public class EntitySharesService {
      */
     public EntitySharePrepareResponse updateEntityShares(GRN ownedEntity, EntityShareRequest request, User sharingUser) {
         final String userName = sharingUser.getName();
-        final List<GrantDTO> existingGrants = grantService.getForTarget(ownedEntity, grnRegistry.newGRN("user", sharingUser.getName()));
+        final List<GrantDTO> existingGrants = grantService.getForTarget(ownedEntity);
 
         // Update capabilities of existing grants (for a grantee)
         existingGrants.stream().filter(grantDTO -> request.grantees().contains(grantDTO.grantee())).forEach((g -> {
@@ -162,7 +162,7 @@ public class EntitySharesService {
     }
 
     private ImmutableSet<ActiveShare> getActiveShares(GRN ownedEntity, User sharingUser) {
-        final List<GrantDTO> activeGrants = grantService.getForTarget(ownedEntity, grnRegistry.newGRN("user", sharingUser.getName()));
+        final List<GrantDTO> activeGrants = grantService.getForTarget(ownedEntity);
 
         return activeGrants.stream()
                 .map(grant -> ActiveShare.create(grnRegistry.newGRN("grant", grant.id()).toString(), grant.grantee(), grant.capability()))
--- a/graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
+++ b/graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
@@ -19,12 +19,12 @@ package org.graylog.security.shares;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import org.apache.shiro.subject.Subject;
-import org.graylog.security.BuiltinRoles;
+import org.graylog.security.BuiltinCapabilities;
 import org.graylog.security.DBGrantService;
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
```diff
@@ -98,32 +98,32 @@ public EntitySharePrepareResponse prepareShare(GRN ownedEntity,
      * The grants in the request are created or, if they already exist, updated.
      *
      * @param ownedEntity the target entity for the updated grants
-     * @param request     the request containing grantees and their roles
+     * @param request     the request containing grantees and their capabilities
      * @param sharingUser the user executing the request
      */
     public EntitySharePrepareResponse updateEntityShares(GRN ownedEntity, EntityShareRequest request, User sharingUser) {
         final String userName = sharingUser.getName();
-        final List<GrantDTO> existingGrants = grantService.getForTarget(ownedEntity, grnRegistry.newGRN("user", sharingUser.getName()));
```
mpfz0r, line 106: did we reach a decision regarding this?
In my book we were still filtering the current user..
This also needs to be handled by the frontend now. (If they don't send back your own grant, it will get deleted)

graylog2-server/src/main/java/org/graylog2/utilities/GRNRegistry.java
```diff
@@ -39,6 +39,7 @@
     // TODO find a way to unify these
     private static final ImmutableSet<GRNType> BUILTIN_TYPES = ImmutableSet.<GRNType>builder()
             .add(GRNType.create("collection", "collections:"))
+            .add(GRNType.create("capability", "capabilities:"))
```
mpfz0r, line 42: does this need a permissions prefix? capabilities will never be entities that are managed with permissions.
Same goes for roles.
I'm not even sure we should be using GRNs for that, actually..

graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
bernd: Good point. I actually saw this as a bug. :smile: My understanding was, that we need to allow that a user can remove its own ownership for an entity. That means the sharing user needs to be in the active shares and shown in the list of users in the share dialog. If we exclude the current user from the active shares, we need a different way for allowing a user to remove its own ownership.

graylog2-server/src/main/java/org/graylog2/utilities/GRNRegistry.java
bernd: Not sure, but I thought it also didn't hurt at the moment. I also don't see a reason *not* to use GRNs for that. What's your thinking behind that?

bernd: To clarify: My thinking was, that we might need to allow modification of capabilities (e.g. adding new permissions from plugins).

Now that I spelled it out... Extending capabilities will probably an internal thing without user interaction, so we also don't need permissions for that. :smile:

graylog2-server/src/main/java/org/graylog/security/shares/EntitySharesService.java
mpfz0r: From a UX perspective, I'd say this will clutter the UI, because it's a rare condition.
But we can keep the API generic. The frontend could still hide it..

graylog2-server/src/main/java/org/graylog2/utilities/GRNRegistry.java
mpfz0r: > Not sure, but I thought it also didn't hurt at the moment. I also don't see a reason _not_ to use GRNs for that. What's your thinking behind that?

Just for simplicity. Currently we could be having three easily readable constants in the Grants table: `View`, `Manage`, `Own`

mpfz0r: > To clarify: My thinking was, that we might need to allow modification of capabilities (e.g. adding new permissions from plugins).
> 
> Now that I spelled it out... Extending capabilities will probably an internal thing without user interaction, so we also don't need permissions for that. smile

having capabilities to modify capabilities :see_no_evil:  reaching tourning completeness :-D

bernd: > Just for simplicity. Currently we could be having three easily readable constants in the Grants table: `View`, `Manage`, `Own`

Okay, sounds good. So I would update the code to only use simple keywords `view`, `manage` and `own` for the capabilities. If we ever need to use GRNs, it should be a simple migration. :+1:

PR #8518, opened 2020-07-09
Implement dependency permission check

bernd:
Also rename the `missing_dependencies` field in the sharing response to `missing_permissions_on_dependencies`.
The previous name was wrong and didn't represent the content of the field.


graylog2-server/src/main/java/org/graylog/security/UserAuthorizer.java
```diff
@@ -0,0 +1,56 @@
...
+    public interface Factory {
+        UserAuthorizer create(GRN grantee);
+    }
+
+    private final Subject subject;
+
+    @Inject
+    public UserAuthorizer(DefaultSecurityManager securityManager, @Assisted GRN grantee) {
+        this.subject = new Subject.Builder(securityManager)
+                .principals(new SimplePrincipalCollection(grantee.entity(), "UserAuthorizer"))
```
mpfz0r, line 40: This will potentially build a "user" with the principal of a team..
We need to solve this sooner or later..

bernd: I refactored the code to also handle teams in the future.

PR #8520, opened 2020-07-09
Add new endpoint to retrive a paginated list of users

kmerz:
## Description
Prior to this change, the users were only receivable
all in one chunk.

This change will get the users paginated.

The old '/system/users' endpoint is marked as deprecated.

## How Has This Been Tested?
- Via API Browser

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -152,11 +179,7 @@ public UserList listUsers() {
 
         // among all active sessions, find the last recently used for each user
         //noinspection OptionalGetWithoutIsPresent
```
bernd, line 181: This `noinspection` comment should be moved to the `getLastSessionForUser()` method.

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -171,6 +194,72 @@ public UserList listUsers() {
...
+        List<UserDTO> users = result.stream().map(userDTO -> {
+            UserDTO.Builder builder = userDTO.toBuilder()
+                    .fillSession(lastSessionForUser.getOrDefault(userDTO.username(), Optional.empty()));
+            if (Optional.ofNullable(userDTO.roles()).isPresent()) {
+                builder.roles(userDTO.roles().stream().map(roleNameMap::get).collect(Collectors.toSet()));
+            }
+            return builder.build();
+        }).collect(Collectors.toList());
+
+        final PaginatedList<UserDTO> userDTOS = new PaginatedList<>(users, result.pagination().total(),
```
bernd, line 250: This variable is unused. The `users` variable is put into the `UserPageListResponse`. Is that correct?

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -171,6 +194,72 @@ public UserList listUsers() {
...
+        } catch (org.graylog2.database.NotFoundException e) {
+            throw new NotFoundException("Couldn't find roles: " + e.getMessage());
+        }
+
+        final UserDTO adminUser = getAdminUserDTO(lastSessionForUser, roleNameMap);
+
+        List<UserDTO> users = result.stream().map(userDTO -> {
+            UserDTO.Builder builder = userDTO.toBuilder()
+                    .fillSession(lastSessionForUser.getOrDefault(userDTO.username(), Optional.empty()));
+            if (Optional.ofNullable(userDTO.roles()).isPresent()) {
```
bernd, line 244: `UserDTO#roles()` is not marked as nullable so it should always have a value. Why are you using the `Optional.ofNullable().isPresent()` construct here?

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -171,6 +194,72 @@ public UserList listUsers() {
...
+                builder.roles(userDTO.roles().stream().map(roleNameMap::get).collect(Collectors.toSet()));
+            }
+            return builder.build();
+        }).collect(Collectors.toList());
+
+        final PaginatedList<UserDTO> userDTOS = new PaginatedList<>(users, result.pagination().total(),
+                result.pagination().page(), result.pagination().perPage());
+        return UserPageListResponse.builder()
```
bernd, line 252: I think we can use the `PaginatedResponse` return value here. That would make the response similar to what we return from other paginated endpoints. (with the custom `UserPageListResponse` we have a `pagination` key inside the response, `PaginatedResponse` doesn't have that) The `admin_user` will move to under the `context` key in the response, but I think that's okay.

```json
return PaginatedResponse.create("users", userDTOS, query, Collections.singletonMap("admin_user", adminUser));
```

graylog2-server/src/main/java/org/graylog2/users/RoleServiceImpl.java
```diff
@@ -159,6 +163,24 @@ public boolean exists(String roleName) {
         }
     }
 
+    @Override
+    public Map<String, Role> findIdMap(Set<String> roleIds) throws NotFoundException {
+        final DBObject query = BasicDBObjectBuilder.start()
+                .push(ID)
+                .append("$in", roleIds.stream().map(ObjectId::new).collect(Collectors.toSet()))
+                .get();
```
bernd, line 171: You can use the `DBQuery.in()` helper here:

```suggestion
        final DBQuery.Query query = DBQuery.in(ID, roleIds);
```

Mongojack also takes care of the `ObjectId` conversion.

UPGRADING.rst
```diff
@@ -13,6 +13,15 @@ We therefore mark '/streams' as deprecated. Users who use this endpoint for scri
 their scripts to the format of the new endpoint, so they only need to to change the URL when '/streams/paginated' will become
 '/streams'.
 
+Deprecation of API endpoint for unpaginated listing of users
```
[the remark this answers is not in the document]
bernd: We decided to keep the deprecation warning. If we deprecate all non-paginated endpoints in the future is a different discussion.

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -171,6 +194,72 @@ public UserList listUsers() {
...
+    @RequiresPermissions(RestPermissions.USERS_LIST)
+    @Produces(MediaType.APPLICATION_JSON)
+    public UserPageListResponse getPage(@ApiParam(name = "page") @QueryParam("page") @DefaultValue("1") int page,
+                                        @ApiParam(name = "per_page") @QueryParam("per_page") @DefaultValue("50") int perPage,
+                                        @ApiParam(name = "query") @QueryParam("query") @DefaultValue("") String query,
+                                        @ApiParam(name = "sort",
+                                                value = "The field to sort the result on",
+                                                required = true,
+                                                allowableValues = "title,description")
+                                            @DefaultValue(StreamImpl.FIELD_TITLE) @QueryParam("sort") String sort,
```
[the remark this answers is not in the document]
bernd: Lol, that wrong field *was* on the `@DefaultValue`. :joy:

PR #8396, opened 2020-06-22
Save sidebar isPinned setting as a user preference

linuspahl:
_This PR is based on https://github.com/Graylog2/graylog2-server/pull/8294 which needs to be merged first_

With https://github.com/Graylog2/graylog2-server/pull/8294 we've implemented a sidebar option to display the sidebar as an overlay or inline. With this PR we are saving this setting as a user preferences.

An important point we need to discuss:
- The user preferences can't be changed for the admin, by using the API.

Fixes https://github.com/Graylog2/graylog2-server/issues/8395

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #8527, opened 2020-07-10
Save system user sidebar pinning in local storage and differentiate setting for searches and dashboards

linuspahl:
This is a follow-up PR for https://github.com/Graylog2/graylog2-server/pull/8396 which saves the sidebar layout setting as a user preference. Because a user preference can't be changed for the system admin, we've decided to save this setting in the local storage, for this type of users.

With this PR we also implement another setting for sidebar pinning on dashboards. This way a user is able to define different sidebar layout settings for dashboards and e.g. saved searches.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.jsx
```diff
@@ -16,13 +20,28 @@ type Props = {
...
-export const defaultLayoutConfig = (userPreferences?: UserPreferences) => ({
-  sidebar: {
-    isPinned: userPreferences?.searchSidebarIsPinned ?? false,
-  },
-});
+export const defaultLayoutConfig = (currentUser: ?User, userPreferences?: UserPreferences) => {
+  let dashboardSidebarIsPinned = false;
+  let searchSidebarIsPinned = false;
+
+  if (currentUser?.id === 'local:admin') {
```
linuspahl, line 30: I could not find a function like `isSystemAdmin(user)`. I prefer a separate function, but was not sure where to place it. Where would you place this kind of function?

linuspahl:
@dennisoelkers Thanks for the good feedback. I extracted the logic of the `SearchPageLayoutProvider` to a separate component (`SearchPageLayoutState`) which provides a getter and setter function. In the beginning I was not completely sure which logic I should extract and which should stay. Initially the `SearchPageLayoutState` contained only a getter and setter: https://github.com/Graylog2/graylog2-server/pull/8527/commits/0d75e533141c40ed885513a4fbec424396f7c145

These changes already reduced the complexity, but in the end I moved every logical part to the `SearchPageLayoutState`  component.

PR #8538, opened 2020-07-14
Create a basic AuthzRole resource

kmerz:
Prior to this change, we did not have a paginated access
to all roles and to the roles of a user.

This change is adding a new resource which will query all
roles and the roles of a user.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRoleDTO.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRoleDTO.java
@@ -1,0 +1,77 @@
+package org.graylog.security.authzroles;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.auto.value.AutoValue;
+import org.mongojack.Id;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRoleDTO.java
```diff
@@ -0,0 +1,93 @@
...
+    @Id
+    @ObjectId
+    @Nullable
+    @JsonProperty(FIELD_ID)
+    public abstract String id();
+
+    @JsonProperty(FIELD_NAME)
+    public abstract String name();
+
+    @JsonProperty(FIELD_DESCRIPTION)
```
mpfz0r, line 49: Custom roles can have an empty description.

```suggestion
    @JsonProperty(FIELD_DESCRIPTION)
    @Nullable
```

graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRoleDTO.java
```diff
@@ -0,0 +1,93 @@
...
+        @Id
+        @ObjectId
+        @JsonProperty(FIELD_ID)
+        public abstract Builder id(String id);
+
+        @JsonProperty(FIELD_NAME)
+        public abstract Builder name(String name);
+
+        @JsonProperty(FIELD_DESCRIPTION)
+        public abstract Builder description(String description);
```
mpfz0r, line 82: same here
```suggestion
        public abstract Builder description(@Nullable String description);
```

PR #8554, opened 2020-07-15
type check unary NOT operator

kroepke:
## Description
The type checker did not properly check the subexpression type for the NOT operator, which could lead to runtime errors
NOT binds closer than comparison (==), which is a common source of mistakes
in this case a string-valued function as being put into `NOT to_string($message.field) == "content"` which parses as `(NOT to_string($message.field)) == "content"`
the type checker did not pick up that `to_string($message.field)` wasn't boolean-valued and the interpreter failed at runtime due to a class cast exception.

After this change, the type checker requires the expression passed to NOT to be of boolean type and makes this a parse-time error

this should not have any incompatibility issues, because it never executed successfully (unless the expression was short-circuited). in any case, rules like this are severely broken and need to be fixed by the user

## Motivation and Context
The parser erroneously allowed an invalid operation to go through, which would always fail at runtime, leading to dropped messages.

The correct way to express `NOT to_string($message.field) == "content"` is to write `NOT (to_string($message.field) == "content")` or to use the `!=` operator instead of `NOT x == y`.

## How Has This Been Tested?
Added a test case rule that fails with a runtime exception before the fix, and properly returns a parse exception after the fix.

## Types of changes

## Checklist:


kroepke:
This should at least be backported to 3.3.

There's an open question on whether we need to somehow scan rules that now are invalid due to this change because users need to update them to make them work after this patch.
Those rules were never good because they could potentially lose data, but now the entire rule wouldn't run.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/parser/PipelineRuleParser.java
```diff
@@ -758,6 +758,15 @@ public void exitOr(RuleLangParser.OrContext ctx) {
             checkBinaryExpression(ctx);
         }
 
+        @Override
+        public void exitNot(RuleLangParser.NotContext ctx) {
+            final Expression expression = parseContext.expressions().get(ctx.expression());
+            Class type = expression.getType();
+            if (!Boolean.class.equals(type)) {
```
mpfz0r, line 765: I might be nitpicking here, but wouldn't it be more correct to use:
```suggestion
            if (!Boolean.class.isAssignableFrom(type)) {
```

mpfz0r:
> This should at least be backported to 3.3.
> 
> There's an open question on whether we need to somehow scan rules that now are invalid due to this change because users need to update them to make them work after this patch.
> Those rules were never good because they could potentially lose data, but now the entire rule wouldn't run.

Those broken rules are now silently ignored after server startup, right? 
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java#L139

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/parser/PipelineRuleParser.java
kroepke: We could do that but Boolean is a final class, so it's effectively identical :)

kroepke:
> > This should at least be backported to 3.3.
> > There's an open question on whether we need to somehow scan rules that now are invalid due to this change because users need to update them to make them work after this patch.
> > Those rules were never good because they could potentially lose data, but now the entire rule wouldn't run.
> 
> Those broken rules are now silently ignored after server startup, right?
> [`master`/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java#L139](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java?rgh-link-date=2020-07-16T14%3A31%3A43Z#L139)

Yeah, that's what I'm worried about.

I'm not sure what the best way of dealing with this is, we could log this prominently, hoping people find it. I'm not sure this warrants a notification of some sort. @bernd what do you think?

bernd:
> > > This should at least be backported to 3.3.
> > > There's an open question on whether we need to somehow scan rules that now are invalid due to this change because users need to update them to make them work after this patch.
> > > Those rules were never good because they could potentially lose data, but now the entire rule wouldn't run.
> > 
> > 
> > Those broken rules are now silently ignored after server startup, right?
> > [`master`/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java#L139](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java?rgh-link-date=2020-07-16T14%3A31%3A43Z#L139)
> 
> Yeah, that's what I'm worried about.
> 
> I'm not sure what the best way of dealing with this is, we could log this prominently, hoping people find it. I'm not sure this warrants a notification of some sort. @bernd what do you think?

@kroepke @mpfz0r If there is a chance that this breaks rules which work in 3.3, I would rather not backport this fix to 3.3 and only do it for 4.0. We can document this as known issue in the 3.3 pipeline docs to let people know.

If we still want to backport this, I think we need verbose logging and a system notification to make sure people see it.

kroepke:
> @kroepke @mpfz0r If there is a chance that this breaks rules which work in 3.3, I would rather not backport this fix to 3.3 and only do it for 4.0. We can document this as known issue in the 3.3 pipeline docs to let people know.

Good point. How about a compromise:

In 3.3 we are backporting the check, but do not fail the parsing, but rather log errors with hints on how to fix it.
In 4.0 we fail the parsing and also complain loudly.

bernd:
> Good point. How about a compromise:
> 
> In 3.3 we are backporting the check, but do not fail the parsing, but rather log errors with hints on how to fix it.
> In 4.0 we fail the parsing and also complain loudly.

@kroepke @mpfz0r Good idea! I like it! :+1:

kroepke:
Cool, I'll add some extra logging and will create a separate 3.3 PR then

mpfz0r:
> Cool, I'll add some extra logging and will create a separate 3.3 PR then

@kroepke This would implement the "complain loudly part":
```
--- graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java
+++ graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/ConfigurationStateUpdater.java
@@ -136,6 +136,7 @@ public class ConfigurationStateUpdater {
             try {
                 rule = pipelineRuleParser.parseRule(ruleDao.id(), ruleDao.source(), false, commonClassLoader);
             } catch (ParseException e) {
+                log.warn("Ignoring non parseable rule <{}/{}> with errors <{}>", ruleDao.title(), ruleDao.id(), e.getErrors());
                 rule = Rule.alwaysFalse("Failed to parse rule: " + ruleDao.id());
             }
             ruleNameMap.put(rule.name(), rule);
```

bernd (approved):
Tested locally again. Works and even found another broken rule in my setup. :smile:

PR #8573, opened 2020-07-17
Roles in user edit

kmerz:
## Description
Add Roles selection to User Edit

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #8590, opened 2020-07-21
Teams edit

kmerz:
##  Description
Add Teams Section to the UserDetails and UserEdit page.

#8573 needs to be merged first.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/users/UserDetails/TeamSection.jsx
+++ b/graylog2-web-interface/src/components/users/UserDetails/TeamSection.jsx
@@ -1,0 +1,41 @@
+// @flow strict
+import * as React from 'react';
+import styled from 'styled-components';
+import { PluginStore } from 'graylog-web-plugin/plugin';
+
+import User from 'logic/users/User';
+
+import SectionComponent from '../SectionComponent';
+
+type Props = {
+  user: User,
+};
+
+const Header = styled.h4`
+  margin-bottom: 5px;
+`;
+
+const NoEnterpriseComponent = () => (
+  <SectionComponent title="Teams">
+    <Header>No enterprise plugin found</Header>
+    <p>To use the <b>Teams</b> functionality you install the Graylog <b>Enterprise</b> plugin.</p>
+  </SectionComponent>
+);
+
+const TeamsSection = ({ user }: Props) => {
+  const teamsPlugin = PluginStore.exports('teams');
+
+  if (!teamsPlugin || teamsPlugin.length <= 0) {
+    return <NoEnterpriseComponent />;
+  }
+
+  const { TeamEditComponent } = teamsPlugin[0];
+
+  return (
+    <SectionComponent title="Teams">
+      <TeamEditComponent user={user} readOnly />
+    </SectionComponent>
+  );
+};
+
+export default TeamsSection;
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/components/users/UserDetails/UserDetails.jsx
```diff
@@ -28,8 +29,8 @@ const UserDetails = ({ user }: Props) => {
           <SettingsSection user={user} />
         </div>
         <div>
-          <SectionComponent title="Teams">Children</SectionComponent>
           <RolesSection user={user} />
+          <TeamsSection user={user} />
```
kyleknighted, line 33: Could the `user` be added to a Context Provider so that we do not have to pass it to every component?

graylog2-web-interface/src/components/users/UserEdit/TeamsSection.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+type Props = {
+  user: User,
+};
+
+const Header = styled.h4`
+  margin-bottom: 5px;
+`;
+
+const NoEnterpriseComponent = () => (
```
kyleknighted, line 18: This looks to be the same as `src/components/users/UserDetails/TeamSection.jsx:18`

Can we add a link to `/system/enterprise` so that users can sign up for Enterprise?

graylog2-web-interface/src/components/users/UserDetails/UserDetails.jsx
kmerz: I will discuss this with @linuspahl on monday

graylog2-web-interface/src/components/users/UserEdit/TeamsSection.jsx
kmerz: No because there is not `/system/enterprise` since there is no Enterprise Plugin installed

graylog2-web-interface/src/components/users/UserDetails/UserDetails.jsx
linuspahl: The advantages are not huge, but I agree we could something like a `UserDetailsContext`.

graylog2-web-interface/src/components/users/UserEdit/RolesSelector.jsx
```diff
@@ -0,0 +1,123 @@
...
+      <Formik onSubmit={onUpdate}
+              initialValues={{ role: undefined }}>
+        {({ isSubmitting, isValid, errors }) => (
+          <Form>
+            <FormElements>
+              <Field name="role" validate={_isRequired('role')}>
+                {({ field: { name, value, onChange } }) => (
+                  <StyledSelect inputProps={{ 'arial-label': 'Search for roles ' }}
+                                onChange={(role) => {
+                                  onChange({ target: { value: role, name } });
```
linuspahl, line 94: I should have mentioned it in the roles select PR, but let's return the function here `onChange={(role) => onChange({ target: { value: role, name } })`. Just in case the related component needs to work with the return value of `onChange` in the future.

graylog2-web-interface/src/components/users/UserDetails/TeamSection.jsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const Header = styled.h4`
+  margin-bottom: 5px;
+`;
+
+const NoEnterpriseComponent = () => (
+  <SectionComponent title="Teams">
+    <Header>No enterprise plugin found</Header>
+    <p>To use the <b>Teams</b> functionality you install the Graylog <b>Enterprise</b> plugin.</p>
```
linuspahl, line 21: *`you need to install`

PR #8613, opened 2020-07-23
Buttons rendering faster

kyleknighted:
## Description

- Removed `map` of variants and generate based on bsStyle
- Introduced [React.memo](https://reactjs.org/docs/react-api.html#reactmemo) which allows memoizing:
  - > If your function component renders the same result given the same props, you can wrap it in a call to React.memo for a performance boost in some cases by memoizing the result.
- Removed `chroma` generated colors every time Button is rendered

## Motivation and Context

Performance concerns on Grok Patterns page as well as new forms being built within Formik

image

## Types of changes

## Checklist:


graylog2-web-interface/src/components/graylog/Button.jsx
```diff
@@ -1,13 +1,78 @@
-import React, { forwardRef } from 'react';
+import React from 'react';
 import styled, { css } from 'styled-components';
 // eslint-disable-next-line no-restricted-imports
 import { Button as BootstrapButton } from 'react-bootstrap';
 
-const StyledButton = styled(BootstrapButton)(({ theme }) => css`
-  ${theme.components.button}
-`);
+const buttonStyles = (bsStyle = 'default') => css(({ theme }) => {
```
dennisoelkers, line 6: I do think it's still unnecessary to compute styles over and over on every initial render of a button component, especially considering that the possible values of `bsStyle` is a rather small set. Using `memo` improves re-rendering times, computing button styles once and stuffing computed styles into the theme would also improve the initial rendering.

graylog2-web-interface/src/components/graylog/Button.jsx
```diff
@@ -1,13 +1,78 @@
...
+          background-color: ${disabledBackground};
+          border-color: ${disabledBorder};
+          color: ${disabledColor};
+        }
+      }
+    }
+  `;
+});
+
+const Button = React.memo(styled(BootstrapButton)(({ bsStyle }) => css`
```
dennisoelkers, line 73: Before this change, the `Button` component was wrapped in a `forwardRef` call. Why is this not needed anymore?

kyleknighted:
Closed in favor of #8618

PR #8619, opened 2020-07-24
Expose GRNPermissions to frontend

### LINKED ISSUE #8603: Find a way to export entity ownership permissions to the frontend
The newly introduced wwnership permissions are not compatible with the existing permissions,
because they use GRNs instead of simple IDs.

Currently we return only `WildcardPermissions` to the frontend in `UserServiceImpl`

mpfz0r:
The frontend needs to know which entities a user
currently owns. This is expressed with a new Permissions
type, that is incompatible with the existing WildcardPermission.

Introduce a new field `grn_permissions` into the UserSummary object.

The serialized format of a GRNPermission is

`<permission>:<grn>`

e.g.: `entity:own:grn::::dashboard:5e2afc66cd19517ec2dabaf3`

Fixes #8603


graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -609,13 +611,25 @@ private UserSummary toUserResponse(User user,
             lastActivity = session.getLastAccessTime();
             clientAddress = session.getHost();
         }
+        List<WildcardPermission> wildcardPermissions;
+        List<GRNPermission> grnPermissions;
+        if (includePermissions) {
+            wildcardPermissions = userService.getPermissionsForUser(user).stream()
```
bernd, line 617: How about creating `UserService#getWildcardPermissionsForUser` and `UserService#getGRNPermissionsForUser` to avoid duplication of this code block in `RolesResource`?

mpfz0r: good point. will do :+1:

PR #8622, opened 2020-07-27
Fix performance problems with react-select and formik

kmerz:
Formik is rerendering every key stroke. But react-select
is rendering rather slow. So we use `FastField` which is
implementing shouldComponentUpdate for that field.

Now react-select is only rendered when necessary.


linuspahl:
@kmerz How did you measured the performance improvement and does it make sense to implement `FastField` for `CapabilitySelect` as well?

PR #8652, opened 2020-07-29
Use GRNs instead of Strings in AvailableGrantee model

mpfz0r:
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1676

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/grn/GRNRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog/grn/GRNRegistry.java
@@ -34,7 +34,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 @Singleton
 public class GRNRegistry {
     // TODO: Find a better name for the "everyone" grantee GRN type
-    public static final GRN GLOBAL_USER_GRN = BUILTIN_TEAM.newGRNBuilder().entity("everyone").build();;
+    public static final GRN GLOBAL_USER_GRN = GRNTypes.BUILTIN_TEAM.newGRNBuilder().entity("everyone").build();;
 
     private final ConcurrentMap<String, GRNType> REGISTRY = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/grn/GRNRegistry.java
```diff
@@ -33,7 +33,10 @@
  */
 @Singleton
 public class GRNRegistry {
-    public static final String GLOBAL_USER_GRN = "grn::::builtin-team:everyone"; // TODO: Find a better name for the "everyone" grantee GRN type
+    // TODO: Find a better name for the "everyone" grantee GRN type
+    public GRN GLOBAL_USER_GRN() {
```
bernd, line 37: Why are you converting this into a method instead of keeping it as a constant? If we use a method, we should adhere to the method naming conventions.

mpfz0r: yeah, not sure what I did there :-D

graylog2-server/src/main/java/org/graylog/grn/GRNRegistry.java
```diff
@@ -33,7 +33,8 @@
  */
 @Singleton
 public class GRNRegistry {
-    public static final String GLOBAL_USER_GRN = "grn::::builtin-team:everyone"; // TODO: Find a better name for the "everyone" grantee GRN type
+    // TODO: Find a better name for the "everyone" grantee GRN type
+    public final GRN GLOBAL_USER_GRN = newGRN(GRNTypes.BUILTIN_TEAM, "everyone");
```
bernd, line 37: How about using this? That way we can make it `static` and don't have a public instance field that looks like a constant.
 
```suggestion
    public static final GRN GLOBAL_USER_GRN = BUILTIN_TEAM.newGRNBuilder().entity("everyone").build();;
```

mpfz0r: ah, much nicer

PR #8672, opened 2020-07-31
Enable Search After for ES7 exports by adding a field alias for gl2_message_id in event indices

### LINKED ISSUE #7986: Use "search after" instead of scrolling as default for export
As described in #7985, our initial attempt to use "search after" for all exports has failed due to problems with events and old messages, both of which don't have `gl2_message_id`, which we depend on for breaking ties.

As soon as we include support for Elasticsearch versions that support [field aliases](https://www.elastic.co/blog/introducing-field-aliases-in-elasticsearch) (>= 6.4), we can add `gl2_message_id` as an alias to event indices. We can also document how to add an alias to older indices for customers that wish to export data from them.

We will then switch back to "search after" as default for exports, to avoid the drawbacks of scrolling, namely:
- load on Elasticsearch clusters 
- implicit limit on amount of used indices, because scrolling works with a bodyless `GET`, meaning the generated URL can get too long, if too many indices are used

Refs #7354
Refs #7985

alex-konn:
## Description
This PR enables Search After for exports using the Elasticsearch 7 storage driver. The `Scroll` implementation of the `RequestStrategy` interface was completely removed.
A migration was added, which adds this field alias to all event indices. In addition the alias was added to `EventsIndexMapping7`, so that all newly created event indices will get it.

Fixes #7986

## Motivation and Context
We only had both, because we initially wanted to always use Search After, but then realized that our event indices were lacking the `gl2_message_id` field, meaning there was no common tie breaker field between event and message indices.
In 6.4 Elasticsearch introduced field aliases, which we can now leverage to add `gl2_message_id` to event indices without duplicating data. 
This was only done for the ES7 driver, because field aliases are not supported for all minor versions of ES 6. I think that keeping things simple for the ES6 driver is more valuable than implementing it there only for versions >= 6.4. If we really want, we could add that in a follow-up PR.

## How Has This Been Tested?
- Added unit and integration tests
- Manual testing on local machine

## Types of changes


graylog2-server/src/main/java/org/graylog2/storage/VersionAwareStorageModule.java
```diff
@@ -31,18 +31,7 @@
...
-import org.graylog2.storage.providers.IndexFieldTypePollerAdapterProvider;
-import org.graylog2.storage.providers.IndexToolsAdapterProvider;
-import org.graylog2.storage.providers.IndicesAdapterProvider;
-import org.graylog2.storage.providers.MessagesAdapterProvider;
-import org.graylog2.storage.providers.MoreSearchAdapterProvider;
-import org.graylog2.storage.providers.NodeAdapterProvider;
-import org.graylog2.storage.providers.SearchesAdapterProvider;
-import org.graylog2.storage.providers.V20170607164210_MigrateReopenedIndicesToAliasesClusterStateAdapterProvider;
-import org.graylog2.storage.providers.V20200730000000_AddGl2MessageIdFieldAliasForEventsElasticsearchAdapterProvider;
+import org.graylog2.storage.providers.*;
```
dennisoelkers, line 34: Please do not use wildcard imports.

graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200730000000_AddGl2MessageIdFieldAliasForEventsTest.java
```diff
@@ -27,10 +27,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
dennisoelkers, line 30: Please do not use wildcard imports.

PR #8679, opened 2020-08-03
Add get a single role from the AuthzRolesResource

kmerz:
## Motivation
Prior to this change, there was no way to get a single Role from the
new AuthzRolesResource. But we would like to display a single role
readonly in the frontend.

## Description
This change will search for a role by id and return it.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
@@ -95,7 +95,6 @@ public class AuthzRolesResource extends RestResource {
     @ApiOperation(value = "Get a single role")
     @Path("/{roleId}")
     @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -88,6 +90,18 @@ public AuthzRolesResource(PaginatedAuthzRolesService authzRolesService) {
         return PaginatedResponse.create("roles", result, query);
     }
 
+
+    @GET
+    @ApiOperation(value = "Get a single role")
+    @Path("/{roleId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @RequiresPermissions(RestPermissions.ROLES_READ)
```
mpfz0r, line 98: this would require the user to have a `roles:read:*` permission.
The specific one underneath is better. I'd just remove this line.
```suggestion
```

PR #8691, opened 2020-08-04
Implementing ES node discovery for ES7.

dennisoelkers:
## Motivation and Context

This PR is adding node discovery for Elasticsearch 7. It uses the default `Sniffer` module which hooks into the rest client and replaces the nodes list with the one retrieved from the server.

In addition, this change is reimplementing a feature from Jest that allows to filter the retrieved list of nodes based on a given
`attribute:value` setting in the config file. For this, it implements the `FilteredElasticsearchNodesSniffer` that wraps around the stock `ElasticsearchNodesSniffer` and filters the nodes list it retrieves based on the `elasticsearch_discovery_filter` setting.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/assembly.xml
+++ b/graylog-storage-elasticsearch7/assembly.xml
@@ -17,6 +17,7 @@
             <includes>
                 <include>org.graylog:graylog-storage-elasticsearch7</include>
                 <include>org.graylog.shaded:elasticsearch7</include>
+                <include>org.graylog.shaded:elasticsearch-rest-client-sniffer7</include>
             </includes>
         </dependencySet>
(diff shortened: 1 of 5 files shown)


dennisoelkers:
Yes, that is the desired behavior. In case of node discovery, the `elasticsearch_hosts` setting is seeding the nodes list. We also need it, because the way the nodes discovery works is that it is polling the cluster stats for the nodes currently present and replaces the current nodes list of the client with these (after filtering them).

kmerz (approved):
Code LGTM :+1: 
Tested Node discovery: Stopped and restarted nodes and the discovery adjusted accordingly :heavy_check_mark:

PR #8692, opened 2020-08-04
Making LDAP TLS/SSL tests faster and more reliable.

dennisoelkers:
## Motivation and Context

This change is doing two things for the LDAP authenticator SSL/TLS tests:

1. Making them more reliable
    Before this change, running the complete test class worked, but running individual tests could result in test failures. This is related to different exception classes being thrown, depending on if a connection was attempted before or not. The real root cause for this is unknown. The exceptions are thrown in two different places:
   - When writing the initial packet (`InvalidConnectionException`), or
   - When parsing the response (`LdapProtocolErrorException`)

    As I was unable to find out the underlying cause, the assertions for failed connections were relaxed. Instead of checking for the exact exception class, we are now checking if an exception was thrown at all.  Together with the positive cases (connection works -> no exception is thrown), this should still provide a reliable test coverage.

2. Making them faster
    Before this change, the container was started/stopped for each test (method). This change is now starting the container once for the complete class, reducing runtime significantly (on my machine from 14.895s down to 1.114s).

## Types of changes

## Checklist:


dennisoelkers:
Wrong target branch, closing.

PR #8707, opened 2020-08-05
Migrate dashboard and stream roles to grants

### LINKED ISSUE #8594: Migrate existing streams and dashboard roles and permissions into grants
## Description

The existing streams and dashboard roles need to be migrated into the new grants system.

For every dashboard or stream permission in an existing role, we create grants for each user that has the role assigned. The migrated permissions will be removed from the role. If the role doesn't have any permissions left after the migration, the role will be removed from each user and deleted.

## Tasks

- [x] Explode role permissions into grants for each user
- [x] Remove stream and dashboard permissions from migrated roles
- [x] Remove role from assigned users and delete role when there are no permissions left in a role
- [ ] Migrate stream and dashboard permissions in user objects into grants

mpfz0r:
This migrates the currently UI configurable roles
for dashboards and streams into grants.

If the permissions of a role are empty after the migration, it gets removed
completely.

If entities has permissions other than read and edit, we don't migrate
it, so we don't end up having permissions on two different places for a
single entity.

Fixes #8594 

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
+++ b/graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
@@ -106,16 +106,20 @@ public class DBGrantService extends PaginatedDbService<GrantDTO> {
                 DBQuery.in(GrantDTO.FIELD_GRANTEE, grantees))).toArray();
     }
 
-    public GrantDTO create(GrantDTO grantDTO, @Nullable User currentUser) {
+    public GrantDTO create(GrantDTO grantDTO, String userName) {
         final ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
-        final String userName = requireNonNull(currentUser, "currentUser cannot be null").getName();
-
         return super.save(grantDTO.toBuilder()
                 .createdBy(userName)
                 .createdAt(now)
                 .updatedBy(userName)
                 .updatedAt(now)
                 .build());
+    }
+
+    public GrantDTO create(GrantDTO grantDTO, @Nullable User currentUser) {
+        final String userName = requireNonNull(currentUser, "currentUser cannot be null").getName();
+
+        return create(grantDTO, userName);
     }
 
     public GrantDTO update(GrantDTO updatedGrant, @Nullable User currentUser) {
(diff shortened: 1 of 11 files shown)


graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
```diff
@@ -89,10 +89,8 @@ public DBGrantService(MongoConnection mongoConnection,
                 DBQuery.in(GrantDTO.FIELD_GRANTEE, grantees))).toArray();
     }
 
-    public GrantDTO create(GrantDTO grantDTO, @Nullable User currentUser) {
+    public GrantDTO create(GrantDTO grantDTO, String userName) {
         final ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
-        final String userName = requireNonNull(currentUser, "currentUser cannot be null").getName();
```
bernd, line 94: Since we have a `String` now, I think we should check that the `userName` is not null **and** not empty.

graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
```diff
@@ -89,10 +89,8 @@ public DBGrantService(MongoConnection mongoConnection,
                 DBQuery.in(GrantDTO.FIELD_GRANTEE, grantees))).toArray();
     }
 
-    public GrantDTO create(GrantDTO grantDTO, @Nullable User currentUser) {
+    public GrantDTO create(GrantDTO grantDTO, String userName) {
```
bernd, line 92: How about adding a test for this? :slightly_smiling_face:

graylog2-server/src/test/java/org/graylog2/migrations/V20200803120800_MigrateRolesToGrantsTest.java
```diff
@@ -0,0 +1,309 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class V20200803120800_MigrateRolesToGrantsTest {
```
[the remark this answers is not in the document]
mpfz0r: cool! thanks

kmerz (changes requested):
For some reason not all of my test roles get deleted.

I tried to debug it, but there they get deleted :(

Let's have a look tomorrow together!

kmerz (approved):
Code looks excellent :heavy_check_mark: 
Works with my :computer: :heavy_check_mark:

PR #8712, opened 2020-08-06
Add missing routes to AuthzRolesResource

kmerz:
## Motivation
Prior this change, there was no way to get all users which had a certain
role assigned. Also the user wanted to add and remove users from the
role.

## Description
With this change three new routes were introduced:
   - get all users for a role
   - add a role to a user
   - remove a role from a user

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
@@ -23,13 +23,20 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import org.apache.shiro.authz.annotation.RequiresAuthentication;
 import org.apache.shiro.authz.annotation.RequiresPermissions;
+import org.graylog2.audit.AuditEventTypes;
+import org.graylog2.audit.jersey.AuditEvent;
 import org.graylog2.database.PaginatedList;
+import org.graylog2.plugin.database.ValidationException;
+import org.graylog2.plugin.database.users.User;
 import org.graylog2.rest.models.PaginatedResponse;
 import org.graylog2.search.SearchQuery;
 import org.graylog2.search.SearchQueryField;
 import org.graylog2.search.SearchQueryParser;
 import org.graylog2.shared.rest.resources.RestResource;
 import org.graylog2.shared.security.RestPermissions;
+import org.graylog2.shared.users.UserService;
+import org.graylog2.users.PaginatedUserService;
+import org.graylog2.users.UserOverviewDTO;
 
 import javax.inject.Inject;
 import javax.validation.constraints.NotBlank;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -126,7 +180,53 @@ public AuthzRoleDTO get(@ApiParam(name = "roleId") @PathParam("roleId") @NotBlan
...
+        final PaginatedList<AuthzRoleDTO> result = authzRolesService.findPaginatedForUser(
+                searchQuery, page, perPage,sort, order, username);
         return PaginatedResponse.create("roles", result, query);
     }
+
+    @PUT
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation("Add a user to a role")
+    @AuditEvent(type = AuditEventTypes.USER_UPDATE)
+    @Path("{roleId}/assignee/add/{username}")
```
bernd, line 192: I think we can use `PUT <role-id>/assignee/<username>` here:
```suggestion
    @Path("{roleId}/assignee/{username}")
```

graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -126,7 +180,53 @@ public AuthzRoleDTO get(@ApiParam(name = "roleId") @PathParam("roleId") @NotBlan
...
+        roles.add(roleId);
+        user.setRoleIds(roles);
+        userService.save(user);
+
+        return Response.ok().build();
+    }
+
+    @PUT
+    @ApiOperation("Remove a member to a team")
+    @Path("{roleId}/assignee/remove/{username}")
```
bernd, line 213: And `DELETE <role-id>/assignee/<username>` here:

```suggestion
    @DELETE
    @ApiOperation("Remove a member to a team")
    @Path("{roleId}/assignee/{username}")
```

graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -126,7 +180,53 @@ public AuthzRoleDTO get(@ApiParam(name = "roleId") @PathParam("roleId") @NotBlan
...
+        if (user == null) {
+            throw new NotFoundException("Cannot find user with name: " + username);
+        }
+        authzRolesService.get(roleId).orElseThrow(() -> new NotFoundException("Cannot find role with id: " + roleId));
+        Set<String> roles = user.getRoleIds();
+        roles.add(roleId);
+        user.setRoleIds(roles);
+        userService.save(user);
+
+        return Response.ok().build();
```
bernd, line 208: I think we should return a `204` if we don't want to return anything. (or make the method `void` instead of returning `Response`, that's the same)
```suggestion
        return Response.noContent().build();
```

graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -126,7 +180,53 @@ public AuthzRoleDTO get(@ApiParam(name = "roleId") @PathParam("roleId") @NotBlan
...
+        if (user == null) {
+            throw new NotFoundException("Cannot find user with name: " + username);
+        }
+        authzRolesService.get(roleId).orElseThrow(() -> new NotFoundException("Cannot find role with id: " + roleId));
+        Set<String> roles = user.getRoleIds();
+        roles.remove(roleId);
+        user.setRoleIds(roles);
+        userService.save(user);
+
+        return Response.ok().build();
```
bernd, line 230: See comment above.
```suggestion
        return Response.noContent().build();
```

graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
```diff
@@ -126,7 +180,53 @@ public AuthzRoleDTO get(@ApiParam(name = "roleId") @PathParam("roleId") @NotBlan
...
+
+        final User user = userService.load(username);
+        if (user == null) {
+            throw new NotFoundException("Cannot find user with name: " + username);
+        }
+        authzRolesService.get(roleId).orElseThrow(() -> new NotFoundException("Cannot find role with id: " + roleId));
+        Set<String> roles = user.getRoleIds();
+        roles.remove(roleId);
+        user.setRoleIds(roles);
+        userService.save(user);
```
bernd, line 228: We have quite some duplication in `#addUser` and `#removeUser`. How about creating a private method to avoid the duplication? :slightly_smiling_face: (e.g. `updateUserRoles(username, roles -> roles.remove(roleId))` or something similar)

PR #8770, opened 2020-08-11
Fix publicPath for file-loader rules in plugin webpack configs

bernd:
Assets like SVG or WOFF files are accessible at the default "/assets"
path. This works for files in the server core but doesn't work for files
in plugins.

For plugins the path needs to be "/assets/plugin/(plugin-name)" to make
sure the web server serves the files from the plugin JAR instead of the
server JAR.

This affects .svg, .woff, .woff2 and .ttf files because those are using
the "file-loader" webpack loader. Images like .png, .gif and .jpg are
using the "url-loader", not sure if they are affected by that bug as
well.

**Note:** This PR should be backported to 3.3 once it has been merged.

### DIFF
--- a/graylog2-web-interface/packages/graylog-web-plugin/src/PluginWebpackConfig.js
+++ b/graylog2-web-interface/packages/graylog-web-plugin/src/PluginWebpackConfig.js
@@ -14,6 +14,23 @@ function getPluginFullName(fqcn) {
   return `plugin.${fqcn}`;
 }
 
+function prefixAssetPaths(fqcn, config) {
+  const rules = (config.module || {}).rules || [];
+
+  rules
+    .filter(rule => rule.loader === 'file-loader') // Only modify file-loader options!
+    .forEach(rule => {
+      if (!rule.options) {
+        rule.options = {};
+      }
+
+      // For file-loader loaders in plugins the default publicPath of
+      // "/assets" doesn't work so we need to make sure we set it to
+      // the plugin's asset path.
+      rule.options.publicPath = '/assets/plugin/' + fqcn;
+    });
+}
+
 function PluginWebpackConfig(fqcn, _options, additionalConfig) {
   const options = merge(defaultOptions, _options);
   const VENDOR_MANIFEST = require(path.resolve(_options.web_src_path, 'manifests', 'vendor-manifest.json'));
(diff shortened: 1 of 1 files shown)


bernd:
I tested this with a production build but I would appreciate if someone else would test them as well so we make sure this doesn't break anything else. Thank you!

bernd:
Okay, let's hold off with this. @thll Just tested a SVG file in the enterprise repository **without** the patch and the path was correct. :thinking:

thll:
This appears to be a special case of the enterprise plugin where the asset path seems to be correct even without the patch applied.
This is not the case for other plugins though. I've tested with `graylog-plugin-collector` where the asset path is broken without the patch. Just checking now, if this patch fixes it for `graylog-plugin-collector` and if it does, we can probably go ahead with this fix.

bernd:
> This appears to be a special case of the enterprise plugin where the asset path seems to be correct even without the patch applied.
> This is not the case for other plugins though. I've tested with `graylog-plugin-collector` where the asset path is broken without the patch. Just checking now, if this patch fixes it for `graylog-plugin-collector` and if it does, we can probably go ahead with this fix.

Thanks for testing! That's funny, I wonder how this works. :smile:

thll:
> Just checking now, if this patch fixes it for graylog-plugin-collector and if it does, we can probably go ahead with this fix.

Yes, this appears to be working. So if we are willing to ignore that the enterprise plugin behaves differently we can merge it.

bernd:
@thll I found the reason why enterprise is working: https://github.com/Graylog2/graylog-plugin-enterprise/blob/master/enterprise/src/web/webpack-entry.js

Gaya:
Makes sense. Basically they same as you are doing, Bernd. In my opinion a patch during the deployment process makes more sense because it only needs to have the assetsPath set to the plugin folder then. Could be that in the future it would bundle everything at the same time instead of per plugin, and they the existing solution in enterprise would fail.

TLDR: my opinion is to go with this PR, remove current enterprise solution.

bernd:
@thll @Gaya I believe the fix in the `webpack-entry.js` file of the enterprise plugins is correct and this PR is not. This PR wouldn't work with a different app prefix. (see `AppConfig.gl2AppPathPrefix()` in [`webpack-entry.js`](https://github.com/Graylog2/graylog-plugin-enterprise/blob/0a333c397bcbb3d2e38077836d1911f8e2fea1dc/enterprise/src/web/webpack-entry.js#L9))

So I am closing this and open a new one in the cloud plugin.

PR #8777, opened 2020-08-12
Make the admin owner of every entity

### LINKED ISSUE #8727: Admin wildcard permission ("*") doesn't work for GRNPermission checks
## Description

The ownership check in `RestResourceWithOwnerCheck` is using a `GRNPermission` instead of a `WildcardPermission`. That means even an admin user (with a `*` permission) needs an `own` grant in the grants table to be able to share an entity.

Admin users shouldn't need a grant for an entity to be able to share it.

mpfz0r:
Convert WildcardPermission("*") to AllPermission 

WildcardPermissions are not comparable with other types of permissions.
In case of the admin permission "*" we want it to imply also
a GRNPermission, which is used for handling entity ownership.

Having the AllPermission allows a User to "own" every entity
inside Graylog.

Improve error message

Fixes #8727 

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/permissions/GRNPermission.java
+++ b/graylog2-server/src/main/java/org/graylog/security/permissions/GRNPermission.java
@@ -19,7 +19,6 @@ package org.graylog.security.permissions;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.auto.value.AutoValue;
 import org.apache.shiro.authz.Permission;
-import org.apache.shiro.authz.permission.WildcardPermission;
 import org.graylog.grn.GRN;
 
 @AutoValue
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog/security/permissions/GRNPermission.java
```diff
@@ -33,7 +34,14 @@ public static GRNPermission create(String type, GRN target) {
 
     @Override
     public boolean implies(Permission p) {
-        // By default only supports comparisons with other GRNPermission
+        // The admin permission also implies access to every GRNPermission
+        if ((p instanceof WildcardPermission)) {
```
bernd, line 38: This doesn't work, unfortunately. The permission `p` that is passed in is the `entity:own` `GRNPermission`: `GRNPermission{type=entity:own, target=grn::::stream:000000000000000000000003}`

That would mean we would need this check in the `WildcardPermission` class instead. (which we cannot do because it's part of shiro)

I am afraid we need to find a different solution.

mpfz0r: You're right. It only worked because I was testing it with the admin user itself, not with a regular user that has the admin role.

Issue #8392, opened 2020-06-19
Race condition on `/system/configurations` page


kyleknighted:
## Current Behavior

![image](https://user-images.githubusercontent.com/122591/85170598-3e81ed80-b233-11ea-814f-4285a5aab171.png)

```

Warning: Failed prop type: The prop `config` is marked as required in `UrlWhiteListConfig`, but its value is `null`.
    in UrlWhiteListConfig (created by ConfigurationsPage)
    in ConfigurationsPage (created by ConnectStoresWrapper[ConfigurationsPage] stores=configurations,currentUser)
    in ConnectStoresWrapper[ConfigurationsPage] stores=configurations,currentUser (created by LoadableComponent)
    in LoadableComponent (created by RouterContext)
    in div (created by Col)
    in Col (created by AppWithoutSearchBar)
    in div (created by Row)
    in Row (created by Context.Consumer)
    in StyledComponent (created by AppWithoutSearchBar__StyledRow)
    in AppWithoutSearchBar__StyledRow (created by AppWithoutSearchBar)
    in div (created by Grid)
    in Grid (created by AppContentGrid)
    in div (created by Context.Consumer)
    in StyledComponent (created by AppContentGrid__Container)
    in AppContentGrid__Container (created by AppContentGrid)
    in AppContentGrid (created by AppWithoutSearchBar)
    in AppWithoutSearchBar (created by RouterContext)
    in div (created by AppWithGlobalNotifications)
    in AppWithGlobalNotifications (created by RouterContext)
    in RuntimeErrorBoundary (created by App)
    in ReportedErrorBoundary (created by withRouter(ReportedErrorBoundary))
    in withRouter(ReportedErrorBoundary) (created by App)
    in ScratchpadProvider (created by App)
    in App (created by ConnectStoresWrapper[App] stores=currentUser)
    in ConnectStoresWrapper[App] stores=currentUser (created by RouterContext)
    in RouterErrorBoundary (created by RouterContext)
    in RouterContext (created by Router)
    in Router (created by AppRouter)
    in AppRouter (created by LoggedInPage)
    in CurrentUserPreferencesProvider (created by LoggedInPage)
    in LoggedInPage (created by LoadableComponent)
    in LoadableComponent (created by AppFacade)
    in AppFacade (created by ConnectStoresWrapper[AppFacade] stores=currentUser,server,sessionId)
    in ConnectStoresWrapper[AppFacade] stores=currentUser,server,sessionId
[...]

PR #8778, opened 2020-08-12
Fix race conditions on system/configuration

Gaya:
Will prevent problem occurring in #8392

The `state.loaded` only checks _if_ there is some config available, not which.

This patch will allow each configuration block to load individually as data comes in.

## Types of changes

## Checklist:


graylog2-web-interface/src/pages/ConfigurationsPage.jsx
```diff
@@ -29,12 +29,14 @@ const EVENTS_CONFIG = 'org.graylog.events.configuration.EventsConfiguration';
...
 
 class ConfigurationsPage extends React.Component {
-  state = {
-    loaded: false,
-  }
-
   checkLoadedTimer = undefined
 
+  constructor() {
+    super();
```
kyleknighted, line 35: Don't we need to pass `props` through?

```suggestion
  constructor(props) {
    super(props);
```

Gaya: Right! I always forget that. Been aged since I used component classes 😅

PR #8819, opened 2020-08-18
Removing tests snapshotting layout excessively.

dennisoelkers:
## Motivation and Context

This change is removing snapshots containing too much layout, which is not part of the system under test. Those are mostly tests which render a component and throw all of the returned markup against `toMatchSnapshot`, resulting in huge snapshots, which do not provide us a benefit going beyond (ideally) a check that the component renders at all.

The additional downsides of these are:

  * Polluting PRs which change theming or central components, easily blowing up n * 10 loc changes to n * 10000 loc changes, making a proper code review impossible and effectively shadowing the actual changes.
  * Due to the size of the snapshots, test negatives are dismissed too easily during development. Instead of investigating big diffs, snapshots are just updated, rendering the mechanism ineffective.
  * The existence of snapshot tests suggest (falsely) proper coverage of the SUT. Instead, behavior-focused tests should be used.

Tests which used these snapshots where left in place, as they do actually test something: if the rendering of the component succeeds at all. Therefore the test expecations in those was changed to use `.toExist()`/`.not.toBeNull()` instead of `.toMatchSnapshot()`.

This should not suggest that snapshot tests should not be used _at all_. They are useful for cases where isolated, stable results are checked. This includes functions operating on bigger data structures, where generating test expectations using snapshots are feasible and the results are stable.

## Types of changes

## Checklist:


Gaya:
Will this interfere with #8813?

dennisoelkers:
@Gaya: Generally no, but the one which is merged first forces the other one to be rebased. Merging this one first makes parts of #8113 obsolete, merging #8113 before is probably easier, as rebasing this one is straight-forward.

Gaya:
@dennisoelkers then I think it's up to you and @kyleknighted to figure it out. Looks good to me!

kmerz (approved):
LGTM :+1: 
Lets snaps go the way of the dodo! :heavy_check_mark:

PR #8824, opened 2020-08-19
Caching singleton instance of `JestClient` in `JestClientProvider`.

dennisoelkers:
## Motivation and Context

This change is made to avoid initializing a `JestClient` instance when it is not used at all (e.g. when the ES6 storage driver is not used).  Declaring the binding as eager singleton or even with the singleton scope results in an eager call to `JestClientProvider#get`, resulting in a creation of a `JestClient` instance and initializing the node discovery.
    
To avoid this, this change is removing any singleton declaration for the binding and caches the instance in the provider instead, always returning the same one.

## Types of changes

## Checklist:


mpfz0r (approved):
Tested and LGTM

PR #8829, opened 2020-08-20
Remove legacy users and roles

kmerz:
## Description
Remove old Users and Roles pages from the Authentication Page

## Motivation
We implemented new pages so the old ones are obsolete.


linuspahl:
Looks good to me, two things we've talked about:
* Changing the permissions for 
```
<IfPermitted permissions={['users:list', 'roles:read']} anyPermissions>
    <NavigationLink path={Routes.SYSTEM.AUTHENTICATION.OVERVIEW} description="Authentication" />
</IfPermitted>
```
* Changing the description for the authentication page
`Configure Graylog's authentication providers and manage the active users of this Graylog cluster`

PR #8836, opened 2020-08-20
Display confirm dialog on entity share modal save if grantees select is not empty

### LINKED ISSUE #8835: Implement confirm dialog for entity share modal close button, if the grantees assignment select is not empty
When you assign a new grantee in the entity share modal it is easy to forget if you clicked on the "Assign grantee" button. We should implement a confirm dialog for the save button, which appears if the grantees assignment select is not empty.

linuspahl:
When you assign a new grantee in the entity share modal it is easy to forget if you clicked on the "Assign grantee" button. With this PR we've implement a confirm dialog which appears if the grantees assignment select is not empty on save.

Fixes #8835


graylog2-web-interface/src/components/permissions/EntityShareModal.jsx
```diff
@@ -25,18 +25,40 @@ const EntityShareModal = ({ description, entityId, entityType, entityTitle, onCl
...
+        setDisableSubmit(false);
+
+        return;
+      }
+    }
+
+    EntityShareActions.update(entityGRN, payload).then(() => {
+      setDisableSubmit(true);
+      onClose();
+    });
```
linuspahl, line 61: I tried to add a more complex test but this is not possible with due to the mocking and usage of the `EntityShareStore`. We will be able to handle the mocking better when we use the return value of our actions inside the `EntityShareModal` instead of the state in the `EntityShareState`

PR #8847, opened 2020-08-21
Fix flickering of theme

kmerz:
## Motivation
Prior to this change, when opening a popover, the opened popover would
flicker white then back to dark when in dark mode.

## Description
This change will do not set a default state for the mode. Which will
prevent the flickering.

### DIFF
--- a/graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
+++ b/graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
@@ -28,7 +28,7 @@ const GraylogThemeProvider = ({ children }: Props) => {
   const [theme, setTheme] = useState();
   const [userPreferences, setUserPreferences] = useState();
   const [themeColors, setThemeColors] = useState();
-  const [mode, setMode] = useState();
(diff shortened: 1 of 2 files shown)


dennisoelkers (changes requested):
Wouldn't it be better to keep the implicit `undefined` for the theme state? This would allow us to express explicitly on a typing level that the theme might not be set.

kyleknighted:
@dennisoelkers I like `undefined` as well before the render, but not a fan of the typings allowing `undefined` because I felt like an undefined mode in the theme should cause an error as well. What do you propose as a good solution? Is the type allowing an `undefined` the best way?

dennisoelkers:
@kyleknighted: IMO we should keep a strict, non-optional typing in `ThemeProvider`, but for the `GraylogThemeProvider` we already kind of have a non-defined initial state, we just make it implicit by using `''`. Instead we should allow `mode` to be `undefined` and deal with that case  in the related places.

kmerz:
This is fixed by another PR.

PR #8848, opened 2020-08-21
Set font size defined in theme for query input

### LINKED ISSUE #8720: make the keyboard cursor tight in the search textbox
## What?
 the keyboard cursor is not tight in the search box. To my eyes, it looks like , there's a space already in there, when I finish typing `AND`.
compare keyboard cursor with  `ms word` or in github feature request `text area`

keyboard_cursor

## Why?
i had to type backspace , to fix the space. 

## Steps to reproduce
type inside the search text area.

## Your Environment

* Graylog Version:4.0.0-SNAPSHOT (branch: master)
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

linuspahl:
This PR fixes a problem with long queries where uncontrollable whitespace is being added before the cursor.
image
In the above screenshot `query` is selected, but it looks like the selected text is `ery  `.

Fixes #8720.

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
@@ -105,4 +105,6 @@ QueryInput.defaultProps = {
   placeholder: '',
 };
 
-export default withPluginEntities(withTheme<ThemeInterface, Props, typeof undefined>(QueryInput), { completers: 'views.completers' });
+const mapping = { completers: 'views.completers' };
+
+export default withPluginEntities<$Rest<Props, {| theme: ThemeInterface |}>, typeof mapping>(withTheme(QueryInput), mapping);
--- a/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
@@ -16,11 +16,11 @@ import type { Completer } from './SearchBarAutocompletions';
 type Props = {|
   completerFactory: (Array<Completer>) => AutoCompleter,
   completers: Array<Completer>,
-  disabled: boolean,
+  disabled?: boolean,
   onBlur?: (string) => void,
   onChange: (string) => Promise<string>,
   onExecute: (string) => void,
-  placeholder: string,
+  placeholder?: string,
   theme: ThemeInterface,
   value: string,
 |};
@@ -105,4 +105,4 @@ QueryInput.defaultProps = {
   placeholder: '',
 };
 
-export default withTheme(withPluginEntities(QueryInput, { completers: 'views.completers' }));
+export default withPluginEntities(withTheme<ThemeInterface, Props, typeof undefined>(QueryInput), { completers: 'views.completers' });
--- a/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
+++ b/graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
@@ -4,7 +4,7 @@ import { useCallback, useMemo } from 'react';
 import { withTheme } from 'styled-components';
 import PropTypes from 'prop-types';
 
-import type { ThemeInterface } from 'theme';
+import { type ThemeInterface, themePropTypes } from 'theme';
 import withPluginEntities from 'views/logic/withPluginEntities';
 import UserPreferencesContext from 'contexts/UserPreferencesContext';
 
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
```diff
@@ -104,4 +104,4 @@ QueryInput.defaultProps = {
   placeholder: '',
 };
 
-export default withPluginEntities(QueryInput, { completers: 'views.completers' });
+export default withTheme(withPluginEntities(QueryInput, { completers: 'views.completers' }));
```
linuspahl, line 107: Initially I wrapped `QueryInput` with `withTheme`,  but it results in a flow error I can't fix easily:
```
Cannot call withPluginEntities because property completers is missing in object type [1] in the first argument.
[prop-missing]

     src/views/components/searchbar/QueryInput.jsx
     105│   placeholder: '',
     106│ };
     107│
     108│ export default withPluginEntities(withTheme(QueryInput), { completers: 'views.completers' });
     109│

     flow-typed/npm/styled-components_v4.x.x.js
 [1] 175│   declare export function withTheme<Theme, Config: {...}, Instance>(Component: React$AbstractComponent<Config, Instance>): React$AbstractComponent<$Diff<Config, ThemeProps<Theme | void>>, Instance>
```

I prefer the above structure and we do it similar in comparable cases where we use `withPluginEntities`  (e.g. in combination with `withRouter`), but I see no other disadvantage regarding the current structure.

dennisoelkers: It is fine this way, imo. It should work both ways around though, so I took a quick look into the issue. There are two things: 

a) we are still using `styled-components` for v4. They do not work with functional components (like `QueryInput` is), so the type inference for the component does not work. I was able to force it by passing explicit types for the generics like this:

```
withPluginEntities(withTheme<ThemeInterface, Props, typeof undefined>(QueryInput), { completers: 'views.completers' })
```

b) updating the typedefs to v5 fixed this issue and made it work with `withPluginEntities(withTheme(QueryInput)` out of the box. Can you do the typedef update either in this or a separate PR?

Hope this helps.

graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
```diff
@@ -12,19 +14,20 @@ import SearchBarAutoCompletions from './SearchBarAutocompletions';
...
-  value: string,
-  completers: Array<Completer>,
   completerFactory: (Array<Completer>) => AutoCompleter,
+  completers: Array<Completer>,
+  disabled: boolean,
   onBlur?: (string) => void,
   onChange: (string) => Promise<string>,
   onExecute: (string) => void,
   placeholder: string,
+  theme: ThemeInterface,
```
kyleknighted, line 24: Could we add `theme: themePropTypes.isRequired` to the `propTypes` object as well?

graylog2-web-interface/src/views/components/searchbar/QueryInput.jsx
linuspahl: Thanks again for the feedback, I've tested solution b) locally by using https://github.com/Graylog2/graylog2-server/pull/8866 and I am still receiving the flow type error with the `styled-components` v5 interface definition:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/views/components/searchbar/QueryInput.jsx:108:16

Cannot call withPluginEntities because property completers is missing in object type [1] in the first argument.
[prop-missing]

     src/views/components/searchbar/QueryInput.jsx
     105│   placeholder: '',
     106│ };
     107│
     108│ export default withPluginEntities(withTheme(QueryInput), { completers: 'views.completers' });
     109│

     flow-typed/npm/styled-components_v5.x.x.js
 [1] 223│   declare export function withTheme<Theme, Config: { ... }, Instance>(

```
Do you have an idea, why this behaves differently? 

I also Implemented solution a) and realised we need to define explicit types when using `withPluginEntities`. Under these circumstances I prefer the following solution:
```
const mapping = { completers: 'views.completers' };

export default withPluginEntities<$Rest<Props, {| theme: ThemeInterface |}>, typeof mapping>(withTheme(QueryInput), mapping);

dennisoelkers (approved):
LGTM and works! Good work!

PR #8853, opened 2020-08-24
Fix darkmode color of tableheader of data tables

kmerz:
## Description
Fix darkmode color of tableheader of data tables

### Before
Graylog (11)

### After
Graylog (10)


kyleknighted:
Light Mode
image

PR #8864, opened 2020-08-25
Improve `GraylogThemeProvider` rendering by streamlining state handling.

dennisoelkers:
## Motivation and Context

This PR is refactoring the `GraylogThemeProvider` component's state handling to reduce the number of renderings required until it stabilizes (initially and after a theme change was triggered) and improve its readability.

For this, it removes the cascade of dependant `useEffect` + setStates and replaces them with a) straight variable assignments or b) `useMemo`. This allows us to reduce rendering counts from initially 10 and 4 upon a theme change to initially 3 and a single render for a theme change. For better readibility, the handling of the user theme preference (getting and setting it) is extracted into the `useCurrentThemeMode` hook.

Some additional cleanup was performed, also the `debounce` was removed as it seemed to be unneeded, because the toggle is disabled while the new theme is loading. If it turns out to be still required, it should be changed to use `{ leading: true, trailing: false }` options, so it is preventing subsequent clicks but does not waste time by waiting for the trailing edge.

## Types of changes

## Checklist:


kmerz (approved):
LGTM :+1: 
Works on my :computer:

graylog2-web-interface/src/components/navigation/ThemeModeToggle.jsx
```diff
@@ -2,7 +2,7 @@
 import * as React from 'react';
 import { useEffect, useState } from 'react';
 import styled, { css, withTheme, type StyledComponent } from 'styled-components';
-import debounce from 'lodash/debounce';
+import { defer } from 'lodash';
```
kyleknighted, line 5: To enable tree-shaking in lodash, you have to import functions using syntax like `import foo from 'lodash/foo'`. 
`import { foo } from 'lodash'` will not tree-shake.
```suggestion
import defer from 'lodash/defer';
```

dennisoelkers: :white_check_mark:

PR #8868, opened 2020-08-25
Search Popovers and Tooltips

kyleknighted:
## Description

- Reposition DayPicker error Tooltip
- Update Popover right and left arrow colors to match the background
- Updated Save Search Popover buttons to be smaller and resized width to make sure they all fit in one line

## Motivation and Context

Some issues with Popover and Tooltip within the Search bar 

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
+++ b/graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
@@ -37,13 +37,13 @@ const createTheme = (mode, themeColors, changeMode): ThemeInterface => {
   };
 };
 
-const GraylogThemeProvider = ({ children }: Props) => {
-  const colorScheme = usePrefersColorScheme();
-  const [mode, setCurrentThemeMode] = useCurrentThemeMode(colorScheme);
+const GraylogThemeProvider = ({ children, defaultMode }: Props) => {
+  const [mode, setCurrentThemeMode] = useCurrentThemeMode();
+  const themeMode = defaultMode ?? mode;
 
-  const themeColors = colors[mode];
+  const themeColors = colors[themeMode];
 
-  const theme = useMemo(() => (themeColors ? createTheme(mode, themeColors, setCurrentThemeMode) : undefined), [themeColors]);
+  const theme = useMemo(() => (themeColors ? createTheme(themeMode, themeColors, setCurrentThemeMode) : undefined), [themeMode, themeColors]);
 
   return theme ? (
     <ThemeProvider theme={theme}>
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
```diff
@@ -37,13 +37,13 @@ const createTheme = (mode, themeColors, changeMode): ThemeInterface => {
   };
 };
 
-const GraylogThemeProvider = ({ children }: Props) => {
-  const colorScheme = usePrefersColorScheme();
-  const [mode, setCurrentThemeMode] = useCurrentThemeMode(colorScheme);
+const GraylogThemeProvider = ({ children, defaultMode }: Props) => {
+  const [mode, setCurrentThemeMode] = useCurrentThemeMode();
+  const themeMode = defaultMode ?? mode;
```
kyleknighted, line 42: Maybe a conditional here, if `defaultMode` is set, no reason to do the `useCurrentThemeMode`?

dennisoelkers: That would break the first rule for hooks, not calling them from the top-level/unconditionally imo. Might work right now, but could introduce subtle bugs in the future if `GraylogThemeProvider` gets more state.

graylog2-web-interface/src/theme/GraylogThemeProvider.jsx
```diff
@@ -37,13 +37,12 @@ const createTheme = (mode, themeColors, changeMode): ThemeInterface => {
   };
 };
 
-const GraylogThemeProvider = ({ children }: Props) => {
-  const colorScheme = usePrefersColorScheme();
-  const [mode, setCurrentThemeMode] = useCurrentThemeMode(colorScheme);
+const GraylogThemeProvider = ({ children, defaultMode }: Props) => {
+  const [mode, setCurrentThemeMode] = defaultMode ? [defaultMode, () => {}] : useCurrentThemeMode();
```
dennisoelkers, line 41: IMO this is a no-go for hooks, which should always be called unconditionally. Instead, we could pass in the `defaultMode` and use it accordingly.

graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchForm.jsx
```diff
@@ -43,7 +43,7 @@ const SavedSearchForm = (props: Props) => {
                 placement="left"
                 target={target}>
         <Popover title="Name of search" id="saved-search-popover">
-          <form onSubmit={stopEvent}>
+          <form onSubmit={stopEvent} style={{ width: '210px' }}>
```
dennisoelkers, line 46: Should we extract this into a styled component instead of using an inline style?

dennisoelkers (changes requested):
Fixes the issues I was seeing, added two comments inline.

dennisoelkers (approved):
LGTM and works! Thanks for the fix!

PR #8878, opened 2020-08-27
Add type definition for user create and update payload

linuspahl:
Right now the payload for the user create and update request is typed as `any`. This PR adds proper type definitions for the actions and store methods.


kmerz (approved):
Code LGTM :+1: 
Works on my machine :+1:

PR #8880, opened 2020-08-27
Auto-sense Elasticsearch version if no version is set.

dennisoelkers:
## Motivation and Context

This change is adding auto-sensing for the Elasticsearch version if the `elasticsearch_version` configuration field is not set. The config field is kept, but optional now. It can be used to explicitly override the chosen Elasticsearch storage driver implementation if auto-sensing fails.

If the config setting is not present, early at startup version probing will be attempted by iterating over the list of configured Elasticsearch nodes and requesting their root resource (`/`). All fields of the response will be ignored, besides the `$.version.number` field, which contains the server version as a string. The first successful attempt where the request succeeds and the retrieved version can be parsed properly will be used to determine the remote version and therefore which storage driver module is used. If no request succeeds, one or more errors will be logged and startup will fail.

## Types of changes

## Checklist:


kmerz (approved):
Works as designed :heavy_check_mark: 
Code LGTM :+1:

PR #8882, opened 2020-08-27
Wrap app with GlobalStylesProvider to amend global css in-flight

kyleknighted:
## Description

Moved Global Theme Styles to a Context Provider so a single instance can be manipulated 

## Motivation and Context

Global styles were duplicating on error pages

## Types of changes

## Checklist:


graylog2-web-interface/src/pages/LoadingPage.jsx
```diff
@@ -1,15 +1,20 @@
...
 import authStyles from 'theme/styles/authStyles';
-import GlobalThemeStyles from 'theme/GlobalThemeStyles';
+import { GlobalStylesContext } from 'contexts/GlobalStylesProvider';
 
 const LoadingPage = ({ text }) => {
+  const { addGlobalStyles } = useContext(GlobalStylesContext);
+
+  useEffect(() => {
+    addGlobalStyles(authStyles);
+  }, []);
```
linuspahl, line 14: We need to reset the additional styles here as well.

graylog2-web-interface/src/contexts/GlobalStylesProvider.jsx
```diff
@@ -1,6 +1,31 @@
...
+import { createContext, useState } from 'react';
+import PropTypes from 'prop-types';
+import { createGlobalStyle, css, type CSSRules } from 'styled-components';
+
+type Props = {
+  children: React.Node,
+};
+
+type StyleTypes = {
+  addGlobalStyles: (?CSSRules) => ?string,
```
linuspahl, line 12: This should be `(?CSSRules) => void`

graylog2-web-interface/src/contexts/GlobalStylesProvider.jsx
```diff
@@ -1,6 +1,31 @@
...
+};
+
+export const GlobalStylesContext = createContext<StyleTypes>({ addGlobalStyles: () => {} });
+
+const GlobalStylesProvider = ({ children }: Props) => {
+  const [additionalStyles: string, addGlobalStyles] = useState();
```
linuspahl, line 18: Let's unify this type with the type used in line 12 and define the type like this: `const [additionalStyles, addGlobalStyles] = useState<?string>();`

linuspahl (changes requested):
Found one bug with the `LoadingPage` and some nitpicks, otherwise it looks good! I really like the refactorings.

As described the global styles are not being rendered twice on the error page anymore, which fixes the occasionally failing `RuntimeErrorPage.test.jsx`.

graylog2-web-interface/src/pages/LoadingPage.jsx
kyleknighted: Reset in `LoggedInPage` during useEffect mount

dennisoelkers:
@kyleknighted: Could you please make the title more helpful in understanding what this PR is doing?

linuspahl (approved):
Your update fixes the problem I had previously. I saw that the loading page background image is still being displayed, when the navigation is already visible. This looked slightly better before, but we can may improve this in the future and it is not a blocker for me.

This is also related to a topic I thought about while looking at these changes. One disadvantage with this implementation is that you have no control, in a component, what global styles you are overwriting or resetting with `addGloblalStyles`. 

I had the idea to define a key for the styles you want to add / remove, this could also improve the initially described behaviour. I gave it a try: https://github.com/Graylog2/graylog2-server/compare/GlobalStylesProvider...GlobalStylesProvider-alternative?expand=1

But right now the state does not persist, e.g. when the `LogginPage` is displayed after the LoadingPage`. Would be great to have a look together at this alternative and the problem.

linuspahl:
@kyleknighted Yesterday we talked about merging this PR once we've implemented the changes described in my review comment. Right now I prefer to do this in a follow-up PR because, this PR fixes the `RuntimeErrorPage.test.jsx` which is something other PR's will benefit from.

PR #8885, opened 2020-08-28
Prevent to remove last owner

### LINKED ISSUE #8841: Ensure that the last owner cannot be removed from EntityShareDialog
## Why
The backend will throw an error if the last owner will be remove from an entity share.
That error we could handle, but is not nice UX wise.

## What
We should somehow mark the last user and prevent the user to delete him in the first place!

kmerz:
## Motivation
The backend provides us with a validation which ensures that the last owner can't be removed.

## Description
- Extend `EntityShareState` value class for `ValidationResults`
- Add new component which will display the `ValidationResult` if the validation `failed`
- Disable the submit button if the Validation `failed.

Fixes #8841 


graylog2-web-interface/src/components/permissions/ValidationError.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+const ValidationError = ({ validationResult, availableGrantees }: Props) => {
+  const pastOwners = validationResult.errorContext.selectedGranteeCapabilities.toJS().map(
+    (grn) => availableGrantees.find((grantee) => grantee.id === grn)?.title,
+  ).join(', ');
+
+  return (
+    <Container bsStyle="danger">
+      <List>
+        <li>{`Removing the following owners ${pastOwners} will leave the entity ownerless.`}</li>
```
linuspahl, line 38: We could structure this message similar to the dependency warning: 
```
<li>
  Removing the following owners will leave the entity ownerless:</br>
  {pastOwners.map((owner, key) => (
    <span key={owner.id}>
      {_cap(owner.type)} <i>{owner.title}</i>
      {key !== pasOwners.size - 1 && ', '}
    </span>
  ))}
</li>
```
`pastOwners` is an array of Grantees in this case.
This way the message is still readable, even if there are many previous owners, the owner names can be identified easier, because of the `<i>` and it is consistent compared with the dependency warning.

graylog2-web-interface/src/components/permissions/ValidationError.jsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+`;
+
+type Props = {
+  validationResult: ValidationResult,
+  availableGrantees: GranteesList,
+};
+
+const ValidationError = ({ validationResult, availableGrantees }: Props) => {
+  const pastOwners = validationResult.errorContext.selectedGranteeCapabilities.toJS().map(
```
linuspahl, line 31: `.toJS()` is not necessary in this case.

graylog2-web-interface/src/components/permissions/EntityShareSettings.jsx
```diff
@@ -111,14 +116,20 @@ const EntityShareSettings = ({
                       selectedGrantees={selectedGrantees}
                       title="Current collaborators" />
       </Section>
+      {validationResults && validationResults.failed && (
```
linuspahl, line 119: `{validationResults?.failed && (` would be suitable as well.

linuspahl (commented):
Works very well, only added some nitpicks and suggestions.

linuspahl (approved):
LGTM! We can merge the PR when the checks finished. Looks like the linter warnings are unrelated.

PR #8886, opened 2020-08-28
Define theme input background color for Scratchpad textarea

linuspahl:
As described in the title this PR defines the theme input background color for the Scratchpad textarea. Right now the text is not visible, using the darkmode:
image


linuspahl:
I am closing this PR, because I came to the conclusion that there is no disadvantage in defining the background color for all textareas in `GlobalThemeStyles`.

PR #8910, opened 2020-09-03
Remove GlobalStylesProvider and use createGlobalStyles

kyleknighted:
## Description

Removes `GlobalStylesProvider`, replaced with directly using Styled-Components `createGlobalStyles`	where needed

## Motivation and Context

Login/Loading pages and sometimes after an error, the background was not resetting properly.
First we attempted the difficult way, then we realized it was being overthought and went for the simple solution.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/LoadingPage.jsx
+++ b/graylog2-web-interface/src/pages/LoadingPage.jsx
@@ -1,16 +1,21 @@
+// @flow strict
+import * as React from 'react';
 import PropTypes from 'prop-types';
-import React from 'react';
 import { createGlobalStyle } from 'styled-components';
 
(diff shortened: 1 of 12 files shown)


graylog2-web-interface/src/pages/LoadingPage.jsx
```diff
@@ -1,20 +1,19 @@
 import PropTypes from 'prop-types';
-import React, { useContext, useEffect } from 'react';
+import React from 'react';
```
linuspahl, line 2: I prefer to use `import * as React from 'react';` in general.

linuspahl (changes requested):
Found a variable which needs a different name, otherwise it looks good.

linuspahl (approved):
Looks good to me!

PR #8926, opened 2020-09-07
Add is_cloud flag and disable "No input running" warning

danotorrey:
## Description
Add an `is_cloud` configuration flag. When enabled, certain in functionality/behavior in Graylog can be disabled or altered as appropriate for running in the Cloud environment.

This PR also skips checks in the `ClusterHeathCheckThread`, which currently only raise a System Notification when no inputs are running on a node. This notification is not appropriate for Cloud, since Forwarders/Forwarder Inputs are used instead of Inputs. So, a node in Graylog Cloud can be healthy when no inputs are running on it.

## How Has This Been Tested?
Enabled the `is_cloud` flag and observed that the system notification is not raised.

Disabled the `is_cloud` flag, and observed that the notification is raised.

## Types of changes

Closes https://github.com/Graylog2/graylog-plugin-cloud/issues/359

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
@@ -52,7 +52,6 @@ public class ClusterHealthCheckThread extends Periodical {
     @Override
     public void doRun() {
         if (isCloud) {
-            LOG.info("Skipping run of ClusterHealthCheckThread, since contained checks are not applicable for Cloud.");
             return;
         }
         try {
@@ -111,7 +110,7 @@ public class ClusterHealthCheckThread extends Periodical {
     public int getInitialDelaySeconds() {
         // Wait some time until all inputs have been started otherwise this will trigger a notification on every
         // startup of the server.
-        return 1;
+        return 120;
     }
 
     @Override
--- a/graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
@@ -26,6 +26,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 
 /**
  * @author Dennis Oelkers <dennis@torch.sh>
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
```diff
@@ -35,18 +36,25 @@
...
         this.notificationService = notificationService;
         this.inputRegistry = inputRegistry;
         this.nodeId = nodeId;
+        this.isCloud = isCloud;
     }
 
     @Override
     public void doRun() {
+        if (isCloud) {
+            LOG.info("Skipping run of ClusterHealthCheckThread, since contained checks are not applicable for Cloud.");
```
bernd, line 55: I think we don't need to log this to info. If we do this in every place we use the flag, we will have a lot of messages at some point. :wink:

graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
```diff
@@ -103,7 +111,7 @@ public boolean startOnThisNode() {
     public int getInitialDelaySeconds() {
         // Wait some time until all inputs have been started otherwise this will trigger a notification on every
         // startup of the server.
-        return 120;
+        return 1;
```
bernd, line 114: Why did you reduce this to `1`?

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
danotorrey: Ack, accidental commit. Thanks for seeing this.

graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
```diff
@@ -35,18 +36,24 @@
...
         this.notificationService = notificationService;
         this.inputRegistry = inputRegistry;
         this.nodeId = nodeId;
+        this.isCloud = isCloud;
     }
 
     @Override
     public void doRun() {
+        if (isCloud) {
+            return;
```
bernd, line 55: I meant moving the log statement to debug. Now there is no explanation why we do this. :wink:

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/periodical/ClusterHealthCheckThread.java
danotorrey: Thanks. This was the original intention, but somehow I left it at info. Makes sense, I see that now.

bernd (approved):
LGTM. :+1: Feel free to merge after the build is green.

PR #8932, opened 2020-09-08
Add flow type definitions, create test and use styled-components for EmptyEntity

linuspahl:
With this PR we are adding flow type definitions, creating a test and using styled-components for the `EmptyEntity` component.


ousmaneo (approved):
Legit. Good Improvement.

PR #8933, opened 2020-09-08
Add flow type definitions for Wizard

linuspahl:
This PR adds flow type definitions for the Wizard component.


ousmaneo (approved):
Legit !

PR #8940, opened 2020-09-09
Add shared form helper components, e.g. to wrap the common Input with a formik Field.

linuspahl:
This PR adds some helper components for our forms. This includes:
- `FormikInput`: Wraps the common `Input` component with a formik `Field`
- `FormikFormGroup`: Uses the `FormikInput` to display a label and an input with a specific layout
image
- `ReadOnlyFormGroup`: It is a common use case that some fields are not editable (e.g. on a details page) but needs to be displayed with the same layout like the `FormikFormGroup`.
image

These components simplify the usage of the formik `Field`. E.g. the full name input can be added with:

```
<FormikFormGroup label="Full Name"
                 name="full_name"
                 maxLength={200}
                 required
                 help="Give a descriptive name for this account, e.g. the full name." />
```

Please note that the `FormikInput` does not support multiple checkboxes with the same name yet, when needed it can be implemented like this: https://codesandbox.io/s/328038x19q


graylog2-web-interface/src/components/common/FormikFormGroup.jsx
```diff
@@ -0,0 +1,31 @@
...
+type Props = {
+  label: string,
+  name: string,
+  type?: string,
+  help?: string,
+  validate?: (string) => ?string,
+};
+
+const FormikFormGroup = ({ label, name, type, help, validate, ...rest }: Props) => (
+  <FormikInput {...rest}
```
Gaya, line 15: If you start with `{...rest}` you can never overwrite all the following props, is that intended?

graylog2-web-interface/src/components/common/ReadOnlyFormGroup.jsx
```diff
@@ -0,0 +1,63 @@
...
+  return '-';
+};
+
+const ReadOnlyFormGroup = ({ label, value, help }: Props) => (
+  <Row>
+    <LabelCol sm={3}>
+      {label}
+    </LabelCol>
+    <ValueCol sm={9}>
+      {readableValue(value)}
```
Gaya, line 53: You're only rendering text here in a container that looks like a form input. Maybe an `<input disabled />` would work better?

Gaya (approved):
Implementation looks good to me, just some minor remarks to be considered.

graylog2-web-interface/src/components/common/FormikFormGroup.jsx
linuspahl: In this case I have to structure it like that, otherwise flow complains:
`Cannot spread props because Flow cannot determine a type for  props [1].  rest of object pattern [2] is inexact, so it may contain `value` with a type that conflicts with `value`'s definition in  props [1]. Try making  rest of object pattern [2] exact.`

graylog2-web-interface/src/components/common/ReadOnlyFormGroup.jsx
linuspahl: In the context of the screenshot I posted, a disabled input would probably be the better option. But we will mainly use this component on a details pages where we just need this form group column and label styling. On these pages I prefer to display just text instead of a disabled input.

PR #8976, opened 2020-09-16
Reduce auth image background size - 47% smaller

kyleknighted:
Compressing NASA control background image from authentication steps


Gaya (approved):
🏋️‍♂️

PR #8977, opened 2020-09-16
Fix path traversal in swagger resource.

### LINKED ISSUE #8986: Path traversal vulnerability in documentation resource
**This vulnerability does not apply to the standard installation methods.**

When starting Graylog with the JVM `-cp` classpath options, instead of using the default `-jar` for the bundled classes, the API REST endpoint `/api-browser/.*` would allow path traversal to allow reading files on the server's filesystem.

For example, when starting graylog-server like:

```
/usr/local/openjdk-8/bin/java -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:NewRatio=1 -XX:MaxMetaspaceSize=256m -server -XX:+ResizeTLAB -XX:+UseConcMarkSweepGC -XX:+CMSConcurrentMTEnabled -XX:+CMSClassUnloadingEnabled -XX:+UseParNewGC -XX:-OmitStackTraceInFastThrow -cp /usr/share/graylog:/usr/share/graylog/graylog.jar -Dlog4j.configurationFile=/usr/share/graylog/data/config/log4j2.xml -Djava.library.path=/usr/share/graylog/lib/sigar/ -Dgraylog2.installation_source=docker org.graylog2.bootstrap.Main server -f /usr/share/graylog/data/config/graylog.conf
```
The use of `-cp /usr/share/graylog:/usr/share/graylog/graylog.jar` triggers the vulnerability.

The check for `..` elements in the resulting file path for documentation resources is performed too early in the resolution process. As a result, an attacker can craft a URL that can contain `..` elements and thus read files the server process has access to.

# Attribution
Many thanks to Florian Hauser and Christian Fünfhaus of Code White for disclosing this vulnerability.

dennisoelkers:
## Motivation and Context

Prior to this change, the `/api-browser/.*` resource allowed to retrieve static assets used by swagger. Some of these were prefixed with a `global/` prefix, that needs to be stripped away before locating the file. The string replacement used for this was used unbounded (replacing every occurence of the prefix, not just the first one), so it could be used to create URLs containing `..`. The check if the route contains `..` was performed _before_ the string replacement, so it could effectively be used to evade it.

This change is:

1. Replacing _only the first_ occurrence of `global/`
2. Performing the check if the route contains `..` _after_ the replacement and _right before_ the file is tried to be opened.

Fixes #8986.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/documentation/DocumentationBrowserResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/documentation/DocumentationBrowserResource.java
@@ -98,14 +98,14 @@ public class DocumentationBrowserResource extends RestResource {
     @GET
     @Path("/{route: .*}")
     public Response asset(@PathParam("route") String route) throws IOException {
-        // Directory traversal should not be possible but just to make sure..
-        if (route.contains("..")) {
-            throw new BadRequestException("Not allowed to access parent directory");
-        }
-
         // Remove path globalModePath before we serve swagger resources
         if (route.startsWith("global/")) {
-            route = route.replace("global/", "");
+            route = route.replaceFirst("global/", "");
+        }
+
+        // Trying to prevent directory traversal
+        if (route.contains("..")) {
+            throw new BadRequestException("Not allowed to access parent directory");
         }
         final URL resource = classLoader.getResource("swagger/" + route);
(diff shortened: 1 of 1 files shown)


bernd (approved):
I tested this in development and production setups including the non-standard server startup commands.

PR #8981, opened 2020-09-16
Reset Consumes annotation to accept everything for (un)schedule

bernd:
Since these endpoints don't accept a request body, we can just accept
any content-type. This unbreaks calling these endpoints with our swagger
API browser.

(This should be backported to the 3.3 branch)


linuspahl (approved):
Looks good to me, it fixes the problem with swagger. Enabling and disabling an event definition in the UI still works.

I will create a backport for 3.3.

PR #8995, opened 2020-09-18
Avoid URL synchronization if view is not currently dirty.

dennisoelkers:
## Motivation and Context

Before this change, the URL synchronization hook unconditionally added query/timerange/streams information to the current URL if the current view is a search (and not a dashboard). This lead to an unnecessary URL change if a saved search is loaded (as the persisted query/timerange/streams are identical to the current state).

This change is checking if the current view is dirty. If not, no change is made.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
+++ b/graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
@@ -31,9 +31,9 @@ const extractTimerangeParams = (timerange: TimeRange) => {
 };
 
 export const syncWithQueryParameters = (query: string, action: (string) => mixed = history.push) => {
-  const { view } = ViewStore.getInitialState() || {};
+  const { view, dirty } = ViewStore.getInitialState() || {};
 
-  if (view && view.type === View.Type.Search) {
+  if (view && view.type === View.Type.Search && dirty) {
     const { queries } = view.search;
 
     if (queries.size !== 1) {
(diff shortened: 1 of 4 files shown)


linuspahl (changes requested):
I found one problem which occurs when using the browser navigation:
search-example

If a user changes his search multiple times, he can still go back to a previous state by using the bowser navigation, but he will never get back to the initial state. (If he did not executed the first search manually). It behaves the same way for saved searches and streams.

graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
```diff
@@ -31,9 +31,9 @@ const extractTimerangeParams = (timerange: TimeRange) => {
 };
 
 export const syncWithQueryParameters = (query: string, action: (string) => mixed = history.push) => {
-  const { view } = ViewStore.getInitialState() || {};
+  const { view, dirty } = ViewStore.getInitialState() || {};
```
kyleknighted, line 34: I do not think there is a way for `ViewStore.getInitialState()` to return falsy, so we may not need the fallback `|| {}`

```suggestion
  const { view, dirty } = ViewStore.getInitialState();
```

graylog2-web-interface/src/views/hooks/SyncWithQueryParameters.js
```diff
@@ -31,9 +31,9 @@ const extractTimerangeParams = (timerange: TimeRange) => {
...
 
-  if (view && view.type === View.Type.Search) {
+  if (view && view.type === View.Type.Search && dirty) {
```
kyleknighted, line 36: As noted in a DM, optional chaining was not available when this was initially written, but could be simplified a little bit with 
```suggestion
  if (view?.type === View.Type.Search && dirty) {
```

kyleknighted (approved):
Couple notes, but LGTM

dennisoelkers:
Closing this, as it is not a proper solution, as [pointed out](https://github.com/Graylog2/graylog2-server/pull/8995#pullrequestreview-493480066) by @linuspahl

PR #9006, opened 2020-09-21
Make RolesSelector reusable by removing the connection to users

linuspahl:
This PR makes the `RolesSelector` reusable by removing the connection to users. This way it can not only be used to assign roles for users but also for teams.

Have a look at the enterprise PR for more information.
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1772


kmerz (approved):
LGTM :+1: 

PR #9007, opened 2020-09-21
Add authentication services infrastructure

bernd:
This PR adds the authentication services infrastructure including an authentication service for MongoDB and LDAP authentication.

Includes the changes from #8949 so we can close that one.

Closes #8949


graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceAuthenticator.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    public AuthServiceResult authenticate(AuthServiceCredentials authCredentials) {
+        final Optional<AuthServiceBackend> activeBackend = authServiceConfig.getActiveBackend();
+
+        if (activeBackend.isPresent()) {
+            final AuthServiceResult result = authenticate(authCredentials, activeBackend.get());
+            if (result.isSuccess()) {
+                return result;
+            }
+            // TODO: Do we want the fallback to the default backend here? Maybe it should be configurable?
```
kmerz, line 52: My thoughts to this:

- LDAP server logs to graylog
- LDAP server is "broken"
- You want to look at the problem but you can't log in to graylog since LDAP is "broken"
- A fallback user with reader rights (so other than admin) would be helpful

bernd: @kmerz One fallback user that is always there (at least at the moment) is the root user account. People can also still add users to MongoDB as fallback admin accounts if the external provider breaks.

This comment was more about if we always want the fallback to the MongoDB backend or are there reasons to not do that. To make you use case work, we need the fallback. (and it's also that way at the moment)

kmerz (approved):
Code LGTM :+1:

PR #9015, opened 2020-09-22
Quote search query values for regexp operators

bernd:
Avoids errors when the search query contains regexp meta characters like
`\`.


kmerz (approved):
Code LGTM :+1: 
Works as expected :heavy_check_mark:

PR #9028, opened 2020-09-24
Fix empty search result in stream

kmerz:
## Motivation
Prior to this change, when the user was executing a search on streams
which resulted in no streams. The search bar was removed and the user
could not reset the search.

## Description
This change will display the Alert which will show that there are no
streams found below the search bar in stead of replacing it.


dennisoelkers (approved):
Good one!

PR #9030, opened 2020-09-24
Abstracting link components by reexporting them.

dennisoelkers:
## Motivation and Context

This PR is creating an abstraction around our `react-router`-specific
link components (`Link` & `LinkContainer`) by reexporting them from an
own module. This allows us to replace them easily across core & plugins
when we are replacing `react-router` or `react-router-bootstrap` in a
future step.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/events/events/Events.jsx
+++ b/graylog2-web-interface/src/components/events/events/Events.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import lodash from 'lodash';
 import { Link } from 'react-router';
-import { LinkContainer } from 'react-router-bootstrap';
+import { LinkContainer } from 'components/graylog/router';
 import styled, { css } from 'styled-components';
 
 import { Alert, Col, Label, OverlayTrigger, Row, Table, Tooltip, Button } from 'components/graylog';
--- a/graylog2-web-interface/src/components/events/events/Events.jsx
+++ b/graylog2-web-interface/src/components/events/events/Events.jsx
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import lodash from 'lodash';
-import { Link } from 'components/graylog/router';
-import { LinkContainer } from 'components/graylog/router';
 import styled, { css } from 'styled-components';
 
+import { Link, LinkContainer } from 'components/graylog/router';
 import { Alert, Col, Label, OverlayTrigger, Row, Table, Tooltip, Button } from 'components/graylog';
 import { EmptyEntity, IfPermitted, PaginatedList, Timestamp, Icon } from 'components/common';
 import Routes from 'routing/Routes';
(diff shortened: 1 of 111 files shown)


graylog2-web-interface/src/components/events/events/Events.jsx
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import lodash from 'lodash';
-import { Link } from 'react-router';
-import { LinkContainer } from 'react-router-bootstrap';
 import styled, { css } from 'styled-components';
 
+import { Link, LinkContainer } from 'components/graylog/router';
```
Gaya, line 6: Right on! Saved a line of code 🤓

graylog2-web-interface/src/views/components/views/MissingRequirements.jsx
```diff
@@ -1,9 +1,9 @@
 // @flow strict
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-// $FlowFixMe: imports from core need to be fixed in flow
-import { LinkContainer } from 'react-router-bootstrap';
 
+// $FlowFixMe: imports from core need to be fixed in flow
```
Gaya, line 5: Are these FlowFixMe still needed?

Gaya (approved):
This approach looks safe and good to me. I'd do the same thing.

Is there a way to ignore the weblinter on jenkins since you're touching this many old files?

graylog2-web-interface/src/views/components/views/MissingRequirements.jsx
dennisoelkers: No, good catch!

dennisoelkers:
@Gaya: We _could_ introduce a flag for Jenkins to skip the linter job for jobs like this, or backports where we want to keep the amount of changes minimal. But I am not sure if this would be good, because this way the reviewer has to look and think twice if the linter job is red. You can still merge the PR if you have the same opinion for the linter job to be ok to fail. Does that make sense?

Gaya:
> @Gaya: We _could_ introduce a flag for Jenkins to skip the linter job for jobs like this, or backports where we want to keep the amount of changes minimal. But I am not sure if this would be good, because this way the reviewer has to look and think twice if the linter job is red. You can still merge the PR if you have the same opinion for the linter job to be ok to fail. Does that make sense?

Can we already merge even though the linter is red? I have mixed feelings about that, but it would be the same as putting in a flag to disable linting.

But yeah, I am okay with being able to merge even though the linter says no, maybe it's up to the person creating the PR to put in a comment that it's only touching minor parts of old files and fixing the linting errors makes no sense?

dennisoelkers:
> Can we already merge even though the linter is red?

Yes, that should work.

> [...] maybe it's up to the person creating the PR to put in a comment that it's only touching minor parts of old files and fixing the linting errors makes no sense?

I would think that the author should add a comment if it is not obvious. There might be cases where the author thinks it is obvious, while the reviewer does not, but that would just mean that the reviewer might ask if the linter is red on purpose or not.

Gaya:
100% agree with this methodology.

PR #9034, opened 2020-09-24
Move Teams share url to enterprise

kmerz:
## Motivation
Prior to this change, the teams share url was accessed in the core
but we want to strictly avoid having enterprise code leaked into
core.

## Description
This change implements the teams share url and action in the
TeamsStore/Action/Domain

Also Fix type on type declaration

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1785


linuspahl (approved):
Looks good to me, thanks for the update.

PR #9041, opened 2020-09-25
Remove `add-module-exports` babel plugin.

dennisoelkers:
## Motivation and Context

This plugin was used before mainly for interop between ancient Typescript parts and core. Recently, it was only needed for tests doing `require()` calls without a `default` suffix (we should get rid of `require` in tests anyway).

This change is removing the plugin and changes test calls.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/babel.config.js
+++ b/graylog2-web-interface/babel.config.js
@@ -3,13 +3,16 @@ module.exports = {
   plugins: [
     '@babel/plugin-syntax-dynamic-import',
     '@babel/plugin-proposal-class-properties',
-    'add-module-exports',
     'babel-plugin-styled-components',
   ],
   env: {
     test: {
       presets: ['@babel/env'],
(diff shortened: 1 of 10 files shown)


Gaya:
@dennisoelkers is it possible that the plugins are failing on Jennkins because of this change? Other than that I can only encourage changes like this.

dennisoelkers:
@Gaya: Yes, this PR requires the referenced plugin PR. If you want to test it, you should check out both simultaneously. The plugin PR is referencing this PR for jenkins, so its build is green.

Gaya:
> @Gaya: Yes, this PR requires the referenced plugin PR. If you want to test it, you should check out both simultaneously. The plugin PR is referencing this PR for jenkins, so its build is green.

Hmmm, it's showing a Maven build error here. Or is this not the build we should be looking at with referenced builds?

Gaya (approved):
Checked and tested. Looks good!

PR #9042, opened 2020-09-25
Fix authentication service configuration updates

bernd:
Previously we didn't actually save the updated configuration in the API resource.

Also add `AuthServiceBackend#prepareConfigUpdate()` to allow backends to modify the configuration before it's being updated in the database. (e.g. to check what needs to happen with the system user password in LDAP)


linuspahl (approved):
Works as expected!

PR #9048, opened 2020-09-28
Unify wizard navigation style

linuspahl:
This PR defines the `EventDefinitionFrom` wizard navigation style as the default for all horizontal wizards:

image

It looks better than the default style, which is being used in two plugins. The default style looks like this:
image

This PR also improves the responsiveness of the navigation.
Before:
image

After:
image

There are some ESLint warnings, but they are not related to this PR.


kmerz (approved):
Code LGTM :+1: 
Works on my machine!

PR #9049, opened 2020-09-28
Install formik as a vendor module

linuspahl:
Installing `formik` as a vendor module is necessary to combine form components from core and e.g. the enterprise plugin.

If we would install `formik` in both projects we could not create a `<Form>` in the enterprise plugin and e.g. use `FormikInput` from `components/common`. The reason is that any component in core that uses the formik context would get an empty context, as the context from core is different from the one in enterprise. Webpack bundles both context files into separate modules with different ids, we need to make sure that any usage of a context shared between core and plugins uses the exact same module in both the webpack build and tests.

This change is not necessary for current graylog development version, but will be in the future.


dennisoelkers (approved):

dennisoelkers:
@linuspahl: Looks good! After you have resolved the conflict, I will merge it.

PR #9059, opened 2020-09-29
Add error prop for `Input` to be able to display an error message together with the input help text

linuspahl:
## Motivation and Context
The `Input` component currently allows defining a help text. We are also using this prop to display errors for an input. This way we always hide the help text when want to display an error. There are many cases where the help text is especially helpful when there is an error.

With this PR we are adding an extra error prop to be able to display both messages at the same time.

This change does not affect the current usage. I already adjusted the `ContentPackSelect` to demonstrate how to apply the change for existing components.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1790

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/InputDescription.jsx
+++ b/graylog2-web-interface/src/components/common/InputDescription.jsx
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 
 import { HelpBlock } from 'components/graylog';
 
-const ErrorMessage = styled.span(({ theme, hasError }) => hasError && `
+const ErrorMessage = styled.span(({ theme }) => `
   color: ${theme.colors.variant.danger};
 `);
 
@@ -29,14 +29,17 @@ const InputDescription = ({ help, error }: Props) => {
 
   return (
     <HelpBlock>
-      <ErrorMessage hasError={!!error}>
-        {error}
-      </ErrorMessage>
+      {error && (
+        <ErrorMessage>
+          {error}
+        </ErrorMessage>
+      )}
       {(!!error && !!help) && <br />}
-      <HelpMessage>
-        {help}
-
-      </HelpMessage>
+      {help && (
+        <HelpMessage>
+          {help}
+        </HelpMessage>
+      )}
     </HelpBlock>
   );
 };
--- a/graylog2-web-interface/src/components/common/InputDescription.jsx
+++ b/graylog2-web-interface/src/components/common/InputDescription.jsx
@@ -29,7 +29,8 @@ const InputDescription = ({ help, error }: Props) => {
   return (
     <Wrapper hasError={!!error}>
       <HelpBlock>
-        {error && <>{error}<br /></>}
+        {error}
+        {(error && help) && <br />}
         {help}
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/common/InputDescription.jsx
```diff
@@ -0,0 +1,56 @@
...
+import styled, { type StyledComponent } from 'styled-components';
+
+import type { ThemeInterface } from 'theme';
+import { HelpBlock } from 'components/graylog';
+
+const Wrapper: StyledComponent<{hasError: boolean}, ThemeInterface, HTMLSpanElement> = styled.span(({ theme, hasError }) => `
+  ${hasError ? `
+    color: ${theme.colors.variant.danger};
+  ` : ''};
+`);
```
linuspahl, line 13: I made sure the text color changes when there is an error, this way the component can be used independently of the  `Input` component. I tried to style the `HelpBlock`, but it is not possible because the `hasError` attribute would be added to the DOM.

graylog2-web-interface/src/components/common/InputDescription.jsx
```diff
@@ -0,0 +1,56 @@
+// @flow strict
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import styled, { type StyledComponent } from 'styled-components';
+
+import type { ThemeInterface } from 'theme';
+import { HelpBlock } from 'components/graylog';
+
+const Wrapper: StyledComponent<{hasError: boolean}, ThemeInterface, HTMLSpanElement> = styled.span(({ theme, hasError }) => `
```
Gaya, line 9: I find that putting a conditional with a template string and using a template string again can be confusing. Personally I think this reads better:

```suggestion
const Wrapper: StyledComponent<{hasError: boolean}, ThemeInterface, HTMLSpanElement> = styled.span(({ theme, hasError }) => hasError ? `color: ${theme.colors.variant.danger};` : '';
```

Gaya (approved):
Looks alright to me, just a few questions.

Did we change the colour of the help text when the help text was actually an error before?

I like the idea of help and error both being displayed, but when we get an error, the help is also displayed with the error colour in this implementation. This might be confusing.

graylog2-web-interface/src/components/common/InputDescription.jsx
linuspahl: Thanks, thats a good hint!

linuspahl:
@Gaya Thanks for having a look, regarding the help text and error being displayed in the same color:
I thought about this as well and I personally prefer to only display the error text in red. I've adjusted it as suggested, if we see any problem in the future we can still change this behaviour.

The main reason I did not adjusted this already, is because it's the bootstrap way. No matter what `bsStyle` (e.g. `success`) you define for a form group it will also apply the color (for this example `green` ) for the help text. 
One the one hand its easier to identify the state of an input, on the other hand it can be confusing.

Gaya:
@linuspahl Jenkins is failing still 😢

kmerz (approved):

PR #9067, opened 2020-10-02
Add ownership check to saved search/dashboard permission checks.

### LINKED ISSUE #9068: Saved searches/dashboards permission checks do not respect ownership.
The permission checks used to check if a user has access to a saved search/dashboard do consider the user's permissions (associated with its user object) or any view sharings that might exist from other users. In addition to that, every saved search/dashboard has an optional `owner` field, which is not respected. So a user might be the owner of a saved search/dashboard and has initially created it, but lacks permissions/view sharings and is not allowed to see/open it.

[z#4115]

dennisoelkers:
## Motivation and Context

Before this change, the following checks were performed when a saved search/dashboard is listed/opened:

  - Does the user have permissions for the saved search/dashboard?
  - Is the saved search/dashboard shared with the user?

If one of them holds true, access is permitted. Although the object has an optional `owner` field, its presence and equality with the current user's name is not checked.

This did not cause any issues in general, because the permissions are added dynamically to the current user when a saved search/dashboard is created, so their was no reliance on the existence of an ownership check.

During the migration of legacy saved searches, their owner field was migrated too, but no permissions were added to this user. Therefore, migrated saved searches were visible only to the admin user, but not to the user creating them initially.

This change is now adding a check additionally to the previously existing ones, that checks for the presence of the `owner` field of a saved search/dashboard and compares it to the current user's name, permitting access if they are equal. This allows the users initially creating legacy saved searches seeing and accessing their migrated counterparts.

Fixes #9068.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/DashboardsResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/DashboardsResource.java
@@ -24,16 +24,12 @@ import io.swagger.annotations.ApiParam;
 import org.apache.shiro.authz.annotation.RequiresAuthentication;
 import org.graylog.plugins.views.search.views.ViewDTO;
 import org.graylog.plugins.views.search.views.ViewService;
-import org.graylog.plugins.views.search.views.sharing.IsViewSharedForUser;
-import org.graylog.plugins.views.search.views.sharing.ViewSharing;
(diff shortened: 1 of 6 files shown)


kmerz (approved):
Code LGTM :+1: 
- Created a view with a user
- removed the permissions from the user
- view still loads

PR #9070, opened 2020-10-02
Migrate user permissions into grants

### LINKED ISSUE #8919: Migrate user permissions into grants
The currently existing permissions that are directly stored within the user collection
should also be migrated to grants (where applicable)

The conversion should be done for Dashboards and Streams:
View + Edit  -> Manage 
View             -> Viewer

mpfz0r:
Fixes #8919


kmerz (approved):
Code LGTM :+1: 

wrong browser tab

kmerz (changes requested):
Did not work on my machine!

kmerz (approved):
Code LGTM :+1: 
Tested with different configurations and it worked :heavy_check_mark:

graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
```diff
@@ -136,6 +141,27 @@ public GrantDTO create(GRN grantee, Capability capability, GRN target, String cr
...
+        final List<GrantDTO> existingGrants = getForTargetAndGrantee(target, grantee);
+        if (existingGrants.isEmpty()) {
+            return create(grantee, capability, target, creatorUsername);
+        }
+        // This should never happen
+        Preconditions.checkState(existingGrants.size() == 1);
+
+        final GrantDTO grantDTO = existingGrants.get(0);
+        // Only upgrade capabilities: VIEW < MANAGE < OWNER
+        if (capability.ordinal() > grantDTO.capability().ordinal()) {
```
bernd, line 158: I am not sure if using the #ordinal method of the enum will bite us at one point in the future. The value will change if the enums are reordered at some point. (e.g. someone wants them to be sorted alphabetically and refactors the order) How about adding an explicit priority value to each enum instead?

bernd (changes requested):
Only one comment:

graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
mpfz0r: fair enough. :+1:

bernd (approved):

PR #9082, opened 2020-10-05
Check password length of user creation before submit

### LINKED ISSUE #8913: User Creation password lenght check is only triggered after submit
We should check if the password length is corrected on unfocus the password input. Otherwise it is a bit annoying when you typed the password 2 times only to find that it is not long enough.

kmerz:
## Motivation
Prior to this change, the length of th password was checked when the
users submited the user creation. Which means he typed the password
already 2 times.

## Description
This change will check the password length from the beginning.

Fixes #8913 


linuspahl (approved):
Looks good to me!

PR #9085, opened 2020-10-05
Stop pivot config form submit event propagation.

### LINKED ISSUE #9087: pivot number of aggregation values can't be changed
## Expected Behavior
changing the number of items in the aggregation should not blank the results and it needs to be possible to save the number of aggregation numbers. 

## Current Behavior
3.3.6 introduced a bug 

1. create top n values 
2. edit the widget (select items 25)

* Graylog Version: 3.3.6

dennisoelkers:
## Motivation and Context

Before this change, when configuring the settings for a row/column pivot in the widget edit mode, submitting the form resulted in a double submit. First, the pivot config form is submitted and its config is changed. Then, the same submit event is propagated to the overlying widget edit form (as defined in `SearchBarForm`). The `onSubmit` handler in there is submitting the current widget again, but with stale pivot config settings, resulting in a mismatching view/search constellation. This results then in search results not being resolvable, as the widget mapping is based on a stale search.

This change is now preventing the pivot config form submit event to be propagated, therefore limiting its effect to the initial submit only.

Fixes #9087.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/PivotConfiguration.jsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/PivotConfiguration.jsx
@@ -33,6 +33,7 @@ export default class PivotConfiguration extends React.Component {
 
   _onSubmit = (e) => {
     e.preventDefault();
+    e.stopPropagation();
     const { onClose } = this.props;
     onClose(this.state);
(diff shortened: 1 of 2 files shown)


linuspahl (approved):

PR #9088, opened 2020-10-05
Updating to @testing-library/react v11.0.4.

dennisoelkers:
This change is updating `@testing-library/react` to the next major
version. All tests passed before and after.


linuspahl (approved):
Looks good to me.

PR #9104, opened 2020-10-07
Sync cloud branch with current 3.3. branch

edmundoa:
Merges 3.3 into cloud-3.3. This time with the right base branch :)


edmundoa:
The linter errors are also in 3.3, so I think we should not be fixing them here.

Gaya (approved):
LGTM

PR #9105, opened 2020-10-07
Hide CSV Adapter in cloud instances

edmundoa:
Using CSV adapters in cloud will have different requirements, so this
plugin should not be available.

Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/437


thll (approved):
LGTM

Issue #8792, opened 2020-08-14
Enable multi-select for TeamsSelector, RolesSelector and UsersSelector


linuspahl:
Currently the `TeamsSelector`, `RolesSelector` and `UsersSelector` only support the assignment of one item, but it would improve the usability to be able to assign multiple items. 

Here is an overview of the necessary changes:
- User edit page
  - `RolesSelector`: Needs to be changed in the frontend
  - `TeamsSelector`: Needs to be changed in the frontend and backend https://github.com/Graylog2/graylog2-server/issues/9029
- Team edit page
  - `UsersSelector`: Uses same endpoints like `TeamsSelector`
  - `RolesSelector`: Needs to be changed in the frontend
- Role edit page
  - `UsersSelector` Needs to be changed in the frontend and backend https://github.com/Graylog2/graylog2-server/issues/9029

kmerz:
Fixed with #9107 and others

PR #9107, opened 2020-10-07
Bulk update roles

kmerz:
## Motivation
A user might want to update more then one role on a user or a team with one request. To make this possible the User Edit and User Create should make role bulk update possible

## Description
The backend code for user was already accepting the roles for a user in a bulk only the frontend code needed to be adjusted.
For the role there was need for backend adjustments so that the backend accepts multiple users and iterate for the users and adds the roles for that users. 

Refs #8792

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1805


linuspahl (approved):
Looks good to me!

Issue #8906, opened 2020-09-02
Add explanation for disabled share buttons


bernd:
There is currently no explanation why a share button is disabled:

![Peek 2020-09-02 11-38](https://user-images.githubusercontent.com/461/91966229-d90dab00-ed11-11ea-8732-282264d51a9a.gif)

There can be several reasons for a share button to be disabled. (unsaved search, search/dashboard not owned by the user)

One approach could be to update the `title` attribute to explain why the button is disabled. @linuspahl had another suggestion:

> In general: We talked about displaying an info icon next to the text which explains in a popover that you need to be owner to e.g. share a dashboard. That is one possible solution. I prefer it over a read-only icon.

PR #9113, opened 2020-10-08
Add explanation for disabled share buttons

linuspahl:
Right now users do not know why they can't share entities like dashboards. With this PR we are adding an info icon next to every disabled share action. Hovering over the help icon will open a popover which explains that only owners of the entity can share it.

FIxes: https://github.com/Graylog2/graylog2-server/issues/8906


graylog2-web-interface/src/components/common/HoverForHelp.jsx
```diff
@@ -0,0 +1,39 @@
+// @flow strict
```
linuspahl, line 1: I moved this file with `git mv`, not sure why it is being tracked as a new file.

kmerz (approved):
Code LGTM :+1: 
Worked fabulously :dancing_women: on my machine!

PR #9135, opened 2020-10-12
Upgrade `react-router` to v5.

dennisoelkers:
## Motivation and Context

This PR is upgrading `react-router` from v3 to v5, as well was `react-router-bootstrap`. It reuses the previously introduced abstraction (`withParams` / `withLocation`) to centralise changes. The biggest change is the restructuring of the `AppRouter` component. The most notable changes for route configuration are:

  - A `Switch` component is required when only a single route should match (instead of rendering all components where the route matches)
  - Instead of specifying a parent component in a parent route, the parent component itself now needs to wrap the `Route` component
  - The `location` prop does not include the `query` object anymore. If hooks are used, the `useQuery` hook should be used instead of `useLocation` to retrieve it, but it should be noted that `useQuery` provides a string and does not do any parsing, unlike before. `withLocation` is merging a parsed `query` object into `location` to provide backwards-compatible behavior.
  - The `WrappingContainer` component used to wrap all components in tests needs to include a `Router` component as well, in order to establish the `react-router` context

Configuration of plugin routes should stay the same as before. Also specifying a parent component (and allowing `null` for it) should be the same for plugin routes.

Additional changes include:

  - Adding `create-react-class` as a dependency to `graylog-web-plugin`. Before, it was transitively included by `react-router`, now it needs to be specified explicitly
  - Typing `react-router` properly now

## Types of changes

## Checklist:


linuspahl (approved):
Look good to me!

PR #9165, opened 2020-10-15
Expose `isCloud` to frontend

thll:
## Description

Exposes an `isCloud` flag via `config.js`.

When running in production mode, the flag is mirroring the `is_cloud` server configuration parameter which is by default `false` and has to be enabled explicitly.

When running in development mode (`env DEVELOPMENT=true`), the flag is `false` by default but will be automatically set to `true` when the cloud plugin is present.

## Motivation and Context

When running in a cloud context, certain UI elements need to be different than for an on-premise deployment. With this flag exposed, the UI has an easy way to know about the context.

## How Has This Been Tested?

To see if this works in development mode I altered the `DEV` badge we show in the header. If the cloud context is detected, it now looks like this:
image

Checked that this works also in the production build by temporarily adding `console.log` debugging statements and building with maven.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/navigation/Navigation.jsx
+++ b/graylog2-web-interface/src/components/navigation/Navigation.jsx
@@ -90,7 +90,7 @@ const Navigation = ({ permissions, fullName, location, loginName }) => {
 
         {
         AppConfig.gl2DevMode()
-          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">DEV</Badge>
+          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''} DEV</Badge>
         }
       </Navbar.Header>
 
--- a/graylog2-web-interface/src/components/navigation/Navigation.jsx
+++ b/graylog2-web-interface/src/components/navigation/Navigation.jsx
@@ -90,7 +90,7 @@ const Navigation = ({ permissions, fullName, location, loginName }) => {
 
         {
         AppConfig.gl2DevMode()
-          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''} DEV</Badge>
+          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">{AppConfig.isCloud() ? 'CLOUD' : ''} DEV</Badge>
         }
       </Navbar.Header>
 
@@ -124,7 +124,7 @@ const Navigation = ({ permissions, fullName, location, loginName }) => {
           AppConfig.gl2DevMode()
             && (
               <InactiveNavItem className="dev-badge-wrap">
-                <Badge bsStyle="danger" className="dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''} DEV</Badge>
+                <Badge bsStyle="danger" className="dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''}DEV</Badge>
               </InactiveNavItem>
             )
           }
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/components/navigation/Navigation.jsx
```diff
@@ -90,7 +90,7 @@ const Navigation = ({ permissions, fullName, location, loginName }) => {
 
         {
         AppConfig.gl2DevMode()
-          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">DEV</Badge>
+          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''} DEV</Badge>
```
Gaya, line 93: I'd remove the prefix space here

There is already a space before `DEV` so I think we can leave it out of the `'CLOUD '` string, or remove the space before `DEV`
```suggestion
          && <Badge bsStyle="danger" className="small-scrn-badge dev-badge">{AppConfig.isCloud() ? 'CLOUD' : ''} DEV</Badge>
```

graylog2-web-interface/src/components/navigation/Navigation.jsx
```diff
@@ -124,7 +124,7 @@ const Navigation = ({ permissions, fullName, location, loginName }) => {
           AppConfig.gl2DevMode()
             && (
               <InactiveNavItem className="dev-badge-wrap">
-                <Badge bsStyle="danger" className="dev-badge">DEV</Badge>
+                <Badge bsStyle="danger" className="dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''} DEV</Badge>
```
Gaya, line 127:
```suggestion
                <Badge bsStyle="danger" className="dev-badge">{AppConfig.isCloud() ? 'CLOUD ' : ''}DEV</Badge>
```

graylog2-web-interface/src/util/AppConfig.js
```diff
@@ -30,6 +30,15 @@ const AppConfig = {
     return this.appConfig().rootTimeZone;
   },
 
+  isCloud() {
+    if (typeof (IS_CLOUD) !== 'undefined') {
```
Gaya, line 34: You can remove the parenthesis after `typeof`
```suggestion
    if (typeof IS_CLOUD !== 'undefined') {
```

graylog2-web-interface/webpack.combined.config.js
```diff
@@ -16,7 +16,11 @@ const globOptions = {
   nodir: true,
 };
 
-const pluginConfigs = process.env.disable_plugins === 'true' ? [] : glob.sync(pluginConfigPattern, globOptions).map(config => `${globCwd}/${config}`);
+const pluginConfigs = process.env.disable_plugins === 'true' ? [] : glob.sync(pluginConfigPattern, globOptions).map((config) => `${globCwd}/${config}`);
+
+if (pluginConfigs.filter((config) => config.includes('graylog-plugin-cloud/server-plugin')).length > 0) {
```
Gaya, line 21: `Array.some()` will return true if there is any element which matches the condition. So no need to filter and then count the amount of elements.

```suggestion
if (pluginConfigs.some((config) => config.includes('graylog-plugin-cloud/server-plugin'))) {
```

graylog2-web-interface/webpack.combined.config.js
```diff
@@ -29,10 +33,9 @@ function getPluginName(pluginConfig) {
     // If a package.json file exists (should normally be the case) use the package name for pluginName
     const pkg = JSON.parse(fs.readFileSync(packageConfig, 'utf8'));
     return pkg.name.replace(/\s+/g, '');
-  } else {
-    // Otherwise just use the directory name of the webpack config file
-    return path.basename(path.dirname(pluginConfig));
   }
+  // Otherwise just use the directory name of the webpack config file
```
Gaya, line 37: Much better indeed!

Gaya (changes requested):
Looks good and think this will suffice for what we want to achieve.

Just a couple of minor remarks on the implementation.

I'll use this for hiding elements in the application.

graylog2-web-interface/webpack.combined.config.js
thll: Thanks!

Gaya (approved):
Looks good!

PR #9170, opened 2020-10-15
Declare custom ES image as compatible w/ testcontainers.

dennisoelkers:
Starting with testcontainers v1.15.0-rc2, it expects specific images compatible with their Elasticsearch container support to be used with it. Therefore we need to declare the OSS image we are using for integration tests to be compatible.


mpfz0r (approved):
:+1:

PR #9171, opened 2020-10-15
Remove permission from whitelist check

mpfz0r:
Otherwise users which possess the new creator roles
cannot create objects that perform a URL whitelist check.
(e.g. creating a HTTPEventNotification)

Thus, we should allow any authenticated user to check if URLs are whitelisted.


thll (changes requested):
Makes sense and the changes look good!

@mpfz0r could you take care of the ESLint errors?

thll (approved):
LGTM

PR #9172, opened 2020-10-15
Correct typo on SystemAdminOverview page

kroepke:
correct typo `can only be edit` -> `can only be edited`


linuspahl (approved):
Thanks for the fix! I will merge the PR, once all checks passed, just to be sure.

PR #9179, opened 2020-10-15
Merge GrantsAuthorizingRealm and MongoDbAuthorizationRealm

bernd:
This fixes an issue when authorizing a user with a GRN principal. The
old MongodbAuthorizationRealm didn't handle GRN principals and the
GrantsAuthorizingRealm didn't load the user's permission and roles.


mpfz0r (approved):
Works and LGTM

PR #9185, opened 2020-10-16
Add GranteeService.getVisibleUsers()

mpfz0r:
Graylog users are visible to other users if the user that is running
the query has `users:read` permissions on certain (or all) users.

The `users:read` permission can be conveniently assigned by using the
new `User Inspector` role.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1833

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/testing/UserServiceExtension.java
+++ b/graylog2-server/src/test/java/org/graylog/testing/UserServiceExtension.java
@@ -1,0 +1,48 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog.testing;
+
+import org.graylog.testing.mongodb.MongoDBExtension;
+import org.graylog.testing.mongodb.MongoDBTestService;
+import org.graylog2.shared.users.UserService;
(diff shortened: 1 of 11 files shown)


graylog2-server/src/test/java/org/graylog/testing/UserServiceExtension.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+        return Objects.equals(UserService.class, parameterContext.getParameter().getType());
+    }
+
+    @Override
+    public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) throws ParameterResolutionException {
+        final Class<?> parameterType = parameterContext.getParameter().getType();
+
+        if (UserService.class.equals(parameterType)) {
```
bernd, line 41: Should we only inject `TestUserService` instances? That would make it clear that the user of the extension gets a `TestUserService` and not a fully featured `UserServiceImpl`. (we might want to rename the extension to `TestUserServiceExtension` in that case, too)

```suggestion
        if (TeamUserService.class.equals(parameterType)) {
```

bernd (changes requested):
LGTM :+1: Only one question about the `UserServiceExtension`

graylog2-server/src/test/java/org/graylog/testing/UserServiceExtension.java
mpfz0r: Makes sense :+1:

graylog2-web-interface/src/logic/permissions/GRN.js
```diff
@@ -25,7 +25,7 @@ export const getShowRouteFromGRN = (grn: string) => {
       return Routes.dashboard_show(id);
     case 'event_definition':
       return Routes.ALERTS.DEFINITIONS.edit(id);
-    case 'event_notification':
+    case 'notification':
```
bernd, line 28: Is this change intended?

bernd (changes requested):

graylog2-web-interface/src/logic/permissions/GRN.js
mpfz0r: oops, no. but after a merge it will be gone. It's in master already

bernd: Okay :+1:

bernd (approved):

PR #9188, opened 2020-10-16
Fix plugin routes with `null` parent components.

dennisoelkers:
## Motivation and Context

Our application router allows plugins to register routes with a `null` parent component, which allows them to be rendered outside of our application chrome. This got broke in #9135, as react-router v5 is now rendering all matching routes (unless `Switch` is used) and the scoping of the `Switch` components and the plugin routes got messed up.

This change is now restoring the original behavior by

  a) placing the `RouterErrorBoundary` outside of the `Switch` clause.
  b) putting the application chrome components (`App`/`AppWith...`) inside the catch-all route for non-null-parent components.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/routing/AppRouter.jsx
+++ b/graylog2-web-interface/src/routing/AppRouter.jsx
@@ -121,13 +121,13 @@ const AppRouter = () => {
 
   return (
     <Router history={history}>
-      <Switch>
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Looks good to me, works as expected.

PR #9204, opened 2020-10-19
Show friendlier page when user has no stream access.

### LINKED ISSUE #9174: Display helpful page when reader user is missing stream permissions
## Expected Behavior
The "you don't have any stream permissions" page should look less threatening and explain the situation better.

## Current Behavior
Currently a user with only the `Reader` role sees the following page directly after log in:
image

It looks like an error to the user and there's no indication of what they should do next or if they did something wrong.

1. Create a reader user
2. Log in as that user

## Your Environment

* Graylog Version: 4d38edcefee9092d50c8f781112b2af41b776cca

dennisoelkers:
## Motivation and Context

This small change is using a different layout for the page a user is
redirected to when a search is started, but the user does not have
access to any stream. Before this change, it was the error page layout,
which indicates to the user that this an error and not a condition.
After this change, a friendlier variant is used.

Fixes #9174.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/UserHasNoStreamAccess.jsx
+++ b/graylog2-web-interface/src/pages/UserHasNoStreamAccess.jsx
@@ -1,14 +1,28 @@
 // @flow strict
 import * as React from 'react';
 
-import ErrorPage from 'components/errors/ErrorPage';
+import { DocumentTitle, Icon } from 'components/common';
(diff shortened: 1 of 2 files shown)


kmerz (approved):
Love it :heart: 
Works on my machine! :heavy_check_mark:

PR #9216, opened 2020-10-20
Event definition share dialog

kmerz:
## Description
Add share button to Event Definitions and create a read only page for a user with view permissions.

## Motivation and Context
If a user wants to verify the settings for a event definition (as a revisor) then he needs a read only page.

If a user is missing stream or notification permissions a warning will be displayed but the page otherwise rendered


linuspahl (approved):
Could not find any errors!
An Idea I had while looking at these changes: In another iteration we can also display the Status, Last execution, Next execution, Next timerange on the details page.

mpfz0r (approved):
Works for me :)

Issue #9192, opened 2020-10-16
UI issues with usernames that contain "@" characters


bernd:
## Description

There are UI issues in several places when a username contains a `@` character. The user edit and user access token edit pages don't work, for example.

The issue is that the UI is encoding the `@` value in the username twice.

Example for the username `jane@example.com`:

The API receives the following username in the URL from the UI: `api/users/jane%2540example.com`

The Java API framework automatically decodes the value to `jane%40example.com`. But that user cannot be found because the `@` is still escaped.

In the `UsersStore.js` file we use `encodeURIComponent()` in serveral places to encode the `username` in the URL.
I guess the HTTP client we use is also encoding the URL again and converts `%40` to `%2540`.

### Possible Solutions

1. Ensure that usernames are only encoded once
1. Use the user ID instead of the username (requires backend changes)

## Steps to reproduce

1. Create a user with username `jane@example.com`
1. Try to load the user's edit page
1. Try to edit the user's API access tokens

## Your Environment

* Graylog Version: 4.0.0-SNAPSHOT

PR #9217, opened 2020-10-20
Use user ID instead of username for user management API calls

mpfz0r:
## Description
Previously most user management related API calls were based on the username, with this PR we are using the user id instead. This way we can use the user id for all user management related URLs like `/system/users/[user-id]/edit`.

Here is an overview of all endpoint changes
- user update `/users/${username}`  =>` /users/${userId}`
- user password change `/users/${username}/password`  => `/users/${userId}/password`
- user token create and delete `/users/${username}/tokens/${tokenName}`  => `/users/${userId}/tokens/${tokenName}`
- user tokens list `/users/${username}/tokens/${tokenName}` => `/users/${userId}/tokens/${tokenName}`

Only the user delete endpoint did not changed

Fixes: https://github.com/Graylog2/graylog2-server/issues/9238
Fixes: https://github.com/Graylog2/graylog2-server/issues/9192


linuspahl:
@mpfz0r Thanks for the change, it's good to keep the current endpoint which uses the username, because we still need it in one case.

It would be very helpful if we can adjust `/shares/user/${username}` => `/shares/user/${userId}` in addition.

mpfz0r:
> @mpfz0r Thanks for the change, it's good to keep the current endpoint which uses the username, because we still need it in one case.

Sure,  that's why I didn't delete it ;-)

> It would be very helpful if we can adjust `/shares/user/${username}` => `/shares/user/${userId}` in addition.

Doing this, would need a kludge, because those paths are in conflict with each other.
We could try to detect an ID from a username, but in theory, there could be still a username that looks like an ID ;-)

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -173,6 +174,32 @@ public UserSummary get(@ApiParam(name = "username", value = "The username to ret
...
+    @ApiResponses({
+            @ApiResponse(code = 404, message = "The user could not be found.")
+    })
+    public UserSummary getbyId(@ApiParam(name = "userId", value = "The userId to return information for.", required = true)
+                           @PathParam("userId") String userId,
+                           @Context UserContext userContext) {
+
+        final User user = loadUserById(userId);
+        final String username = user.getName();
+        // If a user has permissions to edit another user's profile, it should be able to see it.
```
bernd, line 190: How about putting the handling code into a separate method so we don't duplicate the same code in `get` and `getbyId`? :smile:

graylog2-web-interface/src/actions/users/UsersActions.js
```diff
@@ -40,13 +40,14 @@ export type PaginatedUsers = PaginatedList<UserOverview> & {
...
-  update: (username: string, request: any) => Promise<void>,
-  delete: (username: string) => Promise<void>,
-  changePassword: (username: string, request: ChangePasswordRequest) => Promise<void>,
-  createToken: (username: string, tokenName: string) => Promise<Token>,
-  loadTokens: (username: string) => Promise<Token[]>,
-  deleteToken: (username: string, tokenId: string, tokenName: string) => Promise<void>,
+  load: (userId: string) => Promise<User>,
+  loadByUsername: (username: string) => Promise<User>,
+  update: (userId: string, request: any) => Promise<void>,
+  delete: (username: string, fullName: string) => Promise<void>,
```
bernd, line 46: Why don't we delete users by ID and still use the username for that?

bernd (changes requested):

graylog2-web-interface/src/pages/UserTokensEditPage.jsx
```diff
@@ -31,16 +31,16 @@ const PageTitle = ({ fullName }: {fullName: ?string}) => (
   </>
 );
 
-const _loadTokens = (username, currentUser, setTokens) => {
-  if (isPermitted(currentUser?.permissions, [`users:tokenlist:${username}`])) {
-    UsersDomain.loadTokens(username).then(setTokens);
+const _loadTokens = (userId, currentUser, setTokens) => {
+  if (isPermitted(currentUser?.permissions, [`users:tokenlist:${userId}`])) {
```
kmerz, line 35: for permission we need to use the `username`

kmerz (changes requested):
One slight nitpick, did not yet test.

graylog2-web-interface/src/pages/UserTokensEditPage.jsx
linuspahl: Good catch

graylog2-web-interface/src/actions/users/UsersActions.js
mpfz0r: I didn't want to touch the delete code, because that also needs changes to the UserService.delete() which has these weird functionality that allows to remove multiple users with the same name..

bernd: @mpfz0r How about creating a `UserService#deleteById` method and a `DELETE /users/id/<id>` endpoint so we can leave the other code untouched?

mpfz0r: @bernd that I can do, of course

mpfz0r: ok, done. @linuspahl there is also user deletebyId endpoint now

kmerz (approved):
LGTM :+1:

bernd (approved):

PR #9223, opened 2020-10-21
Offer GIM as a template option for index set configs.

dennisoelkers:
## Goal

As a Graylog user, I would want to be able to configure an index set to use the Graylog Information Model mapping template, so every new index created in this index set automatically has the proper mappings.

## Description of Changes

This PR is adding a mapping template class similar to our standard mapping template, which includes the field types and namings as defined in the Graylog Information Model. A new possible value (`generic`) is added for the `index_template_type` field of the index set config, which can be used to enable the usage of this template type for a given index set.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
@@ -305,7 +305,7 @@ public abstract class GIMMapping extends IndexMapping {
     private static final Set<String> DATE_FIELDS = ImmutableSet.<String>builder()
             .add("event_created")
             .add("event_start")
-            .add("event_recieved_time")
+            .add("event_received_time")
             .add("file_created_date")
             .build();
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
@@ -401,7 +401,8 @@ public abstract class GIMMapping extends IndexMapping {
                                         "char_filter", Collections.emptyList(),
                                         "filter", Collections.singleton("lowercase")
                                 )
-                        )
+                        ),
+                        "analyzer", analyzerKeyword()
                 )
         );
     }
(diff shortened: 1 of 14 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
```diff
@@ -58,7 +58,9 @@
         @JsonProperty("messages")
         MESSAGES,
         @JsonProperty("events")
-        EVENTS
+        EVENTS,
+        @JsonProperty("generic")
+        GENERIC
```
bernd, line 63: What's the reasoning behind calling this `generic`? Wouldn't it be better to call this `gim` or `graylog-schema` or something similar?

And should we include a version number? Or how would we handle different versions of the schema?

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/indexer/indexset/IndexSetConfig.java
dennisoelkers: We had mentioned on the side that we would call the type "generic" in our last meeting about it, at least that is what I had written down. I cannot recall the reasoning though. So, I would rename it to `gim` - and include the versioning, so it would be `gim-v1` for now.

bernd (changes requested):
LGTM :+1: Only one small issue:

graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
```diff
@@ -0,0 +1,443 @@
...
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ * This file is part of Graylog.
```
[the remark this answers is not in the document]
dennisoelkers: :white_check_mark:

bernd (approved):

PR #9225, opened 2020-10-21
Implement PaginatedDataTable component

linuspahl:
## Description

This PR adds the `PaginatedDataTable` component to our `common` components directory. The `DataTable` component is only able to display the provided rows. When we have a list which is not getting paginated by the backend, we can't just combine the `DataTable` with the `PaginatedList`. This PR adds a component which handles this use case and ensures that the tables pagination works correctly in combination with the filter.

### DIFF
--- a/graylog2-web-interface/src/components/common/DataTable/DataTable.jsx
+++ b/graylog2-web-interface/src/components/common/DataTable/DataTable.jsx
@@ -197,17 +197,18 @@ class DataTable extends React.Component {
 
     return (
       <div>
-        <Filter customFilter={customFilter}
-                label={filterLabel}
-                id={id}
-                rows={rows}
-                displayKey={displayKey}
-                filterBy={filterBy}
-                filterSuggestions={filterSuggestions}
-                filterKeys={filterKeys}
-                onDataFiltered={this.filterDataRows}>
-          {children}
-        </Filter>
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/components/common/DataTable/DataTable.jsx
```diff
@@ -197,17 +197,18 @@ class DataTable extends React.Component {
...
+                  rows={rows}
+                  id={id}
+                  displayKey={displayKey}
+                  filterBy={filterBy}
+                  filterSuggestions={filterSuggestions}
+                  filterKeys={filterKeys}
+                  onDataFiltered={this.filterDataRows}>
+            {children}
+          </Filter>
+        )}
```
linuspahl, line 211: I tried to implement the following approach:
```
const TableFilter = customFilter ?? Filter

return (<div> 
  <TableFilter
      ...
```

This way the `customFilter` has access to all props the `Filter` component receives. This caused a problem where the filter input value got removed after submitting the filter. I tried to fix the problem with `useCallback` / `useMemo`, but due to the dependency chain, I would have to split up the pagination state of every component which uses the `DataTable` with a `customFilter`.

kmerz (approved):
LGTM :+1:

PR #9230, opened 2020-10-21
Don't sign any artifacts in the distribution module

bernd:
This fixes an error in release builds.

Also switch packaging to pom and skip source, javadoc, install and
deploy.


juju2112 (approved):
I approve.

PR #9233, opened 2020-10-22
Fix field types of current query if no stream is selected.

dennisoelkers:
## Motivation and Context

Before this change, the `FieldTypesStore` returns empty field types for the current query, if the current query has no streams selected (which results in all streams being queried).

This change is now fixing the `QueryFiltersStore` to return queries with an empty filter as well, and fixes the `FieldTypesStore` to return all field types for queries which have an empty filter.

## How Has This Been Tested?

Start a new search. Enter any field name in the query input, see if it states that it is "not in streams" in the auto-completion.

## Types of changes

## Checklist:


kmerz (approved):
LGTM :+1: 
Fixed on my machine :heavy_check_mark:

PR #9245, opened 2020-10-22
Fix typo

lennartkoopmann:
Fixing a typo in this very important and possibly first change since 2014 or so.


kyleknighted (approved):
Fix tests and LGTM 👏

linuspahl:
@kyleknighted You probably did not pushed the fix for the test? I just fixed the test, to ensure we do not merge a PR with a failing test.

Issue #9149, opened 2020-10-13
Share dialog fails if a grantee cannot be found


bernd:
I just ran into the following error while testing #9123. This happened when I opened the share dialog for a dashboard.

The problem is that user `5f85af1f11075861c403dbae` doesn't exist anymore and so it doesn't show up in `available_grantees`.

While this is a consistency issue in the backend, I think the frontend shouldn't fail that hard.

```
Cannot find grantee with id grn::::user:5f85af1f11075861c403dbae in available grantees

Stack Trace:

Error: Cannot find grantee with id grn::::user:5f85af1f11075861c403dbae in available grantees
    at http://127.0.0.1:39765/app.js:78355:17
    at http://127.0.0.1:39765/app.js:93053:46
    at http://127.0.0.1:39765/app.js:92904:56
    at http://127.0.0.1:39765/app.js:93053:36
    at http://127.0.0.1:39765/app.js:92869:35
    at http://127.0.0.1:39765/app.js:91414:16
    at ArrayMapNode.iterate.HashCollisionNode.iterate (http://127.0.0.1:39765/app.js:91763:11)
    at Map.__iterate (http://127.0.0.1:39765/app.js:91412:32)
    at ToKeyedSequence.__iterate (http://127.0.0.1:39765/app.js:92867:25)
    at KeyedIterable.mappedSequence.__iterateUncached (http://127.0.0.1:39765/app.js:93052:23)
    at seqIterate (http://127.0.0.1:39765/app.js:90641:16)
    at KeyedIterable.Seq.__iterate (http://127.0.0.1:39765/app.js:90311:14)
    at ToIndexedSequence.__iterate (http://127.0.0.1:39765/app.js:92904:25)
    at IndexedIterable.mappedSequence.__iterateUncached (http://127.0.0.1:39765/app.js:93052:23)
    at seqIterate (http://127.0.0.1:39765/app.js:90641:16)
    at IndexedIterable.IndexedSeq.__iterate (http://127.0.0.1:39765/app.js:90357:14)
    at IndexedIterable.toArray (http://127.0.0.1:39765/app.js:94295:23)
    at List (http://127.0.0.1:39765/app.js:92102:62)
    at IndexedIterable.toList (http://127.0.0.1:39765/app.js:94363:14)
    at EntityShareState.get (http://127.0.0.1:39765/app.js:78359:10)
    at EntityShareSettings (http://127.0.0.1:39765/app.js:134651:48)
[...]

PR #9256, opened 2020-10-23
Don't expose shares with invisible grantees

mpfz0r:
Filter the activeShares result so it contains only grantees
that are visible to the requesting user.


mpfz0r:
Refs #9149

bernd (approved):

PR #9257, opened 2020-10-23
Saved Search: Hide Save button for missing edit perissions

### LINKED ISSUE #8894: "⭐ Save"  button is active for shared (read-only) saved searches
## Expected Behavior

Accessing a saved search that I'm not a owner of, should indicate that it is read-only.

## Current Behavior

Clicking on the search button and trying to save the search, gives a permissions error.

## Possible Solution

I didn't dig too deep, but it seems that the `view.owner` is always set to the current user.
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchControls.jsx#L47

kmerz:
## Motivation
Prior to this change, the Save button was shown to a user which only had
view permissions to a Saved Search. If he clicked on that button the
result was an error page.

## Description
This change will hide the 'Save' and 'Save as' button and shows a
'Create New' button to make clear he has to create a new Search which he
can edit.

Fixes #8894


linuspahl (approved):
Looks good to me

PR #9269, opened 2020-10-26
Fix the shareable link copy button

### LINKED ISSUE #9254: Copy sharable URL into clipboard does not actually copy anything
When trying to use the "Copy" button here, the value does not end up in my clipboard. Also, the URL looks wrong because it points to my base URL but not a stream?

2020-10-22-143028_581x108_scrot

## Your Environment

* Graylog Version: 4.0.0-beta.2

kmerz:
## Motivation
Prior to this change, clicking on the copy to clipboard button would
not work since the clipboard buttone could not find it's container.

## Description
This change will save the container in a ref and pass it to the
clipboard as a prop.

Fixes #9254


linuspahl (approved):
Works as expected.

PR #9271, opened 2020-10-26
Render conditionals for Cloud specific elements

Gaya:
We need to be able to hide certain UI elements when conditions apply.

## Description
In the case of cloud we need to be able to hide elements and disable whole pages altogether. This requires a conditional check, a higher order component that allows for easy hiding, and a function to filter any pages from registering in the app's menu and routes.

## Motivation and Context
To allow hiding / changing certain parts of the UI

## How Has This Been Tested?
Unit tested and tested using to hide specific elements in the UI and complete pages / routes.

## Types of changes

## Checklist:


graylog2-web-interface/src/util/conditional/isCloud.js
```diff
@@ -0,0 +1,9 @@
+// @flow strict
+
+import AppConfig from '../AppConfig';
+
+function isCloud(): boolean {
```
edmundoa, line 5: I'm failing to see why we need this at the moment, could you please explain what the reasoning behind it is?

graylog2-web-interface/src/util/conditional/filterMenuItems.js
```diff
@@ -0,0 +1,24 @@
+// @flow strict
+
+import isCloud from './isCloud';
+
+type MenuItem = { path: string };
+
+function filterMenuItems(
+  menuItems: Array<MenuItem>,
+  toFilter: Array<string>,
```
edmundoa, line 9: Could we make it clearer that the array should contain the `path`s to filter? I think renaming the parameter to `pathsToFilter` would be enough.

edmundoa (commented):
I added a couple comments but it looks good to me overall.

graylog2-web-interface/src/util/conditional/filterMenuItems.js
Gaya: But wouldn't that imply that the array just contains paths, and not the whole menu items? I'd change it to `Array<string>` and just ask for the paths, makes it more reusable too. What do you think?

graylog2-web-interface/src/util/conditional/isCloud.js
Gaya: It was a fallback for when we didn't have the `AppConfig`. I'll refactor it out.

graylog2-web-interface/src/util/conditional/filterMenuItems.js
edmundoa: I think making both arguments take the same data structure makes it clearer, so I like changing them to `Array<string>` and passing paths in this particular case.

Gaya: Will do that, thanks!

edmundoa (approved):
LGTM 👍

PR #9272, opened 2020-10-26
Display correct sharable URL on entity overview pages

### LINKED ISSUE #9270: Sharable URL not correct when opening share modal on overview pages.
##  Context
Sharable URL displayed inside the share modal is not correct when opening the modal for an entry of an overview, like the dashboards list. The result is always based the current URL and we need to generate the URL based on the `entityGRN`.

linuspahl:
## Description

As described in #9270 the sharable URL displayed inside the share modal is not correct when opening the modal for an entry of an overview, like the dashboards list. With this PR the URL is always based on the `entityGRN` and not the current location.

Fixes #9270

## Types of changes

### DIFF
--- a/graylog2-web-interface/.vscode/settings.json
+++ b/graylog2-web-interface/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "cSpell.words": [
+    "formik"
+  ]
+}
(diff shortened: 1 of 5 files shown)


kmerz (approved):
Code looks fresh to me :+1: 
Happy to had run successfully on my machine :heavy_check_mark:

PR #9273, opened 2020-10-26
Prevent empty string in directory service update and create password payload

### LINKED ISSUE #9183: LDAP system user password cannot be empty
## Expected Behavior
In 4.0 the System DN and password should be optional.

## Current Behavior
Leaving the password empty leads to an error message from the API validation:
image

## Possible Solution
Make the optional fields optional in the REST DTO.

1. Create a LDAP authentication service
2. Leave System DN/password empty
3. Click "test connection"

## Your Environment

* Graylog Version: 4d38edcefee9092d50c8f781112b2af41b776cca

linuspahl:
## Description
Previously it was possible that the directory service payload contains a password with an empty string.

This was the case when a user:
1. Fills out the password input in the server config step
2. Goes to the another step
3. Goes back to the server config step
4. Removes the previously defined password
5. Triggers a connection test

With this PR we ensure that we do not submit an empty string for the password value.

Fixes #9183

## Types of changes


kmerz (approved):
LGTM :+1: 

PR #9275, opened 2020-10-26
Check periodically if running ES is compatible with initial ES version.

### LINKED ISSUE #9246: Upgrade to ES 7.x from 6.8 requires reboot in order for Graylog search to work
## Expected Behavior
Upgrade to ES 7.x from 6.8 should not require a reboot of the Graylog Servers for Search to work properly

## Current Behavior
Upgraded Graylog to 4.0, upgraded ES to 7.x, and got an error on the search page of a boolean type incompatibility in both the Message Count and All Messages Boxes.  Restarted the service and then began getting stream permission issues on the search screen even though stream permissions had not changed and user had rights to all streams.  Rebooted all Graylog servers in cluster, and everything worked upon the servers coming back up.

1. Upgrade GL 3.3.x to 4.0
2. Upgrade ES 6.8 to 7.x
3. Issue will be apparent on the search screen after the upgrade

## Context
This issue would require unnecessary down time of the servers and could lead to an increase in support cases do to how catastrophic the failures appear to be even though the fix is a simple reboot

## Your Environment
This occurred in both Aaron's environment and the SE dev environment in AWS

* Graylog Version: 4.0
* Java Version: 8
* Elasticsearch Version: 7.7 and 7.9
* MongoDB Version:4.0 and 4.2
* Operating System: Ubuntu
* Browser version: Chrome Latest

### LINKED ISSUE #9259: Periodically check if detected ES version is the same major version than current ES version
## What?

Graylog is currently probing ES nodes once during startup to find out the current ES version.
Graylog should periodically do this and compare major versions. If the currently running major version is not matching the one detected during startup, it should raise a notification.

## Why?

As a Graylog admin, I would like to get a heads up if the current ES version is not matching the one which was detected during startup, as this could indicate malfunctions.

dennisoelkers:
## Motivation and Context

Starting with 4.x, the ES version is probed during startup and the code corresponding to the major version is bound. If ES is upgraded to a different major version while Graylog is running, the ES code bound might not be compatible to the ES version running. This can result in indexing/search errors.

To inform users about this condition, this PR is adding a periodical running every 30 seconds, which probes for the current ES version and compares it against the one probed initially during startup. If both are incompatible, a system notification is raised. If they are compatible, it is fixed. If either the version is overridden manually or no version can be probed, no change is performed (although an error is logged in the latter case).

An initial idea was to start the periodical with a delay of 30 seconds, but this would lead to the notification persisting after restarting Graylog, which might irritate users. Therefore the periodical is started immediately, resulting in a (mostly) unneeded first probe.

Fixes #9259.
Fixes #9246.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
@@ -1,0 +1,124 @@
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Graylog is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Graylog.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.graylog2.periodical;
+
+import org.graylog2.notifications.Notification;
+import org.graylog2.notifications.NotificationService;
+import org.graylog2.plugin.Version;
+import org.graylog2.plugin.periodical.Periodical;
+import org.graylog2.storage.ElasticsearchVersion;
+import org.graylog2.storage.versionprobe.VersionProbe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.net.URI;
+import java.util.List;
+import java.util.Optional;
+
+public class ESVersionCheckPeriodical extends Periodical {
+    private static final Logger LOG = LoggerFactory.getLogger(ESVersionCheckPeriodical.class);
+    private final Version initialElasticsearchVersion;
+    private final Optional<Version> versionOverride;
+    private final List<URI> elasticsearchHosts;
+    private final VersionProbe versionProbe;
+    private final NotificationService notificationService;
+
+    @Inject
+    public ESVersionCheckPeriodical(@ElasticsearchVersion Version elasticsearchVersion,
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public boolean startOnThisNode() {
+        return true;
+    }
+
+    @Override
+    public boolean isDaemon() {
+        return false;
```
bernd, line 78: I think this can run in the daemon thread pool. What do you think?

graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
```diff
@@ -0,0 +1,125 @@
...
+
+        final Optional<Version> probedVersion = this.versionProbe.probe(this.elasticsearchHosts);
+
+        probedVersion.ifPresent(version -> {
+            if (compatible(this.initialElasticsearchVersion, version)) {
+                notificationService.fixed(Notification.Type.ES_VERSION_MISMATCH);
+            } else {
+                LOG.warn("Elasticsearch version currently running (" + version.toString() + ") is incompatible with the " +
+                        "one Graylog was started with (" + initialElasticsearchVersion.toString() + ")" +
+                        " - a restart is required!");
```
bernd, line 111: Tiny nitpick: use placeholder instead of manual string concat makes it a bit more readable. :smile: 
```suggestion
                LOG.warn("Elasticsearch version currently running ({}) is incompatible with the " +
                        "one Graylog was started with ({}) - a restart is required!", version, initialElasticsearchVersion);
```

bernd (approved):
LGTM apart from two smaller things :+1:

graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
dennisoelkers: Indeed! :)

graylog2-server/src/main/java/org/graylog2/periodical/ESVersionCheckPeriodical.java
dennisoelkers: Good point!

bernd (approved):

PR #9278, opened 2020-10-26
HTTP header authentication realm

bernd:
This implements the core functionality of the legacy [SSO Plugin](https://github.com/Graylog2/graylog-plugin-auth-sso).

It offers the possibility to use a trusted HTTP header to login a user.


graylog2-server/src/main/java/org/graylog2/security/realm/HTTPHeaderAuthenticationRealm.java
```diff
@@ -0,0 +1,161 @@
...
+            if (isBlank(username)) {
+                LOG.warn("Skipping request with trusted HTTP header <{}> and blank value", config.usernameHeader());
+                return null;
+            }
+
+            if (inTrustedSubnets(headersToken.getRemoteAddr())) {
+                return doAuthenticate(username);
+            }
+
+            LOG.warn("Request with trusted HTTP header <{}={}> received from <{}> which is not in the trusted subnets: {}",
```
mpfz0r, line 97:
```suggestion
            LOG.warn("Request with trusted HTTP header <{}={}> received from <{}> which is not in the trusted subnets: <{}>",
```

mpfz0r (approved):

PR #9279, opened 2020-10-26
Fix session activity indicator for users

### LINKED ISSUE #9175: User `session_active` attribute is not being updated instantly
The `session_active` attribute of an user is not being updated instantly, when a user logs in or out. The status can be viewed on the users overview and user details page.

* Graylog Version: Current development version for 4.0

### LINKED ISSUE #9241: Status for logged on users is incorrect
## Expected Behavior

When multiple users are logged into a cluster, their status should show as green.

## Current Behavior

Currently, if multiple users are logged into the UI, they're not showing as logged in.
Screen Shot 2020-10-22 at 12 08 13 PM

In this case, me, Joe and Or were logged into the UI.

bernd:
The session activity indicators on the users overview and the user
details page were broken because of the following:

1. The `GET /users/<username>` endpoint didn't load the session
   information for the given user
2. The code tried to find the user sessions by username but we recently
   switched to user IDs for the principal session key

This change also refactors the session lookup to avoid code duplication
in several places.

Fixes #9175
Fixes #9241

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -150,8 +150,8 @@ public class UsersResource extends RestResource {
     @ApiOperation(value = "Get user details", notes = "The user's permissions are only included if a user asks for his " +
             "own account or for users with the necessary permissions to edit permissions.")
     @ApiResponses({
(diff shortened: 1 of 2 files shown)


bernd:
@linuspahl @mpfz0r You probably want to use the "Hide whitespace changes" option to review this. Sorry... :wink:

mpfz0r (approved):

PR #9285, opened 2020-10-27
Ensure a user has at least reader or admin role

### LINKED ISSUE #8752: Ensure that at least 'Reader' or 'Admin' role is present for a user
A user needs at least `reader` or `admin` role. But right now a  admin can remove all roles from a user in the new user edit page.

kmerz:
## Motivation
Prior to this change, an admin could remove all roles from a user which
would render the user useless.

## Description
This change will check if the roles after a deletion will at least
contain a reader or an admin role and otherwise prevent deletion and
showing a warning.

Fixes #8752


linuspahl:
@kmerz These changes look good to me. What do you think about implementing the same warning for the `RoleEdit` page? Right now an admin can edit the reader and admin role and remove a user who has only this one role.

linuspahl (approved):
Looks good to me!

Issue #9244, opened 2020-10-22
Teams / Permissions header navigation does not clearly indicate active page


lennartkoopmann:
![2020-10-22-112221_311x126_scrot](https://user-images.githubusercontent.com/35022/96901497-f977fd00-1458-11eb-9a9d-f9c45841051a.png)
![2020-10-22-112211_278x119_scrot](https://user-images.githubusercontent.com/35022/96901500-fa109380-1458-11eb-9fa9-eed1bae922bd.png)

Multiple buttons in the header navigation of new teams and permissions pages are not clearly indicating what page I am currently on, leading to severe confusion about where I am or what else I can do. In the beginning I even thought buttons are broken because the page would not change after a click, only to find out that I was on that page already.

## Your Environment

* Graylog Version: 4.0.0-beta.2

PR #9286, opened 2020-10-27
Provide active className for Buttons wrapped with LinkContainer if link target equals current pathname

linuspahl:
## Description
With https://github.com/Graylog2/graylog2-server/pull/9135 we've added an own implementation of `LinkContainer`, because the `LinkContainer` provided by `react-router-bootstrap` changed its behaviour with a previous update.
Instead of defining `onClick` for its children it transformed them into a link.

With this PR we are adding the functionality that `Buttons` wrapped with `LinkContainer` receive the `active` class, if the target of the `LinkContainer` equals the current pathname.

We are also removing not needed `className="active"` props from `Buttons` wrapped with `LinkContainer`.
There are still some buttons where we need to set this prop manually.

Please note there is a linter warning which is not related to these changes.

Fixes: #9244

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/graylog/router.jsx
+++ b/graylog2-web-interface/src/components/graylog/router.jsx
@@ -1,17 +1,16 @@
 // @flow strict
 import * as React from 'react';
 import { useCallback } from 'react';
-import { Link } from 'react-router-dom';
+import { Link, useLocation } from 'react-router-dom';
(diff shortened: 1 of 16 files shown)


kmerz (approved):
LGTM :+1: 
Works like a charm on my machine :heavy_check_mark:

PR #9287, opened 2020-10-27
Per request authorization cache

mpfz0r:
- Revert the event based cache invalidation

-  Only cache authorization results within a request context

 - Introduce a per-request random ID by using a commonly known HTTP
   header: X-Request-Id
   This ID is either set by clients, load balancers, etc.
   If it's not found, we create our own.

 - The request headers were already accessible within Shiro's ThreadContext.
   But this is now refactored into a separate filter: ShiroRequestHeadersBinder

 - Finally add the request ID into the AuthorizationCacheKey

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1872


bernd (approved):

bernd:
Waiting for the build to turn green.

PR #9169, opened 2020-10-15
Remove IntelliJ's `@NotNull` annotations.

dennisoelkers:
## Motivation and Context

In some of our tests usages of IntelliJ's `@NotNull` annotation have slipped in. This change is removing them and replacing them with `javax.annotation.Nonnull`.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #9292, opened 2020-10-28
Remove Jetbrains NotNull annotation

danotorrey:
Use the `javax.annotation.Nonnull` annotation instead of `org.jetbrains.annotations.NotNull` which is not available in all environments. The same change was made in https://github.com/Graylog2/graylog2-server/pull/9169 and ports the compatible `3.3` file changes over to `cloud-3.3`.

This incorrect import only caused the following error in some environments, such as running the Graylog Server with the [Docker tool](https://github.com/Graylog2/graylog-project-internal/blob/5dbe21d3b37fca04570620bf177a3525e1104fb3/runner/README.md).

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/security/ldap/LdapConnectorSSLTLSIT.java
+++ b/graylog2-server/src/test/java/org/graylog2/security/ldap/LdapConnectorSSLTLSIT.java
@@ -22,7 +22,6 @@ import org.apache.directory.ldap.client.api.LdapNetworkConnection;
 import org.assertj.core.api.Assertions;
 import org.graylog2.rest.models.system.ldap.requests.LdapTestConfigRequest;
 import org.graylog2.security.DefaultX509TrustManager;
-import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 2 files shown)


danotorrey:
Closing this PR, and will instead selectively port https://github.com/Graylog2/graylog2-server/pull/9169 over to cloud-3.3.

bernd (changes requested):
Thanks! Can you please add the `org.jetbrains.annotations.NotNull` class to the `config/forbidden-apis/signatures.txt` file? That way we will fail the build in the future if that class is used again.

danotorrey:
Reopening this PR, and will manually adjust the annotation to use the static-checking `javax.annotation.Nonnull` as was done in https://github.com/Graylog2/graylog2-server/pull/9169. I cannot directly port this PR to cloud-3.3, since it contains `4.0`-only changes.

> Can you please add the `org.jetbrains.annotations.NotNull` class to the `config/forbidden-apis/signatures.txt` file? 
That way we will fail the build in the future if that class is used again.

Thanks @bernd. I was wondering if we had support for this. Will do.

bernd:
@danotorrey @dennisoelkers just told me that forbiddenapis doesn't work in this case because it cannot find the annotation class. He tried that already.

bernd (approved):

Issue #9249, opened 2020-10-22
Explain how to share entities on user edit page.


linuspahl:
## What?
Previously, sharing entities like dashboards and streams has been part of the user edit page, now it can be done while editing / viewing an entity.

We plan to explain this change on the user edit and details page, by adding an alert above the roles section. The alert should also contain an example of the share button.

bernd:
@linuspahl We might also want to consider adding some kind of "inverted" share dialog to users and teams pages. That way a user with the appropriate permissions can go to a user account or team and create shares for entities that user is an owner of.

PR #9297, opened 2020-10-28
Display info about permissions update on user edit and details page

linuspahl:
## Description
With Graylog 4.0 we've update the permissions system. Previously, sharing entities like dashboards and streams has been part of the user edit page, now it can be done with the share button of an entity, which is e.g part of the entities overview page.

With this PR we explain this change on the user edit and details page, so users will understand where they can grant permissions for an entity.

Initially I planned to include the share button component inside the info text, but this way the text would take up even more space.

Fixes: #9249

## Types of changes


kmerz (approved):
LGTM :+1: 
Worked on my computer :heavy_check_mark:

PR #9299, opened 2020-10-28
Trim string of username when creating a user

### LINKED ISSUE #8593: Username allows trailing white space characters
I've been several times able to create a new user that has a trailing whites space characters. It's probably caused by copying & pasting from Microsoft Office products (they tend to add space to selections). Graylog probably should make sure that the username is trimmed (both left and right). Because username can't be modified I've had to delete the user and recreate it to recrify the situation.

kmerz:
## Motivation
Prior to this change, if a user was creating a user with leading spaces
the leading spaces were stored into the backend and the user would have
to login with this leading spaces.

## Description
This change will trim the username input before submiting it to the
server.

Fixes #8593


linuspahl (approved):
Looks good to me!

PR #9303, opened 2020-10-29
Add target as href to `LinkContainer` children.

### LINKED ISSUE #9298: Navigation Tabs do not hold actual link to new page which does prevent opening in new tab
## Expected Behavior
When clicking to open a nav item in a new tab then the user expects the desired page opened in that new tab.

## Current Behavior
When opening a nav item with the middle click (open in new tab) the new tab opens with the same page as we started.

## Possible Solution
Contain the actual target link instead of `#`.

1. Open new tab from navigation header item

## Your Environment
* Graylog Version: graylog 4.0 beta-1

dennisoelkers:
## Motivation and Context

Before this change, the `LinkContainer` component did add an `onClick` handler to its children to perform a route change without reloading the app. This lead to the element being created not having an `href` prop, so clicking on it and opening it in a new tab did not work.

This change is adding an `href` prop containing the value of the `to` prop to its children, so the html element contains a valid target.

Fixes #9298.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/graylog/LinkContainer.test.jsx
+++ b/graylog2-web-interface/src/components/graylog/LinkContainer.test.jsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { render } from 'wrappedTestingLibrary';
+import { render, waitFor } from 'wrappedTestingLibrary';
 import { fireEvent } from '@testing-library/dom';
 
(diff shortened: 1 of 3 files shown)


kmerz (approved):
Code Looks solid to me :+1: 

PR #9307, opened 2020-10-29
Avoid using throw by displaying an Alert

### LINKED ISSUE #8756: Be more fault tolerant when dealing with errors in the new users and role pages.
On some places we throw an error (`thorw new Error()`) when we do not find an item. But we can be more tolerant with that so the user can fix that issue.
Also when dealing with errors while fetching data. If e.g. the request fails witch fetches all users for the UsersOverview, we receive `undefined` as a result for the action. In this case we will defined `undefined` for the UsersOverview component state, which results in frontend errors.

kmerz:
## Motivation
Prior to this change, when we had an runtime error we rendered the error
page to the user even though he might have been able to workarround an
issue.

## Description
This change implements a ErrorAlert component which is closable and
which will display an runtime error or warning. It replaces the
warnings already implemented and does also replace the `throw` errors
for users and roles components.

Fixes #8756

### DIFF
--- a/graylog2-web-interface/src/components/permissions/RolesSelector.jsx
+++ b/graylog2-web-interface/src/components/permissions/RolesSelector.jsx
@@ -10,7 +10,7 @@ import type { PaginatedRoles } from 'actions/roles/AuthzRolesActions';
 import AuthzRolesDomain from 'domainActions/roles/AuthzRolesDomain';
 import Role from 'logic/roles/Role';
 import { Button } from 'components/graylog';
-import { Select, Spinner } from 'components/common';
+import { Select, Spinner, ErrorAlert } from 'components/common';
 
 type Props = {
   assignedRolesIds: Immutable.Set<string>,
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/src/components/permissions/RolesSelector.jsx
```diff
@@ -44,7 +44,7 @@ const _options = (roles, assignedRolesIds, identifier) => roles
   .toArray()
   .map((r) => ({ label: r.name, value: r.name, role: r }));
 
-const _assignRole = (selectedRoleName, roles, onSubmit, setSelectedRoleName, setIsSubmitting) => {
+const _assignRole = (selectedRoleName, roles, onSubmit, setSelectedRoleName, setIsSubmitting, setError) => {
```
linuspahl, line 47: Lets reset the error somewhere, otherwise it will not get removed on a successful request.

graylog2-web-interface/src/components/roles/RoleEdit/UsersSelector.jsx
```diff
@@ -88,7 +89,9 @@ const UsersSelector = ({ role, onSubmit }: Props) => {
     })));
 
     if (!userOverview) {
-      throw new Error(`Unable to find user with name ${user} in ${users.map((u) => u.username).join(', ')}`);
+      setError(`This is an exceptional error! Unable to find user with name ${user} in ${users.map((u) => u.username).join(', ')}`);
```
linuspahl, line 92: We should somewhere reset the error in this case as well.

linuspahl (changes requested):
Just added a suggestion.

linuspahl (approved):
Looks good to me!

PR #9309, opened 2020-10-29
Clean-up ESLint configuration

linuspahl:
## Description
This PR is cleaning up the ESLint configuration, by moving all settings like `env` and `import/resolver` from the `.eslintrc.json` to the `eslint-config-graylog` package.

This way we only need to extend the `eslint-config-graylog` either in the `.eslintrc` or in the `eslintConfig` of the  `package.json`.

## Motivation and Context
Currently there are some plugins which reference the `.eslintrc.json` of the core project in their `lint` script.
`"lint": "eslint --ext js,jsx src --config ../../graylog2-server/graylog2-web-interface/.eslintrc.json"`

It is cleaner to move all settings to the `eslint-config-graylog` and only rely on this package to configure ESLint  in plugins.
This is also necessary to enable the built-in ESLint editor support for these projects.

Please note we need to merge this PR together with the related plugin PR's to ensure we can run ESLint in these projects correctly.

## Types of changes


dennisoelkers (changes requested):
Looks in general and helps cleaning up the config. I think we need the explicit `"parser": "babel-eslint"` declaration in `eslint-config-graylog`, at least it was present in the config files the PR is removing.

dennisoelkers (approved):

PR #9310, opened 2020-10-29
Include grants for "everyone" in entities for user overview

bernd:
Refs Graylog2/graylog-plugin-enterprise#1718

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1885


mpfz0r (approved):
LGTM.
We can still extend the `ShareResponse` with the info on how something is shared with a user..

PR #9313, opened 2020-10-30
Add missing binding of GRNRegistry for journal decode command

### LINKED ISSUE #9304: Journal Decode command broken
## Expected Behavior

`$ java -jar graylog.jar journal decode 0 -f ./graylog.conf` should not throw a binding error

## Current Behavior

```
$ java -jar graylog.jar journal decode 0 -f ./graylog.conf
2020-10-29 11:40:12,268 WARN : org.graylog2.shared.security.tls.DefaultTLSProtocolProvider - JRE doesn't support all default TLS protocols. Changing <[TLSv1.2, TLSv1.3]> to <[TLSv1.2]>
2020-10-29 11:40:12,686 ERROR: org.graylog2.bootstrap.CmdLineTool - Guice error (more detail on log level debug): Could not find a suitable constructor in org.graylog.grn.GRNRegistry. Classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private.
Exception in thread "main" com.google.inject.CreationException: Unable to create injector, see the following errors:

1) Could not find a suitable constructor in org.graylog.grn.GRNRegistry. Classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private.
  at org.graylog.grn.GRNRegistry.class(GRNRegistry.java:37)
  while locating org.graylog.grn.GRNRegistry
    for the 4th parameter of org.graylog2.shared.bindings.providers.ObjectMapperProvider.<init>(ObjectMapperProvider.java:84)
  at org.graylog2.shared.bindings.ObjectMapperModule.configure(ObjectMapperModule.java:45)

1 error
	at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:543)
	at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:159)
	at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:106)
	at com.google.inject.Guice.createInjector(Guice.java:87)
	at org.graylog2.shared.bindings.GuiceInjectorHolder.createInjector(GuiceInjectorHolder.java:34)
	at org.graylog2.bootstrap.CmdLineTool.setupInjector(CmdLineTool.java:381)
	at org.graylog2.bootstrap.CmdLineTool.run(CmdLineTool.java:196)
	at org.graylog2.bootstrap.Main.main(Main.java:50)
```

1. Run the above command

## Your Environment

* Graylog Version: `4.0.0-SNAPSHOT-20201028172840`
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

thll:
## Description

The object mapper needs a binding to a GRNRegistry. In the context of
the journal decode command we don't care about GRNs, so we just
add a binding to an empty registry.

## Motivation and Context

Fixes #9304 

## How Has This Been Tested?

Tested from the IDE that the decode command works again

## Types of changes

## Checklist:


mpfz0r:
I'm wondering if we should add this to the ObjectMapperModule instead.
This way we can prevent missing out on this dependency.
Maybe @bernd has an opinion on this? 

```
--- graylog2-server/src/main/java/org/graylog2/shared/bindings/ObjectMapperModule.java
+++ graylog2-server/src/main/java/org/graylog2/shared/bindings/ObjectMapperModule.java
@@ -18,6 +18,7 @@ package org.graylog2.shared.bindings;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
+import org.graylog.grn.GRNRegistry;
 import org.graylog2.plugin.inject.Graylog2Module;
 import org.graylog2.shared.bindings.providers.ObjectMapperProvider;
 import org.graylog2.shared.plugins.GraylogClassLoader;
@@ -41,6 +42,7 @@ public class ObjectMapperModule extends Graylog2Module {
         // the ObjectMapperProvider requires at least an empty JacksonSubtypes set.
         // if the multibinder wasn't created that reference will be null, so we force its creation here
         jacksonSubTypesBinder();
+        bind(GRNRegistry.class).toInstance(GRNRegistry.createWithBuiltinTypes());
         bind(ClassLoader.class).annotatedWith(GraylogClassLoader.class).toInstance(classLoader);
         bind(ObjectMapper.class).toProvider(ObjectMapperProvider.class).asEagerSingleton();
     }
```

thll:
> I'm wondering if we should add this to the ObjectMapperModule instead.
> This way we can prevent missing out on this dependency in the future.

I'd prefer that as well, but then we'll have to remove the binding here: https://github.com/Graylog2/graylog2-server/blob/2a1cf59086f5e182e45436e50fc7808e6a5ef721/graylog2-server/src/main/java/org/graylog/grn/GRNModule.java#L30
because otherwise it will be bound twice, which Guice doesn't allow. And removing it from that module doesn't really look right. After all this is the module I would use to look for the binding.
We could split `GRNModule` into a `GRNModule` and a `GRNTypesModule` but then we still have to make sure to always bind the `GRNModule` in both the server and the journal command. So that doesn't buy us much.

Any other ideas?

bernd:
@thll @mpfz0r How about using `install(new GRNModule())` in `ObjectMapperModule`? The `install()` method checks if the given module has already been installed. To make sure that check works, we need to override the `equals()` method in `GRNModule` or we create a constant with a singleton instance of the module and make the constructor private.

thll:
> @thll @mpfz0r How about using `install(new GRNModule())` in `ObjectMapperModule`? The `install()` method checks if the given module has already been installed. To make sure that check works, we need to override the `equals()` method in `GRNModule` or we create a constant with a singleton instance of the module and make the constructor private.

I like the idea. I'm gonna try how that looks like.

thll:
`install(new GRNModule())` and making the module a singleton would probably work, but the problem is that the `GRNModule` will pull too many dependencies into the journal decode command. I've split the module so that we can install the types only for the server and have the registry bound via the object mapper bindings for both server and journal command.

mpfz0r (approved):
LGTM. Thanks :+1:

bernd (approved):
I am not super happy with binding the GRN module in the object mapper module, but I also don't have a better idea at the moment.

PR #9342, opened 2020-11-03
Add migration for legacy LDAP/AD configuration

### LINKED ISSUE #9078: Create migration to convert LDAP/AD configuration to the new format
Existing LDAP and Active Directory configurations using the old system should be migrated to the new authentication service system.

bernd:
This creates a migration to create a new authentication service based on
an existing legacy LDAP/AD configuration.

Since the legacy LDAP/AD configuration doesn't contain all information
we need for a new authentication service, we guess some config fields
and create a system notification to make sure an admin will notice it
and reviews and activates the migrated configuration.

Closes #9078

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.migrations;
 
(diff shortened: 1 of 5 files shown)


mpfz0r (approved):
OK for the backend part. (Frontend works fine as well)

kmerz (approved):
LGTM :+1:

PR #9355, opened 2020-11-04
Display an warning if a grantee was not found

kmerz:
## Motivation
Prior to this change, if grantee was not found for the share dialog the
gui threw an error.

## Description
This change will display the grantees ID together with an exclamation
icon and a note that this is an error.

Graylog (19)

Fixes #9355


linuspahl (approved):
Looks good to me!

PR #9356, opened 2020-11-04
Fix JDK TLS protocol detection

bernd:
Previously we only checked the protocols that are enabled by default
instead of checking all available.

This allows the usage of TLSv1.3 in OpenJDK 8u272 and later.

See: https://blog.adoptopenjdk.net/2020/10/adoptopenjdk-8u272-1109-and-1501-available/


mpfz0r (approved):
Cool. Thanks. Tested with `testssl.sh`

PR #9360, opened 2020-11-04
Hide edit profile and password section for external user

### LINKED ISSUE #9222: Avoid modification of external users
## Expected Behavior

Admins shouldn't be able to exit external users. (synced by LDAP, AD, and others)

## Current Behavior

It's currently possible to edit external users.

kmerz:
## Motivation
Prior to this change, synced (external) users had the profile and
password section on their edit page, which should be in sync with the
configured authentication service and not be managed by graylog itself.

## Description
This change will hide these sections so a admin does not change anything
and expects a different password.

Fixes #9222

### DIFF
--- a/graylog2-web-interface/src/components/users/UserEdit/UserEdit.jsx
+++ b/graylog2-web-interface/src/components/users/UserEdit/UserEdit.jsx
@@ -1,6 +1,7 @@
 // @flow strict
 import * as React from 'react';
 import { useContext } from 'react';
+
 import UsersDomain from 'domainActions/users/UsersDomain';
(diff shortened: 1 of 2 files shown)


bernd (changes requested):
@kmerz Works for me, thank you! :+1: 

Should we add a notification to the user edit page to indicate that the user is an external user and thus cannot be edited?
In 3.3 we had this for the password section. How about creating a more generic one for the new user edit page?

image

linuspahl (approved):
Looks good to me!

bernd (approved):

PR #9365, opened 2020-11-05
Improve UserEdit and RoleEdit tests by splitting them up

linuspahl:
## Description
Previously the `UserEdit.test.jsx` and `RoleEdit.test.jsx` failed occasionally. The main problem was, that we are doing multiple requests on this page and some of them do not trigger an UI change which can be handled with `screen.findBy*`. We would have to wrap each promise mock in an act in every test, to make sure we handle them correctly. But even in this case some tests would run longer that one second.

With this PR we are splitting the both tests up in multiple files, one for each section.


kmerz (approved):
Code LGTM :+1:

PR #9366, opened 2020-11-05
Remove unusual domain actions usage in auth backend edit and create component

linuspahl:
## Description
Previously, we've implemented closures for some authentication domain actions, to be able to control the notification, when updating and creating an authentication service. This was necessary because there are cases were we do not want to trigger a notification.
```
AuthenticationDomain.create(backendCreateNotificationSettings)(payload);
```
This is an unusual usage which can be confusing, we now call the authentication actions directly and tigger the user notification inside the create and update component.

## Types of changes


kmerz (approved):
Code LGTM :+1: 

PR #9367, opened 2020-11-05
Fix Input error color for Inputs which use the help prop to display errors

linuspahl:
## Description
With https://github.com/Graylog2/graylog2-server/pull/9059 we've added the `error` prop for our `Input` component, to be able to display errors next to the help text.

With these changes we've defined a color for help text, to ensure the color does not change when the related `Input` has the prop `bsStyle="error"`.

A side effect is that errors of `Input`'s which still use `help` prop, do not have the correct color:
image

With this PR we are removing this styling change. 
This also means `Input`'s which already use the new error prop will look like this: 
image
(The help text has the error color when an error occurs)

But this is still a better solution than the alternative in terms of usability. I had a look at the cases where we use the `help` prop to display an error, but it is not a suitable solution to adjust all these cases right now.

## Types of changes

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1914

### DIFF
--- a/graylog2-web-interface/src/components/common/InputDescription.jsx
+++ b/graylog2-web-interface/src/components/common/InputDescription.jsx
@@ -9,10 +9,6 @@ const ErrorMessage = styled.span(({ theme }) => `
   color: ${theme.colors.variant.danger};
 `);
 
-const HelpMessage = styled.span(({ theme }) => `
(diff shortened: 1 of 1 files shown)


kmerz (approved):
LGTM :+1: 

PR #9374, opened 2020-11-05
Implement and enforce user account status

### LINKED ISSUE #9077: Implement a user status
It's currently not possible to enable/disable users in our database. Authentication services will have this capability and we should be able to mirror that status when syncing users.

We should probably use an actual status enum instead of a simple flag so we can introduce more stauses later. (e.g. `DELETED`)

mpfz0r:
Fixes #9077


graylog2-server/src/main/java/org/graylog/security/authservice/ldap/LDAPEntry.java
```diff
@@ -35,6 +35,8 @@
 
     public abstract String base64UniqueId();
 
+    public abstract ADUserAccountControl userAccountControl();
```
bernd, line 38: See comment in `LDAPUser`.

graylog2-server/src/main/java/org/graylog/security/authservice/ldap/LDAPUser.java
```diff
@@ -22,6 +22,8 @@
 public abstract class LDAPUser {
     public abstract String base64UniqueId();
 
+    public abstract ADUserAccountControl userAccountControl();
```
bernd, line 25: I don't think we should put AD specific code into the `LDAPEntry` and `LDAPUser` objects. How about using the `AccountStatus` instead? That makes it independent from AD. We could extract the `AccountStatus` into a separate enum out of the `User` interface to make it a bit more generic.
What do you think?

I also think we should only have an `accountStatus()` method in `LDAPUser` instead of also putting it into the `LDAPEntry` object. The `LDAPEntry` object is supposed to be generic. We can move the `userAccountControl` parsing into the `createLDAPUser()` method in `UnboundLDAPConnector`.

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
```diff
@@ -257,13 +258,17 @@ private ADAuthServiceBackendConfig buildTestConfig(@Nullable AuthServiceBackendD
...
-                    AD_OBJECT_GUID, user.base64UniqueId(),
-                    config.userNameAttribute(), user.username(),
-                    config.userFullNameAttribute(), user.fullName(),
-                    "email", user.email()
-            ));
+            userDetails.put("user_details", ImmutableMap.<String, String>builder()
+                    .put("dn", user.dn())
+                    .put("account_disabled", String.valueOf(user.userAccountControl().accountIsDisabled()))
+                    .put("normal_user_account", String.valueOf(user.userAccountControl().isUserAccount()))
+                    .put("password_expired", String.valueOf(user.userAccountControl().passwordExpired()))
```
bernd, line 265: How about only putting in the `userAccountControl` attribute as key and then all values as named flags?

```
"userAccountControl": "ACCOUNTDISABLED|NORMAL_ACCOUNT|PASSWORD_EXPIRED"
```
I think that would make it a bit clearer to AD admins. `account_disabled`, `normal_user_account` and `password_expired` are not really AD attributes.

graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackend.java
```diff
@@ -77,23 +77,24 @@ public LDAPAuthServiceBackend(UnboundLDAPConnector ldapConnector,
...
                     .authServiceId(backendId())
+                    .accountIsEnabled(true)
                     .base64AuthServiceUid(userEntry.base64UniqueId())
                     .username(userEntry.username())
                     .fullName(userEntry.fullName())
                     .email(userEntry.email())
                     .defaultRoles(backend.defaultRoles())
                     .build());
 
+            if (!authCredentials.isAuthenticated()) {
```
bernd, line 91: Can you please add a comment about why we first provision and then authenticate?

bernd (changes requested):
First round of review.

graylog2-server/src/main/java/org/graylog/security/authservice/ldap/LDAPUser.java
mpfz0r: I could remove it from `LDAPEntry`, because it would be still accessible within the attributes map.
But I'm not sure about using `AccountStatus` for that. A simple enum cannot reflect the `userAccountControl` flags.

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
mpfz0r: sure, I can do that.

graylog2-server/src/main/java/org/graylog/security/authservice/ldap/LDAPUser.java
bernd: Ah, right. But do we need all the flags? We are only interested in the the account status here, no?

mpfz0r: If we want to display them in the test result.
I don't see any other way right now :-/

bernd: Okay, if it's just for the test result I am fine with leaving the flags out. :smile: We can still show if a user is enabled/disabled in the test result.

mpfz0r: ok. :wastebasket:

graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackend.java
mpfz0r: I've changed my mind about changing this. We'll have to refactor this at some point anyhow.
I'm leaving this untouched, as this is too close to release

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
```diff
@@ -98,12 +98,17 @@ public ADAuthServiceBackend(UnboundLDAPConnector ldapConnector,
                     return Optional.empty();
                 }
             }
+            if (!userEntry.accountIsEnabled()) {
```
bernd, line 101: How about moving this up in front of the `if (!authCredentials.isAuthenticated()) {` line? That way we get the warning when a disabled user tries to log in. Otherwise we will never see the warning because the authentication doesn't work.

bernd (changes requested):
Tested with local users and AD users. Works! :+1: 

Three smaller comments:

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
mpfz0r: sure. makes sense

bernd (approved):

PR #9377, opened 2020-11-05
Do not wrap group sync details plugin with a section component.

linuspahl:
## Description
With this PR we are changing the structure of the group sync details section on the auth backend details page.
These changes are necessary to display the trigger team sync button for this section.

We are also moving the generic `BackendConfigDetails` component from the `ldap` dir to the `directoryService` dir.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1890


bernd (approved):
Works for me :+1:

PR #9378, opened 2020-11-05
Fix read handling of corrupted journal segments

### LINKED ISSUE #9305: Journal reader will stall when segment is truncated
## Expected Behavior

If a journal segment is truncated in a way that it abruptly ends in the middle of an entry, the journal reader should skip the broken entry and continue reading at the start of the next journal segment.

## Current Behavior

When reaching the end of the first segment, the journal reader will silently stop reading indefinitely. This leads to the journal filling up but no more messages being processed.

## Possible Solution

When the journal reader attempts to read a truncated entry, the Kafka stack will return nothing. Currently we are treating this as if we have reached the end of the journal. We should however check if the journal contains more entries (e.g. in subsequent segments) and increment our read offset until we are able to successfully read entries again.

1. Start a Graylog server with an empty journal (delete `data/journal`)
2. Stop elasticsearch
3. Send 100.000 messages, each 2048 bytes, to the server, e.g. using the RAW TCP input
4. Stop the Graylog server
5. Open the first segment of the journal (`data/journal/messagejournal-0/00000000000000000000.log`) in an editor and truncate the last entry
6. Start elasticsearch
7. Start the Graylog server. Notice that it processes some messages (up to and including offset 47342 in my case) but then stops processing altogether.

* Graylog Version: `e231e93f4cf63e54a54850d83b30e7d4609cd619`
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

bernd:
Previously the journal reader got stuck waiting when reading
from an offset that returned an empty result.

In that case the journal reader assumes that it reached the
end of the journal and waits for more messages to come in.
Once new messages have been written to the journal, the
reader is trying again with the same offset.

With a corrupted log segment, reading the same offset will
return an empty result again. Even though there are valid
messages in new segments after the corrupted one.

This put the journal reader in an infinite loop until
the corrupted journal segmet gets deleted manually or users
delete the complete journal.

The code is now checking if there are more messages after
the current read offset when an empty result is received
from the read operation. In that case it increases the
offset one by one and retries the read until the result is
not empty anymore or the end of the log has been reached.

We reproduced this issue with a journal where the first
segment was truncated at the end. Later log segments
contained valid messages.

* Rename fault-tolerant read method to `readNext`

So that the JournalDecode command uses the old method. Otherwise
decoding a faulty offset explicitly with `journal decode` would decode a
 message at the next non-faulty offset, which is probably not expected.

* Add test to simulate truncated segment

Co-authored-by: Bernd Ahlers <bernd@graylog.com>

Fixes #9305


thll:
@bernd: tried to fix the race condition we had. Can you please check if this fixes it for you?

thll:
@bernd I updated the PR to be in sync with `master`. Do you think we should merge this after the `4.1` release? If so, could you please review the fix or delegate to someone else? If we decide it's too risky to do the change, I'd like to close this as a "won't fix" rather than having it lying around unmerged even longer.

thll (approved):

Issue #9382, opened 2020-11-06
Explain reason why an unsaved search cannot be shared


bernd:
An unsaved search cannot be shared and we grey out the button. There is no explanation _why_ it cannot be shared.

![image](https://user-images.githubusercontent.com/461/98343287-e39a3880-2011-11eb-8ff6-cc955117d888.png)

It would be nice if we can add a small hover or popover like we do for share buttons when a non-owner tries to share something.

linuspahl:
Fixed with https://github.com/Graylog2/graylog2-server/pull/9383

PR #9383, opened 2020-11-06
Display explanation for share button why unsaved searches and dashboards can not be shared

linuspahl:
## Description
As described in https://github.com/Graylog2/graylog2-server/issues/9382 an unsaved search cannot be shared and there is no explanation why it cannot be shared. With this PR we are showing an explanation, that a search / dashboards needs to be saved first.


kmerz (changes requested):
A user with manage access does not get a correct message:
Graylog (20)

kmerz (approved):
Code LGTM :+1: 
(Problems with theme will be addressed in another PR)

PR #9385, opened 2020-11-06
Display synced users section on every authentication service details page

bernd:
Previously the synced users section was only available on the details page of an active authentication service.
With this PR we are refactoring the endpoint to get users for an auth service. It's now possible to get users for every authentication backend instead of only the active one. This way we can display the section on every authentication service details page.

Fixes Graylog2/graylog-plugin-enterprise#1912


graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
```diff
@@ -138,6 +166,55 @@ public void delete(@ApiParam(name = "backendId", required = true) @PathParam("ba
         dbService.delete(config.id());
     }
 
+    @GET
+    @Path("{backendId}/users")
+    @ApiOperation("Get paginated users for active authentication service backend")
```
mpfz0r, line 171:
```suggestion
    @ApiOperation("Get paginated users for an authentication service backend")
```

graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
```diff
@@ -138,6 +166,55 @@ public void delete(@ApiParam(name = "backendId", required = true) @PathParam("ba
...
+    @RequiresPermissions({RestPermissions.AUTH_SERVICE_GLOBAL_CONFIG_READ, RestPermissions.USERS_LIST})
```
mpfz0r, line 172: I'd prefer it if we used `USERS_READ`. 
Or even `GranteeService.getVisibleUsers()`

graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
```diff
@@ -138,6 +166,55 @@ public void delete(@ApiParam(name = "backendId", required = true) @PathParam("ba
...
+                Collections.singletonMap("roles", createRoleContext(userList.delegate()))
+        );
+    }
+
+    private Map<String, Object> createRoleContext(List<UserOverviewDTO> userList) {
+        final Set<String> roleIds = userList.stream()
+                .flatMap(user -> user.roles().stream())
+                .collect(Collectors.toSet());
+        try {
+            return roleService.findIdMap(roleIds).values()
```
mpfz0r, line 201: should be checking `roles:read`

mpfz0r (changes requested):

kmerz (approved):
Code LGTM :+1: 

graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
bernd: Good catch, thanks! :+1:

graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
bernd: I opted for `USERS_READ`. I think we need to have a larger discussion about fine grained permission checks. The UI somehow needs to handle this as well. We probably need placeholder user objects if the requesting user is not allowed to read a user. Otherwise there would be no indicator that there are users for that backend.

graylog2-server/src/main/java/org/graylog/security/authservice/rest/AuthServiceBackendsResource.java
bernd: thanks!

mpfz0r (approved):

PR #9391, opened 2020-11-06
Consistently Load the Proper Theme

kyleknighted:
## Description

Initially the Theme was being generated before the user was loaded. 
This has been updated by creating a User and Theme context wrapper that we can then reuse as necessary to verify that we can loading in the proper User preferences before generating the theme

## Motivation and Context

Theme would not always load the same one in user preferences consistently and some Popovers and Tooltips were not loading in the same theme as other parts of the app.

image
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/graylog/Popover.jsx
+++ b/graylog2-web-interface/src/components/graylog/Popover.jsx
@@ -1,9 +1,9 @@
-import React, { useContext } from 'react';
+import React from 'react';
 // eslint-disable-next-line no-restricted-imports
 import { Popover as BootstrapPopover } from 'react-bootstrap';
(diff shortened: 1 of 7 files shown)


linuspahl (approved):
Looks good to me!

Two parts we could improve in a follow-up PR:
As discussed we are currently setting a default theme in the `UserPreferencesContext`, if we remove this default, the local storage value is being used. This way the page loading state would already use the selected theme and would probably not differ from the theme type defined in the users preferences.

We are currently displaying a user notification when changing the theme. Imo we do not need this notification. This can be configured by providing another parameter for `PreferencesStore.saveUserPreferences`.

PR #9394, opened 2020-11-07
Make sure to use the OSS ES container for tests

bernd:
Due to the version constraint of "^6.0.0" we only used the OSS container
for ES 6.x.

Adjust ClientES7#existingTemplates to not use a wildcard ("*") argument
when creating the GetIndexTemplatesRequest object.
Using the wildcard argument with no existing index templates returns a
404 from ES and throws an exception.

This fixes the IndicesGetAllMessageFieldsES7IT test when switching to
the OSS container. The regular ES container is automatically creating
indices for x-pack features so there are always existing indices and
index templates. The OSS container doesn't create any indices and
templates by default and so the #existingTemplates call failed with a
404.


bernd:
@dennisoelkers Can you please check if the change in `ClientES7#existingTemplates` could have any unwanted consequences?  Thank you! :slightly_smiling_face:

dennisoelkers (approved):
LGTM and works!

PR #9409, opened 2020-11-09
Show users in roles overview

### LINKED ISSUE #9219: Show number of users and user excerpt in roles overview
## What?

It would be great if we can show the number of users (and a user list excerpt) that have a role assigned in the overview, similar to the teams overview page:

image

image

## Why?

This would help to get a quick overview of how many users have a role assigned.

kmerz:
## Description
This PR will display the users of a role on the roles overview page.

## Motivation and Context
All other overview pages display their relationships. So the user expects to see the users of a role on the roles overviewpage.

Graylog - Roles Overview

Fixes #9219

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authzroles/AuthzRolesResource.java
@@ -56,8 +56,10 @@ import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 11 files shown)


linuspahl (changes requested):
Could not find any problems while testing these change, but let's discuss an implementation detail.

You've added `users` to the `Roles` class. Reading the code I would expect that we also have `users` when we fetch one role, but it is only the case when we fetch all roles paginated. We could e.g. create a new class like `RoleOverview` or implement these changes similar to teams, where we return the context as a separate value when fetching the paginated teams, which we use in the teams overview. I prefer the second solution.

linuspahl (approved):
Looks good to me!

PR #9411, opened 2020-11-09
Enable TypeScript in webpack build.

dennisoelkers:
**Note: This must not be merged before the `4.0` branch was branched off.**

## Goal

This PR is enabling TypeScript transpilation in the webpack build. It is supposed to not intervene with the build as it is in any way, but to allow optional transpilation of `.ts(x)` files. Type checking is performed during the development build when run with `yarn start`. TypeScript modules can import all other modules from core. A full type check of the project is not meant to work yet and will return a huge amount of errors (mostly due to `.js(x)` files typed with flow containing type annotations).

We have chosen to use the babel preset to transpile TypeScript, which allows us to stick with `babel-loader` for both `.js(x)` and `.ts(x)`. To compensate its lack of type checking, the `ForkTsCheckerWebpackPlugin` webpack plugin is used, which decouples emitting code and type checking, allowing us to benefit from faster incremental builds in dev mode.

The TypeScript config currently has pretty lax settings to allow a smoother transition. When we have ported existing flow code to TypeScript, we can enable more settings and gradually improve type strictness.

~~**Note:** We are currently using TypeScript `4.1.1-rc`, due to a bug in `4.0.5` and earlier versions (fixed in Microsoft/TypeScript#40596), resulting in an exception being thrown during type checking. TypeScript 4.1 will be released at or around Nov 17th. We will either switch to it when it is released or we will switch to 4.0 if we have performed code changes that will make our code parseable without an exception again.~~

We are setting `allowJs` to `false` now, to avoid `tsc` tripping over type annotations in flow-typed files. Once we have migrated flow typings to TypeScript, we can enable it again. This also fixes the issue where `v4.0.5` was crashing due to a parsing issue, so we can go back it (instead of `v4.1.1-rc`).

## Motivation and Context

This is the first, preparatory step in migrating the flow-typed parts of the code base to TypeScript. Preparatory steps are meant to not interrupt the general development process or produce merge conflicts with other PRs.

## How Has This Been Tested?

A basic page component was written in TypeScript and a route for it was added. That route properly serves the page component. Afterwards, the commit containing that change was reverted. To test it again, the revert commit (7852910ab5c910716c12edd1d3a2d617a77f137f) can be reverted.

## Types of changes

## Checklist:


Gaya (approved):
Tested and works perfectly!

Issue #9315, opened 2020-10-30
Cannot load node metrics page under high load situations


edmundoa:
Under certain high load situations a node may not be fast enough to return its metrics. At the moment we don't handle that case graciously, so we end up loading an error page with this message:

```
Cannot read property 'length' of undefined

Stack Trace:

TypeError: Cannot read property 'length' of undefined
    at t.render (https://<redacted>/assets/aa9d5b41-102.53c5a65f985b27d783a7.js:1:1441)
    at Wi (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:69874)
    at Pi (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:69669)
    at Ms (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:105468)
    at uc (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:96715)
    at sc (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:96640)
    at Qs (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:93670)
    at https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:45312
    at exports.unstable_runWithPriority (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:42:3988)
    at Xo (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:45021)
    at Ko (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:45257)
    at Uo (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:45192)
    at Gs (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:90455)
    at Object.enqueueSetState (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:26:48977)
    at t.y.setState (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:34:1296)
    at a.n (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:18:2700)
    at a.emit (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:133:220231)
    at f.trigger (https://<redacted>/assets/vendor.8d6aa5835c8302c41e2f.js:18:3581)
[...]

PR #9421, opened 2020-11-10
Sync cloud with 3.3

edmundoa:
Pulls a fix for https://github.com/Graylog2/graylog2-server/issues/9315 into `cloud-3.3`.


edmundoa:
This is now stale and I think we don't need it any more, closing.

PR #9424, opened 2020-11-10
Fix condition for session extension

mpfz0r:
The session wasn't "touched" if the `X-Graylog-No-Session-Extension`
header was missing.
(And that's what the frontend is doing)

Bug introduced with #9287


bernd (approved):
Tested and works. :+1: And thanks @kmerz for finding it! :100:

PR #9426, opened 2020-11-10
Display and manage user account status in the frontend

### LINKED ISSUE #9077: Implement a user status

### LINKED ISSUE #9413: Allow modification and display user status in the web interface
## Description

With https://github.com/Graylog2/graylog2-server/pull/9374 we've implemented the backend code for the user status. Now we need to display the status in the user overview and on the details / edit page. The user overview and also the edit page should contain an action to change this status. This action should only be available for internal users. 

With these changes we should also remove the enable / disable buttons form the synced users overview on the authentication details page and the related actions, since they are no longer needed

kmerz:
## Motivation
If a user is on vacation a often used routine is to disable a user account. Right now the backend did support disable and enabling a user account but the frontend did not use it.

## Description
- Show user enable and disable status in users details page and useroverivew
- Add a enable / disable menu item to a user in the useroverview page
- Also: Add missing permission check in the backend resource

Fixes #9413

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -563,6 +563,7 @@ public class UsersResource extends RestResource {
 
         final User.AccountStatus newStatus = User.AccountStatus.valueOf(newStatusString.toUpperCase(Locale.US));
         final User user = loadUserById(userId);
+        checkPermission(RestPermissions.USERS_EDIT, user.getName());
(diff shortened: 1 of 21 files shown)


mpfz0r (approved):
Tested and LGTM

linuspahl (approved):
Looks good to me!

PR #9427, opened 2020-11-10
Do not create new component for plugin routes with parent on every AppRouter render

linuspahl:
## Description
When a plugin route has a parent component we wrap the component with its parent in AppRouter -> renderPluginRoute.
When creating the route we provided the component for the Route component by using the Route component prop.

Because of this combination plugin routes with a parent got newly created every time the AppRouter rendered.
This is not the case when using the render prop of the Route component. Please have a look at the react-router docs for more information.

You can test this change on a dashboard by changing the sidebar pinning (with a non system admin user). Before this change the whole page rerendered sometimes and the sidebar got closed.

## Types of changes


kmerz (approved):
LGTM :+1: 

PR #9434, opened 2020-11-10
Terminate active sessions for disabled users

bernd:
Subscribe to "UserChangedEvent" and terminate active sessions when a
user is deleted or disabled.


mpfz0r (approved):
Works and LGTM

PR #9437, opened 2020-11-11
Prevent disabling your own user account

kmerz:
## Motivation
Prior to this change, a logged in user was able to disable himself (if
he had the necessary permissions).

## Description
This change will hide the menu item `Enable`/`Disable` for the user
himself.

## Also
Add a confirmation dialog for disabling user which explains that
disabling a user also stops its current session.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -103,6 +103,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.Objects.requireNonNull;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.maxBy;
 import static org.graylog2.shared.security.RestPermissions.USERS_EDIT;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -556,11 +557,17 @@ public void changePassword(
...
     public Response updateAccountStatus(
-            @ApiParam(name = "userId", value = "The id of the user whose status to change.", required = true) @PathParam("userId") String userId,
+            @ApiParam(name = "userId", value = "The id of the user whose status to change.", required = true)
+            @PathParam("userId") @NotBlank String userId,
             @ApiParam(name = "newStatus", value = "The account status to be set", required = true,
-                    defaultValue = "enabled", allowableValues = "enabled,disabled,deleted")
+                      defaultValue = "enabled", allowableValues = "enabled,disabled,deleted")
             @PathParam("newStatus") @NotBlank String newStatusString) throws ValidationException {
 
+        final User currentUser = requireNonNull(getCurrentUser(), "currentUser cannot be null");
```
mpfz0r, line 566: We have `@Context UserContext userContext` for this now.

bernd: Hah, you are right! :smile: I will change it. Thank you!

graylog2-web-interface/src/components/users/UsersOverview/UserOverviewItem/ActionsCell.jsx
```diff
@@ -50,9 +52,17 @@ const ReadOnlyActions = ({ user }: { user: UserOverview }) => {
...
-    const newStatus = accountStatus === 'enabled' ? 'disabled' : 'enabled';
-    UsersDomain.setStatus(id, newStatus);
+    // eslint-disable-next-line no-alert
+    if (accountStatus === 'enabled' && window.confirm(`Do you really want to disable user ${fullName}? All current sessions will be terminated.`)) {
+      UsersDomain.setStatus(id, 'disabled');
+
+      return;
+    }
+
+    UsersDomain.setStatus(id, 'enabled');
```
linuspahl, line 65: When disabling a user and closing the confirm dialog, `UsersDomain.setStatus(id, 'enabled');` is being called and the API responds with `Error: Not Modified`

linuspahl (changes requested):

linuspahl (approved):
Looks good to me!

mpfz0r (approved):

PR #9423, opened 2020-11-10
Hide indices on cloud

Gaya:
Hides all pages and mentioning of indices for cloud consumer as they will not be needing those.

## Description

Hides the indices page, the entry in the systems menu, indices index failures on system overview.

## Motivation and Context

Because this feature will not be available for cloud users.

## How Has This Been Tested?

Using a local cloud environment.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #9439, opened 2020-11-11
Hide outputs page on cloud

Gaya:
Users on cloud shouldn't be able to setup outputs

## Description

Hides the system outputs page for cloud users

Please not that this includes #9423, so please wait until this is merged

## Motivation and Context

Users on cloud shouldn't be able to setup outputs

## How Has This Been Tested?

Checking output of pages and menu

## Types of changes

## Checklist:


Gaya:
Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/90

edmundoa:
I noticed that this PR only takes care of System Outputs (`/system/outputs`), but we should also remove Stream Outputs (`/streams/<stream_id>/outputs`), since that would also let users send messages out of Graylog Cloud. @Gaya do we want to also hide that in this PR or better create a new one for it?

Gaya:
I'll add it to this one. Thanks for noticing and letting me know

edmundoa:
I just realised that there's another thing we should hide in the Streams page. It is the Index set a Stream should use. Since it's only a form field and we have modified some things on that page here, I think we can also hide it here, @Gaya what do you think?
Screen Recording 2020-11-16 at 14 11 22

Gaya:
@edmundoa sure, it's going to "default" anyway so I guess that's enough for now. I'll hide the select

graylog2-web-interface/src/components/streams/StreamForm.jsx
```diff
@@ -84,14 +85,15 @@ class StreamForm extends React.Component {
     this.setState(change);
   };
 
-  state = this._getValuesFromProps(this.props);
+  _indexSetSelect = () => {
+    const { indexSetId } = this.state;
 
-  render() {
-    const { title, description, removeMatchesFromDefaultStream, indexSetId } = this.state;
+    if (!AppConfig.isCloud()) {
```
edmundoa, line 91: It should be the other way around, or? :)

```suggestion
    if (AppConfig.isCloud()) {
```

Gaya: Hah! I used that to test the opposite. Good catch!

edmundoa (approved):
Looks good to me :shipit:

Issue #9443, opened 2020-11-11
Make startpage restable for User


kmerz:
## Expected Behavior
A user must be able to reset the startpage. So he can reach the graylog default startpage (Search without streams).

## Current Behavior
We miss a reset button and a once choosen stream as startpage cannot be made undone.

## Possible Solution
- `4.1` Make a input field in the settings section of a user with a reset button
- `4.0` Add a reset button to the header of the User Pages

bernd:
The 4.0 fix got merged, moving this issue to milestone 4.1.0.

linuspahl:
Got closed with https://github.com/Graylog2/graylog2-server/pull/9448

PR #9448, opened 2020-11-11
Add reset startpage button to profile section of user edit 4.0

kmerz:
## Motivation
Prior to this change, there was no way to reset a startpage for a user.

## Description
This change will add the most minimal solution by adding a Reset
Startpage button to the header of the profile section on the user edit page.

## Also
Fix setting the Set as Startpage menuitem for streams.

Graylog - Edit User (3)

Refs #9443


linuspahl (approved):
Looks good to me.

PR #9453, opened 2020-11-11
Show missing menu items for Default Stream

### LINKED ISSUE #9433: It's not possible to configure outputs for the default message stream
Before 4.0 it was possible to configure outputs for the default message stream via one of the big buttons:

image

In 4.0 the "Manage Ouputs" link went into the "More Actions" menu. Since that one is disable for the default message stream, it's not possible anymore to configure an output from the stream page:

image

## Possible Solution

We should probably enable the "More Actions" drop-down menu for the default message stream and disable the options that are not allowed for that stream.

kmerz:
## Motivation
Prior to this change, we completly disabled the `More Actions` of the
`Default Stream` which prevented the user to `Manage Outputs` or to set
the `Default Stream` as Startpage.

## Description
This change will only disable the menu items which are not applicable
for the default stream.

Fixes #9433

### DIFF
--- a/graylog2-web-interface/src/components/streams/Stream.jsx
+++ b/graylog2-web-interface/src/components/streams/Stream.jsx
@@ -214,7 +214,7 @@ class Stream extends React.Component {
       ? <Icon name="cube" title="Created from content pack" /> : null);
 
     const streamControls = (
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Works as expected!

PR #9459, opened 2020-11-12
Verify session against http auth header

mpfz0r:
If the username in the HTTPHeaderAuth request changes,
we need to destroy the session.

Fixes https://github.com/Graylog2/graylog-plugin-auth-sso/issues/35


bernd (changes requested):
I am seeing the following issue when trying to use a proxy that sets the `Remote-User` header. It works fine without the PR changes:

Peek 2020-11-24 19-34

mpfz0r:
@bernd hmm, I just tested this again. and it works without issues for me. maybe we can look at this together?

bernd (approved):
Tested with a fresh build and it works for me now. :+1:

PR #9472, opened 2020-11-12
Update UPGRADING for 4.0

bernd:
Add more upgrading notes.

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -4,79 +4,172 @@ Upgrading to Graylog 4.0.x
 
 .. _upgrade-from-33-to-40:
 
-Deprecation of cluster stats endpoints
-======================================
+.. contents:: Overview
+   :depth: 3
+   :backlinks: top
 
-Starting with v4.0, the cluster stats endpoints are deprecated and will be removed in a future version. Those include:
+.. warning:: Please make sure to create a MongoDB database backup before starting the upgrade to Graylog 4.0!
 
-  - '/system/cluster/stats'
-  - '/system/cluster/stats/elasticsearch'
-  - '/system/cluster/stats/mongo'
+Breaking Changes
+================
 
-[BREAKING] Fixing certificate validation for LDAP servers used for authentication
-=================================================================================
+Changes to the Elasticsearch Support
+------------------------------------
+
+Starting with Graylog v4.0, bigger changes to the Elasticsearch versions supported are happening:
+
+  - Support for Elasticsearch versions prior to v6.8.0 is dropped.
+  - Support for Elasticsearch v7.x is now included.
+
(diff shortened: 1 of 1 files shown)


UPGRADING.rst
```diff
@@ -4,79 +4,156 @@ Upgrading to Graylog 4.0.x
...
+Changes to the Elasticsearch Support
+------------------------------------
+
+Starting with Graylog v4.0, bigger changes to the Elasticsearch versions supported are happening:
+
+  - Support for Elasticsearch versions prior to v6.8.0 are dropped.
+  - Support for Elasticsearch v7.x is now included.
+
+This means that you can upgrade to Graylog v4.0 without an Elasticsearch update only if you have been on at least Elasticsearch v6.8.0 before.
+Additionally, due to the fact that Elasticsearch supports only indices created by the last two major versions (i.e. ES6.8.0+ reads indices created by ES5 & ES6, ES7 reads indices created by ES6 & ES7), you can change to Graylog v4.0 with an Elasticsearch update without reindexing only if you have been on at least Elasticsearch v6.0.0 before.
```
mpfz0r, line 23:
```suggestion
Additionally, due to the fact that Elasticsearch supports only indices created by the last two major versions (i.e. ES6.8.0+ reads indices created by ES5 & ES6, while ES7 reads indices created by ES6 & ES7), you can change to Graylog v4.0 with an Elasticsearch update without reindexing only if you have been on at least Elasticsearch v6.0.0 before.
```
I find it easier to read that way

UPGRADING.rst
```diff
@@ -4,79 +4,156 @@ Upgrading to Graylog 4.0.x
...
+=========================
+
+The following API endpoints are deprecated beginning with 4.0.
+
+Deprecation of cluster stats endpoints
+--------------------------------------
+
+Starting with v4.0, the cluster stats endpoints are deprecated and will be removed in a future version. Those include:
+
+- ``/system/cluster/stats`
```
mpfz0r, line 93:
```suggestion
- ``/system/cluster/stats``
```
I guess that needs two backticks as well

mpfz0r (approved):

UPGRADING.rst
```diff
@@ -4,79 +4,156 @@ Upgrading to Graylog 4.0.x
...
-Starting with v4.0, the cluster stats endpoints are deprecated and will be removed in a future version. Those include:
+Breaking Changes
+================
+
+Changes to the Elasticsearch Support
+------------------------------------
+
+Starting with Graylog v4.0, bigger changes to the Elasticsearch versions supported are happening:
+
+  - Support for Elasticsearch versions prior to v6.8.0 are dropped.
```
dennisoelkers, line 19: I know I am correcting myself, but it should be `is dropped`, not `are dropped` :)

bernd: @dennisoelkers Thanks, I pushed a fix!

PR #9474, opened 2020-11-12
Use the correct permission for role assignment

mpfz0r:
Every user has user:edit permissions for himself!
This prevents users from assigning themselves to any role.


kmerz (approved):
LGTM :+1:

bernd (approved):

PR #9485, opened 2020-11-13
Finish support for multiple LDAP / AD servers

### LINKED ISSUE #1046: Support Multiple LDAP Server
Hi,

I would be nice if we can add more than one LDAP servers.  In case one LDAP server is not responding the interface will still available.

Thanks

mpfz0r:
- Add support to verify against multiple hosts in DefaultX509TrustManager

 - modify connection test to support and test multiple servers

 - remove validation for single server

Fixes #1046


bernd (changes requested):
Thanks for working on this! :+1: 

Right now we are using the `FailoverServerSet` for all configured LDAP servers. I think we should allow users to select a different strategy. The unbound-sdk offers some more `ServerSet` implementations which can be useful in certain scenarios.

mpfz0r:
I think this would be new feature and a separate PR.

bernd:
> I think this would be new feature and a separate PR.

Yeah, good point. I create a feature request for this: https://github.com/Graylog2/graylog2-server/issues/9491

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
```diff
@@ -191,14 +195,42 @@ public AuthServiceBackendDTO prepareConfigUpdate(AuthServiceBackendDTO existingB
...
+            return testSingleConnection(config, config.serverList().get(0));
+        }
+
+        // Test each server separately, so we can see the result for each
+        final List<AuthServiceBackendTestResult> testResults = config.serverList().stream().map(server -> testSingleConnection(config, server)).collect(Collectors.toList());
+
+        if (testResults.stream().anyMatch(res -> !res.isSuccess())) {
+            return AuthServiceBackendTestResult
+                    .createFailure("Test failure",
+                            //testResults.stream().map(AuthServiceBackendTestResult::errors).flatMap(Collection::stream).collect(Collectors.toList()));
```
bernd, line 210: Can we remove this comment?

graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackend.java
```diff
@@ -171,14 +175,42 @@ public AuthServiceBackendDTO prepareConfigUpdate(AuthServiceBackendDTO existingB
...
+            return testSingleConnection(config, config.serverList().get(0));
+        }
+
+        // Test each server separately, so we can see the result for each
+        final List<AuthServiceBackendTestResult> testResults = config.serverList().stream().map(server -> testSingleConnection(config, server)).collect(Collectors.toList());
+
+        if (testResults.stream().anyMatch(res -> !res.isSuccess())) {
+            return AuthServiceBackendTestResult
+                    .createFailure("Test failure",
+                            //testResults.stream().map(AuthServiceBackendTestResult::errors).flatMap(Collection::stream).collect(Collectors.toList()));
```
bernd, line 190: Can we remove this comment?

graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackendConfig.java
```diff
@@ -84,9 +84,6 @@
 
     @Override
     public void validate(ValidationResult result) {
-        if (servers().size() > 1) {
```
bernd, line 87: Same as for the AD config, how about checking if the list is not empty?

graylog2-server/src/main/java/org/graylog2/security/DefaultX509TrustManager.java
```diff
@@ -107,10 +119,12 @@ private void validateHostnames(X509Certificate[] x509Certificates, String s) thr
...
 
     private boolean certificateMatchesHostname(X509Certificate x509Certificate) {
-        try {
-            return this.authorizer.verify(this.host, x509Certificate);
-        } catch (IOException e) {
-            return false;
-        }
+        return this.hosts.stream().anyMatch(host -> {
+            try {
+                return this.authorizer.verify(host, x509Certificate);
```
bernd, line 124: Is it actually okay to just check if the certificate matches any host in our list? I was wondering if this can be an issue because we don't know which host is giving us the certificate. So a host could give us the certificate of another host from our list.

bernd (changes requested):
I tested the failover case by adding two LDAP servers to my database. (no UI support yet)

A few comments:

graylog2-server/src/main/java/org/graylog2/security/DefaultX509TrustManager.java
mpfz0r: That's difficult to implement. I've added a javadoc comment to point this out.
However, in our scenario, if I trust to talk to any of the servers in the list, it shouldn't matter which certificate they provide.

bernd: Thanks! :+1:

bernd (approved):

PR #9489, opened 2020-11-16
Move DEFAULT_PAGINATION to central point

kmerz:
## Motivation
Prior to this change, we defined DEFAULT_PAGINATION for every
list a new.

## Description
This change will define DEFAULT_PAGINATION on a central point for
pagination.

## Also
Fix use of this in URLUtils

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#793

### DIFF
--- a/graylog2-web-interface/src/stores/PaginationTypes.ts
+++ b/graylog2-web-interface/src/stores/PaginationTypes.ts
@@ -47,6 +47,13 @@ export type ListPagination = Pagination & {
   count: number,
 };
 
+export const DEFAULT_PAGINATION = {
+  page: 1,
+  perPage: 10,
+  query: '',
+  total: 0,
+};
+
 export type PaginatedList<ItemType> = {
   list: Immutable.List<ItemType>,
   pagination: ListPagination,
--- a/graylog2-web-interface/src/stores/PaginationTypes.ts
+++ b/graylog2-web-interface/src/stores/PaginationTypes.ts
@@ -49,7 +49,7 @@ export type ListPagination = Pagination & {
   count: number,
 };
 
-export const DEFAULT_PAGINATION = {
+export const DEFAULT_PAGINATION: Pagination = {
   page: 1,
   perPage: 10,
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/stores/PaginationTypes.js
```diff
@@ -31,6 +31,13 @@ export type ListPagination = Pagination & {
   count: number,
 };
 
+export const DEFAULT_PAGINATION = {
+  page: 1,
+  perPage: 10,
+  query: '',
+  total: 0,
+};
```
linuspahl, line 39: Do we need `total` here? The lists we've implemented recently, receive the total amount from the request response.
To ensure the `DEFAULT_PAGINATION` has the correct format, we could type it as `Pagination`.

graylog2-web-interface/src/stores/PaginationTypes.ts
```diff
@@ -28,26 +29,33 @@ export type PaginatedResponseType = {
...
+  query: string,
+  additionalQueries?: AdditionalQueries,
+};
 
 export type ListPagination = Pagination & {
   total: number,
   count: number,
 };
 
+export const DEFAULT_PAGINATION = {
```
linuspahl, line 53: Let's type this as `Pagination`.

linuspahl (changes requested):

linuspahl (approved):

graylog2-web-interface/src/stores/PaginationTypes.ts
```diff
@@ -19,6 +19,7 @@ import { $PropertyType } from 'utility-types';
 
 import type { AdditionalQueries } from 'util/PaginationURL';
 
+/* eslint-disable camelcase */
```
linuspahl, line 22: @kmerz Was about to merge the PR, but just saw this comment is no longer needed. It is a nitpick, but could you remove the comment, now where we disabled the rule and removed all similiar comments?

Issue #8984, opened 2020-09-17
Remove old pluggable authentication realm code


bernd:
Remove the now obsolete pluggable authentication realm code. In 4.0 we will have pluggable authentication services and the actual shiro realms will be hardcoded and in a specific order.

We need to check with @thll if this will work with the cloud okta plugin.

## Tasks

- [ ] Check with @thll if this works for the cloud okta plugin
- [ ] Remove binding related code in `Graylog2Module`
- [ ] Remove old UI to enable/disable and order authentication realms
- [ ] Remove old API resources to enable/disable and order authentication realms
- [ ] Remove `org.graylog2.security.AuthenticationConfig` from `cluster_config`
- [ ] Remove disabled authentication realms from `AuthenticatingRealmModule`
- [ ] Remove dynamic ordered realms from `AuthenticatingRealmModule` and also the code itself (and everything that's obsolete by then)
- [ ] Remove old LDAP code.
  - `org.graylog2.security.ldap.LdapConnector`
  - `LdapSettings` (+ implementation), `LdapSettingsService` (+ implementation)
  - `org.graylog2.shared.security.ldap.LdapEntry`
  - `org.graylog2.security.realm.LdapUserAuthenticator`
  - `org.graylog2.rest.resources.system.ldap.LdapResource`
  - `org.graylog2.periodical.LdapGroupMappingMigration`
- [ ] tbd (check for more tasks)

mchubby:
Hello @bernd @mpfz0r,
We currently use the REST API to regularly update the LDAP binding credentials. Will there be similar endpoints for the new authentication mechanism?

Secondly, I think there's also a mention of Ldap filtered messages in the log4j2.xml file you may want to remove.

Also, it is probably worth mentioning these upcoming changes in the docs for the 3.3 branch.

bernd:
@mchubby Yes, we will have replacement APIs in 4.0. Specifically the following ones:

![image](https://user-images.githubusercontent.com/461/99098611-86862000-25d9-11eb-9bbd-e590c3758490.png)

The removal of these APIs will be in the 4.0 upgrading notes.

mpfz0r:
> Secondly, I think there's also a mention of Ldap filtered messages in the log4j2.xml file you may want to remove.

Thanks for the hint :) I'll remove that.

PR #9490, opened 2020-11-16
Remove logging config for apache ldap client

mpfz0r:
We don't use the apache library anymore.

Refs #8984


bernd (approved):
Thank you, good catch!

PR #9500, opened 2020-11-16
Provide verbose failure support through new AuthServiceRealm

mpfz0r:
If the LDAP/AD authentication cannot be done due to a server
error, throw an `AuthenticationServiceUnavailableException`.
This gives the user (and the audit log) an indicator, that
it wasn't his fault that he couldn't log in.

Refs #8136


bernd (changes requested):
Can you fix the conflict? Thanks! :smile:

bernd (approved):

PR #9506, opened 2020-11-16
Extend license formatter for TypeScript

bernd:
Extend the license formatter configuration to handle TypeScript files.


dennisoelkers (changes requested):
Looks good, but please hold merging until we are done with the TS migration, as we are touching a lot of files. When this is done, we will make a run and add license headers to this PR.

bernd:
@dennisoelkers Alright, thanks for the review! :+1:

dennisoelkers (approved):
Rebased and did a test run, which added missing license headers properly. Looks good!

PR #9520, opened 2020-11-17
A few minor wording adjustments for the new Users pages

danotorrey:
The new Users and Teams feature is really great, and the UI is very well-designed. While testing, I noticed a few very minor nitpicky wording issues. Rather than opening an issue, I just created this PR.


graylog2-web-interface/src/components/users/UserCreate/EmailFormGroup.jsx
```diff
@@ -25,7 +25,7 @@ const EmailFormGroup = () => (
                    maxLength={254}
                    type="email"
                    required
-                   help="Give the contact email address." />
+                   help="Give the contact an email address." />
```
bernd, line 28: I think this was meant in a different way: "an email address to contact the user" "contact email".

I guess we should use something like "The user's email address." (same for the Full Name)

bernd (changes requested):
Thanks for the PR! One note:

graylog2-web-interface/src/components/users/UserCreate/EmailFormGroup.jsx
danotorrey: Ahh yes, good point. I've adjusted it. How does it look now in https://github.com/Graylog2/graylog2-server/pull/9520/commits/7a9cccf556d9e68f4ce06f7584d27104ed425fe0?

bernd (approved):
Thanks, LGTM :+1:

danotorrey:
@bernd I think we can safely ignore the linter error since it's unrelated to this change. After reviewing the linter error with @Gaya, we identified it requires some larger reworking, which we probably don't want to do in this PR and right now. :D

> React Hook useEffect has missing dependencies: 'getValue' and 'onChange'. Either include them or remove the dependency array. If 'onChange' changes too often, find the parent component that defines it and wrap that definition in useCallback. (react-hooks/exhaustive-deps)

I'll resolve the merge conflicts now.

linuspahl (approved):

PR #9543, opened 2020-11-18
DateRange Absolute Limits & Errors

kyleknighted:
## Description

Adding error message for incorrect Absolute time ranges
Disable previous dates if Admin has limit setup
Disable Apply button on errors

image


graylog2-web-interface/src/views/components/searchbar/date-time-picker/AbsoluteRangeField.jsx
```diff
@@ -153,22 +162,36 @@ const fieldUpdate = (value) => {
...
+};
+
+const AbsoluteRangeField = ({ disabled, limitDuration, from, currentTimerange, setDisableApply }: Props) => {
   const range = from ? 'from' : 'to';
   const hourIcon = useRef(TIME_ICON_MID);
+  const [timeRangeError, setTimeRangeError] = useState('');
+
+  useEffect(() => {
+    setTimeRangeError(_isDayAfter(currentTimerange));
+  }, [currentTimerange]);
```
linuspahl, line 180: Why did you choose this kind of validation over the Formik validation?

linuspahl (commented):
When I try to change the time range, by using the input, I get the following error:
`Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`
This is not the case with `feature/DateTimeRange`. Can you reproduce this behaviour?

linuspahl (approved):

PR #9551, opened 2020-11-19
Migrate index of components/common

kmerz:
During migrating `components/common` I missed out to convert `index.jsx` to `index.tsx`.

This is done now


graylog2-web-interface/src/components/common/ErrorAlert.tsx
```diff
@@ -33,8 +33,8 @@ const Container: StyledComponent<{margin: number}, ThemeInterface, HTMLElement>
 `);
 
 type Props = {
-  onClose: () => void,
-  children: JSX.Element | string,
+  onClose: (msg: string | undefined) => void,
```
edmundoa, line 36: I find the following a bit more readable, not sure if we have any conventions on one or the other way:

```
  onClose: (msg?: string) => void,
```

edmundoa (commented):
I think most changes look good. I only have a general suggestion/question.

From what I see in https://react-typescript-cheatsheet.netlify.app/docs/advanced/types_react_api/#typesreact, it seems that `React.ReactElement` is not meant to use as type for the `children` props:

```
`ReactNode` - anything that is renderable inside of JSX, this is NOT the same as what can be rendered by a component!
(...)
`ReactElement` - can be used if you want to pass it to cloneElement aka it's pretty rarely used
```

Also in the examples they compare some of the types one could use for types:
```
  children1: JSX.Element; // bad, doesnt account for arrays
  children2: JSX.Element | JSX.Element[]; // meh, doesn't accept strings
  children3: React.ReactChildren; // despite the name, not at all an appropriate type; it is a utility
  children4: React.ReactChild[]; // better
  children: React.ReactNode; // best, accepts everything
```

From that I would suggest using `React.ReactNode` as type for children props, since it would simplify definitions a bit. That is unless you are using `React.ReactElement` for a particular reason I'm not aware of.

graylog2-web-interface/src/components/common/ErrorAlert.tsx
kmerz: In this case it makes totally sense what you propose!

linuspahl (approved):
Looks good to me. There are some cases were I prefer `ReactNode` over `ReactElement`, e.g. in the `PageHeader` props, but as discussed, we can adjust these cases if TS complains when migrate files to TS which implement these components.

edmundoa (approved):
Looks good to me, as well 👍

PR #9576, opened 2020-11-23
Migrate user and roles pages to Typescript

kmerz:
Migrate user and roles related pages to Typescript


graylog2-web-interface/src/pages/UserEditPage.tsx
```diff
@@ -58,12 +58,19 @@ const UserEditPage = ({ params }: Props) => {
     UsersDomain.load(userId);
   }, [userId]);
 
+  // @ts-ignore We check if user is loaded with the optional
+  const fullName = loadedUser?.fullName ?? '';
```
dennisoelkers, line 62: IMO we can skip these, if we type the `useState` call for `loadedUser`.

dennisoelkers (changes requested):
Looks good, just one issue.

dennisoelkers (approved):
LGTM! :tada:

PR #9587, opened 2020-11-23
Sync cloud 3.3

Gaya:
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/routing/AppRouter.jsx
+++ b/graylog2-web-interface/src/routing/AppRouter.jsx
@@ -215,16 +215,21 @@ const AppRouter = () => {
 
                         <Route exact path={Routes.SYSTEM.GROKPATTERNS} component={GrokPatternsPage} />
 
-                        {!isCloud && <Route path={Routes.SYSTEM.INDEX_SETS.CREATE} component={IndexSetCreationPage} />}
                         {!isCloud && (
-                          <Route path={Routes.SYSTEM.INDEX_SETS.SHOW(':indexSetId')} component={IndexSetPage} />
+                          <Route exact path={Routes.SYSTEM.INDEX_SETS.CREATE} component={IndexSetCreationPage} />
                         )}
                         {!isCloud && (
-                          <Route path={Routes.SYSTEM.INDEX_SETS.CONFIGURATION(':indexSetId')}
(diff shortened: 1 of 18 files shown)


graylog2-web-interface/src/routing/AppRouter.jsx
```diff
@@ -211,13 +214,17 @@ const AppRouter = () => {
...
+                        {!isCloud && <Route path={Routes.SYSTEM.INDEX_SETS.CREATE} component={IndexSetCreationPage} />}
+                        {!isCloud && (
+                          <Route path={Routes.SYSTEM.INDEX_SETS.SHOW(':indexSetId')} component={IndexSetPage} />
+                        )}
+                        {!isCloud && (
+                          <Route path={Routes.SYSTEM.INDEX_SETS.CONFIGURATION(':indexSetId')}
+                                 component={IndexSetConfigurationPage} />
+                        )}
+                        {!isCloud && <Route path={Routes.SYSTEM.INDICES.LIST} component={IndicesPage} />}
+                        {!isCloud && <Route path={Routes.SYSTEM.INDICES.FAILURES} component={IndexerFailuresPage} />}
```
edmundoa, line 227: These routes are missing the `exact` prop they had in 4.0. I'm not sure if that was intended or something lost while cherry-picking the changes.

edmundoa (changes requested):
I noticed there was a change in the Streams page and the link to "Manage Outputs" now moved to a menu. The link does not work, but we should hide the entry as well. Other than that it seems to do what it should 👍 

Screenshot 2020-11-25 at 11 46 21

graylog2-web-interface/src/routing/AppRouter.jsx
Gaya: That's a good one! I'll add them back

Gaya:
> I noticed there was a change in the Streams page and the link to "Manage Outputs" now moved to a menu. The link does not work, but we should hide the entry as well.

Will change! Thanks for letting me know.

edmundoa:
@Gaya I just merged #9537 in case you want to put the change in here as well. I think that's the last pending PR that was based on `cloud-3.3`.

Gaya:
> @Gaya I just merged #9537 in case you want to put the change in here as well. I think that's the last pending PR that was based on `cloud-3.3`.

Awesome! @edmundoa I'll add that one in here as well. Thanks!

edmundoa (approved):
Looks good and works for me 👍

Issue #9607, opened 2020-11-25
Scratchpad text area has wrong background color in dark mode


linuspahl:
## Expected Behavior
The scratchpad text area background should be dark.

## Current Behavior
The scratchpad text area background and font color is currently white.
![image](https://user-images.githubusercontent.com/46300478/100208389-f618db00-2f08-11eb-9be8-217cab474e0c.png)

## Context
We've already created a fix for this problem with https://github.com/Graylog2/graylog2-server/pull/8886 but closed this PR to implement a solution for all `textarea`s.

## Possible Solution
We need to implement a fix for the sctrachpad text area or define the background and font color for all `textarea`s.

## Your Environment

* Graylog Version: 4.0

linuspahl:
Got fixed with #9608 and #9614 (4.0)

PR #9608, opened 2020-11-25
Unify basic `<textarea />`styles with bootstrap text area styles

linuspahl:
**This PR needs a backport for 4.0**

## Description
This PR unifies the basic `textarea` styles like font, background and border color with the bootstrap text area styles.
This fixes e.g. the scratchpad background color in dark mode (https://github.com/Graylog2/graylog2-server/issues/9607).

I thought about adding the `form-control` class to the scratchpad textarea, but this would make it harder to overwrite styles like the font color. Another benefit of this solution is that we do not have to remember this class when adding new `textarea`s.
We should also consider implementing the same change for the  `input` and `select` element selector. I did not include this change in this PR to reduce the amount of possible side effects, which becomes relevant when creating the backport for 4.0.

## Types of changes


Gaya (approved):
👏

PR #9610, opened 2020-11-25
Pulsar

mpfz0r:
TODOS

- [x] Switch configuration between Pulsar, Kafka and NOOP Journal implementation (@mpfz0r)
- [x] Refactor MessageQueue Interfaces? E.g. do we still need a `MessageQueueReader`?

- [x] Implement connection retry logic. Test failure scenarios. (@thll)
- [x] Try to unify Kafka journal offset and pulsar messageId in `Message` class
- [x] Try to reduce conversion steps for processing messages: We currently have many different wrappers around RawMessage objects: `{Pulsar,Kafka}MessageQueueEntry` and `Journal.JournalReadEntry`
- [ ] Decide on Pulsar queue names (tenant, namespace, topic). We could use both, the `cluster-id`-`node-id`. This gives us the possibility to consume topics from multiple nodes in the future. (@mpfz0r ) -> see #9806 
- [x] Make Pulsar URL configurable via server config file. (@mpfz0r )
- [x] Handle `LifeCycle` changes, since nodes can be asked to "pause" processing. (@thll)
- [x] Check if we need to register in `GracefulShutdownService` (@thll )
- [x] Check how to handle failures to acknowledge messages (is the client retrying automatically? do we block until successful?) (@thll) (_Apparently acks are always performed asynchronously without a possibility to block until completed. If acks don't make it to the broker, messages will be redelivered. If this becomes a problem we probably have to solve it by eliminating duplicates manually_)
- [x] Unify Kafka and Pulsar cases (`JournalReader` vs `PulsarMessageQueueReader`) (@thll )

- [x] Benchmark new implementation. 

- [x] Specify throttling support. (first we need to define a quota / limit for Pulsar) (@thll) -> see #9786
- [ ] Implement missing metrics.
      - Make sure to change hardcoded kafka metrics: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/system/processing/MongoDBProcessingStatusRecorderService.java#L161
- [ ] Write tests (if possible and meaningful)
- [ ] Decide on re-delivery policy on ack timeout / negative acks
- [ ] Authentication / Authorization for consuming / producing / managing (@mpfz0r)
- [ ] Evaluate full cluster on real hardware (@mpfz0r)
- [ ] Investigate Forwarder backpressure behaviour when quota is exceeded. _Can we detect if server is blocked on producer and not run into timeout/retries in forwarder?_
- [ ] Modularize throttling logic and implement simplistic version for pulsar. See #9786 for discussion.


mpfz0r:
closed as not planned

PR #9628, opened 2020-11-26
Remove no longer needed `FetchProvider` unmocking in error pages tests

linuspahl:
## Description
In some tests we unmocked the `FetchProvider`, to be able to import the `FetchError`. Both classes shared the same file. This was not the best solution, because it resulted in the following `console.error`s when running these tests. 
```
There was an error fetching a resource: getaddrinfo ENOTFOUND undefined. Additional information: Not available
```

We recently created a separate file for the `FetchError`, which means unmocking the `FetchProvider` is no longer necessary.

## Types of changes


dennisoelkers (approved):
Awesome!

PR #9634, opened 2020-11-26
Implement Highlighting Rule Form

kmerz:
## Description
Make it possible to create and edit a Highlighting rule via a creation form.

## Motivation and Context
To create more complex Highlighting rules we need a form for creating these rules.

Graylog (22)

## Types of changes

## Todo
- chaining conditions
- since the last matching rule wins it must be possible to bring rules into an order


dennisoelkers:
@kmerz: Good work so far! I prefer to show the dialog in a modal instead, as the inline form in the sidebar is too size-constrained imo.

graylog2-web-interface/src/views/components/sidebar/highlighting/HighlightForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+const numberConditionOptions = Object.entries(ConditionLabelMap).map(([value, label]) => ({ value, label }));
+const otherConditionOptions = Object.entries(StringConditionLabelMap).map(([value, label]) => ({ value, label }));
+
+const HighlightForm = ({ onClose, rule }: Props) => {
+  const fieldTypes = useContext(FieldTypesContext);
+  const fields = fieldTypes?.all
+    ? fieldTypes.all
+    : Immutable.List<FieldTypeMapping>();
+  const fieldOptions = fields.map(({ name }) => ({ value: name, label: name })).toArray();
```
dennisoelkers, line 55: We should sort the field names lexicographically using `defaultCompare` here.

dennisoelkers (changes requested):
Looks good so far! Added a couple of comments inline.

One thing I noticed: We cannot create highlights for series (e.g. `count()`) right now through the new highlighting rule form. Also, a highlighting rule configured for a series does show an empty field value in the edit form. We could probably fix this by enabling `allowCreate` on the field select dropdown.

dennisoelkers (approved):
Awesome work! :tada:

PR #9641, opened 2020-11-30
Use plugins from web modules file in dev build.

dennisoelkers:
## Description

Before this change, running `devServer` without `disable_plugins` made it use a glob (`../../graylog-plugin-*/**/webpack.config.js`) to find webpack configs of all plugins in the filesystem. This lead to issues when plugins are merely existing in the fs but are not supposed to be used. (e.g. from a previous `graylog-project` checkout)

After this change, `devServer` looks for a `web-modules.json` file (which is generated by `graylog-project` upon checkout) and uses the plugins referenced in there instead of the glob. If the file is not present, it will fall back to the old behavior of using a glob. If `web-modules.json` is regenerated while the dev server is running, it needs a restart to pick up the changes.

## Motivation and Context

This PR allows it to switch between different manifests using different
sets of plugins without issues.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/webpack.combined.config.js
+++ b/graylog2-web-interface/webpack.combined.config.js
@@ -14,29 +14,47 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
+const fs = require('fs');
 const glob = require('glob');
 const path = require('path');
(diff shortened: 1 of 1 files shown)


bernd (approved):
Works, thank you! :+1:

PR #9679, opened 2020-12-01
Let devServer proxy all "/api" requests to the server API

bernd:
I still had this in my tree from the time we disabled CORS by default a few months ago.

This uses an express proxy middleware to forward all `/api` requests to the server backend. (similar to the "proxy" feature in [webpack-dev-server](https://webpack.js.org/configuration/dev-server/#devserverproxy))

This could simplify the development setup because CORS doesn't need to be enabled anymore.

What do you @Graylog2/frontend folks think? (I am also fine with discarding it :wink:)

### DIFF
--- a/graylog2-web-interface/config.js
+++ b/graylog2-web-interface/config.js
@@ -15,7 +15,7 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 window.appConfig = {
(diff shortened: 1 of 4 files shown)


Gaya (approved):
I like this idea. Makes it all just a little bit less complicated.

PR #9691, opened 2020-12-03
Disabling `camelcase` rule globally, `no-undef` for TS files.

dennisoelkers:
## Motivation and Context

This PR aims to improve our ESLint rules by:

  - disabling the `camelcase` rule globally: we are using camelcase for backend requests/response, therefore need to disable it pretty often. At the same time all developers are aware of our casing convention and rule violations are usually not a problem.
  - disabling the `no-undef` rule for TS files: checking for undefined variables is performed by the type check. At the same time ESLint is not aware of certain globals (e.g. `JSX` which is introduced by ambient react typings)

This PR is also removing inline comments disabling/enabling these rules.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/packages/eslint-config-graylog/index.js
+++ b/graylog2-web-interface/packages/eslint-config-graylog/index.js
@@ -26,6 +26,7 @@ module.exports = {
       parser: '@typescript-eslint/parser',
       plugins: ['@typescript-eslint/eslint-plugin'],
       rules: {
+        'no-undef': 'off',
         'no-use-before-define': 'off',
         '@typescript-eslint/no-use-before-define': ['error'],
         'no-unused-vars': 'off',
--- a/graylog2-web-interface/packages/eslint-config-graylog/index.js
+++ b/graylog2-web-interface/packages/eslint-config-graylog/index.js
@@ -53,6 +53,7 @@ module.exports = {
   ],
   rules: {
     'arrow-body-style': 'off',
+    camelcase: 'off',
     'import/extensions': 'off',
     'import/no-extraneous-dependencies': 'off',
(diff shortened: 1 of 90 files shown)


Gaya:
Honestly not 100% sure about the `no-undef`. Globals are pretty easy to define in a declaration file. I like the `no-undef` as a safe guard or at least making sure that we thought about the types of global variables. As far as I am concerned we can type those as `any` for the time being and better them later.

About the JSX and not importing React: I have no solution for it yet. Going to try and see if I can get it to work in a Create React App project and stricter linting options. Last time my IDE was conflicted about it and didn't know what to do with itself. Going to test it out today.

Camel case: Agree with that one. I am sure we all agree it's only used in API communication.

dennisoelkers:
@Gaya: The thing is that `tsc` reports undefined variables in the same way as ESLint does. The change I did is disabling for `no-undef` only, so ESLint would still report undefined variables in `.js(x)`. If we want to keep `no-undef` though, we can add `JSX` to the globals list.

Gaya:
@dennisoelkers ah, didn't notice it was for `tsc`. I just checked in a CRA project and it seems to work here now out of the box. The only thing it changed was to add `"jsx": "react-jsx",` in my `compilerOptions` in `tsconfig.json`. Maybe this will resolve the `no-undef` there as well. It's worth the try I think. It also updated typescript to `4.1.2`, so I don't know it that's required.

Gaya:
Found out through trying that indeed the `eslint` rule for `no-undef` was disabled in CRA's `eslint`, but it did throw errors in TypeScript for undefined variables. So I am indeed fine with this behaviour!

kmerz (approved):
LGTM :+1:

PR #9692, opened 2020-12-03
  Add state to disable or enable submit form in DataAdapterForm

ousmaneo:
These changes add a new `isFormExternalyValidated`  state to  `DataAdapterForm` to allow children component to control (disable or enable) the form submit button.

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#617

## Motivation and Context

For the CSV File upload plugin, this allows the submit button to be disabled when the file is being uploaded.
## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
+++ b/graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
@@ -108,7 +108,7 @@ class DataAdapterForm extends React.Component {
       // when creating always initially auto-generate the adapter name,
       // this will be false if the user changed the adapter name manually
       generateAdapterName: create,
-      isFormExternalyValidated: true,
+      isFormDisabled: true,
       dataAdapter: {
         id: adapter.id,
         title: adapter.title,
@@ -129,8 +129,8 @@ class DataAdapterForm extends React.Component {
     }
   };
 
-  _setExternalValidationState = (formValidationState) => {
-    this.setState({ isFormExternalyValidated: formValidationState });
+  _setIsFormDisabled = (formValidationState) => {
+    this.setState({ isFormDisabled: formValidationState });
   };
 
   _validate = (adapter) => {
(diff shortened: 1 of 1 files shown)


edmundoa (changes requested):
The change looks fine to me but I think the naming could be improved. I think the prop `isFormExternalyValidated` and even more `setIsFormExternalyValid` are a bit confusing and do not represent what we want to accomplish with the change, which is letting a children component disable form submission. How about renaming props to `isFormDisabled` and `setIsFormDisabled` respectively?

ousmaneo:
Thank you for reviewing @edmundoa. I have updated the naming.

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
```diff
@@ -357,7 +363,7 @@ class DataAdapterForm extends React.Component {
               <fieldset>
                 <Row>
                   <Col mdOffset={3} md={9}>
-                    <Button type="submit" bsStyle="success">{create ? 'Create Adapter'
+                    <Button type="submit" bsStyle="success" disabled={!isFormDisabled}>{create ? 'Create Adapter'
```
Gaya, line 366: I like the `isFormDisabled` name better as @edmundoa suggested, but right here it's telling the opposite is true. If the form is not not disable, it's disabled. I think the default and this check need to be flipped.

Other than that: looks good to me!

ousmaneo: Thank you ! fixed it.

Gaya (approved):
LGTM!

graylog2-web-interface/src/components/lookup-tables/DataAdapterForm.jsx
```diff
@@ -128,6 +129,10 @@ class DataAdapterForm extends React.Component {
     }
   };
 
+  _setIsFormDisabled = (formValidationState) => {
```
edmundoa, line 132: Small nitpick: we still use `formValidationState` as argument here.

edmundoa (approved):
Looks good to me. I added a small nitpick as comment, just in case you want to take care of it. Otherwise, feel free to merge 👍

PR #9700, opened 2020-12-03
Use MessageQueueAcknowledger instead of Journal

mpfz0r:
Log exception instead of throwing it in `acknowledge()`
We don't know how to recover from it anyhow.
So instead of handling that exception everywhere, keep it simple.


thll (approved):
LGTM

We should investigate how the pulsar client behaves when acknowledging fails due to network anomalies or the pulsar cluster being unreachable. I've added this as a TODO to our list.

PR #9701, opened 2020-12-03
Support for better integration of Token creation from Forwarder wizard

edmundoa:
- Extract create token form into a component that we can reuse
- Update Token type definition
- Remove default `expanded` value from `Panel`, since that threw an error in the console log: https://github.com/react-bootstrap/react-bootstrap/blob/v0.33.1/src/Panel.js#L99-L101

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#620


Gaya (approved):
LGMT

ousmaneo (approved):
Tested and works!

PR #9703, opened 2020-12-04
Make messagequeue implementation configurable

mpfz0r:
- Clean up and unify guice modules

- Add a new config option `message_journal_mode = kafka,pulsar`


graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
```diff
@@ -142,6 +145,11 @@ public boolean isMessageJournalEnabled() {
         return messageJournalEnabled;
     }
 
+    public String getMessageJournalMode() {
+        // TODO use enum?
```
thll, line 149: I think that's a good idea.

thll (changes requested):
That works well 👍 

However I have the feeling that the fact that we are treating the Kafka journal differently creates a bit of confusion, at least for me.

Things I find confusing (amongst others):
* We don't have a `KafkaMessageQueueReader` implementation because we still have the `JournalReader`.
* The `Noop...` classes throw exceptions in some cases instead of being really `No-Op`
* The `NoopMessageQueueWriter` is never used because the `InputBufferImpl` will not call it when journalling is disabled.

Maybe there's a way to reduce a bit of complexity. I'll add a TODO to our ever growing list 😀 

The only thing I think we should address right away is making the journal modes an 'enum' as you suggested. I think it will catch misconfiguration a bit earlier instead of throwing a bunch of guice errors.

thll (approved):

PR #9705, opened 2020-12-04
Typing `styled-components`

linuspahl:
## Description
This PR is typing `styled-components`. Initially we installed `@types/styled-components`, but we had to copy the type definitions to adjust the typing of `withTheme`. This is necessary because the type inference did not worked as expected in combination with `defaultProps` and `propTypes`. We can revert this once the changes are part of `@types/styled-components`.

Overview of changes:
- Using the `ThemeInterface` as `styled-component`s `DefaultTheme`
- Removing no longer needed usage of `StyledComponent` type.
- Improving component props based on improved type inference.

Please note, there are some cases where we type the theme prop as `DefaultTheme`, when using `styled`, because it can't be inferred. This will no longer be necessary when all components are written in TypeScript.

I recommend you to read https://styled-components.com/docs/api#typescript.

## Types of changes


graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
```diff
@@ -168,10 +164,8 @@ class SeriesSelect extends React.Component<Props, State> {
               options={options}
               value={series.map(_wrapOption)}
               components={_components}
-              onClose={this._onClose}
-              closeMenuOnSelect={false}
+              closeMenuOnSelect
```
dennisoelkers, line 167: This is breaking the current behavior of the series select, which allows the user to select function and field without having to reopen the select.

graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
```diff
@@ -130,10 +130,6 @@ class SeriesSelect extends React.Component<Props, State> {
     this.setState({ options: suggester.defaults });
   };
 
-  _onClose = () => {
```
dennisoelkers, line 133: Removing this is breaking the behavior which resets the state to selecting a function when closing the select.

graylog2-web-interface/src/components/permissions/GranteesSelector.tsx
```diff
@@ -129,6 +128,7 @@ const GranteesSelector = ({ availableGrantees, availableCapabilities, className,
                                     onChange={(granteeId) => onChange({ target: { value: granteeId, name } })}
                                     optionRenderer={_renderGranteesSelectOption}
                                     options={granteesOptions}
+                                    // @ts-ignore
```
dennisoelkers, line 131: Is there a different way to solve this?

dennisoelkers (changes requested):
Looks good so far, added a few comments inline.

graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
linuspahl: Good catch, that change slipped in.

graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
linuspahl: I removed this part because the `react-select` component does not have a prop `onClose`. But I understand the benefit and will reimplement this function and use the prop `onMenuClose` instead.

graylog2-web-interface/src/components/permissions/GranteesSelector.tsx
linuspahl: With https://github.com/Graylog2/graylog2-server/pull/9713 I changed this part and this `@ts-ignore` should no longer be necessary. I will remove it one of these pull requests.

graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
```diff
@@ -168,10 +168,9 @@ class SeriesSelect extends React.Component<Props, State> {
               options={options}
               value={series.map(_wrapOption)}
               components={_components}
-              onClose={this._onClose}
               closeMenuOnSelect={false}
+              onMenuClose={this._onClose}
               menuShouldScrollIntoView
-              tabSelectsValue
```
dennisoelkers, line 174: Is this removed on purpose?

dennisoelkers (changes requested):

graylog2-web-interface/src/views/components/aggregationbuilder/SeriesSelect.tsx
linuspahl: I removed this prop, because we always set it to `false` in the views `Select`. I tested this case with `tabSelectsValue` and prefer to keep this disabled.

dennisoelkers (approved):
LGTM!

PR #9711, opened 2020-12-07
Focus widget

kmerz:
## Motivation and Context
The user wants to set a single widget in focus by removing all other widgets temporarily and see the focused widget in full screen.
 
## Description
- Add a new CurrentViewState[Store/Actions] which will store the widget in focus
- Render the focused widget instead of the WidgetGrid
- Unset focused Widget before going into big display mode
- Adjust css styling

fullscreen-widget

## Types of changes


graylog2-web-interface/src/views/stores/CurrentViewStateStore.ts
```diff
@@ -47,11 +48,13 @@ export const CurrentViewStateActions: CurrentViewStateActionsType = singletonAct
     titles: { asyncResult: true },
     widgets: { asyncResult: true },
     widgetPositions: { asyncResult: true },
+    focusWidget: { asyncResult: true },
   }),
 );
 
 type CurrentViewStateStoreState = {
   activeQuery: string;
+  focusedWidget: string | undefined | null,
```
dennisoelkers, line 57: Do we need to put this in the `CurrentViewStateStore`? I would suggest creating a separate context for this instead, which provides the current state and the actions to modify it.

dennisoelkers (changes requested):
Works! Unfortunately it creates a scroll bar:

image

graylog2-web-interface/src/views/stores/CurrentViewStateStore.ts
linuspahl: @kmerz We could use the `SearchPageLayoutContext` for this.

linuspahl (changes requested):
Looks good, some thoughts and findings:

- When I open a dashboard I sometimes receive the error `Cannot read property 'id' of undefined` (`Query` - `_renderWidgetGrid`)
- We should make clear that the focus is only a temporary state. I can imagine that a user will not directly understand that this action behaves differently than e.g. the `Stretch width` action.
image
We could show a label (when focused) for the new action. In this case I would move the `Focus / Unfocus widget` before the dropdown icon (in both states). We should also hide (or disable) the `Stretch widget` action when the focus is active.

graylog2-web-interface/src/views/components/SearchResult.tsx
```diff
@@ -56,6 +67,7 @@ type Props = {
 
 const SearchResult = React.memo(({ queryId, searches, viewState }: Props) => {
   const fieldTypes = useContext(FieldTypesContext);
+  const focusedWidget = useContext(WidgetFocusContext);
```
linuspahl, line 70: We need to use ` const { focusedWidget } = useContext(WidgetFocusContext);` here as well, right now `focusedWidget`  is always true.

linuspahl (commented):
As discussed we do not have to adjust the widget actions in the "is focused" view right now, but we should disable actions like stretch and duplicate.

linuspahl (approved):
Looks like the duplicate action is still enabled. I am still approving the PR so my change request will not block the merge next week.

graylog2-web-interface/src/views/components/Query.test.jsx
```diff
@@ -57,14 +64,14 @@ describe('Query', () => {
...
-      <Query results={results}
-             widgetMapping={widgetMapping}
-             widgets={widgets}
-             onToggleMessages={() => {}}
-             queryId="someQueryId"
-             showMessages
-             allFields={Immutable.List()}
-             fields={Immutable.List()} />
+      <SUT results={results}
+           widgetMapping={widgetMapping}
```
dennisoelkers, line 68: Can we move all of the props which are constantly reused to the `SUT` definition? This way we can separate the boilerplate and override in tests what's actually different, which makes it way easier to understand.

graylog2-web-interface/src/views/components/Query.tsx
```diff
@@ -81,6 +87,27 @@ const _renderWidgetGrid = (widgetDefs, widgetMapping, results, positions, queryI
     }
   });
 
+  if (focusedWidget) {
```
dennisoelkers, line 90: I think we should move this to a separate function itself (`renderFocussedWidget`?) and do a conditional in the `Query` function, e.g. `return focussedWidget ? renderFocussedWidget(...) : renderWidgetGrid(...);` - this makes it easy to understand.

dennisoelkers (changes requested):
LGTM, just a few remarks inline.

dennisoelkers (changes requested):
Looks better already, two smaller comments inline.

dennisoelkers (approved):
Works fine, great work! :tada:

PR #9717, opened 2020-12-08
Align message queue reader classes

thll:
* The `JournalReader` becomes `KafkaMessageQueueReader`.
* The Pulsar reader will now also be stopped during graceful shutdown


mpfz0r (approved):
LGTM and works

PR #9721, opened 2020-12-08
Improving `connect` type inference regarding component `propTypes` and `defaultProps`.

linuspahl:
## Description
This PR improves the type inference when using `connect` for a component which also uses `defaultProps`.  Previously we had a problem in the following case:

Component has a prop `exampleProp`. This prop is defined as required in the component `Prop` type, but has a default value. Since it has a default value, the `PropType`s are defining the prop as optional. When we use this component the `exampleProp` is required. With this PR the `exampleProp` is typed as optional.

There is still one case were we do not infer the props perfectly. If the mentioned `exampleProp` is an object and the TypeScript type defines its values as required, we can't define a `PropType` definition which defines as the prop optional, but the attributes are required. With the current inference all attributes are typed as optional.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
+++ b/graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
@@ -55,13 +55,16 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
     newUrlEntry: PropTypes.string,
     onUpdate: PropTypes.func,
     configuration: PropTypes.object,
+    currentUser: PropTypes.exact({
+      permissions: PropTypes.arrayOf(PropTypes.string),
+    }).isRequired,
+    urlType: PropTypes.oneOf(['regex', 'literal', ''] as const),
   };
 
   static defaultProps = {
     newUrlEntry: '',
     onUpdate: () => { return undefined; },
     configuration: {},
-    // eslint-disable-next-line react/default-props-match-prop-types
     urlType: '' as const,
   }
 
--- a/graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
+++ b/graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
@@ -42,7 +42,7 @@ type Props = {
   newUrlEntry: string,
   onUpdate: () => void,
   configuration: {},
-  currentUser: {permissions: Array<string>},
+  currentUser: { permissions: Array<string> },
   urlType: 'regex' | 'literal' | '',
 };
 
@@ -55,15 +55,15 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
     newUrlEntry: PropTypes.string,
     onUpdate: PropTypes.func,
     configuration: PropTypes.object,
-    currentUser: PropTypes.object.isRequired,
-    urlType: PropTypes.string,
+
   };
 
   static defaultProps = {
     newUrlEntry: '',
     onUpdate: () => { return undefined; },
     configuration: {},
-    urlType: '',
+    // eslint-disable-next-line react/default-props-match-prop-types
+    urlType: '' as const,
   }
 
   constructor(props) {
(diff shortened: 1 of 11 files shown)


graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
```diff
@@ -55,15 +55,14 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
     newUrlEntry: PropTypes.string,
     onUpdate: PropTypes.func,
     configuration: PropTypes.object,
-    currentUser: PropTypes.object.isRequired,
-    urlType: PropTypes.string,
```
dennisoelkers, line 59: We can do this instead:

```
    urlType: PropTypes.oneOf(['regex', 'literal', ''] as const),
```

graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
```diff
@@ -55,15 +55,14 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
     newUrlEntry: PropTypes.string,
     onUpdate: PropTypes.func,
     configuration: PropTypes.object,
-    currentUser: PropTypes.object.isRequired,
```
dennisoelkers, line 58: We can do this instead:

```
    currentUser: PropTypes.exact({
      permissions: PropTypes.arrayOf(PropTypes.string),
    }).isRequired,
```

graylog2-web-interface/src/components/common/URLWhiteListFormModal.tsx
```diff
@@ -55,15 +55,14 @@ class URLWhiteListFormModal extends React.Component<Props, State> {
...
-    urlType: PropTypes.string,
   };
 
   static defaultProps = {
     newUrlEntry: '',
     onUpdate: () => { return undefined; },
     configuration: {},
-    urlType: '',
+    // eslint-disable-next-line react/default-props-match-prop-types
```
dennisoelkers, line 64: If we use the suggestion above, we can remove this.

graylog2-web-interface/src/views/components/widgets/Widget.tsx
```diff
@@ -129,14 +129,6 @@ const _editComponentForType = (type) => {
 class Widget extends React.Component<Props, State> {
   static propTypes = {
     id: PropTypes.string.isRequired,
-    view: PropTypes.object.isRequired,
```
dennisoelkers, line 132: I think we might be able to use CustomPropTypes.CurrentView` here.

graylog2-web-interface/src/views/components/widgets/Widget.tsx
```diff
@@ -129,14 +129,6 @@ const _editComponentForType = (type) => {
...
-    widget: PropTypes.shape({
```
dennisoelkers, line 133: Can we use `PropTypes.instanceOf(WidgetModel)` here?

graylog2-web-interface/src/views/components/widgets/Widget.tsx
```diff
@@ -146,7 +138,6 @@ class Widget extends React.Component<Props, State> {
     onSizeChange: PropTypes.func.isRequired,
     onPositionsChange: PropTypes.func.isRequired,
     title: PropTypes.string.isRequired,
-    position: PropTypes.object.isRequired,
```
dennisoelkers, line 149: Can we use `PropTypes.instanceOf(WidgetPosition)` here?

dennisoelkers (changes requested):
Looks good, some nice changes in there! Added some remarks/questions inline.

linuspahl:
@dennisoelkers Thanks for the good feedback! I wrongly assumed `instanceOf` is not working as expected in the mentioned cases, due to a failing test. But I just had to update the test properly.

dennisoelkers (approved):
Good work!

PR #5082, opened 2018-09-07
Return HTTP 400 when creating/updating a user with invalid role names

radykal-com:
## Description
Check all passed roles exist in the system prior to trying to assign them to the user

## Motivation and Context
Fixes #4665 

## How Has This Been Tested?
Try to create or update users through the API, current behaviour returns 500 without any message.
New behaviour returns 400 with a descriptive message about the roles

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #9722, opened 2020-12-08
Return HTTP 400 when creating/updating a user with invalid role names

mpfz0r:
Port of https://github.com/Graylog2/graylog2-server/pull/5082
with an additional fix.

Thank you very much @radykal-com


mpfz0r:
@bernd gonna need your review. can't do it for my own PR :-) But the original one looks fine for me, except for the lowercase problem

bernd (approved):

PR #9729, opened 2020-12-09
Hide nodes page and disable navigation on cloud

Gaya:
## Description

Will disable nodes page and components linking to it.

## Motivation and Context

On cloud we don't want to show node details just yet, so we need to disable them for now.

## How Has This Been Tested?

By running through the code finding mentions of the nodes page and disabling them.

## Types of changes

## Checklist:

Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/97

### DIFF
--- a/graylog2-web-interface/src/components/common/LinkToNode.jsx
+++ b/graylog2-web-interface/src/components/common/LinkToNode.jsx
@@ -22,6 +22,7 @@ import Reflux from 'reflux';
 import { Link } from 'components/graylog/router';
 import StoreProvider from 'injection/StoreProvider';
 import Routes from 'routing/Routes';
+import AppConfig from 'util/AppConfig';
 
 import Icon from './Icon';
 import Spinner from './Spinner';
@@ -58,12 +59,20 @@ const LinkToNode = createReactClass({
       const iconClass = node.is_master ? 'master-node' : '';
       const iconTitle = node.is_master ? 'This is the master node in the cluster' : '';
 
-      return (
-        <Link to={Routes.SYSTEM.NODES.SHOW(this.props.nodeId)}>
+      const content = (
+        <>
           <Icon name={iconName} className={iconClass} title={iconTitle} />
           {' '}
           {node.short_node_id} / {node.hostname}
-        </Link>
+        </>
+      );
+
+      if (AppConfig.isCloud()) {
+        return content;
+      }
+
+      return (
+        <Link to={Routes.SYSTEM.NODES.SHOW(this.props.nodeId)}>{content}</Link>
       );
     }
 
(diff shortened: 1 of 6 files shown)


edmundoa (changes requested):
This works fine except that we are still missing a couple links to the single node page:

- [`NodeName`](https://github.com/Graylog2/graylog2-server/blob/29aa0ab4bd8dca7ef60ab7b5f1b17b805d9e7027/graylog2-web-interface/src/views/components/messagelist/NodeName.jsx#L42) has a link to that page, which is used in the search results after expanding a message
- [`search/MessageDetail`](https://github.com/Graylog2/graylog2-server/blob/29aa0ab4bd8dca7ef60ab7b5f1b17b805d9e7027/graylog2-web-interface/src/components/search/MessageDetail.jsx#L58) also has a link _in theory_. Looking at the usages we never supply the list of Nodes, so I think that part never renders. Still, we should make sure to hide the link, in case I'm wrong. The component is used in the message loader in the Stream Rules management page, and the Pipeline simulator

Gaya:
@edmundoa I'll take a look at those two instances. I think I can disable it quiet easily on those too. Looks very similar too 😅

Gaya:
@edmundoa I was looking for where `<NodeName />` is used, and it's only here:
https://github.com/Graylog2/graylog2-server/blob/cloud-4.0/graylog2-web-interface/src/views/components/messagelist/MessageDetail.jsx#L92-L103

When on cloud the other display is used (received by Forwarder / id)

Created a fallback for `MessageDetails.jsx`

edmundoa:
@Gaya I guess I'm seeing the link with `NodeName` because I have some standard inputs running in parallel to the forwarder inputs. I think we don't want to support that (Inputs page is hidden), so we can forget about that for now. Thank you for checking!

edmundoa (changes requested):
There are some broken tests in Jenkins, other than that I think we are good to go.

Gaya:
@edmundoa ah good one! Fixed the test. Should be green now.

edmundoa (approved):
It's looking good now, thank you! :shipit:

edmundoa:
I'm afraid this is logging a warning in the browser console that I somehow missed:

```
Warning: React does not recognize the `activeKey` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `activekey` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (created by Navigation)
    in ul (created by Nav)
    in Nav (created by Nav)
    in Nav (created by Navigation)
    in div (created by NavbarCollapse)
    in Transition (created by Collapse)
    in Collapse (created by NavbarCollapse)
    in NavbarCollapse (created by Navigation)
```

PR #9730, opened 2020-12-09
Server changes for custom user management

danotorrey:
## Description

Server changes required for https://github.com/Graylog2/graylog-plugin-cloud/pull/626

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#626


danotorrey:
Re-running the build to see if it resolves the error.

thll (commented):
I did a smoke test to check if everything still works with the default implementation when no custom implementation is provided. I didn't come across any issues 👍

pom.xml
```diff
@@ -137,6 +137,7 @@
         <netty.version>4.1.52.Final</netty.version>
         <netty-tcnative-boringssl-static.version>2.0.34.Final</netty-tcnative-boringssl-static.version>
         <okhttp.version>3.14.6</okhttp.version>
+        <okta-sdk.version>3.0.1</okta-sdk.version>
```
mpfz0r, line 140: while this can't create any problems, I was wondering if there isn't a better place for this?

mpfz0r (commented):
Mostly LGTM. I've also successfully cherry picked your changes into current non-cloud master.
No issues there :+1:

pom.xml
thll: @mpfz0r what do you have in mind as a better place? If I'm correct, the goal here is to resolve the dependency of two different plugins to the same version. As the Okta SDK is not used in Graylog core but only in plugins, it might make indeed sense to move it somewhere else. I see two options:
1. Add the version property definition to the plugin parent pom [here](https://github.com/Graylog2/graylog2-server/blob/092d6f7e53963b26ccfad43ad60b3afc3fdbfc6e/graylog-plugin-parent/pom.xml#L38) instead
2. Or manage the dependencies explicitly in the plugin parent pom [here](https://github.com/Graylog2/graylog2-server/blob/092d6f7e53963b26ccfad43ad60b3afc3fdbfc6e/graylog-plugin-parent/pom.xml#L50) and remove the version specifiers from dependencies in the plugin poms.

What do you think?

danotorrey: Thanks for the feedback @mpfz0r. I am also curious if you have in mind any other particular place that might work better.

As far as I know, we currently only define common dependency versions in [graylog2-server/pom.xml](https://github.com/Graylog2/graylog2-server/blob/maser/pom.xml). I agree that, since this dependency is just used in plugins it perhaps _could_ be moved to the plugin-parent. But, we've not done this before with other plugins, so perhaps we need some feedback/agreement from others (eg. Bernd) if we really want start doing this.

graylog2-server/src/main/java/org/graylog/security/authservice/UserDetails.java
```diff
@@ -43,6 +43,8 @@
 
     public abstract String fullName();
 
+    public abstract boolean isExternal();
```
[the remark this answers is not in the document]
danotorrey: Thanks for the feedback @mpfz0r. Here's some of the thinking behind adding this property here, incase that helps the discussion:

The current definition for external users (as I understand) is that they exist in another system of record and are not permitted to be edited in Graylog. Okta users do not fall into this category, since they can be edited in Graylog. The intention of adding the `isExternal` flag to UserDetails is to allow the `ProvisionerService` to provision users with this flag set to false (the Okta case). The `UserDetails` object is just a means of sending user data into the ProvisionerService.

I did not yet have any plan to add any properties to the `UserImpl` persisted DB object, but happy to look at that further if you think we should identify the Okta users somehow.

pom.xml
mpfz0r: > @mpfz0r what do you have in mind as a better place?

None ;-) I'm always a bit lost when it comes to maven and our plethora of pom files.
But it kind of felt wrong putting this into the server...
IMHO the plugin parent pom sounds better. But I'll leave that up to you or @bernd

danotorrey: @bernd Do you have an opinion on this?

danotorrey: I will likely revert this, since we now cannot use the Okta Java SDK with the new Owner Resource Password authentication flow that we are using now to achieve the isolation between Cloud Okta users.

danotorrey: The Okta User Management feature no longer uses the Okta SDK directly. However, the Enterprise Integrations Plugin uses it. I don't see an issue with leaving this here. This might be helpful if the Okta SDK is needed in another plugin in the future (since version conflicts are known to occur with it).

danotorrey: Update: The `okta-sdk.version` has been removed, since we don't need it in this PR: https://github.com/Graylog2/graylog2-server/pull/9730/commits/88ab01b8cf85c43ae799e735a92748486521a176

danotorrey:
See [comment](https://github.com/Graylog2/graylog-plugin-cloud/pull/626#issuecomment-770720485) in Cloud User Management PR. I think once checks pass, this is ready for final review.

thll (approved):
The only thing changed since my last review is the removal of the Okta SDK property from `pom.xml`. This is fine ✅ .

Other than that it's only frontend changes and I can't judge those. But I trust they have been properly reviewed already.

PR #9732, opened 2020-12-09
Improvements in Token list

edmundoa:
This PR adds the following changes to the Token list:
- Replace [`TableList`](https://graylog2.github.io/frontend-documentation/#!/TableList) with [`ControlledTableList`](https://graylog2.github.io/frontend-documentation/#!/ControlledTableList), which gives us more flexibility to customize how to display tokens
- Add last access information to each Token, helping users to know which Tokens may not be used
- Display Token information after creation, making it easier to copy it

Screenshot 2020-12-09 at 12 13 46

Refs: https://github.com/Graylog2/graylog-plugin-cloud/issues/570

### DIFF
--- a/graylog2-web-interface/src/components/users/TokenList.jsx
+++ b/graylog2-web-interface/src/components/users/TokenList.jsx
@@ -16,14 +16,18 @@
  */
 // @flow strict
 import PropTypes from 'prop-types';
-import React, { useState } from 'react';
-import { List } from 'immutable';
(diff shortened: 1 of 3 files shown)


Gaya (approved):
Great idea! I personally always use `(a, b) =>` on sort, but `token1` and `token2` is fine too by me.

ousmaneo (approved):
Tested and Legit!

PR #9746, opened 2020-12-09
Only display tokens immediately after creation

### LINKED ISSUE #9124: As an Admin User Access Tokens  of other users can be viewed in Plain
Dear Graylog Team,

we happiliy read the news about Graylog 3.3 with encrypted tokens https://docs.graylog.org/en/3.3/pages/upgrade/graylog-3.3.html#api-access-token-encryption . Our expection of this news, was that token are now handled similar to passwords. Unfortunely as an Admin one can see tokens of other users in Plaintext. 

## Expected Behavior
Tokens are hashed and salted in MongoDB and can not be seen anywhere in plain after creation.

## Current Behavior
Tokens are visible in Plaintext although they give the same access as the users password. 

## Possible Solution
Handle tokens as passwords and give only during creation of token the possiblity to read the token in plaintext. 

Ideally a User gets an E-Mail if someone creates a token in his usercontext, so he can control if the creation is wanted, additionally it would be a nice feature if tokens could be bound only to specific functions in a usercontext e.g. execution of a specific search. 

1. Assign/create Token to/as  User
2. Admin and User can see the token in Plaintext
3. 
4.

## Context
Tokens should be treated as passwords and should have the same security regulations (salt and hash)

## Your Environment

* Graylog Version: 3.3
* Java Version: %
* Elasticsearch Version: %
* MongoDB Version: % 
* Operating System: %
* Browser version: %

edmundoa:
Remove tokens from the list tokens resource and the token list page. This means that users will only be able to see (and copy) tokens when they are created. If a token is lost, the user will need to create another one.

Please note that this only fixes the issue for cloud. We will need to port it to other branches along with other cloud-specific changes.

Fixes #9124

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/users/responses/TokenSummary.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/users/responses/TokenSummary.java
@@ -21,18 +21,25 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 import org.graylog.autovalue.WithBeanGetter;
-
-import java.util.List;
+import org.joda.time.DateTime;
 
 @JsonAutoDetect
 @AutoValue
 @WithBeanGetter
-public abstract class TokenList {
+public abstract class TokenSummary {
     @JsonProperty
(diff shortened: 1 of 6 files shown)


mpfz0r (approved):
backend looks good

ousmaneo (approved):
Legit !

PR #9747, opened 2020-12-09
Remove div wrapping GlobalThroughput in cloud

edmundoa:
`Nav` passes some props to each one of its children, so using a div
as wrapper means that React logs a warning in the browser console log on
each render, see:
https://github.com/Graylog2/graylog2-server/pull/9729#issuecomment-741892217

Since `GlobalThroughput` returns a `NavItem` and accepts props, we can
pass the `disabled` prop to style it when it is not a link, and also
remove the warning from the console.

Refs #9729


Gaya:
Whoops! My bad! Thanks for fixing

PR #9748, opened 2020-12-09
Date Time Picker variable renaming

kyleknighted:
Updating some of the variable names that were discussed.
Please let me know if you see any others we should modify


linuspahl (approved):
Looks good to me.

PR #9759, opened 2020-12-10
Improve pulsar performance

mpfz0r:
Use `sendAsync()` instead of `send()`.

Sending each message synchronously has obviously a huge performance impact.

Instead of ~500 msg/s we can now produce ~50k msg/s


thll (approved):
LGTM

I tested the behaviour when the pulsar server is not available while producing messages. As expected, the async call becomes blocking after 1000 messages have been queued up in the send buffer of the client. Once the server becomes available again, all messages make it through to the pulsar server.

thll:
@mpfz0r feel free to merge once the build is green

PR #9760, opened 2020-12-10
Hide email sender field for Alert Notifications

edmundoa:
- Make `sender` field in Email Alert Notification optional, since that will use the default set in the Graylog configuration file
- Hide `sender` field in both the Email Alert Notification and the Legacy Email Alarm Callback
- Small linter and UX fixes

Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/576

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
+++ b/graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
@@ -172,11 +172,13 @@ public class EmailAlarmCallback implements AlarmCallback {
     // I am truly sorry about this, but leaking the user list is not okay...
     private ConfigurationRequest getConfigurationRequest(Map<String, String> userNames) {
         ConfigurationRequest configurationRequest = new ConfigurationRequest();
-        configurationRequest.addField(new TextField("sender",
-                "Sender",
-                "",
-                "The sender of sent out mail alerts",
-                ConfigurationField.Optional.OPTIONAL));
+        if (!graylogConfig.isCloud()) {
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
```diff
@@ -172,11 +172,13 @@ public void initialize(Configuration config) throws AlarmCallbackConfigurationEx
     // I am truly sorry about this, but leaking the user list is not okay...
     private ConfigurationRequest getConfigurationRequest(Map<String, String> userNames) {
         ConfigurationRequest configurationRequest = new ConfigurationRequest();
-        configurationRequest.addField(new TextField("sender",
-                "Sender",
-                "",
-                "The sender of sent out mail alerts",
-                ConfigurationField.Optional.OPTIONAL));
+        if (!graylogConfig.isCloud()) {
```
mpfz0r, line 175: This is actually dead code. But it doesn't hurt to do that here as well

mpfz0r (approved):
LGTM

We could make the sender optional in the server anyway.
The notification backend supports that.

graylog2-server/src/main/java/org/graylog2/alarmcallbacks/EmailAlarmCallback.java
edmundoa: Indeed, it seems it will go away in 4.1. Until then someone could still create those through the UI I guess, so better safe than sorry :)

ousmaneo (approved):
Frontend looks good too!

PR #9762, opened 2020-12-10
Allow user access to event annotation for streams he has access to

### LINKED ISSUE #8260: Event Annotation needs permission from "All events" and "All system events"
The Event Annotation option needs permission from "All events" and "All system events"'s Stream and not only permissions from related Stream from Dashboard

kmerz:
## Motivation
Prior to this change, the user needed additional access to `All Events`
and `All system events` to search with Event Annotation.

## Description
This change will change the streams used to check for permission.
Instead of effective streams which contains the Event Streams,
we use the streams the user has access to. We will filter with these
streams later the result of the Event Streams request.

## Types of changes

Fixes #8260


dennisoelkers (approved):
Looks good, works fine!

PR #9763, opened 2020-12-10
Fix Docker bind mounts for resources inside of JAR files

bernd:
In some cases a resource URL references a file inside a JAR file instead
of a file in the file system. To be able to mount these files into
Docker container via testcontainers, we need to extract them into a
temporary file first.

- Move ResourceUtil one package up and make its methods public so we can
  reuse it
- Extract resources into temorary files when resource URL points into a
  JAR file
- Bind each TLS related file in OpenLDAPContainer separately to avoid
  having to add more logic to extract complete directories from
  resources

Fixes Graylog2/graylog-plugin-enterprise#1812


mpfz0r (approved):
LGTM :+1:

PR #9770, opened 2020-12-11
Fix version probe for ES snapshot versions

mpfz0r:
My machine is running elastic `7.9.3-SNAPSHOT`
which couldn't be parsed and prevented the server to start.

- Use the semver parser instead of a hand rolled version which
  cannot deal with prerelease tags

- Also, avoid throwing an exception at this point, which forever blocks
  the AtomicCache. I guess we don't complete the Future in this case.


dennisoelkers (approved):
LGTM! Thanks for the fix!

PR #9774, opened 2020-12-11
Remove remaining eslint-disable-camelcase comments

kmerz:
## Motivation
Prior to this change, we set the eslint config that
it does not bother us with camelcase warnings.

## Description
This change was proudly presented by sed:

`git grep -l camelcase | xargs sed -i '/camelcase/d'`


dennisoelkers (approved):
Thanks!

PR #9782, opened 2020-12-11
Create separate form for date time picker.

linuspahl:
## Description
This PR is creating a separate form for the search date time picker. Separating these fields from the `SearchBarForm` helps to separate the concerns. This way we can simplify the related form state and logic.


kyleknighted:
unfortunately I am still finding bugs in the Relative time range tab on Dashboard.
If you choose All Time, Apply it, then open it
You cannot uncheck All Time, but you can go to Absolute tab and then back to Relative and it is unchecked and 5 Min

PR #9783, opened 2020-12-11
Hide API browser and API docs when running on cloud

edmundoa:
This PR adds an annotation for REST resources that let us mark resources that should be available in on-premise setups but not on cloud. It also uses that annotation to mark the `DocumentationResource` and `DocumentationBrowserResource` as hidden on cloud.

There probably is a better way to do this, but I could not find it with my (little) knowledge in that part of our stack, so I'm open for suggestions if you have any!

Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/91

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/jersey/PrefixAddingModelProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog2/jersey/PrefixAddingModelProcessor.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableMap;
 import org.glassfish.jersey.server.model.ModelProcessor;
 import org.glassfish.jersey.server.model.Resource;
 import org.glassfish.jersey.server.model.ResourceModel;
+import org.graylog2.shared.rest.HideOnCloud;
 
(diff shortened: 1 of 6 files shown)


thll (approved):
Looks good and worked well in a smoke test!

I'd like to hear another opinion about the general approach from @bernd or @mpfz0r before merging.

Also I'm wondering if we should name the annotation `HideInCloud` or `HideForCloud` rather than `HideOnCloud`? 🤷‍♂️

edmundoa:
Took the naming from the frontend, where we have a `HideOnCloud` component, but we don't have to use the same.

mpfz0r (approved):
LGTM. The only other solution I could think of, is packing the resources into a separate module.
But this is better, IMO

mpfz0r:
I'm also fine with the `HideOnCloud` name. From our perspective we are actually _on_ the cloud :-D

thll:
> Took the naming from the frontend, where we have a `HideOnCloud` component, but we don't have to use the same.

let's leave it like it is then!

PR #9784, opened 2020-12-11
Update Absolute time inputs to use number instead of text

kyleknighted:
Allows for using number inputs so time can be driven with keyboard up/down buttons.


ousmaneo (approved):
Tested and Legit

PR #9788, opened 2020-12-11
DateTimePicker - capture Enter and Esc keys to submit or cancel range selection

kyleknighted:
Adds functionality to TimeRangeDropdown that captures the `Enter` key so it does not submit the SearchBarForm and instead just handles the Apply function if `isValid`. 

Similar for `Esc` key, but it Cancels the Dropdown

_One issue I saw is that clicking the Cancel button properly cleans up Formik to not render the `dirty` flag, but the Esc key running the same function does not properly clean up Formik. I am not sure what the cause is here, but a second set of eyes on this behavior would be awesome._


ousmaneo (approved):
Legit, I will review again and merge once the conflict is fixed

ousmaneo (approved):
Looks great!

PR #9792, opened 2020-12-14
Sync cloud-4.0 branch with latest changes in 4.0

edmundoa:
Sync `cloud-4.0` branch with latest changes in `4.0`. This time with the right base branch, or so it seems :)

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise-integrations#444


thll (approved):

thll:
@edmundoa do you want to address the linter error in this PR?

edmundoa:
@thll no, I think that has been there like that for a while and I don't think we should change it here.

thll:
> @thll no, I think that has been there like that for a while and I don't think we should change it here.

all right. merging

PR #9797, opened 2020-12-15
Updating "No Override" content of Date Time Picker on Dashboard

kyleknighted:
## comes in after #9789 

Adds content for Dashboard `No override` state

image


ousmaneo (approved):
Tested and works!

PR #9805, opened 2020-12-15
Fix order of tabs in dashboard after install

### LINKED ISSUE #8264: Content Package Dashboards has tabs order mixed after install
## Expected Behavior

When installing Content Package with dashboards that have tabs, the tabs should keep the original order.

## Current Behavior

The tabs got order mixed after installing the contenta package dashboards.

## Possible Solution

Review how the dashboards are saving tabs and how to keep the order.

1. Create a content package with dashboards that has at least four tabs.
2. Install the content package
3. Check the dashboards tabs

Some of them got mixed and some don't.

## Context

It is not a big issue because we still can duplicate the tabs and rename them to keep the right order. But will be good don't have that trouble.

## Your Environment

* Graylog Version: Graylog 3.3.0+4ea5649 
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kmerz:
## Motivation
Prior to this change, a dashboard installed by a content pack had a
tabs in a random order. Since we cannot change the order of the tabs
this is quite inconvinient for the user.

## Description
This change will do mutiple steps to fix that:
- Use a LinkedHashMap to ensure the order stays fix during content pack
  creation.
- Instead of using the serach QueryIds we use the views QueryIds which
  we previously fixed in the order.
- When loading a view and selecting the activeQuery we also use the
  view.state queryids instead of the search ones.

Fixes #8264

## How Has This Been Tested?
1. Create a Dashboard
1. Create multiple Tabs with names [ 1, 2, 3, 4, 5 ]
1. Create a content pack with this Dashboard
1. Install that dashboard
1. Ensure the tabs order is still the same.

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewDTO.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewDTO.java
@@ -18,18 +18,12 @@ package org.graylog.plugins.views.search.views;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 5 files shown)


dennisoelkers (approved):
LGTM! :tada:

PR #9809, opened 2020-12-15
DateTime Picker - Updating error message styles

kyleknighted:
Making sure errors are similar in all time range options
image
image

image


ousmaneo (changes requested):
It looks great. I just notice a small thing: on the relative range for example when an error is displayed for the "from" block the whole "until" block got pushed a bit to the bottom

ousmaneo (approved):
Legit

PR #9810, opened 2020-12-15
DateTimePicker - Adding limitDuration config to Widget

kyleknighted:
Widget DateTime Picker needed limitDuration from the search config

image


graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
```diff
@@ -83,6 +88,7 @@ const EditWidgetFrame = ({ children }: Props) => {
     return <Spinner text="Loading widget ..." />;
   }
 
+  const limitDuration = moment.duration(config.query_time_range_limit).asSeconds();
```
dennisoelkers, line 91: Do we need checks here if 

1. `config` or 
2. `query_time_range_limit` are `undefined`?

The first would result in an exception, the second forces us to rely on `moment.duration` returning an empty duration if the argument is `undefined`. IMO we should make this more explicit like this:

```suggestion
  const limitDuration = config?.query_time_range_limit !== undefined ? moment.duration(config.query_time_range_limit).asSeconds() : 0;
```

dennisoelkers (changes requested):
Looks good! Some suggestions for smaller improvements inline.

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
```diff
@@ -124,4 +130,8 @@ EditWidgetFrame.propTypes = {
   children: PropTypes.node.isRequired,
 };
 
-export default EditWidgetFrame;
+export default connect(EditWidgetFrame, {
```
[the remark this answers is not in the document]
kyleknighted: Great suggestion! Had to modify slightly, but thanks for the reminder on this hook!

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
```diff
@@ -93,7 +99,7 @@ const EditWidgetFrame = ({ children }: Props) => {
            animation={false}
            dialogComponentClass={EditWidgetDialog}
            enforceFocus={false}>
-      <SearchBarForm initialValues={{ limitDuration: 0, timerange, streams, queryString }}
+      <SearchBarForm initialValues={{ limitDuration, timerange, streams, queryString }}
```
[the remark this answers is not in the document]
kyleknighted: It's not really part of the entire state, and I agree having a wrapping context may be better, but this is already written in such a way across Search and Dashboard, so it would require updating in a few places. Maybe this is something we can include as a follow-up later or do you feel it is important enough to rework all of them?

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
kyleknighted: `moment.duration(undefined | null)` will return `0`. 

If `config` does not return Search Config, then we are already having issues everywhere else with searching because we will not have the necessary data.

If `query_time_range_limit` also does not return something, then we should define this as a default in the SearchConfigStore to always return `0` if it does not exist.

I do agree that there should be some checking though so added the optional chaining to the object

`const limitDuration = moment.duration(config?.query_time_range_limit).asSeconds();`

Maybe something like `const limitDuration = moment.duration(config?.query_time_range_limit).asSeconds() ?? 0;` would be good?

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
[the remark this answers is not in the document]
dennisoelkers: If you keep track of it and rework it before the picker is merged into `master`, then this is fine.

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
dennisoelkers: The last option would be okay for me too.

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
```diff
@@ -71,6 +75,8 @@ const onSubmit = (values, widget: Widget) => {
 };
 
 const EditWidgetFrame = ({ children }: Props) => {
+  const config: SearchesConfig = useStore(SearchConfigStore, ({ searchesClusterConfig }) => searchesClusterConfig);
```
dennisoelkers, line 78: IMO we can remove the explicit `SearchesConfig` type. It should be inferred from `SearchConfigStore`'s type.

dennisoelkers (commented):
Better already! Just one tiny thing.

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
kyleknighted: @dennisoelkers Updated!

graylog2-web-interface/src/views/components/widgets/EditWidgetFrame.tsx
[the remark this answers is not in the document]
kyleknighted: This conversation was followed-up in https://github.com/Graylog2/graylog2-server/pull/9893

dennisoelkers (approved):

PR #9814, opened 2020-12-16
Show timezone setting to the user

### LINKED ISSUE #9813: User cannot edit his own timezone setting
## Expected Behavior
A user should be able to change his own timezone setting since it is not security relevant.

## Current Behavior
Right now the whole settings section is only visible for a admin user.

## Possible Solution
Display the settings section and only hide the Timeout section to the user.

## Notes
This is already fixed in 4.1 development

## Your Environment
* Graylog Version: 4.0

kmerz:
## Motivation
Prior to this change, the user was not able to change his own timezone
settings since they were hidden to him.
    
## Description
This change will show the settings to the user so he can edit them.

## Types of changes

Fixes #9813

### DIFF
--- a/graylog2-web-interface/src/components/users/UserDetails/UserDetails.jsx
+++ b/graylog2-web-interface/src/components/users/UserDetails/UserDetails.jsx
@@ -45,9 +45,7 @@ const UserDetails = ({ user }: Props) => {
         <IfPermitted permissions={`users:edit:${user.username}`}>
           <div>
             <ProfileSection user={user} />
-            <IfPermitted permissions="*">
-              <SettingsSection user={user} />
-            </IfPermitted>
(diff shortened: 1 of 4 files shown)


mpfz0r:
hm, shouldn't this be fixed in `master` first?

kmerz:
> hm, shouldn't this be fixed in `master` first?

@mpfz0r This is already fixed in master as I implemented the Startpage feature. This is where I figured that it might be a problem in 4.0.

mpfz0r (approved):
Works and LGTM. 
But like @dennisoelkers  mentioned, we should really replace the permissions check for `*` with an actual new permission `users:settimeout`. (In the backend as well)

This is fine for `4.0`, but could you do that for `master`? Or create an issue?

PR #9821, opened 2020-12-17
Add common Okta version to parent POM

danotorrey:
## Description

Add common Okta version to parent POM so that the same version can be used across all plugins.

## Motivation and Context

Initially, this will be used to solve a version conflict between the Enterprise Integrations and Cloud plugin usages of the Okta SDK.

## How Has This Been Tested?

This change has no affect on functionality until it is subsequently used within plugins.


danotorrey:
This requires a backport to `cloud-4.0` in order to be used for Cloud. I can create the backport PR when this is merged.

waab76 (approved):
LGTM

PR #9836, opened 2020-12-21
Use plotlys colorscales for Heatmap Aggregation

kmerz:
## Motivation and Context
Plotly provides a list of predefined colorscales for its Heatmap. Instead of hard coding a single color scale
we should provide that list to the use so he can use a colorscale matching to his data.
Also it is often useful to define the range of the scale. For example if we have 100 data points in the range from 1 to 10
and 2 data points in the range 60 to 70, the range of the colorscale is defined from 1 to 70. The 100 data points are now invisible.
If the user now selects the data range from 1 to 11 the 2 data point 2 with 60 and 70 will now still look out but the rest is more
readable.

## Description
- Add a select where we list the colorscales poltly accepts
- Add a Min and Max field so the user can enter the range of the scale
- Add Autoscale which enables of disables the Min, Max fields
- Add a default value which will be rendered if a undefined value was found, until now we used 'None' and draw the hard
  coded background, but we do not have it any longer, so the background will now 
- Add a checkbox to automatically set the default to the smallest value.

Heatmap-Scale

## How Has This Been Tested?
- Created a Heatmap and run through all possible values

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
@@ -15,7 +15,8 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import React, { useCallback } from 'react';
-import { capitalize } from 'lodash';
+
+import { Checkbox } from 'components/graylog';
 
 import HeatmapVisualizationConfig, { COLORSCALES } from 'views/logic/aggregationbuilder/visualizations/HeatmapVisualizationConfig';
 
@@ -30,15 +31,19 @@ const _makeOption = (value) => ({ label: value, value });
 const colorScalesOptions = COLORSCALES.map(_makeOption);
 
 const HeatmapVisualizationConfiguration = ({ config = HeatmapVisualizationConfig.empty(), onChange }: Props) => {
-  const _onChange = useCallback(({ value }) => onChange(config.toBuilder().colorScale(value).build()), [config, onChange]);
+  const _onColorScaleChange = useCallback(({ value }) => onChange(config.toBuilder().colorScale(value).build()), [config, onChange]);
+  const _onReverseScaleChange = useCallback((e) => onChange(config.toBuilder().reverseScale(e.target.checked).build()), [config, onChange]);
 
   return (
     <>
       <span>Color Scheme</span>
       <Select placeholder="Select Color Scheme"
-              onChange={_onChange}
+              onChange={_onColorScaleChange}
               options={colorScalesOptions}
               value={_makeOption(config.colorScale)} />
+      <span>Reverse Scale</span>
+      <Checkbox onChange={_onReverseScaleChange}
+                checked={config.reverseScale} />
     </>
   );
 };
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
@@ -103,6 +103,7 @@ const HeatmapVisualizationConfiguration = ({ config = HeatmapVisualizationConfig
       <Select placeholder="Select Color Scheme"
               onChange={_onColorScaleChange}
               options={colorScalesOptions}
+              isClearable={false}
               value={_makeOption(config.colorScale)} />
       <Checkbox onChange={_onReverseScaleChange}
                 checked={config.reverseScale}>
--- a/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
@@ -90,6 +90,7 @@ const HeatmapVisualizationConfiguration = ({ config = HeatmapVisualizationConfig
   const _onAutoScaleChange = useCallback((e) => onChange(config.toBuilder().autoScale(e.target.checked).build()), [config, onChange]);
   const _onZminChange = useCallback((e) => onChange(config.toBuilder().zMin(_parseEvent(e)).build()), [config, onChange]);
   const _onZmaxChange = useCallback((e) => onChange(config.toBuilder().zMax(_parseEvent(e)).build()), [config, onChange]);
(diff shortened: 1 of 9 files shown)


graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+const HeatmapVisualizationConfiguration = ({ config = HeatmapVisualizationConfig.empty(), onChange: onChangeFunc }: Props) => {
+  const [errors, setErrors] = useState<Error>({});
+  const onChange = useCallback((newConfig) => {
+    if (_validateConfig(newConfig, setErrors)) {
+      onChangeFunc(newConfig);
+    }
+  }, [onChangeFunc]);
+  const _onColorScaleChange = useCallback(({ value }) => onChange(config.toBuilder().colorScale(value).build()), [config, onChange]);
```
dennisoelkers, line 88: We could clean this up a bit by introducing something like this:

```suggestion
  const modifyConfig = useCallback((fn: (builder: HeatmapVisualizationConfigBuilder) => HeatmapVisualizationConfigBuilder) => {
    onChange(fn(config.toBuilder()).build());
  }, [onChange, config]);
  const _onColorScaleChange = useCallback(({ value }) => modifyConfig((builder) => builder.colorScale(value)), [modifyConfig]);
```

The next five callbacks could then reuse `modifyConfig` as well. It would require exporting the type from the `Builder` class in `HeatmapVisualizationConfig` as `HeatmapVisualizationConfigBuilder` though.

graylog2-web-interface/src/views/components/visualizations/heatmap/HeatmapVisualization.tsx
```diff
@@ -134,7 +120,7 @@ const _chartLayout = (heatmapData) => {
   return {
     yaxis: axisConfig,
     xaxis: axisConfig,
-    plot_bgcolor: hasContent ? BG_COLOR : 'transparent',
+    // plot_bgcolor: hasContent ? BG_COLOR : 'transparent',
```
dennisoelkers, line 123: Why is this commented out?

graylog2-web-interface/src/views/components/visualizations/heatmap/HeatmapVisualization.tsx
```diff
@@ -98,11 +78,17 @@ const _transposeMatrix = (z: Array<Array<any>> = []) => {
...
+
+const _formatSeries = (visualizationConfig) => ({ valuesBySeries, xLabels }: {valuesBySeries: ValuesBySeries, xLabels: Array<any>}): ExtractedSeries => {
+  const valuesFoundBySeries = values(valuesBySeries);
   // When using the hovertemplate, we need to provie a value for empty z values.
   // Otherwise plotly would throw errors when hovering over a field.
   // We need to transpose the z matrix, because we are changing the x and y label in the generator function
-  const z = _transposeMatrix(_fillUpMatrix(values(valuesBySeries), xLabels));
+  const defaultValue = visualizationConfig.useSmallestAsDefault
+    ? _findSmallestValue(valuesFoundBySeries)
+    : visualizationConfig.defaultValue ?? 'None';
```
dennisoelkers, line 90: This could use some braces to make the operator precedence clearer.

graylog2-web-interface/src/views/logic/aggregationbuilder/visualizations/HeatmapVisualizationConfig.ts
```diff
@@ -0,0 +1,248 @@
...
+  | 'Portland'
+  | 'Jet'
+  | 'Hot'
+  | 'Blackbody'
+  | 'Earth'
+  | 'Electric'
+  | 'Viridis'
+  | 'Cividis';
+
+export const COLORSCALES = ['Greys', 'YlGnBu', 'Greens', 'YlOrRd', 'Bluered', 'RdBu', 'Reds', 'Blues', 'Picnic', 'Rainbow', 'Portland', 'Jet', 'Hot', 'Blackbody', 'Earth', 'Electric', 'Viridis', 'Cividis'];
```
dennisoelkers, line 41: To avoid duplicating all values in the type, we can do:
```suggestion
export const COLORSCALES = ['Greys', 'YlGnBu', 'Greens', 'YlOrRd', 'Bluered', 'RdBu', 'Reds', 'Blues', 'Picnic', 'Rainbow', 'Portland', 'Jet', 'Hot', 'Blackbody', 'Earth', 'Electric', 'Viridis', 'Cividis'];

type ColorScale = typeof COLORSCALES[number];
```

dennisoelkers (changes requested):
Looks awesome and improves the heatmap a lot! Added some minor comments inline.

graylog2-web-interface/src/views/components/visualizations/heatmap/HeatmapVisualization.tsx
kmerz: It must be deleted, we can't set the BG_COLOR anymore, since we do not have it, with the themes.

graylog2-web-interface/src/views/components/aggregationbuilder/HeatmapVisualizationConfiguration.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const _onZmaxChange = useCallback((e) => modifyConfig((builder: HeatmapConfigBuilder) => builder.zMax(_parseEvent(e))), [modifyConfig]);
+  const _onUseSmallestAsDefaultChange = useCallback((e) => modifyConfig((builder: HeatmapConfigBuilder) => builder.useSmallestAsDefault(e.target.checked)), [modifyConfig]);
+  const _onDefaultValueChange = useCallback((e) => modifyConfig((builder: HeatmapConfigBuilder) => builder.defaultValue(_parseEvent(e))), [modifyConfig]);
+
+  return (
+    <>
+      <span>Color Scheme</span>
+      <Select placeholder="Select Color Scheme"
+              onChange={_onColorScaleChange}
+              options={colorScalesOptions}
```
dennisoelkers, line 105: Can we add `isClearable={false}` here? IMO it does not make sense to show the clear icon in here.

dennisoelkers (commented):
Looks really good now! Just one tiny last remark.

dennisoelkers (approved):

PR #9838, opened 2020-12-21
Save focused widget id in url

### LINKED ISSUE #9728: Support auto-focussing widget from search/dashboard URL
## What?

The URL to a search/dashboard can contain the widget id of a widget contained in the current search/dashboard which is focussed automatically when this search/dashboard is opened.

Refs #9727.

## Why?

This functionality allows a user to share the current focus state with someone else by exchanging the generated URL.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kmerz:
## Motivation
Prior to this change, we did not save the state
of the focused widget in the url.

## Description
This change will sync the focused widget id with the
url so a user can copy and past the state to another
user.

## How Has This Been Tested?
1. Focus a widget
2. Copy url
3. open a new tab 
4. paste url

Fixes #9728 

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
+++ b/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
@@ -15,16 +15,47 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useCallback } from 'react';
+import { useLocation, useHistory } from 'react-router-dom';
 import PropTypes from 'prop-types';
+import URI from 'urijs';
 
 import { useStore } from 'stores/connect';
 import { WidgetStore } from 'views/stores/WidgetStore';
 import WidgetFocusContext from 'views/components/contexts/WidgetFocusContext';
 
+const _syncWithQuery = (query: string, focusedWidget: string) => {
+  const baseUri = new URI(query)
+    .removeSearch('focused');
+
+  if (focusedWidget) {
+    return baseUri.setSearch('focused', focusedWidget).toString();
+  }
+
+  return baseUri.toString();
+};
+
+const getFocusedWidgetFromSearch = (search: string) => {
+  const uri = new URI(search);
+  const { focused } = uri.search(true);
+
+  return focused;
+};
+
 const WidgetFocusProvider = ({ children }: { children: React.ReactNode }): React.ReactElement => {
-  const [focusedWidget, setFocusedWidget] = useState(undefined);
+  const { search, pathname } = useLocation();
+  const query = pathname + search;
+  const history = useHistory();
+  const [focusedWidget, setFocusedWidget] = useState<string | undefined>();
   const widgets = useStore(WidgetStore);
+
+  useEffect(() => {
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -15,27 +15,63 @@
...
+const getFocusedWidgetFromSearch = (search: string) => {
+  const uri = new URI(search);
+  const { focused } = uri.search(true);
+
+  return focused;
+};
+
 const WidgetFocusProvider = ({ children }: { children: React.ReactNode }): React.ReactElement => {
-  const [focusedWidget, setFocusedWidget] = useState(undefined);
+  const { search, pathname } = useLocation();
```
dennisoelkers, line 46: You can use this instead:
```suggestion
  const { focused } = useQuery();
```

`useQuery` should be imported from `routing/useQuery`.
This already does the query string parsing and returns an object containing the query parameters.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -15,27 +15,63 @@
...
+  const query = pathname + search;
+  const history = useHistory();
+  const [focusedWidget, setFocusedWidget] = useState<string | undefined>();
   const widgets = useStore(WidgetStore);
 
+  useEffect(() => {
```
dennisoelkers, line 52: We could extract this into a `useFocusedWidgetIdFromURL()` hook to improve readability.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -15,27 +15,63 @@
...
       setFocusedWidget(undefined);
     }
   }, [focusedWidget, widgets]);
 
-  const updateFocus = (widgetId: string | undefined | null) => (
-    widgetId === focusedWidget
-      ? setFocusedWidget(undefined)
-      : setFocusedWidget(widgetId));
+  const updateFocus = useCallback((widgetId: string | undefined | null) => {
+    const newFocus = widgetId === focusedWidget
```
dennisoelkers, line 67: I do not understand this part. If the `updateFocus` function is called with the widget id of the currently focused widget, `_syncWithQuery` would be called with `undefined` and would remove the `focused` query parameter. Or am I misreading something?

Maybe we should simplify the functionality of the `updateFocus` function. We could let it set:
1. the URL only (which would then be synced with the state in `WidgetContextProvider`)
2. the state only (through `setFocusedWidget`) and synchronize with the URL in `WidgetContextProvider`

dennisoelkers (changes requested):
Looks good and works! Added some remarks inline.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
kmerz: This is again related to the bug, which only @linuspahl could reproduce. I am pretty sure this part is not needed any longer since we are using a ContextProvider and not a Store anymore, but I would like to remove this when @linuspahl can test it.

dennisoelkers (approved):
Looks good and works! :tada:

PR #9841, opened 2020-12-22
Provide default position in case the widget isn't found in positions map

### LINKED ISSUE #9778: Widgets moved from the search page can't be moved to another dashboard page.
## Description
On dashboards it is possible to move a widget to a different page. This is not the case when a widget got copied to the dashboard from the search page.

1. Open `/search`
2. Copy any widget to a dashboard with multiple pages
3. Select "Move To Page" in the widget action menu.
4. Select a page and click "Submit"
5. Nothing happens

The console shows the error:
```
react-dom.development.js:327 Uncaught TypeError: Cannot read property 'toBuilder' of undefined
    at MoveWidgetToTab (MoveWidgetToTab.ts:113)
    at Widget.tsx:211
    at MoveWidgetToTabModal.tsx:52
    at BootstrapModalForm.jsx:102`
```

## Your Environment

* Graylog Version: 4.0.2

kmerz:
## Motivation
Prior to this change, a widget copied from a search was unable to be copied to a different page since the position was missing in the positions map.

## Description
This change will add a safety net in case the position is missing we provide the widgets default size.

## How Has This Been Tested?
1. go to `/search`
2. Click on `Copy to Dashbaord`
3. Wait for `dashboard` to render
4. click on `Move to Page`

Fixes #9778 

Needs Backport to 4.0

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
+++ b/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
@@ -18,12 +18,12 @@ import { widgetDefinition } from 'views/logic/Widgets';
 import WidgetPosition from 'views/logic/widgets/WidgetPosition';
 import Widget from 'views/logic/widgets/Widget';
 
-const GetPositionForNewWidget = (widget: Widget) => {
+const GetPositionForNewWidget = (widget: Widget, oldWidgetPosition: WidgetPosition) => {
   const { defaultHeight, defaultWidth } = widgetDefinition(widget.type);
 
   return WidgetPosition.builder()
-    .height(defaultHeight)
-    .width(defaultWidth)
+    .height(oldWidgetPosition?.height ?? defaultHeight)
+    .width(oldWidgetPosition?.width ?? defaultWidth)
     .col(1)
     .row(1)
     .build();
--- a/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
+++ b/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
@@ -17,15 +17,16 @@
 import { widgetDefinition } from 'views/logic/Widgets';
 import WidgetPosition from 'views/logic/widgets/WidgetPosition';
 import Widget from 'views/logic/widgets/Widget';
-import View from 'views/logic/views/View';
 
-const GetPositionForNewWidget = (widget: Widget, queryId: string, view: View) => {
-  const widgetDef = widgetDefinition(widget.type);
+const GetPositionForNewWidget = (widget: Widget) => {
+  const { defaultHeight, defaultWidth } = widgetDefinition(widget.type);
 
-  const widgetPositionsBuilder = view.state.get(queryId).widgetPositions[widget.id]?.toBuilder()
-    || WidgetPosition.builder().width(widgetDef.defaultWidth).height(widgetDef.defaultHeight);
-
-  return widgetPositionsBuilder.col(1).row(1).build();
+  return WidgetPosition.builder()
+    .height(defaultHeight)
+    .width(defaultWidth)
+    .col(1)
+    .row(1)
+    .build();
 };
 
 export default GetPositionForNewWidget;
--- a/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
+++ b/graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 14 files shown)


dennisoelkers (commented):
Could this be consolidated with the default placement of newly created widgets? This way we would have a central policy for the placement of new widgets (top vs. bottom vs. more intelligent placement)

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+import { widgetDefinition } from 'views/logic/Widgets';
+import WidgetPosition from 'views/logic/widgets/WidgetPosition';
+import Widget from 'views/logic/widgets/Widget';
+import View from 'views/logic/views/View';
+
+const GetPositionForNewWidget = (widget: Widget, queryId: string, view: View) => {
+  const widgetDef = widgetDefinition(widget.type);
+
+  const widgetBuilder = view.state.get(queryId).widgetPositions[widget.id]?.toBuilder()
```
dennisoelkers, line 25: The naming is off here: This is more of a `widgetPositionBuilder` or similar.

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
```diff
@@ -0,0 +1,31 @@
...
+    || WidgetPosition.builder().width(widgetDef.defaultWidth).height(widgetDef.defaultHeight);
+
+  return widgetBuilder.col(1).row(1).build();
```
dennisoelkers, line 28: Why is the position of the widget overriden to `1,1`, even if a position for the widget existed before?

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
```diff
@@ -0,0 +1,31 @@
...
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import { widgetDefinition } from 'views/logic/Widgets';
+import WidgetPosition from 'views/logic/widgets/WidgetPosition';
+import Widget from 'views/logic/widgets/Widget';
+import View from 'views/logic/views/View';
+
+const GetPositionForNewWidget = (widget: Widget, queryId: string, view: View) => {
+  const widgetDef = widgetDefinition(widget.type);
```
dennisoelkers, line 23: This logic is contained in `AddNewWidgetsToPositions` as well. Can we consolidate it?

graylog2-web-interface/src/views/logic/views/__snapshots__/CopyWidgetToDashboard.test.js.snap
```diff
@@ -17,27 +17,33 @@ Object {
         "546b13fe-eae7-4fb8-abb9-e0e3a907c3e6": Object {
           "col": 1,
           "height": 3,
-          "row": 5,
+          "row": NaN,
```
dennisoelkers, line 20: Isn't this an indication that the code has an issue? IMO it should not generate a `NaN` value for the `row`.

graylog2-web-interface/src/views/logic/views/__snapshots__/MoveWidgetToTab.test.ts.snap
```diff
@@ -194,7 +194,85 @@ Object {
...
+      "widgets": Immutable.List [],
+    },
+    "5faea09b-4187-4eda-9d59-7a86d4774c73": Object {
+      "formatting": Object {
+        "highlighting": Array [],
+      },
+      "positions": Immutable.Map {
+        "dead-beef": Object {
+          "col": 1,
+          "height": undefined,
```
dennisoelkers, line 225: Is this correct? `height` & `width` are `undefined`.

dennisoelkers (changes requested):
Added a couple of remarks/questions inline.

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
kmerz: Because it is the position for a new widget, and right now we place the new widgets on top.

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
kmerz: I do not really see how it would it improve the code. :confused:

graylog2-web-interface/src/views/logic/views/GetPositionForNewWidget.ts
dennisoelkers: `widgetBuilder` is either:
  1. a previously existing widget position (retrieved through `view.state.get(queryId).widgetPositions[widget.id]?.toBuilder()`)
  2. a newly created widget position with default dimensions (created through `WidgetPosition.builder().width(widgetDef.defaultWidth).height(widgetDef.defaultHeight)`)

In the first case the widget position should not be overridden IMO.

dennisoelkers (changes requested):
Looks good and works, just some last cleanups.

dennisoelkers (approved):

PR #9844, opened 2020-12-22
Custom chart legend

### LINKED ISSUE #6612: Value actions in View widgets Pie and Bar charts
Please add Value actions in View widgets also for Pie and Bar charts

## Expected Behavior
Views widgets Pie chart and Bar chart should also support Value actions such as Add to query, Exclude from results. Now it is only supported in Data table and Single Number widgets.

## Current Behavior
You can only color in Pie or Bar charts widget in View. 

## Possible Solution
Add small arrow with Value actions menu next to values below chart. Or better, change behaviour of changing color, so you can change color only if you click on color rectable (make sence), but no value. If you click on value below chart, aplication should open Value actions menu.

view-pie-chart-actions

## Your Environment
Graylog 3.1.2 OVA

### LINKED ISSUE #7207: Unstable color assignment in graph widget
## Expected Behavior

The color assignment in a graph widget should be stable for the same values.

## Current Behavior

The colors change from time to time on reload:

Peek 2020-01-16 13-26

This is the configuration:

image

1. Create graph widget with a single row and a single metric
2. Enable auto-reload
3. See color assignment change from time to time

## Your Environment

* Graylog Version: 3.2.0-beta.4-SNAPSHOT (rev cb33786)

kmerz:
## Motivation and Context
The user wants to be able to change the query not only from the query bar and message list, but also from a aggregation widget.
For that the legend which is displaying the fields values is a good starting point.

## Description
- Create our own custom plot legend so we can handle the onClick event for color hint and the value separably
- To provide the colors for the charts we introduce a ColorMapper which will take care of the default colors and saved colors 

value-actions-in-chart

Fixes #6612 
Fixes #7207 by keeping track of field to color map thoughout the session

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/common/ColorPicker.jsx
+++ b/graylog2-web-interface/src/components/common/ColorPicker.jsx
@@ -14,24 +14,33 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-import React from 'react';
-import createReactClass from 'create-react-class';
+import * as React from 'react';
+import { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { SwatchesPicker } from 'react-color';
 
 /**
  * Color picker component that let the user select a color from a list of 95 colors grouped by hue.
  */
-const ColorPicker = createReactClass({
-  propTypes: {
-    /** Indicates the selected color in hexadecimal format. */
-    color: PropTypes.string,
-    /**
+const ColorPicker = ({ onChange, ...rest }) => {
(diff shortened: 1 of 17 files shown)


dennisoelkers (changes requested):
I am seeing some regressions with the color picker popover:
image
And it is now by default going out of bounds for the default histogram's legend:
image

dennisoelkers (changes requested):
The legend of a pie chart is cut off for me in edit mode:

image

dennisoelkers (approved):

aim4r:
This looks awesome, thanks for implementing it !

PR #9852, opened 2020-12-29
Retry read timeouts for Elasticsearch.

dennisoelkers:
## Motivation and Context

Before this change, the retry strategy used for ES requests did not retry for instances of `SocketTimeoutException`, raised in case of read timeouts. This resulted in exceptions being raised when individual nodes were not responding and node discovery was not used. This was e.g. being noticable when performing searches and 1 of 3 nodes was failing, so every 3rd search request was timing out and returning an error instead of being retried and successful and just taking longer.

This change is adding this class to the exceptions being retried, so any request that goes out to a node which is not responding on time is retried.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/jest/GraylogJestRetryHandlerTest.java
+++ b/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/jest/GraylogJestRetryHandlerTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/jest/GraylogJestRetryHandlerTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+/**
+ * This file is part of Graylog.
+ *
+ * Graylog is free software: you can redistribute it and/or modify
```
bernd, line 20: Duplicate license that needs to be removed.

bernd (changes requested):

graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/jest/GraylogJestRetryHandlerTest.java
dennisoelkers: :+1:

bernd (approved):

PR #9868, opened 2021-01-06
Cloud user management integration

ousmaneo:
Server change required to integrate Cloud UI specific user creation form field.

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#654
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
+++ b/graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
@@ -19,6 +19,7 @@ import * as React from 'react';
 import * as Immutable from 'immutable';
 import { useEffect, useState } from 'react';
 import { Formik, Form } from 'formik';
+import { PluginStore } from 'graylog-web-plugin/plugin';
 
 import type { DescriptiveItem } from 'components/common/PaginatedItemOverview';
 import User from 'logic/users/User';
@@ -30,6 +31,7 @@ import { Input } from 'components/bootstrap';
 import { Spinner } from 'components/common';
 import history from 'util/History';
 import Routes from 'routing/Routes';
+import AppConfig from 'util/AppConfig';
 
 import TimezoneFormGroup from './TimezoneFormGroup';
 import TimeoutFormGroup from './TimeoutFormGroup';
@@ -40,10 +42,17 @@ import UsernameFormGroup from './UsernameFormGroup';
 
 import { Headline } from '../../common/Section/SectionComponent';
 
+const isCloud = AppConfig.isCloud();
+
 const _onSubmit = (formData, roles, setSubmitError) => {
   const data = { ...formData, roles: roles.toJS(), permissions: [] };
   delete data.password_repeat;
-  data.username = data.username.trim();
+
+  if (isCloud) {
+    data.email = data.username;
+  } else {
+    data.username = data.username.trim();
+  }
 
   setSubmitError(null);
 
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
```diff
@@ -88,10 +106,47 @@ const UserCreate = () => {
...
+
   if (!users) {
     return <Spinner />;
   }
 
+  const showSubmitError = (errors) => {
+    if (isCloud) {
+      const oktaError = JSON.parse(errors?.additional?.res?.text);
+
+      return oktaError && oktaError?.message;
```
edmundoa, line 144: I think this logic should live in the plugin instead of here.

graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
```diff
@@ -40,10 +42,17 @@ import UsernameFormGroup from './UsernameFormGroup';
...
 
+const isCloud = AppConfig.isCloud();
+
 const _onSubmit = (formData, roles, setSubmitError) => {
   const data = { ...formData, roles: roles.toJS(), permissions: [] };
   delete data.password_repeat;
-  data.username = data.username.trim();
+
+  if (isCloud) {
+    data.email = data.username;
```
edmundoa, line 52: I think this logic should live in the plugin.

graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
edmundoa: Looking at the error message, I also wonder if that error should really be exposed in here.


Toastr notification:
error

Error in form:
error2

I think we should try to extract the Okta API message, since that is what I would expect to see. Looking at the response this may be possible in the server, but I think it's worth exploring it. Do you think that would be possible @danotorrey?

edmundoa (changes requested):
So far I only tested creating users and that is working fine, well done! Added a few changes inline.

graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
```diff
@@ -40,10 +42,19 @@ import UsernameFormGroup from './UsernameFormGroup';
...
+
 const _onSubmit = (formData, roles, setSubmitError) => {
-  const data = { ...formData, roles: roles.toJS(), permissions: [] };
+  let data = { ...formData, roles: roles.toJS(), permissions: [] };
   delete data.password_repeat;
-  data.username = data.username.trim();
+
+  if (isCloud && cloudPlugin) {
+    const onCreate = cloudPlugin?.[0]?.oktaUserForm.onCreate;
+    data = onCreate(data);
```
edmundoa, line 54: I think using the optional chaining operator this way does not help us: Using the operator only in the first two accesses would mean that if one of them is `null` or `undefined`, the error thrown will just occur on the next level (when accessing `onCreate` in this case).

Since using a cloud instance will also use Okta for authentication and user management, I think it's safe to assume that a cloud instance will also have the cloud plugin installed, and not having it is an error itself. Rendering the regular user form in case the plugin is not there for some reason is also wrong, since it disguises an error that would lead to possible data inconsistencies and other failures when sending data to Okta.

From that perspective, my opinion is that we should assume that a cloud instance will always have the `cloud` plugin installed. Therefore, I think it is fine to change things a bit to make code more readable, starting with this change in line 46:

```diff
- const cloudPlugin = isCloud ? PluginStore.exports('cloud') : null; 
+ const oktaUserForm = isCloud ? PluginStore.exports('cloud')[0].oktaUserForm : null; 
```

That would let us access properties of that interface directly in other places, for instance on this block:
```diff
-  if (isCloud && cloudPlugin) {
-    const onCreate = cloudPlugin?.[0]?.oktaUserForm.onCreate;
-    data = onCreate(data);
+  if (isCloud) {
+    const { onCreate } = oktaUserForm;
+    data = onCreate(data);
```

We could also try to display a proper error message if the plugin is not available, for that I would enclose line 46 in a `try catch` statement and handle that situation to display an error message instead of the form in that case.

edmundoa (changes requested):

graylog2-web-interface/src/components/users/UserCreate/UserCreate.jsx
edmundoa: I created https://github.com/Graylog2/graylog2-server/issues/9884 to document this issue, so we can move on with this.

edmundoa (approved):
Great job! Looks good to me 👍

PR #9871, opened 2021-01-07
SQS support proof of concept

thll:
Proof of concept to support SQS as a message bus implementation.

Prerequisites:
- Create a `standard` SQS queue at: https://console.aws.amazon.com/sqs/ 
- If you don't have one yet, create an [access key](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html?icmpid=docs_iam_console#Using_CreateAccessKey) for your AWS account
- Put the credentials for the access key into `~/.aws/credentials` and set your region in `~/.aws/config` following the instructions [here](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/setup.html#setup-credentials)

**Be aware that performing SQS requests might generate costs.**

TODO:
- [ ] Increase throughput (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-throughput-horizontal-scaling-and-batching.html)
  - [ ] Use asynchronous client. Need to downgrade to AWS SDK 1 to make this work.
  - [ ] Check if we additionally need to fan out operations using multiple threads to increase throughput.
- [ ] Consider self-management of queues. (Create appropriate queue on startup if not yet present)
- [ ] Improve configurability. (e.g. AWS region is hardcoded atm)
- [ ] Metrics.


thll:
This was WIP and is not required anymore. Closing.

PR #9873, opened 2021-01-08
Fix date format specifiers in index mapping templates.

### LINKED ISSUE #9690: new java.time date format specifiers after upgrade to 4.0
After major upgrade the log contains

 `2020-12-03T10:09:04.375+01:00 WARN  [RestClient] request [POST http://127.0.0.1:9200/_msearch?typed_keys=true] returned 1 warnings: [299 Elasticsearch-7.10.0-51e9d6f22758d0374a0f3f5c6e8f3a7997850f96 "'y' year should be replaced with 'u'. Use 'y' for year-of-era. Use new java.time date format specifiers."]
`

## Expected Behavior
If it is for backward compatibility it should be muted in log, if not it should be adopted.

## Current Behavior
Spamming this message in log for all requests that return it.

## Possible Solution
Adopt new format or mute the message and adopt it later.

Upgrade to graylog 4.0 and elasticsearch 7.10

## Context
The issue leads to significant increase of log size.

## Your Environment

* Graylog Version:4.0
* Java Version:1.8.0
* Elasticsearch Version:7.10
* MongoDB Version:4.4
* Operating System:Centos 8

dennisoelkers:
## Motivation and Context

**Note: This should be backported to `4.0`.**

Before this change, the pre-ES7 date format specifiers were reused for ES7 to define the format of date fields in index mapping templates.

Unfortunately, this results in warnings being returned in responses to searches operating on date fields, due to a [breaking change introduced in ES7](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#_joda_based_date_formatters_are_replaced_with_java_ones).

This change is now splitting up the date format specifiers and uses joda syntax for ES6 and `java.time` syntax for ES7. In addition, the overly lax exposition of the date format specifiers from the `Tools` class was locked down to prevent them leaking into new places.

**Question:** When this change is used in existing setups, any usage of date fields in indices created before this change will still trigger a warning. Only indices created after this change will prevent the warning. This could be remedied by writing a migration that updates the index mapping templates in a subsequent PR. I am not sure if the result is worth the effort (and the risk of breaking index mapping templates in the migration), any comments?

Fixes #9690.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
@@ -16,24 +16,6 @@
  */
 package org.graylog2.indexer;
 
-import com.google.common.collect.ImmutableMap;
-
-import static org.graylog2.plugin.Message.FIELD_GL2_MESSAGE_ID;
-
-public class EventsIndexMapping7 extends EventsIndexMapping {
-    @Override
-    protected ImmutableMap<String, Object> fieldProperties() {
-        return map()
-                .putAll(super.fieldProperties())
-                .put(FIELD_GL2_MESSAGE_ID, map()
-                        .put("type", "alias")
-                        .put("path", "id")
-                        .build())
-                .build();
-    }
-
-    @Override
-    protected String dateFormat() {
-        return ConstantsES7.ES_DATE_FORMAT;
-    }
+class Constants {
+    static final String ES_DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
 }
--- a/graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
@@ -17,5 +17,5 @@
 package org.graylog2.indexer;
 
 class Constants {
-    static final String ES_DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
+    static final String ES_DATE_FORMAT = "8yyyy-MM-dd HH:mm:ss.SSS";
 }
--- a/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
@@ -352,9 +352,7 @@ public abstract class GIMMapping extends IndexMapping {
         ));
     }
 
-    protected String dateFormats() {
-        return "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd";
-    }
+    protected abstract String dateFormats();
 
     private Map<String, Object> textField() {
         return typeField("text");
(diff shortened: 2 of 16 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
```diff
@@ -0,0 +1,21 @@
...
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog2.indexer;
+
+class Constants {
+    static final String ES_DATE_FORMAT = "yyyy-MM-dd HH:mm:ss.SSS";
```
bernd, line 20: According to the [breaking changes docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#_joda_based_date_formatters_are_replaced_with_java_ones) for ES 7, we can use the `8` prefix for date patterns in ES 6. (supported since 6.7) That way users don't get warnings if they create new indices with ES 6.

We requite at least ES 6.8 to run Graylog now, correct?

bernd (changes requested):
> **Question:** When this change is used in existing setups, any usage of date fields in indices created before this change will still trigger a warning. Only indices created after this change will prevent the warning. This could be remedied by writing a migration that updates the index mapping templates in a subsequent PR. I am not sure if the result is worth the effort (and the risk of breaking index mapping templates in the migration), any comments?

@dennisoelkers I reproduced the deprecation warnings. They only show up for queries that use indices that have been created with ES 6. If queries only use indices created with ES 7, there are no warnings. Even if the year format is `yyyy`.

**UPDATE:** Forgot about the migration question. :smile: 

I don't think we need a migration for the mapping templates. We update the index template anyway before creating a new index.

For users with old indices, I think we should document a workaround for changing the log level for the ES 7 `RestClient` class.

graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
```diff
@@ -348,10 +348,14 @@
     private Map<String, Object> dateField() {
         return merge(typeField("date"), Collections.singletonMap(
                 "format",
-                "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd||basic_date_time||basic_date_time_no_millis||epoch_second||date_time_no_millis||date_hour_minute_second_fraction||epoch_millis"
+                dateFormats() + "||basic_date_time||basic_date_time_no_millis||epoch_second||date_time_no_millis||date_hour_minute_second_fraction||epoch_millis"
         ));
     }
 
+    protected String dateFormats() {
+        return "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd";
```
bernd, line 356: Do we need to add the `8` prefix here as well? Or is it not supported in this format?

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/indexer/Constants.java
dennisoelkers: Correct!

graylog2-server/src/main/java/org/graylog2/indexer/GIMMapping.java
dennisoelkers: Yes, good catch. Added that and tested it with a GIM index on ES6. :+1:

bernd (approved):

PR #9875, opened 2021-01-08
Fix page size change selector in Content Pack list

### LINKED ISSUE #9629: Content Packs Pagination Is Broken

1. Go to System > Content Packs
2. Select anything other than 10, for example 25 pages
3. The "Something went wrong." error page appears with the following error : 
`getPaginationModel(): totalPages should be a number`

## Your Environment

* Graylog Version: 4.0.1
* Operating System: Ubuntu 20.04.1

edmundoa:
The `PageSizeSelect` component passes an event as argument to its `onChange` function but but we were expecting a string containing a number. This change fixes that by extracting the value from the received event.

Fixes #9629


ousmaneo (approved):
Tested and works!
Thank you @edmundoa  for this fix

PR #9877, opened 2021-01-11
Fix pattern used to identify parameters in search/dashboard query strings.

### LINKED ISSUE #9497: Can't "Add to query" or "Remove from results" any items with $ sign at the end
## Expected Behavior
When using aggregation "Show top values" for Windows Domain Controller logs there will be plenty of entries like "TargetUserName" that end with $. For example if a desktop's hostname is "desktop123" it will show up as "desktop123$".
I would expect to filter these items with "Add to query" or "Remove from results".

## Current Behavior
When clicking "Add to query" or "Remove from results" for an entry that ends with a $ sign it won't filter anymore.
instead a message is displayed in a reddish box below the query:

Undeclared query parameters: 
All parameters used in the search query need to be declared before the query can be used:

See above.

## Context
This happens a lot with filtering Windows Domain Controller logs.

## Your Environment
Graylog 3.3.8+e223f85
(Private Build 1.8.0_272 on Linux 5.4.0-52-generic)

dennisoelkers:
## Motivation and Context

**Note:** This PR needs to be backported to `3.3` and `4.0`.

Before this change, a user could write or generate a query string that contains two `$` characters in unrelated places (e.g. because a field name or value contains it), leading to an undeclared parameter being incorrectly identified. An example for this is a query string of

```
foo:bar$ AND some:$value
```

This PR is fixing the regex used to identify parameters in query strings. The regex is now checking for two `$` enclosing a group of alphanumerical (`a-z`, `A-Z`, `0-9` & `_`) characters instead of a group of _any_ characters. This also matches the description in the frontend, which was already showing this specification although it was not enforced like this in the backend.

This PR (for `master`) also includes a refactoring that extracts a `QueryParser` class in order to consolidate backend(ES6/ES7)-agnostic code that was previously duplicated for both storage modules. That refactoring should be left out for the backports.

Fixes #9497.

## Types of changes

## Checklist:

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#1999

### DIFF
--- a/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/views/ElasticsearchBackend.java
+++ b/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/views/ElasticsearchBackend.java
@@ -16,32 +16,21 @@
  */
 package org.graylog.storage.elasticsearch6.views;
 
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
-import com.google.common.graph.Traverser;
 import com.google.inject.name.Named;
 import io.searchbox.client.JestClient;
 import io.searchbox.core.MultiSearch;
 import io.searchbox.core.MultiSearchResult;
 import io.searchbox.core.Search;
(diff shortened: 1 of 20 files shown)


kmerz (approved):
LGTM :+1: 

PR #9882, opened 2021-01-11
Do not set invalid widget id as a focused widget

### LINKED ISSUE #9881: Dashboard URLs which contain a query param for an invalid focused widget can result in an unresponsive page.
## Description
This bug only exists on the current master branch. We recently implemented the possibility to focus a widget. Focusing a widget will add its id as a query param to the URL.

This can result in a broken / unresponsive page  in some cases:
- Create a new dashboard (but do not save it), add a widget, focus the widget, reload the page.
- On a dashboard with e.g. two pages, focus a widget on the second dashboard page and reload the application.
- On a dashboard with e.g. two pages, focus a widget on the first page, open the second page.
- On a dashboard or saved search focus a widget, copy the link, remove the widget, save the dashboard or saved search, open the copied link.
dashboard

kmerz:
## Motivation
Prior to this change, when the URL contained a widget id which was not
part of the current widget set an error occured.

## Description
This change will check if the widget id of the URL belongs to an
existing widget before setting the focusedWidget.

## How Has This Been Tested?
- Open Search
- Focus widget and change id in the URL
- Reload page

## Types of changes

Fixes #9881

### DIFF
--- a/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
+++ b/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
@@ -18,8 +18,9 @@ import * as React from 'react';
 import { useContext } from 'react';
 import { render, screen, fireEvent, waitFor } from 'wrappedTestingLibrary';
 import { useLocation } from 'react-router-dom';
-import { useStore } from 'stores/connect';
+import { asMock } from 'helpers/mocking';
 
+import { useStore } from 'stores/connect';
 import WidgetFocusProvider from 'views/components/contexts/WidgetFocusProvider';
 import WidgetFocusContext from 'views/components/contexts/WidgetFocusContext';
 
@@ -83,8 +84,7 @@ describe('WidgetFocusProvider', () => {
   });
 
   it('should not set focused widget from url if the widget does not exist', async () => {
-    // @ts-ignore this is already mocked above.
-    useStore.mockReturnValue({
+    asMock(useStore).mockReturnValue({
       has: jest.fn(() => false),
     });
 
--- a/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
+++ b/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
@@ -15,14 +15,16 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
+import { Map } from 'immutable';
 import { useContext } from 'react';
 import { render, screen, fireEvent, waitFor } from 'wrappedTestingLibrary';
 import { useLocation } from 'react-router-dom';
 import { asMock } from 'helpers/mocking';
 
-import { useStore } from 'stores/connect';
+import { WidgetStore } from 'views/stores/WidgetStore';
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
```diff
@@ -80,4 +81,23 @@ describe('WidgetFocusProvider', () => {
     renderSUT();
     await screen.findByText('clack');
   });
+
+  it('should not set focused widget from url if the widget does not exist', async () => {
+    // @ts-ignore this is already mocked above.
```
dennisoelkers, line 86: You can use `asMock` in the next line to prevent this.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
```diff
@@ -80,4 +81,23 @@ describe('WidgetFocusProvider', () => {
...
+    useLocation.mockReturnValue({
+      pathname: '',
+      search: 'focused=clack',
+    });
+
+    renderSUT();
+    const clack = screen.findByText('clack');
+
+    waitFor(() => {
+      expect(clack).toBeEmptyDOMElement();
```
dennisoelkers, line 100: Using `waitFor` here will not work. `clack` will always be the same, so it will either work the first time or never.

What are you testing here? If you want to check if focused widget id context does not have a widget id set, you could do:

```
await findByText('undefined');
```

dennisoelkers (changes requested):
Added some remarks inline.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
kmerz: No, I needed to use `waitFor` here, since clack will never be undefined but a Promise with an empty Object. Only when I do a `waitFor` It will return as an `EmptyDomElement` :man_shrugging:

dennisoelkers: The only reason why this works is that the test does not wait for `waitFor` to finish (because of a missing `await` or `return`). This way, `waitFor` generates a promise that will be executed in the background, but the test will just finish before it was resolved.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.test.tsx
```diff
@@ -80,4 +82,18 @@ describe('WidgetFocusProvider', () => {
     renderSUT();
     await screen.findByText('clack');
   });
+
+  it('should not set focused widget from url if the widget does not exist', async () => {
+    asMock(useStore).mockReturnValue({
```
dennisoelkers, line 87: I think we should mock `WidgetStore` instead of `useStore` here. Imagine the following:

 - Instead of `useStore`, the SUT uses something different.
 - The SUT uses a second store with `useStore`.

In both cases the test would break, although the component/functionality works fine.

dennisoelkers (changes requested):
Looks almost perfect now and works fine. Just a last remark ...

dennisoelkers (approved):
Looks good, works fine! :tada:

PR #9893, opened 2021-01-12
DateTimePicker : DateTimeProvider to pass static form values

kyleknighted:
Adds the `DateTimeProvider` and Context to allow for limitDuration, and other potential/future static values, to be used throughout the DateTimePicker


linuspahl (approved):
Looks good to me!

PR #9896, opened 2021-01-13
Support relative time range "until" value in date time picker.

linuspahl:
**This PR is based on https://github.com/Graylog2/graylog2-server/pull/9899 which needs to be merged first**

## Description

With this PR we are enabling the "Until" field for the relative tab in the date time picker, this allows defining e.g. the following time range:

`From: 5 minutes ago until 1 minute ago`

There is one failing test, which also fails on `feature/DateTimeRange`.


linuspahl:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/9946

PR #9901, opened 2021-01-13
DateTimePicker: moving live update

kyleknighted:
Reverting location of Live Update component 
image


linuspahl (approved):
Looks good to me!

PR #9911, opened 2021-01-14
Do not allow selecting same pipeline stage twice

### LINKED ISSUE #6806: Changing a pipeline stage to an existing one deletes that stage
In a pipeline rule, if you change the stage number to one which already exists then the rule you change vanishes.

## Expected Behavior
Manually changing the rule stage to an existing one should either 
1) Replace the stage number, shifting the previous stage down(and all subsequent stages down) 
or
2) Merge the stages together

## Current Behavior
The overlapping stage and all associated rules are dropped/vanished.

## Possible Solutions
Raise an error on "Already existing stage"
Allow overwrite with confirmation
Use visual representation to re-arrange stage orders
Merge existing stage rules and newly-assigned stage rules.
Prompt the user for preferred action

1. Create pipeline stages 1, 2, 3. Assign rules to them.
2. Edit stage 3, change the stage priority to 2.
3. Stage 3 "vanishes"

## Context
I was starting to tweak and re-order some of my pipeline stages.

* Graylog Version: 3.1.2

edmundoa:
Before this change, it was possible to add or edit a pipeline stage using the same number as an existing one, resulting in one of the stages silently disappearing. This PR adds validation for the stage and displays an error indicating the situation where the number selected is already used.

While working on the fix, I realised most pipeline components are in need of some care, so I started modernizing those I wanted to edit to fix the issue, but I ended up doing that for the whole folder. Changes there include moving files to Typescript, adding some type definitions, replacing `createReactClass` and class components with functional components, and switching to styled-components. I know this obscures the fix and makes it harder to check the PR, but I think we will still get some good value out of it. From my perspective it's fine checking if the pages work overall, and if [the fix](https://github.com/Graylog2/graylog2-server/commit/415f3a1f78389a2e73134678cdabaf34ec1fd94c) is working as expected.

Since many of the changes are exclusively for master, I will prepare a basic fix that we can backport to 4.0.

Fixes #6806


edmundoa:
It seems that my linter was not checking all rules I thought it was using, will take another look.

graylog2-web-interface/src/components/pipelines/PipelineDetails.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  return (
+    <div>
+      <Row>
+        <Col md={12}>
+          <div className="pull-right">
+            <PipelineForm pipeline={pipeline} save={onChange} />
+          </div>
+          <h2>Details</h2>
+          <PipelineDl className="dl-horizontal" style={{ marginTop: 10 }}>
```
ousmaneo, line 67: I think this style can go to the styled-component

ousmaneo (changes requested):
Thank you for working on this @edmundoa.
I tested and reviewed the fix and it works great. And had a look at the modernization changes and it looks good too. Only one small thing inline.

ousmaneo (approved):
Legit

PR #9916, opened 2021-01-15
Do not allow selecting same pipeline stage twice

### LINKED ISSUE #6806: Changing a pipeline stage to an existing one deletes that stage

edmundoa:
Check if selected stage number already exists in pipeline, displaying an error in that case and not allowing to submit the form. This will prevent users of accidentally overriding one of their existing stages.

Additionally improve the initial suggestion for new stages. Before we always displayed `0` when creating a stage, now we base that value in the existing stages for the pipeline.

Fixes #6806

Backport of #9911 for 4.0

### DIFF
--- a/graylog2-web-interface/src/components/pipelines/StageForm.jsx
+++ b/graylog2-web-interface/src/components/pipelines/StageForm.jsx
@@ -88,6 +88,13 @@ const StageForm = createReactClass({
     this.modal.close();
   },
 
+  _isOverridingStage() {
+    const { pipeline } = this.props;
+    const { initialStageNumber, stage } = this.state;
+
+    return (stage.stage !== initialStageNumber && pipeline.stages.some(({ stage: s }) => s === stage.stage));
+  },
+
   _saved() {
     this._closeModal();
   },
@@ -96,7 +103,9 @@ const StageForm = createReactClass({
     const { stage } = this.state;
     const { save } = this.props;
 
-    save(stage, this._saved);
+    if (!this._isOverridingStage()) {
+      save(stage, this._saved);
+    }
   },
 
   _getFormattedOptions(rules) {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/pipelines/StageForm.jsx
```diff
@@ -93,11 +90,6 @@ const StageForm = createReactClass({
 
   _saved() {
```
ousmaneo, line 91: For the saving if I override a stage number I'm still able to save. there is small part of the code is master that dealt with checking `isOverridingStage` that is missing here.

ousmaneo (changes requested):
Hi thank you for the back port, It works apart from a small thing below.

graylog2-web-interface/src/components/pipelines/StageForm.jsx
edmundoa: Good catch! I will fix it 👍

ousmaneo (approved):
Legit

PR #9918, opened 2021-01-15
DateTimePicker limits for Absolute & Keyword

kyleknighted:
Fixes some tests and updates validation to verify Absolute and Keyword ranges do not go outside scope of Admin's query limits.

Updated colors for easier readability in Noir theme

Reduce width of Popover 

Update permission of Configuration link for Relative ranges

image
image
image

image


linuspahl (approved):
Looks good to me!

Issue #9854, opened 2020-12-29
License Expiration Warning breaks Layout


lennartkoopmann:
See screenshot:

![Capture](https://user-images.githubusercontent.com/35022/103300604-d885d800-49c4-11eb-9bda-d89dfdf62c87.PNG)

* I did not test this on 4.x
* I did not test this with other notification types

## Your Environment

* Graylog Version: 3.3.9-SNAPSHOT+7889cad
* Browser version: Firefox 84.0.1

PR #9921, opened 2021-01-18
Fixing global notification placement on search pages

linuspahl:
**Please test and merge this PR together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/2010.**

## Description
This PR fixes the global notification placement on search pages.

Before:
image

After:
image

The main challenge was to define a page content layout which (by default) displays the page content in a scrollable container, but also allows displaying a child with a dynamic height of 100% (full height of the page content container).
The required changes result in a different footer placement. By default the footer is now attached to the bottom of the page. If the page content height is larger than the page content container, the footer will still be displayed below the content.

There are two cases we define calculated height: `height: calc(100vh - 50px);`. They will be replaced in a future app layout refactoring.

I've tested these changes with Chrome, Safari and Firefox.

Fixes: https://github.com/Graylog2/graylog2-server/issues/9854

## Types of changes


graylog2-web-interface/src/components/notifications/GlobalAppNotifications.tsx
```diff
@@ -0,0 +1,48 @@
+/*
```
linuspahl, line 1: Even though I used `git mv` before I applied the component changes, the reference got lost. If you think it is worth the effort I can give it another try.

kmerz: I do not think this will do much since we do a squash later on and git can keep so much..

PR #9924, opened 2021-01-18
Fix page size change selector in Content Pack list (4.0)

linuspahl:
This is a backport of https://github.com/Graylog2/graylog2-server/pull/9875 for 4.0.

The `PageSizeSelect` component passes an event as argument to its `onChange` function but but we were expecting a string containing a number. This change fixes that by extracting the value from the received event.

Refs #9629


edmundoa (approved):
Looks good to me 👍

PR #9937, opened 2021-01-19
DateTimePicker: Validate on Mount

kyleknighted:
Validates the `limitDuration` with the rendered Time to verify that the User still has access to those ranges before allowing the Search in the UI.

image
image
image


graylog2-web-interface/src/views/components/SearchBar.tsx
```diff
@@ -147,7 +172,7 @@ const SearchBar = ({
                                          title="Search query syntax documentation"
                                          text={<Icon name="lightbulb" />} />
                     </div>
-                    <SearchButton disabled={disableSearch || isSubmitting || !isValid}
+                    <SearchButton disabled={disableSearch || isSubmitting || !isValid || exceedsDuration(limitDuration, timerange)}
```
linuspahl, line 175: You can use `isOverLimit` here as well.

linuspahl (changes requested):
I really like the idea of validating the limit on mount.

Some thoughts and findings:
- When I have a limit of 30 minutes and I open a search URL which defines a relative range of 50 minutes, I can open the date time picker and go to the absolute tab. in this case the time range is being migrated to a relative one correctly, but the validation message is not being displayed. When I tested this case it was sometimes not possible to set an absolute time range.
- Did you thought about using the `TimeRangeDropdown` form validation (for `nextTimeRange`) for the `SearchBarForm` (`timerange`)? I do not see a disadvantage right now.

linuspahl (commented):
Looks good to me! I just saw that if there is an error with the time range in any related form, we are always setting the error for the `nextTimeRange` field, even though e.g. the `SearchBarForm` does not have this field. Technically it neither has a `timerange` `Field`. But I still prefer to set the error for the field which is invalid. There is no direct benefit, but this change will may help us in the future.
We could add a `fieldName` parameter for `dateTimeValidate` to achieve this.

linuspahl (approved):
Looks good to me!

PR #9942, opened 2021-01-20
DateTimePicker: Validate Keyword

kyleknighted:
Disable `Apply` button until keyword is fully parsed
Fix issue where blank keyword was being sent to API


linuspahl:
Looks good to me!

PR #9950, opened 2021-01-21
 Add test for TimeRangeDropdown

ousmaneo:
Add TimeRangeDropdown test

## Checklist:


kyleknighted (approved):
LGTM but I made some changes to get aysnc to stop timing out. @linuspahl Could you take one more look to verify all is good?

linuspahl (approved):
Looks good to me, I saw that the `Absolute tab has Accordion` test needs like 7 seconds to complete. I can imagine that this test will timeout sometimes, but it does not seem to be a problem currently and I could not find a way to improve it.

PR #9951, opened 2021-01-21
DateTimePicker: Converting time to user's timezone

kyleknighted:
## ⚠️ #9949 Need to merge first

Updating uses of `moment()` and replaced with `DateTime.now()` so it uses the proper timezone.


linuspahl (commented):
Looks good to me! I will approve this PR when it can be merged.

linuspahl (approved):
Looks good to me, could not find any errors. I tests all time range types with different users (with and without a configured time zone).

PR #9961, opened 2021-01-25
Fix waiting for recovery for red indices.

dennisoelkers:
## Motivation and Context

When one or more ES nodes are in flood stage, cycling the deflector fails due to not being able to create a new index and toggling the deflector alias. During this step, the code waits for a new index to be created and healthy. Unfortunately, the `waitForRecovery` method used internally does not deal properly with indices being red. The response we are receiving when querying the health of a red index is set to unsuccessful, triggering our abstracted error handling mechanism which is throwing an exception instead of passing through the `red` health state.

This change is now using a different, less safe way of executing the health query request and returning the response as is. This allows waiting for recovery on red indices and timing out instead of throwing an exception.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java
+++ b/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java
@@ -362,7 +362,12 @@ public class IndicesAdapterES6 implements IndicesAdapter {
 
     @Override
     public HealthStatus waitForRecovery(String index) {
-        final Health.Status status = waitForStatus(index, Health.Status.YELLOW);
+        return waitForRecovery(index, 30);
+    }
+
+    @Override
+    public HealthStatus waitForRecovery(String index, int timeout) {
+        final Health.Status status = waitForStatus(index, Health.Status.YELLOW, timeout);
         return mapHealthStatus(status);
     }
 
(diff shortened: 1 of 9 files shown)


graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java
```diff
@@ -362,25 +362,35 @@ public IndexRangeStats indexRangeStatsOfIndex(String index) {
...
 
-    private Health.Status waitForStatus(String index, @SuppressWarnings("SameParameterValue") Health.Status clusterHealthStatus) {
+    private Health.Status waitForStatus(String index, @SuppressWarnings("SameParameterValue") Health.Status clusterHealthStatus, int timeout) {
         final Health request = new Health.Builder()
                 .addIndex(index)
                 .waitForStatus(clusterHealthStatus)
+                .timeout(timeout)
                 .build();
-        final JestResult jestResult = JestUtils.execute(jestClient, request, () -> "Couldn't read health status for index " + index);
+        final JestResult jestResult = JestUtils.executeUnsafe(jestClient, null, request, () -> "Couldn't read health status for index " + index);
```
bernd, line 384: Hmm, I think this is an issue because most consumers of `Indices#waitForRecovery(String)` expect the method to throw an error and don't use the return value. We would need to adjust all consumers to check the return value and abort or retry their execution flow if an index is red or the timeout kicks in. 

image

bernd (changes requested):

bernd (approved):

PR #9965, opened 2021-01-25
Submitting `TimeRangeDropDown` form on enter.

linuspahl:
## Description
With this change we are creating a `KeyCapture` component and implementing it for the `TimeRangeDropDown`.
`KeyCapture` allows to define functions for keys which are getting executed when the key gets pressed.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/KeyCapture.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const KeyCapture = ({ children, keys } : Props) => {
+  useEffect(() => {
+    Object.entries(keys).forEach(([key, onKeyPress]) => {
+      Mousetrap.bind(key, onKeyPress);
+    });
+
+    return () => {
+      Mousetrap.reset();
```
kyleknighted, line 33: > The reset method will remove anything you have bound to mousetrap.

This concerns me. What happens when we have multiple instances of `KeyCapture` used in any given render? Will this reset all different instances or do it know to only reset the existing ones within this single instance? 

Maybe we could extract the similar `bind` loop and allow for unbind of the same keys instead of reseting all keys?

graylog2-web-interface/src/components/common/KeyCapture.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import React from 'react';
+import { render, fireEvent, screen } from 'wrappedTestingLibrary';
+
+import KeyCapture from './KeyCapture';
+
+describe('<KeyCapture />', () => {
```
kyleknighted, line 22: Do we need a test to verify that on unmount the key is no longer bound to the function?

kyleknighted (changes requested):

graylog2-web-interface/src/components/common/KeyCapture.tsx
linuspahl: Good point, we still have to reset all bindings for a key (https://github.com/ccampbell/mousetrap/issues/280) But it is still definitely better than resetting all bindings for all keys.

kyleknighted (approved):
looks great!

PR #9971, opened 2021-01-26
Retrying bulk indexing if target alias has invalid target(s).

### LINKED ISSUE #9955: Graylog failes silent if Elastic in Cluster Block Exception [FORBIDDEN/12/index read-only / allow delete (api)] and index gets rotated - messages are lost
Graylog failes silent in GUI and drops messages if elastic is in [FORBIDDEN/12/index read-only / allow delete (api)] and the current index gets cycled 

## Expected Behavior
Fail with message in Graylog GUI, Log messages as dropped or do not drop messages. 

## Current Behavior
GUI gives the feedback, as if everything works well, In/Out messages are displayed, in Journal no high amound of unprocessed messages present. 
In System Overview no failed index attempts in last 24h get displayed, but no messages are indexed. 
Stacktraces are thrown in logfile:
```
Couldn't point deflector to a new index
org.graylog2.indexer.ElasticsearchException: Couldn't remove alias log_deflector from indices [log_28]

blocked by: [FORBIDDEN/12/index read-only / allow delete (api)];
	at org.graylog2.indexer.cluster.jest.JestUtils.specificException(JestUtils.java:110) ~[graylog.jar:?]
	at org.graylog2.indexer.cluster.jest.JestUtils.execute(JestUtils.java:60) ~[graylog.jar:?]
	at org.graylog2.indexer.cluster.jest.JestUtils.execute(JestUtils.java:65) ~[graylog.jar:?]
	at org.graylog2.indexer.indices.Indices.removeAliases(Indices.java:661) ~[graylog.jar:?]
	at org.graylog2.indexer.MongoIndexSet.cleanupAliases(MongoIndexSet.java:352) ~[graylog.jar:?]
	at org.graylog2.periodical.IndexRotationThread.checkAndRepair(IndexRotationThread.java:149) ~[graylog.jar:?]
	at org.graylog2.periodical.IndexRotationThread.lambda$doRun$0(IndexRotationThread.java:76) ~[graylog.jar:?]
	at java.lang.Iterable.forEach(Iterable.java:75) [?:?]
	at org.graylog2.periodical.IndexRotationThread.doRun(IndexRotationThread.java:73) [graylog.jar:?]
	at org.graylog2.plugin.periodical.Periodical.run(Periodical.java:77) [graylog.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:834) [?:?]
```

```
Couldn't get active write index
org.graylog2.indexer.indices.TooManyAliasesException: More than one index in deflector alias: [log_76, log_28]
	at org.graylog2.indexer.indices.Indices.aliasTarget(Indices.java:355) ~[graylog.jar:?]
	at org.graylog2.indexer.MongoIndexSet.getActiveWriteIndex(MongoIndexSet.java:204) ~[graylog.jar:?]
	at org.graylog2.indexer.fieldtypes.IndexFieldTypePollerPeriodical.lambda$schedule$4(IndexFieldTypePollerPeriodical.java:201) ~[graylog.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:834) [?:?]
```

[log_76, log_28] difference in number comes from cycling after time and not being alerted by this error, so that we have quite a time till someone missed their loglines. 

## Possible Solution
Keep messages in internal journal, and give system message in GUI that things are messed up. 

1. Set current write index to read only/ cluster block 
2. Rotate active index in Graylog GUI/Rest API
3. Described issue should occure
4.

## Context
Messages are lost silently and systems behaves as if everything works fine. 

## Your Environment

* Graylog Version: 3.3
* Java Version: 11
* Elasticsearch Version: 6.8
* MongoDB Version: 4.2
* Operating System: RHEL
* Browser version: doesn't matter

### LINKED ISSUE #9962: Indexing fails silently if target index is an alias with none or multiple targets
## Expected Behavior

If indexing fails because the target alias points towards none or multiple targets, it should retry indexing and keep the messages in the journal until indexing succeeds.

## Current Behavior

If indexing fails because the target alias points towards none or multiple targets, it silently drops the messages.

1. Toggle Flood Stage on ES
2. Trigger deflector cycle

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

Prior to this PR, if one of the write targets in a bulk indexing request is an alias with zero or multiple targets, indexing failed and messages were silently dropped. This results in an error type being returned from ES different to the ones we handled before. Before this we had:

  - requests failing completely: due to general errors (e.g. networking errors), indicated by an `IOException` being thrown and no response being returned
  - requests failing partially (one or all individual bulk items failing): due to specific errors (e.g. mapping errors), with no exception being thrown and returning a response which contains error states for individual bulk items

In this case, we have a different scenario which also differs between the ES version involved:

  - for ES6, no exception is being thrown, but the response returned does not contain failed items
  - for ES7, a generic `ElasticsearchException` thrown which is not retried

This PR is doing these things to address this:

  - parsing the result/the exception and throwing a specific `InvalidWriteTargetException` if the request fails generally and
    contains the related error message
  - make sure that the correct target index is extracted from the error
  - make the log format of the exception helpful and consistent between ES6 and ES7

and last but not least:

  - retry an `InvalidWriteTargetException` similar to an `IOException`

**Attention:** This changes the behavior of cases where a deflector points to zero or multiple targets. In these cases, indexing is retried and **blocks all further indexing until this scenario is resolved.** All messages coming in at this condition are buffered in the journal until either the situation is fixed or the journal flows over. This is similar to the behavior we have implemented for an ES node being in flood stage.

In general, this scenario most probably occurs related to one or more ES nodes having been in flood stage, while an index cycle is happening. In this condition, the deflector alias might get assigned a second target but the previous target cannot be removed due to the index it points to having an index block. When the index block is lifted, the situation is resolved automatically due to our deflector alias cleanup code. This is mostly relevant for ES6, because ES7 releases index blocks due to flood stage kicking in automatically when flood stage is over.

The behavior is undesirable in some situations, e.g. when one deflector is broken (zero or multiple targets) and cannot be repaired, while other deflectors are fine and indexing requests for both need to be processed. Requests including the broken deflector as a target for individual bulk items will be retried over and over while others (which could be processed fine) are starving. Solving this would require us to do retrying out of band (with a separate queue for it) while continuing to process the journal.

Fixes #9955, fixes #9962.
 
## How Has This Been Tested?

To reproduce the situation where this logic is effective, we need these conditions:

  - a deflector alias pointing to two indices
  - one of the indices the deflector alias is pointing at must have an index block, so the cleanup code cannot remove its alias

The easiest way to produce this, is to switch the ES cluster to flood stage. This can be done by `PUT`ing this request to
`/_cluster/settings`:

```
{
	"transient": {
		"cluster.routing.allocation.disk.watermark.low": "1%",
		"cluster.routing.allocation.disk.watermark.high": "1%",
		"cluster.routing.allocation.disk.watermark.flood_stage": "1%"
	}
}
```

Afterwards, a manual deflector cycle should be triggered.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ParsedElasticsearchException.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ParsedElasticsearchException.java
@@ -24,7 +24,7 @@ import java.util.regex.Pattern;
 @AutoValue
 abstract class ParsedElasticsearchException {
     private static final Pattern exceptionPattern = Pattern
-            .compile("ElasticsearchException\\[Elasticsearch exception \\[type=(?<type>[\\w_]+), (?:reason=(?<reason>.+?)(\\]+;|\\]$))");
+            .compile("(ElasticsearchException\\[)?Elasticsearch exception \\[type=(?<type>[\\w_]+), (?:reason=(?<reason>.+?)(\\]+;|\\]$))");
 
     abstract String type();
     abstract String reason();
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -53,6 +53,11 @@ import java.util.stream.Collectors;
 
 @Singleton
 public class Messages {
+    public interface IndexingListener {
+        void onRetry(long attemptNumber);
+        void onSuccess(long delaySinceFirstAttempt);
+    }
+
     private static final Logger LOG = LoggerFactory.getLogger(Messages.class);
 
     private static final Duration MAX_WAIT_TIME = Duration.seconds(30L);
@@ -69,7 +74,7 @@ public class Messages {
     static final WaitStrategy exponentialWaitMilliseconds = WaitStrategies.exponentialWait(MAX_WAIT_TIME.getQuantity(), MAX_WAIT_TIME.getUnit());
 
     @SuppressWarnings("UnstableApiUsage")
-    private static final Retryer<List<IndexingError>> BULK_REQUEST_RETRYER = RetryerBuilder.<List<IndexingError>>newBuilder()
+    private static final RetryerBuilder<List<IndexingError>> BULK_REQUEST_RETRYER_BUILDER = RetryerBuilder.<List<IndexingError>>newBuilder()
             .retryIfException(t -> t instanceof IOException)
             .withWaitStrategy(WaitStrategies.exponentialWait(MAX_WAIT_TIME.getQuantity(), MAX_WAIT_TIME.getUnit()))
             .withRetryListener(new RetryListener() {
@@ -81,8 +86,7 @@ public class Messages {
                         LOG.info("Bulk indexing finally successful (attempt #{}).", attempt.getAttemptNumber());
                     }
                 }
-            })
-            .build();
+            });
 
     private final LinkedBlockingQueue<List<IndexFailure>> indexFailureQueue;
     private final MessagesAdapter messagesAdapter;
@@ -110,10 +114,18 @@ public class Messages {
     }
 
     public List<String> bulkIndex(final List<Map.Entry<IndexSet, Message>> messageList) {
-        return bulkIndex(messageList, false);
+        return bulkIndex(messageList, false, null);
+    }
+
+    public List<String> bulkIndex(final List<Map.Entry<IndexSet, Message>> messageList, IndexingListener indexingListener) {
+        return bulkIndex(messageList, false, indexingListener);
     }
 
     public List<String> bulkIndex(final List<Map.Entry<IndexSet, Message>> messageList, boolean isSystemTraffic) {
+        return bulkIndex(messageList, isSystemTraffic, null);
+    }
+
+    public List<String> bulkIndex(final List<Map.Entry<IndexSet, Message>> messageList, boolean isSystemTraffic, IndexingListener indexingListener) {
         if (messageList.isEmpty()) {
             return Collections.emptyList();
         }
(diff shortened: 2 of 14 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
```diff
@@ -196,7 +213,12 @@ private void waitBeforeRetrying(long attempt) {
         }
     }
 
-    private List<IndexingError> runBulkRequest(List<IndexingRequest> indexingRequestList, int count) {
+    @SuppressWarnings("UnstableApiUsage")
+    private List<IndexingError> runBulkRequest(List<IndexingRequest> indexingRequestList, int count, IndexingListener indexingListener) {
+        final Retryer<List<IndexingError>> BULK_REQUEST_RETRYER = indexingListener == null
+                ? BULK_REQUEST_RETRYER_BUILDER.build()
+                : BULK_REQUEST_RETRYER_BUILDER.withRetryListener(retryListenerFor(indexingListener)).build();
```
bernd, line 220: The `.withRetryListener()` modifies the `BULK_REQUEST_RETRYER_BUILDER` builder instead of returning a completely new builder instance. So every time we have a non-null `indexingListener`, we add another listener instance to the builder.

It's currently only used in tests, but the methods are public so nothing keeps us from using them in the future and then we accumulate objects. :slightly_smiling_face:

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ParsedElasticsearchException.java
```diff
@@ -24,7 +24,7 @@
 @AutoValue
 abstract class ParsedElasticsearchException {
     private static final Pattern exceptionPattern = Pattern
-            .compile("ElasticsearchException\\[Elasticsearch exception \\[type=(?<type>[\\w_]+), (?:reason=(?<reason>.+?)(\\]+;|\\]$))");
+            .compile("(ElasticsearchException\\[)?Elasticsearch exception \\[type=(?<type>[\\w_]+), (?:reason=(?<reason>.+?)(\\]+;|\\]$))");
```
bernd, line 27: Please add a test with an example message for the updated pattern. Thank you!

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
dennisoelkers: Good catch!

dennisoelkers: :heavy_check_mark:

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ParsedElasticsearchException.java
dennisoelkers: :heavy_check_mark:

bernd (approved):

PR #9981, opened 2021-01-28
Use aws sdk version 2 for interacting with SQS

thll:
WIP

Upgrade the AWS SDK to version 2.
We can't use the `AmazonSQSBufferedAsyncClient` anymore but first experiments suggest that we can get comparable throughput with significantly lower CPU load by using the `NettyNioAsyncHttpClient`.

TODO:
- [ ] Provide internal buffering/batching for single message commit calls
- [ ] Revise error handling
- [ ] Prevent hot loops during error conditions
- [ ] Slow down reading when queue is empty
- [ ] Ensure clean shutdown sequence
- [ ] Add metrics for acknowledger

### DIFF
--- a/graylog-project-parent/pom.xml
+++ b/graylog-project-parent/pom.xml
@@ -75,9 +75,9 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
-                <groupId>com.amazonaws</groupId>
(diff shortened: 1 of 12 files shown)


thll:
We will not continue working on an SQS integration. Closing this PR.

PR #9993, opened 2021-01-29
Allow non-string options in Select component.

dennisoelkers:
## Motivation and Context

Prior to this change, the value part of options passed to the `Select` component must be strings. For non-numeric strings, upon selecting an option an exception was thrown due to `.split()` being called on something that does not implement it.

This PR is fixing this, by testing if the selected value is a string or not, avoiding `split` in case it is not.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.tsx
+++ b/graylog2-web-interface/src/components/common/Select.tsx
@@ -473,7 +473,7 @@ class Select extends React.Component<Props, State> {
     if (value && allowCreate) {
       formattedValue = this._formatInputValue(value);
     } else {
-      formattedValue = (value instanceof String
+      formattedValue = (typeof value === 'string'
         ? (value ?? '').split(delimiter)
         : [value])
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/common/Select.tsx
```diff
@@ -473,7 +473,10 @@ class Select extends React.Component<Props, State> {
     if (value && allowCreate) {
       formattedValue = this._formatInputValue(value);
     } else {
-      formattedValue = (value || '').split(delimiter).map((v) => options.find((option) => option[valueKey || ''] === v));
+      formattedValue = (value instanceof String
```
linuspahl, line 476: E.g. the stream select does not display the selected options when I select multiple streams. It looks like the component receives the value as a string literal which is not an instance of `String`. It works with `typeof value === 'string'`.

linuspahl (changes requested):

linuspahl (approved):

PR #9996, opened 2021-02-01
Include all file extensions in lint runs.

dennisoelkers:
During the centralisation of the jest config module (#9548), we removed the `--ext` parameter to eslint in our yarn script definitions.  Unfortunately, this makes eslint to lint `.js`, `.ts` & `.tsx` files only, due to eslint using `js` per default and merging it with `ts`/`tsx` from the `overrides` section of the `eslint-config-graylog` module.

In order to include `jsx` files again, this PR adds an empty `overrides` section including `js` & `jsx` explicitly to the `eslint-config-graylog` module.


linuspahl (approved):
Works as expected.

PR #10000, opened 2021-02-02
Add support for Highlighting for Timestamps and Booleans

### LINKED ISSUE #9835: Search Highlight does not work with Date/Times
In latest core, if you try and highlight a Date/Time string, it does not properly match up and highlight the item. 

Note the `ingest_time` in the attached screenshot. 

image

kmerz:
## Motivation
Prior this PR we did not support highlighting for Timestamps and booleans.

## Description
- Use `render` component for booleans in Typespecific value so the highlighting will be used.
- Same for timestamps, but here we need to pass the render function to the Timestamp component
- For timestamps we need to take care of the highlighting length since the compared string is shorter then
  the rendered timestamp.

## How Has This Been Tested?
- Highlight a timestamp
- Highlight a boolean
- Ensure they are not treated as numbers

## Types of changes

Fixes #9835 

## Caution
#9634 needs to be merged first and then this PR needs a rebase!

### DIFF
--- a/graylog2-web-interface/src/components/common/Timestamp.jsx
+++ b/graylog2-web-interface/src/components/common/Timestamp.jsx
@@ -19,6 +19,22 @@ import React from 'react';
 
 import DateTime from 'logic/datetimes/DateTime';
 
+export const formatDateTime = (dateTime, format, tz, relative = false) => {
+  if (relative) {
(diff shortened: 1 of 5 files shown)


linuspahl (approved):
Looks good to me!

PR #10005, opened 2021-02-02
Remove classpath scanning to improve JerseyService startup speed

bernd:
This reduces the startup time of the `JerseyService` by 83% on my development machine when running the production artifacts.
(3.5s vs 20.5s)

Since scanning package paths for resources is pretty slow in Jersey, we now register all REST resources explicitly via `addSystemRestResource()` in the guice bindings. This will also allow us to disable resources via feature flags and runtime profiles in the future.

I compared the output of the `PrintModelProcessor` with both implementations and we get the same resources registered with the new one. (tested in a production artifact deployment including enterprise plugins)

In addition to the resource changes I also had to adjust the full-backend tests to get rid of a race condition that was shadowed by the slow server startup before.


bernd:
@dennisoelkers @mpfz0r @thll The server startup is now so fast that the full-backend tests failed due to a race condition. :joy: I pushed a fix for that. (see: https://github.com/Graylog2/graylog2-server/pull/10005/commits/832cfd6f2971bd94fbc790b847b4a008990d55fe)

thll:
Code looks good and I didn't find any issues when trying it out 👍 

I would approve the PR but as we've started a discussion about abandoning the scanning and start binding resource classes explicitly, I'll leave it be for now.

bernd:
@dennisoelkers @mpfz0r @thll I updated the PR to replace package scanning with explicit resource bindings.

bernd:
This needs more work. The API browser is broken with this because it's also doing package scanning.

bernd:
@dennisoelkers @mpfz0r @thll Okay, this should be ready for review again. I adjusted the API documentation browser to use the list of resource classes instead of doing package scanning.

graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
```diff
@@ -471,8 +475,22 @@ protected void registerJacksonSubtype(Class<?> klass, String name) {
...
-        final Multibinder<RestControllerPackage> restControllerPackages = Multibinder.newSetBinder(binder(), RestControllerPackage.class);
-        restControllerPackages.addBinding().toInstance(RestControllerPackage.create(packageName));
+    /**
+     * Adds given API resource as a system resource. This should not be used from plugins!
+     * Plugins should use {@link org.graylog2.plugin.PluginModule#addRestResource(Class)} instead to ensure the
+     * addition of the path prefix.
+     *
+     * @param restResourceClass the resource to add
+     */
+    protected void addSystemRestResource(Class<?> restResourceClass) {
```
dennisoelkers, line 485: Would it make sense to extend `Graylog2Module` with something like `CoreModule` that `PluginModule` does not extend from and put `addSystemResource` in there? I am trying to avoid that plugin authors make use of this methods because it's available and e.g. suggested in the IDE.

bernd: Yeah, that's a good point.

`PluginModule` currently extends `Graylog2Module` and lots of modules in the server extend `Graylog2Module`. We also have several modules in the server which extend `PluginModule`.

So we would need to introduce a `CoreModule` that extends `Graylog2Module` and then switch all server modules that extend `PluginModule` over to `CoreModule` to get access to the API resource binding helper. That means we probably have to push down some helpers from `PluginModule` to `Graylog2Module`, depending on what the server modules that currently extend `PluginModule` need.

Correct?

Sounds like a separate PR. :smile:

dennisoelkers: Alright, I am fine with doing that in a separate Pr.

dennisoelkers (approved):

PR #10012, opened 2021-02-03
Creating common component `CopyToClipboardCapture` which allows to format content on copy

linuspahl:
## Description
This PR adds a component called `CopyToClipboardCapture` which allows to format content on copy before it gets stored in the clipboard. The component receives a function (`formatSelection`) which gets called every time the user copies encapsulated content (`children`).

## Motivation and Context
We need this functionality for the following case. When a grid is styled with `display: flex;` or `display: grid;` and a user copies a row, each cell is being separated by a line break. With this component we are able to remove the line breaks when necessary.


kmerz (approved):
LGTM :+1: 

PR #10025, opened 2021-02-08
Fixing gap between loading spinner icon and text

linuspahl:
## Description
This PR reimplements the gap between the text and icon of the loading spinner. It is now defined by a margin instead of the space character.

Before:
 the loading spinner
image

After:
image


kmerz (approved):
LGTM :+1:

PR #10032, opened 2021-02-09
Allow dropdown in responsive table

Gaya:
## Description

Attempts to patch the problem with https://github.com/Graylog2/graylog-plugin-cloud/issues/703

It's not the best solution, but it's a good trade off for not having to spend a long time fixing Bootstrap issues.

## Motivation and Context

This solution makes it possible to use the context menu as you'd expect in full width mode.

In responsive mode it requires the user to scroll inside of the table a bit.

## How Has This Been Tested?
In Firefox and on mobile

Normal:
image

Responsive:
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/graylog/DropdownButton.css
+++ b/graylog2-web-interface/src/components/graylog/DropdownButton.css
@@ -1,0 +1,10 @@
+@media (max-width: 992px) {
+    .table-responsive .dropdown-menu {
+        position: static !important;
+    }
+}
+@media (min-width: 993px) {
+    .table-responsive {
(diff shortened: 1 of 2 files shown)


kyleknighted:
Can you just toss that in the styled-component instead of the CSS file?

Gaya:
@kyleknighted not really as it's scoped to the DropdownButton alone. This affects also the responsive table.
Just wanted to only let it happen when you include a dropdown button

Gaya:
Kyle and I have decided to move the solution into the component it was happening in instead of in core.

PR #10049, opened 2021-02-11
Defining generic type definition for `GlobalOverride` search type options.

linuspahl:
## Description
The type definition for the search type options in `GlobalOverride` currently only consider properties for a paginated list.
`{ limit: number, offset: number }`

This PR is adds a type definition for a scroll parameter.
`{ after: [any] }`

## Motivation and Context
We need this new type definition for a search type which uses a scroll position instead of pagination to receive the messages for a specific position in a list.

### DIFF
--- a/graylog2-web-interface/src/views/logic/search/GlobalOverride.ts
+++ b/graylog2-web-interface/src/views/logic/search/GlobalOverride.ts
@@ -18,31 +18,37 @@ import * as Immutable from 'immutable';
 
 import type { QueryString, TimeRange } from '../queries/Query';
 
-export type MessageListOptions = {
-  [searchTypeId: string]: {
-    limit: number;
-    offset: number;
-  };
+export type PaginatedListOptions = {
+  limit: number,
+  offset: number,
+};
+
+export type ScrollListOptions = {
+  after: Array<any>,
+}
+
+export type SearchTypeOptions = {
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/logic/search/GlobalOverride.ts
```diff
@@ -18,31 +18,37 @@ import * as Immutable from 'immutable';
...
+  limit: number,
+  offset: number,
+};
+
+export type ScrollListOptions = {
+  after: Array<any>,
+}
+
+export type SearchTypeOptions = {
+  [searchTypeId: string]: ScrollListOptions | PaginatedListOptions
```
kmerz, line 31: I do not think this is a good solution:
1. we are leaking plugin code here.
2. It does not take in the possibility that there might be other plugin authors which we do not know from.
3. Also the function itself does not care what is inside of `searchTypes`


```suggestion
  [searchTypeId: string]: any
```

kmerz (commented):
Just my thoughts.

graylog2-web-interface/src/views/logic/search/GlobalOverride.ts
linuspahl: That is a good point, I tried to define the type definitions as generic as possible, but I am also fine with using `any` in this case.

kmerz (approved):
Nice :heart:

PR #10057, opened 2021-02-12
Archiving multiple indices at once now creates only one job providing a List of indices to be archived

GoerkemSahin:
If applied, this commit will open one job for every set of indices to be archived instead of opening one job for every single index in the set of indices

This change was made because the creation of jobs for each index in the set of indices was faster than the first job took to finish, resulting in it blocking the next jobs in queue from executing. Now the single job runs through the complete set of indices

https://github.com/Graylog2/graylog-plugin-enterprise/issues/1405

The run() method of ArchiveCreateJob now takes a List of indices which it will iterate through instead of single index

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2047


graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexCountBasedRetentionStrategy.java
```diff
@@ -89,19 +90,29 @@ private void runRetention(IndexSet indexSet, Map<String, Set<String>> deflectorI
...
+                .stream()
+                .skip(orderedIndices.size() - removeCount)
+                // reverse order to archive oldest index first
+                .collect(Collectors.toCollection(LinkedList::new)).descendingIterator().
+                forEachRemaining(indexName -> orderedIndicesDescending.add(indexName));
+
+        String indexNames = "";
+        for(String index : orderedIndicesDescending) {
+            if(!indexNames.isEmpty()){
+                indexNames.concat(",");
```
mpfz0r, line 106: This doesn't work. You'd need to assign the result to the variable again.
However, there is a much simpler way of doing this:
`indexNames = orderedIndicesDescending.stream().collect(Collectors.joining(","));`

graylog2-server/src/test/java/org/graylog2/indexer/retention/strategies/AbstractIndexCountBasedRetentionStrategyTest.java
```diff
@@ -81,7 +82,7 @@ public void setUp() throws Exception {
             }
 
             @Override
-            protected void retain(String indexName, IndexSet indexSet) {
+            protected void retain(List indexName, IndexSet indexSet) {
```
mpfz0r, line 85: Please use a generic type here. IntelliJ also should've marked this as a warning

mpfz0r (changes requested):
I've left a few comments.
Can you please also sign the CLA? Otherwise we cannot merge it.

graylog2-server/src/test/java/org/graylog2/indexer/retention/strategies/AbstractIndexCountBasedRetentionStrategyTest.java
mpfz0r: `List<String> indexNames`

mpfz0r (changes requested):
one generic type missing

mpfz0r (approved):

mpfz0r:
Gonna ignore the frontend build failures :shrug:

PR #10058, opened 2021-02-12
Allow passing timeout multiplier for frontend tests.

dennisoelkers:
## Motivation and Context

Before this change, we were using default timeouts for frontend tests. These are:

  - 5s for each jest test
  - 1s for async wait utils in testing-library

On lower-end machines (e.g. github actions), these timeouts are sometimes too aggressive. Instead of raising timeouts in general, we want to raise timeouts only for specific environments.

Therefore, this PR allows passing a `TIMEOUT_MULTIPLIER` environment variable to the `yarn test` run. It is being parsed as a number (either int or float) and applied to the defaults.

E.g. `TIMEOUT_MULTIPLIER=2 yarn test` results in:

  - 10s for each jest test
  - 2s for async wait utils in testing-library

It also works the other way, `TIMEOUT_MULTIPLIER=0.5 yarn test` results in:

  - 2.5s for each jest test
  - 0.5s for async wait utils in testing-library

Unfortunately, it was not possible to contain this change completely in `jest-preset-graylog`, because:

  - testing-library needs to be configured in each module that uses it, therefore requiring to run a script from `graylog2-web-interface`'s `setupFilesAfterEnv`
  - if `setupFilesAfterEnv` is specified in the parent module, the section from the preset will be ignored, therefore requiring to add `jest-enzyme` to the `setupFilesAfterEnv` section in `graylog2-web-interface` again

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -31,6 +31,10 @@
     "preset": "jest-preset-graylog",
     "setupFiles": [
       "<rootDir>/test/setup-jest.js"
+    ],
+    "setupFilesAfterEnv": [
+      "jest-enzyme",
+      "<rootDir>/test/configure-testing-library.js"
     ]
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


linuspahl (approved):
Looks good to me, works as expected!

PR #10063, opened 2021-02-15
Updates to include dynamic plugin UI settings in generated config.js at startup

waab76:
## Description
Some plugins have dynamic UI settings data that needs to be provided to the UI at startup time via inclusion in the generated config.js file.  This change adds a new `PluginUISettingsProvider` interface and a method for binding instances of that interface such that the data they provide will be included when the config.js file is generated at server startup.

## How Has This Been Tested?
This change has been tested locally using an in-development Enterprise plugin.

Screen Shot 2021-02-16 at 4 26 54 PM

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
@@ -52,6 +52,7 @@ import org.graylog2.plugin.lookup.LookupDataAdapterConfiguration;
 import org.graylog2.plugin.outputs.MessageOutput;
 import org.graylog2.plugin.security.PasswordAlgorithm;
 import org.graylog2.plugin.security.PluginPermissions;
+import org.graylog2.web.PluginUISettingsProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -463,6 +464,14 @@ public abstract class Graylog2Module extends AbstractModule {
 
     protected Multibinder<ConstraintChecker> constraintCheckerBinder() {
         return Multibinder.newSetBinder(binder(), ConstraintChecker.class);
+    }
+
+    protected Multibinder<PluginUISettingsProvider> pluginUISettingsProviderBinder() {
+        return Multibinder.newSetBinder(binder(), PluginUISettingsProvider.class);
+    }
+
+    protected void addPluginUISettingsProvider(Class<? extends PluginUISettingsProvider> uiSettingsProviderClass) {
+        pluginUISettingsProviderBinder().addBinding().to(uiSettingsProviderClass);
     }
 
     private static class DynamicFeatureType extends TypeLiteral<Class<? extends DynamicFeature>> {}
--- a/graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
@@ -466,14 +466,6 @@ public abstract class Graylog2Module extends AbstractModule {
         return Multibinder.newSetBinder(binder(), ConstraintChecker.class);
     }
 
-    protected Multibinder<PluginUISettingsProvider> pluginUISettingsProviderBinder() {
-        return Multibinder.newSetBinder(binder(), PluginUISettingsProvider.class);
-    }
-
-    protected void addPluginUISettingsProvider(Class<? extends PluginUISettingsProvider> uiSettingsProviderClass) {
-        pluginUISettingsProviderBinder().addBinding().to(uiSettingsProviderClass);
-    }
-
     private static class DynamicFeatureType extends TypeLiteral<Class<? extends DynamicFeature>> {}
 
     private static class ContainerResponseFilterType extends TypeLiteral<Class<? extends ContainerResponseFilter>> {}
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
```diff
@@ -465,6 +466,14 @@ protected void registerJacksonSubtype(Class<?> klass, String name) {
         return Multibinder.newSetBinder(binder(), ConstraintChecker.class);
     }
 
+    protected Multibinder<PluginUISettingsProvider> pluginUISettingsProviderBinder() {
+        return Multibinder.newSetBinder(binder(), PluginUISettingsProvider.class);
+    }
+
+    protected void addPluginUISettingsProvider(Class<? extends PluginUISettingsProvider> uiSettingsProviderClass) {
```
bernd, line 473: We usually add the plugin related module methods into `PluginModule`. Is there a specific reason to put the method into this file?

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -43,14 +48,21 @@
...
                              HttpConfiguration httpConfiguration,
-                             Engine templateEngine) {
+                             Engine templateEngine,
+                             Set<PluginUISettingsProvider> settingsProviders,
+                             ObjectMapper objectMapper) {
         this.configuration = requireNonNull(configuration, "configuration");
         this.httpConfiguration = requireNonNull(httpConfiguration, "httpConfiguration");
         this.templateEngine = requireNonNull(templateEngine, "templateEngine");
+        this.pluginUISettings = requireNonNull(settingsProviders, "settingsProviders").stream()
+                .collect(Collectors.toMap(PluginUISettingsProvider::pluginSettingsKey, PluginUISettingsProvider::pluginSettingsJson));
```
bernd, line 64: What happens if two plugins register with the same `pluginSettingsKey`? Does the collect fail or will it overwrite a previous key?

That said, do we need to think about namespacing for the keys? (e.g. with `PluginMetaData#getUniqueId?`)

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -68,7 +80,16 @@ public String get(@Context HttpHeaders headers) {
...
             "serverUri", baseUri.resolve(HttpConfiguration.PATH_API),
-            "appPathPrefix", baseUri.getPath());
+            "appPathPrefix", baseUri.getPath(),
+            "pluginUISettings", buildPluginUISettings());
         return templateEngine.transform(template, model);
     }
+
+    private String buildPluginUISettings() {
+        try {
+            return objectMapper.writeValueAsString(pluginUISettings);
```
bernd, line 90: This will write the `pluginSettingsJson`, which is a JSON string, as JSON string again and thus escaping it. Is that intended? Should we use a `Map<String, Object>` instead of a JSON string as `PluginUISettingsProvider`?

bernd (changes requested):
I know it's still a draft, but I have some questions. :slightly_smiling_face:

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
dennisoelkers: If we do not want to couple it to the unique id (which is by convention a fqcn, which is rather cryptic), we should at least use a `MapBinder` with the key being the key in the generated config.js. This would prevent silently overwriting the same key and fail at runtime instead if two plugins try to fill the same key.

graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
waab76: It went here mostly due to me not knowing the right place to put it.  I will move it.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
waab76: Need to give some thought to namespacing.  I agree that fqcn is probably too cryptic.  Maybe something similar to Java package naming like <your tld>.<your company/org>.<plugin name>.<specifier> such that the one I'm building for the Theme Customizer might be `org.graylog.themecustomizer.themeoverrides`

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
waab76: Yeah, I will work on a way to make this both more flexible and more serializable

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
waab76: Updated documentation on pluggable interface to include key naming conventions to avoid a collision.

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
waab76: We have to do the JSON string conversion before passing data to the template engine because the template engine will do the conversion properly.

waab76:
Branches seem to be in a bad state.  Will re-create as a new PR.

PR #10070, opened 2021-02-16
Fix race condition pipeline

### LINKED ISSUE #9341: Pipeline Rule Direct Link Race Condition on load
## Expected Behavior

When you link directly to a Pipeline Rule, it should load the form with the proper rule within the Editor

## Current Behavior

Some sort of race condition on load, when the link is loaded, it will on occasion actually load the Rule content, but mostly loads in a blank Editor with no content

1. Create a Pipeline Rule
2. Grab the link to the rule and open in new tab
3. Refresh a few times to see it sometimes load, sometimes not
4. Internal Testing Link that I am able to recreate issue - https://smd-dev.torch.sh/system/pipelines/rules/5f22d4151b43f44a1e7a092c

Gaya:
## Description
This PR fixes #9341

## Motivation and Context
The pipeline rule context and source code editor didn't play well together. In some instances the source code of pipeline rules was not loaded into the source code editor.

Because of the nature of how the context and source code editor were implemented it sometimes didn't work correctly.

It problem lies in the way the rule values are set on the source code editor. After debugging I came to the conclusion that the source code editor was not loaded, but the rule source had and thus wanted to update a non-existing source code editor.

This fix will trigger the context to apply the rule source on the source code editor should it load after the rule has been loaded.

I am using `useState` to get the desired effect. It's a bit weird, but it retriggers the effects inside the Context correctly. If we were to fix the way source code editor and the context work we'd have to refactor them to be controlled inputs instead of uncontrolled.

## How Has This Been Tested?
Was tested using the reproduction steps of #9341

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/rules/RuleContext.jsx
+++ b/graylog2-web-interface/src/components/rules/RuleContext.jsx
@@ -14,7 +14,7 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-import React, { createContext, useEffect, useRef, useCallback } from 'react';
+import React, { createContext, useEffect, useRef, useCallback, useState } from 'react';
 import PropTypes from 'prop-types';
 
 import CombinedProvider from 'injection/CombinedProvider';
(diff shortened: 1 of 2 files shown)


kmerz (approved):
- Code LGTM and I cannot come up with something better. :+1: 
- Tested the Code thoroughly and worked on my machine :heavy_check_mark:

PR #10072, opened 2021-02-16
Set right input IDs and references

edmundoa:
Before this change, components for auto-generated configuration forms did not associate their labels and inputs correctly, setting different values for the label's `for` attribute and the input's `id`. This represents a usability issue, and also throw errors when testing these components with `react-testing-library`'s `getByLabelText` method.

This PR ensures all auto-generated labels and inputs are connected through the same id values.

Refs Graylog2/graylog-plugin-cloud#741

### DIFF
--- a/graylog2-web-interface/src/components/configurationforms/NumberField.jsx
+++ b/graylog2-web-interface/src/components/configurationforms/NumberField.jsx
@@ -69,20 +69,20 @@ const NumberField = createReactClass({
   },
 
   render() {
-    const { typeName } = this.props;
-    const { field } = this.props;
+    const { field, title, typeName } = this.state;
     const isRequired = !field.is_optional;
     const validationSpecs = this.validationSpec(field);
+    const fieldId = `${typeName}-${title}`;
 
     // TODO: replace with bootstrap input component
     return (
       <div className="form-group">
-        <label htmlFor={`${typeName}-${field.title}`}>
+        <label htmlFor={fieldId}>
           {field.human_name}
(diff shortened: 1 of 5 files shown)


edmundoa:
These files are pretty old and would benefit from a refactoring, including fixing linter issues. I think it would be better to do that on master and leave this change as small as it can be, since the main goal of this PR is to support Graylog2/graylog-plugin-cloud#741.

ousmaneo:
Hi, while testing this PR I encountered some error while creating new input 
`NumberField.jsx:71 Uncaught TypeError: Cannot read property 'field' of null
    at Object.NumberField_render [as render] (NumberField.jsx:71)`
This happens when I choose an Input that has a Number field in its form.

edmundoa:
@ousmaneo I'll have a look though I'm not sure if it's related to the changes.

graylog2-web-interface/src/components/configurationforms/NumberField.jsx
```diff
@@ -69,20 +69,20 @@ const NumberField = createReactClass({
   },
 
   render() {
-    const { typeName } = this.props;
-    const { field } = this.props;
+    const { field, title, typeName } = this.state;
```
ousmaneo, line 72: It seems the issue is caused by this line here it should be `this.props` instead of `this.state`.

edmundoa: Fixed it, that was a mistake on my side.

ousmaneo (approved):
Legit !

PR #10073, opened 2021-02-16
Add confirm when navigating away from editing pipeline rule

### LINKED ISSUE #8650: Pipeline Rule Edit page does not warn when navigating away
## Expected Behavior
When creating a new or editing an existing Pipeline rule, warn the user with an "Are you sure? All changes will be lost." yes/no Javascript `confirm` if they attempt to navigate away from the Pipeline Rule Edit page with unsaved changes.

## Current Behavior
When creating a new Pipeline rule, and accidentally clicking the Back button in the browser, the navigation occurs immediately. When clicking Forward again, the content of the rule is lost.

This happened to be, when using the left-right scroll on with my trackpad in the rule body. I scrolled too far left, and it triggered the Back action. Now I am being really careful about that to not accidentally lose my rule.

1. Create a new Pipeline rule. Don't save it.
2. Click the Back button in the browser.
3. Click Forward. The rule content is gone.

## Context

This is a UX-related request. I believe it's a general practice to warn users if they are about to lose unsaved changes.

## Your Environment

* Graylog Version: Graylog 3.3.3-SNAPSHOT

Gaya:
## Description

Fixes #8650

Adds a `<ConfirmLeaveDialog />` to the pipeline source edit page.

## Motivation and Context

To warn users who are editing a pipeline rule that changes can be lost when navigating away.

## How Has This Been Tested?

In browser

## Types of changes

## Checklist:


graylog2-web-interface/src/components/rules/RuleForm.jsx
```diff
@@ -34,19 +34,33 @@ const RuleForm = ({ create }) => {
...
+    setIsDirty(true);
     handleDescription(event.target.value);
   };
 
+  const handleSourceChange = () => {
+    setIsDirty(true);
+  };
+
   const handleCancel = () => {
     history.goBack();
```
edmundoa, line 65: It seems that this is resetting any changes made in the editor:

https://user-images.githubusercontent.com/716185/108244521-97ea4580-714f-11eb-92f4-b12b360c7457.mp4

edmundoa (changes requested):
This is a good improvement, thank you for working on it! Besides the inline issue, the PR also needs a rebase to fix a smaller conflict.

graylog2-web-interface/src/components/rules/RuleForm.jsx
Gaya: Oh no! Seems like the whole Pipeline Context is biting our ass once more 😅

Maybe I need to more this dirty flag in there too. 

Good find!

Gaya:
@edmundoa it was a harder problem to tackle than expected. Had to make the rule source more controlled inside of the rules context and pass the `onChange` binding directly to the component.

All in all, this makes the rules editor a bit more predictable luckily.

Tested it with the different situation where stuff went wrong before (race conditions etc)

edmundoa (approved):
It works and I actually like it better this way 👍

PR #10084, opened 2021-02-17
Make message queue implementations pluggable

thll:
## Description

Creates distinct interfaces for the purpose of "reading from", "writing to" and "acknowledging message receipt on" a message queue.
Makes implementations pluggable by providing means to register additional implementations and activating them in the config file.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise-integrations#487

## Motivation and Context

Forward-port of https://github.com/Graylog2/graylog2-server/pull/10091

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #10091, opened 2021-02-18
Make message queue implementations pluggable (cloud-4.0)

thll:
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise-integrations#480,Graylog2/graylog-plugin-cloud#745

## Description

Creates distinct interfaces for the purpose of "reading from", "writing to" and "acknowledging message receipt on" a message queue.
Makes implementations pluggable by providing means to register additional implementations and activating them in the config file.

## Motivation and Context

We aim to support message queue implementations other than `disk`.
See Graylog2/graylog-plugin-cloud#211

This PR contains most of https://github.com/Graylog2/graylog2-server/pull/10014 with only the (now pluggable) implementations taken out.

## How Has This Been Tested?

Basic manual tests to see if server starts up and only runs the activated implementation.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/plugin/BaseConfiguration.java
```diff
@@ -65,6 +66,9 @@
     @Parameter("message_journal_enabled")
     private boolean messageJournalEnabled = true;
 
+    @Parameter(value = "message_journal_mode", validators = StringNotBlankValidator.class)
```
mpfz0r, line 69: this could be empty if the journal is disabled

thll: You mean someone would specify it like this?
```
message_journal_enabled = false
message_journal_mode =
```
That will fail indeed. Just omitting the option entirely will work:
```
message_journal_enabled = false
```
I'll make it more resilient then to support the first case.

graylog2-server/pom.xml
```diff
@@ -165,6 +165,11 @@
             <groupId>com.fasterxml.jackson.module</groupId>
             <artifactId>jackson-module-jsonSchema</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-scala_2.13</artifactId>
+            <version>2.9.10</version>
```
bernd, line 171: Why do we hard code the version here? I guess that should use our regular `jackson.version` property?

graylog2-server/src/main/java/org/graylog2/bindings/CustomScopes.java
```diff
@@ -0,0 +1,40 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog2.bindings;
+
+import com.google.inject.Key;
+import com.google.inject.Provider;
+import com.google.inject.Scope;
+import com.google.inject.Scopes;
+
+public class CustomScopes {
```
bernd, line 24: I guess this class can go?

graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
```diff
@@ -388,6 +392,22 @@ public void configure(Binder binder) {
         }
     }
 
+    /**
+     * Set up a separate injector, containing only the configuration bindings. It can be used to look up configuration
+     * values in modules at binding time.
+     */
+    protected Injector setupConfigInjector(NamedConfigParametersModule configModule) {
+        try {
+            return Guice.createInjector(Stage.PRODUCTION, Collections.singletonList(configModule));
```
bernd, line 401: How about enabling explicit bindings for the config injector? That way we make sure that we don't depend on anything unwanted. Or is this not possible with the config bindings?

graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
```diff
@@ -193,10 +197,16 @@ public void run() {
         final List<String> arguments = ManagementFactory.getRuntimeMXBean().getInputArguments();
         LOG.info("Running with JVM arguments: {}", Joiner.on(' ').join(arguments));
 
-        injector = setupInjector(configModule, pluginBindings, binder -> binder.bind(ChainingClassLoader.class).toInstance(chainingClassLoader));
+        configInjector = setupConfigInjector(configModule);
+        if (configInjector != null) {
+            final PluginBindings pluginBindings = new PluginBindings(plugins, configInjector);
+            injector = setupInjector(configModule, pluginBindings,
```
bernd, line 203: Is it possible to use the `configInjector` as parent for the regular injector so we avoid duplicating the config instances?

bernd (changes requested):

graylog2-server/pom.xml
thll: @mpfz0r I guess this is a question for you.

graylog2-server/src/main/java/org/graylog2/bindings/CustomScopes.java
thll: yes, thanks for noticing

graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
thll: good idea, i'll try that!

graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
thll: I like the idea and I tried it, but had some trouble with duplicate binding. I don't think it's worth the effort at the moment and we can still change it later.

graylog2-server/pom.xml
mpfz0r: This is a workaround for https://github.com/FasterXML/jackson-bom/pull/38
I'll add a comment.

We cannot use our `jackson.version` because that module hasn't been released:

`Could not find artifact com.fasterxml.jackson.module:jackson-module-scala_2.13:pom:2.9.10.20200411 in sonatype-nexus-releases (https://oss.sonatype.org/content/repositories/releases)`

mpfz0r:
I think this looks good now :+1: 
 Given the size of this PR, I think we should not add more changes to it, but use another branch.

thll:
@bernd I think this PR is now in a state where we can wrap it up. Do you care to have a look at the bits that make the core configuration available to plugins?

We won't merge the PR just yet but the plan is to not add any more changes (except critical fixes of course)

bernd (changes requested):
Looks good so far. I haven't tested this yet, will do so on my test setup once we have a PR for the master branch. :+1: 

Only a single comment:

graylog2-server/src/main/java/org/graylog2/shared/buffers/JournallingMessageHandler.java
```diff
@@ -83,53 +56,26 @@ public void onEvent(RawMessageEvent event, long sequence, boolean endOfBatch) th
...
             // copy to avoid re-running this all the time
-            final List<Journal.Entry> entries = Lists.newArrayList(transform(batch, converter));
+            final Filter metricsFilter = new Filter();
+            final List<RawMessageEvent> entries = batch.stream().map(metricsFilter).filter(Objects::nonNull).collect(Collectors.toList());
 
-            // Remove all null values returned from the converter (might happen if the Converter throws an exception)
-            entries.removeAll(NULL_SINGLETON);
+            // The filter computed the latest receive timestamp of all messages in the batch so we don't have to
+            // call the update on the recorder service for every message. (less contention)
+            processingStatusRecorder.updateIngestReceiveTime(metricsFilter.getLatestReceiveTime());
```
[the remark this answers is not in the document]
mpfz0r: Should I fix this comment as well then?

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/system/processing/ProcessingStatusRecorder.java#L27-L28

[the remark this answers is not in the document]
bernd: @mpfz0r Good point! I guess I got confused by the previous code and made some false assumptions. :smile: 

According to the comment in `ProcessingStatusRecorder` updating the status before writing to the journal is correct.

thll:
The corresponding PR in `master` (https://github.com/Graylog2/graylog2-server/pull/10084) has been approved merged. We should consider this PR now "closed". Subsequent changes need to go into separate PRs which need to be ported to `master` separately.

mpfz0r (approved):

mpfz0r:
@bernd can you approve this change?

bernd (approved):

PR #10092, opened 2021-02-18
Fixing counting of messages in index for ES7.

dennisoelkers:
## Motivation and Context

**Note:** This PR should be backported to `4.0`.

Prior to this change, the `Indices#numberOfMessages` implementation for ES7 was using the same JSON path as in ES6, although the structure of the response has changed. This, in combination with an overly relaxed parsing of the response lead to the method always returning 0.

The consequence of this is that any configured count-based rotation policy was never triggered, due to the threshold never being crossed.

This PR is adding a test for the method, fixes the JSON path and makes sure that if the node is not found, it will raise an exception instead of returning 0.

## Types of changes

## Checklist:


mpfz0r (approved):
LGTM and works on my machine with ES 7.10.2

PR #10110, opened 2021-02-22
Adding `jest-preset-graylog` to dependabot config.

dennisoelkers:
This PR is adding the required config to track dependencies for `jest-preset-graylog` in dependabot.

In addition, it replaces the usage of the `ready-for-review` label (which is generally superseded by using the `draft` status in github) with the `dependencies` label (which indicates a mere dependency change in a PR).


linuspahl (approved):
Looks good to me.

PR #10111, opened 2021-02-22
Fix rendering of trend color in report generation

kmerz:
## Motivation
The report generation is using the chrome print to pdf functionallity
which will turn all backgrounds in css to white.

## Description
We add an !important to the css rule (and explain why) and add the
magical webkit rule which tells chrome to keep the background as it is.

## Types of changes

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/1997


dennisoelkers (approved):

dennisoelkers (changes requested):
Two suggestions:

1. The font color is now off in the PDF for the trend indicator:

image

2. Can we just disable the linter hint in `Trend.tsx`?

dennisoelkers (approved):
Good work!

PR #10138, opened 2021-02-24
Replace `loadable-components` with `React.lazy`.

dennisoelkers:
## Motivation and Context

This PR is a minor optimization to remove the need for an external dependency like `loadable-components` and using stock React functionality instead.

## How Has This Been Tested?

The web app was build and tested in the browser while looking at the network tab to ensure that code splitting is still working.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/routing/loadAsync.tsx
+++ b/graylog2-web-interface/src/routing/loadAsync.tsx
@@ -60,7 +60,7 @@ class AsyncLoaderErrorBoundary extends React.Component<{ errorComponent: React.C
   }
 }
 
-export default <T, >(f: () => Promise<{ default: React.ComponentType<T> }>) => (props: T) => {
+export default <T extends React.ComponentType<any>>(f: () => Promise<{ default: T }>) => (props: React.ComponentPropsWithRef<T>) => {
   const LazyComponent = React.lazy(f);
 
   return (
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/routing/loadAsync.tsx
```diff
@@ -14,22 +14,60 @@
...
 
 type Props = {
-  error: {
-    message: string;
-  };
+  error: Error;
 };
 
-const ErrorComponent: React.FC<Props> = ({ error }: Props) => <div>Loading component failed: {error.message}</div>;
+const ErrorComponent: React.ComponentType<Props> = ({ error }: Props) => <div>Loading component failed: {error.message}</div>;
```
Gaya, line 31: Isn't this a `React.FunctionComponent`?  Why did you change it into a `React.ComponentType`?

Gaya (approved):
Great! Much better this way. Just a small question and one test is failing on my end (and jenkins)

graylog2-web-interface/src/routing/loadAsync.tsx
dennisoelkers: `React.ComponentType` is the abstraction, it covers both class and functional components. We do not really care in this case if it's a class or functional component, so I captured that in its type.

Gaya: Makes sense! Thanks for clarifying, never used / seen it before.

Gaya (approved):
LGTM!

PR #10159, opened 2021-02-25
Add report styling for a widget

kmerz:
## Motivation
Prior to this change, all widgets had the same width in a report.

## Description
This change will give message list a broader width so it has more space
to show big messages.

## Types of changes


linuspahl (approved):
Looks good to me. I saw that we also define a default width for the widgets which is also `600`, but I agree that there is no disadvantage in being explicit.

PR #10160, opened 2021-02-25
Add support for dynamic data from UI plugins to be included in generated config.js

waab76:
## Description
Some plugins have dynamic UI settings data that needs to be provided to the UI at startup time via inclusion in the generated config.js file.  This change adds a new `PluginUISettingsProvider` interface and a method for binding instances of that interface such that the data they provide will be included when the config.js file is generated at server startup.

This is a re-do of #10063 

## How Has This Been Tested?
This change has been tested locally using an in-development Enterprise plugin.

Screen Shot 2021-02-16 at 4 26 54 PM

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/PluginModule.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/PluginModule.java
@@ -58,6 +58,7 @@ import org.graylog2.plugin.periodical.Periodical;
 import org.graylog2.plugin.rest.PluginRestResource;
 import org.graylog2.plugin.security.PasswordAlgorithm;
 import org.graylog2.plugin.security.PluginPermissions;
+import org.graylog2.web.PluginUISettingsProvider;
 
 import javax.ws.rs.ext.ExceptionMapper;
 import java.util.Collections;
@@ -322,5 +323,14 @@ public abstract class PluginModule extends Graylog2Module {
         install(new FactoryModuleBuilder().implement(AuthServiceBackend.class, backendClass).build(factoryClass));
         authServiceBackendBinder().addBinding(name).to(factoryClass);
         registerJacksonSubtype(configClass, name);
+    }
+
+    protected MapBinder<String, PluginUISettingsProvider> pluginUISettingsProviderBinder() {
+        return MapBinder.newMapBinder(binder(), String.class, PluginUISettingsProvider.class);
+    }
+
+    protected void addPluginUISettingsProvider(String providerKey,
+                                               Class<? extends PluginUISettingsProvider> uiSettingsProviderClass) {
+        pluginUISettingsProviderBinder().addBinding(providerKey).to(uiSettingsProviderClass);
     }
 }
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
```diff
@@ -68,7 +78,19 @@ public String get(@Context HttpHeaders headers) {
...
+            "pluginUISettings", buildPluginUISettings());
         return templateEngine.transform(template, model);
     }
+
+    private String buildPluginUISettings() {
+        Map<String, Object> pluginUISettings = settingsProviders.entrySet().stream().collect(Collectors.toMap(
+                Map.Entry::getKey,
+                entry -> entry.getValue().pluginSettings()));
+        try {
+            return objectMapper.writeValueAsString(pluginUISettings);
```
dennisoelkers, line 91: For debugging purposes, it would be nice to pretty-print this. We can do it using `ObjectWriter#withPrettyPrinter`.

dennisoelkers (changes requested):
Looks good so far, just one tiny nitpick!

graylog2-server/src/main/java/org/graylog2/web/resources/AppConfigResource.java
waab76: Here's what `config.js` looks like with the update in place:

Screen Shot 2021-03-01 at 11 33 24 AM

I'll push a new commit shortly

graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
```diff
@@ -52,6 +52,7 @@
 import org.graylog2.plugin.outputs.MessageOutput;
 import org.graylog2.plugin.security.PasswordAlgorithm;
 import org.graylog2.plugin.security.PluginPermissions;
+import org.graylog2.web.PluginUISettingsProvider;
```
bernd, line 55: Unrelated change?

graylog2-server/src/main/java/org/graylog2/plugin/PluginModule.java
```diff
@@ -323,4 +324,13 @@ protected void addAuthServiceBackend(String name,
         authServiceBackendBinder().addBinding(name).to(factoryClass);
         registerJacksonSubtype(configClass, name);
     }
+
+    protected MapBinder<String, PluginUISettingsProvider> pluginUISettingsProviderBinder() {
+        return MapBinder.newMapBinder(binder(), String.class, PluginUISettingsProvider.class);
```
bernd, line 329: We need to make sure the map binder exists even when there is no plugin that's calling `addPluginUISettingsProvider()`. Otherwise the server will not start.

Similar to the following. We can initialize the settings map binder there as well.

https://github.com/Graylog2/graylog2-server/blob/e8dd88cf71296c13c30b5ba4de6f559ca9500a21/graylog2-server/src/main/java/org/graylog2/shared/bindings/PluginBindings.java#L42-L46

bernd (changes requested):
Looks good! Only two comments:

graylog2-server/src/main/java/org/graylog2/plugin/PluginModule.java
waab76: Will add.

graylog2-server/src/main/java/org/graylog2/plugin/inject/Graylog2Module.java
waab76: Yes, that was a leftover from a previous version.  Will clean up.

bernd (approved):
Looks good from my point of view. :smile: :+1: Thank you!

kyleknighted (approved):
implementation is working as expected!

dennisoelkers (approved):

PR #10158, opened 2021-02-25
Integrate User First/Last Name backend

danotorrey:
Integrate Graylog User First and Last Name into `cloud-4.0` https://github.com/Graylog2/graylog2-server/pull/10074

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#768,Graylog2/graylog-plugin-enterprise#2085

PR #10162, opened 2021-02-26
Integrate User First/Last Name UI

ousmaneo:
Integrate Graylog User First and Last Name UI into cloud-4.0

See #10158


danotorrey (approved):
I can't say much about the code, but I test it, and it works well 👍

graylog2-web-interface/src/components/users/UserCreate/FullNameFormGroup.jsx
```diff
@@ -19,12 +19,12 @@ import * as React from 'react';
 
 import { FormikFormGroup } from 'components/common';
 
-const FullNameFormGroup = () => (
-  <FormikFormGroup label="Full Name"
-                   name="full_name"
+const LastNameFormGroup = () => (
```
edmundoa, line 22: Is this component still needed? The same folder has a new `LastNameFormGroup` component doing exactly the same.

edmundoa (changes requested):
A small nitpick, otherwise it looks good to me.

graylog2-web-interface/src/components/users/UserCreate/FullNameFormGroup.jsx
ousmaneo: Thank you for pointing this out, this not needed anymore.

edmundoa (approved):
LGTM 👍

PR #10183, opened 2021-03-02
Support IPinfo database files in GeoIP lookup data adapter

bernd:
Add support for the standard location and ASN IPinfo database files.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/MaxmindDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/MaxmindDataAdapter.java
@@ -28,7 +28,6 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Multimap;
 import com.google.common.net.InetAddresses;
 import com.google.inject.assistedinject.Assisted;
-import com.maxmind.geoip2.DatabaseReader;
 import com.maxmind.geoip2.exception.AddressNotFoundException;
 import com.maxmind.geoip2.model.AsnResponse;
 import com.maxmind.geoip2.model.CityResponse;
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog/plugins/map/geoip/MaxmindDataAdapter.java
```diff
@@ -245,6 +254,55 @@ protected LookupResult doGet(Object key) {
...
+                }
+            case IPINFO_STANDARD_LOCATION:
+                try {
+                    final IPinfoStandardLocation ipInfo = reader.ipInfoStandardLocation(addr);
+                    if (ipInfo == null) {
+                        LOG.debug("No IPinfo location data for IP address {}, returning empty result.", addr);
+                        return LookupResult.empty();
+                    }
+
+                    return LookupResult.multi(ipInfo.coordinates(), ImmutableMap.builder()
```
thll, line 289: Can we be sure that we get data for all these fields back from the DB? E.g. is `city` guaranteed to be always provided?  I couldn't find an example for incomplete data, but what do you think about being a bit more cautious? We could make fields in `IPinfoStandardLocation` nullable/optional and construct the map with only those values which are present here.

thll (changes requested):
Looks good overall and I couldn't find any issues when testing this.

Found just two minor things worth taking care of.

graylog2-server/src/test/java/org/graylog/plugins/map/geoip/MaxmindDataAdapterTest.java
```diff
@@ -41,18 +40,24 @@
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
-        MaxmindDataAdapterTest.CityDatabaseTest.class,
-        MaxmindDataAdapterTest.CountryDatabaseTest.class
-})
+                            MaxmindDataAdapterTest.CityDatabaseTest.class,
+                            MaxmindDataAdapterTest.CountryDatabaseTest.class,
+                            MaxmindDataAdapterTest.AsnDatabaseTest.class,
+                            MaxmindDataAdapterTest.IPinfoStandardLocationDatabaseTest.class,
```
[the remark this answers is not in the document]
bernd: Good catch, thank you!

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/MaxmindDataAdapter.java
bernd: Aye, I made all properties nullable. :+1:

thll (approved):
LGTM

PR #10186, opened 2021-03-02
Update Netty to the latest version

danotorrey:
Update Netty dependency to the latest version. 

This was recently done in https://github.com/Graylog2/gelfclient/pull/52, so the same update is being done in the graylog2-server core while we're looking at this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,7 +136,7 @@
         <mongojack.version>2.10.1</mongojack.version>
         <natty.version>0.13</natty.version>
         <netty.version>4.1.59.Final</netty.version>
-        <netty-tcnative-boringssl-static.version>2.0.34.Final</netty-tcnative-boringssl-static.version>
+        <netty-tcnative-boringssl-static.version>2.0.36.Final</netty-tcnative-boringssl-static.version>
         <okhttp.version>3.14.6</okhttp.version>
         <okta-sdk.version>3.0.1</okta-sdk.version>
         <opencsv.version>2.3</opencsv.version>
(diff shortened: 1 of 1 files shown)


danotorrey:
A quick smoke-test of the application was successful.

pom.xml
```diff
@@ -135,7 +135,7 @@
         <mongodb-driver.version>3.12.1</mongodb-driver.version>
         <mongojack.version>2.10.1</mongojack.version>
         <natty.version>0.13</natty.version>
-        <netty.version>4.1.52.Final</netty.version>
+        <netty.version>4.1.59.Final</netty.version>
         <netty-tcnative-boringssl-static.version>2.0.34.Final</netty-tcnative-boringssl-static.version>
```
mpfz0r, line 139: Would you mind als updating netty-tcnative to `2.0.36`? 
We've had problems with that in the past, so it's a good practice to update them together

mpfz0r: For reference: https://github.com/netty/netty-tcnative/releases

danotorrey: Thanks for the suggestion. Done.

bernd (changes requested):
Netty 4.1.60 just got released including a fix for a possible security problem.

See: https://netty.io/news/2021/03/09/4-1-60-Final.html

mpfz0r (approved):
Works for me. Tested HTTPS and TLS enabled beats input

mpfz0r:
@bernd can you approve or remove your change request?

bernd (approved):

PR #10192, opened 2021-03-04
Fix type inference for functions with a dash as a parameter

kmerz:
## Motivation and Context
When using a data table with big numbers and the series function contained a field with a dash `-` the type of the value was not correctly recognized which lead to the problem that no digit grouping was applied.

## Description
The `isFunction` function from Series.ts was a regex which would test for `\w` (word characters `[A-Z,a-z,_]`) which is not including the `-`. This PR changes the function to also include any character to the next `,` or `)` for the group.

## How Has This Been Tested?
- add a field containing a `-`
- use the `sum` function on that field
- check that the type is now `long` instead of `unknown`.

## Also
This PR will also add tests which were written during debugging the problem

## Types of changes


dennisoelkers (approved):
Looks good, works well!

Issue #10099, opened 2021-02-19
Maintain widget state on widget focus.


linuspahl:
## What?
We recently implemented the option to focus a widget. In this case the widget is remounting which resets the widget state but the related widget search type result stays the same. This can be problematic in the following case:

1. Create a message table widget
2. Click on the second page (in the message table pagination)
3. Focus the widget

The widget will now use the initial state and mark the first page as active, while it displays the messages of the second page. Maintaining the state in this case is currently relevant for the message table and log view.

One possible solution could be saving the widget state in the `WidgetFocusContext`, e.g. before the widget unmounts.

PR #10193, opened 2021-03-04
Preventing widget unmount on widget focus.

linuspahl:
## Description
Previously on widget focus we replaced the `WidgetGrid` with the selected widget. This way the widget got unmounted and lost its state. As described in https://github.com/Graylog2/graylog2-server/issues/10099 this was a problem for e.g. the message table, which stores information like the current page in its state.

With this PR we are only changing the widget and widget grid style on widget focus. This way not only the selected widgets keeps its focus, but also all other widgets. They also do no loose e.g. their scroll position.

While working on these changes I tried out different approaches. E.g. I tried to change the widget position and the react grid layout row height, when a widget is being focused. But all these approaches were more complex and had more side effects than the current solution.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10099

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/ReactGridContainer.jsx
+++ b/graylog2-web-interface/src/components/common/ReactGridContainer.jsx
@@ -189,7 +189,7 @@ class ReactGridContainer extends React.Component {
   };
 
   static defaultProps = {
-    animate: true,
+    animate: false,
     className: undefined,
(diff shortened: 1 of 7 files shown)


kmerz (approved):
Code LGTM :+1: 

PR #10206, opened 2021-03-05
Disable no-standalone-expect for each.test blocks

kmerz:
## Motivation
With the new jest-linter we get linting errors when using test blocks
like:

```
each([
  [1, 1, 2],
  [1, 2, 3],
  [2, 1, 3],
]).test('returns the result of adding %d to %d', (a, b, expected) => {
  expect(a + b).toBe(expected);
});
```

## Description
This PR will disable warnings for such tests.


kmerz:
Not needed any longer, we will replace `jest-each` with `it.each` from jest

Issue #9742, opened 2020-12-09
Widget edit mode is displayed in "full screen" mode instead of modal


dennisoelkers:
When a widget is edited, it should use the full screen instead of a modal. This gives the editing controls more space.

This could include exploring different options like:

 - Reusing widget focusing

PR #10212, opened 2021-03-08
Display aggregation builder in widget focus mode instead of modal.

linuspahl:
## Description
We recently introduced the functionality to focus a widget. With this PR we are using the focus mode for the widget aggregation builder. Before this change we've displayed the aggregation builder (also known as edit mode) inside a modal.

Before:
image

After:
image

Please note that we are determining the widget focus state based on the query params. When a widget is focused we are adding the query params: `?focusedId=widget-id&focusing=true`
On widget edit we are adding the query params: `?focusedId=widget-id&editing=true`

If the user focused a widget first and switched to the edit mode afterwords, they look like: `?focusedId=widget-id&focusing=true&editing=true`. This way we know if we should still focus the widget after leaving the edit mode.

Fixes: #9742

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -15,67 +15,123 @@
...
 
-      setFocusedWidget(paramFocusedWidget);
+      setFocusedWidget(nextFocusedWidget);
+    }
+  }, [focusedWidget, setFocusedWidget, widgets, focusUriParams]);
+};
+
+const useCleanupQueryParams = ({ focusUriParams, widgets, query, history }) => {
+  useEffect(() => {
+    if ((focusUriParams?.id || focusUriParams?.editing || focusUriParams?.focusing) && !widgets.has(focusUriParams.id)) {
```
dennisoelkers, line 78: I think the intention here is to clear parameters if a) an invalid widget id is focused or b) if `editing`/`focusing` are specified, but no `id`. If this is the case, we can probably make this a bit more explicit by stating this as:

```suggestion
    if ((focusUriParams?.id && !widgets.has(focusUriParams.id) 
        || (focusUriParams?.id === undefined)) {
```

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -15,67 +15,123 @@
...
 import { useStore } from 'stores/connect';
 import useQuery from 'routing/useQuery';
 import { WidgetStore } from 'views/stores/WidgetStore';
-import WidgetFocusContext from 'views/components/contexts/WidgetFocusContext';
+import WidgetFocusContext, { FocusRequest } from 'views/components/contexts/WidgetFocusContext';
 
-const _syncWithQuery = (query: string, focusedWidget: string) => {
-  const baseUri = new URI(query)
-    .removeSearch('focused');
+type QueryParamsUpdate = FocusRequest & { id: FocusRequest['id'] | undefined };
```
dennisoelkers, line 29: I think we can define the different possible states (focused widget, no focused widget, editing widget) here to make invalid states impossible.

graylog2-web-interface/src/views/components/widgets/WidgetActionsMenu.test.tsx
```diff
@@ -0,0 +1,271 @@
...
+    list: dashboardList,
+    pagination: {
+      total: 2,
+      page: 1,
+      per_page: 10,
+      count: 2,
+    },
+  };
+
+  afterEach(() => {
```
dennisoelkers, line 97: Do we still need those two calls?

graylog2-web-interface/src/views/components/widgets/WidgetActionsMenu.tsx
```diff
@@ -0,0 +1,233 @@
...
+  > * {
+    margin-right: 5px;
+
+    :last-child {
+      margin-right: 0;
+    }
+  }
+`;
+
+const _updateDashboardWithNewSearch = (dashboard: View, dashboardId: string) => ({ search: newSearch }) => {
```
dennisoelkers, line 60: We can simplify this a bit. We can add the `Search` as a third parameter and use an arrow function in line 83.

dennisoelkers (changes requested):
Looks good, works great! Added a couple of suggestions inline.

graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
linuspahl: I've added explicit types for the context state and the query params update. We need slightly more code for the query params update, but it also improved the readability.

dennisoelkers (approved):

PR #10216, opened 2021-03-08
Fix content-pack auto installation

bernd:
This fixes the automatic content pack installation when using the
"content_packs_auto_install" config option.

The ContentPackService#installContentPack expects a username as third
argument, not a user ID. We've been using "local:admin" which is the
local admin's pseudo user ID. UserService#load has a special case for the
local admin user but is expecting Configuration#getRootUsername to be
used and not "local:admin".
So the user service tried to find a user with name "local:admin" in the
database instead of returning the special admin user object.

That fails with:
  `Caused by: java.lang.IllegalStateException: Cannot load user <local:admin> from db`

The ContentPackLoaderPeriodical is now using
Configuration#getRootUsername instead of hard coding "local:admin".

Fixes Graylog2/graylog-docker#153


kmerz (approved):
Code LGTM :+1: 

PR #10244, opened 2021-03-11
Add new Allowed Paths configuration parameter and use in CSV File adapter

danotorrey:
## Description
Establish a new server configuration parameter `allowed_auxiliary_paths`, which contains an optional set of allowed paths for Graylog data files. If provided, certain operations in Graylog will only be permitted if the data file(s) are located in the
specified paths. All subdirectories of indicated paths are allowed by default. This provides an additional layer of security, and allows administrators to control where in the file system Graylog users can select files from. It protects against the potential inspection of arbitrary files in the file system from the Graylog user interface.

The first usage (as implemented in this PR) is within the CSV File lookup adapter.

## Motivation and Context

In Graylog Cloud, we will temporarily enable the CSV File lookup adapter to allow support of Graylog-made content packs that depend on CSV File lookup adapter (for example, Illuminate). This PR adds minimal security protection to prevent from using the CSV File adapter to inspect the contents or arbitrary files on the file system.

## Usage in Cloud
Once this change is merged to `master`, then it will be ported to `cloud-4.0` when the CSV File lookup adapter is re-enabled.

## How Has This Been Tested?

* Started the CSV File adapter with no allowed paths specified.
* Started the CSV File adapter with a allowed path specified that contains the file.
* Started the CSV File adapter with a allowed path specified that does not contains the file.
* Tested to verify that relative paths are working correctly.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
```diff
@@ -17,11 +17,16 @@
...
 import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.ValidatorMethod;
+import com.github.joschi.jadconfig.validators.PathExecutableValidator;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
 public abstract class PathConfiguration {
+    public static final String CSV_FILE_LOOKUP_DIR = "csv_file_lookup_dir";
```
bernd, line 28: How about using a more generic name? That way we can also use it for other data adapters that load files from the file system. (like the MaxMind adapter)

I would suggest `trusted_data_file_paths`. Let's use plural and allow a comma separated list of paths to be set. (we can use a `StringSetConverter` for the `@Parameter` to automatically set a set for the configured paths -- see `enabled_tls_protocols` option in `Configuration`)

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
danotorrey: Thanks lot for the feedback @bernd! I like this. Will do.

graylog2-server/src/test/java/org/graylog2/configuration/converters/PathSetConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test(expected = ParameterException.class)
+    public void testConvertFromNull() {
+        converter.convertFrom(null);
+    }
+
+    @Test
+    public void testConvertTo() {
+        Set<Path> sortedSet = new HashSet<>();
```
mpfz0r, line 57: a HashSet isn't sorted though.
We might sort the set in the `convertTo` method...

graylog2-server/src/main/java/org/graylog2/lookup/TrustedPathChecker.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public boolean fileIsInTrustedPath(String filePath) throws IOException {
+        if (trustedPaths.isEmpty()) {
+            return true;
+        }
+
+        // Get the absolute file path (resolve all relative paths and symbolic links).
+        // The path.toFile().getCanonicalFile() calls accomplishes this.
+        final Path csvFilePath = Paths.get(filePath).toFile().getCanonicalFile().toPath();
```
mpfz0r, line 44: This has still csv in it. How about
```suggestion
        final Path absoluteFilePath = Paths.get(filePath).toFile().getCanonicalFile().toPath();
```

graylog2-server/src/main/java/org/graylog2/lookup/TrustedPathChecker.java
```diff
@@ -0,0 +1,53 @@
...
+        for (Path trustedPath : trustedPaths) {
+            final Path absoluteTrustedPath = trustedPath.toFile().getCanonicalFile().toPath();
+            if (csvFilePath.startsWith(absoluteTrustedPath)) {
```
mpfz0r, line 47:
```suggestion
            if (absoluteFilePath.startsWith(absoluteTrustedPath)) {
```

graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
mpfz0r: Sorry for nitpicking, but I'm not sure `trusted` is the right word here. 
There is nothing trustworthy about these paths themselves.
The facts are kind of backwards: We trust Graylog to access files in these paths.
How about `accessible_data_paths`?
Or more specifically `accessible_auxiliary_data_paths`?

mpfz0r (changes requested):
apart from the naming (see comment) I have a few minor change requests

graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
bernd: @mpfz0r The "trusted" naming was my suggestion and was the first that came to my mind. :smile: 

I agree, "trusted" could be misleading. I like your suggestions. :+1:

graylog2-server/src/main/java/org/graylog2/lookup/TrustedPathChecker.java
bernd: We can also just use `Paths.get(filePath).toAbsolutePath().normaliize()` or `Paths.get(filePath).toRealPath()`. The latter one is throwing an exception if the path doesn't exist.

Either of these avoids the conversion to a `File` object and back.

mpfz0r: Sounds good. I'd avoid throwing an exception on missing files or paths here, though.
That's not the responsibility of this check

graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
danotorrey: @mpfz0r @bernd Thanks for the suggestions. I agree on "trusted" being possibly confusing. Good idea to find a better name.

"accessible" sounds too much like [accessibility](https://en.wikipedia.org/wiki/Computer_accessibility) to me, which is another topic all-together. 

What do you think about using "allowed" instead? I also think "allowed" is more correct for the specific effect of the setting (we're _allowing_ users to access them because of the setting). This is also consistent with allow list/block lists naming, which seems common.

I get the feeling that "auxiliary" might be too general, almost like it could be used for anything. "data_file" seemed a bit more specific (like it's intended to be used for things in Graylog that use files), but I am not sure about that. I am open to using auxiliary, but do you have any final thoughts or other suggestions for this part?

mpfz0r: `allowed` is fine with me. I agree that `auxiliary` is very generic. I've tried to distinguish that new setting from the `data_dir` setting.
We could call it `allowed_data_adapter_files` but I was hoping we could use that for other assets we get from the filesystem in the future..

danotorrey: @mpfz0r Good point, putting "adapter" in there is too specific. We will want to use this for other purposes too. Perhaps "auxiliary is really the best choice. I'll go with that unless anyone has other ideas.

graylog2-server/src/main/java/org/graylog2/lookup/TrustedPathChecker.java
danotorrey: That's a much better approach. I've changed to this.

graylog2-server/src/test/java/org/graylog2/configuration/converters/PathSetConverterTest.java
danotorrey: I've fixed the naming to not reference sorting. Thanks for seeing this. If it makes sense to you, This converter can continue to be unsorted.

graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public boolean isInAllowedPath(String filePath) {
+        if (allowedPaths.isEmpty()) {
+            return true;
+        }
+
+        // Get the absolute file path (resolve all relative paths and symbolic links).
+        // The path.toFile().getCanonicalFile() calls accomplishes this.
+        final Path absoluteCsvFilePath = Paths.get(filePath).toAbsolutePath().normalize();
```
mpfz0r, line 44:
```suggestion
        final Path absoluteFilePath = Paths.get(filePath).toAbsolutePath().normalize();
```

graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
```diff
@@ -0,0 +1,53 @@
...
+        for (Path allowedPath : allowedPaths) {
+            final Path absoluteAllowedPath = allowedPath.toAbsolutePath().normalize();
+            if (absoluteCsvFilePath.startsWith(absoluteAllowedPath)) {
```
mpfz0r, line 47:
```suggestion
            if (absoluteFilePath.startsWith(absoluteAllowedPath)) {
```

graylog2-server/src/test/java/org/graylog2/configuration/converters/PathSetConverterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    public void testConvertTo() {
+        Set<Path> set = new HashSet<>();
+        set.add(Paths.get("/some-dir"));
+        set.add(Paths.get("/another-dir"));
+        set.add(Paths.get("/finally-dir"));
+
+        Assert.assertEquals("", converter.convertTo(new HashSet<Path>()));
+        Assert.assertEquals("/another-dir,/some-dir,/finally-dir", converter.convertTo(set));
```
mpfz0r, line 63: That test is not stable, see comment about `convertTo`

graylog2-server/src/main/java/org/graylog2/configuration/converters/PathSetConverter.java
```diff
@@ -0,0 +1,54 @@
...
+                     .collect(Collectors.toSet());
+    }
+
+    @Override
+    public String convertTo(Set<Path> value) {
+        if (value == null) {
+            throw new ParameterException("String list of Paths must not be null.");
+        }
+
+        return value.stream().map(Path::toString).collect(Collectors.joining(","));
```
mpfz0r, line 52: The result of that could be changing, because a Set has no order.
We could add a sort to the stream.

mpfz0r (changes requested):

graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
danotorrey: Doh, of course. Thanks for seeing this. My head is still in CSV land :)

graylog2-server/src/main/java/org/graylog2/configuration/converters/PathSetConverter.java
danotorrey: @mpfz0r Thanks a lot for the feedback. Sorting has been added to the converter in https://github.com/Graylog2/graylog2-server/pull/10244/commits/ab4992166ec556213e7183db3a04af047ec99b3e

graylog2-server/src/main/java/org/graylog2/lookup/TrustedPathChecker.java
mpfz0r: > We can also just use `Paths.get(filePath).toAbsolutePath().normaliize()` or `Paths.get(filePath).toRealPath()`. The latter one is throwing an exception if the path doesn't exist.
> 
> Either of these avoids the conversion to a `File` object and back.

Sorry, I've spoke to soon. This solution does not resolve symlinks, which is bad for a security related check.
I suggest we do this:

```
diff --git graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
index abfca75cda..110169defe 100644
--- graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
+++ graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java
@@ -16,15 +16,20 @@
  */
 package org.graylog2.lookup;
 
+import org.slf4j.Logger;
+
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.SortedSet;
 
 @Singleton
 public class AllowedAuxiliaryPathChecker {
+    private static final Logger LOG = LoggerFactory.getLogger(AllowedAuxiliaryPathChecker.class);
 
     private final SortedSet<Path> allowedPaths;
 
@@ -40,7 +45,13 @@ public class AllowedAuxiliaryPathChecker {
 
         // Get the absolute file path (resolve all relative paths and symbolic links).
         // The path.toFile().getCanonicalFile() calls accomplishes this.
-        final Path absoluteFilePath = Paths.get(filePath).toAbsolutePath().normalize();
+        final Path absoluteFilePath;
+        try {
+            absoluteFilePath = Paths.get(filePath).toAbsolutePath().toRealPath().normalize();
+            LOG.warn("Could not check allowed path. Returning false.", e);
+            return false;
+        }
         for (Path allowedPath : allowedPaths) {
             final Path absoluteAllowedPath = allowedPath.toAbsolutePath().normalize();
             if (absoluteFilePath.startsWith(absoluteAllowedPath)) {

```

mpfz0r (changes requested):
Works, but fails to detect symlinks that point somewhere else

danotorrey:
> Works, but fails to detect symlinks that point somewhere else

Thanks for finding that @mpfz0r. I'll investigate if we can resolve those too. I still think resolving to actual real paths is best. I'll let you know what I find out.

danotorrey:
@mpfz0r Symbolic links are now resolved in https://github.com/Graylog2/graylog2-server/pull/10244/commits/39615f91bd317e066723a52758a4658213d3d828. The logic got more complicated, since it's not possible to resolve symbolic links without the possibility of throwing exceptions. But, all of the cases are covered by tests. Let me know if anything there can be improved.

graylog2-server/src/test/java/org/graylog2/lookup/AllowedAuxiliaryPathCheckerTest.java
```diff
@@ -22,36 +22,76 @@
...
+        final Path permittedPath = validPath();
+        when(filePath.startsWith(eq(permittedPath))).thenReturn(true);
+
+        pathChecker = new AllowedAuxiliaryPathChecker(new TreeSet<>(Collections.singleton(permittedPath)));
+        assertTrue(pathChecker.fileIsInAllowedPath(filePath));
+    }
+
+    private Path validPath() throws IOException {
+        final Path filePath = mock(Path.class);
+        when(filePath.toRealPath()).thenReturn(filePath);
```
mpfz0r, line 52: I don't think that's a good way to test this.
The only code that's really worth having the test for is to see if `toRealPath()` is working as expected.
If we mock this, the test isn't worth much.
And `startsWith()` shouldn't be mocked either..
I would stick with how the test was done before this change, and use actual temporary files and symlinks
for the test

graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
```diff
@@ -82,6 +86,9 @@ public void doStart() throws Exception {
         if (isNullOrEmpty(config.path())) {
             throw new IllegalStateException("File path needs to be set");
         }
+        if (!pathChecker.fileIsInAllowedPath(Paths.get(config.path()))) {
+            throw new IllegalStateException("The specified CSV file is not in an allowed path.");
+        }
```
mpfz0r, line 91: I just noticed that we are missing this check in `doRefresh()`
So, if you wait after the check interval of the data adapter, the file gets loaded anyway.

mpfz0r (changes requested):

mpfz0r:
Gee, thanks github for silently dropping my comments :angry: 

@danotorrey can you try to merge the current master into this branch?
I'm seeing unrelated conflicts. Something is wrong here.

danotorrey:
> @danotorrey can you try to merge the current master into this branch?
> I'm seeing unrelated conflicts. Something is wrong here.

@mpfz0r Just merged in `master` Let me know if the conflicts persist. Thanks for the additional feedback. I will address it now.

graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
danotorrey: Good find. Addressed in https://github.com/Graylog2/graylog2-server/pull/10244/commits/edb88c44e1a1d7fa067baf9acbb21ab72142bf20

graylog2-server/src/test/java/org/graylog2/lookup/AllowedAuxiliaryPathCheckerTest.java
danotorrey: Thanks a lot for the feedback @mpfz0r. This was a good change to improve the tests a lot. I've just pushed https://github.com/Graylog2/graylog2-server/pull/10244/commits/2717d95ce5f392da860d0b843e8fe0e9c3cdab19 which uses real paths and files. Specific usage of symbolic links is not included, but there is now coverage of the operative part that resolves the link (see https://github.com/Graylog2/graylog2-server/blob/2717d95ce5f392da860d0b843e8fe0e9c3cdab19/graylog2-server/src/test/java/org/graylog2/lookup/AllowedAuxiliaryPathCheckerTest.java#L100). Let me know what you think.

mpfz0r (approved):
Works and LGTM :+1:

mpfz0r:
@bernd could you update your review?

bernd (approved):
Approve to unblock. :smile:

PR #10247, opened 2021-03-12
Drain input buffer when shutting down server

thll:
## Description

When the server shuts down, it drains the process buffer and the output buffer and only then continues with the shutdown sequence.

This change adds the input buffer to the list of buffers to be drained and waited upon during server shutdown.

When elasticsearch is down, we skip buffer draining, because we don't want to wait for the process and output buffers to be empty (which wouldn't happen). This made a small refactoring of the buffer synchronizer necessary so that we can still wait for the input buffer to be drained in that case.

Must only be merged after Graylog2/graylog-plugin-cloud#793 is merged.

Should be forward-ported to `master`.

## Motivation and Context

Draining the input buffer is usually a really fast operation. Once no more events are added to the buffer, it will be empty pretty much instantly because flushing out the remaining events to the on-disk journal doesn't take much time.

By moving towards enabling alternative journal implementations, the situation might be a bit different though.

If a journal implementation experiences slowness, e.g. due to network issues, draining the input buffer might take longer than usual. If the server shutdown sequence completes faster than it takes the input buffer to be drained, the messages in the input buffer will be lost. Waiting for the buffer to be drained should reduce the chance of losing messages significantly.

## How Has This Been Tested?

Manually tested by shutting down after artificially slowing down journal write operations and/or stopping elasticsearch.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/buffers/Buffers.java
+++ b/graylog2-server/src/main/java/org/graylog2/buffers/Buffers.java
@@ -67,15 +67,18 @@ public class Buffers {
      */
     @Deprecated
     public void waitForEmptyBuffers() {
-        waitForEmptyBuffers(DEFAULT_MAX_WAIT, TimeUnit.SECONDS);
-    }
-
-    public void waitForEmptyBuffers(EnumSet<Type> bufferTypes) {
-        waitForEmptyBuffers(EnumSet.of(PROCESS, OUTPUT), DEFAULT_MAX_WAIT, TimeUnit.SECONDS);
+        waitForEmptyBuffers(EnumSet.of(PROCESS, OUTPUT));
     }
 
     /**
-     * @deprecated Usse {@link #waitForEmptyBuffers(EnumSet, long, TimeUnit)} instead
+     * Wait until the buffers of the given types have been drained or abort after 30 seconds
+     */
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/buffers/Buffers.java
```diff
@@ -22,47 +22,76 @@
...
 
+    /**
+     * @deprecated Use {@link #waitForEmptyBuffers(EnumSet)} instead
+     */
+    @Deprecated
     public void waitForEmptyBuffers() {
         waitForEmptyBuffers(DEFAULT_MAX_WAIT, TimeUnit.SECONDS);
     }
 
+    public void waitForEmptyBuffers(EnumSet<Type> bufferTypes) {
```
mpfz0r, line 73: why would we need this method? `bufferTypes` is not used

mpfz0r (changes requested):
One small thing, rest looks good

graylog2-server/src/main/java/org/graylog2/buffers/Buffers.java
thll: Thanks for noticing. The method was meant as a replacement for `#waitForEmptyBuffers()`. Fixed the unused parameter and started to actually use the method.

mpfz0r (approved):

PR #10248, opened 2021-03-12
Add an increasing sequence_nr to Random HTTP message generator

mpfz0r:
This makes it easier to detect lost messages


thll (approved):
LGTM. Very helpful for debugging 👍

PR #10251, opened 2021-03-12
 Load cloud messageloaders in LoaderTabs

ousmaneo:
These changes allow loading and using CloudRecentMesssageLoader to load a Forwarder input's message to test Stream rule.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/messageloaders/LoaderTabs.jsx
+++ b/graylog2-web-interface/src/components/messageloaders/LoaderTabs.jsx
@@ -17,6 +17,7 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
 import * as Immutable from 'immutable';
+import { PluginStore } from 'graylog-web-plugin/plugin';
 
 import { Col, Tab, Tabs } from 'components/graylog';
 import connect from 'stores/connect';
@@ -24,6 +25,7 @@ import StoreProvider from 'injection/StoreProvider';
 import ActionsProvider from 'injection/ActionsProvider';
 import MessageShow from 'components/search/MessageShow';
 import MessageLoader from 'components/extractors/MessageLoader';
+import AppConfig from 'util/AppConfig';
 
 import RawMessageLoader from './RawMessageLoader';
 import RecentMessageLoader from './RecentMessageLoader';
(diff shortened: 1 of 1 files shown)


edmundoa (changes requested):
In first place, I think these changes allow us to solve the issue and they go in a good direction, that's great! However, I have the impression that we could make them work in a more consistent way.

As it is now, we are fixing `RawMessageLoader` to work on cloud, while we hide `RecentMessageLoader` and replace it with a custom component. Since both components could be used on their own and also grouped in `LoaderTabs`, I think we should use the same approach in both cases, that is: either cloud provides custom components for both, or we allow both components to be extended from cloud. I think the second approach would mean less work for us right now, but there are advantages and disadvantages with both approaches. What do you think?

Other than that, I found some minor issues that I added inline.

ousmaneo:
> In first place, I think these changes allow us to solve the issue and they go in a good direction, that's great! However, I have the impression that we could make them work in a more consistent way.
> 
> As it is now, we are fixing `RawMessageLoader` to work on cloud, while we hide `RecentMessageLoader` and replace it with a custom component. Since both components could be used on their own and also grouped in `LoaderTabs`, I think we should use the same approach in both cases, that is: either cloud provides custom components for both, or we allow both components to be extended from cloud. I think the second approach would mean less work for us right now, but there are advantages and disadvantages with both approaches. What do you think?
> 
> Other than that, I found some minor issues that I added inline.

Hi Thank you for reviewing. 
My first implementation was to create a CloudRecentMessage Loader and ForwarderInputDropdown for the Stream rule simulator. Then I had to add some change to Forwarder input to make it work with Pipeline simulator. I think the second approach you suggested is better I will revisit the RecentMessageLoader so the implementation is more consistent.

edmundoa (changes requested):
Almost good to go, just a couple smaller comments.

edmundoa:
Could you please add the license header to `Types.js`? It looks like the build is failing for that reason:

```
14:43:06  [WARNING] Missing header in: /home/jenkins/workspace/apshots_graylog2-server_PR-10251/.repos/graylog2-server/graylog2-web-interface/src/components/messageloaders/Types.js
```

edmundoa (approved):
I think we are good to go, now we need to wait for the green build. Thank you for your patience :)

PR #10266, opened 2021-03-16
Remove the Collector and AWS plugins from the list of expected Cloud plugins

danotorrey:
Remove the Collector and AWS plugins from the list of expected Cloud plugins. They were removed from the manifests in https://github.com/Graylog2/graylog-project-internal/pull/58

Fixes this failing integration test:
```
[2021-03-16T10:58:59.494Z] [ERROR]   BackendStartupIT.loadsDefaultPlugins:66 
[2021-03-16T10:58:59.494Z] Expecting:
[2021-03-16T10:58:59.494Z]   <["Elasticsearch 7 Support",
[2021-03-16T10:58:59.494Z]     "Threat Intelligence Plugin",
[2021-03-16T10:58:59.494Z]     "Elasticsearch 6 Support"]>
[2021-03-16T10:58:59.494Z] to contain exactly in any order:
[2021-03-16T10:58:59.494Z]   <["Threat Intelligence Plugin",
[2021-03-16T10:58:59.494Z]     "Collector",
[2021-03-16T10:58:59.494Z]     "AWS plugins",
[2021-03-16T10:58:59.494Z]     "Elasticsearch 6 Support",
[2021-03-16T10:58:59.494Z]     "Elasticsearch 7 Support"]>
[2021-03-16T10:58:59.494Z] but could not find the following elements:
[2021-03-16T10:58:59.494Z]   <["Collector", "AWS plugins"]>
```

PR #10268, opened 2021-03-16
Fix test after removing plugins in cloud manifest

edmundoa:
After https://github.com/Graylog2/graylog-project-internal/pull/58, cloud tests are failing, since some plugins are no longer available. This PR removes them from the test, so builds are green again.


juju2112:
I think this is covered in #10266.

PR #10288, opened 2021-03-18
Support for displaying cloud version number

edmundoa:
This PR adds some features to support https://github.com/Graylog2/graylog-plugin-cloud/issues/295:

- Plugins can now set a git commit ID in their versions by providing a `properties` file
- Allow plugins to customize the page footer
- Add more flexibility to `Version.js`, so consumers can decide which parts of the version use


bernd (approved):

ousmaneo (approved):
Legit

PR #10293, opened 2021-03-19
Support configuration of group by / row & column pivots in new aggregation controls.

linuspahl:
## Description
This PR extends the new aggregation controls by:
- Implementing the components to configure the different group by options.
- When editing an existing widget, transforming the configured row & column pivots and the column rollup option for the group by form elements.
- Updating the widget config row & column pivots and column rollup option after submitting the group by settings.

Overview of the different configuration options:
image

To finish this implementation we need to:
- [x] Validate the configuration options.
- [x] Test if the widget configuration is always being updated properly.
- [x] Polish the exiting group by components (mainly by splitting up `elementsConfiguration/GroupBy` in multiple components).
- [x] Find the best place for `DatePivotConfig` and `ValuesPivotConfig` in `GroupByElements`. In the past I've added and used them in `Pivot.ts` but this would require many changes in other files.
- [x] Write tests for the new functionality, similar to `__tests__/AggregationWizard.metric.test`
- [x] Check if we can centralize the default values for new grouping. Relevant for the `AggregationElementSelect` / `Add grouping` button and `GroupBy` -> `onChangeField` function.

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
@@ -69,7 +69,7 @@ const _onElementCreate = (
 ) => {
   const aggregationElement = aggregationElementsByKey[elementKey];
 
-  if ('addEmptyElement' in aggregationElement) {
+  if (aggregationElement?.addEmptyElement) {
     setValues(aggregationElement.addEmptyElement(values));
   } else {
     setValues({
--- a/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
@@ -67,11 +67,14 @@ const _onElementCreate = (
   values: WidgetConfigFormValues,
   setValues: (formValues: WidgetConfigFormValues) => void,
 ) => {
+  const aggregationElement = aggregationElementsByKey[elementKey];
+  const newElement = 'createEmpty' in aggregationElement ? aggregationElement.createEmpty() : {};
+
   setValues({
     ...values,
     [elementKey]: [
       ...(values[elementKey] ?? []),
-      {},
+      newElement ?? {},
     ],
   });
 };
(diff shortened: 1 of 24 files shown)


graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
```diff
@@ -67,13 +67,19 @@ const _onElementCreate = (
...
-  setValues({
-    ...values,
-    [elementKey]: [
-      ...(values[elementKey] ?? []),
-      {},
-    ],
-  });
+  const aggregationElement = aggregationElementsByKey[elementKey];
+
+  if ('addEmptyElement' in aggregationElement) {
```
dennisoelkers, line 72: We can also just do:
```suggestion
  if (aggregationElement?.addEmptyElement) {
```

graylog2-web-interface/src/views/components/aggregationwizard/aggregationElements/GroupByElement.ts
```diff
@@ -14,16 +14,190 @@
...
+    if (grouping.interval.value && grouping.interval.value <= 0) {
+      groupByError.interval = 'Must be greater than 0.';
+    }
+  }
+
+  return groupByError;
+};
+
+const validateValuesGrouping = (grouping: ValuesGrouping): GroupByError => {
+  const groupByError = {} as GroupByError;
```
dennisoelkers, line 62: We can also do:

```suggestion
  const groupByError: GroupByError = {};
```
It does not make any difference here, just uses less force to assert the type. `as` should be used as a last resort.

graylog2-web-interface/src/views/components/aggregationwizard/aggregationElements/GroupByElement.ts
```diff
@@ -14,16 +14,190 @@
...
+  return groupByError;
+};
+
+const hasErrors = <T extends {}>(errors: Array<T>): boolean => {
+  return errors.filter((error) => Object.keys(error).length > 0).length > 0;
+};
+
+const validateGrouping = (grouping: GroupByFormValues): GroupByError => {
+  if ('interval' in grouping) {
+    const dateGrouping = grouping as DateGrouping;
```
dennisoelkers, line 85: This is not needed. The compiler knows that `grouping` is a `DateGrouping` here.

dennisoelkers (changes requested):
Looks good, works great! Just some minor nitpicks.

dennisoelkers (approved):
Looks awesome now! :tada:

PR #10317, opened 2021-03-25
Restore ability to use the CSV File adapter in Cloud

### LINKED ISSUE #790: RAW UDP input don't extract source
After updating to ver 0.92 from 0.91 Graylog2 won't extract source from RAW UDP input. Function to ovveride source is not working.

danotorrey:
## Description

Restores the ability to use the CSV File lookup adapter in Cloud with added protections for where files can be lookup up from (porting #10244 to `cloud-4.0`).

Closes https://github.com/Graylog2/graylog-plugin-cloud/issues/790

## Motivation and Context

The CSV File adapter is needed for content packs (specifically Illuminate). Adding it back to Cloud allows the content packs to be installed and used.

## How Has This Been Tested?

Started the app, and verified that the CSV File adapter can be used. Also, smoke-tested to verify that the patch checking from #10244 is working.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/PathConfiguration.java
@@ -17,11 +17,17 @@
 package org.graylog2.configuration;
 
 import com.github.joschi.jadconfig.Parameter;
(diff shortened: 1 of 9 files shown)


mpfz0r (approved):
Works and LGTM

Issue #7796, opened 2020-03-31
Keyword Time Range always uses UTC timezone


dennisoelkers:
The keyword time range used by our search is creating a new `Parser` instance in the `NaturalDateParser` class using the `UTC` timezone. This returns incorrect/unexpected results for time specifications. E.g. "yesterday 00:00:00" results in "02:00:00" for CET.

There are two options:

  - refactoring the time range classes so keyword time ranges have access to the timezone of the user who is executing the search
  - **adding a `timezone` property to the keyword range which allows setting a timezone**

Refs #7358

cawfehman:
perhaps related to this bug, or perhaps just more of a data point.  The keyword yesterday is not behaving as I would think.  If I run a search using keyword "yesterday"  today April 6, 2020, I would think it would tell Graylog to limit the search to April 5, 2020 from 00:00:00 to 23:59:59.  Instead it returns results from the previous 24 hours, beginning with the current time.

searching for "yesterday" at 11:23:00am returns results from April 5, 2020 11:23:00 through April 6, 2020 11:22:59.

those are the same results as when searching for the "last 24 hours".  That is working as expected.

again, not sure if related or another data point, or something else.

radykal-com:
Hi @dennisoelkers,

My PR is almost working, I'm a newbie with react (well, with any frontend technology) and I'm unable to load the currentUser to get the timezone.

It's always loading the default one: https://github.com/Graylog2/graylog2-server/pull/7911/files#diff-247bc39966476e445d297a7d2c315b78R127
If I change the default one to any other it works as expected, but it never loads the one from the user.

Can you check what is wrong with it please? 

Thanks you,

radykal-com:
@cawfehman as per your question, seems the problem is in the underlying library Graylog uses to parse natural language into date ranges.
Check: http://natty.joestelmach.com/doc.jsp
Most of the dates are interpreted as relative, so yesterday is like "from 24 hours ago"

cawfehman:
@radykal-com Thanks for the info... but there are alot of issues IMO with having yesterday == from 24 hours ago.  For example, If it's 1pm and I want to do a search from something that happened at 12pm, selecting yesterday shouldn't return those results and yet it does.  Likewise... searching for something that occured at 12pm the previous day, yesterday should return those results and it doesn't.   

Taking a quick look at the link you sent seems to indicate that the relative dates (like yesterday) should only take into account the date and not the time. so from that, yesterday should be just that.

[...]

PR #10320, opened 2021-03-25
Respect User TimeZone in Keyword Search

janheise:
## Motivation and Context
Added the timezone as a parameter to the keyword-search.
Added the timezone as a parameter to the NaturalDateParser

Fixes: #7796

## How Has This Been Tested?
Manual tests

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/logic/datetimes/DateTime.ts
+++ b/graylog2-web-interface/src/logic/datetimes/DateTime.ts
@@ -47,7 +47,7 @@ class DateTime {
   }
 
   static fromDateTimeAndTZ(dateTime, timezone) {
-    return new DateTime(moment(dateTime).tz(timezone));
+    return new DateTime(moment.utc(dateTime).tz(timezone));
   }
 
   static _cleanDateTimeString(dateTimeString) {
--- a/graylog2-web-interface/src/logic/datetimes/DateTime.ts
+++ b/graylog2-web-interface/src/logic/datetimes/DateTime.ts
@@ -46,6 +46,10 @@ class DateTime {
     return new DateTime(moment.utc(dateTime));
   }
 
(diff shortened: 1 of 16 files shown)


janheise:
Question: I also changed the "Legacy" Service. Should that probably be reverted?

graylog2-web-interface/src/logic/datetimes/DateTime.ts
```diff
@@ -46,6 +46,10 @@ class DateTime {
     return new DateTime(moment.utc(dateTime));
   }
 
+  static fromDateTimeAndTZ(dateTime, timezone) {
+    return new DateTime(moment(dateTime).tz(timezone));
```
dennisoelkers, line 50: This way it is actually parsing `dateTime` in the current timezone, instead of UTC. If e.g. the timezone is UTC, it will subtract the difference from the local timezone to UTC twice.

The effect of this is that there is a bug in the keyword time range tab:

  - Login as user with a UTC timezone configured (e.g. `admin`)
  - Open keyword time range tab
  - Enter e.g. `last week`
  - From/To dates will be shifted by 2 hours

A possible fix would look like this:
```suggestion
    return new DateTime(moment.utc(dateTime).tz(timezone));
```

dennisoelkers (changes requested):
Looks good so far! Just one issue remaining, comment inline.

graylog2-web-interface/src/logic/datetimes/DateTime.ts
janheise: thanks for spotting this! I double checked.

janheise:
Thanks for spotting the mistake in DateTime.ts

dennisoelkers (approved):
Works well! Great job, thanks! :tada:

PR #10321, opened 2021-03-25
Unset old group by form values when changing the type of the field

kmerz:
## Motivation
Prior to this change, the frontend included the old form values in the
request to the backend when applying a new config. For example if the
field type changed from time to value, we do not need the interval
values anymore but only the limit form value.

## Description
This change will set the old values to undefined before setting the new
default values.

## How Has This Been Tested?
- Edit a histogram widget
- Change the GroupBy from `timestamp` to a value fieldtype like `action`
- Apply those changes

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupBy/FieldComponent.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupBy/FieldComponent.tsx
@@ -38,6 +38,8 @@ const FieldComponent = ({ index, fieldType }: Props) => {
 
     if (fieldType !== newFieldType) {
       if (newFieldType === 'time') {
+        setFieldValue(`groupBy.groupings.${index}.limit`, undefined, false);
+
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupBy/FieldComponent.tsx
```diff
@@ -38,13 +38,16 @@ const FieldComponent = ({ index, fieldType }: Props) => {
 
     if (fieldType !== newFieldType) {
       if (newFieldType === 'time') {
+        setFieldValue(`groupBy.groupings.${index}.limit`, undefined, false);
+
         setFieldValue(`groupBy.groupings.${index}.interval`, {
```
dennisoelkers, line 43: Maybe we are also able to do:
```suggestion
        setFieldValue(`groupBy.groupings.${index}`, { interval: {
```

dennisoelkers (commented):
One suggestion which might simplify it a bit.

dennisoelkers (approved):
Looks good, works great!

PR #10342, opened 2021-03-29
Replace superagent with native fetch API.

dennisoelkers:
## Description

This PR is replacing our usage of the `superagent` & `superagent-bluebird-promise` dependencies with the native [`fetch API`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

## Motivation and Context

When we started to migrate the Graylog Web Interface to an SPA, there was no native API that allowed us to communicate with the REST API in a friendly way. Therefore we employed `superagent` to handle this task for us. This has changed in the way that the fetch API is now around for quite some time and allows us to do everything that `superagent` does, without an external dependency.

As we do want to support IE11 now, we need to provide a `fetch` as well as a `Promise` polyfill for now. These are added and imported at startup. If we decide to drop support for IE11, we can easily remove those again.

## How Has This Been Tested?

This has been tested with Chrome, Firefox and (most importantly) IE11 on Windows 10.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -72,6 +72,7 @@
     "numeral": "^2.0.6",
     "opensans-npm-webfont": "1.0.0",
     "plotly.js": "^1.58.4",
+    "promise-polyfill": "^8.2.0",
     "qs": "^6.3.0",
     "react-ace": "9.2.1",
     "react-color": "^2.14.0",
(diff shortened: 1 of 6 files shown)


kmerz (approved):
Code LGTM :+1: 
Worked on my machine! :+1: 
On Linux with Chrome and Firefox :heavy_check_mark:

ousmaneo (approved):
Legit!
Tested on Safari and Brave 👍

Gaya:
Love it! 🔥

PR #10343, opened 2021-03-29
Add posibility to disable actions from plugin

kmerz:
## Motivation
Prior to this change, actions introduced by plugins where not able to be
disabled by the plugin.

## Description
This change will add a interface to check if the rendered action should
be disabled or not.

## How Has This Been Tested?
Tested with Enterprise plugin

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/ExtraWidgetActions.tsx
+++ b/graylog2-web-interface/src/views/components/widgets/ExtraWidgetActions.tsx
@@ -32,15 +32,13 @@ const ExtraWidgetActions = ({ onSelect, widget }: Props) => {
   const pluginWidgetActions = usePluginEntities('views.widgets.actions');
   const extraWidgetActions = useMemo(() => pluginWidgetActions
     .filter(({ isHidden = () => false }) => !isHidden(widget))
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/views/components/widgets/ExtraWidgetActions.tsx
```diff
@@ -32,13 +32,15 @@ const ExtraWidgetActions = ({ onSelect, widget }: Props) => {
   const pluginWidgetActions = usePluginEntities('views.widgets.actions');
   const extraWidgetActions = useMemo(() => pluginWidgetActions
     .filter(({ isHidden = () => false }) => !isHidden(widget))
-    .map(({ title, action, type }) => {
+    .map(({ title, action, type, disabled }) => {
```
dennisoelkers, line 35: We can also use a default value (`() => false`) for `disabled`. This would make the check in line 41 simpler and ensures that there is always a boolean value returned.

dennisoelkers (changes requested):
Looks good, just one minor suggestion.

dennisoelkers (approved):
Looks good!

PR #10344, opened 2021-03-29
Cleanup message export format filter.

dennisoelkers:
## Motivation and Context

This change is performing some cleanup for the message export format filter. It is not a named filter anymore, because it requires to be run at a pre matching stage in order to mutate the incoming request.
Therefore the `@SupportOnlyValidExportFormats` is removed for both the filter and the `MessagesResource` REST class.

This also removes this message at startup:

```
2021-03-29 16:23:26,417 WARN : org.glassfish.jersey.server.ProcessingProvidersConfigurator - @PreMatching provider, class org.graylog.plugins.views.search.rest.MessageExportFormatFilter, also annotated with a name binding annotation. Name binding will be ignored.
```

## Types of changes

## Checklist:


janheise:
I tried to reproduce the "error" (e.g. the mutation of the request) by removing @PreMatching and did an export / ran through the filter. I had no problems. Was the last line of the filter (replacing HttpHeaders.ACCEPT) the line that's supposed to create an error? This is the only mutation of the request I could see and changing the headers should be explicitly allowed here (see spec or other examples) - and worked for me. For performance and other reasons, I would stick to the NameBinding if possible.

dennisoelkers:
Raising the error is not the actual issue. Replacing the `Accept`-header is. It is possible to mutate headers from a post matching filter, but at this time output writers are already bound and will not change anymore, even if e.g. the `Accept`-header does. This is why this filter needs to run at the pre-matching stage, unfortunately.

janheise:
Thx, I created a Testcase for that and validated accordingly.

janheise (approved):
good work!

PR #10364, opened 2021-04-01
Fix HeatmapVisualization with empty data and smallest default value

kmerz:
## Motivation
Prior to this change, if the default value was set with the smallest
value and the provided data was empty, we raised an stacktraced.

## Description
This change will ensure the empty data does not get accessed where
nothing is defined.

## How Has This Been Tested?
Create a HeatmapVisualization and select a timerange without data.

## Types of changes


dennisoelkers (approved):
Looks good and works!

PR #10370, opened 2021-04-05
Add Mongo-based storage for Data Adapter data.

waab76:
## Description
Some Data Adapters download data from HTTP endpoints and store that data in memory.  This change adds an option for downloaded data sets to be stored in MongoDB instead, hopefully freeing up more memory.

## How Has This Been Tested?
This has been tested locally with a new Abuse.ch URLhaus Data Adapter.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/db/LookupStorage.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/db/LookupStorage.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/db/LookupStorage.java
```diff
@@ -0,0 +1,151 @@
...
+import javax.inject.Inject;
+
+import java.util.Optional;
+
+import static org.graylog2.lookup.dto.LookupStorageDto.FIELD_LOOKUP_KEY;
+import static org.graylog2.lookup.dto.LookupStorageDto.FIELD_UPDATED_AT;
+
+/**
+ * This class provides MongoDB-backed storage for data retrieved by Data Adapters.  Each Data Adapter's data will
+ * be stored in a unique MongoDB collection named "lut_data_adapter_storage_Data-Adapter-ID".  The init() method should
```
bernd, line 42: Where do you see the benefit of creating a new collection for each data adapter instead of using a single table?

waab76: While it shouldn't make any real difference in query speed, I thought it would make things easier if anything went wrong and a human had to look at what was going on in MongoDB.

bernd: @waab76 Thanks, sounds good! I read up on that topic [here](https://docs.mongodb.com/manual/core/data-model-operations/#large-number-of-collections) as well and there doesn't seem to be any issues with large number of collections.

If we have one collection per data adapter we can even use a [TTL index](https://docs.mongodb.com/manual/tutorial/expire-data/) to expire data automatically.

waab76:
We decided to move this to a different repo. Closing this review.  Functionality will appear in a different review.

PR #10371, opened 2021-04-05
Adaptable Dashboard Tabs

kyleknighted:
## Description

- Moves tabs to a Dropdown depending on how wide the browser is
- If hidden tab is chosen, it is locked to the end of available tabs to show it's active

Known issues:
- Adding a new Tab is seemingly random, so initial addition will lock to the end, but after a new Tab is selected, it will go back to it's random position. Potentially resolvable from API on the way it returns data
- Rapid continual browser resizing can cause tabs to not take 100% width of the available area, but seems to be a very edge-case and potentially not a real world problem

## Motivation and Context

Too many tabs on a Dashboard leads to tabs wrapping to the next line and generally a poor experience and visual representation of our brand

image
image
image
image
image

## Types of changes

## Checklist:


kmerz (approved):
LGTM :+1: Worked on my machine.

graylog2-web-interface/src/views/components/QueryTabs.tsx
```diff
@@ -15,128 +15,204 @@
...
-  static propTypes = {
-    children: PropTypes.node,
-    onRemove: PropTypes.func.isRequired,
-    onSelect: PropTypes.func.isRequired,
-    onTitleChange: PropTypes.func.isRequired,
-    queries: PropTypes.object.isRequired,
-    selectedQueryId: PropTypes.string.isRequired,
-    titles: PropTypes.object.isRequired,
-  }
+  let maxWidth = moreBtn.offsetWidth + newBtn.offsetWidth + 30; // magic number is PageContentLayout__Container padding
```
linuspahl, line 101: I looks like we do not need to consider the page content layout padding here.

linuspahl (commented):
Code looks good to me and it works well, I like the new page title styling.

The only suggestion I have, let's display background color for the active page title. We could use the same background color for the three dots, if one of the hidden pages is active. I saw that we are already adjusting the dots color, but this way it is easier to understand that one of the hidden pages is active.

dennisoelkers (changes requested):
Testing it fails for me with this, when creating/opening a dashboard:

```
Uncaught DOMException: Failed to execute 'querySelectorAll' on 'Element': 'li:not(.dropdown):not(.query-tabs-new):not(.dropdown-menu li)' is not a valid selector.
    at adjustTabs (http://localhost:8080/app.js:66888:32)
```

dennisoelkers (changes requested):
I got it working on Firefox and tested it a little bit. It looks better, but it also requires a second click when going to a later tab. 

Important: When a tab is active, we should not hide its title. When a tab is selected from the ellipsis menu, the title is hidden right now.

graylog2-web-interface/src/views/components/QueryTabs.tsx
```diff
@@ -15,128 +15,241 @@
...
+             className="query-tabs-new">
+      <Icon name="plus" />
+    </NavItem>
+  );
+
+  const dropDownTabs = queryTabs('menuItem');
+
+  useEffect(() => {
+    adjustTabs();
+    window.addEventListener('resize', throttle(adjustTabs, 250));
```
dennisoelkers, line 208: A potentially easier solution would also be to use the `SizeMe` component from `react-sizeme` instead, which could potentially also simplify the logic used to generate the tabs. At least we can hand off the size provisioning and handling of window resizes to it.

dennisoelkers (changes requested):
One remark regarding size calculation, also: can you write some tests for it?

linuspahl (approved):
Looks good to me!

kmerz:
I tested the version again. And beside the known issues I could not see any problems. :heavy_check_mark:

dennisoelkers (approved):
Works greats and looks good now! Thanks for the patience of everyone involved! :heart:

PR #10376, opened 2021-04-06
Add sorting to aggregation wizard.

### LINKED ISSUE #9740: Aggregation Builder: Implement Sorting Actions
The aggregation builder should allow creating/showing new sorting actions, configured by:
 - field/series name
 - direction

Refs #9736.

dennisoelkers:
## Motivation and Context

This PR is adding the sorting section to the aggregation wizard, which allows the user to configure a sorting of the displayed values by selecting from a list of configured metrics and fields which are in use by the aggregation.

This PR is also slightly changing the `toConfig` signatures. Instead of passing around an `AggregationWizardConfig`, its `Builder` is used. The main reason is that `Builder.build` is performing some validation (checking if the configured sorts do match the configured pivots) that rely on the object being complete before it's built, but it also has some performance benefits by avoiding unneeded object/builder creations.

Fixes #9740.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/AggregationWizard.tsx
@@ -93,9 +93,9 @@ const _onSubmit = (formValues: WidgetConfigFormValues, onConfigChange: (newConfi
     }
 
     return toConfig;
-  }).reduce((prevConfig, toConfig) => toConfig(formValues, prevConfig), AggregationWidgetConfig.builder().build());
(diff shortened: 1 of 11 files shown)


linuspahl (commented):
Looks good to me! Could not find any problems. What do you think about filtering (or validating) the field select options based on the already configured sorts? I can't think of a case where it makes sense to configure a sort direction for a specific field twice.

dennisoelkers:
Yes, I had that in mind for a follow-up PR, good suggestion!

linuspahl (approved):

PR #10377, opened 2021-04-06
Reducing height of widget configuration components in aggregation builder.

linuspahl:
## Description
The new aggregation builder layout displays all widget config options in one column. With this PR we are reducing the height of the configuration component to make better use of the available space.

Example:
image


dennisoelkers (approved):
Looks good, improves things!

PR #10379, opened 2021-04-07
Always show elements configuration actions in aggregation builder, by positioning them as sticky.

linuspahl:
## Description
With this PR we are displaying the "Apply Changes" button of the element configuration column as sticky. We want to ensure that the user does not have to scroll to be able to submit the configuration changes.

I've displayed a box shadow for the actions row which includes the "Apply changes" button.
image
Otherwise it is harder to understand which content can be scrolled. We are hiding this box shadow, when the user scrolled to the very bottom. I've used an `IntersectionObserver` to check if the user scrolled to the bottom. It has many benefits over an `onScroll` solution. Not only in terms of performance, we also do not need to worry about updating the styling when the user resizes the browser.

Before this change:
image

After this change:
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/ElementsConfigurationActions.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/ElementsConfigurationActions.tsx
@@ -17,14 +17,14 @@
 import * as React from 'react';
 import { useEffect, useState } from 'react';
 import { useFormikContext } from 'formik';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
 import { ButtonToolbar } from 'components/graylog';
 import Button from 'components/graylog/Button';
 
 import type { WidgetConfigFormValues } from './WidgetConfigForm';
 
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/views/components/aggregationwizard/ElementsConfigurationActions.tsx
```diff
@@ -0,0 +1,99 @@
...
+  bottom: 0px;
+  height: 5px;
+  z-index: 0;
+`;
+
+const useIsStuck = (): {
+  setVisibilityIndicatorRef: (ref: HTMLDivElement) => void,
+  isStuck: boolean
+} => {
+  const [visiblilityIndicatorRef, setVisibilityIndicatorRef] = useState(null);
```
kmerz, line 59:
```suggestion
  const [visibilityIndicatorRef, setVisibilityIndicatorRef] = useState(null);
```

kmerz:
Just a small nitpick otherwise it looks fine!

dennisoelkers (approved):
Looks good, works well! :tada:

PR #10381, opened 2021-04-07
Fix permission check for notifcations in event definitions (#10380)

mpfz0r:
## Motivation
Prior to this change, the permissions check was broken for notification
when displaying the event definition summary or edit it.

## Description
This change will fix the template string which was missing a colon.

## Notes
Special kudos to @mpfz0r

(cherry picked from commit b49ce7b0801bf0d2e84da346cb284737f878151e)


kmerz (approved):
LGTM :+1:

PR #10383, opened 2021-04-07
Always show element select in aggregation builder by positioning it as sticky.

linuspahl:
## Description
With https://github.com/Graylog2/graylog2-server/pull/10379 we've positioned the elements config actions button sticky, to ensure they are always visible. This PR demonstrates the same change for the element select. We want to ensure that the user does not have to scroll to be able to add an aggregation element.

This is still a draft PR because we need to discuss if we really want position the element select this way. With the current implementation we have less space available for the elements configuration. Imo it is ok, but another solution could be to always display the element select at the bottom next to the "Apply Changes" button.

Examples with 1366 x 768:
Before: 
image
After: 
image


linuspahl:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/10480

PR #10387, opened 2021-04-08
Simplify identification of each aggregation element in aggregation builder.

linuspahl:
## Description
We've recently updated the aggregation builder layout, the config options are now displayed on the left side of the visualization.
image

In comparison with the previous implementation (where every config option had a fixed position) it is now slightly harder to identify the position aggregation element.

With this PR we are simplifying the identification by adjusting the layout as follows:
image

We are also:
- replacing the "Add an element section" buttons with the plus icon next to the title, to reduce the height of each aggregation element.
- Changing the position of the "Rollup columns" option.

## Checklist:


kmerz (approved):
Code LGTM :+1: 
Works on my machine! :heavy_check_mark:

PR #10390, opened 2021-04-08
Displaying aggregation element section actions in a row with section form to reduce section height.

linuspahl:
## Description
With this PR we are displaying the aggregation element section actions in one row with the section form components.
The less height a section needs the more widget config options can be displayed.
In a future iteration we can add drag handles for the sort and group by sections in this columns a well.

Before:
image

After:
image


dennisoelkers (approved):
Looks good, works well!

PR #10391, opened 2021-04-08
CSV File adapter: Use AllowedAuxiliaryPathChecker in File path UI validation

danotorrey:
## Description

Use the [AllowedAuxiliaryPathChecker](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/lookup/AllowedAuxiliaryPathChecker.java#L30) (and therefore the `allowed_auxiliary_paths` server configuration file parameter) for the CSV Lookup adapter UI validation for the File path field. This ensures that the user cannot use the UI validation on the File path field to inspect if arbitrary directories and files exist on the file system in the Cloud context as described in https://github.com/Graylog2/graylog-plugin-cloud/issues/830. The AllowedAuxiliaryPathChecker was already being used in the process of starting the CSV File adapter (see https://github.com/Graylog2/graylog2-server/pull/10244), and this adds the same check for the UI validation.

## Motivation and Context

Fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/830, which is a security concern in the Cloud context.

## How Has This Been Tested?

Tested in the UI to verify that when `allowed_auxiliary_paths` values are specified, the UI will not attempt to perform validation for file paths that are outside of the specified permitted directories.

Also tested that when `allowed_auxiliary_paths` values are not specified, that the file existence and readability checks are performed normally as expected.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
@@ -303,7 +303,7 @@ public class CSVFileDataAdapter extends LookupDataAdapter {
             final Path path = Paths.get(path());
             if (!context.getPathChecker().fileIsInAllowedPath(path)) {
                 errors.put("path", ALLOWED_PATH_ERROR);
-                return;
+                return Optional.of(errors);
             }
 
             if (!Files.exists(path)) {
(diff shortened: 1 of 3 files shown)


danotorrey:
Still needs test coverage.

danotorrey:
Work in progress. There are some issues with the check that I need to think through. I'll have to resume that in the morning.

patrickmann (approved):
Tested behavior locally through the UI - looks good. 
Test coverage is also OK, since the actual path checking is already being tested in AllowedAuxiliaryPathCheckerTest.

mpfz0r (approved):

Issue #9958, opened 2021-01-25
Inputs with auto generated TLS keys fail with key password


mpfz0r:
If an input is enabled with TLS, but no keys are provided Graylog automatically generates a self-signed
key and cert for this input.
This key is an unencrypted PKCS#8 key, regardless if a password for the key has been configured for the input.

When trying to connect to an input that has above conditions (tls enabled, no keys, but password)
the code will try to decrypt the ephemeral key with the password and fail.

```
java.lang.IllegalArgumentException: File does not contain valid private key: /tmp/keyutil_0.0.0.0:null_5991041306884395369.key
	at io.netty.handler.ssl.SslContextBuilder.keyManager(SslContextBuilder.java:350) ~[netty-handler-4.1.52.Final.jar:4.1.52.Final]
	at io.netty.handler.ssl.SslContextBuilder.forServer(SslContextBuilder.java:107) ~[netty-handler-4.1.52.Final.jar:4.1.52.Final]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$3.createSslEngine(AbstractTcpTransport.java:329) ~[classes/:?]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$3.call(AbstractTcpTransport.java:305) ~[classes/:?]
	at org.graylog2.plugin.inputs.transports.AbstractTcpTransport$3.call(AbstractTcpTransport.java:301) ~[classes/:?]
	at org.graylog2.plugin.inputs.transports.NettyTransport$1.initChannel(NettyTransport.java:105) ~[classes/:?]
	at io.netty.channel.ChannelInitializer.initChannel(ChannelInitializer.java:129) [netty-transport-4.1.52.Final.jar:4.1.52.Final]
	at io.netty.channel.ChannelInitializer.handlerAdded(ChannelInitializer.java:112) [netty-transport-4.1.52.Final.jar:4.1.52.Final]
	at io.netty.channel.AbstractChannelHandlerContext.callHandlerAdded(AbstractChannelHandlerContext.java:938) [netty-transport-4.1.52.Final.jar:4.1.52.Final]
	at io.netty.channel.DefaultChannelPipeline.callHandlerAdded0(DefaultChannelPipeline.java:609) [netty-transport-4.1.52.Final.jar:4.1.52.Final]
[...]

PR #10400, opened 2021-04-09
Inputs with auto generated TLS keys fail with key password #9958

patrickmann:
When a password is provided for a TLS-enabled input with self-signed cert, use that password to encrypt the key file.

## Description
If an input is enabled with TLS but no keys are provided, Graylog automatically generates a self-signed
key and cert for this input. If a password is provided, we now use it to encrypt the generated key. 

## Motivation and Context
Fixes <https://github.com/Graylog2/graylog2-server/issues/9958>
Previously, the generated key was never encrypted, even when the user provided a password. This led to a failure when accessing such an input, since the code tried to decrypt the key whenever a password was available.

## How Has This Been Tested?
Tested that a TLS input with self-signed cert can process messages, both with and without password.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/util/KeyUtil.java
```diff
@@ -174,4 +199,76 @@ public static PrivateKey readPrivateKey(Path path) throws IOException {
...
+     * @return the corresponding private key
+     */
+    public static PrivateKey privateKeyFromFile(String password, File keyFile) throws IOException, PKCSException, OperatorCreationException {
+        PrivateKey privateKey;
+
+        JcaPEMKeyConverter converter = new JcaPEMKeyConverter().setProvider("BC");
+
+        // Be sure to specify charset for reader - don't use plain FileReader
+        Object object;
+        final InputStream inputStream = Files.newInputStream(keyFile.toPath());
```
mpfz0r, line 254: I would put this in a try-with block, because that actually opens a file

graylog2-server/src/main/java/org/graylog2/shared/security/tls/KeyStoreUtils.java
```diff
@@ -31,4 +31,5 @@ private KeyStoreUtils() {
 
         return stream.toByteArray();
     }
+
```
mpfz0r, line 34: this file could stay unmodified

graylog2-server/src/test/resources/org/graylog2/plugin/inputs/transports/util/server.key.pem.pkcs8.protected
```diff
@@ -0,0 +1,30 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
mpfz0r, line 1: For completeness you could also add a PBE1 encrypted key, like here:

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/test/resources/org/graylog2/shared/security/tls/key-enc-pbe1.p8

graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/AbstractTcpTransport.java
```diff
@@ -269,8 +273,14 @@ public void stop() {
             try {
                 final SelfSignedCertificate ssc = new SelfSignedCertificate(configuration.getString(CK_BIND_ADDRESS) + ":" + configuration.getString(CK_PORT));
                 certFile = ssc.certificate();
-                keyFile = ssc.privateKey();
-            } catch (CertificateException e) {
+
+                if (!Strings.isNullOrEmpty(tlsKeyPassword)) {
+                    keyFile = KeyUtil.generatePKSC8PrivateKey(tlsKeyPassword.toCharArray(), ssc.key());
```
mpfz0r, line 278: It would be nice if we could replace the existing unencrypted key file.
Leaving the old unencrypted file in the temp directory provides no security benefit

mpfz0r (changes requested):
Works great :+1: 
However, I have a few change requests..

graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/util/KeyUtil.java
mpfz0r: I've actually meant to wrap the `InputStream`, not the `InputStreamReader` in a try with, since that's the one holding the file handle

mpfz0r (changes requested):
Two more comments

mpfz0r (approved):
LGTM :+1:

PR #10402, opened 2021-04-09
Restructure and unify aggregation element and aggregation element section naming.

linuspahl:
## Description
In the past I've understood the relation between aggregation elements and aggregation element sections as follows:

- Aggregation element: A collection of aggregation element sections, e.g. multiple metrics.
- Aggregation element section: One part of an aggregation element, e.g. one metric.

Having a discussion about the naming I've realized the following relation makes more sense:

- Aggregation element: One part of an aggregation element section, e.g. one metric.
- Aggregation element section: A collection of aggregation elements, e.g. multiple metrics.

While there are some parts where we already followed this structure, there were still some cases (attributes of `AggregationElementType` + aggregation wizard components) which had to be adjusted accordingly.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/aggregationwizard/ElementsConfiguration.tsx
```diff
@@ -21,7 +21,7 @@ import styled from 'styled-components';
 
 import AggregationWidgetConfig from 'views/logic/aggregationbuilder/AggregationWidgetConfig';
 
-import ElementConfigurationContainer from './elementConfiguration/ElementConfigurationContainer';
+import ElementConfigurationSection from './elementConfigurationSections/ElementConfigurationSection';
```
dennisoelkers, line 24: Renaming the directory is not necessary in my opinion. Also, I am getting repeatedly confused by the `elementConfiguration` vs. `aggregationElements` directories. The first contains the components for the configuration form, the latter the definition for its `AggregationElement` type (I had to look it up right now, otherwise I would not have remembered it).

Maybe we should get rid of them and create `group-by`, `metric`, `sort` & `visualization` directories instead, which contain the required components for their configuration and an `index.tsx` exporting their `AggregationElement` definition? This way we would also have a nicer separation, because we know that we do not import anything but that definition from their directories outside of them.

dennisoelkers (changes requested):
Some suggestions for improvement.

graylog2-web-interface/src/views/components/aggregationwizard/ElementsConfiguration.tsx
linuspahl: I like the idea to create a directory for each aggregation element. I suggest changing this in a follow-up PR, after we've merged https://github.com/Graylog2/graylog2-server/pull/10404 which is based in this PR. For now I just reverted the `elementsConfiguration` directory name change.

dennisoelkers (approved):
Looks good, improves things! :tada: :pray:

Issue #10319, opened 2021-03-25
Lost messages in Message Processing in case of errors - Debug Message Processing


mburni:
Graylog server log shows repeatedly the message:

`2021-03-23T16:39:09.905+01:00 WARN  [ProcessBufferProcessor] Unable to process message <e9172411-8bed-11eb-a9db-005056a14fc5>: java.lang.ClassCastException: Cannot cast java.lang.String to org.joda.time.DateTime`

Other Issues like [5637](https://github.com/Graylog2/graylog2-server/issues/5637) and [7363](https://github.com/Graylog2/graylog2-server/issues/7363) describes, that they are caused by incoming messages which cannot be processed by the ProcessBufferProcessor caused by a cast error in a pipeline rule or extractor.

We would like to figure out which of our hundreds of pipeline functions and extractors causes this and what the incoming message looks like.
Unfortunately only the message ID is logged and the incoming message is not stored in elasticsearch. 

## Expected Behavior
Dropped messages are logged to server log.

## Current Behavior
Messages are dropped during runtime errors from pipeline functions and extractors. A warning message containing the message ID is logged in the server log. The original message is lost.

## Possible Solution
Extend the the [logmessage](https://github.com/Graylog2/graylog2-server/blob/3.3.11/graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/ProcessBufferProcessor.java#L124) to log the original incoming message, which cannot be processed.

Logging might be configurable, to enable/disable logging for the ProcessBufferProcessor subsystem at a defined loglevel.

## Context

Messages with processing errors are lost and we cannot figure out which ones. So we are not able to fix the pipeline rule or extractor which is the cause.
We have a complex environment with 7 graylog server many different inputs, some with extractors and very many pipeline rules.

It's not an option to disable inputs because messages may be lost. Disabling extractors or pipeline functions will cause messages to be processed, enriched, parsed and stored wrong.

Writing all received messages to the server logfile for debugging purposes as suggested in [this discussion](https://community.graylog.org/t/debugging-message-processing/8652/6) is not an option, because our graylog cluster processes up to 15.000 msgs/sec from different sources of the IT-infrastructure and this will flood the log files and slow down message processing.

## Your Environment

* Graylog Version: 3.3.11 (planing migration to 4.0.5)
* Java Version: 1.8
* Elasticsearch Version: 6.8.6
* MongoDB Version: 4.2
* Operating System: Red Hat 7

[...]

PR #10421, opened 2021-04-12
Log failed messages and full stack trace on processing errors

bernd:
Previously there was no way to find out which messages failed in
processing and where they failed.

This change adjusts the log levels in ProcessBufferProcessor to allow
stack trace and message content logging on DEBUG level.

Refs #10319

This should be backported to 4.0 and 3.3 once it has been merged.


thll (approved):
looks good!

PR #10426, opened 2021-04-13
Execute search only for focused widget

### LINKED ISSUE #10250: Aggregation Builder: Only execute search for edited Widget
## What?
Only execute the search for the widget the user is editing. Do not execute a search for all widgets.

Use `SearchActions.reexecute` for that.

## Why?
Right now when the search is executed for all widgets instead for the one the user is working on, all widget will be rerendered, even though they are not displayed at the moment.

kmerz:
## Motivation
Prior to this change, whenever a search was executed the search was
executed for all widgets, regardless if they are seen or not.

## Description
This change will add a SearchAction which sets a filter by which the
search types will be filtered before beeing send to the server.
This filter will be set from the widget focus provider when ever a
widget gets focused or edited.

## How Has This Been Tested?
- Set a widget into focus and execute a search
- Check the result via webconsole

## Types of changes

Fixes #10250 

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2192

### DIFF
--- a/graylog2-web-interface/src/views/actions/SearchActions.ts
+++ b/graylog2-web-interface/src/views/actions/SearchActions.ts
@@ -40,6 +40,8 @@ export type SearchExecutionResult = {
   widgetMapping: WidgetMapping,
 };
 
+export type FilterSearchTypes = string[];
+
 type SearchActionsType = RefluxActions<{
   create: (search: Search) => Promise<CreateSearchResponse>,
   execute: (state: SearchExecutionState) => Promise<SearchExecutionResult>,
@@ -51,6 +53,7 @@ type SearchActionsType = RefluxActions<{
   refresh: () => Promise<void>,
   get: (searchId: SearchId) => Promise<SearchJson>,
   parameters: (parameters: (Array<Parameter> | Immutable.List<Parameter>)) => Promise<View>,
+  setFilterSearchTypes: (FilterSearchTypes) => Promise<void>,
 }>;
 
 const SearchActions: SearchActionsType = singletonActions(
@@ -76,6 +79,9 @@ const SearchActions: SearchActionsType = singletonActions(
     },
     refresh: {
       asyncResult: true,
+    },
+    setFilterSearchTypes: {
+      asyncResult: false,
     },
   }),
 );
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/actions/SearchActions.ts
```diff
@@ -51,6 +53,7 @@ type SearchActionsType = RefluxActions<{
   refresh: () => Promise<void>,
   get: (searchId: SearchId) => Promise<SearchJson>,
   parameters: (parameters: (Array<Parameter> | Immutable.List<Parameter>)) => Promise<View>,
+  setFilterSearchTypes: (FilterSearchTypes) => Promise<void>,
```
dennisoelkers, line 56: The parameter needs a name in TypeScript, it should be:
```suggestion
  setFilterSearchTypes: (searchTypes: FilterSearchTypes) => Promise<void>,
```

Without a variable name, it assigns an implicit `any` to the parameter. We can err out later, after we have enabled `noImplicitAny` in the TS config, but this requires some grunt work on our legacy code before.

dennisoelkers, line 56: In my understanding the action name is misleading. While we are effectively filtering the executed search types at some point, the intention is to filter the search types of a _widget_ to be executed. The naming here implies that we need to pass _search type ids_, but judging from the code in `SearchStore`, we need to pass a list of _widget ids_ here.

graylog2-web-interface/src/views/stores/SearchStore.ts
```diff
@@ -55,6 +55,7 @@ export type SearchStoreState = {
   search: Search,
   result: SearchResult,
   widgetMapping: WidgetMapping,
+  filterSearchTypes: FilterSearchTypes | undefined | null;
```
dennisoelkers, line 58: Maybe we do want to limit this to `FilterSearchTypes | null`? `null` would be used if it should not be filtered at all. This way we could ensure that `undefined` is not passed accidentally due to a programmatic error here (e.g. an optional chain that is short-circuiting).

dennisoelkers (changes requested):
Some nitpicking regarding names and types :)

linuspahl (approved):
Looks good to me, could not find any problems.

dennisoelkers (approved):
Looks good, works well!

PR #10428, opened 2021-04-13
Fix case inconsistencies with ES cluster health status

### LINKED ISSUE #10407: ES Cluster Status shows red, but UI shows green
## Expected Behavior

When an ES cluster is RED, the UI should have the information in red.

## Current Behavior

When restarting a cluster and waiting for shards to be assigned, the status shows as RED, but the UI shows green as the background for the message, which is a bit confusing:

image

In my case, all I did was restart all the nodes on my ES cluster simultaneously.

## Context
It's a bit confusing to have a red cluster status but the UI showing green. 

## Your Environment

* Graylog Version: 4.0.6
* Java Version: 1.8
* MongoDB Version: Percona Mongodb 4.0.2
* Operating System: Ubuntu 20.04
* Browser version: https://www.whatsmybrowser.org/b/E7ZOS

edmundoa:
This PR makes makes two changes in the way we report ES cluster health status and handle it in the web interface.

First, it fixes an inconsistency in the returned case for the health status between Elasticsearch versions. This is what the API returned before the changes, now it will use lower case in both cases:

**ES6**
`GET /api/system/indexer/cluster/health`

```
{"status":"green","shards":{"active":60,"initializing":0,"relocating":0,"unassigned":0}}
```

**ES7**
`GET /api/system/indexer/cluster/health`

```
{"status":"GREEN","shards":{"active":10,"initializing":0,"relocating":0,"unassigned":0}}
```

Additionally, it ensures the UI works in a case independent way, so the correct status is displayed regardless of the returned case.

Fixes #10407 

This PR needs to be backported into 4.0.

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ClusterAdapterES7.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ClusterAdapterES7.java
@@ -54,6 +54,7 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import java.util.Collection;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


dennisoelkers (approved):
Looks good, works well!

PR #10434, opened 2021-04-14
Refactor removal of aggregation elements and enable removal of sort elements.

linuspahl:
## Motivation and Context
With this PR we are:
- including the element title in each element delete icon title. This improves the user experience slightly and makes it easier to test the removal of an element.
- replacing the aggregation element method `onRemove` with  Formiks array helper method `remove`, where possible.
- enabling the removal of sort elements in the aggregation builder. This is already possible for other aggregation elements, but the functionality was missing for sort elements.

## How Has This Been Tested?
I will add tests for the sort element in a follow-up PR.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/SortConfiguration.tsx
```diff
@@ -33,7 +39,9 @@ const SortConfiguration = () => {
                     <div>
                       {sort.map((s, index) => (
                         // eslint-disable-next-line react/no-array-index-key
-                        <ElementConfigurationSection key={`sort-${index}`}>
+                        <ElementConfigurationSection key={`sort-${index}`}
+                                                     onRemove={() => removeSort(index)}
```
dennisoelkers, line 43: We can simplify this a bit by using formik's array helpers and just do this:
```suggestion
                                                     onRemove={() => remove(index)}
```

where `remove` is coming from the arguments of the render prop of `FieldArray`.

dennisoelkers (changes requested):
Looks good, works well, just one suggestion.

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/SortConfiguration.tsx
linuspahl: Good point, I've implemented this change for every aggregation element. It was possible for all elements, but `GroupBy`. We mainly need the `onRemove` attribute for the `GroupBy` element, to be able to remove all form values for the `groupBy` key, when we remove the last grouping. Otherwise we would display the group by section.

We could also use the array helper for groupings, remove the `onRemove` aggregation element attribute for `GroupBy` as well and implement an aggregation element attribute which indicates if we want to display the related section. But I personally prefer to cleanup the form values for a specific element type, when we remove all related elements.

kmerz (approved):
Code LGTM :+1: 

dennisoelkers (approved):
Looks good, works great!

PR #10437, opened 2021-04-14
Allow using the same LDAP attribute for different config settings

### LINKED ISSUE #10069: Teams Migration: Duplicate ID Attribute is not possible
We would like to use the UID, or the DN returned from the LDAP to use as ID Attribute in Graylog. If we test a connection or use an LDAP Browser, we can see that these fields are returned, but Graylog won't use them for its ID Attribute.

For Teams the same, we would like to use our CN as ID Attribute, the CN gets returned in a response, but Graylog won't recognize for ID Attribute:

image

## Expected Behavior

The ID attributes should be used multiple times, and Graylog should be returned to the field.

## Current Behavior

When attributes are assigned for one time, it works fine, but it does not assign the same attributes in second/multiple cases.

Here are some screenshots for reference:

It looks like the Graylog would not accept that the same attribute is used twice.
Working:
image

Not Working:
image

## Possible Solution

NA

Same as Expected behavior

## Context

Ticket# 844544

## Your Environment

* Graylog Version: 3.3 (Planning to migrate on 4.0 )
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

bernd:
Some users want to use the same attribute (e.g. "cn") for different
config settings. (e.g. name and full name)

Also don't skip the unique ID attribute when collecting user attributes.
This is an issue when the same attribute is used for the unique
attribute and another config setting (note recommended) because the
attribute will be missing from user details.

Adjust the LDAP login test output to use a special key for the unique ID
attribute to avoid confusion because of the base64 encoded value.

Fixes #10069

This needs to be backported to 4.0 once it has been reviewed, tested and merged.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackend.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authservice/backend/LDAPAuthServiceBackend.java
@@ -41,6 +41,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 import java.security.GeneralSecurityException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


mpfz0r (approved):
LGTM.
I was wondering if it makes sense to do the same for the AD backend?

PR #10438, opened 2021-04-14
Don't run index jobs for missing indices

bernd:
The SetIndexReadOnlyJob, SetIndexReadOnlyAndCalculateRangeJob,
OptimizeIndexJob and CreateNewSingleIndexRangeJob jobs
might throw an exception when running "indices.isClosed(indexName)" when
the index doesn't exist anymore.

Some of these jobs are executed with a delay so it might happen that the
index retention already deleted the index when these jobs are started.

This adds a "indices.exist(indexName)" check at the very beginning.


dennisoelkers (approved):
Looks good, thanks!

PR #10447, opened 2021-04-15
Fix fetch provider with empty response body

kmerz:
## Motivation
Prior to this change, the fetch provider was always trying to parse a json body
even if the body was empty. This failed of course.

## Description
This change will check if the body is empty before trying to parse the
response body.

## How I tested it
- Create a content pack and don't get an error when click on the final create
- Delete a content pack and don't get an error when clicking on confirm


edmundoa (approved):
Worked on my tests (creating/deleting users) 👍

dennisoelkers (approved):
Looks good, precise and works! Great work! :tada:

mpfz0r:
just noticed that this breakage while disabling event definitions. glad it's fixed :+1:

PR #10450, opened 2021-04-16
Fix rendering of WorldMapVisualization in reports

kmerz:
## Motivation
Prior to this change, we used a prop to pass the interactive context
value to the MapVisualization component. But this prop was never set so
the defaultProp (true) was used.

## Description
This change will remove the prop and use the InteractiveContext.Consumer
directly just like in GenericPlot, to ensure interactive is always set
correctly.

## How Has This Been Tested?
- Create a world map on a dashboard
- Add this world map widget alone on a report
- Render the report

## Types of changes

Needs backport to 4.0.x and perhaps more.


dennisoelkers (approved):
To the point. Good work! :tada:

PR #10253, opened 2021-03-15
Replace @babel/polyfill with core-js.

dennisoelkers:
## Motivation and Context

`@babel/polyfill` is deprecated in favor of separate inclusion of `core-js` and `regenerator-runtime`. This results in a warning during `yarn install` and will stop receiving updates at some point.

This PR is replacing the usage of `@babel/polyfill` with an explicit usage of `core-js`. Instead of including it in app bundles, a separate `polyfill` bundle is generated, which should be fully cachable.

Future improvements could include building the polyfill bundle bases on the browserslist query for a smaller asset size.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #10451, opened 2021-04-16
Fix cloud login page on IE 11

edmundoa:
Port changes from #10253 into `cloud-4.0`. The new polyfill provides an implementation from `Object.fromEntries`, which was the cause of Graylog2/graylog-plugin-cloud#846.

Fixes Graylog2/graylog-plugin-cloud#846.


Gaya (approved):
I cannot test in IE11, but the changes look okay and were previously tested anyway.

PR #10459, opened 2021-04-19
Separate `useStore` store registration and props mapping logic.

dennisoelkers:
## Motivation and Context

This PR is separating the store registration and the props mapping logic
in the `useStore` custom hook. This is avoiding a bug where passing a
props mapper function as an inline arrow function leads to missed state
updates.

Passing a props mapper function as a constant (either by extracting it
to a constant scoped outside of the component calling `useStore`, or
using `useCallback` if it is using props/state from the component) is
still recommended, but merely for performance reasons and not for
correctness.

## Types of changes

## Checklist:


kmerz (approved):
Code LGTM :+1: 

PR #10472, opened 2021-04-20
Wrap system configuration elements in error boundary.

dennisoelkers:
## Motivation and Context

Before this change, an individual core or plugin configuration element could raise an error and render the whole configurations page unusable.

This PR is adding an error boundary which is used to contain errors from each configuration element. If a single configuration element throws an error, a fallback component will be rendered instead, showing the actual error. In case of the plugin configuration elements, the config type (an fqcn) will be shown instead of the title. This gives at least an indication where the error is coming from. The rest of the page is still usable in case of an individual configuration element failing.

In addition, a general overhaul of the `ConfigurationsPage` was performed, migrating it to TypeScript and turning it into a functional component. Also, some quirks were removed, like polling the configuration prop to set the `loaded` state, which was replaced with a
promise handler.

Before:
image

After (Core configuration element failing):
image

After (Plugin configuration element failing):
image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/ConfigurationsPage.tsx
+++ b/graylog2-web-interface/src/pages/ConfigurationsPage.tsx
@@ -14,11 +14,12 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-import React from 'react';
+import * as React from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
-import { PluginStore } from 'graylog-web-plugin/plugin';
 import { chunk } from 'lodash';
 import { ErrorBoundary } from 'react-error-boundary';
+import { SystemConfiguration } from 'views/types';
 
 import { Col, Row } from 'components/graylog';
 import { DocumentTitle, PageHeader, Spinner } from 'components/common';
@@ -33,6 +34,7 @@ import UrlWhiteListConfig from 'components/configurations/UrlWhiteListConfig';
 import {} from 'components/maps/configurations';
 import style from 'components/configurations/ConfigurationStyles.lazy.css';
 import { Store } from 'stores/StoreTypes';
+import usePluginEntities from 'views/logic/usePluginEntities';
 
 import DecoratorsConfig from '../components/configurations/DecoratorsConfig';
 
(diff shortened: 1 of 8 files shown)


graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+    ];
+
+    if (isPermitted(permissions, ['urlwhitelist:read'])) {
+      promises.push(ConfigurationsActions.listWhiteListConfig(URL_WHITELIST_CONFIG));
+    }
+
+    const pluginPromises = pluginSystemConfigs
+      .map((systemConfig) => ConfigurationsActions.list(systemConfig.configType));
+
+    Promise.all([...promises, ...pluginPromises]).then(() => setLoaded(true));
```
edmundoa, line 155: I like this change, way cleaner than before! However I see a problem here: when one of the configurations can't load (maybe there is some network issue or the server threw an exception), none of the configurations will be rendered and the page will stay in a loading state.

I think the easiest solution would be replacing `Promise.all` with [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled), and display an error message in those configurations that couldn't load. `Promise.allSettled` is quite recent but has already [~90% browser support](https://caniuse.com/mdn-javascript_builtins_promise_allsettled) and our `promise-polyfill` version should already support it. 

Alternatively we could let each configlet fetch its own configuration. That has some drawbacks on its own, but would decouple all fetches and would also make more explicit the relationship between each configuration and its data.

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { useStore } from 'stores/connect';
+import CombinedProvider from 'injection/CombinedProvider';
+import { isPermitted } from 'util/PermissionsMixin';
+import SearchesConfig from 'components/configurations/SearchesConfig';
+import MessageProcessorsConfig from 'components/configurations/MessageProcessorsConfig';
+import SidecarConfig from 'components/configurations/SidecarConfig';
+import EventsConfig from 'components/configurations/EventsConfig';
+import UrlWhiteListConfig from 'components/configurations/UrlWhiteListConfig';
+import {} from 'components/maps/configurations';
+import style from 'components/configurations/ConfigurationStyles.lazy.css';
```
edmundoa, line 34: These styles should eventually be used through styled components but I think we should take that into another PR. Probably also improving some margins and styling, since the page needs some for a long time.

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+  title: string,
+}
+
+const Boundary = ({ children, title }: BoundaryProps) => (
+  <ErrorBoundary FallbackComponent={(props) => <ErrorFallback title={title} {...props} />}>
+    {children}
+  </ErrorBoundary>
+);
+
+const ConfigletContainer = ({ children, title }: BoundaryProps) => (
```
edmundoa, line 75: I think the component is way clearer now that before but I also have the impression that the component would benefit of splitting some of its inline components into their own files, making it easier to re-use them and also reducing the amount of concepts inside the file. `ConfigletContainer` and its related components would be a good candidate, I think, since they provide a valuable component for other parts of the application rendering configuration options, and would make `ConfigurationsPage` more focused on its own purpose. What do you think?

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -0,0 +1,240 @@
...
+type PluginConfigRowsProps = {
+  configuration: Record<string, any>,
+  systemConfigs: Array<SystemConfiguration>,
+};
+
+const PluginConfigRows = ({ configuration, systemConfigs }: PluginConfigRowsProps) => {
+  const pluginConfigs = useMemo(() => _pluginConfigs(systemConfigs, configuration), [configuration, systemConfigs]);
+
+  // Put two plugin config components per row.
+  const configRows = chunk(pluginConfigs, 2)
```
edmundoa, line 116: Didn't know about `chunk`, I like it 👍

edmundoa (changes requested):
I like the changes and I think make the component more comprehensible and resilient 👍  Besides a couple inline comments, I was wondering: the PR was inspired by an error in one of the configlets but I cannot reproduce it now. Did you manage to fix it or was it some temporary error?

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
dennisoelkers: Great finding! I checked that our polyfill supports `Promise.allSettled` and will use that instead. Thanks a lot for including a suggestion for a solution in your comment!

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
dennisoelkers: Great idea. I had the same impression, but was a bit hesitant to include that step in the PR. Thanks for the motivation! I extracted the `ConfigletContainer` and `PluginConfigRows` components.

edmundoa: I think it looks much cleaner now, thank you! 👍

edmundoa (approved):
Looks good to me. Thank you for the improvements!

PR #10479, opened 2021-04-21
Allow changing sort of grouping and sort aggregation elements.

linuspahl:
## Motivation and Context
With this PR we are implementing the `SortableList` component for the sort and group by aggregation elements.
This way a user can change their order, using drag and drop.
image

Please note:
- I tried using Formiks field array helper (`swap`) instead of `setFieldValues`, to update the form state after changing the sort of aggregation elements.
Unfortunately this results in the following form error state: `errors: { aggregationElementKey: [undefined]  }`. This is may a Formik bug, there are multiple Formik issues related to this problem. Our aggregation element `validation` methods behave correctly.

- Compared with the default `SortableList`, dragging an item is like 100ms slower. I've implemented `React.memo` for some components, like `Sort.tsx` and `GroupBy.tsx` and it already improved the performance a lot. If we still experience performance problems with the production version we can try to improve it even more.
 
## Types of changes

## Checklist:

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2201


graylog2-web-interface/src/views/components/aggregationwizard/__tests__/AggregationWizard.groupBy.test.tsx
```diff
@@ -31,6 +31,9 @@ import dataTable from 'views/components/datatable/bindings';
 
 import AggregationWizard from '../AggregationWizard';
 
+const timeout = (Number(process.env.TIMEOUT_MULTIPLIER) || 1) * 15000;
+jest.setTimeout(timeout);
```
dennisoelkers, line 35: Instead of setting this globally, we can also pass the timeout as a third parameter to `it` (see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L380)). That makes it a bit cleaner and decoupled from `jest`. I could also not say if calling `jest.setTimeout` in one test has an impact on timeouts in other tests.

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupBy.tsx
```diff
@@ -41,10 +41,13 @@ const GroupBy = ({ index }: Props) => {
...
+                         name={`groupBy.groupings.${index}.limit`}
+                         type="number"
+                         bsSize="small" />
       )}
     </Wrapper>
   );
 };
 
-export default GroupBy;
+export default React.memo(GroupBy);
```
dennisoelkers, line 53: Tiny nitpick: I would prefer wrapping with `React.memo` at the place of the function declaration instead of wrapping the export, that makes it more visible. It is a bit different with `connect`, but it's usually a multi-line statement that is less subtle than `React.memo`, which is easy to overlook.

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupByConfiguration.tsx
```diff
@@ -73,21 +73,22 @@ const GroupByConfiguration = () => {
...
-                              <GroupBy index={index} />
-                            </ElementConfigurationContainer>
-                          );
-                        })}
-                      </div>
-                    </>
+                    <SortableList items={groupBy?.groupings}
+                                  onMoveItem={(newGroupings) => setFieldValue('groupBy.groupings', newGroupings)}
+                                  customListItemRender={({ index, dragHandleProps, draggableProps, className, ref }) => (
+                                    <ElementConfigurationContainer key={`grouping-${index}`}
```
dennisoelkers, line 79: Wouldn't it be better to include the random id in the key, instead of the array index? The array index is prone to changes when reordering the list, but the random id is stable?

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/SortConfiguration.tsx
```diff
@@ -25,24 +26,25 @@ import { WidgetConfigFormValues } from 'views/components/aggregationwizard/Widge
...
-                                                       elementTitle={SortElement.title}>
-                          <Sort index={index} />
-                        </ElementConfigurationContainer>
-                      ))}
-                    </div>
-                  </>
+                  <SortableList items={sort}
+                                onMoveItem={(newSort) => setFieldValue('sort', newSort)}
+                                customListItemRender={({ index, dragHandleProps, draggableProps, className, ref }) => (
+                                  <ElementConfigurationContainer key={`sort-${index}`}
```
dennisoelkers, line 37: Same as above, using random id instead of index?

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/Sort.tsx
```diff
@@ -121,4 +121,4 @@ const Sort = ({ index }: Props) => {
   );
 };
 
-export default Sort;
+export default React.memo(Sort);
```
dennisoelkers, line 124: Same tiny nitpick as above, wrapping declaration with `React.memo` instead of export?

dennisoelkers (changes requested):
Looks good, works great! Added just a couple of (optional) suggestions/nitpicks.

graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupByConfiguration.tsx
linuspahl: You are right, we should include the id now where it is possible.

graylog2-web-interface/src/views/components/aggregationwizard/__tests__/AggregationWizard.groupBy.test.tsx
linuspahl: I like the approach, I will give it a try!

Regarding `jest.setTimeout`: "it only affects the test file from which this function is called"

dennisoelkers (approved):

PR #10505, opened 2021-04-26
Fixing route leave prompt for focused mode on dashboards.

dennisoelkers:
## Motivation and Context

Prior to this change, the route leave prompt asking the user for confirmation if the current view is dirty and the route is left was unconditionally triggered for any route history action. This means it is also triggered when query parameters are being added to the current URL.

With the introduction of the focused mode and its use during widget editing, this lead to the phenomenon that the prompt was always triggered when a widget on an unsaved dashboard was focused, e.g. for editing.

This change is now adding a check to the `Prompt` component used to generate the confirmation prompt, checking if the new location is a different path than the current one. Therefore it is not triggered if only the query parameters are changing.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/ConfirmLeaveDialog.tsx
+++ b/graylog2-web-interface/src/components/common/ConfirmLeaveDialog.tsx
@@ -16,7 +16,7 @@
  */
 import * as React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


kmerz (approved):
Code LGTM :+1: 

PR #10508, opened 2021-04-26
Timerange Info for dashboard widgets

kmerz:
## Description
Display the configured time range of a widget in dashboards.

## Motivation and Context
Having a different time range for each widget configured on a dashboard, can make it hard to remember those settings.
So we now display that information in the footer of the widget.

## How Has This Been Tested?
- Create a widget
- Set relative/absolute/keyword timerange
- Check the result in the dashboard

## Types of changes

Graylog (2)

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/TimerangeInfo.tsx
+++ b/graylog2-web-interface/src/views/components/widgets/TimerangeInfo.tsx
@@ -40,7 +40,7 @@ const Wrapper = styled.div(({ theme }) => css`
 const TimerangeInfo = ({ className, widget, activeQuery, widgetId }: Props) => {
   const configuredTimerange = timerangeToString(widget.timerange || DEFAULT_TIMERANGE);
   const { result, widgetMapping } = useStore(SearchStore);
-  const searchTypeId = widgetMapping.get(widgetId).first();
+  const searchTypeId = widgetId ? widgetMapping.get(widgetId).first() : undefined;
   const effectiveTimerange = activeQuery && searchTypeId ? result?.results[activeQuery].searchTypes[searchTypeId].effective_timerange : {};
   const effectiveTimerangeString = timerangeToString(effectiveTimerange);
 
--- a/graylog2-web-interface/src/views/components/widgets/TimerangeInfo.tsx
+++ b/graylog2-web-interface/src/views/components/widgets/TimerangeInfo.tsx
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


graylog2-web-interface/src/views/components/widgets/TimerangeInfo.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const Wrapper = styled.div(({ theme }) => css`
+  font-size: ${theme.fonts.size.tiny};
+  color: ${theme.colors.gray[30]};
+  width: max-content;
+`);
+
+const TimerangeInfo = ({ className, widget }: Props) => {
+  const configuredTimerange = timerangeToString(widget.timerange);
```
linuspahl, line 36: After creating a widget, the widget time range is `undefined` and we display the `DEFAULT_TIMERANGE` in the widget edit mode. Let's display the same info in the widget footer (instead of nothing. One side effect, we need to wrap `TimerangeInfo` with `IfDashboard`.

graylog2-web-interface/src/views/logic/queries/TimeRangeToString.ts
```diff
@@ -0,0 +1,60 @@
...
+    .fromNow();
+};
+
+const relativeTimeRangeToString = (timerange: RelativeTimeRange): string => {
+  if (isTypeRelative(timerange)) {
+    return readableRange(timerange, 'range');
+  }
+
+  return `${readableRange(timerange, 'from')} - ${readableRange(timerange, 'to')}`;
+};
```
linuspahl, line 34: I saw that a widget time range like `5 minutes ago - Now` results in `All Time`.
We can fix it with:


```suggestion
const relativeTimeRangeToString = (timerange: RelativeTimeRange): string => {
  if (isTypeRelativeWithStartOnly(timerange)) {
    return `${readableRange(timerange, 'range')} - Now`;
  }

  return `${readableRange(timerange, 'from')} - ${readableRange(timerange, 'to', 'Now')}`;
};

```

With the previous time range picker changes the widget time range no longer contains a `range` attribute, but I see a benefit in taking care of this case, as well.

graylog2-web-interface/src/views/components/widgets/Widget.tsx
```diff
@@ -268,6 +275,9 @@ class Widget extends React.Component<Props, State> {
               {visualization}
             </WidgetErrorBoundary>
           )}
+          <WidgetFooter>
+            <TimerangeInfo widget={widget} />
+          </WidgetFooter>
```
linuspahl, line 280: I prefer to hide this info, when editing a widget.

linuspahl (changes requested):

linuspahl (approved):
Looks good to me!

PR #10512, opened 2021-04-27
Enable full backend tests for plugins

thll:
## Description

This change adds some degree of flexibility to the full backend tests to allow creating tests for plugins outside of the `graylog2-server` repository.

Tests can now specify
* which plugin jars they want to include when running the server
* which maven project needs to be built when executing the tests
* MongoDB fixtures they require to be imported into MongoDB before the server is started

### Things to improve
I had some things in mind which would be nice to improve but which I'd like to safe for future iteration because I didn't get to them
* Don't build the jars before starting the server container but let Maven compute a classpath based on the `target` directories and mount it into the container.
* Support some kind of meta-annotation, so that plugin tests don't need to specify the now numerous parameters of `@ApiIntegrationTest` for each and every test class.

## Motivation and Context

Creating tests for plugins outside of the `graylog2-server` repository was not possible because the server was started with a fixed set of plugins.

## How Has This Been Tested?

Tested by creating a test for an external plugin and verifying that the plugin was loaded.

## Types of changes

## Checklist:


thll:
Tests are failing on Jenkins. I'll investigate.

PR #10515, opened 2021-04-27
Make working with ValidationResults a bit more pleasant

mpfz0r:
This allows us to use it in a Builder like pattern


dennisoelkers (approved):
Totally!

PR #10520, opened 2021-04-28
Setting up `react-query` initially.

dennisoelkers:
## Motivation and Context

This PR is doing solely infrastructure work by adding the `react-query` package and provides a context with a `QueryClient` for a consistent configuration throughout the application.

The `react-query` package is installed through the `graylog-web-plugin` package for version consistency across core & plugins. It is also added to the `vendor` bundle, so it is effectively a singleton across core & plugins, preventing multiple instances to be initialized which would prevent an effective usage of the client cache.

This allows us to use `react-query` throughout the application. The targeted use case is to use it for backend API communication in places where no store exists yet, the store pattern does not match the intended purpose (e.g. because no state should be held), but client caching is desired. Ideally, new functionality is declared through a custom hook thinly wrapping the `useQuery` call, which can be parameterised for the proper request to be formed.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me, the configuration is reasonable.

PR #10525, opened 2021-04-29
Implement Cancellable Promise

### LINKED ISSUE #10488: GrokPatterns Page does not load patterns
## What
When loading the grok patterns page no patterns will be displayed, instead an error is printed in the console:
```
GrokPatterns.jsx:82 Uncaught (in promise) TypeError: _this.loadPromise.isCancelled is not a function
    at GrokPatterns.jsx:82
```

kmerz:
## Motivation
Prior to this change, we rely in our code on the cancellbale promise
interface from bluebird promise. But we want to use the native promise
implementation.

## Description
This change will add a wrapper class arround the promise from native
`window.fetch` which will set a cancel flag which can be accessed later
to check if a promise was cancelled.

## Note
This does not mean the then chain will be stopped. This cannot be done
with the native promise implementation. But we can check if someone
wants us to ignore the result.

Fixes #10488

## How Has This Been Tested?
- Go to the grok patterns page and see the grok patterns load again.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/logic/rest/CancellablePromise.test.ts
+++ b/graylog2-web-interface/src/logic/rest/CancellablePromise.test.ts
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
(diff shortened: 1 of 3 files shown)


linuspahl (approved):
Could not find any problem!

Issue #10213, opened 2021-03-08
Aggregation Builder: Collapse dashboard filter and adjust deactivation of widget query controls on widget edit.


linuspahl:
## Description
With https://github.com/Graylog2/graylog2-server/pull/10212 we changed the way we are displaying the widget edit mode / aggregation builder.
Before this change we've displayed the aggregation builder inside a modal, now we are using the widget focus mode:
![image](https://user-images.githubusercontent.com/46300478/110337156-e5a6ef00-8025-11eb-8333-504cd6587233.png)

The new layout changes the relation between the dashboard filter and the widget query controls, because it is possible to adjust the dashboard filter while editing a widget.

We want to:
- Hide the dashboard filter on widget edit
- Adjust the way we are disabling the widget query controls on widget edit.
Currently we are disabling all controls, when any dashboard filter is defined:

![image](https://user-images.githubusercontent.com/46300478/116383904-960bc500-a817-11eb-970d-c056b2b6bc27.png)
Our plan is to disable only the widget time range picker when a time range dashboard filter is defined, because the filter overrides the widget time range. In addition we want to display the search query dashboard filter next to widget search query, because both queries get combined.

![image](https://user-images.githubusercontent.com/46300478/116384448-2518dd00-a818-11eb-8761-6e322ba1adfb.png)

PR #10527, opened 2021-04-29
Only disable widget query controls partially when dashboard filter is defined.

linuspahl:
## Motivation and Context
Before this change we've disabled all widget query controls (time range picker, query input and streams select) when any dashboard filter (time rang or search query) is defined. 
image

With this PR we are only disabling the widget query controls partially:
- When a dashboard time range is defined, we are disabling the widget time range picker, because the widget time range is being overridden.
- When a dashboard search query is defined we are displaying the dashboard search query next to the widget query input, because both queries are being combined.

image

These changes make it more clear what effect each dashboard filter has. Controls like the widget search query or the widget steam select are now always configurable, even when a dashboard filter is defined.

Please note:
- I implemented the `QueryInput` component for the query filter info in the aggregation builder. I thought about creating another read only version of the `QueryInput`, because the component contains some code which is not required when the input is disabled. Since this component would contain a lot of redundant code, I decided to reuse the `QueryInput` component.
- I am using `timerangeToString` to display the query time range override in the aggregation builder, because the resulting time range has a smaller width than the `TimeRangeDisplay` (which also displays the from and to label). This way it is also easier to differentiate between the widget time range and the dashboard filter preview.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10213

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/QueryOverrideInfo.tsx
+++ b/graylog2-web-interface/src/views/components/QueryOverrideInfo.tsx
@@ -26,6 +26,7 @@ const Wrapper = styled.div`
   width: 30%;
   display: flex;
   align-items: center;
+
   &::before {
     box-shadow: 17px 0 16px -16px rgb(0 0 0 / 40%) inset;
   }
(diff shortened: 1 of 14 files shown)


kmerz (approved):
Code LGTM :+1: 
Was not able to break it :hammer:

dennisoelkers (approved):
Looks great, works well! :tada:

PR #10529, opened 2021-04-29
Do not display pluggable view header elements in aggregation builder.

linuspahl:
## Motivation and Context

Before this change we displayed the pluggable view header elements at the top
of the search page and at the top of the aggregation builder.
This was necessary because we previously displayed the aggregation builder in a
modal. Now were we display the aggregation builder inline, below the
search bar, we no longer need to display them in the aggregation builder
as well.

## Checklist:


kmerz (approved):
LGTM :+1: 

PR #10537, opened 2021-04-30
Fix filtering for `id` on paginated resources

kmerz:
## Motivation and Context
The stream id is used all over the place when referencing a stream (e.g event message) but there is NO WAY
to find a stream based on its id.

## Description
Add a id filter option to the streams overview page and fix it for everything else.
Also add a Query Helper to stream overview.

## How Has This Been Tested?
- search with `id: 5f4dfb9c69be46153b9a9a7b`

Graylog - Streams (1)

## Types of changes

## Note
Suggest a backport to 3.3


linuspahl (commented):
What do you think about mentioning the id in the query helper info for (at least) the users and teams overview as well?

graylog2-server/src/main/java/org/graylog2/search/SearchQueryParser.java
```diff
@@ -259,6 +260,8 @@ FieldValue createFieldValue(SearchQueryField field, String quotedStringValue, bo
                 return new FieldValue(Integer.parseInt(pair.getLeft()), pair.getRight(), negate);
             case LONG:
                 return new FieldValue(Long.parseLong(pair.getLeft()), pair.getRight(), negate);
+            case ID:
```
bernd, line 263: Can you please rename this to `OBJECT_ID` to make it a bit more explicit? :slightly_smiling_face: Thank you!

bernd (changes requested):
Thank you! Can you please merge in the current master branch? There are some conflicts. Thank you!

bernd (approved):
LGTM :+1:

linuspahl (changes requested):
When a user searches for an id and there is no results, the search results in a bad request. For example: `id:search-id`. This is the case for every entity (saved searches, streams...).
It would be optimal if we keep the request result consistent, in comparision with other query fields.

kmerz:
> When a user searches for an id and there is no results, the search results in a bad request. For example: `id:search-id`. This is the case for every entity (saved searches, streams...).
> It would be optimal if we keep the request result consistent, in comparision with other query fields.

This is because the `id:search-id` is not a valid id and the response is a illegal argument. We should handle that better but this will not part of this release anymore. I will open a separate issue for that.

linuspahl (approved):

PR #10543, opened 2021-05-03
Add validation that a sort does not work on timebased charts

### LINKED ISSUE #10416: Aggregation Wizard: Disable Sorting for time based visualization
## What?
Disable the sorting for timestamp when the visualization is based on plotly.

## Why?
Currently the user is able to change the sorting direction of a timestamp based widget, but viualization based on plotly will ignore the sorting direction we put in the datat.

kmerz:
## Motivation
Prior to this change, the user could configure sort elements on the
aggregation widget. But it either had not effect or it only broke the
visualization on timebase graphic charts.

## Description
This change will introduce a validaton to check if the sort element was
introduced together with a non table visualization and a timebase row
groupby.

## How Has This Been Tested?
- Add a sorting to a histogram

## Types of changes

Fixes #10416


linuspahl (approved):
Looks good to me, works as expected.

PR #10392, opened 2021-04-08
Support + in index set prefixes.

dennisoelkers:
## Description

**Note:** This PR requires a backport to `4.0.`.

This PR is making a couple of changes to support the + character in index set prefixes. The required changes were:

  - Properly URL encode `+` in names when creating the index template
  - Quote `+` when creating the regular expression used to match indices     against the prefix

## Motivation and Context

When an index set is configured with a `+` in the index set prefix, all kinds of issues show up, making it effectively unusable. Indices are not identified to belong to this index set, creation of new indices for this index set do not have the proper index template.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #10544, opened 2021-05-03
Do not quote index match pattern for index management check pattern.

dennisoelkers:
## Motivation and Context

**Note:** This needs to be backported to `4.0`.

In #10392 we added quoting for the patterns used to identify if an index is managed by an index set. This works fine unless `IndexSetConfig#indexMatchPattern` is defined. The index match pattern can contain special characters and regex terms deliberately, which must not be quoted. The index match pattern is currently used by archiving, which lead to it not being able to identify indices belonging to the restored archives index set.

This PR is removing quoting for the index match pattern. Any usage of the index match pattern must ensure that all special characters are quoted properly.

In a future PR I would suggest that the index match pattern can not be set through the REST API. It is currently not exposed through the frontend, the only use case is a programmatic one in archiving and it can introduce a lot of issues if the input is not quoted properly, while the benefit of it's exposure through the REST API is very small and the ability to introduce very subtle issues is big.

Fixes Graylog2/graylog-plugin-enterprise#2216.

## Types of changes

## Checklist:


thll (approved):
LGTM

PR #10567, opened 2021-05-04
Encode usernames used in URLs - 4.0

### LINKED ISSUE #10530: Usernames with / not encoded in API calls
We have usernames with / in them (I know, bad choice), e.g. we have tom and tom/admin, dick and dick/admin to separate privileged and unprivileged accounts. These are in ldap, and break the role assignment page on Systems/Users and teams/edit.

## Expected Behavior
Role assignment page should load

## Current Behavior
Role assignment page does not load and errors with a 404 because the "/" isn't encoded in the api call.
When replacing the "/" in the APl call with %2F things work as expected.

## Possible Solution
Encode certain characters in usernames that are valid in e.g. LDAP but not in URLs like "/"

1. Create a local user with a / in the name, e.g. test/admin
2. Try editing the user and watch role assignment not loading and leading to a 404

## Context

Cannot change the /admin users in the ldap system. 

## Your Environment
* Graylog Version: 4.0.6
* Java Version:  openjdk-11-jdk-headless 11.0.11+9-1~deb10u1
* MongoDB Version: 4.2.13
* Operating System: Debian 10.9
* Browser version: Firefox 88

edmundoa:
Usernames may contain some URL reserved characters that may break API
calls. This PR fixes those instances where we were not yet encoding
usernames.

Fixes #10530, backport of #10566

### DIFF
--- a/graylog2-web-interface/src/stores/roles/AuthzRolesStore.js
+++ b/graylog2-web-interface/src/stores/roles/AuthzRolesStore.js
@@ -61,13 +61,19 @@ const _responseToPaginatedUserList = ({ count, total, page, per_page, query, use
   },
 });
 
+const encodeApiUrl = (apiRoute: (...args: Array<string>) => { url: string }, uriParams: Array<string> = []): string => {
+  const encodedUriParams = uriParams.map((param) => encodeURIComponent(param));
+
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Looks good as well! Thanks for taking care.

PR #10573, opened 2021-05-05
Handle session expiration properly in frontend.

dennisoelkers:
## Motivation and Context

Through the transition from superagent to the native fetch API (#10342), a different behavior for request errors was introduced. A non-successful request does not trigger a promise rejection automatically, so an additional `catch` call in the promise chain is required in order to handle e.g. session expirations (which result in a 401 for subsequent requests) properly.

This change is correcting handling of errored responses, adds tests and cleans up the `FetchError` class by removing explicit `console.error` calls in its constructor, leaving this to the environment for uncaught promise rejections, but adding proper `name` & `message` fields so the class cleanly extends the `Error` type.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/logic/errors/FetchError.ts
+++ b/graylog2-web-interface/src/logic/errors/FetchError.ts
@@ -32,6 +32,8 @@ type Additional = {
 }
 
 export default class FetchError extends Error {
+  name: 'FetchError';
+
   responseMessage: string;
 
   additional: Additional;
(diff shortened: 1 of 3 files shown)


mpfz0r (approved):
This fixes it for me.
Funny enough I found a different session bug in the backend as well :-/

kmerz (approved):
Code LGTM :+1: 

PR #10574, opened 2021-05-05
Do not execute validation of widget config form twice when changing value of field inside `FieldArray`.

linuspahl:
## Motivation and Context

By default `validateOnChange` of the `Formik` and `FieldArray` component is
`true`. One side effect is that the custom validate method is being
executed twice when changing any value. With this change we are only
executing the validation once, by setting `validateOnChange` to `false`
for the `FieldArray`s.

This change will improve the experience when typing text into an input, because the input is more responsive.

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupByConfiguration.tsx
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/elementConfiguration/GroupByConfiguration.tsx
@@ -71,6 +71,7 @@ const GroupByConfiguration = () => {
         </Field>
       )}
       <FieldArray name="groupBy.groupings"
+                  validateOnChange={false}
                   render={() => (
                     <SortableList items={groupBy?.groupings}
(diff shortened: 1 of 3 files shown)


dennisoelkers (commented):
The change looks straight-forward to me. A question that came to my mind: Right now we are using one big validation function at the top of the formik form, which branches off into validating individual elements (groupings, metrics, sorts, ...). Is it possible to destructure it into smaller functions which are then passed as `validate` props to the `FieldArray` components?

This is not a blocker for this PR, but might be an option for a future PR.

linuspahl:
@dennisoelkers The `FieldArray` component does not have a `validate` prop. Of course we could achieve this by implementing the validation for the `Field` components, but I prefer using the `Formik` validate method. As we have seen in the past, using only the `Formik` validate method makes it easier to combine different validation rules and to include other field values. The good news, executing our current validation needs very little time.

dennisoelkers (approved):

PR #10575, opened 2021-05-05
Fixing message loading when testing message against stream.

### LINKED ISSUE #10545: Search: Test Against Stream option renders error page
## Expected Behavior

- Open a search result item
- Click `Test Against Stream`
- Choose stream to test against
- Receive UI with results

## Current Behavior

After choosing the stream to test again, an error is displayed `e.submitButton.click is not a function`

## Possible Solution

`MessageLoader.jsx` is the component with `this.submitButton.click();`

This is referenced from `LoaderTabs.jsx:56` which is actually calling `this.messageLoader.submit(messageId, index);` on componentDidMount

1.Open a search result item
2.Click `Test Against Stream`
3.Choose stream to test against
4.Error Page

dennisoelkers:
## Motivation and Context

As reported in #10545, testing a message against a stream results in an exception being thrown. This is caused by a prior change that changes the ref target which is used to submit a form.

This PR is changing the way the message is loaded. Instead of triggering a synthetic click event on a submit button, the function which loads messages is called directly.

In addition, the `MessageLoader` component is migrated to TypeScript and a functional component, cleaning up its inner workings, hopefully making it more robust. The `submit`-function which was available on the component was removed (calling functions on components is considered an anti-pattern), instead it is now possible to pass `messageId` and `index` props for the component. If they are provided, the inputs are pre-populated with their values and the corresponding message is fetched.

Fixes #10545.

## Types of changes

## Checklist:


edmundoa:
@dennisoelkers I think we introduced some conflicts when merging the cloud changes into master, would you mind rebasing the PR? Thank you!

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <div className="message-loader">
+      {hideText || (
+        <p>
+          Wrong example? You can{' '}
+          <Button bsStyle="link" bsSize="sm" onClick={toggleMessageForm}>load another message</Button>.
+        </p>
+      )}
+      {hidden || (
```
edmundoa, line 113: Using `hidden` will not allow users to display the form when clicking on "Load another message", I think this line should use `isHidden` instead:

```suggestion
      {isHidden || (
```

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    loadMessage(messageId, index);
+  }, [index, loadMessage, messageId]);
+
+  const toggleMessageForm = useCallback(() => {
+    const newIsHidden = !isHidden;
+    setIsHidden(newIsHidden);
+
+    if (!newIsHidden) {
+      messageIdRef.current?.focus?.();
```
edmundoa, line 101: This unfortunately does not work as expected. When this code runs, the `input` component is not rendered and its `ref` is therefore not set.

Before, we focussed the input after the state was set and the component was updated:

```
    this.setState({ hidden: !this.state.hidden }, this._focusMessageLoaderForm);
```

I think the best approach would be to move this if statement into its own `useEffect`, where it can run when `isHidden` changes.

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
```diff
@@ -0,0 +1,134 @@
...
+    </form>
+  </div>
+);
+
+type Props = {
+  hidden: boolean,
+  hideText: string,
+  onMessageLoaded: (data: any) => void,
+  messageId: undefined | string,
+  index: undefined | string,
```
edmundoa, line 50: Do we also want to include these new props in the prop types definition? I would also suggest setting their default values to `''`, so we don't see these warnings when editing the inputs:

```
Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

edmundoa: By the way, while playing with the component I noticed that with the current type definition, IntelliJ believes that both `messageId` and `index` are mandatory props:

Screenshot 2021-05-17 at 12 20 04


When I changed the type definition to the one below, the warning was gone and ESLint also complained that those optional props should have a default value, so I think that may be a slightly better option for setting optional props. What do you think?

```suggestion
  messageId?: string,
  index?: string,
```

edmundoa (changes requested):
I added a couple inline issues and comments. Other than that, the changes fix the original issue and I like the more modular design 👍

graylog2-web-interface/src/components/messageloaders/LoaderTabs.jsx
```diff
@@ -134,12 +129,14 @@ class LoaderTabs extends React.Component {
...
     if (this._isTabVisible('messageId')) {
+      const { messageId, index } = this.props;
+
       messageLoaders.push(
         <Tab key="messageId" eventKey={this.TAB_KEYS.messageId} title="Message ID" style={{ marginBottom: 10 }}>
           <div style={{ marginTop: 5, marginBottom: 15 }}>
             Please provide the id and index of the message that you want to load in this form:
           </div>
-          <MessageLoader ref={(messageLoader) => { this.messageLoader = messageLoader; }}
+          <MessageLoader messageId={messageId} index={index}
```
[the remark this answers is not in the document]
dennisoelkers: Thanks. I have no idea why eslint did not fix this on save :/

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
```diff
@@ -0,0 +1,134 @@
...
+      messageIdRef.current?.focus?.();
+    }
+  }, [isHidden]);
+
+  return (
+    <div className="message-loader">
+      {hideText || (
+        <p>
+          Wrong example? You can{' '}
+          <Button bsStyle="link" bsSize="sm" onClick={toggleMessageForm}>load another message</Button>.
```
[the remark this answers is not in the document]
dennisoelkers: Way better!

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
dennisoelkers: :white_check_mark:

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
dennisoelkers: :white_check_mark:

graylog2-web-interface/src/components/extractors/MessageLoader.tsx
dennisoelkers: :white_check_mark:

edmundoa (approved):
Looks good to me, thank you for the fix and refactoring! 👍

Issue #10168, opened 2021-03-01
Updating Output Configuration Breaks Setup


phaabe:
After a change in the IP address in an TCP output configuration Graylog seems to stop working correctly (the visualization stops showing messages). The CPU consumption jumps up to 200% (as it is usually around 10%).
Only a restart of the Graylog service makes it work with the new configuration.

## Expected Behavior
Updating the output configuration shouldn't affect much except forwarding. 

## Current Behavior
Visualization stops showing messages.

## Your Environment

* 2 VCPU
* 4 GB RAM
* ~ 300k GELF messages per day (about half of them are forwarded through the specific output)

## Conclusion
I don't think this is normal behavior. Let me know if I can provide more information or if there is something wrong in principle with this setup.

diesaulov:
Hi, @phaabe! Thank you for reporting the bug. May we ask to provide further details:

1. the version of Graylog in use at the time
2. values of the following properties from `server.conf`: `output_module_timeout` and `ring_size`
3. for how long did you wait before restarting the server?
4. ideally would be great to see server logs starting from the moment when you change the Output configuration until the restart

BR,
Danylo

PR #10584, opened 2021-05-06
Blocked OutputBufferProcessor on Output update / removal

diesaulov:
## Description
This PR simply counts down the latch where it was missed to do so.

## Motivation and Context
During investigation of https://github.com/Graylog2/graylog2-server/issues/10168 it turned out
that, IF at least one of non-default Outputs was requested to shut down,
THEN all available instances of `OutputBufferProcessor` get blocked (for
the period configured in `output_module_timeout`) while waiting for the
non-default Outputs to hand over a message down the flow.

It causes troubles in the following discovered scenarios:

 1) an Output is deleted from a stream
 2) Output's configuration is changed, what implies Output's restart

As result it leads to the default Output (ES Output) getting blocked
until the lock is released after reaching `output_module_timeout`. It
explains 'the visualization stops showing messages' mentioned in the
reported bug (no new messages in Elastic Search). Also if ring buffer's
capacity of OutputBuffer is fully exhausted, one can observe high CPU
utilization due to the blocked pipeline.

TODO:

 1) Follow up with the bug reporter and sort out values of the following
 config props: `ring_size` and `output_module_timeout`
 2) Discuss whether it's a good idea to keep communication with the
 default and customer-provided Outputs in the same thread

## How Has This Been Tested?
The changes have been tested manually. Covered the aforementioned 
scenarios of Output removal and Output's configuration update.

To achieve the state of the blocked pipeline the following configuration
was used:
* `output_module_timeout` = 120000
* `ring_size` = 8192

Apart from that, the inputs have to be heavily loaded (tested with 5 
parallel processes sending logs in a loop with no delay).

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/buffers/processors/OutputBufferProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog2/buffers/processors/OutputBufferProcessor.java
@@ -171,10 +171,12 @@ public class OutputBufferProcessor implements WorkHandler<MessageEvent> {
     private Future<?> processMessage(final Message msg, final MessageOutput output, final CountDownLatch doneSignal) {
         if (output == null) {
             LOG.error("Output was null!");
+            doneSignal.countDown();
             return Futures.immediateCancelledFuture();
         }
         if (!output.isRunning()) {
             LOG.debug("Skipping stopped output {}", output.getClass().getName());
+            doneSignal.countDown();
             return Futures.immediateCancelledFuture();
         }
 


danotorrey:
I recall experiencing a problem before where an edit to a Forwarder Output resulted in a failed output restart. Then, the Output system became blocked indefinitely until a server restart. I suspect the cause was in this area of the code, but I've not had a chance to test this PR yet.

mpfz0r (approved):
I'm not completely sure that this fixes the issue the customer was seeing.
It sounded like he could only resolve it with a server restart.
If the default timeout were used, it should be resolved after 10 seconds.

However, this change is definitely correct by itself :+1: 
Good job

diesaulov:
> 
> 
> I'm not completely sure that this fixes the issue the customer was seeing.
> It sounded like he could only resolve it with a server restart.
> If the default timeout were used, it should be resolved after 10 seconds.
> 
> However, this change is definitely correct by itself 👍
> Good job

Me too. I will follow up with the reporter the values of `output_module_timeout ` and `ring_size `.

Issue #9554, opened 2020-11-19
Lookup tables - `purge key` should purge it from all cluster members


jalogisch:
## Expected Behavior

When purging a key from the caches in the lookup tables that should happen on all nodes and not only the one you are connected too.

![image](https://user-images.githubusercontent.com/404238/99664170-17407e00-2a68-11eb-91f2-458d401b98c3.png)

That is a potential problem as it is not clear that this is done only on the node you are connected too and not on all members of the cluster. When you are using a cache that is used by all cluster members that is not a problem, but if you have only local mem caches this can be a problem. 

## Current Behavior
Deletion is only done on the node that you are connected.

## Possible workround
Change the settings of the cache will end in a restart of the caches and with that a purge of all cached content. 
Purging the key via API endpoint that runs to all nodes in the cluster.

## Context
As long as there is no cache that is used by all cluster members this can cause Graylog nodes to have different results from DSV lookups.

l0calhost:
+1, i currently need to workaround with a shell script manually deleting on every single cluster node.

PR #10605, opened 2021-05-10
[issue-9554] Lookup tables - purge key should purge it from all cluster members

diesaulov:
## Description

To make cluster-wide Lookup Table Cache purging possible a new API endpoint was introduced in scope of this PR. The FE was adjusted accordingly, however error handling could be improved.

## Motivation and Context
This is a fix for https://github.com/Graylog2/graylog2-server/issues/9554

## How Has This Been Tested?

Manually verified:
- Unauthorized access to the new API endpoint
- Cache purging in multi-server environment

Unit tests:
- Covered `ClusterLookupTableResource` with unit tests

An example of response with successful API calls:
```
> POST /api/cluster/system/lookup/tables/Test/purge HTTP/1.1
< HTTP/1.1 200 OK
{
  "ae90e60b-2598-4d93-9e69-d35ba556849c": {
    "server_error_message": null,
    "response": {
      "code": 204,
      "entity": null,
      "success": true,
      "error_text": null
    },
    "call_executed": true
  },
  "356cacbe-b107-41d0-80c6-9ab8f91d7db6": {
    "server_error_message": null,
    "response": {
      "code": 204,
      "entity": null,
      "success": true,
      "error_text": null
    },
    "call_executed": true
  }
}
```

An example of response with failed API calls:
```
> POST /api/cluster/system/lookup/tables/Test123/purge HTTP/1.1
< HTTP/1.1 200 OK
{
  "ae90e60b-2598-4d93-9e69-d35ba556849c": {
    "server_error_message": null,
    "response": {
      "code": 404,
      "entity": null,
      "success": false,
      "error_text": "{\"type\":\"ApiError\",\"message\":\"Lookup table <Test1> not found\"}"
    },
    "call_executed": true
  },
  "356cacbe-b107-41d0-80c6-9ab8f91d7db6": {
    "server_error_message": null,
    "response": {
      "code": 404,
      "entity": null,
      "success": false,
      "error_text": "{\"type\":\"ApiError\",\"message\":\"Lookup table <Test1> not found\"}"
    },
    "call_executed": true
  }
}
```
An example of response with server error:
```
{
  "ae90e60b-2598-4d93-9e69-d35ba556849c": {
    "server_error_message": "Some server error!",
    "response": null,
    "call_executed": false
  },
  "356cacbe-b107-41d0-80c6-9ab8f91d7db6": {
    "server_error_message": null,
    "response": {
      "code": 204,
      "entity": null,
      "success": true,
      "error_text": null
    },
    "call_executed": true
  }
}
```

## Remarks
- Further improvements: https://github.com/Graylog2/graylog2-server/issues/10622

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterLookupTableResource.java
```diff
@@ -0,0 +1,114 @@
...
+    @POST
+    @Timed
+    @Path("tables/{idOrName}/purge")
+    @ApiOperation(value = "Purge Lookup Table Cache on the cluster-wide level")
+    @NoAuditEvent("Cache purge only")
+    @RequiresPermissions(RestPermissions.LOOKUP_TABLES_READ)
+    public List<LookupTableCachePurgingNodeResponse> performPurge(
+            @ApiParam(name = "idOrName") @PathParam("idOrName") @NotEmpty String idOrName,
+            @ApiParam(name = "key") @QueryParam("key") String key) {
+
```
mpfz0r, line 87: I think this can largely simplified by using `getForAllNodes` from the `ProxiedResource`
Kind of like:
```
+        return getForAllNodes(remoteResource -> remoteResource.performPurge(idOrName, key), createRemoteInterfaceProvider(RemoteLookupTableResource.class));
```
This means the response also has to change.
I'd suggest this is good enough for our purpose:
```
    public Map<String, Optional<Boolean>> performPurge(
```
We can change the `LookupTableResource` endpoint to return a `boolean` instead of a `void`

mpfz0r (changes requested):
The approach itself is good.
But we can do this without introducing so much new code.
-> See inline comment

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
```diff
@@ -0,0 +1,44 @@
...
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog2.rest.resources.system.responses;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class LookupTableCachePurgingNodeResponse {
```
bernd, line 22: We usually use auto-value based classes for these kind of objects. It would be great if you can switch this class over to an auto-value. Thanks!

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
```diff
@@ -0,0 +1,44 @@
...
+    public final String nodeId;
+    public final boolean success;
+    public final String message;
+
+    @JsonCreator
+    private LookupTableCachePurgingNodeResponse(
+            @JsonProperty("success") boolean success,
+            @JsonProperty("nodeId") String nodeId,
```
bernd, line 30: We use `snake_case` instead of `camelCase` for our field names in JSON. So this should be `node_id`.

bernd (changes requested):
Thanks for the PR! :slightly_smiling_face: 

Is there a specific reason why you use an additional cluster resource instead of creating an event via the cluster event bus? That way we wouldn't need an additional resource class that fans out to all other nodes and the existing `/purge` endpoint would do the right thing. With a cluster event it's also easier to trigger a cluster wide cache purge from within the system. (we need this in an upcoming enterprise change)

mpfz0r:
> Thanks for the PR! slightly_smiling_face
> 
> Is there a specific reason why you use an additional cluster resource instead of creating an event via the cluster event bus? That way we wouldn't need an additional resource class that fans out to all other nodes and the existing `/purge` endpoint would do the right thing. With a cluster event it's also easier to trigger a cluster wide cache purge from within the system. (we need this in an upcoming enterprise change)

The advantage of having a cluster resource is the possibility to respond with a meaningful status.
The cluster event is just fire and forget.
Besides, I think Rob is already working on the cluster event for his PR.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
mpfz0r: If we use what I proposed earlier, we wouldn't need a new Response class

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterLookupTableResource.java
diesaulov: That was the first thing I had tried, but decided to retreat because a) it entails an API change b) `getForAllNodes` yields the same result for `No Content` responses and for failed requests.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
diesaulov: But does it really make sense to use `@AutoValue` for such simple objects like API Request/Response? There is no need in generating a `hashCode`, `equals` and `toString`.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
diesaulov: 1f1e1810e759d42ac57c9e3956202c653e2c9ba7

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterLookupTableResource.java
diesaulov: On top of that with this custom approach we provide more fine-grained error messages.

mpfz0r: for a) it should be fine if we go from `no content` to returning a response.

I agree that it's not as fine-grained, but still good enough.
It also uses a established response for cluster api calls.

diesaulov: As agreed here is a generic solution for future generations: 2633ea14194847156939d7f1fb3381cb586364d6

graylog2-server/src/main/java/org/graylog2/rest/resources/system/responses/LookupTableCachePurgingNodeResponse.java
diesaulov: ﻿Not relevant anymore: ﻿﻿https://github.com/Graylog2/graylog2-server/commit/2633ea14194847156939d7f1fb3381cb586364d6

bernd (approved):
Thanks for the PR updates! :slightly_smiling_face: 

I personally think we should refactor this at some point to ensure that the regular lookup table endpoint purges the caches on all nodes correctly instead of relying on a cluster version of that endpoint and having an endpoint that shouldn't be called directly. Also we now have two ways of purging the caches on all nodes, one by calling the regular lookup table endpoint on all nodes via the cluster endpoint and a cluster event used by internal systems.

Since we are close to the 4.1 release, I think we should go ahead an merge the current state and do the refactoring later.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -135,37 +204,80 @@ protected ProxiedResource(@Context HttpHeaders httpHeaders,
...
+        @JsonProperty("success")
+        public abstract boolean isSuccess();
+
+        @JsonProperty("server_error_message")
+        @Nullable
+        public abstract String serverErrorMessage();
+
+        @JsonProperty("response")
+        @Nullable
+        public abstract MasterResponse<ResponseType> response();
```
mpfz0r, line 243: I would rename `MasterResponse` to something like `NodeResponse`

It is not just from master nodes.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
```diff
@@ -135,37 +204,80 @@ protected ProxiedResource(@Context HttpHeaders httpHeaders,
...
+    }
+
+    /**
+     * This wrapper is intended to provide additional server error information
+     * if something went wrong beyond the actual API HTTP call.
+     */
+    @AutoValue
+    public static abstract class CallResult<ResponseType> {
+
+        @JsonProperty("success")
```
mpfz0r, line 234: This is bit misleading, because we have two keys called `success` now

```{
    "cbb04ecf-ced3-47a5-9573-4f83b2dce963": {
        "response": {
            "code": 404,
            "entity": null,
            "error_text": "{\"type\":\"ApiError\",\"message\":\"Lookup table <609aa4683373f118e6dc17ed> not found\"}",
            "success": false
        },
        "server_error_message": null,
        "success": true
    }
}
```

maybe we could call this `request_delivered` instead?

mpfz0r (changes requested):
Two small things. Rest works and LGTM :+1:

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
diesaulov: Yes. Initially it was called `NodeResponse`, but later I had to revert it not to break plugins using this class.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
diesaulov: https://github.com/Graylog2/graylog2-server/pull/10605/commits/e6fdd831b6e5deffc17c2168651afb5614627012

mpfz0r (approved):

PR #10612, opened 2021-05-11
Allow emptying relative range value inputs in date time picker when entering new value.

linuspahl:
## Motivation and Context

Before this change it was not possible to temporary empty the relative
range value inputs. When a user tried to remove all numbers, we
automatically inserted a 1. We mainly did this because we are always
updating the form state `onChange`. If we would define for example
`undefined` for the relative range value, we would loose the
information which range type is selected. The range type is
being inferred based on the time range form value.

For example if the form state value for the relative range is 300, the
`RelativeRangeSelect` displays: 5 (range value) minutes (range type).

With this PR we are implementing a separate state for the `RelativeRangeSelect` which controls the
relative range inputs. We are still updating the form state
`onChange`. Every time the form state changes we are synchronizing the
`RelativeRangeSelect`. When a user empties the range value input, we are
defining `null` for the range form state. By implementing a separate state, we
are able to remember the previous selected range type and consider the
type when the user defines a completely new range value.

Please note, this PR is a suggestion for a short-term solution, long-term we should split up the relative range states in the date time picker form and use e.g. `from: { value: 5, type 'minutes' }` instead of `from: 300` (seconds) and justmigrate the value when opening or submitting the date time picker. This way we do not need to manage two states.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/date-time-picker/RelativeRangeSelect.tsx
+++ b/graylog2-web-interface/src/views/components/searchbar/date-time-picker/RelativeRangeSelect.tsx
@@ -15,6 +15,7 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
+import { useEffect, useState } from 'react';
 import { Field, useFormikContext } from 'formik';
 import styled, { css } from 'styled-components';
 import moment from 'moment';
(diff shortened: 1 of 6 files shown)


kmerz (approved):
Code LGTM :+1: 
What a relief! :heart:

PR #10619, opened 2021-05-12
Hide PlotLegend with only one element

### LINKED ISSUE #10581: Hide legend of widget visualization when aggregation has only one metric.
## What?
We should hide the widget visualization legend when the aggregation has only one metric.

## Why?
The widget description will provide enough information in most cases. This way the visualization itself will receive more space, which is especially helpful for widgets with a comparable small height.

kmerz:
## Motivation
Prior to this change, we always displayed the plot legend. But this is
taking valuable space and the description should already mention what is
displayed.

## Description
This change tests if there is only one element to display and hides the
plotlegend otherwise. If the widget is in edit mode the plot legend will always be shown so the user can edit the color.

Fixes #10581

## How Has This Been Tested?
- Created a aggregation with only one metric and row pivot
- Added another aggregation with one metric, a row pivot and a column pivot.

Graylog (9)

## Types of changes


dennisoelkers (changes requested):
For the record: @kmerz and me just had a quick chat and agreed upon doing the check slightly differently, by showing the legend if either more than one metric or at least one column pivot is present.

dennisoelkers (approved):
Looks good, works well! :tada:

dennisoelkers (approved):
Looks good, works great! :)

PR #10621, opened 2021-05-12
Change back REQUEST_HEADERS access to public

edmundoa:
In https://github.com/Graylog2/graylog2-server/pull/10578 we changed the `REQUEST_HEADERS` `public` access modifier to `private`, causing the cloud master build to fail, since one of the classes uses that constant in its code. This PR reverts that change to fix the build.


mpfz0r (approved):
Oh, good catch. Thanks! We'll need to fix that in 4.0 as well

mpfz0r:
@edmundoa or is this due to our cloud merge into master?
where do we use this?

PR #10625, opened 2021-05-12
Configurations style

edmundoa:
Some styling improvements in the Configurations page:

- Replace CSS file with styled-component
- Adds some extra margin to `dd` elements so their titles don't push them
- Adds some space after each configuration block, making it clearer where "update" buttons belong

Before:

2021-05-12 17 29 29 localhost d6da13e6bbf5

After:

2021-05-12 17 19 17 localhost 1887ed2203d2


graylog2-web-interface/src/components/indices/IndicesConfiguration.jsx
```diff
@@ -22,21 +22,12 @@ import { Row, Col } from 'components/graylog';
 import IndexMaintenanceStrategiesSummary from 'components/indices/IndexMaintenanceStrategiesSummary';
 import 'components/indices/rotation'; // Load rotation plugin UI plugins from core.
 import 'components/indices/retention'; // Load rotation plugin UI plugins from core.
-import style from 'components/configurations/ConfigurationStyles.lazy.css';
```
edmundoa, line 25: I looked through our rotation and retention plugins and could not find any of these styles being used in there, so I think we don't need the file anymore.

dennisoelkers (approved):
Looks way better. Thanks for the cleanup! :tada:

PR #10637, opened 2021-05-14
Better swagger specs through extracting inline models.

dennisoelkers:
## Description

This PR is aiming to improve the generated swagger specs by extracting inline models and recursively generate typings for them. Before this change, there were a couple of scenarios where we did not generate a schema for a used type, but instead a string literal was used:

**Before:**
post-parameter-before
before-roles

Through these changes, we can now resolve these type names and generate inline schemas which can be used to display the structure of a type:

**After:**
post-parameter-after

after-roles

This change does not make type introspection for swagger perfect, there are still remaining cases where no useful type information is handed out, but it improves significantly for the majority of cases.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/model/ModelId.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/model/ModelId.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.contentpacks.model;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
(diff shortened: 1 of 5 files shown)


bernd (approved):
I still hope we can switch to the official swagger generator at some point instead of putting more energy into our custom generator. :smile: This seems to improve the generator in some cases, so I think it should go in. :+1: Thank you!

dennisoelkers:
@bernd: For sure. I also looked again at switching over (and migrating to OpenAPI 3.0) before implementing this and considered it too much work for now. I would love to work on this in the near future with you though, if you are interested.

bernd:
@dennisoelkers Sounds good! @kroepke did some work on this recently so we can build on that.

PR #10644, opened 2021-05-17
Fix creatable select focus and "create value" option label problem.

linuspahl:
## Motivation and Context

With this PR we are fixing two problems with the createble select (`common/Select` component with `allowCreate` props):

-  Display "Create value" option label correctly when using `displayKey` select prop.
- Prevent loosing focus when removing content of creatable select input.
Before this change it was possible to loose the react creatable select focus
after removing its content, because we updated the component state unnecessarily.
With this change we ensure that we only update the state when necessary.

## Types of changes

## Checklist:


dennisoelkers (approved):
Looks good, works fine!

PR #10651, opened 2021-05-17
With configured Version-probing, wait for ES to come up if not available straight away

### LINKED ISSUE #10580: Continue probing for ES nodes instead of bailing out on startup
In the current state, the ES version probing bails out and exits the Graylog server if none of the configured ES nodes is reachable. This is not ideal for setups where race conditions exist that could cause the ES cluster to come up after the Graylog server (more input in #9550).

Instead, we should continue probing indefinitely. We could do this with a fixed or adaptive delay between each probing round.

janheise:
## Description
Adds two parameters to the Elasticsearch (ES) version probe:
- elasticsearch_version_probe_attempts
- elasticsearch_version_probe_delay

If Graylog can not connect to ES on the first try, it retries according to the given parameters.

## Motivation and Context
Especially in container environments, you can not always be sure that all services come up or are ready in a certain order. Graylog is resilient in regard to this problem regarding MongoDB and if you have disabled ES version probing. 
If you want to probe for the ES version and the ES-instance is not ready to connect, this PR adds that you can retry for a certain amount of tries and also specify the interval to wait between connection attempts.
fixes #10580

## How Has This Been Tested?
manually, so far

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchClientConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchClientConfiguration.java
@@ -48,11 +48,11 @@ public class ElasticsearchClientConfiguration {
     @Parameter(value = "elasticsearch_idle_timeout")
     Duration elasticsearchIdleTimeout = Duration.seconds(-1L);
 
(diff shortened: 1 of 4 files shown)


dennisoelkers (approved):

PR #10653, opened 2021-05-17
Use more secure TLS default ciphers

mpfz0r:
Our defaults (even with TLS 1.2 and 1.3 only)
were still susceptible to LOGJAM, LUCKY13 and a possible DoS attack.

We programatically use secure defaults for the JDK and for Netty with OpenSSL

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/2202


bernd (approved):
LGTM, thank you! :+1:

PR #10673, opened 2021-05-19
Prevent unnecessary search execution when unfocusing widget.

dennisoelkers:
## Motivation and Context

When introducing functionality, which limits search execution to the widget which is currently edited, a change was added that performs a search execution when leaving edit mode. Unfortunately the condition which was used to determine if widget editing is left was too broad, so it is performed whenever the widget focusing state changes. This leads to e.g. double search execution when loading a new/saved search.

This PR is changing the condition to check if the previous state indicates widget editing mode and does not perform a new search execution otherwise.

## Types of changes

## Checklist:


linuspahl (approved):
Works as expected.

Issue #10548, opened 2021-05-03
ContentPack download  from the list of ContentPacks does not work during dev on localhost


janheise:
## Expected Behavior
When you select "more actions"->"download" next to a ContentPack, you should be able to download the JSON using the "Save link as" option from the context menu on the link in the popup.

## Current Behavior
"Save link as" does not work during development.

## Steps to Reproduce (for bugs)

1. Select "Content Packs" from the "System" Menu
2. Select "more actions" on the right for any Content Pack
3. Select "download"
4. try to follow the download instructions in the Popup

## Context
As a developer, this is needed to check download/installation of ContentPacks

## Your Environment

* Graylog Version: 4.1-Snapshot
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System: Linux
* Browser version: Chrome, Chromium, FireFox

bernd:
I can reproduce that. This would fix it, not sure if that's the right way, though. :wink: 

```diff
diff --git graylog2-web-interface/src/components/content-packs/ContentPackDownloadControl.jsx graylog2-web-interface/src/components/content-packs/ContentPackDownloadControl.jsx
index 4619780dac..36b05087b8 100644
--- graylog2-web-interface/src/components/content-packs/ContentPackDownloadControl.jsx
+++ graylog2-web-interface/src/components/content-packs/ContentPackDownloadControl.jsx
@@ -42,11 +42,15 @@ class ContentPackDownloadControl extends React.Component {
   _getDownloadUrl() {
     const { contentPackId, revision } = this.props;
 
-    const url = new URI(URLUtils.qualifyUrl(
+    let url = new URI(URLUtils.qualifyUrl(
       ApiRoutes.ContentPacksController.downloadRev(contentPackId, revision).url,
     ));
 
     if (URLUtils.areCredentialsInURLSupported()) {
+      // If we set username and password for a relative URL, we get a broken link. (e.g. "username:password@/api/...")
+      if (url.is('relative')) {
+        url = url.absoluteTo(location.href);
+      }
       url
         .username(SessionStore.getSessionId())
         .password('session');
```

PR #10675, opened 2021-05-20
Fix and centralize qualifying relative URLs with credentials.

dennisoelkers:
## Motivation and Context

We are adding credentials to backend API-links in a couple of places to allow the browser to retrieve files directly from the backend. This happens by setting the session id as username, the literal string `session` as password.

This works fine, unless the backend API prefix is a relative URL. In this case invalid URLs are being generated. This makes it impossible for developers to test functionality like this using e.g. `devServer.js`.

This PR is changing this by:

  - checking if `AppConfig.gl2ServerUrl()` is a relative URL. If it is, it will be qualified by using scheme, hostname & port from `window.location
  - creating a `qualifyUrlWithSessionCredentials` function which can be used to augment any given backend API endpoint with current session credentials. This allows us to centralize this repeated pattern.

In addition, the PR is migrating `URLUtils` to TS and adds tests.

Fixes https://github.com/Graylog2/graylog2-server/issues/10548

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -130,6 +130,7 @@
     "@types/plotly.js": "1.54.8",
     "@types/react-beautiful-dnd": "^13.0.0",
     "@types/react-plotly.js": "^2.2.4",
+    "@types/urijs": "^1.19.15",
     "assets-webpack-plugin": "^5.1.2",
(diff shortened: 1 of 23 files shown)


linuspahl (approved):
Looks good to me!

PR #10676, opened 2021-05-20
Centralize reused spacings in theme.

linuspahl:
## Motivation and Context

This PR is creating a new section for the styled components theme called
spacings. By centralizing the spacings in the theme we can easily reuse them whenever we need to. It will
also make it easier to define margins and paddings relative to our commonly used spacings.

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/theme/margins.ts
+++ b/graylog2-web-interface/src/theme/margins.ts
@@ -14,22 +14,26 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-// eslint-disable-next-line no-restricted-imports
-import { Row as BootstrapRow } from 'react-bootstrap';
-import styled, { css } from 'styled-components';
+import PropTypes from 'prop-types';
 
-export const RowContentStyles = css(({ theme }) => css`
-  background-color: ${theme.colors.global.contentBackground};
-  border: 1px solid ${theme.colors.variant.lighter.default};
-  margin-bottom: ${theme.margins.row}px;
-  border-radius: 4px;
-`);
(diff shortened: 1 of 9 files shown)


graylog2-web-interface/src/theme/margins.ts
```diff
@@ -0,0 +1,39 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import PropTypes from 'prop-types';
+
+export type Margins = {
```
linuspahl, line 19: Maybe you have a better name in mind than `margins`? I thought about others like `spacings` or `gabs`, but imo `margins` is more descriptive.

kyleknighted (changes requested):
I like this concept! I think `Spacing` or `Gap` is probably a better name so that way we’re not setting padding from something called margins

Also, would like to assess some generated responsive spacing so that way we can have either something like “spacing-sm, spacing-med” etc
Or we could go Tailwind-esk with `spacing-N` and generate a number of sizes from minuscule to huge that can be used

I would also like to avoid defining `px` within the css and instead have the theme handle the scale.

graylog2-web-interface/src/theme/margins.ts
linuspahl: I renamed the section to `spacings`.

linuspahl:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/10719

PR #10652, opened 2021-05-17
Improving `SearchBar` and `DashbordSearchBar` responsiveness.

linuspahl:
## Motivation and Context

This PR is Improving the `SearchBar` and `DashbordSearchBar` responsiveness. The changes mainly improve the presentation of the search / dashboard action buttons.

Before (search):
image

After (search):
image

Before (search with small resolution):
image

After (search with small resolution):
image

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #10683, opened 2021-05-21
TMP: Testing which change results in a failing PR check.

linuspahl:
## Description
This is a temporary branch to test which change of https://github.com/Graylog2/graylog2-server/pull/10652 results in a failing PR check.


linuspahl:
It looks like adding and removing the search bar `TopRow` component fixed the failing test.

PR #10684, opened 2021-05-21
Fix apache httpclient to version 4.5.13

thll:
Managing and fixing the http client to a specific version works
around resolving the dependency to different versions in plugins.

This would cause unexpected problems as described in
https://github.com/aws/aws-sdk-java-v2/issues/1795


mpfz0r (approved):
I think this is the best solution we have for now :+1: 
It would be great, if maven would allow us to just resolve dependencies to the newest available version :crying_cat_face:

PR #10686, opened 2021-05-21
Forwarder Plugin activation toggle

diesaulov:
Since currently there is no way to inject plugin's config into its plugin bean at the bootstrap phase (what's needed in its turn not to load some particular modules depending on the configuration), and it's only possible for the master `Configuration`, as a temporary solution it was decided to include this toggle property in `Configuration`. Later it's expected to be migrated into the according plugin config.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ExposedConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ExposedConfiguration.java
@@ -88,6 +88,9 @@ public abstract class ExposedConfiguration {
     @JsonProperty("gc_warning_threshold")
     public abstract String gcWarningThreshold();
 
+    @JsonProperty("forwarder_enabled")
+    public abstract boolean forwarderEnabled();
+
     public static ExposedConfiguration create(Configuration configuration) {
         return create(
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/ExposedConfiguration.java
```diff
@@ -88,6 +88,9 @@
     @JsonProperty("gc_warning_threshold")
     public abstract String gcWarningThreshold();
 
+    @JsonProperty("forwarder_enabled")
+    public abstract boolean forwarderEnabled();
```
mpfz0r, line 92: Do we actually need to expose that flag in the configuration API?
I think just having `isForwarderEnabled()` should be fine

diesaulov: For the frontend to be able to decide whether to render/load or not related UI elements/code.

mpfz0r: Ah, makes sense :+1:

mpfz0r (approved):

PR #10687, opened 2021-05-21
Make all WildcardPermission case sensitive

mpfz0r:
When introducing the new permissions system in 4.0
we started using object instead of string permissions.

We did not notice that WildcardPermissions are by default case insensitive.
For most cases that didn't make a difference,
but returning the self-editing permissions to a user with a
non-lowercase name would change from something like

`users:edit:HORST` to `users:edit:horst`

Which would break the frontend permission checks
for these cases.

-> Introduce a new `CaseSensitiveWildcardPermission` class
   that changes this default.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10700

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2257


mpfz0r:
@bernd can you find time to look at this?
I think this should go in before testing day

bernd (approved):
LGTM :+1:

Issue #10665, opened 2021-05-18
Add cluster wide cachePurge() support


mpfz0r:
## What?
Add support to `https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/lookup/CachePurge.java

to send purge events to the ClusterEventBus and cause a cluster wide cache invalidation.
E.g. `purgeAllClusterWide()`

## Why?
For DataAdapters where only the instance on the master node is
regularly refreshing the contents[1], we need a way to invalidate the lookuptable caches on all cluster nodes.
Otherwise the non-master nodes in the cluster would still run with stale caches and report false lookup values.

[1] Those DataAdapters don't exist yet, but are in the making

PR #10692, opened 2021-05-24
Add cluster-wide lookup table cache purge

waab76:
## Description
As detailed in #10665, for DataAdapters where only a single instance is regularly refreshing the contents, we need a way to invalidate the lookuptable caches on all cluster nodes. Otherwise the non-master nodes in the cluster would still run with stale caches and report false lookup values.

This change adds that capability with the `AdapterSharedStoresUpdated` event type.

## How Has This Been Tested?
This has been tested in my local development environment.  Verified that when a Data Adapter sends an `AdapterSharedStoresUpdated` event on the event bus, all other nodes clear the appropriate lookup table caches.

## Types of changes

## Checklist:


mpfz0r (approved):
LGTM

PR #10697, opened 2021-05-25
Fix widget config form validation on aggregation element remove.

### LINKED ISSUE #10679: Aggregation builder form validation not being executed after removing an aggregation element.
## Description
We are always validating the aggregation builder form after a change. Currently this is not the case when removing an aggregation element (one exception is the `GroupByElement`).
This behaviour is related to the following change: https://github.com/Graylog2/graylog2-server/pull/10574

 On possible solution is replacing `FomikArray.remove()` with `Formik.setFieldValue()`.
 
This problem can be reproduced by creating e.g. a bar chart and removing the metric.

* Graylog Version: current development version for 4.1

linuspahl:
## Motivation and Context

Before this change the widget config form validation was not being
executed when removing an aggregation element. This is only the case
when we define `validateOnChange` for the related `FormikArray` component.
We recently disabled `validateOnChange` for the `FormikArray`s (#10574), because a side effect
was that the validation is being executed twice, when changing a related
input. Because of this conditions the most pragmatic approach is using
`setFieldValue` instead of the `FomikArray` method `remove`. I also
tried defining a `validationSchema` for the form, but it has the same
side effects.

Fixes #10679

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/aggregationwizard/aggregationElements/MetricElement.ts
+++ b/graylog2-web-interface/src/views/components/aggregationwizard/aggregationElements/MetricElement.ts
@@ -101,6 +101,10 @@ const MetricElement: AggregationElement = {
   }),
   toConfig: (formValues: WidgetConfigFormValues, configBuilder: AggregationWidgetConfigBuilder) => configBuilder
     .series(metricsToSeries(formValues.metrics)),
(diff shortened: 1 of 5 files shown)


kmerz (approved):
Code LGTM :+1: 

PR #10699, opened 2021-05-25
Feature/full backend tests/expose docker network

thll:
Exposes the docker network that is used for the backend test containers.

This is useful if a test needs to spin up an additional container which has to be reachable by the Graylog server.


mpfz0r (approved):
LGTM

PR #10701, opened 2021-05-25
Don't skip assembly step in full backend tests

thll:
Skipping the assembly step improves test time but it's required to
properly build e.g. the jar of the `graylog-storage-elasticsearch7`
module.

Before this change you could end up with broken tests by performing the following steps:

1. `$ mvn clean compile -DskipTests=true -Dskip.web.build -Dforbiddenapis.skip=true -Dmaven.javadoc.skip=true`
1. Run `BackendStartupIT` from your IDE

You'll notice that the test fails because the server doesn't start. The logs show the following error:
```
java.lang.ClassNotFoundException: org.graylog.shaded.elasticsearch7.org.elasticsearch.action.search.SearchResponse
```


bernd (approved):
Oh, too bad. :smile: LGTM :+1:

PR #10703, opened 2021-05-25
Fix permission check for notifications in event definitions for non alert managers

FlorinPeter:
## Description
Users that do not have **Alerts Manager** role assigned but have instead **Event Definition Creator** and **Event Notification Creator** assigned are not able to choose a notification because the permission **eventnotifications:read** is missing.

## Motivation and Context
As both roles (**Alerts Manager** and **Event Notification Creator**) have the permission **eventnotifications:create** this will allow users with with the roles **Event Definition Creator** and **Event Notification Creator** to choose a notification

## How Has This Been Tested?
1. create user with roles Reader, User Inspector, Event Definition Creator and Event Notification Creator
2. create a notification
3. create an event definition
4. choose a notification

## Types of changes

Needs backport to 4.0.x

### DIFF
--- a/graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationsForm.jsx
+++ b/graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationsForm.jsx
@@ -89,7 +89,7 @@ class NotificationsForm extends React.Component {
       );
     }
 
-    if (!isPermitted(currentUser.permissions, 'eventnotifications:read')) {
+    if (notifications.length < 1) {
       return (
         <Row>
           <Col md={6} lg={5}>


FlorinPeter:
@kmerz can you please have a look?

kmerz:
Hey Peter,

first of all thank you for your contribution! 

The bug you identified is totally valid. But the quick fix you suggest is not going far enough. Instead of changing the permission check should rather change the permission for **Event Definition Creator** and give him the missing `eventnotifications:read` permission. That would need a migration. So if you want you could do that, otherwise it would be nice if you could open up an issue and we will take a look into it soon.

FlorinPeter:
@kmerz maybe I am wrong but if we grant the the permission ```eventnotifications:read``` to the role **Event Definition Creator** then every user that owns that role will see other users ```eventnotifications``` and this was not my target as we are quite happy with the new rights model that was introduced with 4.x.
It is totally fine if an admin can see all ```eventnotifications``` but if a user did not explicit share the ```eventnotification``` with other users then it should not be visible to others, or am I wrong?

bernd:
@FlorinPeter @kmerz Yeah, we cannot assign `eventnotifications:read` to "Event Definition Creator" because that would allow users with that role to read **all** notifications, as @FlorinPeter said.

I think we can just do the following:

```diff
--- graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationsForm.jsx
+++ graylog2-web-interface/src/components/event-definitions/event-definition-form/NotificationsForm.jsx
@@ -89,7 +89,7 @@ class NotificationsForm extends React.Component {
       );
     }
 
-    if (!isPermitted(currentUser.permissions, 'eventnotifications:read')) {
+    if (notifications.length < 1) {
       return (
         <Row>
           <Col md={6} lg={5}>
```

(untested)

The `notifications` variable contains all the notifications that the calling user has access to. This is ensured by the server API resource. That means if there are notification objects in `notifications`, the active user has access to them and we can show them.

What do you think? Do I miss anything?

FlorinPeter:
@bernd thx for picking this up, I tested your suggested change based on 4.0.8 and it works as expected.
@kmerz I updated my PR to the suggested change of @bernd 
I am looking forward to see this PR got merged if there is nothing in addition from your side.

kmerz (approved):
Code LGTM :+1: 

Thanks a lot for your contribution!

PR #10708, opened 2021-05-25
Migrate Okta auth backend to enterprise plugin

patrickmann:
## Description
Allow on-prem Enterprise users to login and authenticate with OIDC/Okta, in addition to the existing methods (LDAP, AD, MongoDB).

### frontend
- Added the ability to bind Okta backend to available backends next to Directory based auth backends
- Fetches current active auth backend before login to check which form to render
- If a form is registered on the active backend then it will show it
- Ability to switch between active auth login form or default in UI

### backend
A new, unauthenticated endpoint allows the UI to obtain the necessary data to perform the OIDC auth flow.

## Motivation and Context
Cloud already supports OIDC/Okta for both authentication and team sync. In this first step we are migrating the authentication feature to on-prem Enterprise users, to improve consistency between the versions.

## How Has This Been Tested?
- unit tests
- manual API testing
- local UI testing
It is important to test this change in various plugin configurations
1. core: behavior should be unchanged, i.e. only LDAP/AD/DB are supported
2. enterprise: OIDC/Okta is available as an additional authentication service
3. enterprise + cloud: behaviour is unchanged from Cloud, i.e. OIDC/Okta is the only available authentication service

## Types of changes

## Checklist:

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2261


graylog2-web-interface/src/pages/LoginPage.jsx
```diff
@@ -33,9 +35,33 @@ const LoginPageStyles = createGlobalStyle`
...
+  const [activeBackend, setActiveBackend] = useState();
+  const [isDetermined, setIsDetermined] = useState(false);
+  const [useFallback, setUseFallback] = useState(false);
+
+  const isCloud = AppConfig.isCloud();
+  const registeredLoginComponents = PluginStore.exports('loginProviderType');
+  const loginComponent = registeredLoginComponents.find((c) => c.type === activeBackend);
+  const hasCustomLogin = loginComponent && loginComponent.formComponent;
+
+  useEffect(() => {
```
dennisoelkers, line 56: This works fine and there is nothing wrong with it. We could make it a bit simpler by using `react-query`. In this case it would look like this:

```
  const { data: activeBackend, isSuccess: isDetermined } = useQuery('activeBackendType', AuthenticationDomain.loadActiveBackendType);
```

I am mapping parts of the response to the names you used (`activeBackend`, `isDetermined`), but this is just for illustration and you can also use the more idiomatic name (`isSuccess`), plus you get more (`isError`/`error`/...).

As a plus, you could wrap this in a custom hook straight-away:
```
const useActiveBackendType = () => useQuery('activeBackendType', AuthenticationDomain.loadActiveBackendType);

[...]

const LoginPage = () => {
[...]
  const { data: activeBackend, isSuccess: isDetermined } = useActiveBackendType();
```

This could be reused in other places then as well.

Gaya: Thanks for the tip @dennisoelkers. Let me dig into this. Never used `react-query`, but it looks like how I'd do hooks that fetch data in my own projects.

Gaya: Updated with you suggestions @dennisoelkers

thll (approved):
Backend changes look good.

graylog2-web-interface/webpack/vendor-module-ids.json
```diff
@@ -1078,7 +1078,7 @@
       "node_modules/core-js/library/modules/_parse-int.js": 147,
       "node_modules/core-js/library/modules/_string-trim.js": 148,
       "node_modules/react-transition-group/utils/PropTypes.js": 251,
-      "node_modules/@babel/runtime/helpers/extends.js": 441,
+      "node_modules/@babel/runtime/helpers/extends.js": 440,
```
dennisoelkers, line 1081: This change should not be in here, afaics. There are no changes to packages included in the vendor bundle in this PR, so it should not change the vendor module ids.

dennisoelkers (changes requested):
See inline comment. Changes to `vendor-module-ids.json` in this PR should be reverted.

graylog2-web-interface/webpack/vendor-module-ids.json
edmundoa: Agreed, will take care of it 👍

dennisoelkers (approved):

graylog2-web-interface/src/pages/LoginPage.jsx
```diff
@@ -92,6 +113,11 @@ const LoginPage = () => {
         <LoginPageStyles />
         {formatLastError()}
         {renderLoginForm()}
+        {hasCustomLogin && !isCloud && (
+          <StyledButton as="a" onClick={() => setUseFallback(!useFallback)}>
+            {`Login with ${useFallback ? loginComponent.type.replace(/^\w/, (c) => c.toUpperCase()) : 'default method'}`}
```
edmundoa, line 118: A small thing with this is that it will show up while we log the user in using Okta. Right now I have no quick fix for it, so I'll create an issue for it.

graylog2-web-interface/src/pages/LoginPage.jsx
```diff
@@ -68,18 +99,16 @@ const LoginPage = () => {
   };
 
   const renderLoginForm = () => {
-    const loginComponent = PluginStore.exports('loginProviderType');
-
-    if (loginComponent.length === 1) {
-      return React.createElement(loginComponent[0].formComponent, {
+    if (!useFallback && hasCustomLogin) {
+      return React.createElement(loginComponent.formComponent, {
```
edmundoa, line 103: When an error occurs loading the component there's no way for the user to login back into Graylog. I'll write an issue for this.

edmundoa (approved):
Besides the two issues mentioned, I think this looks good now. Good job! 👍

PR #10715, opened 2021-05-26
Reset page when searching on steam overview

### LINKED ISSUE #10639: Searching from Streams pages > 1 results in "No streams found"
## Expected Behavior

When searching for streams from pages > 1, I expect that I'll be able to have results returned. 

## Current Behavior

Presently, when searching on any page > 1 from Streams, a message is returned: "No streams found"
CleanShot 2021-05-14 at 09 35 25

1. Run a 4.X instance
2. Create multiple streams
3. Navigate to page 2
4. Attempt to search for a any stream

See HS-411104332

kmerz:
## Motivation
Prior to this change, if a user was on page 2 on the stream overview and
he was searching for an item on the first page, nothing was shown.

## Description
This change we reset the page when sending the search request which is
the correct behavior.

Fixes #10639 

## How Has This Been Tested?
- Got to Streams Overview
- Go to the second page
- Search for a stream from the first page

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/streams/StreamComponent.jsx
+++ b/graylog2-web-interface/src/components/streams/StreamComponent.jsx
@@ -101,7 +101,7 @@ class StreamComponent extends React.Component {
 
   _onSearch = (query, resetLoadingCallback) => {
     const { pagination } = this.state;
-    const newPagination = Object.assign(pagination, { query: query });
+    const newPagination = { ...pagination, query: query, page: 1 };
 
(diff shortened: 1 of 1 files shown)


linuspahl (commented):
@kmerz Works as expected! What do you think about resetting the pagination when clicking on `Reset` as well? It would be consistent with lists we've added recently.

We only need to implement the same changes for `_onReset`.

linuspahl (approved):

PR #10562, opened 2021-05-04
Limit field types retrieval to timerange of query.

dennisoelkers:
## Motivation and Context

This PR is improving field types retrieval for the search. Prior to this PR, all field types for the current set of streams were retrieved. This could lead to issues when a field type has conflicting types across all indices, because restricting the time range to indices where the field type is consistent would not help, as the field type in the UI would still be a compound type.

This PR is now passing the current time range of the search/widget along with the field types retrieval query, so only relevant field types are retrieved. This helps with correctness _and_ performance of field types retrieval.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2219

PR #10729, opened 2021-05-27
Avoid fetching widget field types if not on dashboard.

dennisoelkers:
## Motivation and Context

In #10562, we introduced a more specific way to retrieve field types according to a timerange and stream selection. This is being used to fetch more precise field types for widgets on dashboards. Unfortunately, the `WidgetFieldTypesContextProvider` was used unconditionally to wrap all widgets, even if the current context is that of a Saved Search and not a Dashboard. On Saved Searches, field types are always the same due to the identical root query, therefore we do not need to fetch field types for each individual widget.

This PR is conditionally wrapping the widget component with the widget-specific field types provider if on a dashboard, and a fragment if on a search. This reduces the amound of overall field type requests.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/WidgetComponent.tsx
+++ b/graylog2-web-interface/src/views/components/WidgetComponent.tsx
@@ -14,7 +14,7 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-import React from 'react';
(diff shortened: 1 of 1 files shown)


kmerz (approved):
Fixes the problem :+1: 
Code LGTM :+1: 

PR #10730, opened 2021-05-27
Fix date format in outgoing traffic graphs

### LINKED ISSUE #10051: Change Outgoing Traffic in UI to use UTC instead of localtime for accounting
## Expected Behavior

When comparing outgoing traffic from the UI with traffic from `/api/system/cluster/traffic`, the values should be the same. 

## Current Behavior

Outgoing traffic in the UI is displayed using localtime vs `/api/system/cluster/traffic` which uses UTC. This leads to confusion when attempting to get those values to match up, as they never will.

## Possible Solution

Change the UI to use UTC for that specific part of the display instead of the user's localtime. 

See HS-315483976

kmerz:
## Motivation
Prior this change, we displayed the timestamps on the graph in the
timezone of the browser. That was because we gave plotly unix timestamps
without timezone information.

## Description
This change will use ISOString format to give plotly the desired
timezone (UTC). This leads to the effect that now time will be
displayed. Which is okay since it should describe a daily summary.

Fixes #10051

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/cluster/TrafficGraph.tsx
+++ b/graylog2-web-interface/src/components/cluster/TrafficGraph.tsx
@@ -37,7 +37,7 @@ const TrafficGraph = ({ width, traffic }: Props) => {
   const dailyTraffic = ndx.dimension((d) => moment(d.ts).format('YYYY-MM-DD'));
 
   const dailySums = dailyTraffic.group().reduceSum((d) => d.bytes);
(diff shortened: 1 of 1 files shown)


dennisoelkers (approved):
Works fine and improves things! :tada:

I have manually validated the daily summaries shown against the traffic records in my MongoDB. They are aggregated correctly and are aligned with the days as of UTC.

The only thing that's left - which is outside of the scope of this PR - is that the values are shown in powers of 10 and not of 2, while users would probably expect the latter. That means that a daily total of `209081770` bytes are shown as `209.1M` in the frontend, while users probably expect it to be shown as `199.4Mb`.

PR #10734, opened 2021-05-28
Fix request path matching in MessageExportFormatFilter

bernd:
The code accidentally matched API browser requests because the path
matching wasn't explicit enough.

This change adds the "api/" prefix to the checked target path and also
switches from "contains" to "endsWith".


dennisoelkers (approved):
Thanks for fixing the issue! :tada:

Issue #10671, opened 2021-05-19
Map center/zoom level of map visualization is not saved anymore


dennisoelkers:
Since the implementation of the new aggregation builder wizard, the map center/zoom level of a map visualization is not saved anymore.

PR #10736, opened 2021-05-28
Fixing world map viewport changes by maintaining related visualization config state in `WidgetConfigForm`.

linuspahl:
## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/10671 the center and zoom level of a map visualization is not being saved anymore, when editing the widget. This error occured because we were maintaining the complete widget config in the `WidgetConfigForm` state. The `WidgetConfigForm` was only responsible for the configuration of aggregation elements (left column in the aggregation wizard builder).

With this PR we are also wrapping the visualization with the `WidgetConfigForm` in the aggregation wizard builder.
The visualization component can update the visualization config by calling the provided `onVisualizationConfigChange` prop. 

We are also:
- implementing the following inputs for the world map visualiztion config form: zoom, center x, center Y.
- adjusting the previously used `AggregationControls` component to still support editing the world map using the old aggregation builder.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10671 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/visualizations/worldmap/bindings.tsx
+++ b/graylog2-web-interface/src/views/components/visualizations/worldmap/bindings.tsx
@@ -26,17 +26,29 @@ export type WorldMapVisualizationConfigFormValues = {
   centerY: number,
 };
 
+const DEFAULT_FORM_VALUES = {
+  zoom: 1,
+  centerX: -0.3515602939922709,
+  centerY: 0.703125,
+};
+
 const worldMap: VisualizationType = {
   type: WorldMapVisualization.type,
   displayName: 'World Map',
   component: WorldMapVisualization,
   config: {
-    createConfig: () => ({ zoom: 1, centerX: 0, centerY: 1 }),
-    fromConfig: (config: WorldMapVisualizationConfig) => ({
-      zoom: config?.viewport.zoom,
-      centerX: config?.viewport.center[0],
-      centerY: config?.viewport.center[1],
-    }),
+    createConfig: () => DEFAULT_FORM_VALUES,
+    fromConfig: (config: WorldMapVisualizationConfig) => {
+      if (!config) {
+        return DEFAULT_FORM_VALUES;
+      }
+
+      return {
+        zoom: config.viewport.zoom,
+        centerX: config.viewport.center[0],
+        centerY: config.viewport.center[1],
+      };
+    },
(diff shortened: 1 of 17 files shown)


kmerz:
As discussed lets catch the case where a WorldMapVisualizationConfig is `undefined`.

linuspahl:
@kmerz We threw the error when a previously configured world map widget had no config (custom viewport). We are now defining proper default form values in this case as well.

graylog2-web-interface/src/views/types.d.ts
```diff
@@ -114,6 +114,9 @@ type BooleanField = BaseField & {
 
 type NumericField = BaseRequiredField & {
   type: 'numeric',
+  componentProps?: {
```
dennisoelkers, line 117: I think we can flatten this. A `NumericField` can have an optional `step` key. It can be used inside the component rendering that field to pass that to the input. This way we have defined properties for the fields, which helps a lot when refactoring, because a user cannot just pass any prop to the field.

graylog2-web-interface/src/views/components/visualizations/worldmap/bindings.tsx
```diff
@@ -16,12 +16,68 @@
...
 
+export type WorldMapVisualizationConfigFormValues = {
+  zoom: number,
+  centerX: number,
+  centerY: number,
+};
+
+const DEFAULT_FORM_VALUES = {
+  zoom: 1,
+  centerX: -0.3515602939922709,
```
dennisoelkers, line 31: From what I can see, we can just use `0`/`0` for `centerX`/`centerY`, having the same result, but being slightly less magical.

dennisoelkers (changes requested):
Looks good, works great! Just two suggestions for minor improvements.

graylog2-web-interface/src/views/components/visualizations/worldmap/bindings.tsx
linuspahl: These are the default `react-leaflet` values, which we used in the past if the user did not changed the default viewport. But you are right, we can also just `0`/`0`, since the difference is not noticeable.

dennisoelkers (approved):
Looks good, works great! :tada: I love that you went the extra mile to add inputs for lat/long/zoom!

Issue #10583, opened 2021-05-06
Improve overflow of select dropdowns in aggregation builder.


linuspahl:
## What / Why?
Currently the select dropdowns in the aggregation element configuration column are not overflowing the parent container:
![image](https://user-images.githubusercontent.com/46300478/117324027-b3800500-ae8f-11eb-916b-35c95cd60fd5.png)

In most cases it would improve the user experience when the dropdown is being displayed above the parent container.

There is a good chance that this change will also fix the following presentational bug which can occur when both scroll bar are being used:

![image](https://user-images.githubusercontent.com/46300478/117325174-c515dc80-ae90-11eb-9fca-50f1f1c87ed5.png)

PR #10755, opened 2021-05-31
Improving overflow of select menus in aggregation elements column of aggregation wizard.

linuspahl:
**Please note, this Pr is based on https://github.com/Graylog2/graylog2-server/pull/10672 which needs to be merged first.**

## Motivation and Context

Before this change a menu of an element configuration select could
be cut off by their container. Opening a menu resulted in a scroll bar when
the select is at the bottom of the aggregation elements column and the
dropdown is too large.

image

With this change we are displaying the dropdown in a portal to display it
independently from the container. We are also defining the option `menuPlacement="auto"` for each aggregation element select, for a more dynamic placement.

image

I tried using aggregation wizard container for the select portals instead of using `document.body`. But it resulted in a strange placement of the menus
image

Using the aggregation wizard container would make the related component more encapsulated and easier to test. Currently wen need to define the following option `{ container: document.body };` when calling `select` from `react-select-event`.

Fixes https://github.com/Graylog2/graylog2-server/issues/10583

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.tsx
+++ b/graylog2-web-interface/src/components/common/Select.tsx
@@ -128,6 +128,11 @@ const menu = (base) => ({
   zIndex: 5,
 });
 
+const menuPortal = (base) => ({
+  ...base,
+  zIndex: 'auto',
+});
+
 const singleValueAndPlaceholder = ({ theme }) => (base) => ({
   ...base,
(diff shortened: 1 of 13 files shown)


dennisoelkers (approved):
Great change! Fixes the issue really well :champagne:

PR #10757, opened 2021-05-31
Fix handling of reopened indices and their aliases.

### LINKED ISSUE #10572: Can't get reopened indices with wildcards
The indices resource tries to find all reopened indices by using
a wildcard. E.g. `restored-archive-*reopened`

With ES6 this doesn't work at all, because the code resolves aliases manually and the match
is ignoring wildcards:
https://github.com/Graylog2/graylog2-server/blob/master/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java#L196

With ES7 this does work, because it uses the ES `_alias` api.
https://github.com/Graylog2/graylog2-server/blob/master/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java#L137
However, it triggers an exception as soon as there are multiple indices.
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/indexer/indices/Indices.java#L148

## Expected Behavior
Both ES versions should return all aliases for wildcards

## Current Behavior
ES6 doesn't
ES7 only works for one alias

## Possible Solution

Change the `resolveAlias()` API to reflect how ES aliases are structured.

1. Restore multiple archives
2. Call `https://127.0.0.1:9000/api/system/indexer/indices/reopened` endpoint
3. Either receive an exception or no result

dennisoelkers:
## Motivation and Context

Prior to this change, retrieving the index sets (and their wildcards) containing reopened indices was broken, due to a number of issues.

This PR is fixing:

  - retrieving the list of targets for an index wildcard
  - checking if an index set wildcard contains reopened indices

This unbreaks the indices overview page after multiple indices were restored.

Fixes #10572.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java
+++ b/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/IndicesAdapterES6.java
@@ -66,6 +66,7 @@ import org.graylog.shaded.elasticsearch5.org.elasticsearch.search.aggregations.b
 import org.graylog.shaded.elasticsearch5.org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.graylog.shaded.elasticsearch5.org.elasticsearch.search.sort.FieldSortBuilder;
 import org.graylog.shaded.elasticsearch5.org.elasticsearch.search.sort.SortBuilders;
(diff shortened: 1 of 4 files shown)


mpfz0r (approved):
Works and LGTM

PR #10768, opened 2021-06-02
Improve `suppressConsole` usage in frontend tests and fix two silently failing tests.

linuspahl:
## Motivation and Context

This PR is improving the `suppressConsole` usage in tests by:
- wrapping only the necessary code with `suppressConsole`. Before this change the `StreamPermissionErrorPage.test` and `ConfigurationsPage.test` failed silently because the `expect` call was part of the `suppressConsole` callback.
-  always returning a Promise when using `suppressConsole`. In the `ConfigurationsPage.test` we are throwing an error after changing the component state when rendering the component. To handle this error properly we have to wait with the reset of `console.error` until we display the error message.
- always using `await` when using `suppressConsole`. This is now necessary to ensure we are resetting `console.error` in time. But it also ensures that the test fails when we have to use a check (like `expect` or `findBy*`) in the callback of `suppressConsole` and the check fails.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/ConfigurationsPage.test.tsx
+++ b/graylog2-web-interface/src/pages/ConfigurationsPage.test.tsx
@@ -19,6 +19,7 @@ import { render, screen } from 'wrappedTestingLibrary';
 import asMock from 'helpers/mocking/AsMock';
 import suppressConsole from 'helpers/suppressConsole';
 import MockStore from 'helpers/mocking/StoreMock';
+import { configuration as mockConfiguration } from 'fixtures/configurations';
 
 import ConfigurationsPage from 'pages/ConfigurationsPage';
 import usePluginEntities from 'views/logic/usePluginEntities';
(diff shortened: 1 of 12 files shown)


graylog2-web-interface/src/pages/ConfigurationsPage.test.tsx
```diff
@@ -33,28 +34,44 @@ jest.mock('stores/decorators/DecoratorsStore', () => ({
...
+  listMessageProcessorsConfig: jest.fn(() => Promise.resolve()),
+}));
+
+jest.mock('stores/streams/StreamsStore', () => ({
+  StreamsActions: { listStreams: () => Promise.resolve() },
+}));
 
 const ComponentThrowingError = () => {
-  throw Error('Boom!');
+  throw new Error('Boom!');
```
dennisoelkers, line 52: Changing this is fine, but both are identical: `Thus the function call Error(…) is equivalent to the object creation expression new Error(…) with the same arguments.` ([Source](https://262.ecma-international.org/6.0/#sec-error-constructor))

graylog2-web-interface/test/helpers/suppressConsole.ts
```diff
@@ -23,14 +23,17 @@
...
  * @param {function} fn - the function which should be called after disabling `console.error` and before restoring it.
  */
-const suppressConsole = (fn: () => void) => {
+const suppressConsole = <T extends () => any>(fn: T): Promise<ReturnType<T>> => {
   /* eslint-disable no-console */
   const originalConsoleError = console.error;
   console.error = () => {};
 
-  fn();
+  return Promise.resolve(fn()).then((result) => {
```
dennisoelkers, line 31: What happens when the promise is rejected? From what I can see, `console.error` will not be reset in that case. Maybe use `finally`?

graylog2-web-interface/test/helpers/suppressConsole.ts
```diff
@@ -23,14 +23,17 @@
  *
  * @param {function} fn - the function which should be called after disabling `console.error` and before restoring it.
  */
-const suppressConsole = (fn: () => void) => {
+const suppressConsole = <T extends () => any>(fn: T): Promise<ReturnType<T>> => {
```
dennisoelkers, line 26: A slightly simpler way to type this would be:
```suggestion
const suppressConsole = <R>(fn: () => R): Promise<R> => {
```

dennisoelkers (changes requested):
Looks good, just a couple of nitpicks.

graylog2-web-interface/test/helpers/suppressConsole.ts
linuspahl: Good point.

dennisoelkers (approved):
Great work! :tada:

PR #10740, opened 2021-05-28
Allow quick configuration of search time range by implementing relative range preset select for all search bars.

linuspahl:
## Motivation and Context

While the recent update of the search date time picker improves the time range selection a lot, a user currently needs more clicks to change the time range. In many cases a user just wants to quickly extend the time range for the current search. With this PR we are implementing the relative range preset select next to the date time picker drop down button. This way a user just needs two clicks to extend or reduce the current time range.

image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #10769, opened 2021-06-02
Use correct time range format when selecting relative range preset in search bar.

linuspahl:
## Motivation and Context

Before this change we've updated the time range with e.g. `{ from: 300 }`.
This resulted in an error when selecting the `All Time` option.
We are now using the following format: `{ range: 300 }`. With this
format we are updating the time range in the same way like the `TimeRangeDropdown`.

Refs: https://github.com/Graylog2/graylog2-server/pull/10740

## Types of changes

## Checklist:

- x ] My code follows the code style of this project.


dennisoelkers (approved):
Looks great, fixes the issue! :tada:

Issue #10787, opened 2021-06-04
First dashboard page is not being displayed as first page on some dashboards.


linuspahl:
## Expected Behavior

The first page on a dashboard should always be displayed as the first page.

## Current Behavior

The first page on a dashboard is not always being displayed as the first page. Screenshot after opening a dashboard:

![image](https://user-images.githubusercontent.com/46300478/120771631-b4a06280-c51f-11eb-9c73-9e0c23623f3d.png)

## Steps to Reproduce (for bugs)

It is possible that it only happens with existing dashboards and not with a newly created one.

## Context

This can be related to: https://github.com/Graylog2/graylog2-server/pull/10739

## Your Environment

* Graylog Version: Current development version for 4.1

linuspahl:
Regarding exiting dashboard: the error occurred only for dashboards were I reproduced the bug we fix with #10739.
Regarding new dashboards: the error still occurs but during my tests only when creating a dashboard where the pages have no custom name.

PR #10790, opened 2021-06-04
Always define first search query as active query when loading view.

linuspahl:
## Motivation and Context

With this change we are defining the active query based on `view.search.queries` instead on `view.state`.
This way we can ensure that the active query (responsible for the active dashboard page) is always the first query.

Fixes https://github.com/Graylog2/graylog2-server/issues/10787

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/stores/ViewStore.ts
+++ b/graylog2-web-interface/src/views/stores/ViewStore.ts
@@ -157,7 +157,7 @@ export const ViewStore: ViewStoreType = singletonStore(
       /* Select selected query (activeQuery) or first query in view (for now).
          Selected query might become a property on the view later. */
       const queries = view.state.keySeq().toList();
-      const firstQueryId = queries.first();
+      const firstQueryId = view?.search?.queries?.first()?.id ?? queries.first();
       const selectedQuery = this.activeQuery && queries.find((id) => (id === this.activeQuery)) ? this.activeQuery : firstQueryId;
 
       this.selectQuery(selectedQuery);


graylog2-web-interface/src/views/stores/ViewStore.ts
```diff
@@ -157,7 +157,7 @@ export const ViewStore: ViewStoreType = singletonStore(
       /* Select selected query (activeQuery) or first query in view (for now).
          Selected query might become a property on the view later. */
       const queries = view.state.keySeq().toList();
-      const firstQueryId = queries.first();
+      const firstQueryId = view?.search?.queries?.first()?.id ?? queries.first();
```
linuspahl, line 160: I implemented `queries.first()` as a fallback to ensure we are definitely not breaking anything.

graylog2-web-interface/src/views/stores/ViewStore.ts
```diff
@@ -157,7 +157,7 @@ export const ViewStore: ViewStoreType = singletonStore(
...
       const selectedQuery = this.activeQuery && queries.find((id) => (id === this.activeQuery)) ? this.activeQuery : firstQueryId;
 
       this.selectQuery(selectedQuery);
```
linuspahl, line 163: Do you think we should also ensure that `views.state` has the same order like `view.search.queries` on the backend side? For example to avoid similar problems in the future.

dennisoelkers: Two thoughts:

  - In the future we should not rely on the order of the queries and store the tab order explicitly in the view.
  - For now, we should have only a single point where the order is derived from and that is the `queries` part of the `Search`.

PR #10796, opened 2021-06-07
Update Prometheus exporter default port

bernd:
The previous port has been grabbed by someone else in the meantime. The
new port has been added to the Prometheus default port allocations wiki
page now.


danotorrey (approved):
LGTM

PR #10797, opened 2021-06-07
Support PKCS#1 keys for Inputs

mpfz0r:
With PR #10400 a custom Keyfile loader was introduced,
to properly handle different key formats.
However, it was lacking support for old PKCS#1 key files.

Refs #9958


bernd (approved):
Fixes the issue for me. Thanks for adding a test! :+1:

patrickmann:
So many key formats :) Good catch.

patrickmann (approved):

PR #10798, opened 2021-06-07
Introducing `hasErrors` flag and prevent showing search result if set.

dennisoelkers:
## Motivation and Context

This PR is avoiding rendering the `SearchResult` component if the `hasErrors` flag is set. This flag is introduced in this PR and is set by the function which handles the response of the search parsing call.  This can be used e.g. to avoid showing the rendering result when there are undeclared parameters present.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/Search.tsx
+++ b/graylog2-web-interface/src/views/components/Search.tsx
@@ -15,7 +15,7 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
-import { useCallback, useEffect, useContext } from 'react';
+import { useCallback, useEffect, useContext, useState } from 'react';
 import * as Immutable from 'immutable';
 import styled, { css } from 'styled-components';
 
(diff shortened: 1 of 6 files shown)


linuspahl (commented):
Looks good to me, could not find any runtime errors. Two topics we should discuss:

1. When a user defines a parameter for a widget on a dashboard and the parameter is not declared we are directly displaying the "Undeclared query parameters" warning and hiding the aggregation builder. This change is fine in most cases, but it can be annoying when the parameter name contains a typo. In this case users have to define the parameter, before they can adjust the query string.
2. Currently it is still possible to create a widget, even when the search result is not visible. I can imagine that it can be confusing if you click on e.g. the "Message Count" button and nothing happens.

dennisoelkers:
@linuspahl: Good points. I addressed 1. by showing widgets et al if a search result is present, even if the query has errors. This way we avoid rendering an incomplete search result (showing a spinner) if a new search with a query containing an undeclared parameter in the URL, but also show the aggregation builder etc. if a search query has been executed before.

To address 2., I guess we should disable some functionality if `hasErrors` is `true`. That includes the `Create` section of the sidebar as well as the field actions in the fields list. We should do this as a followup PR though, imo.

linuspahl:
@dennisoelkers Sounds good, I created an issue for the 2. topic. https://github.com/Graylog2/graylog2-server/issues/10813

linuspahl (approved):
Looks good to me!

PR #10811, opened 2021-06-09
Unify relative range preset dropdown usage in search bar and date time range picker.

linuspahl:
## Motivation and Context

We recently implemented a relative range preset select for the search bar (https://github.com/Graylog2/graylog2-server/pull/10740.) In a follow up PR (https://github.com/Graylog2/graylog2-server/pull/10769) we tried to unify the search time range update of the relative range preset dropdown with the date time picker.

This PR is cleaning up both changes, we are now:
- also using the `RangePresetSelect` component for the date time picker. The component is already written in a way it can be reused, but we forgot to implement it for the date time picker.
- fixing the format of the time range update when selecting a range preset. We are now:
Using the following format when selecting the preset "all messages": `{ type: "relative", range: 0 }` and in every other case: `{ type: "relative", from: 300 }`. This way the time range update of the relative range preset dropdown is unified with the update triggered by the date time range picker. Before this change the relative preset dropdown always used the following format: `{ type: "relative", range: 300 }`.
I realized this change is necessary, after finding following bug: When selecting a relative range preset option like "30 minutes" twice, the search bar form was dirty and it was not possible to get rid of the dirty state by submitting the form.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/TimerangeForForm.ts
```diff
@@ -22,8 +22,21 @@ import type { TimeRange } from 'views/logic/queries/Query';
 import type { SearchBarFormValues } from 'views/Constants';
 import { isTypeRelativeWithStartOnly, isTypeRelativeWithEnd } from 'views/typeGuards/timeRange';
 
+import { TimeRangeDropDownFormValues } from './searchbar/date-time-picker/TimeRangeDropdown';
+
 const formatDatetime = (datetime) => datetime.toString(DateTime.Formats.TIMESTAMP);
 
+export const onSettingSearchBarFormTimeRange = (nextTimeRange: TimeRangeDropDownFormValues['nextTimeRange']) => {
```
dennisoelkers, line 29: The naming implies that this is just an event handler, but I think what it is actually doing is normalizing a time range by converting the "all messages" time range. If my impression is correct, I would suggest moving this to a separate module and rename it to something like `normalizeIfAllMessagesRange`.

graylog2-web-interface/src/views/components/TimerangeForForm.ts
```diff
@@ -22,8 +22,21 @@ import type { TimeRange } from 'views/logic/queries/Query';
...
+  if (isTypeRelativeWithEnd(nextTimeRange) && nextTimeRange.from === RELATIVE_ALL_TIME) {
```
dennisoelkers, line 30: I think we can also create a `isAllMessagesRange` predicate. This way we can also encapsulate the usage of the constant.

dennisoelkers (changes requested):
Looks good, just some code style suggestions.

graylog2-web-interface/src/views/components/TimerangeForForm.ts
linuspahl: Good idea, I created a helper function, but did not started to implement it everywhere, to keep this PR small.
I am not completely sure about the best place for this type of functions and created a separate file in `views/logic/queries`.

dennisoelkers (approved):
Thanks! Looks good and works great now! :tada:

As a future improvement, I would suggest bundling up the different functions used for timeranges (`normalizeIfAllTimesRange`, `isAllTimesRange`, ...) in a `timerange` module.

PR #10812, opened 2021-06-09
Update bouncycastle to version 1.69

thll:
This bumps our bouncycastle dependencies to the most recent version.

Prevents downgrading of transitive bouncycastle dependencies e.g. in https://github.com/Graylog2/graylog-plugin-enterprise/pull/2278.


mpfz0r (approved):
Did some testing. LGTM

Issue #7429, opened 2020-02-10
Unify data table widget styling for all browsers.


linuspahl:
The `DataTable` widget currently looks different in Chrome

![image](https://user-images.githubusercontent.com/46300478/74148263-57cfb900-4c05-11ea-93ef-4d382d8e2f3a.png)
![image](https://user-images.githubusercontent.com/46300478/74148322-7a61d200-4c05-11ea-999a-09f363fa6e26.png)

compared with Safari / Firefox:

![image](https://user-images.githubusercontent.com/46300478/74148292-6e761000-4c05-11ea-9f0a-862d64ce3b34.png)

In comparison with other widgets like the message table, we should also remove the padding / margin of the data table.

![image](https://user-images.githubusercontent.com/46300478/74148402-a2513580-4c05-11ea-83db-9c67187459c8.png)

## Your Environment
* Graylog Version: Graylog 3.2.1

PR #10814, opened 2021-06-09
Removing not needed spacings in data table and message table widget.

linuspahl:
## Motivation and Context

Before this change the message table and data table widget had some not need spacings.
Removing these spacings unifies the styling with other widgets and for different browser.

Before: 

image

After:

image

Fixes https://github.com/Graylog2/graylog2-server/issues/7429


kmerz (approved):
Code LGTM :+1: 

PR #10845, opened 2021-06-14
Do not hide legends for pie charts

### LINKED ISSUE #10837: Pie charts in GL 4.1 beta no longer display value keys
I might be missing the obvious, but it seems that the pie charts in `GL 4.1 beta` no longer display value keys (screenshots below).

Here is my widget configuration :

image

And here is the resulting chart after clicking on `Apply Changes` (notice how there are no value keys displayed under the chart) : 

image

## Environment :

* Graylog Version: 4.1 beta 1
* Java Version: 8u292
* MongoDB Version: 4.0.24
* Operating System: Ubuntu Server 20.04

kmerz:
## Motivation
Prior to this change, we wanted to hide the legend if only one label
was presented to the user. The evaluation of that state did not take
into account that the pie chat needs to always show the labels.

## Description
This change will add a neverHide flag to the PlotLegend component and
that flag will be used by the PieChart component.

This PR also adds field value actions for the pie chart legend.
image

Fixes #10837
Refs https://github.com/Graylog2/graylog2-server/issues/6612

## How Has This Been Tested?
- Create a Pie Chart with top values widget and see that the Legend is presented
- The histogram still does not show the legend

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/visualizations/PlotLegend.test.tsx
+++ b/graylog2-web-interface/src/views/components/visualizations/PlotLegend.test.tsx
@@ -49,7 +49,7 @@ const columnPivots = [Pivot.create('field1', 'unknown')];
 const config = AggregationWidgetConfig.builder().series([Series.forFunction('count')]).columnPivots(columnPivots).build();
 
 // eslint-disable-next-line react/require-default-props
-const SUT = ({ chartDataProp = chartData, plotConfig = config }: { chartDataProp?: Array<{ name: string, }>, plotConfig?: AggregationWidgetConfig }) => (
+const SUT = ({ chartDataProp = chartData, plotConfig = config, neverHide = false }: { chartDataProp?: Array<{ name: string, }>, plotConfig?: AggregationWidgetConfig, neverHide?: boolean }) => (
   <WidgetFocusContext.Provider value={{
     focusedWidget: undefined,
     setWidgetFocusing: jest.fn(),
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/views/components/visualizations/PlotLegend.test.tsx
```diff
@@ -108,10 +108,17 @@ describe('PlotLegend', () => {
...
     render(<SUT chartDataProp={[{ name: 'name1' }]} plotConfig={plotConfig} />);
 
     expect(screen.queryByText('name1')).not.toBeInTheDocument();
   });
+
+  it('should not hide with a single value if configured', async () => {
+    const plotConfig = AggregationWidgetConfig.builder().series([Series.forFunction('count')]).build();
+    render(<SUT chartDataProp={[{ name: 'name1' }]} plotConfig={plotConfig} neverHide />);
+
+    expect(screen.queryByText('name1')).toBeInTheDocument();
```
dennisoelkers, line 122: In order to make this work regardless when it is rendered, I would suggest doing this:
```suggestion
    await screen.findByText('name1');
```

dennisoelkers (changes requested):
It shows the legend again for a pie chart, but its entries are not clickable. In addition, one minor suggestion for the test.

dennisoelkers (approved):
Looks good, works as expected!

PR #10849, opened 2021-06-14
Do not render column roll-up label as value action in PlotLegend

### LINKED ISSUE #10840: Value actions should be limited/hidden for functions in charts
## Current Behavior

When creating a chart with at least one column grouping, the roll up functions are shown as traces in the chart and in the legend. The legend entry is clickable and shows value actions which do not make sense for the function and should be limited/hidden.

image

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kmerz:
## Motivation
Prior to this change, the Plot legend would render the column roll-up
label as Value action and with that the user would get a menu offered
which would not really make sense.

## Description
This change will try to find out if the label it renders is a series
function and will not render it with Value Actions.

## Fixes also
This fixes also a problem that when multiple series where selected by
the user the value actions did got presented even though they were
broken.

## Known Issue
If the series function will be renamed to a value which exists in the
chart as a value, the value is not clickable and the plot legend will
only display the label one time (as does plotly renders the it).

Fixes #10840

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/visualizations/PlotLegend.test.tsx
+++ b/graylog2-web-interface/src/views/components/visualizations/PlotLegend.test.tsx
@@ -78,7 +78,7 @@ describe('PlotLegend', () => {
     await screen.findAllByLabelText('Color Hint');
   });
 
-  it('should set a color when clicking on the color hing', async () => {
+  it('should set a color when clicking on the color hint', async () => {
     render(<SUT />);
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Looks good to me, could not find an error!

PR #10850, opened 2021-06-14
Heatmap: Clear min and max on auto scale use

kmerz:
## Motivation
Prior to this change, if the user switched back from using min and max
values to auto scale the auto scale setting was ignored and instead min
and max were still used.

That is because plolty does not know of a auto scale option but only the
absense of min and max.

## Description
This change will clear the values min and max when auto scale is used.

## How Has This Been Tested?
- Edit a heatmap to use min and max instaed of auto scale
- switched back to auto scale
- 
- profit

## Types of changes


dennisoelkers (approved):
Looks good, fixes the issue! :tada:

Issue #10858, opened 2021-06-15
It should be possible to remove all groupings of an aggregation.


linuspahl:
## Current Behavior

Currently it is not possible to remove all groupings of an aggregation.

![groupings](https://user-images.githubusercontent.com/46300478/122045343-e3e78700-cddd-11eb-864c-ebeab18556e6.gif)

## Steps to Reproduce (for bugs)

1. Open the search page
2. Edit the default bar chart
3. Try to remove all groupings

## Your Environment

* Graylog Version: 4Graylog 4.1.0-beta.1

PR #10859, opened 2021-06-15
Fixing deletion of all groupings of an aggregation.

linuspahl:
## Description

As described in https://github.com/Graylog2/graylog2-server/issues/10858 it is currently not possible to remove all groupings of an aggregation element. This error occurred because we removed the `goupby` key of the `WidgetConfigForm` state when removing all groupings. We are only considering aggregation element changes on submit when its key exists in the form state.

I thought about a solution which ensures we are not having a similar problem in the future, e.g. when adding an aggregation element. But I could not find a good one, since I prefer to only consider aggregation element changes when their key exists in the form sate. Currently we can just ensure this error is not occurring for our existing aggregation elements, by writing a unit test for each one.

Fixes https://github.com/Graylog2/graylog2-server/issues/10858

## Types of changes

## Checklist:


dennisoelkers (approved):
Looks good, fixes the issue! :tada:

PR #10860, opened 2021-06-15
Add error boundary to plugin authentication form

edmundoa:
This allows us to handle errors on the plugin code, show information about it and fallback to the regular login form.

Screenshot 2021-06-15 at 14 44 04

Fixes Graylog2/graylog-plugin-enterprise#2363

Changes should be backported into 4.1.

### DIFF
--- a/graylog2-web-interface/src/pages/LoginPage.jsx
+++ b/graylog2-web-interface/src/pages/LoginPage.jsx
@@ -15,9 +15,11 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


edmundoa:
In the end we didn't manage to put this in beta 2. The situation is now less likely to occur, though, since we now have validation in the Okta authentication service.

In order to reduce the risk of breaking the final release, and since this change is still valuable (it may catch an unknown on a plugin that would render the login page unusable), I suggest waiting on this until 4.1.0 is released. We can then merge it into master and backport it to 4.1 so it is included in 4.1.1.

edmundoa:
@ousmaneo I think this can now be reviewed, thank you!

ousmaneo (approved):
Test and works 👍

Issue #10830, opened 2021-06-11
System menu drop-down doesn't close when clicking some entries


bernd:
## Expected Behavior

Clicking any entry in the "System" menu should close the menu.

## Current Behavior

When clicking some entries in the "System" menu the menu doesn't close:

![Peek 2021-06-11 10-35](https://user-images.githubusercontent.com/461/121657978-3143bb80-caa1-11eb-91b7-929ec72c4e01.gif)

## Steps to Reproduce (for bugs)

1. Open "System" menu
2. Click "Inputs"
3. See that menu doesn't close

## Your Environment

* Graylog Version: 4.1.0-beta.1
* Operating System: Ubuntu 20.04 LTS
* Browser version: Google Chrome 91.0.4472.101

PR #10862, opened 2021-06-15
Always close system menu when selecting link in system menu dropdown.

linuspahl:
## Description

We recently wrapped some menu items in the system menu dropdown with the `HideOnCloud` component.
One side effect was that we no longer closed the dropdown when selecting any wrapped menu item.

The reason:
When defining a child for the `react-bootstrap` component `NavDropdown`, the child automatically receives props like `active`, `onKeyDown`, `onSelect`. By wrapping a menu item in the system menu drop down with `HideOnCloud`, the component `HideOnCloud` receives these props, but not its children.

This PR is fixing the bug by forwarding the props.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10830

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/util/conditional/HideOnCloud.test.tsx
+++ b/graylog2-web-interface/src/util/conditional/HideOnCloud.test.tsx
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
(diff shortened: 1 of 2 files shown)


ousmaneo (approved):
Legit 👍

PR #10865, opened 2021-06-16
Always update all widgets after unfocusing

kmerz:
## Motivation
Prior to this change, we did only update all searches after we left the
focus mode from editing. But when we switched a message table to a log
view and unfocused the log view other widgets were not part of the newly
generated search result.

## Description
This change will update all widgets after leaving the unfocusing mode.

## Types of changes

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/2359

### DIFF
--- a/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
+++ b/graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
@@ -84,10 +84,7 @@ const useSyncStateWithQueryParams = ({ focusedWidget, focusUriParams, setFocused
       setFocusedWidget(nextFocusedWidget);
       const filter = nextFocusedWidget?.id ? [nextFocusedWidget.id] : null;
       SearchActions.setWidgetsToSearch(filter);
-
-      if (focusedWidget?.editing && filter === null) {
-        SearchActions.executeWithCurrentState();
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -84,10 +84,7 @@ const useSyncStateWithQueryParams = ({ focusedWidget, focusUriParams, setFocused
       setFocusedWidget(nextFocusedWidget);
       const filter = nextFocusedWidget?.id ? [nextFocusedWidget.id] : null;
       SearchActions.setWidgetsToSearch(filter);
-
-      if (focusedWidget?.editing && filter === null) {
-        SearchActions.executeWithCurrentState();
-      }
+      SearchActions.executeWithCurrentState();
```
linuspahl: As discussed:

With this change we are currently executing the search twice when opening the search page.

We are now also executing the search when we focus / edit a widget, aas a result we loos the scroll position in the log view widget or the current page of the message table when focusing a widget.

One possible solution could be changing `focusedWidget?.editing` to `focusedWidget?.focusing`.

linuspahl (approved):
Looks good to me!

PR #10867, opened 2021-06-16
Generate new search when only `trend` setting is changed.

dennisoelkers:
## Motivation and Context

Prior to this PR, when enabling only the boolean `trend` setting of a numeric visualization and nothing else, we did not generate a new search and the `trend` search type was not retrieved. This lead to empty trending information. The root cause of this was that we generally leave out the `visualization` and `visualizationConfig` keys of an aggregation config when determining equality to decide if we need a new search or not.

This change is now implementing `equalsForSearch` methods in the `VisualizationConfig` classes (defaulting to return `true`, as visualization configs should generally not need to generate a new search) and in `NumberVisualizationConfig` (returning a boolean depending on the presence and value of the `trend` setting of both configs). Also, the `visualizationConfig` key is added to the checked attributes. This leads to generating a new search when the `trend` setting is flipped.

The issue can be easily reproduced:

  - start a new search/dashboard
  - create a `Message Count` predefined aggregation
  - edit it
  - toggle only(!) the `trend` setting and apply it
  - see that trending information is empty (`-- / --`)

Before:
before-trend-change

After:
after-trend-change

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/stores/__tests__/EqualityCheck.fixtures.ts
+++ b/graylog2-web-interface/src/stores/__tests__/EqualityCheck.fixtures.ts
@@ -16,6 +16,12 @@
  */
 import { List, Map } from 'immutable';
 
+import AggregationWidgetConfig from 'views/logic/aggregationbuilder/AggregationWidgetConfig';
+import NumberVisualization from 'views/components/visualizations/number/NumberVisualization';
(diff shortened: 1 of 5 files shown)


linuspahl (approved):
Looks good to me, fixes the error and I could not find a problem!

Issue #10885, opened 2021-06-17
Absolute time range of dashboard widgets has wrong format, when applying time range changes by clicking on "Apply changes".


linuspahl:
## Expected Behavior

We should always use the same format for absolute time ranges when updating the time range of a widget.

## Current Behavior

When selecting an absolute time range for a dashboard widget, a user can apply this change by clicking on the search bar button (in the widget edit mode) or by clicking on the "Apply Changes" button:

![image](https://user-images.githubusercontent.com/46300478/122424368-f3f29880-cf8e-11eb-812f-f20ad557999a.png)

Currently the format of the time range differs, depending on the clicked button.
1. When clicking on the search button (working):
```
"timerange": {
    "type": "absolute",
    "from": "2021-06-17T14:38:44.176Z",
    "to": "2021-06-17T14:43:44.176Z"
},
```
2. when clicking on "Apply changes" (not working):
```
"timerange": {
    "type": "absolute",
    "from": "2021-06-17 14:38:02.860",
    "to": "2021-06-17 14:43:02.860"
},
```

This results in an error, because the widget will be saved with an invalid time range.

## Steps to Reproduce (for bugs)

1. edit a dashboard widget
2. select an absolute time range (but do not save the widget or submit the search yet)
3. click on the "Apply Changes" button

## Your Environment

* Graylog Version: Graylog 4.1.0-beta.2

PR #10888, opened 2021-06-18
Always normalize dashboard widget time range correctly when updating widget.

linuspahl:
## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/10885 there is currently a bug, when updating the time range of a dashboard widget by using the "Apply Changes" button. This bug does not occur when a user applies the time range changes, by executing the search first.

The bug occurred because we were not normalizing the widget time range correctly. This PR by always normalizing the dashboard widget time range, no matter if the changes are being applied by clicking on the search or "Apply Changes" button.

Fixes: https://github.com/Graylog2/graylog2-server/issues/10885

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/widgets/Widget.aggregations.test.tsx
+++ b/graylog2-web-interface/src/views/components/widgets/Widget.aggregations.test.tsx
@@ -47,9 +47,18 @@ import FieldTypesContext from '../contexts/FieldTypesContext';
 import ViewTypeContext from '../contexts/ViewTypeContext';
 
 const testTimeout = applyTimeoutMultiplier(30000);
+const mockedUnixTime = 1577836800000; // 2020-01-01 00:00:00.00
 
 jest.mock('./WidgetHeader', () => 'widget-header');
 jest.mock('./WidgetColorContext', () => ({ children }) => children);
+
+jest.mock('moment-timezone', () => {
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/components/widgets/Widget.aggregations.test.tsx
```diff
@@ -47,10 +47,19 @@ import FieldTypesContext from '../contexts/FieldTypesContext';
...
 jest.mock('./WidgetHeader', () => 'widget-header');
 jest.mock('./WidgetColorContext', () => ({ children }) => children);
 
+jest.mock('moment-timezone', () => {
+  const momentMock = jest.requireActual('moment-timezone');
+  momentMock.tz.setDefault('UTC');
+  momentMock.tz.guess = () => 'UTC';
+
+  return momentMock;
+});
```
linuspahl, line 61: We need to define the timezone for moment here, because we are using `moment.tz.guess` in `DateTime.getBrowserTimezone`. Unfortunately we can not use `AppConfig.rootTimeZone()` in this case.

We should include this mock in our jest setup in a separate PR. In the same step we can implement a constant for   the root time zone and use it to mock `moment.tz` and `AppConfig.rootTimeZone()`

dennisoelkers (approved):
Looks good and fixes the issue for me! :tada:

PR #10889, opened 2021-06-18
Adjusting to changed error messages in ES7 for flood stage/missing alias target

### LINKED ISSUE #10887: Message ingestion is not retried when ES is in flood stage for ES7
## Expected Behavior

When messages are supposed to be ingested into ES while it is in flood stage, those message ingestion attempts will be retried when they fail.

## Current Behavior

Messages which are ingested into ES7 while it is in flood stage will be lost.

Initial investigation points out that the error message returned by ES has changed and therefore this error will not be identified as retriable.

1.
2.
3.
4.

## Your Environment

* Graylog Version: 4.1.0-beta.2
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

Between ES 7.8.0 and 7.10.1, the error messages returned by the server for these case:

  - a message cannot be ingested to an index because the index/cluster is in flood stage
  - a message cannot be ingested to an alias, because the alias is missing a target

have changed. Therefore our error checking code does not identify these cases as being retriable, resulting in message loss.

This PR addresses these cases and adds the changed error messages in addition to the previous ones in order to support both versions.

This circumstance could have been detected if our integration tests would have used newer versions of ES for testing. Therefore, the ES version for the tests was bumped. In the future, we should complete our plans to run integration tests against a matrix of versions, including earliest and latest version.

Fixes #10887.

Note: There are additional cornes cases during flood stage when an index is rotated/the deflector is cycled (either manually or automatically). While this works, it currently generates quite some noise in the logs looking ugly, because a lot of exceptions are thrown in different places. The indices/deflectors are cleaned up successfully after the ES cluster was recovered though, therefore this was not cleaned up at this stage, because of the risk of accidental regressions. It will be cleaned up after the release to maintain a readable log during these situations.

## How Has This Been Tested?

Manual testing using this procedure:

  - Triggering flood stage by decreasing disk watermark thresholds for ES
  - Ingesting one or more messages
  - Resetting disk watermask tresholds
  - Waiting for ES to reset index blocks automatically
  - Check if messages have been ingested and ingestion continues

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/MessagesAdapterES7.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/MessagesAdapterES7.java
@@ -64,6 +64,9 @@ public class MessagesAdapterES7 implements MessagesAdapter {
     static final String UNAVAILABLE_SHARDS_EXCEPTION = "unavailable_shards_exception";
     static final String PRIMARY_SHARD_NOT_ACTIVE_REASON = "primary shard is not active";
 
+    static final String ILLEGAL_ARGUMENT_EXCEPTION = "illegal_argument_exception";
+    static final String NO_WRITE_INDEX_DEFINED_FOR_ALIAS = "no write index is defined for alias";
+
     private final ElasticsearchClient client;
     private final Meter invalidTimestampMeter;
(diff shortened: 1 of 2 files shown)


bernd (approved):
LGTM :+1: Thank you!

janheise (approved):
LGTM

PR #10890, opened 2021-06-18
Releasing objects retained by RawMessageEvents

diesaulov:
Even though the input ring buffer is capped, the unreleased references occupy memory which could be used otherwise. The fix covers both scenarios: with and without journalling.


bernd:
@diesaulov Thank you, good catch! :+1: This needs some careful testing and review to make sure it doesn't break anything.

This is not supposed to go in before 4.1 is release. We might do a stable backport, though.

thll:
@diesaulov good catch indeed! Looks like nulling the fields was in there before but it got lost during implementation of https://github.com/Graylog2/graylog-plugin-cloud/issues/211.

mpfz0r:
Nice catch!
I wouldn't talk about `leaking` in the PR description.
Maybe something like `explicitly release RawMessageEvents`?

mpfz0r (commented):
Once this change is in, we need to update the submodules in the forwarder again.
Also, this should be ported to cloud-4.0

thll (approved):

mpfz0r (approved):

PR #10899, opened 2021-06-21
Mock moment time zone for all tests.

linuspahl:
## Motivation and Context

Before this change it was possible that `moment.tz.guess()` returns a different time zone, when running the frontend tests, depending the environment (e.g. local dev environment vs. GitHub actions).

To avoid this problem in the future we are now mocking the moment time zone for all tests.


kmerz (approved):
LGTM :+1:

PR #10902, opened 2021-06-21
Configure more dynamic default menu placement for `Select` component.

linuspahl:
**Please note:  This PR needs to be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/2399**

## Motivation and Context

With this PR we are changing the default for the `menuPlacement` prop of the common `Select` component from `bottom` to `auto`. When a select menu can not be displayed at the bottom of a select, because of the browser window height, it will be display at the top instead.

Before (opening menu creates a scroll bar):
image

After: 
image

## How Has This Been Tested?

Tested with different use cases.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2399

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.tsx
+++ b/graylog2-web-interface/src/components/common/Select.tsx
@@ -358,7 +358,7 @@ class Select extends React.Component<Props, State> {
     value: undefined,
     valueKey: 'value',
     valueRenderer: undefined,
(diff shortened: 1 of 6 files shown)


kmerz (approved):
Code LGTM :+1: 

PR #10905, opened 2021-06-21
Replace hyphens with underscored in metric names

### LINKED ISSUE #10904: Invalid Prometheus metric names
## Overview

The default core mappings file for the new Prometheus exporter [prometheus-exporter.yml](https://github.com/Graylog2/graylog2-server/blob/046210b213befd3be47a8d7d89fdc7b2891ce06c/graylog2-server/src/main/resources/prometheus-exporter.yml) contains some metric names with hyphens (-) in the name. Hyphens are not allowed, and underscores should be used instead. 

According to the [Prometheus documentation](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels):
> may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.

## Suggested fix for 4.1
Replace the hyphens with underscores.

danotorrey:
Fixes #10904 

Replaces hyphens with underscores in the default core [prometheus-exporter.yml](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/resources/prometheus-exporter.yml) mappings file.

### DIFF
--- a/graylog2-server/src/main/resources/prometheus-exporter.yml
+++ b/graylog2-server/src/main/resources/prometheus-exporter.yml
@@ -904,22 +904,22 @@ metric_mappings:
   - metric_name: "output_executor_service_submitted"
     match_pattern: "org.graylog2.initializers.OutputSetupService.executor-service.submitted"
 
-  - metric_name: "gelf_chunk_aggregator_complete-messages"
+  - metric_name: "gelf_chunk_aggregator_complete_messages"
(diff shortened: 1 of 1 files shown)


danotorrey:
FYI @juju2112 This is needed for the next RC.

bernd (approved):
LGTM :+1: Thank you!

PR #10915, opened 2021-06-22
Improve error logging for auth service connection test

patrickmann:
Provide a user-friendly error message when testing the connection to an auth service backend.
Avoid logging an exception, when the service is merely misconfigured.

This is related to PR [enterprise/2403](https://github.com/Graylog2/graylog-plugin-enterprise/pull/2403)


thll (approved):
This looks good and we should add this change to provide a safety-net. But I'm wondering if we can still avoid hitting the problem it in the first place.

If the backend is wrongly configured, then the UI should have triggered a validation of the backend before issuing the `testConnection` request. Also, the backend should catch all exceptions that could happen within `testConnection` on its own and transform them into errors within the `AuthServiceBackendTestResult`.

Feel free to merge this **after** the `4.1` release.

PR #10919, opened 2021-06-22
Always display widget title in line with drag handle and widget actions.

linuspahl:
## Motivation and Context

Before this change it was that the widget header elements are displayed in three rows, depending on the widget title length.
With this change we ensure that all elements are displayed "inline".

Before:
image

After: 
image


kmerz (approved):
:heart: 
LGTM :+1: 

PR #10923, opened 2021-06-22
Implement `User` class for value of `CurrentUserProvider`.

linuspahl:
**Please note: This PR needs to be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/2409**

## Motivation and Context

Before this change, the current user provided by the `CurrentUserProvider` was a simple object, with this change we are using the `User` class for the related context. This has many benefits, for example we can now use the `Builder` and we can use camel case instead of snake case attributes.

In another iteration we can implement the `User` class for the state of the `CurrentUserStore`.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2409


graylog2-web-interface/src/components/common/IfPermitted.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+      expectToRenderChildren();
+    });
+  });
+
+  it('passes props to children', () => {
+    const Foo = jest.fn(() => <p>Something else!</p>) as React.ElementType;
+    const Bar = jest.fn(() => <p>Something else!</p>) as React.ElementType;
+
+    render((
+      // @ts-ignore
```
linuspahl, line 235: What could be a good way to allow any attribute for the `IfPermitted` component? Currently we are getting an error because of `something` and `otherProps`.

I tried to type the custom props using a generic type, but it resulted in a solution where TypeScript thinks the child component also have the `permissions` prop.

kmerz (approved):
Code LGTM :+1: 

PR #10927, opened 2021-06-23
Display widget edit action as action icon instead of menu item in widget actions dropdown.

linuspahl:
## Motivation and Context

With this PR we are changing the placement of the widget edit action. The widget edit action is now a widget action icon instead of a menu item in the widget action menu. This way the user needs less clicks to edit a widget.

We are also changing the position of the widget stretch icon, mainly to keep the placement of most other icons consistent when focusing a widget (and hiding the stretch icon).

Before:
image

After:
image


linuspahl:
On a dashboard we also display the replay search action as an icon. What do you think about displaying this action as a menu item in the widget action menu instead? This way the widget action would be slightly cleaner.

image

kmerz (approved):
:heart: 
Code LGTM :+1: 

Issue #9904, opened 2021-01-14
Remove remaining legacy views permissions (frontend).


linuspahl:
## Description
In the past we migrated the legacy views permissions defined in https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/LegacyViewsPermissions.java but some of them still exists in the frontend code. (`views/Permissions` and `event-definition-types/FilterForm`)

We need to check if their usage results in an undesired behaviour and remove them.

* Graylog Version: 4.0

PR #10933, opened 2021-06-24
Display info for undeclared parameters in event definitions filter step only if user has `lookuptables:read` permission.

linuspahl:
## Description

With this PR we are adjusting the permissions check for the undeclared parameters info in the event definitions filter step.
The undeclared parameters info appears, when the filter search query contains an undeclared parameter:

image

Before this change we only displayed this info for users with the following permissions:
```
[
  'streams:read',
  'extendedsearch:create',
  'extendedsearch:use',
];
```

The `extendedsearch` permissions are legacy view permissions which are no longer supported. Related issue: https://github.com/Graylog2/graylog2-server/issues/9904. This did not caused a bug, because only admin users were (and are) able to define query parameters.

Because these query parameters are only usable, when a user has access to lookup tables, we are now:
- only displaying the undeclared parameter info when a user has the following permissions: `lookuptables:read`
- are displaying an info that only admin users are able to declare parameters, when a user does not have this permission. Admins are currently the only users with this permission.

### DIFF
--- a/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
+++ b/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
@@ -66,7 +66,7 @@ class FilterForm extends React.Component {
   );
 
   _parseQuery = lodash.debounce((queryString) => {
-    if (!this._userCanViewLookupTables) {
+    if (!this._userCanViewLookupTables()) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


mpfz0r (changes requested):
With this change it is impossible for non-admin users to create event definitions.
Even without using query parameters.
I think this is a move in the wrong direction.

graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterForm.jsx
```diff
@@ -231,6 +226,14 @@ class FilterForm extends React.Component {
       return onChange('config', newConfig);
     };
 
+    if (!this._userCanViewLookupTables()) {
+      return (
+        <Alert bsStyle="info">
+          Only Admins are able to declare Query Parameters from Lookup Tables.
```
mpfz0r, line 232:
```suggestion
          This account lacks permission to declare Query Parameters from Lookup Tables.
```
I think we should phrase this differently. If, for example we decide to give the alerts manager role the lookuptable:read permission.

mpfz0r (approved):
LGTM :+1: Feel free to merge when green

PR #10935, opened 2021-06-24
Removing legacy usage view permissions `extendedsearch:use` and `view:use` in the frontend.

linuspahl:
## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/9904 we migrated the legacy views permissions defined in https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/LegacyViewsPermissions.java but some of them still exist in the frontend code.

With this PR we are removing the legacy views permissions. There are still some legacy permissions in `event-definition-types/FilterForm` which will be removed with https://github.com/Graylog2/graylog2-server/pull/10933

Fixes: https://github.com/Graylog2/graylog2-server/issues/9904

### DIFF
--- a/graylog2-web-interface/src/views/bindings.tsx
+++ b/graylog2-web-interface/src/views/bindings.tsx
@@ -21,7 +21,6 @@ import { PluginExports } from 'graylog-web-plugin/plugin';
 import Routes from 'routing/Routes';
 import App from 'routing/App';
 import AppConfig from 'util/AppConfig';
-import * as Permissions from 'views/Permissions';
 import { MessageListHandler } from 'views/logic/searchtypes/messages';
 import { MessageList } from 'views/components/widgets';
 import AddToTableActionHandler from 'views/logic/fieldactions/AddToTableActionHandler';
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/bindings.tsx
```diff
@@ -128,9 +127,9 @@ const exports: PluginExports = {
     { path: showSearchPath, component: ShowViewPage, parentComponent: App },
     { path: `${Routes.unqualified.stream_search(':streamId')}/new`, component: NewSearchRedirectPage, parentComponent: null },
     { path: Routes.unqualified.stream_search(':streamId'), component: StreamSearchPage, parentComponent: App },
-    { path: extendedSearchPath, component: NewSearchPage, permissions: Permissions.ExtendedSearch.Use, parentComponent: App },
+    { path: extendedSearchPath, component: NewSearchPage, parentComponent: App },
 
-    { path: viewsPath, component: ViewManagementPage, permissions: Permissions.View.Use },
+    { path: viewsPath, component: ViewManagementPage },
```
linuspahl, line 132: The `permissions` attribute of a plugin route is currently not taken into account.

dennisoelkers (approved):

PR #10937, opened 2021-06-25
Remove unused byte code generator for the pipeline language

bernd:
This has never been properly tested and was never enabled by default.
Since the code makes it harder to understand the pipeline
infrastructure, we remove it now to help people who have to dive into
the code for debugging and maintenance.


kroepke (approved):
lgtm

PR #10939, opened 2021-06-25
Set default value for matchingType in StreamDTO

### LINKED ISSUE #10514: After upgrade to Graylog 4.x, the Streams page does not load (StreamDTO - Null matchingType)
## Expected Behavior

The Streams page should load successfully.

## Current Behavior

  After upgrading to Graylog 4.x, the Stream page does not load. The following stacktrace is displayed in the Graylog server log:

```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.graylog2.streams.StreamDTO`, problem: Null matchingType
at org.graylog2.streams.StreamDTO.create(StreamDTO.java:141) ~[graylog.jar:?]
at org.graylog2.streams.$AutoValue_StreamDTO.<init>($AutoValue_StreamDTO.java:71) ~[graylog.jar:?]
at org.graylog2.streams.AutoValue_StreamDTO.<init>(AutoValue_StreamDTO.java:19) ~[graylog.jar:?]
at org.graylog2.database.PaginatedDbService.findPaginatedWithQueryFilterAndSort(PaginatedDbService.java:159) ~[graylog.jar:?]
at org.graylog2.streams.PaginatedStreamService.findPaginated(PaginatedStreamService.java:48) ~[graylog.jar:?]
```

This indicates that the `streams.matching_type` collection value is null, but the `StreamDTO` requires a value.

## Possible Solution

Allow the [StreamDTO.matchingType()](https://github.com/Graylog2/graylog2-server/blob/29aa0ab4bd8dca7ef60ab7b5f1b17b805d9e7027/graylog2-server/src/main/java/org/graylog2/streams/StreamDTO.java#L66) to be Optional or nullable. Also, provide a default value in the create or builder methods. The default can use the same logic as [StreamImpl](https://github.com/Graylog2/graylog2-server/blob/29aa0ab4bd8dca7ef60ab7b5f1b17b805d9e7027/graylog2-server/src/main/java/org/graylog2/streams/StreamImpl.java#L223) uses currently.

This issue occurs mostly for older Graylog installations that are upgraded to 4.1 or beyond.

## Workaround
To workaround the issue, provide a `streams.matching_type` value equal to `AND` in MongoDB for any that currently have a null value.

See user suggestions here: 
https://github.com/Graylog2/graylog2-server/issues/9637#issuecomment-827091749
https://github.com/Graylog2/graylog2-server/issues/9637#issuecomment-769005192

## Your Environment

* Graylog Version: 4.x

thll:
Streams have a `matching_type` field, which got introduced in version `1.2.0` with https://github.com/Graylog2/graylog2-server/commit/a7bac4a1f05c823fd4c8487eeb0116319d35cfae. 

Streams which have been created before that wouldn't have that field set in MongoDB. This is gracefully handled by adding a default value, when the field is missing.

When pagination was added to the streams page in version `4.0.0` it wasn't considered that very old records could have the field missing. This would break the "streams" page.

This PR aims at fixing that by setting a default value for `matching_type` when de-serializing a stream from MongoDB for pagination.

Fixes #10514

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamDTO.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamDTO.java
@@ -19,23 +19,22 @@ package org.graylog2.streams;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
 import org.bson.types.ObjectId;
 import org.graylog.autovalue.WithBeanGetter;
+import org.graylog2.plugin.streams.Stream;
 import org.graylog2.plugin.streams.StreamRule;
 import org.graylog2.rest.models.alarmcallbacks.requests.AlertReceivers;
 import org.graylog2.rest.models.streams.alerts.AlertConditionSummary;
-import org.graylog2.rest.models.system.outputs.responses.OutputSummary;
 
 import javax.annotation.Nullable;
 import java.util.Collection;
-import java.util.Collections;
-
-import static com.google.common.base.MoreObjects.firstNonNull;
 
 @AutoValue
 @WithBeanGetter
 @JsonAutoDetect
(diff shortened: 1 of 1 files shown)


kmerz (approved):
Code LGTM :+1: 

PR #10954, opened 2021-06-30
Only restrict TLS to secure ciphers with TLS1.2 or greater

### LINKED ISSUE #10944: Allowed TLS Cipher Suites too tight
## Expected Behavior

Configurable TLS Cipher Suites or a 'common' list of allowed cipher suites

## Current Behavior

Only some TLS Ciphers are active since: https://github.com/Graylog2/graylog2-server/pull/10653
```

* TLS 1.0 Cipher suites:
     Attempted to connect using 80 cipher suites; the server rejected all cipher suites.

 * SSL 2.0 Cipher suites:
     Attempted to connect using 7 cipher suites; the server rejected all cipher suites.

 * TLS 1.2 Cipher suites:
     Attempted to connect using 158 cipher suites.

     The server accepted the following 6 cipher suites:
        TLS_RSA_WITH_AES_256_GCM_SHA384                   256                      
        TLS_RSA_WITH_AES_128_GCM_SHA256                   128                      
        TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384             256       ECDH: prime256v1 (256 bits)
        TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256             128       ECDH: prime256v1 (256 bits)
        TLS_DHE_RSA_WITH_AES_256_GCM_SHA384               256       DH (2048 bits) 
        TLS_DHE_RSA_WITH_AES_128_GCM_SHA256               128       DH (2048 bits) 

     The group of cipher suites supported by the server has the following properties:
       Forward Secrecy                    OK - Supported
       Legacy RC4 Algorithm               OK - Not Supported

     The server has no preferred cipher suite.


 * TLS 1.1 Cipher suites:
     Attempted to connect using 80 cipher suites; the server rejected all cipher suites.

 * OpenSSL CCS Injection:
                                          OK - Not vulnerable to OpenSSL CCS injection

 * TLS 1.3 Cipher suites:
     Attempted to connect using 5 cipher suites.

     The server accepted the following 2 cipher suites:
        TLS_AES_256_GCM_SHA384                            256       ECDH: x25519 (253 bits)
        TLS_AES_128_GCM_SHA256                            128       ECDH: x25519 (253 bits)

     The group of cipher suites supported by the server has the following properties:
       Forward Secrecy                    OK - Supported
       Legacy RC4 Algorithm               OK - Not Supported

     The server is configured to prefer the following cipher suite:
        TLS_AES_256_GCM_SHA384                            256       ECDH: x25519 (253 bits)
```

1.
2.
3.
4.

## Context

Communication to LDAP or ES is interrupted because of strict TLS Ciphers. 
## Your Environment

* Graylog Version: 4.1.0
* Java Version: JDK 11 Adopt Open JDK
* Elasticsearch Version: 
* MongoDB Version:
* Operating System:
* Browser version:

mpfz0r:
The changes made in #10653 also apply to cases where Graylog
uses TLS as a client.

This makes it impossible to create TLS connections
to legacy software that is only supporting old and insecure ciphers.
Most prominently ES 6.8.x

This changes gives the user the possiblity to re-enable old cipher
suites by changing our default and allowing TLS protocols < 1.2.

This can be done with the `enabled_tls_protocols` setting.
E.g. by setting `enabled_tls_protocols=TLSv1.1, TLSv1.2`

This change had to introduce a separate configuration bean,
to avoid triggering anything that might initialize the default SSLContext.
As for example would the MongoDBConfiguration.validat() method do.

Fixes #10944

/jenkins-pr-deps  Graylog2/graylog-plugin-enterprise#2427,Graylog2/graylog-plugin-enterprise-integrations#584

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -158,6 +158,10 @@ public class Configuration extends BaseConfiguration {
     @Parameter(value = "deactivated_builtin_authentication_providers", converter = StringSetConverter.class)
     private Set<String> deactivatedBuiltinAuthenticationProviders = Collections.emptySet();
 
+    // This is needed for backwards compatibility. The setting in TLSProtocolsConfiguration should be used instead.
+    @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
+    public Set<String> enabledTlsProtocols = null;
+
     @Parameter(value = "is_cloud")
     private boolean isCloud = false;
 
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -29,7 +29,6 @@ import com.github.joschi.jadconfig.validators.PositiveLongValidator;
 import com.github.joschi.jadconfig.validators.StringNotBlankValidator;
 import org.graylog2.plugin.BaseConfiguration;
 import org.graylog2.security.realm.RootAccountRealm;
-import org.graylog2.shared.security.tls.DefaultTLSProtocolProvider;
 import org.graylog2.utilities.IPSubnetConverter;
 import org.graylog2.utilities.IpSubnet;
 import org.joda.time.DateTimeZone;
@@ -158,9 +157,6 @@ public class Configuration extends BaseConfiguration {
 
     @Parameter(value = "deactivated_builtin_authentication_providers", converter = StringSetConverter.class)
     private Set<String> deactivatedBuiltinAuthenticationProviders = Collections.emptySet();
-
-    @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
-    private Set<String> enabledTlsProtocols = DefaultTLSProtocolProvider.getDefaultSupportedTlsProtocols();
 
     @Parameter(value = "is_cloud")
     private boolean isCloud = false;
@@ -322,10 +318,6 @@ public class Configuration extends BaseConfiguration {
 
     public Set<String> getDeactivatedBuiltinAuthenticationProviders() {
         return deactivatedBuiltinAuthenticationProviders;
-    }
-
-    public Set<String> getEnabledTlsProtocols() {
-        return enabledTlsProtocols;
     }
 
     @ValidatorMethod
--- a/graylog2-server/src/main/java/org/graylog2/Configuration.java
+++ b/graylog2-server/src/main/java/org/graylog2/Configuration.java
@@ -160,7 +160,7 @@ public class Configuration extends BaseConfiguration {
 
     // This is needed for backwards compatibility. The setting in TLSProtocolsConfiguration should be used instead.
     @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
-    public Set<String> enabledTlsProtocols = null;
+    private Set<String> enabledTlsProtocols = null;
 
     @Parameter(value = "is_cloud")
     private boolean isCloud = false;
@@ -322,6 +322,14 @@ public class Configuration extends BaseConfiguration {
 
     public Set<String> getDeactivatedBuiltinAuthenticationProviders() {
         return deactivatedBuiltinAuthenticationProviders;
+    }
+
+    /**
+     * This is needed for backwards compatibility. The setting in TLSProtocolsConfiguration should be used instead.
+     */
+    @Deprecated
+    public Set<String> getEnabledTlsProtocols() {
+        return enabledTlsProtocols;
     }
 
     @ValidatorMethod
--- a/graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
@@ -62,16 +62,12 @@ public class TLSProtocolsConfiguration {
      * If it's not configured (null and the default) return a secure set of supported TLS protocols.
      */
     public Set<String> getEnabledTlsProtocols() {
-        return getEnabledTlsProtocols(enabledTlsProtocols);
-    }
-
-    public static Set<String> getEnabledTlsProtocols(Set<String> configuredTlsProtocols) {
-        if (configuredTlsProtocols != null) {
-            if (configuredTlsProtocols.isEmpty()) {
-                return DefaultTLSProtocolProvider.getSupportedTlsProtocols();
+        if (enabledTlsProtocols != null) {
+            if (enabledTlsProtocols.isEmpty()) {
+                return DefaultTLSProtocolProvider.getAllSupportedTlsProtocols();
             }
-            return configuredTlsProtocols;
+            return enabledTlsProtocols;
         }
-        return DefaultTLSProtocolProvider.getDefaultSupportedTlsProtocols();
+        return DefaultTLSProtocolProvider.getSecureTLSProtocols();
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
@@ -1,0 +1,41 @@
+package org.graylog2.configuration;
+
+import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.converters.StringSetConverter;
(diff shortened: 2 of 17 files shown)


graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/AbstractTcpTransport.java
```diff
@@ -142,12 +143,12 @@ public AbstractTcpTransport(
             EventLoopGroup parentEventLoopGroup,
             EventLoopGroupFactory eventLoopGroupFactory,
             NettyTransportConfiguration nettyTransportConfiguration,
-            org.graylog2.Configuration graylogConfiguration) {
+            TLSProtocolsConfiguration tlsConfiguration) {
```
bernd, line 146: Hmm, this is a breaking API change. :thinking:  That means all TCP transports need to be adjusted to work. Also 3rd party ones. I guess we have to add the setting to `org.graylog2.Configuration` as well to avoid this. It's a bit ugly but I don't see any other way, unfortunately. :confused:

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public Set<String> getConfiguredTlsProtocols() {
+        return enabledTlsProtocols;
+    }
+
+    /**
+     * Retrieve the enabled TLS protocols setting.
+     * @return
+     * If the setting is explicitly configured (not null) return that.
+     * If it's configured to an empty set, return all supported protocols by the JVM.
```
bernd, line 45: How can this be configured to an empty set in the config file?

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ * Configuration bean for enabled TLS protocols.
+ *
+ * This was extracted to a separate configuration bean
+ * so that it can be parsed in the early server startup phase individually.
+ * Parsing the entire server configuration might trigger the initialization of the default SSLContext,
+ * which needs to happen after a `jdk.tls.disabledAlgorithms` setting is applied.
+ */
+public class TLSProtocolsConfiguration {
+    @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
+    private Set<String> enabledTlsProtocols = null;
```
bernd, line 35: I think we need to document the new default and the new behavior in the `misc/graylog.conf` file.

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public Set<String> getConfiguredTlsProtocols() {
```
bernd, line 37: Can you please add a `@Nullable` annotation. Thanks!

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
mpfz0r: `enabled_tls_protocols=`

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
mpfz0r: it hasn't really changed, it is just "lazy evaluated" by calling `getEnabledTlsProtocols()`

mpfz0r: but we should mention that you have to enable TLSv1.1 or older to activate insecure CBC ciphers

graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -159,8 +158,9 @@
     @Parameter(value = "deactivated_builtin_authentication_providers", converter = StringSetConverter.class)
     private Set<String> deactivatedBuiltinAuthenticationProviders = Collections.emptySet();
 
+    // This is needed for backwards compatibility. The setting in TLSProtocolsConfiguration should be used instead.
     @Parameter(value = "enabled_tls_protocols", converter = StringSetConverter.class)
-    private Set<String> enabledTlsProtocols = DefaultTLSProtocolProvider.getDefaultSupportedTlsProtocols();
+    public Set<String> enabledTlsProtocols = null;
```
bernd, line 163: Can we make this private again and add a getter with a `@Deprecated` annotation? That makes it clear in the IDE that the usage is deprecated. I think adding the annotation to the public field doesn't highlight it in the IDE.

graylog2-server/src/main/java/org/graylog2/shared/security/tls/DefaultTLSProtocolProvider.java
```diff
@@ -38,16 +38,20 @@
...
         final Set<String> tlsProtocols = Sets.newHashSet(DEFAULT_TLS_PROTOCOLS);
+        final Set<String> supportedProtocols = getSupportedTlsProtocols();
+        if (tlsProtocols.retainAll(supportedProtocols)) {
+            LOG.warn("JRE doesn't support all default TLS protocols. Changing <{}> to <{}>", DEFAULT_TLS_PROTOCOLS, tlsProtocols);
+        }
+        defaultSupportedTlsProtocols = tlsProtocols;
+        return defaultSupportedTlsProtocols;
+    }
+
+    public synchronized static Set<String> getSupportedTlsProtocols() {
```
bernd, line 49: What do you think about renaming the methods? I find it really hard to remember what the difference between `getSupportedTlsProtocols` and `getDefaultSupportedTlsProtocols` is. :smile: 

How about `getAllSupportedTlsProtocols` and `getSecureSupportedTlsProtocols`? (or something similar)

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
bernd: Ah right, thanks. Should have read the comment in `misc/graylog.conf`... :smile:

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return
+     * If the setting is explicitly configured (not null) return that.
+     * If it's configured to an empty set, return all supported protocols by the JVM.
+     * If it's not configured (null and the default) return a secure set of supported TLS protocols.
+     */
+    public Set<String> getEnabledTlsProtocols() {
+        return getEnabledTlsProtocols(enabledTlsProtocols);
+    }
+
+    public static Set<String> getEnabledTlsProtocols(Set<String> configuredTlsProtocols) {
```
bernd, line 68: Can this be private? It doesn't seem to be used outside of this class.

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/plugin/inputs/transports/AbstractTcpTransport.java
mpfz0r: I've added a backwards compatible constructor

graylog2-server/src/main/java/org/graylog2/Configuration.java
mpfz0r: I thought not adding a getter would send the same message.. 
But sure I could do that

graylog2-server/src/main/java/org/graylog2/shared/security/tls/DefaultTLSProtocolProvider.java
mpfz0r: good idea :+1:

graylog2-server/src/main/java/org/graylog2/configuration/TLSProtocolsConfiguration.java
mpfz0r: actually, this can be moved back into the method above.
I thought I'd need that, but didn't

bernd (approved):
LGTM :+1: Thanks!

PR #10964, opened 2021-07-02
Missing Forwarder gRPC server-side metrics

diesaulov:
While coping with https://github.com/Graylog2/forwarder/issues/32 it turned out that the Forwarder gRPC server-side metrics are declared with a wrong package name resulting into having no metrics for the gRPC calls. This PR addresses this issue and also removes irrelevant Forwarder client-side metrics.


thll (approved):
LGTM. We should also backport this to `4.1`.

PR #10977, opened 2021-07-05
Issue 10552 dashboard page context

### LINKED ISSUE #10552: Dashboard: Widget edit URL is only directly opening edit mode if widget is part of the first dashboard page.
## Expected Behavior

While editing a Widget, the URL is appended with `/dashboards/<dash-id>?focusedId=<query-id>&editing=true` when I copy and paste this URL, it should load the edit screen for the focused widget.

## Current Behavior
While editing a Widget, the URL is appended with `/dashboards/<dash-id>?focusedId=<query-id>&editing=true` but if I copy the URL and attempt to share it with a colleague, the URL is redirected back to `/dashboards/<dash-id>`, when the widget is not part of the first dashboard page.

kmerz:
## Description
Implement context which updates the URL of the browser when changing the page of a Dashboard.

## Motivation and Context
A user wants to be able to send a dashboard with a certain page tab open to a colleague.

## Also
Fixes #10552  

## Note
I stumbled into a problem here when `SizeMe` returned undefined. We checked if `SizeMe` is returning a value and if not did not render the children but instead returned `null`. This did not seem to work any longer. But when returning `<div />` instead everything is turning out fine.
The assumption is, that since we return `null` there will be no rerendering. :man_shrugging: 

## How Has This Been Tested?
- Switch pages
- Create new Page
- Close Page
- Focus a widget in a 2nd page and reload

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/QueryBar.tsx
+++ b/graylog2-web-interface/src/views/components/QueryBar.tsx
@@ -15,7 +15,7 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
-import { useCallback, useContext, useEffect } from 'react';
+import { useCallback, useContext } from 'react';
 import PropTypes from 'prop-types';
 import * as Immutable from 'immutable';
 import * as ImmutablePropTypes from 'react-immutable-proptypes';
@@ -39,12 +39,10 @@ const onTitleChange = (queryId, newTitle) => TitlesActions.set('tab', 'title', n
 
 const onSelectQuery = (queryId) => (queryId === 'new' ? NewQueryActionHandler() : ViewActions.selectQuery(queryId));
 
-const onCloseTab = (queryId: string, currentQuery: string, queries: Immutable.OrderedSet<string>) => {
+const onCloseTab = (queryId: string, currentQuery: string, queries: Immutable.OrderedSet<string>, setDashboardPage: (page: string | undefined) => void) => {
   if (queries.size === 1) {
     return Promise.resolve();
   }
-
-  let promise;
 
   if (queryId === currentQuery) {
     const indexedQueryIds = queries.toIndexedSeq();
@@ -53,12 +51,10 @@ const onCloseTab = (queryId: string, currentQuery: string, queries: Immutable.Or
     const newQuery = indexedQueryIds.filter((currentQueryId) => (currentQueryId !== queryId))
       .get(newQueryIdIndex);
 
-    promise = ViewActions.selectQuery(newQuery);
-  } else {
-    promise = Promise.resolve();
+    setDashboardPage(newQuery);
   }
 
-  return promise.then(() => QueriesActions.remove(queryId)).then(() => ViewStatesActions.remove(queryId));
+  return QueriesActions.remove(queryId).then(() => ViewStatesActions.remove(queryId));
 };
 
 type Props = {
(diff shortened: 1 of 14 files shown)


graylog2-web-interface/src/views/components/QueryBar.tsx
```diff
@@ -68,13 +65,21 @@ type Props = {
...
 const QueryBar = ({ queries, queryTitles, viewMetadata }: Props) => {
   const { activeQuery } = viewMetadata;
-  const onRemove = useCallback((queryId) => onCloseTab(queryId, activeQuery, queries), [activeQuery, queries]);
+  const { setDashboardPage } = useContext(DashboardPageContext);
+
+  const onSelectPage = useCallback((pageId) => {
+    setDashboardPage(pageId);
+
+    return onSelectQuery(pageId);
+  }, [setDashboardPage]);
```
linuspahl, line 74: An alternative solution could be, removing `onSelectQuery` and changing  `onSelectPage ` to:

```
  const onSelectPage = useCallback((pageId) => {
    if (pageId === 'new') {
      return NewQueryActionHandler().then((newPage) => {
        setDashboardPage(newPage.id);
      });
    }

    setDashboardPage(pageId);

    return ViewActions.selectQuery(pageId);
  }, [setDashboardPage]);
```

This way we are syncing the URL after creating a new page. Currently we are not defining the `page` URL query param after creating a new page. Another benefit, the `DashboardPageContext` will be slightly less complex, because we do not have to consider that the page id can be `new`.

linuspahl (commented):
Works very well, just added some suggestions.

graylog2-web-interface/src/views/components/contexts/DashboardPageContextProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+const _clearURI = (query) => new URI(query)
+  .removeSearch('page');
+
+const _updateQueryParams = (
+  newPage: string | undefined,
+  query: string,
+) => {
+  let baseUri = _clearURI(query);
+
+  if (newPage && newPage !== 'new') {
```
linuspahl, line 37: We can now remove the check `newPage !== 'new'`

linuspahl (changes requested):
Thanks for applying the feedback, there is just one resulting change left.

graylog2-web-interface/src/views/components/contexts/DashboardPageContextProvider.tsx
linuspahl: @kmerz I saw that you removed the line `if (newPage && newPage !== 'new') {` completely. I prefer to keep `if (newPage) {` because we are calling this function when calling `unSetDashboardPageParam`, which means `newPage` can be `undefined`. Currently we would add the URL query param `page` without any value in this case.

linuspahl:
@kmerz With the last tests I saw that we are not updating the URL when we duplicate a dashboard page. While looking at the code, I had the following idea. We could create a method for the `ViewStore` which takes care of setting the active query id. We can call this method wherever we set `this.activeQuery` in the `ViewStore` right now.

On a dashboard page we can define a listener for this method which takes care of updating the URL query param.
This way we can ensure we are always updating the URL, when the active query id changes.

linuspahl (approved):
Looks good to me! As discussed we take care of https://github.com/Graylog2/graylog2-server/pull/10977#issuecomment-882353577 in a follow-up PR.

Issue #10132, opened 2021-02-23
`events` and `system_events` index sets selectable as target for streams


jalogisch:
## Expected Behavior
It should not be possible to select structured internal index sets as target for streams. 

## Current Behavior
it is possible to select the internal indices `events` and `system_events` as targets for streams. 

## Your Environment

* Graylog Version: 4.0

[z#844639]

Issue #9794, opened 2020-12-14
outputs: `all events` and `all system events` can have outputs assigned


jalogisch:
## Expected Behavior
The option to assign outputs to a stream should be possible only when this will work.

## Current Behavior
You can select and assing outputs for `all events` and `all system events` streams what does not work. This gives a wrong idea what you can do with those messages and create a support burden.

## Possible Solution
Disable the output management for those streams where it is not possible.

## Your Environment
* Graylog Version: 3.3 / 4.0

Issue #5994, opened 2019-06-05
Alert Server Changes - Implement message ID field


bernd:
## Description

To efficiently get large number of messages from Elasticsearch, we have to use search-after queries in the future. Pagination or scrolling is not efficient and can lead to high load situations in Elasticsearch.

The usage of search-after queries requires the usage of a tie-breaker field for sorting the results. (see search-after documentation) This tie-breaker field must have a unique value per document in Elasticsearch. The `_id` field has a unique value per document but it's not recommended to use it as a tie-breaker value, because it's not stored as doc value but in fielddata. That means sorting on `_id` requires Elasticsearch to load lots of data into memory. To make this efficient it's recommended to use a doc value based field with a unique value.

Since we don't have such a field at the moment, we have to add one now so we can use search-after queries in the future.

## Implementation Notes

- The new message ID field needs a gl2_ prefix to make sure we don't overwrite user specific fields
- Suggestion is to use `gl2_message_id`
- The value should be mapped as "keyword" in Elasticsearch, using doc values
- Instead of using the `_id` value for the `gl2_message_id` value, we plan to use an ULID instead
  - This results in shorter IDs (26 characters for ULID vs 36 for UUID) and thus reduced storage usage
  - They are lexicographically sortable (time based UUIDs are as well, but you have to use the correct variant - which we sometimes do and sometimes don't)
- We are using ULIDs for as event IDs already, using the following Java implementation: https://github.com/huxi/sulky/tree/master/sulky-ulid

PR #10983, opened 2021-07-06
Storing indexing/processing failures

mpfz0r:
## Issue
 
Refs https://github.com/Graylog2/graylog-plugin-enterprise/issues/2127

## Overview

To implement requirements stated in the issue above we've had to first conduct some preparation works in the core Graylog component. These works include:

1. Designing a failure handling framework
2. Integrating failure producers with the framework
3. Preparing foundation for a new failure index set

### The Framework

All framework classes can be found in `org.graylog.failure` package. Below are essential ones:

* `FailureHandlingService` - a service running in a dedicate thread, whose responsibility is consuming failures submitted via `FailureSubmissionQueue`and handing them over to registered `FailureHandler`s
* `FailureSubmissionQueue` - a blocking FIFO queue, responsible for decoupling failure producers from failure consumers and for managing pressure on the failure handling framework
* `FailureSubmissionService` - a service making submission of known failures easier for failure producers

### Failure producers

* **Indexing failures.** The old failure handling logic was replaced with the new one in `org.graylog2.indexer.messages.Messages`. See the `propagateFailure` method for details
* **Processing failures.** To avoid having too many dependencies on the failure handling framework, when dealing with processing exceptions occurring at different processing stages, it was decided to collect all exceptions inside a `Message` instance (see `org.graylog2.plugin.Message#addProcessingError`) and then later handle them in `ProcessBufferProcessor`. As for now we've agreed to focus on the pipeline exceptions only.

### The new index set / stream

As per requirements all failures are supposed to be persisted in ES. To achieve this it was decided to introduce a new _system_ index and a new _system_ stream based on that index. Even though most of the concrete steps towards index creation can be put in the enterprise plugin, some changes still have to be done in the core classes. For that reason you will notice a new template type in `org.graylog2.indexer.indexset.IndexSetConfig` and a new stream id in `org.graylog2.shared.buffers.processors.Stream`. Also additional works have been performed to maintain limited operations on the _system_ index sets / streams.

### New configuration parameters

* `failure_handling_queue_capacity` - indicates how many failures can be submitted to the failure queue until it gets blocked. By default it's 1000.
* `failure_handling_shutdown_await` - indicates for how long the failure handling service should wait for new failures on shutdown until it considers that the processing is done with its job and no further failures are expected. By default it's 3 seconds.

### Metrics (a breaking change!)

The former metric `org.graylog2.periodical.IndexFailuresPeriodical.queueSize` -> `elasticsearch_output_queue_size` was removed due its misleading name and replaced with the following new metrics:

* `org.graylog.failure.FailureSubmissionQueue.queueSize` -> `failure_queue_size` - indicates current saturation of the failure queue
* `org.graylog.failure.FailureSubmissionQueue.submittedFailures` -> `failure_queue_submitted_failures` + `org.graylog.failure.FailureSubmissionQueue.consumedFailures` -> `failure_queue_consumed_failures` - these 2 metrics are aimed to help monitoring load on the failure handling framework

### Rule condition evaluation errors

For "historical" reasons `NumericExpression` and `LogicalExpression` both bypass the concept of "override `evaluateUnsafe` - call `evaluate`" (which is by default implemented within `Expression`), what in its turn leads to evaluation exceptions being **_propagated_**, instead of being converted to an evaluation error and stored in `EvaluationContext#evalErrors`.

This PR includes a fix which "duplicates" the evaluation exception handling logic for LogicalExpression(s), so that those errors become discoverable by the failure handling framework.

### E2E Tests

Certain pieces of the test code employed by `ElasticsearchE2E` were extracted into utility classes to be available for reuse in the enterprise plugin. 

P.S. It turned out that tests extending `ElasticsearchE2E` are not picked by Maven due to the missing IT suffix. And after enabling them by adding the suffix another issue popped up - the ports cannot be mapped by Docker. This must be investigated! See https://github.com/Graylog2/graylog2-server/pull/11184 for more details.

### Message ID schemas

Not to get lost in the variety of ID schemas presented in Graylog, here is a brief overview of them:

1. Events - `_id` (ULID) + `id` (ULID)
2. Messages - `_id` (UUID) + `gl2_message_id` (ULID)
3. Failures - `_id` (ULID) + `gl2_message_id` (ULID)

The 3rd one is the one the other 2 should be migrated to at some point. For more context see https://github.com/Graylog2/graylog2-server/issues/5994 and https://github.com/Graylog2/graylog2-server/issues/7986

### Events

The same restrictions, implemented for the failures, have been also applied to the Event streams / index sets:

1. Events / System events stream cannot be paused, cannot have rules, cannot have outputs, cannot be deleted, cannot be cloned. The pipelines can be connected to these streams.
2. Events / System events index sets cannot be set as default. These index sets are not available for selection when creating a new stream.

## Testing notes

Functionally for Graylog Open nothing should change after merging this PR - the failures should continue being handled in the same way as they are handled right now  (i.e. persisted in Mongo. Processing failures must be ignored).

## Notes for Reviewers

- [ ] Sync up with the author before merging

---

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/2127
Fixes https://github.com/Graylog2/graylog2-server/issues/9794
Fixes https://github.com/Graylog2/graylog2-server/issues/10132

---

Enterprise Plugin PR: https://github.com/Graylog2/graylog-plugin-enterprise/pull/2442


diesaulov (approved):

graylog2-server/src/main/java/org/graylog/failure/DefaultFailureHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    @Inject
+    public DefaultFailureHandler(IndexFailureService indexFailureService) {
+        this.indexFailureService = indexFailureService;
+    }
+
+    @Override
+    public void handle(FailureBatch failureBatch) {
+        failureBatch.getFailures().forEach(failure ->
+                indexFailureService.saveWithoutValidation(new IndexFailureImpl(ImmutableMap.<String, Object>builder()
+                    .put("letter_id", failure.failedMessage().getMessageId())
```
thll, line 42: This will change the `Letter ID` from UUID to ULID. As @mpfz0r pointed out, the UUID might be helpful to find related information in the server log when users bump their log levels. So we should probably keep the old behaviour.

image

graylog2-server/src/main/java/org/graylog/failure/FailureBatch.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private FailureBatch(List<? extends Failure> failures,
+                        Class<? extends Failure> failureClass) {
+        this.failures = ImmutableList.copyOf(failures);
+        this.failureClass = failureClass;
+        this.failures.forEach(f -> {
+            if (!f.getClass().equals(failureClass)) {
+                throw new IllegalArgumentException("Not all failures from the list are instances of " + failureClass.getName());
+            }
+        });
```
thll, line 41: Wouldn't that be a good candidate to use generics and let the compiler make sure that all failures in the batch are of the same type? E.g. make it

```
public class FailureBatch<T extends Failure>
```

mpfz0r: So at first I liked the idea and I implemented it.
But looking at the diff, I'm not so sure anymore..
Lots of clutter for not much gain. We only provide two static creator methods, and they only accept
the specific failure types.
So :shrug: I could live without generics, in this case..

Here's part of the diff (no enterprise)
```
diff --git graylog2-server/src/main/java/org/graylog/failure/DefaultFailureHandler.java graylog2-server/src/main/java/org/graylog/failure/DefaultFailureHandler.java
index 20db90b556..6248b65726 100644
--- graylog2-server/src/main/java/org/graylog/failure/DefaultFailureHandler.java
+++ graylog2-server/src/main/java/org/graylog/failure/DefaultFailureHandler.java
@@ -36,7 +36,7 @@ public class DefaultFailureHandler implements FailureHandler {
     }
 
     @Override
-    public void handle(FailureBatch failureBatch) {
+    public void handle(FailureBatch<? extends Failure> failureBatch) {
         failureBatch.getFailures().forEach(failure ->
                 indexFailureService.saveWithoutValidation(new IndexFailureImpl(ImmutableMap.<String, Object>builder()
                         .put("letter_id", failure.failedMessage().getId())
@@ -48,7 +48,7 @@ public class DefaultFailureHandler implements FailureHandler {
     }
 
     @Override
-    public boolean supports(FailureBatch failureBatch) {
+    public boolean supports(FailureBatch<? extends Failure> failureBatch) {
         return failureBatch.containsIndexingFailures();
     }
 
diff --git graylog2-server/src/main/java/org/graylog/failure/FailureBatch.java graylog2-server/src/main/java/org/graylog/failure/FailureBatch.java
index d4668828df..c58a670cc7 100644
--- graylog2-server/src/main/java/org/graylog/failure/FailureBatch.java
+++ graylog2-server/src/main/java/org/graylog/failure/FailureBatch.java
@@ -25,13 +25,13 @@ import java.util.List;
  * A structure bearing a batch of failures. It guarantees all failures
  * to be instances of the same class
  */
-public class FailureBatch {
+public class FailureBatch<T extends Failure> {
 
-    private final List<? extends Failure> failures;
-    private final Class<? extends Failure> failureClass;
+    private final List<T> failures;
+    private final Class<T> failureClass;
 
-    private FailureBatch(List<? extends Failure> failures,
-                        Class<? extends Failure> failureClass) {
+    private FailureBatch(List<T> failures,
+                         Class<T> failureClass) {
         this.failures = ImmutableList.copyOf(failures);
         this.failureClass = failureClass;
         this.failures.forEach(f -> {
@@ -46,8 +46,8 @@ public class FailureBatch {
      * @return a batch of indexing failures
      * @throws IllegalArgumentException if not all failures are instances of {@link IndexingFailure}
      */
-    public static FailureBatch indexingFailureBatch(List<IndexingFailure> indexingFailures) {
-        return new FailureBatch(indexingFailures, IndexingFailure.class);
+    public static FailureBatch<IndexingFailure> indexingFailureBatch(List<IndexingFailure> indexingFailures) {
+        return new FailureBatch<>(indexingFailures, IndexingFailure.class);
     }
 
     /**
@@ -55,24 +55,25 @@ public class FailureBatch {
      * @return a batch of processing failures
      * @throws IllegalArgumentException if not all failures are instances of {@link ProcessingFailure}
      */
-    public static FailureBatch processingFailureBatch(List<ProcessingFailure> processingFailures) {
-        return new FailureBatch(processingFailures, ProcessingFailure.class);
+    public static FailureBatch<ProcessingFailure> processingFailureBatch(List<ProcessingFailure> processingFailures) {
+        return new FailureBatch<>(processingFailures, ProcessingFailure.class);
     }
 
     /**
      * Creates a batch containing only one processing failure
+     *
      * @param processingFailure a processing failure to include in this batch
      * @return a batch with the processing failure
      * @throws IllegalArgumentException if the failure is not instances of {@link ProcessingFailure}
      */
-    public static FailureBatch processingFailureBatch(ProcessingFailure processingFailure) {
+    public static FailureBatch<ProcessingFailure> processingFailureBatch(ProcessingFailure processingFailure) {
         return processingFailureBatch(ImmutableList.of(processingFailure));
     }
 
     /**
      * @return a list of failures within the batch. The returned collection is immutable.
      */
-    public List<? extends Failure> getFailures() {
+    public List<T> getFailures() {
         return failures;
     }
 
@@ -106,8 +107,12 @@ public class FailureBatch {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
         final FailureBatch that = (FailureBatch) o;
         return Objects.equal(failures, that.failures) && Objects.equal(failureClass, that.failureClass);
     }
diff --git graylog2-server/src/main/java/org/graylog/failure/FailureHandler.java graylog2-server/src/main/java/org/graylog/failure/FailureHandler.java
index f68b26893d..41cb645036 100644
--- graylog2-server/src/main/java/org/graylog/failure/FailureHandler.java
+++ graylog2-server/src/main/java/org/graylog/failure/FailureHandler.java
@@ -34,7 +34,7 @@ public interface FailureHandler {
      *
      * @param failureBatch a batch of failures, supported by this handler
      */
-    void handle(FailureBatch failureBatch);
+    void handle(FailureBatch<? extends Failure> failureBatch);
 
     /**
      * Guides the master failure handling service, when deciding
@@ -43,7 +43,7 @@ public interface FailureHandler {
      * @param failureBatch a batch of failures to test
      * @return true if the batch can be processed by this handler
      */
-    boolean supports(FailureBatch failureBatch);
+    boolean supports(FailureBatch<? extends Failure> failureBatch);
 
     /**
      * Guides the master failure handling service, when deciding
diff --git graylog2-server/src/main/java/org/graylog/failure/FailureHandlingService.java graylog2-server/src/main/java/org/graylog/failure/FailureHandlingService.java
index 67398eed2a..0f9f720b31 100644
--- graylog2-server/src/main/java/org/graylog/failure/FailureHandlingService.java
+++ graylog2-server/src/main/java/org/graylog/failure/FailureHandlingService.java
@@ -127,7 +127,7 @@ public class FailureHandlingService extends AbstractExecutionThreadService {
         logger.debug("The service has been interrupted.");
     }
 
-    private void handle(FailureBatch failureBatch) {
+    private void handle(FailureBatch<? extends Failure> failureBatch) {
         suitableHandlers(failureBatch)
                 .forEach(handler -> {
                     try {
@@ -149,7 +149,7 @@ public class FailureHandlingService extends AbstractExecutionThreadService {
         }
     }
 
-    private List<FailureHandler> suitableHandlers(FailureBatch failureBatch) {
+    private List<FailureHandler> suitableHandlers(FailureBatch<? extends Failure> failureBatch) {
         final List<FailureHandler> suitableHandlers = suitableHandlers(failureHandlers, failureBatch)
                 .filter(FailureHandler::isEnabled)
@@ -162,7 +162,7 @@ public class FailureHandlingService extends AbstractExecutionThreadService {
         }
     }
 
-    private Stream<FailureHandler> suitableHandlers(Collection<FailureHandler> handlers, FailureBatch failureBatch) {
+    private Stream<FailureHandler> suitableHandlers(Collection<FailureHandler> handlers, FailureBatch<? extends Failure> failureBatch) {
         return handlers.stream()
                 .filter(h -> h.supports(failureBatch));
     }
diff --git graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
index afe07e9c0a..45eed629b8 100644
--- graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
+++ graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
@@ -47,7 +47,7 @@ class FailureSubmissionQueue {
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
-    private final BlockingQueue<FailureBatch> queue;
+    private final BlockingQueue<FailureBatch<? extends Failure>> queue;
     private final Configuration configuration;
     private final Meter submittedFailureBatches;
     private final Meter submittedFailures;
@@ -98,8 +98,8 @@ class FailureSubmissionQueue {
      * @return one batch from the queue. If the queue is empty,
      * waits for a batch to become available.
      */
-    FailureBatch consumeBlocking() throws InterruptedException {
-        final FailureBatch fb = queue.take();
+    FailureBatch<? extends Failure> consumeBlocking() throws InterruptedException {
+        final FailureBatch<? extends Failure> fb = queue.take();
         consumedFailureBatches.mark();
         consumedFailures.mark(fb.size());
         return fb;
diff --git graylog2-server/src/test/java/org/graylog/failure/FailureSubmissionServiceTest.java graylog2-server/src/test/java/org/graylog/failure/FailureSubmissionServiceTest.java
index c3c0039e3c..1d2b54a130 100644
--- graylog2-server/src/test/java/org/graylog/failure/FailureSubmissionServiceTest.java
+++ graylog2-server/src/test/java/org/graylog/failure/FailureSubmissionServiceTest.java
@@ -38,7 +38,7 @@ public class FailureSubmissionServiceTest {
     private final FailureHandlingConfiguration failureHandlingConfiguration = Mockito.mock(FailureHandlingConfiguration.class);
     private final FailureSubmissionService underTest = new FailureSubmissionService(failureSubmissionQueue, failureHandlingConfiguration);
 
-    private final ArgumentCaptor<FailureBatch> failureBatchCaptor = ArgumentCaptor.forClass(FailureBatch.class);
+    private final ArgumentCaptor<FailureBatch<? extends Failure>> failureBatchCaptor = ArgumentCaptor.forClass(FailureBatch.class);
 
     @Test
     public void submitIndexingErrors_allIndexingErrorsTransformedAndSubmittedToFailureQueue() throws Exception {

```

graylog2-web-interface/src/components/event-definitions/event-definition-types/withStreams.jsx
```diff
@@ -25,13 +25,17 @@ export default function withStreams(WrappedComponent, hiddenStreams = []) {
...
+      };
+    }
 
     componentDidMount() {
       StreamsStore.load((streams) => {
-        let filteredStreams = streams;
+        let filteredStreams = streams.filter((s) => s.is_editable);
 
         if (hiddenStreams.length !== 0) {
           filteredStreams = streams.filter((s) => !hiddenStreams.includes(s.id));
```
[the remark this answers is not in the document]
kyleknighted: @ousmaneo Regarding the lint error - the only thing modified in that file was removing a constant that was not used anyway and could've caused confusion in the future. I think this PR is complicated enough and no need to add to that with modifying a mostly untouched file.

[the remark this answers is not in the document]
kyleknighted: @ousmaneo Sorry, misspoke the first time around on the `hiddenStreams`... too many things in my head 😬

We can remove the `hiddenStreams`, but I had left it there incase anyone has reason to hide a stream in the frontend but not necessarily because the backend said that `is_editable` was false.

If you think it should be fully removed, just let me know. Otherwise, leaves some room for additional functionality should we need it.

ousmaneo (approved):
Hi @kyleknighted, Make sense.
I wrote this because after checking usage it was used only to hide `streams` that were in the `constant`.  I agree that it can be additional functionality. Approving 👍

PR #10986, opened 2021-07-06
Deprecate shutting a node down via the API

### LINKED ISSUE #10668: Remove `Graceful shutdown` from `Nodes` options in Graylog
## Expected Behavior

As a Graylog user, I want to have tight control over both Graylog and the underlying systems that it runs on. I expect, to that end, that all process management is done by the OS's process manager.

## Current Behavior

Currently, a Graylog admin/user can perform a graceful shutdown of Graylog, bypassing things like systemd. This can lead to confusion and issues when the user who performed the shutdown from the UI doesn't have underlying system permissions. 

CleanShot 2021-05-18 at 08 58 16@2x

thll:
Deprecates the API endpoints used to shut down a node.
- Adds deprecation annotations to discourage usage in the code
- Removes Swagger annotations to hide the endpoints in the API browser
- Logs a warning when the endpoints are called
- Returns a warning message in the API response where applicable

Refs #10668

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemShutdownResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemShutdownResource.java
@@ -17,9 +17,6 @@
 package org.graylog2.rest.resources.cluster;
 
 import com.codahale.metrics.annotation.Timed;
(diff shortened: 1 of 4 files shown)


edmundoa (approved):
It looks good to me and seems to be working as expected, feel free to merge if you don't need any other opinions 👍

PR #10988, opened 2021-07-06
Typing more views modules.

dennisoelkers:
## Motivation and Context

This PR is migrating lots of modules from the `views` subsystem to use TS. In most cases this consists of:

  - renaming `*.jsx?` to `.tsx?`
  - fixing type errors in module
  - fixing type errors in other modules resulting in previously untyped
    modules now having types

Functionality should not be impacted by this. It was taken care to do as little as possible changes on the value layer and restrict changes to typings as much as possible.

Modules from the `aggregationbuilder` are left out on purpose, as this module will probably be removed in case it's not rolled back.

Some modules are left for migration which are the target of a follow-up PR to avoid increasing the size of this one even more.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2444

## Types of changes

## Checklist:


kmerz (approved):
LGTM :+1: 
Worked well on my machine :heavy_check_mark:

PR #10994, opened 2021-07-07
Make LookupTableService database independent 

bernd:
Create a `LookupTableConfigService` interface and use it in `LookupTableService` to make it possible to create a `LookupTableService` instance that doesn't depend on the database.

Also make the constructor in the `Debug` pipeline function public to allow the creation of custom instances.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
@@ -24,6 +24,7 @@ import org.graylog2.lookup.adapters.DnsLookupDataAdapter;
 import org.graylog2.lookup.adapters.HTTPJSONPathDataAdapter;
 import org.graylog2.lookup.caches.CaffeineLookupCache;
 import org.graylog2.lookup.caches.NullCache;
+import org.graylog2.lookup.db.DBLookupTableConfigService;
 import org.graylog2.plugin.inject.Graylog2Module;
 import org.graylog2.system.urlwhitelist.UrlWhitelistNotificationService;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
```diff
@@ -41,6 +42,7 @@ protected void configure() {
         binder().bind(UrlWhitelistNotificationService.class).in(Scopes.SINGLETON);
         binder().bind(AllowedAuxiliaryPathChecker.class).in(Scopes.SINGLETON);
 
+        bind(LookupTableConfigService.class).to(DBLookupTableConfigService.class).asEagerSingleton();
```
danotorrey, line 45: Is there a particular reason you opted to use an eager singleton here instead of an optional binding (with default)?

danotorrey (approved):
Looks good, and tests successfully. Just one question:

graylog2-server/src/main/java/org/graylog2/lookup/LookupModule.java
bernd: @danotorrey An optional binding would allow us to override a binding when running with a different application profile or loading a plugin. That would override the binding globally, though. In this case we always want to use the `DBLookupTableConfigService` to keep the lookup feature working as it is. We only want the option to create custom `LookupTableService` instances manually that don't depend on database backend configuration.

danotorrey: Ah, of course. Makes sense.

danotorrey:
This looks good as coded, so I will merge it now.

PR #10998, opened 2021-07-08
Implement generic OIDC login

patrickmann:
This is the companion PR to https://github.com/Graylog2/graylog-plugin-enterprise/pull/2410 and https://github.com/Graylog2/graylog-plugin-cloud/pull/944

We introduce more generic naming in support of increased inheritance and code re-use.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2410

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackend.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackend.java
@@ -41,6 +41,19 @@ public interface AuthServiceBackend {
 
     default Optional<AuthenticationDetails> authenticateAndProvision(AuthServiceToken token,
             ProvisionerService provisionerService) {
+        return authenticateAndProvision(token, provisionerService, true);
+    }
+
+    /**
+     * Authenticate user and provision a local user record
+     * @param token
+     * @param provisionerService
+     * @param isExternal user is managed by external service (e.g. Okta) and cannot be modified locally
+     * @return authentication details; empty if a problem was encountered
+     */
+    default Optional<AuthenticationDetails> authenticateAndProvision(AuthServiceToken token,
+                                                                     ProvisionerService provisionerService,
+                                                                     boolean isExternal) {
         log.debug("Cannot authenticate by token. Token-based authentication is not supported by auth service backend " +
                 "type <" + backendTitle() + ">.");
         return Optional.empty();
--- a/graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackend.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackend.java
@@ -41,19 +41,6 @@ public interface AuthServiceBackend {
 
     default Optional<AuthenticationDetails> authenticateAndProvision(AuthServiceToken token,
             ProvisionerService provisionerService) {
-        return authenticateAndProvision(token, provisionerService, true);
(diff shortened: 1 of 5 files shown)


thll:
@patrickmann the diff appears to show many unrelated changes. could you rebase the PR to the current `master` please?

thll:
From the description: 

> /jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2410

I think we shouldn't have this dependency here. The build should pass without it, otherwise we'd have a problem for the non-enterprise version. Can we remove this?

thll:
> From the description:
> 
> > /jenkins-pr-deps [Graylog2/graylog-plugin-enterprise#2410](https://github.com/Graylog2/graylog-plugin-enterprise/pull/2410)
> 
> I think we shouldn't have this dependency here. The build should pass without it, otherwise we'd have a problem for the non-enterprise version. Can we remove this?

My bad, sorry for the noise. I was confusing this with the github actions build step which passes just fine. I added the Jenkins statement again.

graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackend.java
```diff
@@ -41,6 +41,19 @@
...
+    /**
+     * Authenticate user and provision a local user record
+     * @param token
+     * @param provisionerService
+     * @param isExternal user is managed by external service (e.g. Okta) and cannot be modified locally
+     * @return authentication details; empty if a problem was encountered
+     */
+    default Optional<AuthenticationDetails> authenticateAndProvision(AuthServiceToken token,
+                                                                     ProvisionerService provisionerService,
+                                                                     boolean isExternal) {
```
thll, line 56: Do we still need this change? I don't see it being used with the current state of the implementation so I vote for reverting it to keep the interface clean.

graylog2-server/src/main/java/org/graylog/security/authservice/AuthenticationDetails.java
```diff
@@ -32,6 +32,8 @@ public static Builder builder() {
         return Builder.create();
     }
 
+    public abstract AuthenticationDetails.Builder toBuilder();
```
thll, line 35: It's nice to have this, but we don't have code calling this. If we decide to keep the PR for the UI changes, we should leave this change in.

thll (changes requested):
As the backend code depending on this PR has been refactored, I think from a backend perspective this PR is now obsolete. We could keep the change in `AuthenticationDetails.java` though.

I didn't look at the UI changes.

thll (approved):
Backend changes are fine.
Feel free to merge once the frontend changes have been reviewed.

graylog2-web-interface/src/logic/authentication/okta/types.ts
```diff
@@ -20,22 +20,25 @@ import * as Immutable from 'immutable';
 import AuthenticationBackend from 'logic/authentication/AuthenticationBackend';
 
 export interface OktaBackendConfig {
-  type: 'okta';
-  oktaBaseUrl: string;
+  type: string;
+  baseUrl?: string;
+  oktaBaseUrl?: string;
```
edmundoa, line 25: Should we change this types like we did in enterprise? That is, to have one for OIDC and one for Okta.

edmundoa (approved):
LGTM, thank you for all the work! 👍

Issue #8146, opened 2020-05-18
Elasticsearch integration tests should test multiple versions


alex-konn:
## Expected Behavior
We should have integration tests that verify the system works with all supported Elasticsearch versions
By default all integration tests should test all supported major versions.
In addition it should be possible to define one or more version in the implementing  test class, which will be used instead of the default versions.

## Current Behavior
Only one version can be tested

## Context
It's probably worthwhile to do #7655 before, maybe only for Elasticsearch. The MongoDb stuff can be done later.

dennisoelkers:
Fixed in #11011.

PR #11011, opened 2021-07-12
Backend Tests with multiple ES versions

janheise:
## Description, Motivation and Context
WIP, replacement for #8168, adresses #8146

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2714

This PR addresses multiple problems:
- fix defective instance caching using containers in tests
- automated testing against multiple versions/combinations of Elasticsearch/MongoDB
- keep generation of container instances to a minimum

To achieve the goals, a TestEngine as a plugin to JUnit5/jupiter was created, that manages the grouping of tests to combinations of containers and also manages lifecyle.
- a new lifecycle "VM" has been created to signal that the methods of the annotated test-class can run in a global container that
- lifecycle "class" re-creates containers for the given class
- lifecycle "method" has been removed. First, because it should really be avoided - you should clean-up after a test - and second,  because makes the test-engine a lot more complicated. If you want to separate tests in context, create a second test-class with lifecycle "class". 
- all used JUnit-Annotations have been ported, but not the full set
- version numbers for used containers are in one place, but you can specify your own version numbers if needed 

Todos for follow-up PRs:
- check, if mongo-fixtures run correctly in all lifecycles => not perfect right now, maybe use @beforeAll and deprecate the fixtures from the annotation

- inject running version into test class so that you can have version specific handling of things that you know are different between versions? => Integrate @TestInstance(PER_CLASS) so @beforeAll can handle this
- in case only an ES instance is needed, don't start a Mongo instance, too?
## Types of changes

## Checklist:


bernd:
@janheise Is this a replacement for https://github.com/Graylog2/graylog2-server/pull/8168 ?

janheise:
@bernd yes, I'll add it in my comments. with #8168 and my first solution I could not find a way to have full control over the lifecycle of the various docker containers that have to be instantiated (especially to make sure that they are not instantiated in parallel)

bernd:
@janheise Awesome, thank you! :+1:

todvora (approved):
LGTM, Thanks!

PR #11012, opened 2021-07-12
Typing frontend routing components.

dennisoelkers:
## Motivation and Context

This PR is typing the central routing components of the frontend. These include our central `AppRouter` component as well as the `Routes` it uses to manage frontend routes and `ApiRoutes` doing the same to generate backend routes.

## Types of changes

## Checklist:


graylog2-web-interface/src/routing/Routes.tsx
```diff
@@ -19,6 +19,19 @@ import URI from 'urijs';
...
 import AppConfig from 'util/AppConfig';
 import { extendedSearchPath, viewsPath } from 'views/Constants';
+import { TimeRangeTypes } from 'views/logic/queries/Query';
+
+type RoutesRelativeTimeRange = {
+  relative: number
+};
+type RoutesAbsoluteTimeRange = {
+  from: string,
+  to: string,
```
linuspahl, line 29: What about using a `number` for `from` and `to` as well? It just provides a small benefit, but this way we would ensure that we do not define any string for these attributes.

graylog2-web-interface/src/routing/AppRouter.tsx
```diff
@@ -0,0 +1,407 @@
...
+                             component={ShowEventNotificationPage} />
+                      <Route exact
+                             path={Routes.show_alert_condition(':streamId', ':conditionId')}
+                             component={EditAlertConditionPage} />
+                      <Route exact path={Routes.show_alert(':alertId')} component={ShowAlertPage} />
+
+                      {!isCloud && <Route exact path={Routes.SYSTEM.INPUTS} component={InputsPage} />}
+                      {!isCloud && <Route exact path={Routes.node_inputs(':nodeId')} component={NodeInputsPage} />}
+                      {!isCloud && (
+                      <Route exact path={Routes.global_input_extractors(':inputId')} component={ExtractorsPage} />
```
linuspahl, line 196: If it suits you, let's fix the indentations.

linuspahl (commented):
Could not find any problems.

graylog2-web-interface/src/routing/Routes.tsx
linuspahl: This type is actually correct, I mixed it up with the following relative time range type: `{ type: "relative", to: number, from: number }` which is not yet defined. As discussed we can add this type when we need it.

linuspahl (approved):

PR #11016, opened 2021-07-13
Store messages that failed pipeline processing

mpfz0r:
This submits messages that experienced failures
during pipeline processing to the FailureHandlerService.

Introduce new `Indexable#getMessageId()` that returns the message ULID
instead of the deprecated `getId()` that returns a UUID.
While migrating everything away from the UUID to the ULID
is out of scope for this PR, we don't want to introduce new code using
UUIDs.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2461


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/PipelineInterpreter.java
```diff
@@ -170,6 +186,31 @@ public Messages process(Messages messages, InterpreterListener interpreterListen
...
+            message.setFilterOut(true);
+            messageQueueAcknowledger.acknowledge(message);
+            failedMessages.mark();
+            submitFailure(message, processingError);
+        }
+    }
+
+    private void submitFailure(Message message, String error) {
+        try {
+            final String messageJson = objectMapper.writeValueAsString(message.toElasticSearchObject(objectMapper, new Meter()));
```
diesaulov, line 203: This is already the 2nd place where we have to manually perform the serialization before submitting a failure.

What if we in `org.graylog.failure.Failure` change the signature of the method from `String messageJson()` to `Indexable message()` and then perform all the necessary serialization in `EnterpriseFailureHandler`?

mpfz0r: good idea. simpler this way :+1:

diesaulov (approved):

PR #11017, opened 2021-07-13
Do not set 'to' when setting TimeRange to 'All Time'

### LINKED ISSUE #10920: RC1: Datetimepicker until can not be changed if 'All time' is ticked - search afterwards in error state but no feedback in webinterface
## Expected Behavior
Working search and an 'All time' checkbox resulting in overruling 'until'

## Current Behavior

Search can not be executed and returns API Error

1. Disable now checkbox
2. Enable 'All' Checkbox
3. Until can not be changed anymore resulting in strange lookin datepicker? and a now working search
4.

grafik
grafik

```
Cannot construct instance of `org.graylog2.plugin.indexer.searches.timeranges.RelativeRange$Builder`, problem: `from` must be greater than `to`!
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 193] (through reference chain: org.graylog.plugins.views.search.AutoValue_Search$Builder["queries"]->org.graylog.plugins.views.search.AutoValue_Query$Builder["timerange"])
```

## Your Environment

* Graylog Version: 4.1.0 rc.1
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kmerz:
## Motivation
Prior to this change, the backend raised an error if a user switch the
timerange from 'from -> to' to 'All time' since 'to' was still send to
the server, which the server refuesed to accept.

## Description
This change will change the check where the submitted timerange is
checked to be 'All Time'
Before we did not see the timerange as 'All Time' if 'to' was still set,
but to would have been unset AFTER it acknowleged the time range a s
'All Time'.

Fixes #10920.

## How Has This Been Tested?
- set timerange with from and to
- checked 'All time'
- applied the settings
- started a search

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/logic/queries/IsAllMessagesRange.ts
+++ b/graylog2-web-interface/src/views/logic/queries/IsAllMessagesRange.ts
@@ -20,7 +20,7 @@ import { isTypeRelativeWithEnd, isTypeRelativeWithStartOnly } from 'views/typeGu
 import { TimeRange } from 'views/logic/queries/Query';
 
 const isAllMessagesRange = (timeRange: TimeRange) => {
-  return (isTypeRelativeWithEnd(timeRange) && timeRange.from === RELATIVE_ALL_TIME && !timeRange.to) || (isTypeRelativeWithStartOnly(timeRange) && timeRange.range === RELATIVE_ALL_TIME);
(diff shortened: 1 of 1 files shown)


linuspahl (approved):
Looks good to me! I realized we are not updating the preview of the search date time picker properly, when reseting the `until` value, but that is something we should fix in a follow up PR.

PR #11020, opened 2021-07-14
Fix returning count of saved searches list

kmerz:
## Motivation
Prior to this change, when returning the list of saved searches the
pagination info always had a count of zero.

## Description
This change will remove the seemingly unnecessary construction of
PagianationList with empty arrays.

## How Has This Been Tested?
- Use the saved searches end point and check the pagination info

## Types of changes

## Backport
Needs backport to 4.1

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewSummaryService.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewSummaryService.java
@@ -24,9 +24,7 @@ import org.graylog2.search.SearchQuery;
 import org.mongojack.DBQuery;
 
 import javax.inject.Inject;
-import java.util.ArrayList;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewSummaryService.java
```diff
@@ -40,16 +38,22 @@ protected ViewSummaryService(MongoConnection mongoConnection,
     }
 
     private PaginatedList<ViewSummaryDTO> searchPaginated(DBQuery.Query query,
```
janheise, line 40: searchPaginated() seems to be no longer in use anywhere (according to IntelliJ) - why does ist still exist?

janheise:
ViewSummaryService was introduced by me as a copy from ViewService - this class may hold the same error. See: [ViewService](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java)

janheise (approved):
Works!

PR #11022, opened 2021-07-14
Feature flags v1

AntonEbel:
close Graylog2/graylog-plugin-cloud#938

- [ ] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [ ] Sync up with the author before merging

/jenkins-pr-deps Graylog2/graylog-plugin-cloud#952


patrickmann (commented):
Is there a test that reads a config file of flag?

patrickmann (approved):
Looks good

graylog2-server/src/main/java/org/graylog2/featureflag/ImmutableFeatureFlags.java
```diff
@@ -0,0 +1,96 @@
...
+                .collect(Collectors.toMap(e -> toUpperCase(e.getKey()), e -> new FeatureFlag(e.getKey(), e.getValue(), metricRegistry)));
+    }
+
+    @Override
+    public Map<String, String> getAll() {
+        return flags.values().stream().collect(Collectors.toMap(flag -> flag.name, flag -> flag.value));
+    }
+
+    @Override
+    public boolean isOn(String feature, boolean defaultValue) {
```
bernd, line 48: I was wondering what the use case for the `defaultValue` is. According to the pitch a feature flag should be off by default, when would one set the `defaultValue` to true?

AntonEbel: The purpose of the defaultValue was to give the developer some flexibility and to make clear what the result would be if the feature is not present. But you are right I can't think of any use case where I would set the default value to true. Maybe it's not worth it to have this flexibility. If we remove the default parameter we should provide at least some JavaDoc to make clear what is the fallback behavior or come up with good name for the method.

bernd: Sounds good! We can always add a method with a default value later if we see that we need it. Taking the parameter away is way harder once the API is in use. :+1:

patrickmann (approved):

graylog2-server/src/main/java/org/graylog2/featureflag/ImmutableFeatureFlags.java
AntonEbel: I removed the defaultValue parameter.

PR #11023, opened 2021-07-14
Also exclude outputs from search dependencies

### LINKED ISSUE #11021: Search created by user for shared stream with output incorrectly states missing dependency when shared
## Current Behavior

When a user creates a search that references a stream shared with the user which has an output attached, the share button for that search when opened by the admin user incorrectly states a missing dependency for the output.

image

1. Create stream
2. Attach output to stream
3. Share stream with user
4. Log in as User
5. Save search referencing stream
6. Open search as admin
7. Open "Share" dialog

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

mpfz0r:
A search can reference a stream, which can reference an output.
We don't care for these permissions when sharing a search.

Fixes #11021


bernd (approved):
LGTM :+1: 

I think we should create a 4.1 backport for this.

PR #11029, opened 2021-07-16
Draining the failure submission queue upon server shutdown

diesaulov:
Since there might be no second chance to handle failures upon server restart (mainly for 2 reasons: 1) all the messages are acknowledged regardless of their ES indexing result 2) we commit the highest journal offset, what of course might shadow some failures) our goal is to handle as many failures as possible before shutting down.

Apart from that the PR contains some cosmetic and functional improvements.

Issue: https://github.com/Graylog2/graylog-plugin-enterprise/issues/2127


mpfz0r (approved):
LGTM :+1:

PR #11033, opened 2021-07-19
Flag failures and event streams as non-editable. Handle special IndexSets

mpfz0r:
These streams should not be edited (e.g. assigned to a different index)
Also outputs or stream rules don't work with these Streams.

Technically, they could be deleted, but the migration will recreate
them, so that's also pointless.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2442


diesaulov:
A few findings discovered during testing:

1. The API allows to create stream rules and Outputs for the new Failure stream.
2. The pipelines can be connected to the new Failure stream.
3. The failure index set is available for selection while creating a new stream.
4. On the System / Indices page the Failure stream can be selected as default.

mpfz0r:
> A few findings discovered during testing:

should be fixed now

graylog2-server/src/main/java/org/graylog2/contentpacks/EntityDescriptorIds.java
```diff
@@ -58,7 +58,7 @@ public static EntityDescriptorIds of(Collection<EntityDescriptor> entityDescript
     }
 
     public static boolean isDefaultStreamDescriptor(EntityDescriptor descriptor) {
-        return ModelTypes.STREAM_V1.equals(descriptor.type()) && Stream.isDefaultStreamId(descriptor.id().id());
+        return ModelTypes.STREAM_V1.equals(descriptor.type()) && Stream.isSystemStreamId(descriptor.id().id());
```
diesaulov, line 61: Shouldn't we rename the method to `isSystemStreamDescriptor`?

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/rules/StreamRuleResource.java
```diff
@@ -118,7 +118,7 @@ public SingleStreamRuleSummaryResponse update(@ApiParam(name = "streamid", value
                                                   @ApiParam(name = "JSON body", required = true)
                                                   @Valid @NotNull CreateStreamRuleRequest cr) throws NotFoundException, ValidationException {
         checkPermission(RestPermissions.STREAMS_EDIT, streamid);
-        checkNotDefaultStream(streamid, "Cannot update stream rules on default stream.");
+        checkNotEditable(streamid, "Cannot update stream rules on non-editable streams.");
```
diesaulov, line 121: The rules are now disabled for non-editable stream 👍 However outputs can still be created / assigned.

graylog2-server/src/main/java/org/graylog2/contentpacks/EntityDescriptorIds.java
mpfz0r: makes sense :+1:

diesaulov (approved):

PR #11034, opened 2021-07-19
Fix returning count and grand_total of views

kmerz:
## Motivation
Prior to this change, when returning the list of saved searches
or dashboards the pagination info always had a count of zero
and the combined result of dashboards and saved searches count in
grand_total.

## Description
This PR will slightly change the pagination api so grand total also can
have a DBQuery which will be based on the type (search / dashboard) and
will initiate the newly created PaginatedList with the correct count.

## How Has This Been Tested?
- Retrive a list of saved searches and check the count and grand total
- Same for dashboards

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
@@ -39,26 +39,38 @@ public class PaginatedList<E> extends ForwardingList<E> {
 
     /**
      * Creates a PaginatedList
+     *
      * @param delegate the actual entries
-     * @param total the count of all entries (ignoring pagination)
-     * @param page  the page this PaginatedList represents
-     * @param perPage the size limit for each page
+     * @param total    the count of all entries (ignoring pagination)
+     * @param page     the page this PaginatedList represents
+     * @param perPage  the size limit for each page
      */
     public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage) {
         this(delegate, total, page, perPage, null);
     }
 
+    public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, int count) {
+        this(delegate, total, page, perPage, count, null);
+    }
+
     /**
      * Creates a PaginatedList
-     * @param delegate the actual entries
-     * @param total the count of all entries (ignoring pagination)
-     * @param page  the page this PaginatedList represents
-     * @param perPage the size limit for each page
+     *
+     * @param delegate   the actual entries
+     * @param total      the count of all entries (ignoring pagination)
+     * @param page       the page this PaginatedList represents
+     * @param perPage    the size limit for each page
      * @param grandTotal the count of all entries (ignoring query filters and pagination)
      */
     public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, Long grandTotal) {
         this.delegate = delegate;
         this.paginationInfo = PaginationInfo.create(total, delegate.size(), page, perPage);
+        this.grandTotal = grandTotal;
+    }
+
+    public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, int count, Long grandTotal) {
+        this.delegate = delegate;
+        this.paginationInfo = PaginationInfo.create(total, count, page, perPage);
         this.grandTotal = grandTotal;
     }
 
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
```diff
@@ -62,6 +68,12 @@ public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage
         this.grandTotal = grandTotal;
     }
 
+    public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, int count, Long grandTotal) {
+        this.delegate = delegate;
+        this.paginationInfo = PaginationInfo.create(total, count, page, perPage);
+        this.grandTotal = grandTotal;
+    }
+
```
bernd, line 76: In `ViewService#searchPaginated` we use `PaginatedList` as a mutable `List`. This works because `PaginatedList` extends `ForwardingList`. This use case hasn't been anticipated when we created the `PaginatedList` implementation but is a valid use case, of course. (otherwise `PaginatedList` should have been immutable)

The problem here is that we compute the `PaginationInfo` in the constructor. That means the info will be wrong when we pass in an empty `delegate` and later populate it. Your change fixes the issue by passing in the `count`, but I think we should rather compute the `PaginationInfo` when it's requested. I the `ViewService#searchPaginated` use case we know the number of elements that will be in the `delegate`, but there might be use cases where we don't know that when creating the `PaginatedList`.

I would suggest the following change to your last changes. What do you think?

```diff
diff --git graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
index 8b6abe9c85..feb2c309e8 100644
--- graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
+++ graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
@@ -82,8 +82,8 @@ public class ViewService extends PaginatedDbService<ViewDTO> {
         return viewsList.stream()
                 .map(this::requirementsForView)
                 .collect(Collectors.toCollection(() -> viewsList.grandTotal()
-                        .map(grandTotal -> new PaginatedList<ViewDTO>(new ArrayList<>(viewsList.size()), viewsList.pagination().total(), page, perPage, viewsList.size(), grandTotal))
-                        .orElseGet(() -> new PaginatedList<>(new ArrayList<>(viewsList.size()), viewsList.pagination().total(), page, perPage, viewsList.size()))));
+                        .map(grandTotal -> new PaginatedList<ViewDTO>(new ArrayList<>(viewsList.size()), viewsList.pagination().total(), page, perPage, grandTotal))
+                        .orElseGet(() -> new PaginatedList<>(new ArrayList<>(viewsList.size()), viewsList.pagination().total(), page, perPage))));
     }
 
     public PaginatedList<ViewDTO> searchPaginated(SearchQuery query,
diff --git graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
index 6dd369ddec..3073eec26e 100644
--- graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
+++ graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
@@ -32,8 +32,9 @@ import java.util.Optional;
 public class PaginatedList<E> extends ForwardingList<E> {
 
     private final List<E> delegate;
-
-    private final PaginationInfo paginationInfo;
+    private final int total;
+    private final int page;
+    private final int perPage;
 
     private final Long grandTotal;
 
@@ -49,10 +50,6 @@ public class PaginatedList<E> extends ForwardingList<E> {
         this(delegate, total, page, perPage, null);
     }
 
-    public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, int count) {
-        this(delegate, total, page, perPage, count, null);
-    }
-
     /**
      * Creates a PaginatedList
      *
@@ -64,13 +61,9 @@ public class PaginatedList<E> extends ForwardingList<E> {
      */
     public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, Long grandTotal) {
         this.delegate = delegate;
-        this.paginationInfo = PaginationInfo.create(total, delegate.size(), page, perPage);
-        this.grandTotal = grandTotal;
-    }
-
-    public PaginatedList(@Nonnull List<E> delegate, int total, int page, int perPage, int count, Long grandTotal) {
-        this.delegate = delegate;
-        this.paginationInfo = PaginationInfo.create(total, count, page, perPage);
+        this.total = total;
+        this.page = page;
+        this.perPage = perPage;
         this.grandTotal = grandTotal;
     }
 
@@ -80,7 +73,7 @@ public class PaginatedList<E> extends ForwardingList<E> {
     }
 
     public PaginationInfo pagination() {
-        return paginationInfo;
+        return PaginationInfo.create(total, delegate.size(), page, perPage);
     }
 
     public Optional<Long> grandTotal() {
```

bernd (changes requested):
One comment about the `PaginatedList` change. The rest looks good, thank you! :+1:

graylog2-server/src/main/java/org/graylog2/database/PaginatedList.java
bernd: Oh, I think we should put this `PaginatedList` change into master and 4.1 as well because that's a generic issue with the implementation.

bernd (approved):

PR #10941, opened 2021-06-25
Installing type definitions for `react-select`.

linuspahl:
## Motivation and Context

With this PR we are installing `@types/react-select` which improves the type definition for our common `Select` component. Please note, I tried to improve the `Select` type inference, but decided to do it in a follow-up PR: https://github.com/Graylog2/graylog2-server/pull/11036

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2488

PR #11036, opened 2021-07-19
Improving common `Select` type inference.

linuspahl:
## Description

This PR is a follow-up PR for https://github.com/Graylog2/graylog2-server/pull/10941 which installed `@types/react-select`.

This is still a draft PR because the following problem still exist:
- We want to infer the value type based on the provided options, but using `withTheme` in `Select` is breaking this type inference.
- The type inference is also breaking when we define the following prop types:
```
displayKey: Proptypes.string,
valueKey: Proptypes.string,
options: PropTypes.array.isRequired,
```
- Currently the options prop is not being typed correctly. I tried to define `Opts extends Options<LabelKey, ValueKey, OptionValue>`, but it results in a circularly referenced type.

## Types of changes

## Checklist:


linuspahl:
Closing this for now, since the `Select` component has changes a lot in the meantime and it makes more sense to create a new PR for this improvement.

PR #11056, opened 2021-07-22
Create random http message timestamps early

mpfz0r:
This change moves the timestamp creation for the random http message
generator to an earlier point in the processing stack.

Previously the timestamp was created _after_ the message was read from
the journal.
When testing situations where we want to make sure we don't lose any
messages, it is desirable to have messages as a continous flow which is easliy visible
in the histogram.

This change moves the timestamp into the GeneratorState which is
created _before_ the message is written into the disk journal.

This way we don't get a spike of messages with the same timestamp, if
they are read from a bigger disk journal backlog.

Before:
image

After:
image


kmerz (approved):
Code LGTM :+1: 
Worked on my machine!

kmerz (approved):

Issue #11043, opened 2021-07-20
Extend Value Actions for external Actions


kmerz:
## What?
Use the external Actions from the plugin store to extend the existing value actions.

## Why?
The user wants to be able to use external actions from the plugin store.

PR #11061, opened 2021-07-23
Implementing external value actions for field values.

linuspahl:
## Description

With this PR it will be possible to extend the list of field value actions. We now have two types of value actions, our internal common ones which depend solely on the field value and new external value actions which will be displayed depending on e.g. the related field.

Before this change we were only storing field and value action in the plugin store. We are now using a context (`FieldAndValueActionsContext`) to provide the internal and external value and field actions.
In a previous version I created a context (`ExternalValueAstionsContext`) only for external value actions, but I in the end preferred to create one which provides all field and field value actions.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2512

Fixes: #11043

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/contexts/FieldAndValueActionsContext.tsx
```diff
@@ -0,0 +1,34 @@
...
+import type { ActionDefinition } from 'views/components/actions/ActionHandler';
+import { singleton } from 'views/logic/singleton';
+
+export type FieldAndValueActionsContextType = {
+  fieldActions: {
+    internal: Array<ActionDefinition> | undefined,
+  },
+  valueActions: {
+    external: Array<ActionDefinition> | undefined,
+    internal: Array<ActionDefinition> | undefined,
```
linuspahl, line 28: We should discuss if `internal` and `external` is the best wording. While the new "external"  value actions will be displayed depending on the related field, currently only their placement in the value actions dropdown differs.

We could also name these categories `common` and `extended` or `common` and `fieldSpecific`.

graylog2-web-interface/src/views/components/actions/ActionDropdown.tsx
```diff
@@ -0,0 +1,129 @@
...
+                        handlerArgs={handlerArgs}
+                        setOverflowingComponents={setOverflowingComponents}
+                        overflowingComponents={overflowingComponents}
+                        type={type}
+                        onMenuToggle={onMenuToggle} />
+      ))}
+
+      {(externalActions && externalActions.length !== 0) && (
+        <>
+          <MenuItem divider />
```
linuspahl, line 113: Please note, I remove the header for the second dropdown section, because neither "External Actions" nor "Event Actions" really described what these actions are doing. I kept the divider to still separate these external actions from the internal actions.

kmerz: External Actions are IMHO the correct term. The actions are executed on non views/search/dashboard features and are therefore external to that system. In all but one case for now they are even external to graylog.

But we can bring that case to product management or our UX engineer.

graylog2-web-interface/src/views/components/contexts/FieldAndValueActionsContext.tsx
kmerz: Instead of external we could also go for extended since we extend the actions via plugins.

kmerz (approved):
LGTM :+1: 

Issue #938, opened 2015-02-03
Journal uncommitted entries is not a metric


kroepke:
We report the number of uncommitted journal entries via REST but don't have a metric for it.

kroepke:
This is an important metric to see when debugging or tuning, so it should be readily available as a gauge metric even in 1.0.

PR #11070, opened 2021-07-27
Implement feature flag

ousmaneo:
These changes add components and hooks that allow checking for feature flags.
- get feature flag from AppConfig
- add Feature component and useFeature hook
- this change replace the environment variable based feature flag (#6367)

## Motivation and Context

This is part of the Feature flag V1 implementation. It uses `featureFlags` list set in the config.js file by the backend as the source of flags. Therefore this replaces the environment variable version used in #6367. 
ref. #938
## Types of changes

## Checklist:

close Graylog2/graylog-plugin-cloud#938


graylog2-web-interface/src/components/features/Feature.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import useFeature from 'hooks/useFeature';
+
+type Props = {
```
AntonEbel, line 22: I found this [article](https://blog.logrocket.com/comparing-typescript-and-proptypes-in-react-applications/) yesterday. It is possible to infer props and remove the duplicate type definition but we have to install @types/prop-types as a dev dependency to use it. It is only a suggestion.

AntonEbel (approved):
As I'm not so familiar with the code base maybe someone else should also review this PR?

edmundoa (changes requested):
Added some comments while looking at the code. I'm having some troubles with my local environment so I could not yet test the code.

graylog2-web-interface/src/components/features/Feature.tsx
AntonEbel: Yes, I totally agree. It's a broader topic.

graylog2-web-interface/config.js
```diff
@@ -23,4 +23,7 @@ window.appConfig = {
     'org.graylog.plugins.customization.notifications': {},
   },
   isCloud: false,
+  featureFlags: {
+    'legacy-aggregation-wizard': 'on',
```
AntonEbel, line 27: We should also add the legacy-aggregation-wizard feature flag to the feature-flag.config file and document it.

ousmaneo: Thanks, @AntonEbel I will do that. I will also remove it from the config file this was a test.

graylog2-web-interface/src/components/features/Feature.tsx
```diff
@@ -0,0 +1,48 @@
...
+  name: string;
+  fallback?: React.ReactElement,
+  children: React.ReactElement,
+}
+
+const Feature = ({ name, fallback, children }: Props) => {
```
edmundoa, line 28: Could we rename this component to something more meaningful? Since we already have the `IfPermitted` component, I would suggest `IfFeatureEnabled`.

edmundoa (commented):
Only a suggestion, other than that looks fine to me.

graylog2-web-interface/src/components/features/Feature.tsx
dennisoelkers: The issue is that prop types are less expressive than TypeScript. IMO we would want it the other way around: Generating prop types from TypeScript typings. There is https://github.com/milesj/babel-plugin-typescript-to-proptypes, but it does not look too good (no release for a year, some rather unfortunate issues we would run into like https://github.com/milesj/babel-plugin-typescript-to-proptypes/issues/45), so this is still an unsolved question.

Our general idea is this:

  - Always use TypeScript types (when there is no real good argument against that)
  - Use prop types for components which are reused across different contexts

This should strike a good compromise in most cases.

edmundoa (approved):
Looks good to me, thank you! 👍

PR #10277, opened 2021-03-17
Remove aws and collector plugin references from integration test.

juju2112:
This keeps CI jobs from failing due to references to the aws and collector plugins in the integration tests.

## Description
We removed the aws and collector plugins from the cloud-4.0 manifest (Graylog2/graylog-project-internal#58), but `NodeContainerFactory` still attempts to use them. Since they are not there, it creates a directory on the build server.

```
jenkins@jenkins-worker-01:~/workspace/hots_graylog-plugin-cloud_PR-745$ ll .repos
total 16
drwxrwxr-x 4 jenkins jenkins 4096 Mar 17 11:45 ./
drwxrwxr-x 3 jenkins jenkins 4096 Mar 17 11:45 ../
drwxr-xr-x 3 root    root    4096 Mar 16 17:14 graylog-plugin-aws/
drwxr-xr-x 3 root    root    4096 Mar 16 17:14 graylog-plugin-collector/

jenkins@jenkins-worker-01:~/workspace/hots_graylog-plugin-cloud_PR-745$ ll .repos/graylog-plugin-aws/target
total 12
drwxr-xr-x 3 root root 4096 Mar 16 17:14 ./
drwxr-xr-x 3 root root 4096 Mar 16 17:14 ../
drwxr-xr-x 2 root root 4096 Mar 16 17:14 graylog-plugin-aws-4.0.6-SNAPSHOT.jar/

jenkins@jenkins-worker-01:~/workspace/hots_graylog-plugin-cloud_PR-745$ ll .repos/graylog-plugin-collector/target
total 12
drwxr-xr-x 3 root root 4096 Mar 16 17:14 ./
drwxr-xr-x 3 root root 4096 Mar 16 17:14 ../
drwxr-xr-x 2 root root 4096 Mar 16 17:14 graylog-plugin-collector-4.0.6-SNAPSHOT.jar/
```

Since the integration test runs in docker as root, and because the jar isn't there, it is creating a directory named after the filename, and it's owned as root. Because of this, Jenkins is unable to clean the workspace (it doesn't run as root), and subsequent CI runs fail.

This PR removes the references to the plugins we don't use in cloud anymore.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
[...]

PR #11073, opened 2021-07-27
Reimplement #10277: some JARs may not exist in a certain manifest

patrickmann:
Some JARs may not exist in a certain environment. Trying to mount these causes problems for the Jenkins Workspace. 
Refer to https://github.com/Graylog2/graylog2-server/pull/10277 for the details and how we initially solved this on the Cloud branch. 

Now that Cloud and Master branches have merged, we achieve the same result by simply ignoring any non-existant JARs (rather than having a Cloud-specific configuration).


bernd (approved):
Thank you! :+1:

PR #11080, opened 2021-07-28
Add service for user session termination

bernd:
This service checks on startup if a global session termination should
run. By bumping the `TERMINATION_REVISION` in the code we can trigger a
global session termination after a user has updated to the new release.

This replaces the `UserSessionTerminationListener`.

This needs to be backported to 4.1, 4.0 and 3.3.


mpfz0r (approved):
LGTM

graylog2-server/src/main/java/org/graylog2/security/UserSessionTerminationService.java
```diff
@@ -0,0 +1,219 @@
...
+                GlobalTerminationRevisionConfig.class,
+                GlobalTerminationRevisionConfig.initial()
+        );
+
+        if (!globalTerminationRevisionConfig.isOutdated()) {
+            LOG.debug("Global session termination not required");
+            return;
+        }
+
+        final AtomicLong terminatedSessions = new AtomicLong(0);
```
thll, line 145: Out of curiosity: what's the reason to use an `AtomicLong` instead of a primitive type here?

bernd: @thll Haha, lol. This is a leftover from a previous version where I incremented the long inside a lambda. I will update it to use a primitive. Good catch, thanks!

thll (approved):
LGTM and works!

PR #11081, opened 2021-07-28
Don't log session ID in DEBUG messages

bernd:


thll (approved):
LGTM

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -188,7 +188,7 @@ public SessionValidationResponse validateSession(@Context ContainerRequestContex
                 session.setAttribute(HTTPHeaderAuthenticationRealm.SESSION_AUTH_HEADER, usernameHeader.get());
             }
 
-            LOG.debug("Session created {}", session.getId());
+            LOG.debug("Session created");
```
mpfz0r, line 191: how about logging the user here instead?

bernd: Done, thanks!

mpfz0r (approved):

PR #11088, opened 2021-07-29
Add is_editable field to StreamDTO

mpfz0r:
This info is enriched when the Autovalue object is constructed


diesaulov (approved):

kyleknighted (approved):
works great

PR #11092, opened 2021-07-30
Add "match pass" guard for pipeline stages

bernd:
This allows users to define a stage with the following to make sure the
next stage is executed even when no rule in the declared stage matches.

```
pipeline "title"
stage 0 match pass
  rule "test-1"
stage 1 match all
  rule "test-2"
end
```

If the "test-1" rule in stage 0 doesn't match, the pipeline interpreter
will still execute  stage 1.


kroepke (approved):
lgtm

PR #11106, opened 2021-08-03
Exclude non-data nodes from ES floodgate checks

### LINKED ISSUE #11074: ES floodgate check needs to take non-data nodes into account
## Description

The `IndexerClusterCheckerThread` service checks if all Elasticsearch nodes have enough available disk space based on the low/high/floodgate settings in Elasticsearch.

The current implementation doesn't take ES master-only nodes into account. A cluster of 10 ES nodes might have 7 data-only nodes and 3 coordinator nodes. Usually the non-data nodes only have a fraction of the disk space that the data nodes have.

Since the current disk space check doesn't take the node status into account, Graylog might trigger a low/high/floodgate notification. This is most likely to happen when using absolute values for the limits instead of percentages.

So we need a way to handle the disk space check in ES clusters where not all nodes have the same roles. 

HS-492248755
HS-497898143

## Your Environment

* Graylog Version: 4.1.1

thll:
## Description

Fetches an ES node's roles along with its stats and uses the roles to filter nodes when issuing low diskspace warnings.

## Motivation and Context

Fixes #11074 

## How Has This Been Tested?

* Used https://www.elastic.co/guide/en/elasticsearch/reference/7.10/docker.html#docker-compose-file to start a multi-node cluster.
* Modified the docker-compose file to alter node roles
* Changed cluster settings during runtime with
```
$ curl -X PUT "localhost:9200/_cluster/settings?pretty" -H 'Content-Type: application/json' -d'
{
  "transient" : {
    "cluster.routing.allocation.disk.watermark.low" : "1%"
  }
}
'
```
* Observed watermark notifications in Graylog for the nodes holding data.

## Types of changes

## Checklist:

- [z] My code follows the code style of this project.

### DIFF
--- a/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/ClusterAdapterES6.java
+++ b/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/ClusterAdapterES6.java
@@ -102,13 +102,14 @@ public class ClusterAdapterES6 implements ClusterAdapter {
 
     @Override
     public Set<NodeDiskUsageStats> diskUsageStats() {
-        final JsonNode nodes = catNodes("name", "host", "ip", "diskUsed", "diskTotal","diskUsedPercent");
+        final JsonNode nodes = catNodes("name", "role", "host", "ip", "diskUsed", "diskTotal", "diskUsedPercent");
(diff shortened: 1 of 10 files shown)


bernd (approved):
LGTM :+1: Thank you! :pray:

Issue #10385, opened 2021-04-07
Orphaned permissions grants


ChristopherKB:
Currently, when a user is deleted from Graylog, the individual permission grants are not deleted.

A cleanup task would remove grants left over from deleted users.
## Expected Behavior

Graylog should remove grants left over from deleted users.
## Current Behavior

Graylog does not remove grants left over from deleted users.

## Possible Solution

A cleanup task that would identify and delete objects with permissions to no current users.
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

Impact is small, but over time in an environment with high user turnover, it could begin to cause performance or other "bloating" issues.
## Your Environment

* Graylog Version: ALL
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

PR #11121, opened 2021-08-05
Clear grants when user is deleted

thll:
## Description

Upon deletion of a user, clears all grants for users which are not present in the DB anymore.

We scanning the complete grants table and remove all grants that are orphaned. This way we keep a clean grants table even if previously some grants have been missed before. This saves us from writing an additional migration.

I considered using raw mongodb queries and some batching to reduce the number of queries but I think we are fine using the higher level APIs to query for users and delete the grants.

I wouldn't expect any systems with user/grant numbers at such a high level that this could become a problem.

## Motivation and Context

#10385 

## How Has This Been Tested?

Smoke test by creating grants for users/teams and checking which grants get removed when a user is deleted.

## Types of changes

## Checklist:


mpfz0r (approved):
LGTM

PR #11129, opened 2021-08-09
Remove node graceful shutdown menu entry in UI

### LINKED ISSUE #10668: Remove `Graceful shutdown` from `Nodes` options in Graylog

edmundoa:
This PR removes the menu entry in the UI to shutdown a Graylog node, taking care of the UI part of #10668.

Fixes #10668


ousmaneo (approved):
Tested and works! 👍

PR #11130, opened 2021-08-09
Display Tooltips by default

### LINKED ISSUE #10975: Tooltips are not shown
Tooltips in most parts of the web interface seems to not work, and are hidden even if interacting with their trigger elements.

Since https://github.com/Graylog2/graylog2-server/blob/4bc12e0e60437eba1783c54ed6657080b68f8e37/graylog2-web-interface/src/components/graylog/Tooltip.jsx, the `Tooltip` component has a `show` property that is `false` by default. It looks like this broke other `Tooltips` that were relying on them being visible by default.

## Expected Behavior
Interacting with a trigger element should display a tooltip.

## Current Behavior
Interacting with a trigger element does not always display a tooltip.

1. Go to Forwarder list page
2. Hover over Forwarder state
3. That should display a tooltip, but none are displayed

## Your Environment
* Graylog Version: Graylog 4.2.0-SNAPSHOT (96564c1f81ccb5b7e25404c5b0f57128b66b3eac), Graylog 4.1.0

edmundoa:
The `Tooltip` component offers a `show` prop that will decide if the tooltip is rendered or not, with a default value of `false`. However, when introducing that prop and its default value we did not update all usages of the component, which use `OverlayTrigger` to decide when to display or hide the `Tooltip`, and therefore they expect the `Tooltip` to be visible as default when rendered.

This PR changes the default value of `show` to `true`, ensurig `Tooltip`s are shown by default but still having the possibility of hiding a `Tooltip` with a prop if needed.

Fixes #10975

### DIFF
--- a/graylog2-web-interface/src/components/events/events/Events.jsx
+++ b/graylog2-web-interface/src/components/events/events/Events.jsx
@@ -143,7 +143,7 @@ class Events extends React.Component {
 
     return (
       <>
-        <OverlayTrigger placement="top" overlay={tooltip}>
+        <OverlayTrigger placement="top" trigger={['hover', 'click', 'focus']} overlay={tooltip}>
           <EventsIcon name={icon} fixedWidth className={style} />
         </OverlayTrigger>
       </>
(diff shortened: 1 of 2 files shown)


ousmaneo (approved):
Tested and works 👍

PR #11142, opened 2021-08-12
Fixing resolving alias target for index sets with plus in prefix on ES6.

dennisoelkers:
## Motivation and Context

**Note:** This needs a backport to `4.0` and `4.1`.

Prior to this PR, we fixed a couple of issues related to index sets with a plus in their names (or any other character that requires special URL encoding) in #10392. One additional source of error was not found, which prevents resolving the deflector alias to the current index name on ES6.

This error leads to index cycling not working, due to receiving `null` for the current deflector target and passing that as the old index to the cycle alias method, which constructs an ES-command which atomically removes the old alias mapping and creates the new one. Due to the alias target which needs to be removed, the overall command fails, resulting in the complete index cycling to fail.

This change is now applying proper URL encoding in the `resolveAlias` method, resulting in the current target being properly resolved and index cycling to succeed.

## Types of changes

## Checklist:


janheise (approved):
Looks good, works good!

PR #11146, opened 2021-08-12
Add metadata field and accessors on Graylog Message object

danotorrey:
## Description

Adds a new `metadata` field on the Graylog `Message` object to allow data to be stored in the message, that will accompany the message throughout the processing lifecycle. 

## Motivation and Context

In certain feature cases, it can be useful to store state within the message object in Graylog.  Previously, this was only possible to add metadata to messages within explicit fields on the message. This previous approach was less desirable, since fields are intended to be eventually stored in search indices, so careful handling was needed to ensure the field metadata data is not indexed.

With this new approach, the metadata field (which is `null` by default) can be set, retrieved, and cleared through accessor methods (`message.setMetadata()`, `message.getMetadataValue()`, `message.removeMetadata()`)

See https://github.com/Graylog2/graylog-plugin-enterprise/issues/2590 for additional details on the initial motivation for the change.

Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/2590

## How Has This Been Tested?

Tested to verify that Graylog pipelines still process messages successfully, and also verified that messages are indexed properly.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -29,6 +29,7 @@ import com.google.common.collect.Iterators;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.net.InetAddresses;
+import org.apache.commons.lang3.StringUtils;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.messages.Indexable;
 import org.graylog2.plugin.streams.Stream;
@@ -52,6 +53,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


danotorrey:
@bernd I know tomorrow is your last day before vacation, but I added this tonight, just in case you have a moment to peruse to see if this this is what you had in mind for storing Illuminate state. If not, no stress. I'll review with the others. :)

graylog2-server/src/test/java/org/graylog2/plugin/MessageTest.java
```diff
@@ -638,4 +639,18 @@ public void assignUnsupportedTemporalTypeAsTimestamp() {
         message.addField(Message.FIELD_TIMESTAMP, ThaiBuddhistDate.of(0, 4, 19));
         assertThat(message.getTimestamp()).isGreaterThan(new DateTime(2018, 4, 19, 0, 0, 0, 0, DateTimeZone.UTC));
     }
+
+    @Test
+    public void testSetMessageState() throws NoSuchFieldException, IllegalAccessException {
+        final Message message = new Message("message", "source", Tools.nowUTC());
+        final Field stateMap = Message.class.getDeclaredField("metadata");
+        assertNull(stateMap.get(message));
```
bernd, line 647: Please stick to the assertj assertion in this class for consistency. Thanks!

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -953,4 +962,40 @@ public String apply(final Message input) {
...
+     *
+     * @param key   A globally unique string key identifier for the metadata value.
+     * @param value The metadata object value.
+     */
+    public void setMetadata(String key, Object value) {
+        Preconditions.checkArgument(StringUtils.isNotEmpty(key), "A non-empty key is required.");
+        Preconditions.checkNotNull(value);
+        if (metadata == null) {
+            metadata = new HashMap<>();
+        }
```
bernd, line 977: We only initialize the metadata map here but don't store the key and value.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -953,4 +962,40 @@ public String apply(final Message input) {
...
+    }
+
+    /**
+     * Get the metadata value for the specified key.
+     *
+     * @param key The string key for the metadata entry.
+     */
+    public Object getMetadataValue(String key) {
+        return metadata.get(key);
+    }
```
bernd, line 987: How about annotating this with `@Nullable`? This will help during development.

What do you think about adding a `getMetadataValue(String key, Object defaultValue)` method. Might be useful to avoid null checks in consumers of the class.

bernd: We also need a check if the metadata is `null` to avoid a NullPointerException when trying to get a value from an uninitialized map.

graylog2-server/src/test/java/org/graylog2/plugin/MessageTest.java
```diff
@@ -638,4 +639,18 @@ public void assignUnsupportedTemporalTypeAsTimestamp() {
         message.addField(Message.FIELD_TIMESTAMP, ThaiBuddhistDate.of(0, 4, 19));
         assertThat(message.getTimestamp()).isGreaterThan(new DateTime(2018, 4, 19, 0, 0, 0, 0, DateTimeZone.UTC));
     }
+
+    @Test
+    public void testSetMessageState() throws NoSuchFieldException, IllegalAccessException {
+        final Message message = new Message("message", "source", Tools.nowUTC());
+        final Field stateMap = Message.class.getDeclaredField("metadata");
```
bernd, line 646: This breaks in Jenkins and my local environment. I think it's okay to not test the internals here.

bernd (changes requested):
Yep, that's what I had it mind! :+1: A few comments:

danotorrey:
Thanks a lot for the feedback @bernd! I'll address it.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
danotorrey: This all sounds good to me.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -953,4 +962,59 @@ public String apply(final Message input) {
...
+    /**
+     * Get the metadata value for the specified key. If not present, then return the default value.
+     *
+     * @param key The string key for the metadata entry.
+     */
+    @Nullable
+    public Object getMetadataValue(String key, Object defaultValue) {
+        if (metadata == null) {
+            return null;
+        }
```
waab76, line 1003: So if no metadata has been set, then an attempt to get-with-default returns `null` instead of the default?  Is there a reason we wouldn't want to return the provided default value?

danotorrey: Thanks @waab76 Great point and find. This is a bug. I'll fix it and add test coverage.

danotorrey: Thanks for the feedback. Fixed.

waab76 (approved):
LGTM!

mpfz0r (approved):
fine with me as well

danotorrey:
A copy of the PR was created (https://github.com/Graylog2/graylog2-server/pull/11190) to override the "changes requested" flag by Bernd, since he is on vacation.

PR #11149, opened 2021-08-13
Record processing failures from extractors and timestamp conversion

mpfz0r:
- Record ExtractorExceptions at various places

-  Unify Message timestamp conversion and record errors
   There used to be two slightly different timestamp conversion methods.
   Unify both and try to support as many timestmap types as possible.

   If the timestamp cannot be converted, or is null,
   fall back to the current time and record the error using
   a newly introduced FailureCause: InvalidTimestampException

- Remove broken timestamp workaround from clone_message() pipeline function

- Improve exception error log by using `ExceptionUtils#getShortenedStackTrace()` on high volume per message logs.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2598,https://github.com/Graylog2/graylog-plugin-integrations/pull/925


graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
```diff
@@ -106,8 +110,15 @@ public Messages process(Messages messages) {
...
-                    LOG.error("Could not apply filter [" + filter.getName() + "] on message <" + msg.getId() + ">: ",
-                            e);
+                    final String shortError = String.format(Locale.US, "Could not apply filter [%s] on message <%s>",
+                            filter.getName(), msg.getId());
+                    if (LOG.isDebugEnabled()) {
+                        LOG.error("{}:", shortError, e);
+                    } else {
+                        LOG.error("{}:\n{}", shortError, ExceptionUtils.getShortenedStackTrace(e));
+                    }
+                    msg.addProcessingError(new Message.ProcessingError(ProcessingFailureCause.ExtractorException,
```
diesaulov: `ExtractorException` is not applicable here.

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -281,7 +287,14 @@ private void runConverters(Message msg) {
                     }
                 } catch (Exception e) {
                     this.converterExceptions.incrementAndGet();
-                    LOG.error("Could not apply converter [" + converter.getType() + "] of extractor [" + getId() + "].", e);
+                    final String error = "Could not apply converter [" + converter.getType() + "] of extractor <" + getTitle() + " (" + getId() + ")>";
+                    if (LOG.isDebugEnabled()) {
+                        LOG.error(error, e);
+                    } else {
+                        LOG.error("{}:\n{}", error, ExceptionUtils.getShortenedStackTrace(e));
```
diesaulov, line 294: The logging framework must support a special formatting symbol for line breaks making it OS-independent

graylog2-server/src/test/java/org/graylog2/plugin/MessageTest.java
```diff
@@ -669,4 +673,65 @@ public void testMetadataDefault() throws NoSuchFieldException, IllegalAccessExce
...
+        DateTimeUtils.setCurrentMillisSystem();
+
+        final Message message = new Message("message", "source", Tools.nowUTC().minusMinutes(2));
+        final LocalDateTime localDate = LocalDateTime.of(2021, Month.AUGUST, 19, 12, 0);
+        final ZonedDateTime zonedDateTime = ZonedDateTime.of(localDate, ZoneOffset.UTC);
+
+        message.addField(Message.FIELD_TIMESTAMP, zonedDateTime);
+        assertThat(message.getTimestamp()).isInstanceOf(DateTime.class);
+        final DateTimeFormatter formatter = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm").withZone(DateTimeZone.UTC);
+        final DateTime expectedLocalDateEquivalent = formatter.parseDateTime("2021-08-19 12:00");
```
diesaulov, line 731: Why the intermediate step with string parsing is need here?

graylog2-server/src/test/java/org/graylog2/shared/utilities/ExceptionUtilsTest.java
```diff
@@ -71,4 +71,33 @@ public void hasCauseOf_returnsTrueIfTheCauseIsSubtypeOfTheProvidedType() {
         assertThat(ExceptionUtils.hasCauseOf(
                 new RuntimeException("parent", new IOException("asdasd")), IOException.class)).isTrue();
     }
+
+    @Test
+    public void getShortenedStackTrace() {
```
diesaulov, line 76: I believe some other test cases might be out there :D

graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionService.java
```diff
@@ -112,15 +112,16 @@ private void submitProcessingFailure(Message failedMessage, Message.ProcessingEr
             final boolean needsAcknowledgement = !failureHandlingConfiguration.keepFailedMessageDuplicate();
 
             final String message;
+            final String messageId = StringUtils.isBlank(failedMessage.getMessageId()) ? failedMessage.getId() : failedMessage.getMessageId();
```
diesaulov, line 115: Please reflect this change in the unit tests

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
```diff
@@ -77,8 +79,14 @@ public boolean filter(Message msg) {
                 extractor.runExtractor(msg);
             } catch (Exception e) {
                 extractor.incrementExceptions();
-                LOG.error("Could not apply extractor \"" + extractor.getTitle() + "\" (id=" + extractor.getId() + ") "
-                        + "to message " + msg.getId(), e);
+                final String error = "Could not apply extractor <" + extractor.getTitle() + "(" + extractor.getId() + ")>";
```
diesaulov, line 82: Please prove the code with unit tests

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
```diff
@@ -106,8 +110,15 @@ public Messages process(Messages messages) {
...
-                    LOG.error("Could not apply filter [" + filter.getName() + "] on message <" + msg.getId() + ">: ",
-                            e);
+                    final String shortError = String.format(Locale.US, "Could not apply filter [%s] on message <%s>",
+                            filter.getName(), msg.getId());
+                    if (LOG.isDebugEnabled()) {
+                        LOG.error("{}:", shortError, e);
+                    } else {
+                        LOG.error("{}:\n{}", shortError, ExceptionUtils.getShortenedStackTrace(e));
+                    }
+                    msg.addProcessingError(new Message.ProcessingError(ProcessingFailureCause.MessageFilterException,
```
diesaulov, line 120: Please prove the code with unit tests

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -212,7 +214,11 @@ public void runExtractor(Message msg) {
 
             try (final Timer.Context ignored2 = executionTimer.time()) {
                 final Result[] results = run(field);
-                if (results == null || results.length == 0 || Arrays.stream(results).anyMatch(result -> result.getValue() == null)) {
+                if (results != null && results.length == 1 && results[0].exception != null) {
+                    final String error = "Could not apply extractor <" + getTitle() + "(" + getId() + ")>";
+                    msg.addProcessingError(new Message.ProcessingError(
```
diesaulov, line 219: Please prove the code with unit tests

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
```diff
@@ -281,7 +287,14 @@ private void runConverters(Message msg) {
...
                 } catch (Exception e) {
                     this.converterExceptions.incrementAndGet();
-                    LOG.error("Could not apply converter [" + converter.getType() + "] of extractor [" + getId() + "].", e);
+                    final String error = "Could not apply converter [" + converter.getType() + "] of extractor <" + getTitle() + " (" + getId() + ")>";
+                    if (LOG.isDebugEnabled()) {
+                        LOG.error(error, e);
+                    } else {
+                        LOG.error("{}:\n{}", error, ExceptionUtils.getShortenedStackTrace(e));
+                    }
+                    msg.addProcessingError(new Message.ProcessingError(ProcessingFailureCause.ExtractorException,
```
diesaulov, line 296: Please prove the code with unit tests

diesaulov (changes requested):
Even though some points are still to be addressed, the refactoring is highly appreciated 👍

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -122,11 +123,6 @@
      */
     public static final String FIELD_GL2_MESSAGE_ID = "gl2_message_id";
 
-    /**
-     * Can be set when a message timestamp gets modified to preserve the original timestamp. (e.g. "clone_message" pipeline function)
-     */
-    public static final String FIELD_GL2_ORIGINAL_TIMESTAMP = "gl2_original_timestamp";
```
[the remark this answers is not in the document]
mpfz0r: I'm pretty sure that the `clone_message()` pipeline function was the only user of that obscure fix/feature.
Theoretically some unofficial plugin might use that constant, but I highly doubt it.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
         obj.put(FIELD_GL2_ACCOUNTED_MESSAGE_SIZE, getSize());
 
         if (!processingErrors.isEmpty()) {
+            if (processingErrors.stream().anyMatch(processingError -> processingError.getCause().equals(ProcessingFailureCause.InvalidTimestampException))) {
```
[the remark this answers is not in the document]
mpfz0r: oh, indeed. that makes more sense :+1:

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
...
+                invalidTimestampMeter.mark();
+            }
             obj.put(FIELD_GL2_PROCESSING_ERROR,
                     processingErrors.stream()
                             .map(ProcessingError::getDetails)
                             .collect(Collectors.joining(", ")));
         }
 
         final Object timestampValue = getField(FIELD_TIMESTAMP);
+        DateTime dateTime = convertToDateTime(timestampValue);
```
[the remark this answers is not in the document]
mpfz0r: A Message object can be constructed without a timestamp. So if no timestamp was ever added during the processing,
we do it at this point. I don't know if that happens in practice, but changing that behavior might be risky.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
...
             // if the timestamp value is a string, we try to parse it in the correct format.
             // we fall back to "now", this avoids losing messages which happen to have the wrong timestamp format
             try {
-                dateTime = ES_DATE_FORMAT_FORMATTER.parseDateTime((String) timestampValue);
+                dateTime = ES_DATE_FORMAT_FORMATTER.parseDateTime((String) value);
             } catch (IllegalArgumentException e) {
-                LOG.trace("Invalid format for field timestamp '{}' in message {}, forcing to current time.", timestampValue, getId());
-                invalidTimestampMeter.mark();
+                final String error = "Invalid format for field timestamp '" + value + "' in message " + getId() + ", forcing to current time.";
+                LOG.trace(error, e);
```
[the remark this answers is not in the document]
mpfz0r: Not sure, but I think it's fine for the trace loglevel

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
...
         final Object timestampValue = getField(FIELD_TIMESTAMP);
+        DateTime dateTime = convertToDateTime(timestampValue);
+        if (dateTime != null) {
+            obj.put(FIELD_TIMESTAMP, buildElasticSearchTimeFormat(dateTime.withZone(UTC)));
+        }
+
+        return obj;
+    }
+
+    private DateTime convertToDateTime(Object value) {
```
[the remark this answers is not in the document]
mpfz0r: Agreed, but none of this is new code, and those DateTime conversion stuff is a pain to test.
So I don't it's worth to spend probably two more hours on writing a test for that.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
...
-                invalidTimestampMeter.mark();
+                final String error = "Invalid format for field timestamp '" + value + "' in message " + getId() + ", forcing to current time.";
+                LOG.trace(error, e);
+                addProcessingError(new ProcessingError(ProcessingFailureCause.InvalidTimestampException, error, ExceptionUtils.getRootCauseMessage(e)));
                 dateTime = Tools.nowUTC();
             }
+        } else if (value == null) {
+            final String error = "null value for field timestamp in message " + getId() + ", forcing to current time.";
+            LOG.trace(error);
+            addProcessingError(new ProcessingError(ProcessingFailureCause.InvalidTimestampException, error, ""));
```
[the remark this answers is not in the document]
mpfz0r: I don't have anymore details :shrug:

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -433,39 +428,73 @@ public DateTime getTimestamp() {
...
+            dateTime = Tools.nowUTC();
         } else {
             // don't allow any other types for timestamp, force to "now"
-            LOG.trace("Invalid type for field timestamp '{}' in message {}, forcing to current time.", timestampValue.getClass().getSimpleName(), getId());
-            invalidTimestampMeter.mark();
+            final String error = "Invalid type for field timestamp '" + value.getClass().getSimpleName() + "' in message " + getId() + ", forcing to current time.";
+            LOG.trace(error);
+            addProcessingError(new ProcessingError(ProcessingFailureCause.InvalidTimestampException, error, ""));
```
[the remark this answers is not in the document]
mpfz0r: same as above. all the details are in the error message already

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
mpfz0r: Not that I know of.
In a previous version I had
```suggestion
                        LOG.error("{}:{}{}", error, System.lineSeparator(), ExceptionUtils.getShortenedStackTrace(e));
```
But it felt over-engineered because we only run on unix-ish platforms anyway

graylog2-server/src/test/java/org/graylog2/plugin/MessageTest.java
mpfz0r: Seemed the easiest way to construct an equivalent DateTime to me

graylog2-server/src/test/java/org/graylog2/shared/utilities/ExceptionUtilsTest.java
mpfz0r: care to elaborate?

graylog2-server/src/main/java/org/graylog2/filters/ExtractorFilter.java
mpfz0r: Done. Don't you love it if a test is 4 times as much complicated as the actual code? :see_no_evil:

graylog2-server/src/main/java/org/graylog2/messageprocessors/MessageFilterChainProcessor.java
mpfz0r: done

graylog2-server/src/main/java/org/graylog2/plugin/inputs/Extractor.java
mpfz0r: done

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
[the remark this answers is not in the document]
diesaulov: The constructor uses `addRequiredField`, which in its turn transitively calls  `convertToDateTime` in `addField`

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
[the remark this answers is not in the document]
mpfz0r: done

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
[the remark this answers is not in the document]
mpfz0r: I changed it to logging a short exception only

graylog2-server/src/test/java/org/graylog2/plugin/MessageTest.java
mpfz0r: changed that

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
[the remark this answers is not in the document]
mpfz0r: addressed via zoom call

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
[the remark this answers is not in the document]
mpfz0r: addressed via zoom call

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
```diff
@@ -432,40 +421,45 @@ public DateTime getTimestamp() {
         obj.put(FIELD_STREAMS, getStreamIds());
         obj.put(FIELD_GL2_ACCOUNTED_MESSAGE_SIZE, getSize());
 
+        final Object timestampValue = getField(FIELD_TIMESTAMP);
+        DateTime dateTime = convertToDateTime(timestampValue);
+        if (dateTime != null) {
+            obj.put(FIELD_TIMESTAMP, buildElasticSearchTimeFormat(dateTime.withZone(UTC)));
+        }
```
thll, line 428: I find it a bit confusing that we are "converting" the timestamp again. If I'm not missing anything, it's either already in the right format at this point, or it's `null`. So I'd prefer to explicitly handle the `null` case here. I think it would make it more obvious what's going on.

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/DateTimeConverter.java
```diff
@@ -0,0 +1,67 @@
...
+    public static DateTime convertToDateTime(@Nonnull Object value) {
+        if (value instanceof DateTime) {
+            return (DateTime) value;
+        }
+
+        if (value instanceof Date) {
+            return new DateTime(value, DateTimeZone.UTC);
+        } else if (value instanceof ZonedDateTime) {
+            final DateTimeZone dateTimeZone = DateTimeZone.forTimeZone(TimeZone.getTimeZone(((ZonedDateTime) value).getZone()));
+            return new DateTime(Date.from(((ZonedDateTime) value).toInstant()), dateTimeZone);
```
thll, line 45: Is there a reason why for `ZonedDateTime` we now try to preserve the time offset while for `Date`, `OffsetDateTime` and `Instant` we are now removing that information and reverting to UTC?

graylog2-server/src/test/java/org/graylog2/plugin/utilities/date/DateTimeConverterTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test
+    @SuppressForbidden("Comparing twice with default timezone is okay in tests")
+    void convertFromLocalDateTime() {
+        final LocalDateTime input = LocalDateTime.of(2021, Month.AUGUST, 19, 12, 0);
+
+        final DateTime output = DateTimeConverter.convertToDateTime(input);
+
+        // represented in local timezone, but then converted to UTC
+        //final DateTime expectedOutput = new DateTime(2021, 8, 19, 12, 0).withZone(DateTimeZone.UTC);
```
thll, line 108: Leftover comment but it highlights what we need to decide: should we always return UTC or do we keep the time zone (in this specific case the system default zone I assume)

thll (changes requested):
Thanks for the effort to finally fix this!

I only found a couple of minor things.
I still need to test this, but it's looking good so far.

graylog2-server/src/main/java/org/graylog2/inputs/extractors/JsonExtractor.java
```diff
@@ -94,7 +93,12 @@ public JsonExtractor(final MetricRegistry metricRegistry,
 
     @Override
     protected Result[] run(String value) {
-        final Map<String, Object> extractedJson = extractJson(value);
+        final Map<String, Object> extractedJson;
+        try {
+            extractedJson = extractJson(value);
+        } catch (IOException e) {
+            return new Result[]{new Result(null, null, -1, -1, e)};
```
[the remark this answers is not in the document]
mpfz0r: Great idea. Thanks :+1:

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
mpfz0r: I think I agree :thinking:  But how would you write that without duplicating too much code?

We could also drop that null check, because `convertToDateTime` never returns `null`
```suggestion
        final Object timestampValue = getField(FIELD_TIMESTAMP);
        DateTime dateTime = convertToDateTime(timestampValue);
        obj.put(FIELD_TIMESTAMP, buildElasticSearchTimeFormat(dateTime.withZone(UTC)));
```

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/DateTimeConverter.java
mpfz0r: I think it's best to preserve as much as possible.
The actual conversion to UTC will be performed when it's written to elastic

graylog2-server/src/test/java/org/graylog2/plugin/utilities/date/DateTimeConverterTest.java
mpfz0r: Thanks, I was opting for keeping the timezone. So I'll remove the wrong comment

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/DateTimeConverter.java
thll: I agree that we should preserve as much as possible, but unless I'm wrong, with your change we are now removing offset information where it was previously preserved in the `Date`, `OffsetDateTime` and `Instant` case. So to rephrase: let's be consistent by either preserving as much as possible all the time or converting everything to UTC.

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
thll: 🤔  maybe like so?
```
Index: graylog2-server/src/main/java/org/graylog2/plugin/Message.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java	(revision 473e376b96308f200e97d7fdcc5fca78e7555098)
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java	(date 1631031513849)
@@ -422,10 +422,8 @@
         obj.put(FIELD_GL2_ACCOUNTED_MESSAGE_SIZE, getSize());
 
         final Object timestampValue = getField(FIELD_TIMESTAMP);
-        DateTime dateTime = convertToDateTime(timestampValue);
-        if (dateTime != null) {
-            obj.put(FIELD_TIMESTAMP, buildElasticSearchTimeFormat(dateTime.withZone(UTC)));
-        }
+        DateTime dateTime = timestampValue == null ? fallbackForNullTimestamp() : convertToDateTime(timestampValue);
+        obj.put(FIELD_TIMESTAMP, buildElasticSearchTimeFormat(dateTime.withZone(UTC)));
 
         if (processingErrors != null && !processingErrors.isEmpty()) {
             if (processingErrors.stream().anyMatch(processingError -> processingError.getCause().equals(ProcessingFailureCause.InvalidTimestampException))) {
@@ -441,15 +439,6 @@
     }
 
     private DateTime convertToDateTime(Object value) {
-        if (value == null) {
-            final String error = "<null> value for field timestamp in message <" + getId() + ">, forcing to current time";
-            LOG.trace(error);
-            addProcessingError(new ProcessingError(ProcessingFailureCause.InvalidTimestampException,
-                    "Replaced invalid timestamp value in message <" + getId() + "> with current time",
-                    "<null> value provided"));
-            return Tools.nowUTC();
-        }
-
         try {
             return DateTimeConverter.convertToDateTime(value);
         } catch (IllegalArgumentException e) {
@@ -461,6 +450,15 @@
             return Tools.nowUTC();
         }
     }
+
+    private DateTime fallbackForNullTimestamp() {
+        final String error = "<null> value for field timestamp in message <" + getId() + ">, forcing to current time";
+        LOG.trace(error);
+        addProcessingError(new ProcessingError(ProcessingFailureCause.InvalidTimestampException,
+                "Replaced invalid timestamp value in message <" + getId() + "> with current time",
+                "<null> value provided"));
+        return Tools.nowUTC();
+    }
 
     // estimate the byte/char length for a field and its value
     static long sizeForField(@Nonnull String key, @Nonnull Object value) {
@@ -536,7 +534,7 @@
 
         final boolean isTimestamp = FIELD_TIMESTAMP.equals(trimmedKey);
         if (isTimestamp) {
-            final DateTime timeStamp = convertToDateTime(value);
+            final DateTime timeStamp = value == null ? fallbackForNullTimestamp() : convertToDateTime(value);
             final Object previousValue = fields.put(FIELD_TIMESTAMP, timeStamp);
             updateSize(trimmedKey, timeStamp, previousValue);
         } else if (value instanceof String) {

```

graylog2-server/src/main/java/org/graylog2/rest/resources/tools/JsonTesterResource.java
```diff
@@ -98,7 +99,12 @@ private JsonTesterResponse testJsonExtractor(String testString,
             throw new BadRequestException("Invalid extractor configuration", e);
         }
 
-        final Map<String, Object> result = extractor.extractJson(testString);
+        final Map<String, Object> result;
+        try {
+            result = extractor.extractJson(testString);
+        } catch (IOException e) {
+            throw new BadRequestException("Failure running JSON extractor", e);
```
thll, line 106: If we were unable to extract JSON, the UI was previously showing this:
image

Now we are showing an error overlay: 
image

What do you think about adding a bit more information what actually went wrong. Maybe just by including the actual error, or do you think this is too much information? Thinking about something along the lines of

```suggestion
            throw new BadRequestException("Failure running JSON extractor: " + e.getMessage(), e);
```

graylog2-server/src/main/java/org/graylog2/plugin/Message.java
mpfz0r: Thanks :-) Applied :+1:

graylog2-server/src/main/java/org/graylog2/rest/resources/tools/JsonTesterResource.java
mpfz0r: I like it :+1:

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/DateTimeConverter.java
thll: Ok, so we are only "losing" the system default offset which is not worth keeping. I think the solution is good as it is 👍

thll (approved):
Tested and is looking good! 👍

diesaulov (approved):

PR #11152, opened 2021-08-13
Failure queue metrics

diesaulov:
Migrated the `IndexFailuresPeriodical` size metric. The change is not
backward-compatible (it was decided to rename the metric, since the old
name doesn't represent the actual nature of that metric).

Also 2 more metrics were exposed to monitor performance of
failure handling (`failure_queue_submitted_failures` vs
`failure_queue_consumed_failures`)

[graylog-plugin-enterprise/issue-2127] Storing indexing/processing failures

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
+++ b/graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
@@ -16,6 +16,7 @@
  */
 package org.graylog.failure;
 
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import org.graylog2.Configuration;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog/failure/FailureSubmissionQueue.java
```diff
@@ -59,10 +59,14 @@ public FailureSubmissionQueue(Configuration configuration,
                                   MetricRegistry metricRegistry) {
         this.queue = new LinkedBlockingQueue<>(configuration.getFailureHandlingQueueCapacity());
         this.configuration = configuration;
-        this.submittedFailureBatches = metricRegistry.meter(name(FailureSubmissionQueue.class, "submittedFailureBatches"));
+
+            this.submittedFailureBatches = metricRegistry.meter(name(FailureSubmissionQueue.class, "submittedFailureBatches"));
```
mpfz0r, line 63:
```suggestion
        this.submittedFailureBatches = metricRegistry.meter(name(FailureSubmissionQueue.class, "submittedFailureBatches"));
```

mpfz0r (approved):

Issue #9559, opened 2020-11-20
Unhandled exception in REST resource // Cannot deserialize instance of `java.lang.Long` out of START_OBJECT token


eadanilenko:
Hi Team,

Have no idea how it happens and how to reproduce it, but from time to time I do able to see following exception:

```
2020-11-20T12:02:51.677+02:00 ERROR [AnyExceptionClassMapper] Unhandled exception in REST resource
java.lang.RuntimeException: IOException encountered while reading from a byte array input stream
        at org.mongojack.internal.stream.JacksonDBDecoder.decode(JacksonDBDecoder.java:67) ~[graylog.jar:?]
        at com.mongodb.DBDecoderAdapter.decode(DBDecoderAdapter.java:49) ~[graylog.jar:?]
        at com.mongodb.DBDecoderAdapter.decode(DBDecoderAdapter.java:29) ~[graylog.jar:?]
        at com.mongodb.operation.CommandResultArrayCodec.decode(CommandResultArrayCodec.java:52) ~[graylog.jar:?]
        at com.mongodb.operation.CommandResultDocumentCodec.readValue(CommandResultDocumentCodec.java:60) ~[graylog.jar:?]
        at org.bson.codecs.BsonDocumentCodec.decode(BsonDocumentCodec.java:84) ~[graylog.jar:?]
        at org.bson.codecs.BsonDocumentCodec.decode(BsonDocumentCodec.java:41) ~[graylog.jar:?]
        at org.bson.internal.LazyCodec.decode(LazyCodec.java:48) ~[graylog.jar:?]
        at org.bson.codecs.BsonDocumentCodec.readValue(BsonDocumentCodec.java:101) ~[graylog.jar:?]
        at com.mongodb.operation.CommandResultDocumentCodec.readValue(CommandResultDocumentCodec.java:63) ~[graylog.jar:?]
        at org.bson.codecs.BsonDocumentCodec.decode(BsonDocumentCodec.java:84) ~[graylog.jar:?]
        at org.bson.codecs.BsonDocumentCodec.decode(BsonDocumentCodec.java:41) ~[graylog.jar:?]
        at com.mongodb.internal.connection.ReplyMessage.<init>(ReplyMessage.java:51) ~[graylog.jar:?]
        at com.mongodb.internal.connection.InternalStreamConnection.getCommandResult(InternalStreamConnection.java:413) ~[graylog.jar:?]
[...]

PR #11176, opened 2021-08-17
/system/cluster/traffic: Support dots in nodeId field name

patrickmann:
This addresses #9559: Mongo will save field names with embedded dot characters, but it cannot properly update or retrieve them. As a result, retrieving traffic metrics may lead to an IOException.

Solution: escape the problematic characters "$" and "." when saving the field name. 
At this time we do not ever retrieve and expose the field name; but an unescape method is provided in case we do.

As of version 5.0, Mongo also has [built-in support](https://docs.mongodb.com/manual/core/dot-dollar-considerations/#field-names-with-periods--.--and-dollar-signs----) for this issue; however, MongoJack library only supports versions up to 4.x.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/system/NodeId.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/system/NodeId.java
@@ -93,6 +93,14 @@ public class NodeId {
         return id;
     }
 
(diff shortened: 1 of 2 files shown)


mpfz0r (approved):
Tested and LGTM :+1:

PR #11184, opened 2021-08-18
Fixing ignored / broken Elasticsearch E2E tests

devdanylo:
Apart from not being picked by Maven due to the missing IT suffix there is another problem - the ES E2E tests produce expected results only when run in isolation (i.e. one test per run). They don't work as expected when run along with other tests for one essential reason - GraylogBackend is a singleton. That is 1) additional ports cannot be mapped via extraPorts if there is already a running instance of GraylogBackend 2) various versions of ES cannot be tested if there is already a running instance of GraylogBackend. See https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/test/java/org/graylog/testing/completebackend/GraylogBackend.java for details.

It turned out that @janheise is currently working on E2E test framework refactoring which should address the issues explained above. For that reason we've agreed to keep the tests inactive so far and integrate them with the new framework as soon as it's available in master.


janheise:
@diesaulov my first version of the ContainerMatrix-Tests is now merged, I can give you an introduction if you like?

dennisoelkers:
This PR is obsolete - closing.

PR #11185, opened 2021-08-18
Shorten widget titles with ellipsis when overflowing available space.

dennisoelkers:
## Motivation and Context

Prior to this change, a widget's title was always shown in full, even when there was not enough space, leading to unwanted line breaks and even taking over space of the actual visualization.

This change is now implementing shortening of the widget's title and adding an ellipsis when the available space is exceeded.

Before:

image

After:

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/common/EditableTitle.tsx
+++ b/graylog2-web-interface/src/views/components/common/EditableTitle.tsx
@@ -26,6 +26,7 @@ const StyledStaticSpan = styled.span(({ theme }) => css`
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
+  min-width: 20px;
 `);
 
 const StyledInput = styled.input(({ theme }) => css`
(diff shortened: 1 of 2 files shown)


linuspahl (commented):
With a small widget and / or monitor size, the title can disappear:

image
Unfortunately I do not have a good solution for this in mind. It would be great if the title has a `min-width` and we display the actions in a new line, when the widget is too small.

I also noticed we are displaying the message "Double click the title to edit it." when hovering over the title. We should also display the full title here, otherwise the full title can not be displayed easily.

dennisoelkers:
I would say for a small widget it's okay if the title disappears. Showing the actions in a new line would cost too much space from the visualization, especially for small widgets/resolutions. As long as we show the full title on hover, the user is still able to see the title on demand.

linuspahl:
Alright, in this case let's define a min width of e.g. `20px` for the title, to ensure it will never be completely hidden, like in this case: 

image

Compared with the previous widget header it is now possible that we hide widget actions, when a widget is very small. But IMO this is an edge case.
image

linuspahl (approved):
Looks good to me!

PR #11194, opened 2021-08-20
Replacing reflux `ToolsStore` with object of functions.

dennisoelkers:
## Motivation and Context

Although `ToolsStore` is wrapped in a reflux constructor, it does not hold any state and performs only calls to the backend returning the response through the generated promise, so it does not benefit from reflux at all. 

Due to this, and the fact that wrapping it in `Reflux.createStore` swallows the type information for the wrapped
functions, this PR is removing the reflux-specifics and just exports an object of functions.

This fixes typings, removes a couple of `StoreProvider` calls and makes testing a bit easier.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/extractors/extractors_configuration/SplitAndIndexExtractorConfiguration.tsx
```diff
@@ -0,0 +1,138 @@
...
+        return;
+      }
+
+      const preview = (result.cut ? <samp>{result.cut}</samp> : '');
+
+      onExtractorPreviewLoad(preview);
+    });
+
+    promise.finally(() => setTrying(false));
+  }, []);
```
linuspahl, line 75: Currently we are always calling `ToolsStore.testSplitAndIndex` with the initial form values, since there are no dependencies for `useCallback`.

linuspahl (changes requested):
We can also remove ` StoreProvider.getStore('Tools')` in `JSONExtractorConfiguration` and `EditExtractor`.

Since the direct import is the preferred way to import this store, we should imo no longer include the ToolsStore in `registerBuiltinStores`.

graylog2-web-interface/src/components/extractors/extractors_configuration/SplitAndIndexExtractorConfiguration.tsx
dennisoelkers: Good catch!

linuspahl (approved):
Looks good to me!

PR #11197, opened 2021-08-23
Lazy initialize processingErrors list

mpfz0r:
Processing errors should still be rare cases in regular message
processing. Better save the memory for messages without errors.


diesaulov (approved):
👍

PR #11220, opened 2021-08-29
Changes to be able to compile with JDK 11

uskr:
- Most java code changes are related to new errorprone checks
- Changes to EnterpriseMetadataSummary to properly override equals and hashCode
- Changed new Double, deprecated in JDK 11, to Double.valueOf

## Description
Changes necessary to compile the graylog-project, and all sub-projects, with JDK 11.

## Motivation and Context
Since JDK 11 is the latest LTS version, and brings significant performance improvements in many areas, I believe it is too good to pass on the free performance gains.

Please keep in mind I am submitting pull requests to all involved projects below, so the full project can compile and execute properly under JDK 11.

- graylog-project
- graylog2-server
- graylog-plugin-threatintel
- graylog-plugin-integrations
- graylog-plugin-aws

## How Has This Been Tested?
The code changes in itself are minimal.

For my current setup, I compiled and ran all the projects using openjdk 11.0.11 under Ubuntu 20.04. No problems, so far.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-plugin-archetype/src/main/resources/archetype-resources/pom.xml
+++ b/graylog-plugin-archetype/src/main/resources/archetype-resources/pom.xml
@@ -54,10 +54,11 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.release>11</maven.compiler.release>
 
-        <!-- Plugins will not be deployed by default - set to `false` if you actually want to deploy it -->
+    <!-- Plugins will not be deployed by default - set to `false` if you actually want to deploy it -->
         <maven.deploy.skip>true</maven.deploy.skip>
 
         <graylog.version>${project.parent.version}</graylog.version>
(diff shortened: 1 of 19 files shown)


uskr:
Please, keep in mind that, after the proposed changes, you will only be able to compile and run with JDK 11+, not 1.8.

On average, there is a 16% performance increase on switching.

Generally, you do NOT need to compile with JDK11. You can compile with JDK 1.8 and run with 11 to leverage most performance gains. I did not test this with Graylog. Some packages may break this rule and not execute.

JDK 11 brings many new features and APIs that developers can leverage. That was the main reason I decided to implement these changes. I am developing some plugins to Graylog that needed some internal changes to the graylog2-server package, and the new JDK 11 APIs and performance gains were a must.

Just wanted to post some extra background/context to this proposal.

Thank you.

uskr:
Will need a build setup with JDK 11, for the build test to pass.

bernd:
@uskr Thank you very much for the pull-requests! Getting these in will probably take some time because we need to test it a lot. Also, we have to decide when we can make the switch to requiring Java 11. We still have a lot of users and customers that run on Java 8, so this would be a breaking change for them.
We might have to put the 11 build into a maven profile so we can easily enable it for testing. I guess the code changes also work with Java 8.

uskr:
I would believe that to be true.

You can probably take the path of updating all dependencies, specially errorprone, apply the java code fixes, but leave the maven compile changes out. That way you can test everything else, while still on Java 8.

This kind of transition is always tricky, but something you could do to make it smooth, without adding a lot of pressure on anybody, is once you judge the changes are stable, add a second maven profile to compile with Java 11, so you can auto generate packages for java8 and java11, and do this for one major version. You get the best of both worlds, giving people time to move on and having a fallback if anything needs adjustments.

Keep up the great work. I am cheering for you all!

janheise:
@uskr Thank you for your work on this! I close your PR because we're switching to Java 17 with no intermediate steps.

Issue #6663, opened 2019-10-25
Allow creating searches through the API without assigning an id explicitly


alex-konn:
As an API user I would like to create a search through the API without assigning a valid mongo ObjectId beforehand.

Currently POST /views/search returns a 500, if the submitted search object doesn't contain a valid Mongo ObjectId. Invalid IDs at least yield a half-decent error message, omitting the ID gives no error message at all.

## Possible Solution
Assign a new ObjectId to the search object, if it doesn't contain one. If an invalid ID is included, we should return a better error message.

jgrammen-agilitypr:
I am also having this issue when trying to run search via the graylog api
https://community.graylog.org/t/graylog-3-3-9-search-api/18216

all the id fields had to be supplied as "id":"?"
for a valid search to be performed

todvora:
Implemented in #11247

PR #11247, opened 2021-09-07
Simplify search requests, provide default values

### LINKED ISSUE #11234: Simplify Search API to make it usable with minimal request
Ideally we want users to have an easy time operating our API for scripting purposes.

Right now the search API (`/api/views/search/sync`) requires a rather big request to generate a response. Validation and error handling is not ideal supporting a user trying to write a request from scratch.

Ideally we should:

  - try to write a search request from scratch (starting with an empty body)
  - identify missing validations/assertions
  - implement handling of parameters which can be optional so they can be left out

todvora:
The goal of this PR is to simplify search API (/api/views/search/sync) request needed to obtain a basic response.  

## Description
The first obvious road block are the IDs (both MongoDB ObjectId and UUIDs) in the request structure. Currently they are autogenerated by the frontend. But for any REST/CLI usage they are not necessary and may be generated by the backend itself. They are generated only if not provided, so there is no need to adapt the current frontend logic now. 

Completely empty requests ends with no error message and NPEs in the backend, so there need to be validation adapted. 

The (empty) query and default 300s relative time range can be provided by default, the frontend does the same.

## Motivation and Context

Fixes #11234, #6663

Without these UUIDs, a simple request may now look like this:

```
{
   "queries":[
      {
         "query":{
            "type":"elasticsearch",
            "query_string":""
         },
         "timerange":{
            "type":"relative",
            "from":300
         },
         "search_types":[
            {
               "type":"messages"
            }
         ]
      }
   ]
}
```

Adapted error handling to provide an error message in case of empty request body:

image

After providing default query and timerange, this is the shortest valid request:

```
{
   "queries":[
      {     
         "search_types":[
            {
               "type":"messages"
            }
         ]
      }
   ]
}
```

## How Has This Been Tested?
Added unit tests, manually tested.

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
@@ -1,0 +1,46 @@
+package org.graylog.plugins.views;
+
+import io.restassured.specification.RequestSpecification;
+import org.graylog.storage.elasticsearch7.ElasticsearchInstanceES7Factory;
+import org.graylog.testing.completebackend.ApiIntegrationTest;
+import org.graylog.testing.completebackend.GraylogBackend;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.graylog.testing.completebackend.Lifecycle.CLASS;
+import static org.hamcrest.core.IsEqual.equalTo;
+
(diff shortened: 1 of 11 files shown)


dennisoelkers (approved):
Good work! Thanks! :tada:

todvora:
Thank you for the review and merge!

PR #11249, opened 2021-09-07
Execute stale jobs.

### LINKED ISSUE #8612: Event definitions stop working after server is crashed
## Current Behavior

After server is crashed (due to OOM) and started again we stopped getting notifications. 
We've noticed there is no any record in `Alerts & Events`. After digging further we could observe that Event Definitions were left in wrong state after the crash.
On screenshot below (taken at 2020-07-22) below you can see broken **HTTP 500** event definition left in running state with old execution times (2020-07-10). Problem touched all event definitions. 
Second event definition **At least on ERROR message** looks fine since we've disabled and enabled it, which fixed the state.

image

What's more, we've checked some Graylog metrics we collect via Prometheus and found that it touched all scheduled jobs:

image

So I suppose after our Graylog pod restart due to OOM job execution engine couldn't start properly. Unfortunately we didn't collect server logs from Graylog.
## Environment

* Graylog Version:  Graylog 3.3.2+ade4779
* MongoDB Version: 4.1.13
* Environment: kubernetes

AntonEbel:
Locked jobs are now executed if the node no longer exists and a time threshold is exceeded.

## Motivation and Context

The current behavior of graylog is that only the node on which the job was running and not shut down properly can resolve the hanging job on the next startup. For each Graylog node, a node ID is generated at first startup and stored in a file on the file system. In a kubernetes environment old containers are discarded and new ones are created with new node IDs. So the hanging jobs are never resolved from the old nodes.

closes #8612

Auto-unlock for scheduler triggers was already planned in this issue Graylog2/graylog-plugin-enterprise#1057 but with a different implementation.

## Types of changes


graylog2-server/src/main/java/org/graylog/scheduler/JobSchedulerConfiguration.java
```diff
@@ -27,11 +28,19 @@
 @SuppressWarnings({"FieldCanBeLocal", "unused", "WeakerAccess"})
 public class JobSchedulerConfiguration implements PluginConfigBean {
     public static final String LOOP_SLEEP_DURATION = "job_scheduler_loop_sleep_duration";
+    public static final String LOCK_EXPIRATION_TIME_MINUTES = "job_lock_expiration_time_minutes";
```
bernd, line 31: How about using the same prefix as the other option for consistency?
```suggestion
    public static final String LOCK_EXPIRATION_TIME_MINUTES = "job_scheduler_lock_expiration_time_minutes";
```

bernd (changes requested):

thll (approved):
LGTM and works. Nice work!

Handing over to @bernd to give the final OK.

bernd (approved):
Once @thll's https://github.com/Graylog2/graylog2-server/pull/11249#discussion_r742862251 is fixed, I am happy. :smile: 

Thanks a lot for working through all the feedback! :pray:

thll (approved):

PR #11250, opened 2021-09-07
Improve memory profile of `useStore`.

dennisoelkers:
## Motivation and Context

Prior to this change, `useStore` would hold the complete state of the store it connects to in memory. This is useful, because it allows us to check if the new state of the store is equal to the previous one before.  If it is not, the update function can exit early.

The downside of this is that if a store contains a big structure and the consumer is only interested in a very small subset, we can use the props mapper to return only the relevant part of it, but `useStore` would still keep the whole structure in memory. An example for this is:

```
const foo = useStore(FooStore, (state) => state.length);
```

Assuming that the state of `FooStore` is a very large array, keeping a reference of it in `useStore`'s memory is useless, if we are interested in its size only.

This change is now improving `useStore`'s behavior in that it maintains its internal state and references related to the _mapped_ versions of them. This allows us to keep only the relevant parts of the store state.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/EntityListItem.tsx
+++ b/graylog2-web-interface/src/components/common/EntityListItem.tsx
@@ -71,7 +71,7 @@ const EntityListItem = ({ actions, contentRow, description, title, titleSuffix }
     <StyledListItem>
       <Row className="row-sm">
         <Col md={12}>
-          <div className="pull-right hidden-xs">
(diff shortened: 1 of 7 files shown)


dennisoelkers:
This introduces another set of problems. `useStore` will be prospectively replaced by either `redux` or `react-query`, so I am closing this.

PR #11252, opened 2021-09-07
Improve performance when comparing React elements

### LINKED ISSUE #8657: Browser tab freezes after saving a Pipeline Rule and returning to the Pipeline Rules page
## Expected Behavior

That the Pipeline Rules page successfully loads and is responsive after clicking Save & Close on the Edit Pipeline page.

## Current Behavior

After returning to the Pipeline Rules page when clicking Save & Close or even the browser Back button after editing a Pipeline Rule, the entire browser tab freezes. All metrics for the rules are stuck in the "Calculating..." state:

image

For me, this is only reproducible in my local dev environment on the `cloud/cloud-3.3.json`, which is based on `3.3`.

I have not yet tested with a prod jar build, but I plan to. 

1. Edit any pipeline rule.
2. Click Save & Close on the rule (or Back)
3. Browser tab freezes and never recovers. CPU spins up high.

## Your Environment

* Graylog Version: Graylog 3.3.3-SNAPSHOT (`cloud/cloud-3.3.json` manifest)
* Elasticsearch Version: 5.6-alpine
* MongoDB Version: 3.6
* Operating System: Mac OS Catalina 10.15.5 (19F101)
* Browser version: Chrome 84.0.4147.105 (Official Build) (64-bit)

edmundoa:
Performing deep equal comparisons of props may be expensive, specially when comparing React elements passed as children. This gets even worse during development, since React elements will include additional debugging information that also needs to be compared. In my analysis, this caused the tab freezes in the Pipeline Rules page described in #8657.

This PR customises the deep comparison in `isDeepEqual.ts`, using [`react-fast-compare`](https://github.com/FormidableLabs/react-fast-compare) to compare React elements.

You can find performance profiles below but I thought it may be interesting to see a gist of them without all the hassle of loading them, so here are some captures of the critical moment. Before the change:

Screenshot 2021-09-08 at 17 01 35

After the change:

Screenshot 2021-09-08 at 16 54 21

<details>

<summary>Performance profiles</summary>

For reference, I uploaded some performance tests I captured on my browser with different comparison methods:
[new-pipeline-rules-performance-profiles.zip](https://github.com/Graylog2/graylog2-server/files/7129889/new-pipeline-rules-performance-profiles.zip)

- `new-pipeline-rule-lodash.json`: Deep equality using `lodash`, just as we did before the PR
- `new-pipeline-rule-react-fast-compare.json`:  Deep equality of React elements using `react-fast-compare`, as we do in this PR

The pipeline rules used during the performance tests are also included in the zip file.
</details>

Fixes #8657

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -85,6 +85,7 @@
     "react-color": "^2.14.0",
     "react-day-picker": "^7.4.8",
     "react-error-boundary": "^3.1.1",
+    "react-fast-compare": "^3.2.0",
     "react-grid-layout": "^1.2.2",
     "react-immutable-proptypes": "^2.1.0",
     "react-isolated-scroll": "^0.1.1",
--- a/graylog2-web-interface/src/stores/isDeepEqual.ts
+++ b/graylog2-web-interface/src/stores/isDeepEqual.ts
@@ -15,6 +15,7 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import { isEqualWith, isFunction } from 'lodash';
+import { isValidElement } from 'react';
 import isEqual from 'react-fast-compare';
 
 const hasFn = (obj, fn) => (obj && obj[fn] && isFunction(obj[fn]));
@@ -30,7 +31,11 @@ const _isEqual = (first, second) => {
     return first.equals(second);
   }
 
-  return isEqual(first, second);
+  if (isValidElement(first) || isValidElement(second)) {
+    return isEqual(first, second);
+  }
+
+  return undefined;
 };
 
 const isDeepEqual = (first: any, second: any): boolean => isEqualWith(first, second, _isEqual);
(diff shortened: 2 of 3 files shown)


kyleknighted (changes requested):
Code looks great, is it possible to reduce the yarn.lock to only the changes needed for `react-fast-compare`? Unless @dennisoelkers thinks this is good, not sure the caveats of updating dependencies

edmundoa:
@kyleknighted I think the `yarn.lock` changes happen because it gets out of sync with changes and dependency updates from `package.json`. As far as I know, committing those changes would not have any effect on the packages installed by `yarn`, since the changes in `yarn.lock` also occur when running `yarn install` on a clean repository and will therefore also be included in a build.

kyleknighted (approved):

dennisoelkers (approved):
Looks good! Thanks a lot for this, @edmundoa!

PR #11253, opened 2021-09-07
Fix potential race condition in traffic calculation

### LINKED ISSUE #11248: Potential race condition in traffic calculation
Non-thread-safe field variables are used in traffic calculations and can lead to incorrect results.

## Possible Solution

Make field variables thread safe.

Could not be reproduced.

## Context

HS-525674391

AntonEbel:
Make calculation fields of TrafficCounterCalculator volatile.

## Motivation and Context

close #11248 

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/periodical/TrafficCounterCalculator.java
+++ b/graylog2-server/src/main/java/org/graylog2/periodical/TrafficCounterCalculator.java
@@ -35,9 +35,9 @@ public class TrafficCounterCalculator extends Periodical {
     private final NodeId nodeId;
     private final TrafficCounterService trafficService;
     private final MetricRegistry metricRegistry;
-    private long previousInputBytes = 0L;
-    private long previousOutputBytes = 0L;
-    private long previousDecodedBytes = 0L;
+    private volatile long previousInputBytes = 0L;
+    private volatile long previousOutputBytes = 0L;
+    private volatile long previousDecodedBytes = 0L;
     private Counter inputCounter;
     private Counter outputCounter;
     private Counter decodedCounter;


graylog2-server/src/main/java/org/graylog2/periodical/TrafficCounterCalculator.java
```diff
@@ -35,9 +35,9 @@
     private final NodeId nodeId;
     private final TrafficCounterService trafficService;
     private final MetricRegistry metricRegistry;
-    private long previousInputBytes = 0L;
-    private long previousOutputBytes = 0L;
-    private long previousDecodedBytes = 0L;
+    private volatile long previousInputBytes = 0L;
+    private volatile long previousOutputBytes = 0L;
+    private volatile long previousDecodedBytes = 0L;
```
bernd, line 40: Can you elaborate on the thinking behind this change? The class is a Periodical that runs every second. But I don't see how there would be concurrent access to the bytes fields from multiple threads. :thinking:

bernd: Hmm, is it because the periodical might be scheduled on different threads for every execution?

AntonEbel: After a second look at the JavaDoc of [ScheduledThreadPoolExecutor](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/concurrent/ScheduledThreadPoolExecutor.html#scheduleAtFixedRate(java.lang.Runnable,long,long,java.util.concurrent.TimeUnit)), it seems that I was wrong to make the fields volatile. The JavaDoc mentions:

> Successive executions of a periodic task scheduled via scheduleAtFixedRate or scheduleWithFixedDelay do not overlap. **While different executions may be performed by different threads, the effects of prior executions happen-before those of subsequent ones.**

I must have gotten it wrong the first time. It seems that I am not the only one as you can see in this [stackoverflow question](https://stackoverflow.com/questions/55542189/does-an-object-always-see-its-latest-internal-state-irrespective-of-thread). I thought [visibility](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.3.1.4) or a [non-atomic write](https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7) would cause problems. So it's really puzzling what the problem described in the HubSpot ticket could be. I think we can then close the PR and Issue and delete the branch.

bernd (approved):

graylog2-server/src/main/java/org/graylog2/periodical/TrafficCounterCalculator.java
bernd: Okay, I just merged the PR anyways. I shouldn't hurt and there might be an edge case that we missed. Thank you!

Issue #10836, opened 2021-06-11
Clicking on sidebar tooltip opens the sidebar


edmundoa:
While trying to press the search button my pointer slid for a moment to the left and was surprised to see the sidebar open. Trying to find the reason I saw that clicking on the tooltip opens the sidebar as well.

https://user-images.githubusercontent.com/716185/121718479-a5eb1a00-cae2-11eb-8113-e2522ae4ebac.mp4

## Your Environment
* Graylog Version: Graylog 4.1.0-beta.1+1cc9dc

PR #11270, opened 2021-09-10
Only display sidebar tooltip when hovering over sidebar navigation item.

linuspahl:
## Motivation and Context

It is currently possible to hover over and click on a sidebar tooltip.

image

As described in https://github.com/Graylog2/graylog2-server/issues/10836 this can be counterintuitive in some cases.
With this PR we are hiding the sidebar tooltip, one the user no longer hovers over the navigation item.

Fixes: #10836


dennisoelkers (approved):
Looks good, works great! Good work!

PR #11274, opened 2021-09-10
Selecting dashboard page specified in URL immediately at load.

dennisoelkers:
## Motivation and Context

Prior to this change, when a `page` parameter was part of the url when
loading a dashboard, the sequence was like this:

  - loading the dashboard
  - activating the first page during load
  - parsing the URL
  - activating the page id specified in the URL (if present)

This leads to unwanted visual artifacts, the first page will be shown
shortly before switching to the selected page. To avoid this, this PR is
now passing the query part of the URL to the view loader, which will
activate a specified page id right when loading the dashboard.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/stores/ViewStore.ts
+++ b/graylog2-web-interface/src/views/stores/ViewStore.ts
@@ -69,14 +69,14 @@ export const ViewActions: ViewActionsType = singletonActions(
 
 type ViewStoreType = Store<ViewStoreState>;
 
-const _selectedQuery = (queries: Immutable.List<string>, activeQuery: string, queryId: string) => {
+const _selectedQuery = (queries: QuerySet = Immutable.Set(), activeQuery: string, queryId: string): QueryId => {
   const selectedQuery = queryId ?? activeQuery;
 
-  if (selectedQuery && queries.find((id) => (id === selectedQuery))) {
+  if (selectedQuery && queries.find(({ id }) => (id === selectedQuery))) {
     return selectedQuery;
   }
 
-  return queries.first();
+  return queries.first()?.id;
 };
 
 export const ViewStore: ViewStoreType = singletonStore(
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/views/stores/ViewStore.ts
```diff
@@ -156,15 +166,14 @@ export const ViewStore: ViewStoreType = singletonStore(
 
       return promise;
     },
-    load(view: View, isNew = false): Promise<ViewStoreState> {
+    load(view: View, isNew = false, queryId: string = undefined): Promise<ViewStoreState> {
       this.view = view;
       this.dirty = false;
 
-      /* Select selected query (activeQuery) or first query in view (for now).
+      /* Select query id passed thorugh URL, selected query (activeQuery) or first query in view (for now).
```
linuspahl, line 173: *through

graylog2-web-interface/src/views/stores/ViewStore.ts
```diff
@@ -156,15 +166,14 @@ export const ViewStore: ViewStoreType = singletonStore(
...
          Selected query might become a property on the view later. */
       const queries = view.state.keySeq().toList();
-      const firstQueryId = view?.search?.queries?.first()?.id ?? queries.first();
```
linuspahl, line 166: Previously the default `activeQuery` was based on `view?.search?.queries?.first()?.id` to ensure we are always  defining the first search query as the active one: https://github.com/Graylog2/graylog2-server/pull/10790

linuspahl (changes requested):
Works well in general! I just saw we are reverting a little improvement I implemented in the past.

graylog2-web-interface/src/views/stores/ViewStore.ts
dennisoelkers: Thanks, good catch! Fixed it.

graylog2-web-interface/src/views/stores/ViewStore.ts
dennisoelkers: :white_check_mark:

linuspahl (approved):
Looks good to me!

PR #11282, opened 2021-09-13
Self-managed index templates

diesaulov:
In some scenarios we would like to take over the responsibility of the 
index template lifecycle management due to specifics of particular
features. For instance, it's applicable when one wants the same
index template to be employed by multiple index sets.

This commit Introduces a new index template type to tell the index set 
framework that the index template lifecycle is manually managed 
by the calling side.


diesaulov:
Is replaced by https://github.com/Graylog2/graylog2-server/pull/11290

PR #11283, opened 2021-09-13
Refactor `ReactGridContainer` and install `@types/react-grid-container`.

linuspahl:
## Motivation and Context

With this PR we are refactoring the `ReactGridContainer` component by:
- migrating it to TypeScript
- converting it to a functional component
- removing its (layout) state, as far as i can tell there is currently no benefit in maintaining the layout not only as part of the view, but also in the `ReactGridContainer`.
- always defining `useCSSTransforms` for the `react-grid-layout`, since it improves the painting performance.

We are also installing `@types/react-grid-container`.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/ReactGridContainer.tsx
```diff
@@ -0,0 +1,272 @@
...
+    lg: number,
+    md: number,
+    sm: number,
+    xs: number,
+  },
+  draggableHandle?: string,
+  isResizable?: boolean,
+  locked?: boolean,
+  measureBeforeMount?: boolean,
+  onPositionsChange: (newPositions: Array<WidgetPositionJSON>) => void,
```
linuspahl, line 128: Previously we only called this function after checking if the layout changed, because `onLayoutChanged`  "may be triggered when clicking somewhere in a widget, check before propagating the change".

I could not find a case were `onLayoutChanged` will be triggered unintentional.

graylog2-web-interface/src/components/common/ReactGridContainer.tsx
```diff
@@ -0,0 +1,272 @@
...
+      id: widget.i,
+      col: widget.x + 1,
+      row: widget.y + 1,
+      height: widget.h,
+      width: widget.w,
+    });
+  });
+
+  onPositionsChange(newPositions);
+};
```
linuspahl, line 110: Previously we only called this function after checking if the layout changed, because `onLayoutChanged`  "may be triggered when clicking somewhere in a widget, check before propagating the change".

I could not find a case were `onLayoutChanged` will be triggered unintentional.

dennisoelkers (changes requested):
Looks good and works great. The component is now way cleaner, thanks!

There is just one thing left: I think we need to update the usage of the component for the frontend component gallery in [here](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/components/common/ReactGridContainer.md).

dennisoelkers (approved):
Good work!

PR #11287, opened 2021-09-14
Trigger cleanup of containers and networks after all IT tests finished

todvora:
This change ensures that all testcontainers and created networks get removed after all the test finish

## Description
Register a shutdown hook that closes or started resources

## Motivation and Context
Without this change, after each test-suite run there remain a docker network and an elastic container running. With repeated suite runs it consumes all RAM of the machine and crashes.

## How Has This Been Tested?
Run the integration tests several times (may be from your IDE) and observe outputs of following commands:

```
watch docker ps
watch docker network ls
```

Before the change, you should be able to see that both running containers and networks accumulate with each run and are not stopped. After the change you shouldn't see any new networks and no running elastic containers.

image

This is the state after 4 successful test-suite runs, you can see 4 elastic containers running and 4 networks created and not removed.

## Types of changes

## Checklist:


janheise (approved):
LGTM, great work!

todvora:
Thanks for the review and merge!

PR #11290, opened 2021-09-14
Pluggable Index Templates

diesaulov:
Due to the hard-coded nature of the Index Template type, the core server module requires changes every time a new custom Index Template is needed to build a new feature, what violates the Open-Closed principle and makes development less flexible.

This PR replaces the `TemplateType` enum with a string value and adds support of external Index Template providers.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2702

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/shared/bindings/GenericBindings.java
```diff
@@ -90,6 +94,12 @@ protected void configure() {
                 .setDefault()
                 .to(DefaultFailureHandlingConfiguration.class);
 
+        final Multibinder<IndexTemplateProvider> indexTemplateProviderMultibinder = Multibinder.newSetBinder(binder(), IndexTemplateProvider.class);
+
+        indexTemplateProviderMultibinder.addBinding().to(MessageIndexTemplateProvider.class);
+        indexTemplateProviderMultibinder.addBinding().to(EventIndexTemplateProvider.class);
+        indexTemplateProviderMultibinder.addBinding().to(GimIndexTemplateProvider.class);
+
```
bernd, line 102: We usually use a map binder and use the type string as key. Using a set binder and then find factories by iterating over the set and comparing the type works as well, of course. Any specific reason why you didn't choose a map binder?

Apart from the easier lookup, another advantage of the map binder is, that trying to bind the same key twice results in a binding error on server startup. With the set binder we have to check that there is only a single implementation for the same key at runtime and throw an IllegalStateException.

bernd (changes requested):
Thanks for the refactoring! :pray: I will test this tomorrow. One thing I noticed:

graylog2-server/src/main/java/org/graylog2/shared/bindings/GenericBindings.java
diesaulov: https://github.com/Graylog2/graylog2-server/pull/11290/commits/abcd797052c00b60b2e09380fa5085ed706e14f8

bernd (approved):
Tested in my dev environment and LGTM! :+1: Good work, thank you!

I think at some point in the future we can add an `addIndexTemplateType(String, Class<? extends IndexTemplateProvider>)` to `PluginModule` like we do for other pluggable objects to make it easier for plugins to register them.

PR #11244, opened 2021-09-06
Improve re-rendering of widget grid on search page.

dennisoelkers:
## Motivation and Context

This PR is improving rendering performance of the widget grid on the search/dashboard pages by reducing the amount of re-renders. This is achieved by:

  - Pushing state (mostly derived from stores) down to its actual consumption whenever possible, e.g. extracting widget results closely to the visualization component
  - Making callbacks static
  - Extracting components which do not change when the search results are updated and rendering them _next_ to components which need to update

This allowed me to improve rendering times for two scenarios:

  - Updating the search
  - Toggling edit mode on a widget

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #11291, opened 2021-09-15
Passing positions to grid again, instead of through widgets.

dennisoelkers:
## Motivation and Context

This change is effectively reverting a small change performed in #11244, which changed our way of using `react-grid-layout` by passing widget coordinates through widget props instead of a global positions object.

Unfortunately, the `getDerivedStateFromProps` method in the `ReactGridLayout` component does not take the content of the `children` props into account when determining if a new layout needs to be calculated, just the children's keys. This leads to the fact that we cannot update widget positions "from the outside" when passing coordinates through children props.

Therefore, this change is using a global positions map again, making sure that it's aligned with the current state of widgets by retrieving it in one central point in the code.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me, fixes the described problem!

PR #11303, opened 2021-09-17
Ensure filter events have "source_streams" set

### LINKED ISSUE #11301: Permission check problem with filter events when no stream is configured
## Description

Permissions on created events and alerts are checked based on the `source_streams` field in events.

When configuring an event definition of type "Filter & Aggregation", creating events if "Filter has results" and leaving the "Streams" empty, the resulting events have an empty `source_streams` field. That means all users are able to see the events.

This needs to be fixed.

## Your Environment

* Graylog Version: 4.0

bernd:
When an event definition doesn't have any streams configured, this
previously resulted in an empty "source_streams" field in the event
which allowed all users to see the event.

To fix it, we are now using the same way to compute the "source_streams"
for filter events that we are using for aggregation events.

Fixes #11301

This needs to be backported to 4.1 and 4.0.


mpfz0r (approved):
Works and LGTM :+1:

PR #11312, opened 2021-09-20
Create simple common component which centers its children and implement it to fix widget error layout.

linuspahl:
## Motivation and Context

In a previous PR we removed some styles which were still being used for the widget error layout:
image

With this PR we are creating a common component which centers its children. The component can not only be used to fix the widget error styling: 
image

But also for every other case where we need to center content horizontally and vertically.


dennisoelkers (approved):
:pray:

PR #11330, opened 2021-09-23
Retrieve store state 2nd time after registration to avoid race condition

### LINKED ISSUE #11316: Frontend sometimes displays loading spinner instead of actual page, after page refresh.
## Behavior

In some cases the frontend is just displaying the message "We are preparing Graylog for you..." after a page refresh, instead of the actual page.

image

This error can't be reproduced easily, since it only happens occasionally.

## Your Environment

The error occurs with development and production build for 4.2.

dennisoelkers:
## Motivation and Context

Prior to this PR, the `useStore` logic to connect a component with a store looked like this:

  1. retrieve current state from store and set it in state
  2. register listener on store which updates state upon store updates

This could lead to a race condition in some scenarios, where between 1.  & 2. a store update is happening, which is not processed by the `useStore` hook due to no being registered to the store at this time.

In order to prevent this, a 3rd step is added to `useStore`'s logic:

  3. retrieve current state from store again and set it in state

This helps ups keeping up with potential store updates between 1. & 2.  In addition, it does not necessarily mean that an additional rerendering of the component is triggered, if the state retrieved in 1. and 3. is shallowly equal (due to the setter function of `useState` not triggering a rerender when [both values are equal](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update)).

Fixes #11316.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2733

## How Has This Been Tested?

This PR was created mainly to fix the issue in #11316. Therefore, the `AppFacade` component was modified to render a component which is triggering a full page refresh, effectively going into an infinite loop of validating the session -> refreshing the page, until it gets stuck in the condition described in the issue.

After fixing the issue with the changes in this PR, the loop never got stuck in the condition again.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/permissions/EntityShareModal.test.tsx
+++ b/graylog2-web-interface/src/components/permissions/EntityShareModal.test.tsx
@@ -37,13 +37,17 @@ jest.mock('stores/permissions/EntityShareStore', () => ({
   },
   EntityShareStore: {
     listen: jest.fn(),
-    getInitialState: jest.fn(() => ({ state: mockEntityShareState })),
+    getInitialState: jest.fn(),
   },
 }));
 
 jest.setTimeout(10000);
 
 describe('EntityShareModal', () => {
+  beforeEach(() => {
+    asMock(EntityShareStore.getInitialState).mockReturnValue({ state: mockEntityShareState });
+  });
+
   beforeAll(() => {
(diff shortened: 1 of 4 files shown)


linuspahl (approved):
Looks good to me, it is fixing the described problem!

PR #11333, opened 2021-09-23
Proof-of-concept: Make Graylog node easily replaceable

thll:
Reduce state of a Graylog node to make it more easily replaceable in highly dynamic environments.

- Remove dependency on persistent node ID. If a node dies and never comes back, this shouldn't have any negative consequences
  - [ ] Inputs can be tied to a specific node ID. We probably need to remove this option.
  - [ ] The locking mechanism for scheduled jobs relies on a persistent node ID. This can lead to stale jobs if a node dies. We have https://github.com/Graylog2/graylog2-server/pull/11249 to improve that.
 - Remove the `is_master` flag.
   - [ ] Some services will only execute certain tasks when started on the master. Convert each of them to a scheduled job which is executed by any node. The job needs to run only once after being scheduled.
   - [ ] Convert simple "master-only" periodicals to scheduled jobs.
   - [ ] Rewrite `IndexFieldTypePollerPeriodical`. This is one of the fancier periodicals. It internally schedules periodical polling tasks to keep index field types up to date.
   - [ ] Convert archive jobs to scheduled jobs. Add a label mechanism to label nodes accordingly (i.e. with `on-disk-archive` if they have access to the on-disk archive. Only those nodes should be able to execute jobs which are tagged with the `on-disk-archive` label.
   - [ ] Refactor https://github.com/Graylog2/graylog-plugin-aws so that it will start on any node and not just on master (see [here](https://github.com/Graylog2/graylog-plugin-aws/blob/79b4461f9caefc01f927ac41c231ec1483cfc285/src/main/java/org/graylog/aws/AWSPluginMetadata.java#L66)). **Be aware of the Forwarder runtime which doesn't provide access to MongoDB when refactoring.**


thll:
Superseded by #11373

Issue #11355, opened 2021-09-27
Message tables in the dashboard full screen mode display an error.


linuspahl:
## Description

Displaying a message table widget in the dashboard full screen mode results in the following error:
`Cannot read properties of undefined (reading 'security')`. The error is being catched by the widget error boundary. 

## Your Environment

* Graylog Version: Graylog 4.2.0-SNAPSHOT+1c99273

PR #11358, opened 2021-09-27
Render `GlobalContextProviders` for all pages a user can access, when logged in.

linuspahl:
## Description
With this PR we are including the `GlobalContextProviders` in the `LoggedInPage` instead of the `App` component
This way we do not need to implement the `GlobalContextProviders` on pages which are not wrapped by the `App` component, like the `ShowDashboardInBigDisplayMode`.

As described in https://github.com/Graylog2/graylog2-server/issues/11355 we are currently throwing an error for message tables in the dashboard full screen mode, because the  `ShowDashboardInBigDisplayMode` was not wrapped with the `GlobalContextProviders` bevor.

Fixes:  https://github.com/Graylog2/graylog2-server/issues/11355 

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/2752

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change


dennisoelkers (approved):
Looks good, makes sense, fixes the issue! :)

PR #11360, opened 2021-09-28
Add a trace log for invalid GELF messages

mpfz0r:
This gives a user more context when debugging incomplete
GELF messages, where the source IP is not enough.

[HS #553887549]


kmerz (approved):
LGTM :+1:

PR #11362, opened 2021-09-28
Manually stopped inputs do not auto-restart

### LINKED ISSUE #10955: Disable inputs auto starting
## What?

As a Graylog user I would like the inputs to not "auto start" all the time. It would be great to either have a "disable auto start for all inputs" setting or a toggle button where you can select individual inputs from auto starting and select "run with last setting" or something similar.

## Why?

Currently when a node reboots the input will automatically start on that node even if I don’t want it to. In some situations, where we have a surge of messages, the journal may be full and the node will change its status to THROTTLED. While in the THROTTLED state a Kubernetes pod running Graylog will begin to fail the health check against the lbstatus api. When this happens Kubernetes will delete the pod and redeploy another which only makes things worse.

patrickmann:
Currently, a server restart will automatically restart all inputs; even those that had been manually stopped. This is undesirable for several reasons - see linked issue for details.

We now persist the desired state of the input in the DB; and avoid restarting inputs where desired state is `STOPPED`.

Users can choose the legacy behaviour by setting a flag in the `graylog.conf`:
`auto_restart_inputs = true`

Note: 
- [Previously](https://github.com/Graylog2/graylog2-server/pull/3824), we changed the behavior of inputs to not automatically restart on updates, if they were in a stopped state. We must be careful not to regress that change with this PR.
- We must also be careful that a change to the Input classes does not break existing forwarders (see Graylog2/graylog-plugin-enterprise#2923)

resolves #10955 
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2923

### DIFF
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -482,8 +482,8 @@ inputbuffer_ring_size = 65536
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 
-# Manually stopped inputs are not auto-restarted. To re-enable the previous behavior, set autostart_inputs to true.
-#autostart_inputs = true
+# Manually stopped inputs are no longer auto-restarted. To re-enable the previous behavior, set auto_restart_inputs to true.
+#auto_restart_inputs = true
 
 # Enable the message journal.
 message_journal_enabled = true
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -482,6 +482,10 @@ inputbuffer_ring_size = 65536
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 
+# As of version 4.2, manually stopped inputs are not auto-restarted. To re-enable the
+# previous behavior, set autostart_inputs to true.
+#autostart_inputs = true
+
 # Enable the message journal.
 message_journal_enabled = true
 
--- a/misc/graylog.conf
+++ b/misc/graylog.conf
@@ -482,8 +482,7 @@ inputbuffer_ring_size = 65536
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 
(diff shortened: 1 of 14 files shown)


misc/graylog.conf
```diff
@@ -482,6 +482,9 @@ inputbuffer_ring_size = 65536
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 
+# Manually stopped inputs are not auto-restarted. To re-enable the previous behavior, set autostart_inputs to true.
+#autostart_inputs = true
```
boosty, line 486: I think the name might be misleading. It sounds as if this config setting would influence the general autostart behaviour (for new inputs), not just on server restart. Maybe we should use `auto_restart_inputs` instead?

boosty (commented):
Hi @patrickmann, please have a look at my comments for `misc/graylog.conf`. I reopened them, as it seems they were resolved without changing the config option and related comment.

misc/graylog.conf
```diff
@@ -482,6 +482,9 @@ inputbuffer_ring_size = 65536
 inputbuffer_processors = 2
 inputbuffer_wait_strategy = blocking
 
+# Manually stopped inputs are not auto-restarted. To re-enable the previous behavior, set autostart_inputs to true.
```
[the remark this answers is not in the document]
patrickmann: Fixed - forgot to include graylog.conf in the previous commit :/

misc/graylog.conf
patrickmann: Fixed - forgot to include graylog.conf in the previous commit :/

AntonEbel (changes requested):

AntonEbel (approved):
LGTM :+1:

boosty:
Assigned @dulanism for documentation

PR #11364, opened 2021-09-28
Improve positioning of trend bar in number visualization.

dennisoelkers:
## Motivation and Context

Prior to this PR, the trend bar of the number visualization was
positioned starting at 80% of the overall widget height. This could lead
to empty space if the trend bar was smaller than the remaining 20%.

This change is now making it stick to the bottom, leaving more space for
the number itself and looking better.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/visualizations/number/NumberVisualization.tsx
```diff
@@ -38,7 +38,7 @@ import type { CurrentViewType } from '../../CustomPropTypes';
 const GridContainer = styled.div`
   display: grid;
   grid-template-columns: 1fr;
-  grid-template-rows: 4fr 1fr;
+  grid-template-rows: 4fr auto;
```
linuspahl, line 41: We can now use `1fr auto` here. It is not changing the layout it is just slightly easer to understand.

dennisoelkers: You are right! Thanks! :)

linuspahl (approved):

PR #11372, opened 2021-09-29
Avoid not needed search creation on default and stream search page

linuspahl:
## Motivation and Context

Before this change we we created two searches, when opening the default or stream search page via a URL with search query params like: `?q=http_method:GET&rangetype=relative&relative=300`. This was not the case when opening a saved search. This problem occurred because we were creating a search with the default search query first and another one with the search settings provided by the URL query params.

Please note, this bug still exists when the URL query params contain a stream filter. Fixing this bug will require more complex changes, which we should implement in a follow-up PR.

### DIFF
--- a/graylog2-web-interface/src/views/logic/NormalizeSearchURLQueryParams.ts
+++ b/graylog2-web-interface/src/views/logic/NormalizeSearchURLQueryParams.ts
@@ -58,7 +58,7 @@ type StreamsQuery = {
   streams?: string,
 };
 
-export type RawQuery = RawTimeRange & StreamsQuery & { q?: string };
+export type RawQuery = Partial<RawTimeRange> & StreamsQuery & { q?: string };
 
 const _getRange = (query): RawTimeRange => {
   const rangetype = query.rangetype || DEFAULT_RANGE_TYPE;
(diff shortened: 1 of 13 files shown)


graylog2-web-interface/src/views/logic/NormalizeSearchURLQueryParams.ts
```diff
@@ -0,0 +1,143 @@
...
+        to: range.to,
+      } as AbsoluteTimeRange : undefined;
+    case 'keyword':
+      return range.keyword ? {
+        type: range.rangetype,
+        keyword: range.keyword,
+        timezone: range.timezone,
+      } as KeywordTimeRange : undefined;
+    default:
+      // @ts-ignore
```
dennisoelkers, line 108: To avoid using `@ts-ignore` here, we can do two things: 
  - Using the complete `range` to generate the `Error` in the next line (TS is complaining because `range.rangetype` is supposed to be `never` at this point), which could also help for debugging.
  - Using `@ts-expect-error` instead. This is doing the same as `@ts-ignore` (suppressing an error), but in addition it is raising an error if no error is suppressed. This way, we are also being notified when a new possible value for `rangetype` is introduced but it is not handled in a `case`-clause.

dennisoelkers (changes requested):
Looks good so far! Added just one minor remark.

graylog2-web-interface/src/views/logic/NormalizeSearchURLQueryParams.ts
linuspahl: Thanks for the tip, I really like the typing related advantages of `@ts-expect-error`.

dennisoelkers (approved):
LGTM and fixes the issue!

PR #11374, opened 2021-09-29
Handle empty responses on plaintext requests

### LINKED ISSUE #11343: Import Grok Pattern redirects 500 view then back to Grok Pattern page
## Expected Behavior

Do not redirect to 500 page

## Current Behavior

While importing a grok file

```
USERNAME [a-zA-Z0-9._-]+
USER %{USERNAME}
```

The site is redirected to the 500 error page and then re-renders with the Grok Pattern page again with the new grok patterns from the file but also showing UserNotification error

https://user-images.githubusercontent.com/122591/134677545-cf4f7a89-0c59-4ab0-8b5b-93f5107f350a.mp4

1. Upload above grok file

edmundoa:
Graylog server API returns JSON responses also when doing a plaintext
request. Sometimes though, it is possible that we get an empty response
from the server, which will throw an error when trying to parse the JSON
response using the JS `fetch` API.

Since the default response can read a JSON response from the server and
already handles the situation described above, this commit replaces the
custom response handler for plaintext responses with the default one.

Fixes #11343

### DIFF
--- a/graylog2-web-interface/src/logic/rest/FetchProvider.ts
+++ b/graylog2-web-interface/src/logic/rest/FetchProvider.ts
@@ -183,15 +183,7 @@ export class Builder {
     this.body = { body, mimeType: 'text/plain' };
     this.accept = 'application/json';
 
-    this.responseHandler = (resp) => {
-      if (resp.ok) {
-        reportServerSuccess();
-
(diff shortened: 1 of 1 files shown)


kyleknighted (approved):
lgtm

PR #11407, opened 2021-10-01
Include existing URLs in allow list

### LINKED ISSUE #11345: Lookup Table Data Adapter - Add URL Whitelist
## Expected Behavior

To be displayed the entire list of Whitelisted URLs

## Current Behavior

Potential race condition, on occasion while trying to add a URL to the Whitelist from the Data Adapter page, you will get a modal with no other URLs in it, just the new one you're adding now

image

image
image

edmundoa:
Before this change, changing the `newUrlEntry` prop in `URLWhiteListFormModal` would reset the component state, removing any previously configured URLs in the allow list.

This change ensures the component will keep the existing allow list when being reset, and does also work for users with the creator role.

Fixes #11345


ousmaneo (approved):
Tested and works! 👍 
Thank you for the changes

kyleknighted (approved):
https://user-images.githubusercontent.com/122591/139851969-7af84874-1dd5-4385-ab86-f5820ea90a6c.mp4

If you can fix this double update when the URL is changed, that would be awesome. But if it's not feasible within this PR, then this all looks good!

edmundoa:
@kyleknighted both updates are coming from different validations: one is checking if the URL being typed is allowed in `URLWhiteListInput`, the other validates the whole form, which also checks if the URL is allowed, and adds some text alongside the error (that's why the text jumps).

Since not all forms are validated this way, I don't think we can easily circumvent both validations from running in this particular case. I think the easiest way may be to add a flag to let deactivate the `URLWhiteListInput` validation if the field is validated externally. At the same time, I am not sure if that will be creating even more complexity that is masking some other issues in our architecture.

This PR already overstretched it's initial purpose, so I think it's better to tackle the underlying validation issue separately. I will fill an issue for it.

Thank you for the reviews and all the testing!

PR #11413, opened 2021-10-04
Fixed Race Condition with CP Installation and Caches/LUTs

janheise:
**Note:** This PR needs a backport to `4.2`.

## Description
The call that loads all tables linked to a cache has been placed near the actual recreation of these tables to make sure that the contents of that variable are current.

## Motivation and Context
For the Watchlist functionality, a migration script had been created that imports and installs a ContentPack. After installation, 3 keys are to be created in a lookup table. This fails sometimes due to a possible race condition that occurs that saving the cache and the lookup table during import fires an update event on the EventBus but due to the ordering of the code and the runtime during the cache recreation, the lookup table never gets started properly.

## Types of changes

## Checklist:


dennisoelkers (approved):
:+1:

PR #11424, opened 2021-10-05
Replacing `(Action|Store|Combined)Provider` with `singleton`.

dennisoelkers:
## Motivation and Context

Prior to this PR, we were using a mechanism to make sure that stores
have only a single instance across core & plugins. This was implemented
through the `(Action|Store|Combined)Provider` modules. This mechanism
had the following drawbacks:

  - The regular `import` mechanism was circumvented, leading to IDEs not being able to navigate to the store implementations easily
  - All stores/actions had to be retrieved through a call to a function of these modules and must not be imported directly. This has proven to be error-prone lately, because a single type import could instantiate a second store.
  - Importing stores/actions this way broke their typing
  - Mocking actions/stores in tests was even harder, because in many cases a `(Store|Action|Combined)Provider` instance had to be mocked too.
  - Dependency tracking, esp. tree shaking is not possible

Due to this, this PR is now replacing the centralisation of actions/stores with a decentralised approach that is being used in the views subsystem for quite some time, which breaks out the logic of the providers (caching instances under the `window` object with a unique identifier) but wraps individual actions/stores in place, so they can be `import`ed like normal dependencies. This avoids all of the aforementioned drawbacks. This was implemented in multiple steps:

  - Actions/stores were combined into single files (with some exemptions) so we cannot import only one of them
  - Actions/stores are wrapped with `singletonActions`/`singletonStore` call respectively
  - All usages of `*Provider.get*` are replaced with direct imports
  - Test mocks are cleaned up
  - `*Provider` are removed
  - The `builtins` asset is removed
  - The `registerBuiltinStores` function (and its call) are removed

The benefit of this endeavor is that we can restore many of the benefits of plain imports, increase typing and simplify testing in a lot of places. For testing, we can even consider providing default mocks for commonly used stores and/or even enable auto mocking.

## Types of changes

## Checklist:


kyleknighted (approved):
code changes look good and can't seem to find any broken areas of the app! thanks for this dennis!

PR #11443, opened 2021-10-08
Adding ESLint rule to prefer type-only imports.

dennisoelkers:
## Motivation and Context

This PR is adding an ESLint rule which configures it to prefer type-level imports. This means that any (full, without the `type` modifier) import which only imports types from the module, will result in an (auto-fixable) error. All type-only imports will be changed to have the `type` modifier. Type-only imports are generally preferrablle, as they stripped out during bundling.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me, works like a charm.

PR #11526, opened 2021-10-26
Move migrations into a separate command

mpfz0r:
Migrations used to be executed as a periodical.
Having them run by a separate command allows us
more flexibility and robustness in scenarios when
a cluster needs to be bootstrapped or upgraded automatically.
E.g. Kubernetes

While it would be nice to reduce the dependencies for the
MigrateCmd to a minimum, we currently pull in most of the server
dependecies and only remove a few things we don't want to run
together with the migration:

 - Journal
 - Services
 - InputBufferImpl
 - pid file


thll (approved):
LGTM

PR #11535, opened 2021-10-27
Avoid creating compound type for identical physical types but diff. properties

dennisoelkers:
## Motivation and Context

**Note:** This needs to be backported to `4.1` and `4.2`.

Prior to this change, a physical field type of `keyword` and a physical type of `text` were being merged into a logical field type of `string` with a `compound` property. This is not correct, as it is not really a compound type. All we need to do in this case is create the set of shared properties for identical logical types.

This is fixing an issue where a user cannot aggregate on a field which is a `keyword` in one index and a `text` field in another.

## Types of changes

## Checklist:


janheise (approved):
Looks good!

todvora (approved):
Looks good!

PR #11542, opened 2021-10-28
Adding a `latest value` metric.

### LINKED ISSUE #8296: "Last value" widget
## What?
Add the ability to show a single number (or other value) from a search result in a widget.
If the value is a number the widget should also indicate whether the value increased or decreased with respect to the penultimate value.

## Why?
Gauge-like inputs, that are ingested periodically but with low volume are difficult to visualize in Graylog today.
Consider the case when ingesting the number of open threat detections from a service provider API into Graylog.
On a dashboard, you want to display the last valid number in the dashboard's (or widget's) timeframe.

### Query
I believe the data source could be as simple as a messages search type with a descending timestamp sort and a limit of 2 (1 for the last value and 1 for the trending).

## Your Environment

* Graylog Version: 3.3

dennisoelkers:
## Motivation and Context

This PR is adding a `latest value` metric which always shows the latest value for a given field. Bucketing information will be respected, so this can also be the most recent value inside a bucket which is created through pivoting. This metric is very helpful if only non-aggregated, discrete values are of interest, e.g. the most recent status from an asset or threat db, the latest status of a network interface or service.

Trending is not yet supported, but will follow in a future PR.

Fixes #8296.

## Types of changes

## Checklist:


graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/views/searchtypes/pivot/series/ESLatestHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public class ESLatestHandler extends ESPivotSeriesSpecHandler<Latest, TopHitsAggregation> {
+    @Nonnull
+    @Override
+    public Optional<AggregationBuilder> doCreateAggregation(String name, Pivot pivot, Latest latestSpec, ESPivot searchTypeHandler, ESGeneratedQueryContext queryContext) {
+        final TopHitsAggregationBuilder latest = AggregationBuilders.topHits(name).size(1).sort(SortBuilders.fieldSort("timestamp").order(SortOrder.DESC));
+        record(queryContext, pivot, latestSpec, name, MaxAggregation.class);
```
janheise, line 43: probably `TopHits` instead of `MaxAggregation` => see `ESLatestHandler` in ES7 package

janheise (approved):
LGTM!

PR #11562, opened 2021-11-02
Refactor index field type polling

thll:
Refactoring of the index field type poller periodical.

The goal is that we can stop running the periodical on one node and start it on another node. In the previous version, the periodical was starting and maintaining separate scheduled threads, which would keep running, even if we would stop executing the periodical.

After the refactoring, the periodical only maintains the previous execution of each individual poll to determine if an index set needs polling at any given time. If we stop running the periodical, no polling will happen until we start running it again.

As suggested by @bernd

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -22,3 +22,10 @@ The following API endpoints have been removed in 4.3.
 | Endpoint                                    | Description                 |
 | ------------------------------------------- | --------------------------- |
 | `PUT /example/placeholder`                  | TODO placeholder comment    |
+
+## Configuration File Changes
+
+The following configuration option as been removed: `index_field_type_periodical_interval`.
+
+It has been replaced with a new configuration option which allows users to tweak the default full refresh interval for
+index field type information: `index_field_type_periodical_full_refresh_interval`.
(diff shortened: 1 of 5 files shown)


thll:
@bernd I'm a bit unsure about the frequency of "initializing" the field types in `initializeFieldTypes`. If indexes are added to an index set which are not active write indexes, we will miss adding the field types to those indexes until a node gets restarted.

Can we ever run into this situation? It looks to me as if this could be the case in an archive restore. With the old version, we would "fix" this situation after each periodical run every hour (by default). Do you think we need something like this or should we trigger updating the field types after an archive restore explicitly?

thll:
@mpfz0r mentioned, that we need some kind of periodic cleanup task, because indexes might be deleted in elasticsearch triggered externally. This is actually happening in some setups.

[HS #575331695]

bernd:
> @mpfz0r mentioned, that we need some kind of periodic cleanup task, because indexes might be deleted in elasticsearch triggered externally. This is actually happening in some setups.
> 

@thll @mpfz0r I agree, I think we need to execute `initializeFieldTypes` on a regular basis to do cleanup and to add new indices that we might have missed. I guess by default we can do it every minute? We should probably make this configurable to be able to mitigate any issues this might cause on big clusters. (I don't think it will, but well :smile:)

What do you think?

mpfz0r (approved):
Works and LGTM :+1:

PR #11570, opened 2021-11-03
Adhere to data_dir setting for netty libnative

### LINKED ISSUE #11543: The "io.netty.native.workdir" property is not honoring "data_dir" configuration
## Expected Behavior

The `io.netty.native.workdir` property should point to a `libnative` directory inside the configured `data_dir`.

## Current Behavior

The `io.netty.native.workdir` property always points to `data/` and it will create the `data/libnative` directory in the current working directory.

## Possible Solution

The `setNettyNativeDefaults()` method is called in `ServerBootstrap#beforeStart(TLSProtocolsConfiguration)`. This method is called **before** we parse the server configuration file. That's why the native workdir always points to the default `data/` directory. This probably broke when we fixed the TLS configuration settings in the 4.2 development cycle. (did we backport this to 4.1 and 4.0? If so, we need to backport a fix to those versions as well)

It is currently easy to miss that `#beforeStart()` is called before the configuration has been parsed. I think we should introduce a new lifecycle method that is properly named and documented to make it clear that the configuration hasn't been parsed when this method is called. (e.g. `#beforeConfigParse()` or something similar) This should avoid problems like this one in the future.

1. Configure `data_dir` to a custom directory
2. Start the server
3. Observe a `data/libnative` directory created in the current working directory instead of the custom `data_dir`

## Your Environment

* Graylog Version: 4.3.0-SNAPSHOT

mpfz0r:
The `io.netty.native.workdir` property always points to `data/libnative`
because the configuration hasn't been parsed at that stage.

Explicitly parse a PathConfiguration bean and pass it into the
beforeStart() method.

Fixes #11543

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
+++ b/graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
@@ -54,6 +54,7 @@ import org.apache.logging.log4j.core.LoggerContext;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
 import org.graylog2.Configuration;
 import org.graylog2.bootstrap.commands.MigrateCmd;
+import org.graylog2.configuration.PathConfiguration;
 import org.graylog2.configuration.TLSProtocolsConfiguration;
 import org.graylog2.featureflag.FeatureFlags;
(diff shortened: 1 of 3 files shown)


bernd (approved):
Tested, works. Thank you! :+1:

PR #11584, opened 2021-11-05
Give up leader lock at server shutdown

mpfz0r:


graylog2-server/src/main/java/org/graylog2/cluster/leader/StaticLeaderElectionService.java
```diff
@@ -43,4 +43,9 @@ public boolean isLeader() {
         //noinspection deprecation
         return configuration.isMaster();
     }
+
+    @Override
+    public void giveUpLeader() {
+        // Nothing we can do here
```
thll, line 49: It won't have an effect when we only call this upon shutdown. But I guess the following would theoretically be the correct thing to do:

```suggestion
        configuration.setIsMaster(false);
```

thll (approved):
Looks good and works as expected 👍

Issue #11491, opened 2021-10-19
Query Input Assistance: suggestions for values


todvora:
A user needs to know the data model (not only fields and their types, but also common field values) when writing a query. Next step after https://github.com/Graylog2/graylog2-server/issues/11490 is to suggest some values for the fields

Example: `http_method` field and suggestions like `POST`, `GET` or `DELETE`. 

## What to suggest?
* Starting with empty value, recommend some common and most probable values
* Prefix suggestion: Starting with `PO` recommend `POST` as a value
* Recommend least probable values? (How? Many of single-occurence values)

Caution: all of it needs to be able to work across several indices!
If possible write the logic in a way that it can be reused for other places where suggestions would be helpful, like the correlation engine.

## Why?
Increasing user satisfaction by reducing time-to-value. Helping user to create a query without prior knowledge of typical field values.

## Pointers
*  https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-suggesters.html

Steps:

- [ ]  Define an API, start with the current Suggester API/Structure and try to extend that (link an API issue/PR here)
- [ ]  Extend Suggesters category in frontend regarding field values
- [ ]  Implement new suggesters in the backend (link an BE issue/PR here)

PR #11585, opened 2021-11-05
Field value suggestions

todvora:
This is the first part of the #11491 issue, implementing the backend part of the field value suggestions. There will be a separate frontend pull request.

## Description
Provide API to suggest most common values of a field, taking into the account the already typed-in prefix that user could enter. 

## Motivation and Context
see #11491

## How Has This Been Tested?
Added integration test

See comment below

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
@@ -19,19 +19,19 @@ package org.graylog.plugins.views;
 import io.restassured.response.ValidatableResponse;
 import io.restassured.specification.RequestSpecification;
 import org.graylog.testing.completebackend.GraylogBackend;
+import org.graylog.testing.containermatrix.ContainerVersions;
 import org.graylog.testing.containermatrix.annotations.ContainerMatrixTest;
 import org.graylog.testing.containermatrix.annotations.ContainerMatrixTestsConfiguration;
 import org.graylog.testing.utils.GelfInputUtils;
 import org.graylog.testing.utils.SearchUtils;
 
-import java.util.List;
-
 import static io.restassured.RestAssured.given;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.graylog.testing.completebackend.Lifecycle.CLASS;
+import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.core.IsEqual.equalTo;
 
-@ContainerMatrixTestsConfiguration(serverLifecycle = CLASS, extraPorts = SuggestionResourceIT.GELF_HTTP_PORT)
+@ContainerMatrixTestsConfiguration(serverLifecycle = CLASS, extraPorts = SuggestionResourceIT.GELF_HTTP_PORT, esVersions = {ContainerVersions.ES6, ContainerVersions.ES7})
 public class SuggestionResourceIT {
 
     static final int GELF_HTTP_PORT = 12201;
@@ -45,16 +45,16 @@ public class SuggestionResourceIT {
     }
 
     @ContainerMatrixTest
-    void testMinimalisticRequest() {
+    void testMinimalRequest() {
         int mappedPort = sut.mappedPortFor(GELF_HTTP_PORT);
         GelfInputUtils.createGelfHttpInput(mappedPort, GELF_HTTP_PORT, requestSpec);
         GelfInputUtils.postMessage(mappedPort,
-                "{\"short_message\":\"Hello there\", \"host\":\"example.org\", \"facility\":\"test\"}",
+                "{\"short_message\":\"SuggestionResourceIT#testMinimalisticRequest\", \"host\":\"example.org\", \"facility\":\"test\"}",
                 requestSpec);
 
         // mainly because of the waiting logic
-        final List<String> strings = SearchUtils.searchForAllMessages(requestSpec);
-        assertThat(strings.size()).isEqualTo(1);
+        final boolean messagePresent = SearchUtils.searchForMessage(requestSpec, "SuggestionResourceIT#testMinimalisticRequest");
+        assertThat(messagePresent).isTrue();
 
         final ValidatableResponse validatableResponse = given()
                 .spec(requestSpec)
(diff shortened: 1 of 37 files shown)


todvora:
image

image

full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
```diff
@@ -0,0 +1,97 @@
...
+import static io.restassured.RestAssured.given;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.graylog.testing.completebackend.Lifecycle.CLASS;
+import static org.graylog.testing.containermatrix.SearchServer.ES6;
+import static org.graylog.testing.containermatrix.SearchServer.ES7;
+import static org.graylog.testing.containermatrix.SearchServer.OS1;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.core.IsEqual.equalTo;
+
+@ContainerMatrixTestsConfiguration(serverLifecycle = CLASS, extraPorts = SuggestionResourceIT.GELF_HTTP_PORT, searchVersions = {ES6, ES7, OS1})
```
janheise, line 38: change to Lifecyle.VM? also remove searchVersions (all ist default), GELF port is added as a default to, but for transparency, it could be left in

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ElasticsearchClient.java
```diff
@@ -178,4 +189,22 @@ public static RequestOptions withTimeout(RequestOptions requestOptions, Duration
...
+        if (ex.getCause() != null) {
+            final Throwable[] suppressed = ex.getCause().getSuppressed();
+            if (suppressed.length > 0) {
+                final Throwable realCause = suppressed[0];
+                if (realCause instanceof ResponseException) {
+                    try {
+                        final ResponseError err = objectMapper.readValue(((ResponseException) realCause).getResponse().getEntity().getContent(), ResponseError.class);
+                        return Optional.of(err);
+                    } catch (IOException ioe) {
+                        //LOG.warn("Failed to parse exception", e);
```
janheise, line 203: If the exception can be dropped, I would at least write a comment hat suggests so.

graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/testing/ElasticsearchInstanceES7.java
```diff
@@ -47,7 +49,7 @@
     protected ElasticsearchInstanceES7(String image, SearchVersion version, Network network) {
         super(image, version, network);
         this.restHighLevelClient = buildRestClient();
-        this.elasticsearchClient = new ElasticsearchClient(this.restHighLevelClient, false);
+        this.elasticsearchClient = new ElasticsearchClient(this.restHighLevelClient, false, new ObjectMapperProvider().get());
```
janheise, line 52: not 100% sure. maybe this can be injected?

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SuggestionsResource.java
```diff
@@ -0,0 +1,114 @@
...
+                .ifPresent(suggestionsBuilder::error);
+
+        return suggestionsBuilder.build();
+    }
+
+    private Set<String> adaptStreams(Set<String> streams) {
+        if (streams == null || streams.isEmpty()) {
+            return loadAllAllowedStreamsForUser();
+        } else {
+            // TODO: is it ok to filter out a stream that's not accessible or should we throw an exception?
```
janheise, line 90: additionally: not sure if this filter should be done in the service?

graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/testing/ElasticsearchInstanceES7.java
todvora: I am not sure. it's a test class, all instances of everything are manually created around this code. So I did the same. Any better approach, @dennisoelkers? Thx!

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ElasticsearchClient.java
todvora: returned it back, no idea why I commented it out in the first place :see_no_evil:

full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
todvora: Done, good catch, thx!

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SuggestionsResource.java
todvora: this will need some unified handling. I'll prepare a separate PR for the stream loading and filtering, as this is a repeated pattern that gets copy-pasted around.

graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/QueryBackend.java
```diff
@@ -149,4 +151,6 @@ default void validateQueryTimeRange(Query query, SearchConfig config) {
                     throw new IllegalTimeRangeException("Search out of allowed time range limit");
                 });
     }
+
+    SuggestionResponse suggest(SuggestionRequest req);
```
dennisoelkers, line 155: It looks to me as if the feature and its implementations are distinct enough so that it does not need or benefit from coupling it to the `QueryBackend`. IMO it would benefit from being isolated in a separate class/interface.

graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
```diff
@@ -0,0 +1,55 @@
...
+    public abstract String field();
+
+    public abstract String input();
+
+    public abstract List<SuggestionEntry> suggestions();
+
+    public abstract Optional<SuggestionError> suggestionError();
+
+
+    public static Builder builder(final String field, final String input) {
```
dennisoelkers, line 39: I would suggest using specific factory methods instead of a builder when constructing instances. E.g. we could have:

```suggestion
    public static SuggestionResponse forSuggestions(final String field, final String input, final List<SuggestionEntry> suggestions) { ... }
    public static SuggestionResponse forError(final String field, final String input, final SuggestionError error) { ... }
```

We do not really need the flexibility of a builder here, because we have two different cases for creation. Modelling them like this makes it more obvious what the possible states could be.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SuggestionsResource.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Inject
+    public SuggestionsResource(PermittedStreams permittedStreams, QueryEngine queryEngine) {
+        this.permittedStreams = permittedStreams;
+        this.queryEngine = queryEngine;
+    }
+
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation("Validate a search query")
```
dennisoelkers, line 66: This is left over from copying the method?

dennisoelkers (changes requested):
Looks good, works fine, added some suggestions inline!

graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/QueryBackend.java
todvora: Good point, separated the query backend and suggestions

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SuggestionsResource.java
todvora: Indeed, thanks, fixed!

graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
todvora: Make sense, done, thanks!

dennisoelkers (approved):
Perfect! :handshake:

PR #11587, opened 2021-11-05
Downgrade from leader if lock can't be renewed

thll:
Downgrades a node when the leader lock can't be acquired for a whole lock TTL interval because then we are risking the lock being taken by another node.

Also adds a grace period to pause lock acquiring attempts for a while after a forced downgrade happened.


mpfz0r (approved):
LGTM

PR #11586, opened 2021-11-05
Make AutomaticLeaderElectionService timeouts configurable

mpfz0r:
Also introduce new JavaDurationConverter for JADConfig so
we can start moving towards using Java8 Durations
in our code.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2797

PR #11597, opened 2021-11-08
Update MongoDBContainer to version 4.0

mpfz0r:
MongoDB 3.6 is EOL since April 21

Adopt EnforceUnqieGrokPatternsTest to new more detailed Exception
message.

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/testing/mongodb/MongoDBContainer.java
+++ b/graylog2-server/src/test/java/org/graylog/testing/mongodb/MongoDBContainer.java
@@ -30,7 +30,7 @@ import static java.util.Objects.requireNonNull;
  */
 public class MongoDBContainer extends GenericContainer<MongoDBContainer> {
     public static final String DEFAULT_IMAGE = "mongo";
-    public static final String DEFAULT_VERSION = "3.6";
+    public static final String DEFAULT_VERSION = "4.4";
     public static final int MONGODB_PORT = 27017;
     public static final String NETWORK_ALIAS = "mongodb";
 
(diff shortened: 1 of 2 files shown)


graylog2-server/src/test/java/org/graylog/testing/mongodb/MongoDBContainer.java
```diff
@@ -30,7 +30,7 @@
  */
 public class MongoDBContainer extends GenericContainer<MongoDBContainer> {
     public static final String DEFAULT_IMAGE = "mongo";
-    public static final String DEFAULT_VERSION = "3.6";
+    public static final String DEFAULT_VERSION = "4.4";
```
bernd, line 33: Can we make this `4.0`? We still support MongoDB 4.0 so we should use the oldest supported version, I think.

bernd (changes requested):

graylog2-server/src/test/java/org/graylog/testing/mongodb/MongoDBContainer.java
mpfz0r: Fine with me. Testing against newer versions also has its benefits though.
For the leader election, we do this: https://github.com/Graylog2/graylog2-server/pull/11586/commits/94adff71728e30bb5f54a7a106b190efe0bc4ca4

bernd (approved):

bernd (changes requested):
Hmm, the tests fail again. :smile:

bernd (approved):

PR #11600, opened 2021-11-09
Update elasticsearch client version from 5 to 6

todvora:
This PR updates the shaded ES client for the ES6 storage module from ES5 to ES6. 

## Motivation and Context
We don't support ES5 anymore and the client should be ideally in the same version as the storage module. The current implementation still used ES5, because there are no significant changes. But it would be good to have the latest version there to minimize any possible technical debt and use the latest releases to avoid any bugs that may be already fixed in the new versions. 

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2925

## How Has This Been Tested?
There is already an extensive set of test for ES6 storage module, only 3 tests adapted for the slightly changed default structure of the generated queries, all other are passing. 

## Types of changes

## Checklist:


graylog-storage-elasticsearch6/src/test/resources/org/graylog/storage/elasticsearch6/views/everySearchTypeGeneratesOneESQuery.request.ndjson
```diff
@@ -1,4 +1,4 @@
...
+{"from":0,"size":0,"query":{"bool":{"must":[{"bool":{"filter":[{"match_all":{"boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},{"range":{"timestamp":{"from":"2018-08-23 08:02:00.247","to":"2018-08-23 08:07:00.252","include_lower":true,"include_upper":false,"boost":1.0}}},{"terms":{"streams":[],"boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},"aggregations":{"pivot1-series-avg(field1)":{"avg":{"field":"field1"}},"timestamp-min":{"min":{"field":"timestamp"}},"timestamp-max":{"max":{"field":"timestamp"}}}}
```
todvora, line 2: It's really hard to see the difference, but the change is rather minor, only ` "disable_coord": false` added automatically when the query is serialized in the ES6 version. Similar minor and unrelated changes are in the other 2 test resource files.

dennisoelkers (approved):
Looks good from my end! Good work! :)

janheise (approved):
LGTM! Tests run fine locally, also tested manually some of the changes where the classes changed between 5->6

PR #11602, opened 2021-11-09
Improve submitProcessingErrors() checking order

mpfz0r:
Move the checks that don't require asking a service to the beginning.

This way, we only check the configuration for messages that have
errors and can be handled.


mpfz0r:
@bernd I think this might be a canditate for 4.2.x

diesaulov (approved):
LGTM

PR #11620, opened 2021-11-11
Option to configure sharing of entities

### LINKED ISSUE #11112: Add option to disable sharing with everybody or single users
In a Graylog setup with many users and teams, where teams ensure users only get access to data they are allowed to see, the options to share an entity with everybody or with single users are not helpful. Selecting one of those may end up in users getting access to an entity but not other entities it depends on (as described in https://github.com/Graylog2/graylog2-server/issues/9339), and it results in users having long lists of entities they can't access and additional administration complexity.

## What?
Add configuration options to set whether sharing with everyone and sharing with users should be available or not.

## Why?
In some situations sharing with everyone or with single users is not helpful. See HS-507433390.

## Your Environment
* Graylog Version: 4.1.2

patrickmann:
Add configuration options to set whether sharing with everyone and sharing with users should be available or not.

We introduce a new cluster configuration class `org.graylog2.users.UserAndTeamsConfig`, which is accessible via the cluster configuration API `api/system/cluster_config/{config_class}`.
It exposes these settings:
- allow_sharing_with_everyone
- allow_sharing_with_users

See linked issue for more details on the use case.

Resolves #11112

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2939


graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
```diff
@@ -80,12 +93,20 @@ public DefaultGranteeService(UserService userService, GRNRegistry grnRegistry, G
                 .collect(ImmutableSet.toImmutableSet());
     }
 
+    private ImmutableSet<AvailableGrantee> getGlobalGrantees() {
+        final ImmutableSet.Builder<AvailableGrantee> builder = ImmutableSet.<AvailableGrantee>builder();
+        final Optional<UserAndTeamsConfig> config = Optional.ofNullable(clusterConfigService.get(UserAndTeamsConfig.class));
+        if (config.isPresent() && config.get().sharingWithEveryone()) {
```
mpfz0r, line 99: This would not add the `everyone` grantee if the clusterconfig has never been saved.
(but see above suggestion on how to avoid that problem)

graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
```diff
@@ -68,6 +76,11 @@ public DefaultGranteeService(UserService userService, GRNRegistry grnRegistry, G
     }
 
     private ImmutableSet<AvailableGrantee> getAvailableUserGrantees(User sharingUser) {
+        final Optional<UserAndTeamsConfig> config = Optional.ofNullable(clusterConfigService.get(UserAndTeamsConfig.class));
```
mpfz0r, line 79: I'd reccomend using 
`clusterConfigService.getOrDefault(UserAndTeamsConfig.class, UserAndTeamsConfig.create(true, true));`

or even better create a constant:
`clusterConfigService.getOrDefault(UserAndTeamsConfig.class, UserAndTeamsConfig.DEFAULT_VALUES);`

graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
```diff
@@ -80,12 +93,20 @@ public DefaultGranteeService(UserService userService, GRNRegistry grnRegistry, G
                 .collect(ImmutableSet.toImmutableSet());
     }
 
+    private ImmutableSet<AvailableGrantee> getGlobalGrantees() {
+        final ImmutableSet.Builder<AvailableGrantee> builder = ImmutableSet.<AvailableGrantee>builder();
+        final Optional<UserAndTeamsConfig> config = Optional.ofNullable(clusterConfigService.get(UserAndTeamsConfig.class));
```
mpfz0r, line 98: same here. `getOrDefault()` makes things easier

graylog2-server/src/main/java/org/graylog2/users/UserAndTeamsConfig.java
```diff
@@ -0,0 +1,41 @@
...
+public abstract class UserAndTeamsConfig {
+    @JsonProperty
+    public abstract boolean sharingWithEveryone();
+
+    @JsonProperty
+    public abstract boolean sharingWithUsers();
+
+    @JsonCreator
+    public static UserAndTeamsConfig create(
+            @JsonProperty("allow_sharing_with_everyone") boolean sharingWithEveryone,
```
mpfz0r, line 37: These property names are not the same ones that are used when serializing.
I think this only works because of JsonAutoDetect.
But better give everything a consistent naming.

graylog2-web-interface/src/components/configurations/PermissionsConfig.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const _resetConfig = () => {
+    setShowModal(false);
+  };
+
+  return (
+    <div>
+      <h2>Permissions Config</h2>
+      <p>Configuration for Permissions</p>
```
mpfz0r, line 65: I think this should have a more descriptive help text.
E.g.  `These settings can be used to control which entity sharing options are available`

mpfz0r (changes requested):
See comments

graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
patrickmann: We are initializing in the frontend, when the form is opened. But it's a good idea to have defined default behavior in the backend.

graylog2-server/src/main/java/org/graylog2/users/UserAndTeamsConfig.java
patrickmann: Weird that this even works. I'd rather it just failed. Good catch.

graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
patrickmann: Yes, that is a nice idiom

graylog2-web-interface/src/components/configurations/PermissionsConfig.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const _resetConfig = () => {
+    setShowModal(false);
+  };
+
+  return (
+    <div>
+      <h2>Permissions Config</h2>
+      <p>These settings can be used to control which entity sharing options are available</p>
```
edmundoa, line 65: We are missing a dot here:
```suggestion
      <p>These settings can be used to control which entity sharing options are available.</p>
```

graylog2-web-interface/src/components/configurations/PermissionsConfig.tsx
```diff
@@ -0,0 +1,138 @@
...
+                      <div>
+                        <Row>
+                          <Col sm={12}>
+                            <FormikInput type="checkbox"
+                                         name="allow_sharing_with_everyone"
+                                         id="shareWithEveryone"
+                                         label={(
+                                           <LabelSpan>Share with everyone</LabelSpan>
+                                         )} />
+                            <InputDescription help={<>Share with everyone option will be available</>} />
```
edmundoa, line 106: This description is somehow confusing to me: the option will not be available if I disable the toggle. I suggest changing this to: "Control whether it is possible to share with everyone".

edmundoa: By the way, would you mind changing the types for the `InputDescription` props from `ReactElement` to `ReactNode`? That will allow us to pass a `string` directly.

graylog2-web-interface/src/components/configurations/PermissionsConfig.tsx
```diff
@@ -0,0 +1,138 @@
...
+                          </Col>
+                          <Col sm={12}>
+                            <FormikInput type="checkbox"
+                                         name="allow_sharing_with_users"
+                                         id="shareWithUsers"
+                                         label={(
+                                           <LabelSpan>Share with users</LabelSpan>
+                                         )} />
+                            <InputDescription help={<>Share with users options will be available.</>} />
```
edmundoa, line 115: Same as above, I would suggest something like "Control whether it is possible to share with single users".

edmundoa (changes requested):
It looks good and works as expected, thank you! I added a few changes/suggestions inline.

graylog2-web-interface/src/stores/configurations/ConfigurationsStore.ts
```diff
@@ -140,6 +145,24 @@ export const ConfigurationsStore = singletonStore(
...
     },
 
+    listPermissionsConfig(configType) {
+      const promise = fetch('GET', this._url(`/${configType}`)).then((response: PermissionsConfigType) => {
+        this.configuration = {
+          ...this.configuration,
+          [configType]: response || {
+            allow_sharing_with_everyone: true,
+            allow_sharing_with_users: true,
+          },
```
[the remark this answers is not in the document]
patrickmann: Actually Marco made the same suggestion and I have already added defaults to the backend. 
@ousmaneo we should be OK removing this from the frontend.

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -72,6 +75,7 @@ const ConfigurationsPage = () => {
       ConfigurationsActions.list(SIDECAR_CONFIG),
       ConfigurationsActions.list(EVENTS_CONFIG),
     ];
+    promises.push(ConfigurationsActions.listPermissionsConfig(PERMISSIONS_CONFIG));
```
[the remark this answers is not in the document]
ousmaneo: Agree to move it up. I put it there in case we needed to check permission for it.

edmundoa (changes requested):
Would you mind merging the changes in master? I think we need them to make the build pass again. Besides that, it all looks good from the front end side, only this small nitpick is missing: https://github.com/Graylog2/graylog2-server/pull/11620#discussion_r751458821.

graylog2-web-interface/src/stores/configurations/ConfigurationsStore.ts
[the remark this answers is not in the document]
mpfz0r: Well, unfortunately the clusterconfigresource does not return those defaults.
So this actually breaks the UI (the permissions config doesn't show up) when there's no UserAndTeamsConfig in the cluster config collection.

Either we introduce a separate resource that provides the default as fallback,
or we have to keep the duplicate default values in the UI.

mpfz0r (changes requested):
backend looks good now, but this got broken:
https://github.com/Graylog2/graylog2-server/pull/11620#discussion_r753164617

graylog2-web-interface/src/stores/configurations/ConfigurationsStore.ts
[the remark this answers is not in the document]
edmundoa: I don't think providing these defaults is worth adding an extra resource to the backend, so I think we can keep the defaults in the UI. The main issue I can think of doing it this way is that we can end up with different defaults in both back end and front end, but I _guess_ we'll never change those defaults once they are set.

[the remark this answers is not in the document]
boosty: > The main issue I can think of doing it this way is that we can end up with different defaults in both back end and front end

If we go this way, I would at least add a comment in both places to tell that both must be changed together.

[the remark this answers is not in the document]
ousmaneo: Added the default back and a comment saying that it should be the same in the backend.

mpfz0r (approved):

edmundoa (approved):

edmundoa:
The front end looks good to me, although the build is still broken. If the issue is related to this PR, please fix that before merging.

jens-graylog:
Tested and verified.
method:
- setup cluster
- select config page
- disable sharing with everyone
expected result:
unauthorized users are unable to access data
result: works as intended

PR #11628, opened 2021-11-12
Disable extractor in action menu for forwarder messages

ousmaneo:
These changes add isLocalNode to `AdditionnalContext`  and add the value to the validation test to enable or disable the create-extractor ActionContext menu. Since the isLocalNode value is Available in the `MessageDetail` component, the `FormatReceiveBy` component was modified to receive it as a prop instead of computing it.

fix Graylog2/graylog-plugin-cloud#390
## Motivation and Context

When using the forwarder clicking on a field value and choosing `create extractor` throw an error. More generally extractor feature is not available when using the Forwarder.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/hooks/useIsLocalNode.ts
+++ b/graylog2-web-interface/src/hooks/useIsLocalNode.ts
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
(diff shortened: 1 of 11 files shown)


dennisoelkers (approved):
Looks good to me! Some minor suggestion:

 - `useIsLocalNode` is used only in the `views` subsystem. We should maybe move it to some place in `src/views`?
 - `useIsLocalNode` takes a `Message` as first argument. It would be easier to use for other purposes if it would just take a node id as an argument.

ousmaneo:
Thank you for the review, I applied the suggestions. 👍

edmundoa (approved):
Looks good to me, thank you! 👍

PR #11635, opened 2021-11-15
Use view summary when sharing/exporting saved searches/dashboards.

dennisoelkers:
## Motivation and Context

When sharing or exporting saved searches/dashboards, a list of exportable entities is prepared. For
these, the entity excerpt is being used.

Prior to this change, the excerpt for saved searches/dashboards was the
same as for the complete entity (`ViewDTO`). This lead to the complete
view being deserialized, although only id and title were being used. It
also lead to the complete request failing, when only a single entity
cannot be deserialized properly.

With this change, we are now using the `ViewSummaryDTO` class instead.
This allows us to deserialize only a minimum set of fields in order to
present a list of sharable/exportable entitites and avoids breakage in
case a single entity cannot be deserialized completely.

## Types of changes

## Checklist:


janheise (approved):
LGTM!

linuspahl (approved):
Looks good to me!

Issue #9021, opened 2020-09-23
JSON parsing: add shallow parsing up to a certain path depth


kroepke:
## What?
Add JSON parsing function(s) that support shallow parsing a deeply nested JSON document, returning all the key/value pairs up to a certain depth (optionally starting at a JSONPath expression).
The unparsed portion of the subtree would be returned as a string, instead of being parsed, allowing either storing it verbatim or parsing it in a later stage.

## Why?
When dealing with deeply nested documents, especially from sources like AWS services, the number of unique paths in that document can easily be in the tens of thousands, leading to too many fields in an index.
Most of the time those fields aren't important for searching but need to be displayed later on.
Given the variable nature of these documents, it's often hard to impossible to know all the interesting paths beforehand, so parsing just the first one or two levels of the JSON document helps in navigating it, without catastrophically exploding the number of keys.

Removing the extra keys is often not possible, either, because their names are not known ahead of time, so they cannot be easily removed later on.

## Your Environment

* Graylog Version: 3.3.5

boosty:
Related: #8706

pbrzica:
Is there any ETA on this issue?
Would be great to see this included in Graylog 4.

patrickmann:
Could somebody clarify the expected return value with regard to unparsed content below the depth cut-off?
E.g. given the JSON below, what would be the expected result for depth=2? The resulting JSON tree would include only fields `k0`, `k1`, `k2` and not `k11`, `k21`; but how should the unparsed content be returned? These could be fragments from a bunch of different places within the tree and not particularly useful. 

I suggest we just ignore anything below the cut-off. If there are additional requirements in future, we can always extend the function or provide a new function to address that.

```
{
  "k0": "v0",
  "obj1": {
    "k1": "v1",
    "obj11": {
      "k11": "v11"
    }
  },
  "obj2": {
    "k2": "v2",
    "obj21": {
      "k21": "v21"
    }
  }
}
```

boosty:
Hi @patrickmann, I talked about this with @kroepke and @bernd. We like your proposal of ignore anything below the cut-off 👍

boosty:
Assigned @dulanism for documentation (once this is done)

PR #11640, opened 2021-11-16
JSON parsing: add shallow parsing up to a certain path depth

patrickmann:
Adds a new optional parameter to the `parse_json` pipeline function. The parameter specifies the maximum parse depth - we ignore any nodes that are more deeply nested.

See related issues for details and use case.

Resolves Graylog2/graylog2-server#9021

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
@@ -17,19 +17,10 @@
 package org.graylog.plugins.pipelineprocessor.functions.json;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-
-import static com.google.common.collect.Lists.newArrayList;
 
 public class JsonUtils {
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
```diff
@@ -0,0 +1,77 @@
...
+                    }
+                }
+                index++;
+            }
+            // avoid ConcurrentModificationException by delaying removal
+            int offset = 0;
+            for (int i : toRemove) {
+                ((ArrayNode) root).remove(i - offset++);
+            }
+        }
```
thll, line 74: I'm not sure, if I'm missing something but I have the feeling we could maybe simplify this. Could you please check if the following condensed version of the above would work as well?

```suggestion
        final Iterator<JsonNode> elements = root.elements();
        while (elements.hasNext()) {
            final JsonNode node = elements.next();
            if (node.isContainerNode()) {
                if (depth >= maxDepth) {
                    elements.remove();
                } else {
                    deleteBelow(node, maxDepth, depth + 1);
                }
            }
        }
```

thll (changes requested):
Tested it and it worked well.

Didn't find any issues with the code but I've left some suggestions nonetheless.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
patrickmann: You are quite right - I was so focused on the custom removal methods provided by JsonObject and JsonArray that I missed the iterator's remove(). I was wondering why there was nothing at the JsonNode level. :/

thll (approved):
LGTM

PR #11650, opened 2021-11-17
Fix migrate command for running with cloud plugin

mpfz0r:
The cloud plugin gets its Configuration dependency injected.
Duplicating the configuration and disabling the journal
did not affect the cloud plugin. Thus two MessageQueue implementations
were bound.

```
1) [Guice/BindingAlreadySet]: MessageQueueReader was bound multiple times.

Bound at:
1  : PluginModule.bindMessageQueueImplementation(PluginModule.java:376)
      \_ installed by: ServerBindings -> MessageQueueModule
2  : PluginModule.bindMessageQueueImplementation(PluginModule.java:376)
      \_ installed by: PluginBindings -> GraylogCloudModule

```

Simply disabling the journal on the configuration bean ensures
that both modules receive the same setting.

Also, remove uneeded configuration field from MigrateCmd.


thll (approved):
LGTM

PR #11661, opened 2021-11-18
Moves simulation diff logic to backend

kingzacko1:
## Description
The diff of fields added, removed, and changed during pipeline simulation is now created on the back end as opposed to the front end 

## Motivation and Context
Related to Graylog2/graylog-plugin-enterprise#2880. Generating pre/post-processing diffs on the back end makes it easier to unit test. Also logically it just makes more sense to create the diff on the back end vs front end. 

## How Has This Been Tested?
Ran Graylog locally in my development environment, set up a pipeline to change, add, and remove fields from a message, and confirmed that the output on the front end is correct based on the pre/post processed message. Also tested hitting Illuminate sim from command line with help from Michael to ensure all added, removed, and changed fields return properly. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
@@ -55,7 +55,8 @@ public abstract class DecorationStats {
         return Sets.intersection(originalMessage().keySet(), decoratedMessage().keySet())
                 .stream()
                 .filter(key -> !originalMessage().get(key).equals(decoratedMessage().get(key)))
-                .collect(Collectors.toMap(Function.identity(), key -> decoratedMessage().get(key)));
+                .collect(Collectors.toMap(Function.identity(), key
+                        -> ChangedField.builder().before(originalMessage().get(key)).after(decoratedMessage().get(key)).build()));
     }
 
     @SuppressWarnings("unused")
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
@@ -53,9 +53,9 @@ public abstract class DecorationStats {
     @JsonProperty(FIELD_CHANGED_FIELDS)
     public Map<String, Object> changedFields() {
         return Sets.intersection(originalMessage().keySet(), decoratedMessage().keySet())
-            .stream()
-            .filter(key -> !originalMessage().get(key).equals(decoratedMessage().get(key)))
-            .collect(Collectors.toMap(Function.identity(), key -> originalMessage().get(key)));
+                .stream()
+                .filter(key -> !originalMessage().get(key).equals(decoratedMessage().get(key)))
+                .collect(Collectors.toMap(Function.identity(), key -> decoratedMessage().get(key)));
     }
 
     @SuppressWarnings("unused")
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/SimulatorResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/SimulatorResource.java
@@ -30,6 +30,7 @@ import org.graylog2.database.NotFoundException;
 import org.graylog2.plugin.Message;
 import org.graylog2.plugin.rest.PluginRestResource;
 import org.graylog2.plugin.streams.Stream;
+import org.graylog2.rest.models.messages.responses.DecorationStats;
 import org.graylog2.rest.models.messages.responses.ResultMessageSummary;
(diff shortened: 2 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
```diff
@@ -53,9 +53,10 @@
     @JsonProperty(FIELD_CHANGED_FIELDS)
     public Map<String, Object> changedFields() {
         return Sets.intersection(originalMessage().keySet(), decoratedMessage().keySet())
-            .stream()
-            .filter(key -> !originalMessage().get(key).equals(decoratedMessage().get(key)))
-            .collect(Collectors.toMap(Function.identity(), key -> originalMessage().get(key)));
+                .stream()
+                .filter(key -> !originalMessage().get(key).equals(decoratedMessage().get(key)))
```
roberto-graylog, line 57: This method's implementation, as well as the **addedFields**() and **removedFields**() seem out of place in a domain object.  Would these values be best supplied by the builder or an external utility class.

Further, could these methods benefit from unit tests?

roberto-graylog (approved):
I tested locally and the changes seem to perform the intended function.

graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
kingzacko1: If this object was larger and had other things inside of it I'd think the use of a builder or external utility class would make sense. Since those 3 methods are really the only use of this class and they aren't all that verbose, I don't see an issue with them left as is. Doesn't matter to me one way or the other if there happens to be a consensus on refactoring.

I can add some unit tests to validate they are created correctly though.

danotorrey:
@kingzacko1 I think this branch will need the latest `master` merged in to avoid some errors in the https://github.com/Graylog2/graylog-plugin-enterprise/blob/3952373b7f9ebe38d530f385c39cf0143f9a8b71/enterprise/src/main/java/org/graylog/plugins/security/shares/EnterpriseGranteeService.java#L29 file.

graylog2-server/src/main/java/org/graylog2/rest/models/messages/responses/DecorationStats.java
```diff
@@ -53,9 +53,10 @@
...
+                .collect(Collectors.toMap(Function.identity(), key
+                        -> ChangedField.builder().before(originalMessage().get(key)).after(decoratedMessage().get(key)).build()));
```
danotorrey, line 59: I like the approach to use the new `ChangedField` object to keep track of the original and new values. I think it fits well.

Question: I think this `DecoratorStats` class is used in other parts of Graylog. This like appears to change the map values from the original scalar message field values to instances of the `ChangedField` class. Is this compatible with the other usages of the `DecoratorStats` class? For example, https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog2/decorators/DecoratorProcessorImpl.java#L116 which is used as part of the [Search Decorators](https://docs.graylog.org/v1/docs/decorators) feature?

kingzacko1: @danotorrey I was unaware `DecoratorStats` was being used elsewhere. Since it was not being populated at all in normal pipeline simulator diff or the illuminate simulator diff, I just assumed it was safe to make changes to the class. I will check out the features you mentioned as well as confirm there is nowhere else that may be affected with the change. Thank you for pointing this out!

kingzacko1:
@danotorrey the ESLint warnings are on code that has been around prior to this PR. Don't know if we changed our ESLint settings but this was building fine before with the exact same code. The only thing the recent push did was rebase on top of master.

danotorrey:
Thanks @kingzacko1. I am reviewing this once more. Should have that done today.

danotorrey:
@kingzacko1 The linter errors appear unrelated to the changes in this PR, so I think those can be ignored. 

I had nearly finished the review yesterday. I noticed the additional force-push this morning. Just checking, did that include any changes, or was that just merging in latest master?

kingzacko1:
@danotorrey just me rabasing the existing changes onto latest master trying to get all of this to work locally again.

danotorrey (approved):
LGTM! Thanks a lot @kingzacko1!

During development and review, one aspect that we looked into collectively was the change of the `DecoratorStats.changed_fields` from a value map to a collection of `ChangedField` objects and whether or not this could break existing Decorators functionality in the UI. As far as we could could see from reviewing React components and the UI, we did not see any usages that would be broken. 

We could not test this definitively, since the "Show changes" decorator was broken before this PR (see https://github.com/Graylog2/graylog2-server/issues/11753), but we are fairly confident that this will not have an effect once that is fixed.

PR #11664, opened 2021-11-19
Improve submitProcessingErrors() checking order (#11602)

mpfz0r:
Move the checks that don't require asking a service to the beginning.

This way, we only check the configuration for messages that have
errors and can be handled.

(cherry picked from commit 05e354722cc3e465d7fb25d4c76409406d45bd18)


diesaulov (approved):
👍

PR #11666, opened 2021-11-19
Introducing DTOs for Search API, selecting version via content negotiation.

dennisoelkers:
### Decoupling through DTOs

The major goal of this PR is to separate the structures used to interact with the search execution API from the entities used internally. For this, the `Search` and `Query` structures are decoupled from the REST API by introducing `SearchDTO` & `QueryDTO` classes. This allows us to:

  - change the structure of the internally used entities while keeping the request format of the API stable
  - implement different versions of the API without having to change the underlying search engine used

Intentionally, not the complete class hierarchy is duplicated as DTOs.  This should be done when changes are implemented which affects structures which are deeper in the class hierarchy, but would involve a lot of unnecessary churn initially.

### API Versioning

In addition, a sample v2 of the search execution API is introduced to demonstrate the recommended way for search API versioning. 
The following aspects have been considered:

  - We need to find a sweet spot between full decoupling between endpoints/requests/responses and excessive overhead maintaining API versions
  - In most cases the API should evolve instead of being versioned:
    - Adding API endpoints should be done in a backwards-compatible manner
    - Removing API endpoints should happen when 
      - alternatives exist
      - after being deprecated for a timeframe allowing clients to migrate
      - only for releases which allow breaking changes
  - Versioning should happen when an endpoint is supposed to consume/produce multiple versions of the same data, varying in structure
  - Selecting the version should happen through content-negotiation, by specifying `Accept`/`Content-Type`-headers
    - This allows us easier iteration of independently-versioned resources
    - A single-token media-type should be used, in the following format:
      - `application/vnd.graylog.<entity>.v<version>+json`, where `<entity>` is a name for the entity/structure being consumed/produced (e.g. `search`) and `<version>` is the numeric version (e.g. `v2` or `v1.2.3`)
    - There should always be a default endpoint consuming/producing `application/json` to support legacy clients

The example v2 is introducing a breaking change in the query format, which deprecates the `filter` attribute and replaces it with a clearer, more domain-specific `streams` attribute instead.

Existing clients should not be affected. Using the v2 search api requires opt-in from the client by specifically using it for the `Accept`/`Content-Type`-headers.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/SearchJob.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/SearchJob.java
@@ -36,9 +36,6 @@ import java.util.concurrent.CompletableFuture;
 // execution must come before results, as it signals the overall "done" state
 @JsonPropertyOrder({"execution", "results"})
 public class SearchJob {
-    private static final Logger LOG = LoggerFactory.getLogger(SearchJob.class);
-    static final String FIELD_OWNER = "owner";
-
     @JsonProperty
     private final String id;
 
@@ -51,7 +48,7 @@ public class SearchJob {
     @JsonIgnore
     private CompletableFuture<Void> resultFuture;
 
-    private Map<String, CompletableFuture<QueryResult>> queryResults = Maps.newHashMap();
+    final private Map<String, CompletableFuture<QueryResult>> queryResults = Maps.newHashMap();
 
     @JsonProperty("errors")
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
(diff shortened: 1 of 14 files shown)


janheise (approved):
Looks good!

graylog2-server/src/main/java/org/graylog/plugins/views/search/SearchJob.java
```diff
@@ -51,7 +46,7 @@
     @JsonIgnore
     private CompletableFuture<Void> resultFuture;
 
-    private Map<String, CompletableFuture<QueryResult>> queryResults = Maps.newHashMap();
+    final private Map<String, CompletableFuture<QueryResult>> queryResults = Maps.newHashMap();
```
todvora, line 49: Should we change the order to private final?

todvora (approved):
Looks really good, thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/SearchJob.java
dennisoelkers: :white_check_mark:

PR #11534, opened 2021-10-27
Search query validation

todvora:
## Description

With this PR we are implementing a search query validation. In the UI we are displaying an indicator for the validation warnings and errors:

image

It is also possible to validate a query using the API endpoint `/search/validate`

Example payload:
```
{
"query": "(lorem:ipsum OR dolor:sit)"
}
```

Fixes: https://github.com/Graylog2/graylog2-server/issues/11490, https://github.com/Graylog2/graylog2-server/issues/11502, https://github.com/Graylog2/graylog2-server/issues/11503

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#2920

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #11681, opened 2021-11-23
Only display warning for time range in search bar, when range is invalid.

linuspahl:
## Motivation and Context

Before this change we displayed the warning when any validation error occurred.

## How Has This Been Tested?
Tests already exist.


linuspahl:
These changes have been merged together with https://github.com/Graylog2/graylog2-server/pull/11534.

PR #11683, opened 2021-11-23
Remove old blocking migration / reduce verbosity

mpfz0r:
- don't log migration execution by default
 - remove old migration which was used to calculate index ranges for the
   default stream's index for pre-2.2 installations.


thll (approved):
LGTM

PR #11692, opened 2021-11-26
Improve the JavaDurationConverter

mpfz0r:
The main problem with the previous implementation was, that
it would parse jadconfig type 'value+unit' fromats like `10s`

But reproduce the Duration as ISO8601 strings: `PT10S`

When dumping a graylog configuration, these values couldn't be parsed anymore.

 -> Change the converter to accept both formats which uses joda Period as conversion helper.
 -> Add unit test


thll (approved):
LGTM

PR #11695, opened 2021-11-26
 move RuleMetricsConfig to RulesPage

### LINKED ISSUE #11688: Debug metrics button missing on pipeline rules page
## Expected Behavior

The pipeline rules page should show the "Debug Metrics" button like this:

image

## Current Behavior

In 4.3.0-SNAPSHOT the button is gone so users are unable to enable/disable the debug metrics:

image

## Your Environment

* Graylog Version: 4.3.0-SNAPSHOT (rev 330bb8a)

ousmaneo:
These change move rulesMetrics config to RulesPage and fix the Debug metric button not being displayed.

## Motivation and Context

Before these changes The debug metric button on the Rules page was not being shown as described in the issue #11688

## Types of changes

## Checklist:

fix #11688


bernd (commented):
With this PR, the "Debug Metrics" button shows up again. :+1:

graylog2-web-interface/src/pages/RulesPage.tsx
```diff
@@ -68,6 +73,10 @@ const RulesPage = () => {
     }
   }, [isPaginationReady, pagination]);
 
+  useEffect(() => {
+    _loadRuleMetricData(setIsMetricDataLoading, setMetricsConfig);
```
edmundoa, line 77: When the user updates the debug metrics configuration using `RuleMetricsConfigContainer`, the change is not reflected until the page is reloaded. We should instead provide immediate feedback and re-fetch the metrics after the change has been done.

edmundoa (changes requested):
First of all, sorry that it took so long to take a look at this, I somehow missed it. Besides a comment the changes look good and fix the issue 👍 
We also need to rebase the code before merging, could you please take care of it? Thank you!

graylog2-web-interface/src/pages/RulesPage.tsx
```diff
@@ -56,8 +49,20 @@ const _loadData = (pagination: Pagination, setIsLoading, setPaginatedRules) => {
   });
 };
 
+const _loadRuleMetricData = (setIsloading, setMetricsConfig) => {
```
edmundoa, line 52: Small typo:
```suggestion
const _loadRuleMetricData = (setIsLoading, setMetricsConfig) => {
```

graylog2-web-interface/src/pages/RulesPage.tsx
```diff
@@ -95,6 +104,28 @@ const RulesPage = () => {
     };
   };
 
+  const onCloseMetricsConfig = () => {
+    _loadRuleMetricData(setIsMetricDataLoading, setMetricsConfig);
```
edmundoa, line 108: While this works, it also hides the debug metrics button while re-fetching the information. During the first load, not having the button is slightly odd but I think this issue becomes really prominent when working with the modal (since it will load every time it is closed), and specially in a slower connection.



https://user-images.githubusercontent.com/716185/146947279-978451f1-ac6e-4ffc-afe4-e38a4cc06b78.mp4


Why do we hide the button on loading in the first place? I think we should not hide it, either display the standard "Debug metrics" while we load, or show a loading state in the button.

edmundoa (changes requested):

graylog2-web-interface/src/pages/RulesPage.tsx
ousmaneo: Hi, thank you for testing. We were waiting for the metrics config data to show the button which also shows the debug config state. I made the change to show the disabled state when loading.

edmundoa (approved):
Looking great now, thank you! 👍

PR #11705, opened 2021-11-29
add queryString() to BackendQuery Interface

janheise:
## Description
I moved queryString() into BackendQuery Interface.

## Motivation and Context
This change removes a lot of necessary typecasts where they seem uncomfortable. 
The reasoning is, that for the foreseeable future, we'll always have a string-based query language for the backends. If this changes, there will be a lot more and fundamental changes than the - now typecasted - occurences.

## How Has This Been Tested?
There were no tests, the compiler would have shown any mistake.

## Types of changes

## Checklist:


dennisoelkers (approved):
I think the assumption that a backend query always has a query string makes sense. Even if we implement a backend that does not work with query strings at all, this change makes sure that we have a better place to start refactoring, because dependencies on the query string are now explicit.

todvora:
Love it, thanks @janheise!

Issue #8706, opened 2020-08-05
Add Support for parsing JSON Arrays in Processing Pipelines


ChristopherKB:
## What?

We are attempting to switch from json extractors on the input to pipeline processors doing JSON cracking

We need a way to process (flatten?) JSON arrays so that they behave similarly to the extractor, since we won't know the schema beforehand.
## Why?

Bring in logs that are formatted as JSON arrays.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

patrickmann:
**Proposal**: add a new pipeline function `JsonFlatten` which takes a nested JSON string and returns a flattened JSON tree, similarly to the JSON extractor. The resulting flattened tree only contains top-level key/value pairs - there are no more objects or arrays of any kind. This JSON tree can then be further processed by any function taking a JSON tree input, e.g. `to_map`; or transformed to a string via `to_string`.

Nested arrays within a multi-dimensional array are anonymous. We use the corresponding array index to uniquely identify these in the flattened tree.

Input:
- value: string - the JSON source string, arbitrarily nested
- key_separator: string - separator used in concatenating the keys of flattened structures

Examples:
```
Input: {k1:v1, o1:{k11:v11, o11:{k111:v111}}}
Output {k1:v1, o1_k11:v11, o1_o11_k111:v111}

Input: {k1:v1, a1:[[k11:v11, k12:v12],[k21:v21, k22:v22]]}
Output: {k1:v1, a1_0_k11:v11, a1_0_k12:v12, a1_1_k21:v21, a1_1_k22:v22}
```

boosty:
LGTM 👍 

@ChristopherKB Would this solve your use-case?

Would also like to get feedback from @bernd.

bernd:
@patrickmann I like the general idea of a separate function that returns a `JsonNode` similar to `parse_json`. :+1: 

> The function is similar to both JSON Extractor (in the way we concatenate keys in flattening) and `parse_json` (in the input and output data types), so it integrates well with the existing functionality.

The behavior you described is not how the JSON extractor works. The JSON extractor is handling list values differently:

**Input:**
```json
{
  "string": "text value",
  "string_list": [
    "one",
    "two"
  ],
  "object_list": [
    {
      "a": 1,
      "b": 2
    },
    {
      "c": 3,
      "d": 4
    }
  ],
  "nested_object": {
    "bool": true,
    "number": 1,
[...]

PR #11708, opened 2021-11-29
 Add Support for parsing JSON Arrays in Processing Pipelines

patrickmann:
We want to move processing away from extractors and into the pipeline. This PR provides a function that is similar to the JSON Extractor, while addressing issues the extractor has in processing nested containers.

See related issue for more details.

Resolves Graylog2/graylog2-server#8706


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
```diff
@@ -16,11 +16,116 @@
...
+            LOG.debug("Unknown type \"{}\" in key \"{}\"", value.getClass(), key);
+            return Collections.emptySet();
+        }
+    }
+
+    private static Collection<Entry> parseObject(
+            String key, Object value, ObjectMapper mapper, boolean flattenObjects, boolean escapeArrays, boolean deleteArrays)
+            throws JsonProcessingException {
+        @SuppressWarnings("unchecked")
+        final Map<String, Object> mapWithoutNull = Maps.filterEntries((Map<String, Object>) value, REMOVE_NULL_PREDICATE);
```
AntonEbel, line 92: It would be safer to cast the Object directly after the `instanceof `check and adapt the method signature to: 
`parseObject(
            String key, Map<String, Object> value, ObjectMapper mapper, boolean flattenObjects, boolean escapeArrays, boolean deleteArrays)`

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
```diff
@@ -16,11 +16,116 @@
...
+        if (deleteArrays) {
+            // ignore all arrays
+            return Collections.emptySet();
+        } else if (escapeArrays) {
+            // serialize, so it can be re-parsed as valid JSON
+            return Collections.singleton(Entry.create(key, mapper.writeValueAsString(value)));
+        } else {
+            // flatten array using indices for unique keys
+            @SuppressWarnings("unchecked")
+            int listSize = ((List<Object>) value).size();
```
AntonEbel, line 118: same here with the cast.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/json/JsonUtils.java
```diff
@@ -16,11 +16,116 @@
...
+    private static final String KEY_SEPARATOR = "_";
+    private static final String KEY_VALUE_SEPARATOR = ":";
+    private static final String LIST_SEPARATOR = ",";
+    private static final RemoveNullPredicate<Map.Entry<String, Object>> REMOVE_NULL_PREDICATE = new RemoveNullPredicate<>();
+
+    private JsonUtils() {
+    }
+
+    public static JsonNode extractJson(
+            String value, ObjectMapper mapper, boolean flattenObjects, boolean escapeArrays, boolean deleteArrays)
```
AntonEbel, line 54: It would be nice to have a data class with a builder for the three boolean values. I think we use @AutoValue.Builder for builder classes.

AntonEbel (changes requested):

patrickmann:
@AntonEbel thx for the suggestions. Made the updates and removed some non-essential unit tests at the same time.

AntonEbel (approved):
LGTM.

PR #11713, opened 2021-11-30
Give up leader status instantly if lock renewal fails

mpfz0r:
Idea from @bernd 


graylog2-server/src/main/java/org/graylog2/cluster/leader/AutomaticLeaderElectionService.java
```diff
@@ -89,14 +89,11 @@ private void runIteration() {
 
         try {
             isLeader = lockService.lock(RESOURCE_NAME).isPresent();
-            lastSuccess = System.nanoTime();
```
bernd, line 92: I think the `lastSuccess` field is now unused and can be removed.

PR #11719, opened 2021-12-01
Refactor LeaderElectionService start / shutdown

mpfz0r:
- Convert all LeaderElectionService into a true guava Service.
    
     - Remove the LeaderElectionService from the ServiceManager
       so we can shutdown it independently and thus very late
       in the shutdown sequence.

     - Move unlock handling into Service#shutdown() hook.
       This is simpler than to have a separate and confusing
       giveUpLeader() method.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/commands/Server.java
+++ b/graylog2-server/src/main/java/org/graylog2/commands/Server.java
@@ -289,6 +289,7 @@ public class Server extends ServerBootstrap {
             serviceManager.stopAsync().awaitStopped();
 
             leaderElectionService.stopAsync().awaitTerminated();
+
             // Some services might continue performing processing
             // after the Journal service being down. Therefore it's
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog2/commands/Server.java
```diff
@@ -283,6 +288,7 @@ public void run() {
             gracefulShutdown.runWithoutExit();
             serviceManager.stopAsync().awaitStopped();
 
+            leaderElectionService.stopAsync().awaitTerminated();
```
thll, line 291:
```suggestion
            leaderElectionService.stopAsync().awaitTerminated();
            
```

thll (changes requested):
Looks good and works well. Just two nitpicks/suggestions.

thll (approved):
LGTM

PR #11722, opened 2021-12-01
Adds optional override path to CSVFileDataAdapter

kingzacko1:
## Description
Adds an optional overridePath field to CSVFileDataAdapter.Config

## Motivation and Context
To support Graylog2/graylog-plugin-enterprise#2878 there needs to be an option for an override path in CSVFileDataAdapters. 

## How Has This Been Tested?
This is just an addition of a field in the CSVFileDataAdapter.Config class. Since it is nullable it should not affect any existing Data Adapters and I tested that it does not affect any newly created ones.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
@@ -105,7 +105,7 @@ public class CSVFileDataAdapter extends LookupDataAdapter {
         if (!pathChecker.fileIsInAllowedPath(Paths.get(config.path()))) {
             throw new IllegalStateException(ALLOWED_PATH_ERROR);
         }
-        if (config.hasOverridePath() && !pathChecker.fileIsInAllowedPath(Paths.get(config.overridePath()))) {
+        if (config.canBeOverridden() && !pathChecker.fileIsInAllowedPath(Paths.get(Objects.requireNonNull(config.overridePath())))) {
             throw new IllegalStateException(ALLOWED_PATH_ERROR);
         }
         if (config.checkInterval() < 1) {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
```diff
@@ -99,13 +105,20 @@ public void doStart() throws Exception {
         if (!pathChecker.fileIsInAllowedPath(Paths.get(config.path()))) {
             throw new IllegalStateException(ALLOWED_PATH_ERROR);
         }
+        if (config.canBeOverridden() && !pathChecker.fileIsInAllowedPath(Paths.get(Objects.requireNonNull(config.overridePath())))) {
```
roberto-graylog, line 108: Readability-wise, _pathChecker.filesInAllowedPath(Paths.get(Objects.requireNonNull(config.overridePath())_ is too complex.  Also, **Paths.get(Objects.requireNonNull(config.overridePath()** can be evaluated twice, depending on the state of **config.canBeOverridden()**.    It may be best if **Paths.get(Objects.requireNonNull(config.overridePath()** is evaluated and saved to a variable once.

Also, the same operation is found in doRefresh(..)

graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
```diff
@@ -191,6 +223,23 @@ protected void doRefresh(LookupCachePurge cachePurge) throws Exception {
         return newLookupBuilder.build();
     }
 
+    private void populateLookupRefs() throws IOException {
+
+        // populate lookup with CSV file
+        try (final InputStream inputStream = Files.newInputStream(Paths.get(config.path()));
```
roberto-graylog, line 229: These two try-blocks could be extracted to a single method that takes the AtomicReference and a path.

kingzacko1:
Reverting back to draft as these changes may not be necessary

PR #11723, opened 2021-12-01
Don't filter "only once" inputs in persisted inputs iterator

thll:
Rather than filtering the "only once per cluster" inputs in the
iterator of PersistedInputsImpl, we explicitly handle them when dealing
with leader-change events and input launching on server startup.

This is more obvious and hopefully less prone to subtle errors.


graylog2-server/src/main/java/org/graylog2/shared/inputs/InputLauncher.java
```diff
@@ -116,15 +119,32 @@ protected void handleLaunchException(Throwable e, IOState<MessageInput> inputSta
...
     }
+
+
+    public boolean shouldStartAutomatically(MessageInput input) {
+        return configuration.getAutoRestartInputs() || input.getDesiredState().equals(IOState.Type.RUNNING.toString());
+    }
+
+    public boolean leaderStatusInhibitsLaunch(MessageInput input) {
+        if (input.onlyOnePerCluster() && input.isGlobal() && !leaderElectionService.isLeader()) {
+            LOG.info("Not launching 'onlyOnePerCluster' input [{}/{}/{}] because this node is not the leader.",
```
mpfz0r, line 144: I think having this LOG at line 123 is more consistent

mpfz0r (commented):
One small change. rest looks good :+1:

PR #11734, opened 2021-12-03
Add backend support for user-defined TZ in exports

### LINKED ISSUE #9546: Generate reports and CSVs in local time
## What?

As a user, I know that the values I see for log timestamps in my browser are localized to my browser's time. This differs when I generate a report or a CSV, as those values show system time or UTC (z-4329). I'd like to have the option to generate reports/CSVs in my local time, versus system time.

todvora:
This PR adds support for user-defined TimeZone in exports. This allows the api request to contain an optional `time_zone` field which is then used to format timestamps. If none is specified, the time zone of the user is used.

## Motivation and Context
Fixes #9546 

## How Has This Been Tested?
Added integration test, additionally manually verified via API calls (see screenshot bellow)

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/views/export/ElasticsearchExportBackendIT.java
+++ b/graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/views/export/ElasticsearchExportBackendIT.java
@@ -31,6 +31,7 @@ import org.graylog.testing.elasticsearch.SearchServerInstance;
 import org.graylog2.indexer.ElasticsearchException;
 import org.graylog2.plugin.indexer.searches.timeranges.AbsoluteRange;
 import org.graylog2.plugin.indexer.searches.timeranges.InvalidRangeParametersException;
+import org.joda.time.DateTimeZone;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/export/CommandFactory.java
```diff
@@ -104,12 +103,15 @@ private AbsoluteRange toAbsolute(TimeRange timeRange) {
...
+        ExportMessagesCommand.Builder requestBuilder = ExportMessagesCommand.builder()
+                .fieldsInOrder(resultFormat.fieldsInOrder());
 
         if (resultFormat.limit().isPresent()) {
-            requestBuilder.limit(resultFormat.limit().orElseThrow(() -> new IllegalStateException("No value present!")));
+            requestBuilder = requestBuilder.limit(resultFormat.limit().orElseThrow(() -> new IllegalStateException("No value present!")));
+        }
+
+        if (resultFormat.timeZone().isPresent()) {
+            requestBuilder = requestBuilder.timeZone(resultFormat.timeZone().orElseThrow(() -> new IllegalStateException("No time zone present!")));
```
todvora, line 114:
```suggestion
        resultFormat.timeZone().ifPresent(requestBuilder::timeZone);
```

The logic of the condition prevents any possible IllegalStateException, so I'd suggest to simplify the expression. And maybe adapt the very same expression with `limit` above?

todvora (commented):
Besides the small optional expressoin everything looks perfect! Thx!

graylog2-server/src/main/java/org/graylog/plugins/views/search/export/CommandFactory.java
dennisoelkers: I don't have a strong opinion towards this. I tried to avoid assuming that the builder is mutable, but with the requirements of variables being (effectively) final in closures and therefore allowing no reassignment, it results in slightly clumsy code.

todvora: Isn't a builder by definition mutable, building immutable objects as a result? I'd prefer to mutate the builder rather than to re-assign the local variable. And we'd avoid the duplicated check if the timezone is present.

dennisoelkers: I am a bit used to immutable builders which we are using in the frontend. As long as we are using AutoValue for builders, we are fine here too.

janheise (approved):
LGTM, only one small change re: timezones in tests

graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/views/export/ElasticsearchExportBackendIT.java
```diff
@@ -213,6 +214,21 @@ public void sortsByTimestampAscending() {
                 "graylog_0, 2015-01-01T04:00:00.000Z, source-2, Ho");
     }
 
+    @Test
+    public void usesProvidedTimeZone() {
+        importFixture("messages.json");
+
+        ExportMessagesCommand command = commandBuilderWithAllStreams()
+                .timeZone(DateTimeZone.forID("Europe/Vienna"))
```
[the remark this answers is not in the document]
todvora: Good catch, working on it!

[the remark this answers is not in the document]
todvora: Used ACDT, which is UTC+9:30, resulting in a nice timestamps, easily distinguishable from the server UTC TZ.

PR #11736, opened 2021-12-03
Prevent infinite recursion when serializing `IpAddress`.

dennisoelkers:
## Motivation and Context

**Note:** Due to the simplicity of this fix, we should consider backporting it to `4.1` and `4.2`.

Prior to this PR, when a message had an instance of an `IpAddress` (added through a pipeline rule) in one of its fields, indexing that message results in an infinite recursion due to trying to serialize `IpAddress#getAnonymized` (which returns a new, "anonymized" instance of itself and another attempt to serialize that one is started). What is even worse, there is no way to recover from this than deleting the complete journal. All further message indexing will stall until this is resolved.

An example rule to trigger this could look like this:
```
rule "Adding IP field"
when
  true
then
  set_field("client_ip", to_ip("192.168.1.42"));
end
```

This PR is now marking `IpAddress#getAnonymized` to be ignored for JSON serialization and uses `IpAddress#toString` instead to retrieve the JSON value of this class.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ips/IpAddress.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/ips/IpAddress.java
@@ -16,6 +16,8 @@
  */
 package org.graylog.plugins.pipelineprocessor.functions.ips;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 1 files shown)


mpfz0r (approved):
Wow. So this never ever worked before?

dennisoelkers:
I have not researched if it has ever worked before, but it does not look like that.

PR #11738, opened 2021-12-03
Improve "key_value" pipeline function to handle quoted values

### LINKED ISSUE #4751: key value converter vs. key value function
I've created a pipeline with key value function, but I'm not getting the same result as with the key value converter.

```
Example log line:

subject=“TOTAL: Automatische Buchungsvorschlaege Maerz 2018”

result with key_value function:
fm_subject
TOTAL:

result with key_value Converter:
subject
TOTAL: Automatische Buchungsvorschlaege Maerz 2018

```
See also: 

https://community.graylog.org/t/graylog-extraktor-with-lookup-table/5003/11

* Graylog Version:  2.4.3-1 
* Elasticsearch Version: 5.6.6
* MongoDB Version: 3
* Operating System: Ubuntu 16.04 / 18.03.0~ce-0~ubuntu 
* Browser version:

### LINKED ISSUE #11731: Enhance Pipeline Function `key_value` to support double quoted text
## What?
Currently, when using the `key_value` function in the pipelines, quoted text is handled improperly.  For example, if you get a log, were a key value pair is:

country="United States" you get the field `country` and the value is `United`, but the `States` is dropped all together.  Need to have a way to recognize that this is a quoted value, and grab  everything in between the double quotes as the value.

## Why?

Parsing Logs from sources putting in Key Value pairs is easy, but when you run into these issues, you have to create a different solution which can be overlooked if the sample data does not have examples of " " with spaces in there.  

This would cover issues the security team is uncovering with the function.

Similar to issue: #4751

## Your Environment

* Graylog Version: 4.2+
* Illuminate Processing Engine (But is in all of them)

bernd:
The "key_value" pipeline function can now handle quoted values that
contain the configured "delimiter". (by default whitespace)

Example:
The value extracted from "key='a value'" will now be "a value" instead
of "a".

Fixes #11731
Fixes #4751

**Note:** We might want to backport this to 4.2

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/KeyValue.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/strings/KeyValue.java
@@ -76,7 +76,7 @@ public class KeyValue extends AbstractFunction<Map<String, String>> {
         final CharMatcher kvPairsMatcher = splitParam.optional(args, context).orElse(CharMatcher.whitespace());
         final CharMatcher kvDelimMatcher = valueSplitParam.optional(args, context).orElse(CharMatcher.anyOf("="));
 
-        Splitter outerSplitter = Splitter.on(kvPairsMatcher)
+        Splitter outerSplitter = Splitter.on(DelimiterCharMatcher.withQuoteHandling(kvPairsMatcher))
                 .omitEmptyStrings()
                 .trimResults();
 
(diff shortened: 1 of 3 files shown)


thll (approved):
Looks good and works.

Are we OK with not supporting any kind of escaping inside quotes? With the current solution, for an input like:

```extracted="and I'm quoting \"the quote\" for you."```
we will get the value
```"and I'm quoting \"the```

I think we can live with that, but just wanted to mention it in case you think it's an issue.

bernd:
> Are we OK with not supporting any kind of escaping inside quotes? With the current solution, for an input like:
> 
> `extracted="and I'm quoting \"the quote\" for you."` we will get the value `"and I'm quoting \"the`
> 
> I think we can live with that, but just wanted to mention it in case you think it's an issue.

@thll Good catch! I also think we can live with that for now. This fix is already an improvement compared to the current situation.

I will prepare a separate PR once I merged this to propose a possible solution.

PR #11754, opened 2021-12-06
Stop processing further stages after drop_message

### LINKED ISSUE #4855: Pipeline processing will not stop for dropped messages
When trying to filter out noise from event logs, the pipeline rules seem as excellent option, as complex conditions are available. 

## Expected Behavior

When setting up noise reduction pipeline, it would seem logical to make the most voluminous drops at stage 0, and then have less messages to process in later stages.

## Current Behavior

At the moment, there are two things that happen

1. If there is no "empty rule": 
```
rule "empty"
when 
  (1 == 1)
then
end
```
processing does not continue to the next stages for the messages that are **not** dropped.

2. After adding the empty rule to the first stage, all messages are processed on the second stage, even the dropped ones.

## Possible Solution

Two suggestions:

1. Add a new condition to pipeline stage processing: "continue processing on the next step if none of the rules match condition"
2. When message is dropped, stop evaluating other rules for that message

1. Make a stream Windows Events and route a Windows Event log entry there
2. Make a pipeline "Filter unnecessary Windows Events".  Connect it to the Windows Events stream
3. Stage 0, create a rule that drops messages with condition EventID == 4662 (for example - if your test setup creates these events)
4. Stage 1, create a rule that forwards the messages to a specific new stream "TestStream"

## Your Environment

* Graylog Version: 2.4.5
* Elasticsearch Version: 5.6.9
* MongoDB Version: 3.4
* Operating System: CentOs7
* Browser version: any

patrickmann:
`drop_message` has 2 purposes:
- prevent the message from being stored in ES
- improve performance by short-circuiting pipeline processing 

The current implementation does not do a good job on the latter. We are excessively conservative about avoiding unexpected side-effects between pipelines that process the same message. As a result, `drop_message` rarely achieves significant performance gains; and resulting metrics are often confusing to users.

We define the behavior more strictly now - which should also be more intuitive to users:
When `drop_message` is called, we complete processing of that stage; as well as all stages up to that stage number in other pipelines (for that message). Any subsequent stages are skipped, across all pipelines.

Resolves #4855

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -31,7 +31,14 @@ pre 2.2 Graylog installations.
 
 ## Configuration File Changes
 
-The following configuration option as been removed: `index_field_type_periodical_interval`.
+The following configuration option has been removed: `index_field_type_periodical_interval`.
 
 It has been replaced with a new configuration option which allows users to tweak the default full refresh interval for
 index field type information: `index_field_type_periodical_full_refresh_interval`.
+
+## Behaviour Changes
+
+- Pipeline function `drop_message` was modified to provide more performant and 
+predictable results: When `drop_message` is called in a rule, we complete processing of
+the current stage; following stages are skipped. Other pipelines operating on the same message will
+complete stage numbers less than or equal to the aborting stage; higher numbered stages are skipped.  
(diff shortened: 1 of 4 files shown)


thll (changes requested):
Code looks good and works fine.

What do you think about altering the description of the function in `DropMessage.java` a little along the lines of
`Discards a message from further processing` ->  `Discards a message from further processing once the current stage has completed`.

It's very unlikely that anyone is relying on the old behaviour but just to be sure we might want to include a note in `UPGRADING.md` about the change to give users a chance to check their pipeline setup.

thll (approved):

PR #11766, opened 2021-12-07
Deprecate usage of the term "master"

thll:
TODO:
- [ ] `org.graylog2.cluster.leader.StaticLeaderElectionService` and `org.graylog2.commands.Server` are using the `isMaster` to demote a server from leader to follower in case there is a second leader in the cluster on startup. Consider refactoring this by moving the duplicate leader detection logic into the static leader election service itself.
- [ ] Consider refactoring/removing everything related to `org.graylog2.plugin.ServerStatus.Capability`.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
@@ -39,12 +39,27 @@ public abstract class Periodical implements Runnable {
     /**
      * Only start this thread on master nodes?
      *
-     * @return
+     * @deprecated Use {@link #leaderOnly()} instead.
      */
-    public abstract boolean masterOnly();
+    @Deprecated
+    public boolean masterOnly() {
+        return false;
+    }
+
+    /**
+     * Determines if this periodical should run only on the leader node.
+     *
+     * @return {@code false} (default) if this periodical may run on every node. {@code true} if it may run only on the
+     * leader node
+     */
+    public boolean leaderOnly() {
+        // Defaulting to the now deprecated original method to not break existing implementations.
+        return masterOnly();
+    }
 
     /**
      * Start on this node? Useful to decide if to start the periodical based on local configuration.
+     *
      * @return
      */
     public abstract boolean startOnThisNode();
--- a/graylog2-server/src/main/java/org/graylog/events/periodicals/EventNotificationStatusCleanUp.java
+++ b/graylog2-server/src/main/java/org/graylog/events/periodicals/EventNotificationStatusCleanUp.java
@@ -54,7 +54,7 @@ public class EventNotificationStatusCleanUp extends Periodical {
     }
 
     @Override
-    public boolean masterOnly() {
+    public boolean leaderOnly() {
         return true;
     }
 
--- a/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/periodical/LegacyDefaultStreamMigration.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/periodical/LegacyDefaultStreamMigration.java
@@ -52,7 +52,7 @@ public class LegacyDefaultStreamMigration extends Periodical {
     }
 
     @Override
-    public boolean masterOnly() {
+    public boolean leaderOnly() {
         return true;
     }
 
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/periodical/PurgeExpiredConfigurationUploads.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/periodical/PurgeExpiredConfigurationUploads.java
@@ -45,7 +45,7 @@ public class PurgeExpiredConfigurationUploads extends Periodical {
     }
 
     @Override
-    public boolean masterOnly() {
+    public boolean leaderOnly() {
         return true;
     }
 
--- a/graylog2-server/src/main/java/org/graylog/plugins/sidecar/periodical/PurgeExpiredSidecarsThread.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/sidecar/periodical/PurgeExpiredSidecarsThread.java
@@ -50,7 +50,7 @@ public class PurgeExpiredSidecarsThread extends Periodical {
     }
 
     @Override
-    public boolean masterOnly() {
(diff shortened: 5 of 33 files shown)


kyleknighted (commented):
React changes look fine, not worried about linting errors as they are known problems that require more rewriting than this PR is responsible for.
Couple questions on the UPGRADING.md file but nothing to force a change request
I have no insight on the Java changes.

kyleknighted:
@thll thanks for taking care of this 🙏

UPGRADING.md
```diff
@@ -31,10 +37,14 @@ pre 2.2 Graylog installations.
...
 ## Configuration File Changes
 
-The following configuration option has been removed: `index_field_type_periodical_interval`.
-
-It has been replaced with a new configuration option which allows users to tweak the default full refresh interval for
-index field type information: `index_field_type_periodical_full_refresh_interval`.
+| Option                                              | Action      | Description |
+| --------------------------------------------------- | ----------  | ------------ |
+| `is_master`                                         | **removed** | Replaced with `is_leader`. For backwards compatibility, `is_master` will still be evaluated, but `is_leader` takes precedence, if both are configured.|
+| `stale_master_timeout`                              | **removed** | Replaced with `stale_leader_timeout`. For backwards compatibility, `stale_master_timeout` will still be evaluated, but `stale_leader_timeout` takes precedence, if both are configured. |
```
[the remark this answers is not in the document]
mpfz0r: @kyleknighted above describes the API. This the config file.

graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
```diff
@@ -46,9 +46,9 @@ public String getNodeId() {
     }
 
     @Override
-    @JsonProperty("is_master")
-    public boolean isMaster() {
-        return (Boolean) fields.get("is_master");
+    @JsonProperty("is_leader")
+    public boolean isLeader() {
+        return (boolean) fields.get("is_leader");
     }
```
mpfz0r, line 52: I wonder if we should use
`@JsonIgnoreProperties(ignoreUnknown = true)` on NodeImpl, so that old entries with `is_master` don't create a problem?

graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
```diff
@@ -39,12 +39,27 @@
     /**
      * Only start this thread on master nodes?
```
mpfz0r, line 40:
```suggestion
     * Only start this thread on leader nodes?
```

graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
```diff
@@ -39,12 +39,27 @@
...
+    /**
+     * Determines if this periodical should run only on the leader node.
+     *
+     * @return {@code false} (default) if this periodical may run on every node. {@code true} if it may run only on the
+     * leader node
      */
-    public abstract boolean masterOnly();
+    public boolean leaderOnly() {
+        // Defaulting to the now deprecated original method to not break existing implementations.
+        return masterOnly();
```
mpfz0r, line 57: if this one returns false and `masterOnly()` calls `leaderOnly()` we have one /master/ less in our code ;)

mpfz0r (commented):
got a few comments. but mostly LGTM :+1:

UPGRADING.md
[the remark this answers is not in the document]
thll: Exactly. The API is simply exposing some select values of the configuration file, that's why deprecating the field is reflected in both the config file and the API.

UPGRADING.md
```diff
@@ -7,6 +7,12 @@ Please make sure to create a MongoDB database backup before starting the upgrade
 
 ## Breaking Changes
 
+### Breaking API changes ##
+
+| Endpoint                        | Description                 |
+| ------------------------------- | --------------------------- |
+| `GET /api/system/configuration` | The key `stale_master_timeout` has been been replaced with `stale_leader_timeout` in the response object. |
```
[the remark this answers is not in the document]
thll: I thought was OK to make it a breaking change, but was not really sure myself. See PR description. To avoid any potential issues I made it non-breaking as per your suggestion. I don't know of any good way to properly deprecate a field in a response and announce it to clients though. I guess usually you'd switch to a new API version at some point and deprecate the whole endpoint.

graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
thll: 🤔  but all the existing periodicals out there might implement `masterOnly()` because they don't know anything about `leaderOnly()`. As we now have the `PeriodicalsService` evaluating `leaderOnly()` exclusively, we have to call the periodical's implementation of `masterOnly()` and not just assume `false`. Or am I making this too complicated?

graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
thll: As far as I can see we are not deserializing this with Jackson but simply operate on a `Map<String, Object>` so I don't think the annotation would have any effect.

I'm not sure if we are even serializing nodes anywhere, so probably the Jackson annotations are not needed at all. But just to be sure I added a test to check if we properly serialize nodes. In fact we weren't properly serializing the deprecated `is_master` field anymore, so I fixed that.

Can you please check if this makes sense?

mpfz0r: yeah, it does :+1:

graylog2-server/src/main/java/org/graylog2/plugin/periodical/Periodical.java
mpfz0r: Sorry, I was distracted and should've read the comment. Makes sense to do it this way. Even though, I doubt that anyone implemented a master-only periodical.

mpfz0r (approved):

thll:
@dulanism, @boosty: now that this is merged we should adjust our documentation to reflect the change. We should use the new configuration options in our examples and also adjust the wording in general. Please let me know, if you need more information for that.

boosty:
@thll Great! I assigned @dulanism regarding the documentation.

PR #11750, opened 2021-12-06
Fix extractor date converters

patrickmann:
#11149 introduced a regression with regard to date converter. This breaks scenarios that use the date converter to assign a non-standard date to the `timestamp` field.

With this PR, the date converter again works as it did in 4.1, while preserving the newly added functionality from 4.2 (with regard to `timestamp`).

Resolves #11495

Issue #11495, opened 2021-10-20
processing error after migrating to 4.2.0-3


ngvirus:
Graylog-server version: 4.2.0-3
OS: Oracle Linux 8
Java: 1.8.0-openjdk-headless-1.8.0.302.b08-0.el8_4.x86_64
I just upgraded to this version of Graylog and in my processed events i see this error:
`gl2_processing_error: Value <2021-10-20 09:05:39,892> caused exception: Invalid format: "2021-10-20 09:05:39,892" is malformed at ",892".`

I have a very basic GROK extractor that works fine:

`\[%{DATA:timestamp}\]\s\[%{DATA:loglevel}\]\s\[%{DATA:fieldname2}\]\s\[%{DATA:fieldname3}\]\s\[%{DATA:fieldname4}\]\s-\s(?<message>(.|\r|\n)*)$`
In the server log i see this warning which is like 850 lines long repeating the same 6 lines over and over:
2021-10-20T13:13:28.286Z WARN  [ProcessBuffer] Unable to process event MessageEvent{raw=null, message=null, messages=null}, sequence 896092
java.lang.StackOverflowError: null
        at java.util.regex.Pattern$GroupTail.match(Pattern.java:4731) ~[?:1.8.0_302]
        at java.util.regex.Pattern$BranchConn.match(Pattern.java:4582) ~[?:1.8.0_302]
        at java.util.regex.Pattern$CharProperty.match(Pattern.java:3791) ~[?:1.8.0_302]
        at java.util.regex.Pattern$Branch.match(Pattern.java:4618) ~[?:1.8.0_302]
        at java.util.regex.Pattern$GroupHead.match(Pattern.java:4672) ~[?:1.8.0_302]
        at java.util.regex.Pattern$Loop.match(Pattern.java:4799) ~[?:1.8.0_302]

This is how a message looks like.

`[2021-10-20 08:56:06,420] [INFO] [field1] [field2] [field3] - rest of the message`

supahgreg:
Good to hear.  The invalid format error is due to you trying to set the `timestamp` field to something not matching `yyyy-MM-dd HH:mm:ss.SSS` (i.e. comma instead of period before milliseconds; also assumed to be UTC, just in case yours isn't).

miwent:
I don't know what would have changed, but what is the purpose of ```(?<message>(.|\r|\n)*)$``` - are you trying to do multi-line matching?

kokosda:
[...]

PR #11774, opened 2021-12-08
Fix extractor date converters (#11750) (4.2 backport)

mpfz0r:
* Fix regression when using date converter #11495

* get tests passing

* Run timestamp fallback conversion

If the converter fails for a timestamp value,
run the fallback timestamp conversion, to ensure that a Message
always has a valid timestamp.

* Add tests

* Improve gl2_processing_error message

Showing only the details is not helpful in some cases.

* Improve converter code

We don't need to call removeField before addField.
It will just overwrite the value anyway.
This avoids having to recalculate the message size twice.

Only null values don't get set. Those we need to remove explicitly.

* adjust test to new processing error output

* improve comments

Co-authored-by: Marco Pfatschbacher <marco@graylog.com>
(cherry picked from commit d12d84668242858bdd10d1930e11ec4f604302a2)

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -443,7 +443,7 @@ public class Message implements Messages, Indexable {
             }
             obj.put(FIELD_GL2_PROCESSING_ERROR,
                     processingErrors.stream()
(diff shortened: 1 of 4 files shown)


patrickmann (approved):
Looks good - tested locally on 4.2

PR #11778, opened 2021-12-09
Consider fielddata for field types, making them enumerable if enabled.

dennisoelkers:
## Description

Prior to this PR, generating field types relied solely on the type literal returned by the indexer and other properties of the index mapping were not considered. This could lead to e.g. the `source` field, which is mapped as a `text` field, but with `fielddata` enabled, to be mapped as a `string` field which is not `enumerable`.

This change is now taking the `fielddata` property into account. If it is present (therefore explicitly requested), the resulting type will be `enumerable`. If it is not present, it is up to the mapped type if it is `enumerable` or not (e.g. for `keyword` being `enumerable` anyway).

## Motivation and Context

We are relying more and more on field types and their properties to decide if certain functionalities in the frontend (enabling aggregations, providing suggestions etc.) are active or not. These field types must be accurate and reflect the capabilities present on the field as good as possible.

## Types of changes

## Checklist:


janheise (approved):
LGTM!

PR #11781, opened 2021-12-09
Fixing issue when auto-fixing deflector alias with multiple targets

dennisoelkers:
## Motivation and Context

**Note:** This needs a backport to `4.1` and `4.2`.

During index rotation checks, we are also checking if the current deflector alias points to more than one index. If that is the case (for whatever reason), we are trying to repair it by removing the older target. For ES6 this was working before, for ES7 we are using the wrong alias command in the request.

This PR is now fixing the command used and adds an integration test to verify that it is working as expected.

## Types of changes

## Checklist:


todvora (approved):
Looks good, thanks!

janheise (approved):
LGTM!

PR #11787, opened 2021-12-10
Display field value suggestions when writing a search query.

linuspahl:
**Please note:** This PR is based on https://github.com/Graylog2/graylog2-server/pull/11585 which need to be merged first.

## Motivation and Context

With this PR we are extending the suggestions when writing a search query. We already display suggestions for e.g. field names and operator. We are now also displaying suggestions for field names.

The suggestions will be displayed when writing e.g. the query `field_name:` or when selecting a field name suggestion.

image

Fixes: https://github.com/Graylog2/graylog2-server/issues/11491

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
```diff
@@ -41,15 +47,17 @@ type Props = {
...
   placeholder?: string,
+  streams?: Array<string>,
+  timeRange?: TimeRange | NoTimeRangeOverride,
   theme: DefaultTheme,
   value: string,
   warning?: QueryValidationState,
 };
 
-const defaultCompleterFactory = (completers) => new SearchBarAutoCompletions(completers);
+const defaultCompleterFactory = (completers, timeRange, streams) => new SearchBarAutoCompletions(completers, timeRange, streams);
```
linuspahl, line 57: Since the search query is only reflecting the selected time range and streams after submitting the form, we can not use a store inside `FieldValueCompletion` to get the information.

graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.ts
```diff
@@ -0,0 +1,159 @@
...
+      return false;
+    }
+
+    const queryField = getFieldByName(this.currentQueryFields, fieldName);
+
+    if (!queryField || !isEnumerableField(queryField)) {
+      const allFieldType = getFieldByName(this.allFields, fieldName);
+
+      return isEnumerableField(allFieldType);
+    }
```
linuspahl, line 100: I am not completely sure about the best way to identify the most accurate field type. The problem:

The query field types are only being updated, once we submit e.g. the search bar form. When a user changes the streams filter and adjusts the query afterwards (which could trigger the field value suggestions) without submitting the form in the meantime, the query fields are not "up to date". Because of this circumstance I am checking the "all" field types as a fallback, but I can imagine that this can still result in an API error in some cases when the field type properties vary.

graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
```diff
@@ -62,23 +70,32 @@ const handleExecution = (editor, onExecute, value, error) => {
...
+
+const _configureEditor = (editor, completer: AutoCompleter, configuredListeners: React.MutableRefObject<boolean>) => {
+  if (editor) {
     editor.commands.removeCommands(['find', 'indent', 'outdent']);
+    // eslint-disable-next-line no-param-reassign
     editor.completers = [completer];
+
+    if (!configuredListeners.current) {
+      editor.session.on('tokenizerUpdate', (input, { bgTokenizer: { currentLine, lines } }) => {
+        displaySuggestionAfterFieldName(editor, currentLine, lines);
```
linuspahl, line 93: In a previous version, tried a different approach to display the suggestions, mainly by adjusting the regex which identifies the word we want to display completions for and by including the field name in the field value suggestions.
This had some disadvantages:
- it only worked well, when there were no other field name suggestions. When typing in `test-field:` while there are still suggestion like `test-field-1`, the field value suggestions are not being displayed.
- the field names were considered when filtering the suggestions. Imagine there are four field value suggestions when typing in "test-field:". If you start typing "test-field:tes" it will still display all other suggestions, because the prefix of the value is part of the field name.

With the current approach we are opening the suggestions manually, it has the disadvantage, that the `QueryInput` contains code which is related to the field name suggestions related code and that we are not relying on the ace editor logic On the other hand it has the advantage, that we could now refetch the suggestions while typing in the field name.

graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.ts
dennisoelkers: I think we need to make some improvements for this outside of this PR. Can you create an issue so we can track it?

graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  return {};
+};
+
+const getFieldByName = (fields: FieldTypeMappingsList, fieldName: string) => {
+  return fields.find(({ name }) => name === fieldName);
+};
+
+const isEnumerableField = (field: FieldTypeMapping) => {
+  return !!field?.type.properties.find((property) => property === 'enumerable');
```
dennisoelkers, line 70: We can make this a bit easier:
```suggestion
  return field?.isEnumerable() ?? false;
```

graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
dennisoelkers, line 93: Maybe we can generalize the logic here so that we do not couple field name completion to the general completion logic. We could do something like this:
```suggestion
        if (completer.shouldShowCompletions(editor, currentLine, lines)) {
          editor.execCommand('startAutocomplete');
        }
```

Where `SearchBarAutoCompletions#shouldShowCompletions` would delegate the call to the different completers and returns if any one of them wants to show suggestions at this point.

dennisoelkers (changes requested):
Looks good so far and works fine! Added some suggestions for improvements inline.

graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
linuspahl: That is a really good Idea, it worked very well.

graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.ts
linuspahl: Sounds good, I will create an issue.

graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
dennisoelkers: While you are at it, can you type the function arguments?

dennisoelkers (approved):
Looks good now and works great. Waiting for #11585 to be merged before proceeding.

PR #11788, opened 2021-12-10
Implementing cookie authentication for API, using it in frontend.

### LINKED ISSUE #6831: HTTP authentication impossible in front of graylog
## Expected Behavior
It should be possible to use graylog behind a reverse proxy that requires HTTP authentication.

## Current Behavior
Behind a reverse proxy that requires HTTP authentication, a user cannot get past graylog's login screen.

Refresher: HTTP authentication works by sending a `Authorization` header containing a base64-encoded string containing `user:password`; for instance, HTTP authentication for user `foo` and password `bar` means the browser sends `Authorization: basic Zm9vOmJhcg==` on _each_ request to graylog, not only on the login page.

graylog unfortunately uses the same header name for session management; `XmlHttpRequests` are sent to the API using a customized `Authorization` string, containing a session UUID; for instance, a javascript call could include a `Authorization: basic YmRlMzE2ZDMtNDJhYi00OGU3LWFkNTEtMzc2Y2RjMzEyMDFjOnNlc3Npb24=`.

So either the reverse proxy refuses the authentication because it was clobbered by graylog and does not match any user anymore, or graylog API denies access because it expects a session management UUID, not a `user:password` combo.  In any case, the user cannot go past the login page.

## Possible Solutions

Each of those would work for me AFAICT:

* Graylog could use another header name (such as `X-Graylog-Authorization` instead of `Authorization`) for its session management.  To allow backwards compatibility, the HTTP header name could be set via a configuration variable with its default set as `Authorization`.
* Graylog could drop usage of HTTP headers for session management and move that session token to a cookie.
* Graylog could manually pass session UUID as a GET/POST parameter on API requests.
* Assuming a single username cannot have more than one session, graylog could accept a `username:password` `Authorization` as a unique session identifier instead of enforcing a `UUID:session` string.

## Context
Because it is company policy (for security, auditing and other purposes) to enforce HTTP auth for internal services, I cannot currently deploy graylog for said company.

## Your Environment
Running the `graylog-3.1.3-1.ova` appliance.

dennisoelkers:
## Motivation and Context

Prior to this PR, for authenticating with the backend API from the frontend, we were using a token created during session instantiation, keeping it in local storage. This has the drawback that any code that runs in the frontend can potentially extract the session token.

This PR is implementing cookie authentication by setting a new cookie during session creation (on `POST /api/system/sessions`) in addition to returning in the payload. This cookie has the `httpOnly` flag set, which leads to it being used for subsequent backend API requests but it is not possible to read the cookie from anywhere in the frontend.

Fixes #6831.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3013

## How Has This Been Tested?

- [x] Logging in with valid credentials through web interface
- [x] Failing to log in with invalid credentials through web interface
- [x] Using trusted header through trusted proxy
- [x] Failing to use trusted through proxy not in `trusted_proxies`

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/security/realm/CookieAuthenticationRealm.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.shiro.subject.Subject;
+import org.graylog2.plugin.database.users.User;
+import org.graylog2.shared.security.AuthenticationCookieToken;
+import org.graylog2.shared.users.UserService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+
+public class CookieAuthenticationRealm extends AuthenticatingRealm {
```
thll, line 35: Comparing this with `SessionAuthenticator`, I think we need to be careful to not overlook something.

1. We are not calling `ThreadContext.bind(subject)`. This means, that after the authentication filter logs in the subject [here](https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroAuthenticationFilter.java#L48), the session is not available via `Subject#getSession` anymore. We should make extra sure that this is OK before leaving it out.
2. Session extension was left out. Was this intentional?
3. SSO via trusted headers is not working anymore.

I'm wondering if we should maybe go the safe route and just use the `SessionAuthenticator` and hand it a `SessionIdToken` that we create by extracting the session ID from the cookie.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -141,7 +148,11 @@ public JsonNode newSession(@Context ContainerRequestContext requestContext,
         try {
             Optional<Session> session = sessionCreator.create(sessionId, host, authToken);
             if (session.isPresent()) {
-                return sessionResponseFactory.forSession(session.get());
+                final SessionResponse token = sessionResponseFactory.forSession(session.get());
+                return Response.ok()
+                        .entity(token)
+                        .cookie(createSecureCookie(token, requestContext))
+                        .build();
```
thll, line 155: To support SSO via trusted headers, I think we need to trigger setting the cookie also when `#validateSession` is called. Also, when testing SSO, I didn't see a request from the frontend to the `#validateSession` endpoint at all so I guess the frontend needs some adjustments as well.

thll (changes requested):
Next round of review. I think we should get SSO via trusted headers working again, then I'll have another look.

graylog2-server/src/main/java/org/graylog2/security/realm/CookieAuthenticationRealm.java
dennisoelkers: > I'm wondering if we should maybe go the safe route and just use the SessionAuthenticator and hand it a SessionIdToken that we create by extracting the session ID from the cookie. 

Yes, I came to that conclusion after having a second look. I simplified the code to just extract the cookie's value (if existing) before creating the token. This way we can reuse `SessionAuthenticator`, avoid duplication and make sure that the relevant cases are handled.

thll (changes requested):
Simplifying the code and re-using the session token for cookie based auth looks good.

I still can't get trusted header auth to work. Here's how I've tested this:
* Configured trusted headers with System–>Authentication–>Authenticators
* Set `Remote-User: test@graylog.local` header in browser. Session got created but no cookie set.
* Set `Remote-User: test1@graylog.local` header in browser and reload. Browser gets stuck on `We are preparing Graylog for you...` screen. Console shows `GET http://localhost:8080/api/users/test@graylog.local 403 (Forbidden)`
* I'm able to proceed when I delete the `username` key from application storage

It looks as if we are not associating a session with the user in the browser, because no cookie is set. We will also need to clean up the session of a user, when the header indicates that the user has changed for a browser instance. I believe this is how it was working before.

thll:
@dennisoelkers I still can't get the following scenario to work:

1.  Set Remote-User: test@graylog.local header in browser. Session got created and cookie set.
2. Set Remote-User: test1@graylog.local header in browser and reload. Browser shows login screen.

My expectation is, that at step 2 I am logged in as test1@graylog.local. This is essentially the scenario from https://github.com/Graylog2/graylog-plugin-auth-sso/issues/35.

I can only get that to work if I manually delete the cookie and reload.

graylog2-server/src/main/java/org/graylog2/security/realm/HTTPHeaderAuthenticationRealm.java
```diff
@@ -65,24 +66,23 @@ public HTTPHeaderAuthenticationRealm(ClusterConfigService clusterConfigService,
...
+        final RemoteAddressAuthenticationToken remoteAddrToken = (RemoteAddressAuthenticationToken) token;
         final HTTPHeaderAuthConfig config = loadConfig();
 
         if (!config.enabled()) {
             LOG.debug("Skipping disabled HTTP header authentication");
             return null;
         }
 
-        final MultivaluedMap<String, String> headers = headersToken.getHeaders();
-        final Optional<String> optionalUsername = headerValue(headers, config.usernameHeader());
```
thll, line 85: The method `#headerValue` can now be removed as it's unused after this change.

thll (changes requested):
Trusted header auth now works as expected 👍

graylog2-server/src/main/java/org/graylog2/rest/resources/system/CookieFactory.java
```diff
@@ -0,0 +1,109 @@
...
+        final Optional<String> xForwardedProtoHeader = Optional.ofNullable(requestContext.getHeaderString(HEADER_X_FORWARDED_PROTO))
+                .filter(header -> !Strings.isNullOrEmpty(header));
+
+        if (xForwardedProtoHeader.isPresent()) {
+            return xForwardedProtoHeader;
+        }
+
+        final Optional<URI> origin = uriFromHeader(requestContext, HEADER_ORIGIN);
+
+        return origin.map(URI::getScheme);
```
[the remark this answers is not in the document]
dennisoelkers: No, that could actually be misleading, when there is a client <-http-> proxy <-https-> graylog situation. In this case we would interpret the scheme as being https, but the client is actually using http. In this case, a secure cookie would be rejected by the browser.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/CookieFactory.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private NewCookie makeCookie(String value, Date validUntil, ContainerRequestContext requestContext) {
+        final Date now = new Date();
+        final int maxAge = Long.valueOf((validUntil.getTime() - now.getTime()) / 1000).intValue();
+
+        final URI baseUri = baseUriFromRequest(requestContext);
+        final String basePath = Optional.ofNullable(Strings.emptyToNull(baseUri.getPath())).orElse("/");
+
+        final boolean isSecure = schemeFromRequest(requestContext)
+                .map(header -> header.equalsIgnoreCase("https"))
```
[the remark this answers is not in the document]
dennisoelkers: :white_check_mark:

graylog2-server/src/main/java/org/graylog2/security/realm/HTTPHeaderAuthenticationRealm.java
dennisoelkers: :white_check_mark:

thll (approved):
Nice! 👍

bernd (commented):
Are there any changes in this PR that we need to communicate in `UPGRADING.md` so users are aware of it?

linuspahl (approved):
I also tested it with an LDAP user and could not find any problems.

PR #11794, opened 2021-12-10
Normalize time range when validating search query.

linuspahl:
## Motivation and Context

With this change we are using ISO 8601 for times in an absolute
time range. The API expects this format and will throw an
error when using a different format.


dennisoelkers (approved):
Looks good, works as expected!

Issue #9822, opened 2020-12-17
Elasticseach and Kibana related warning spam after upgrade to 7.x


mikkolehtisalo:
If the ElasticSearch has also Kibana installed, after updating ElasticSearch to 7.x Graylog logs are full of spam like this:

```
2020-12-17T12:46:50.780+02:00 WARN  [RestClient] request [GET http://localhost:9200/gl-events_*/_alias?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=false] returned 1 warnings: [299 Elasticsearch-7.10.1-1c34507e66d7db1211f66f3513706fdf548736aa "this request accesses aliases with names reserved for system indices: [.kibana_task_manager, .kibana], but in a future major version, directaccess to system indices and their aliases will not be allowed"]
2020-12-17T12:46:50.781+02:00 WARN  [RestClient] request [GET http://localhost:9200/gl-events_*/_alias?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=false] returned 1 warnings: [299 Elasticsearch-7.10.1-1c34507e66d7db1211f66f3513706fdf548736aa "this request accesses aliases with names reserved for system indices: [.kibana_task_manager, .kibana], but in a future major version, directaccess to system indices and their aliases will not be allowed"]
2020-12-17T12:46:50.789+02:00 WARN  [RestClient] request [GET http://localhost:9200/gl-system-events_*/_alias?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=false] returned 1 warnings: [299 Elasticsearch-7.10.1-1c34507e66d7db1211f66f3513706fdf548736aa "this request accesses aliases with names reserved for system indices: [.kibana_task_manager, .kibana], but in a future major version, directaccess to system indices and their aliases will not be allowed"]
2020-12-17T12:46:50.791+02:00 WARN  [RestClient] request [GET http://localhost:9200/gl-system-events_*/_alias?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open&allow_no_indices=false] returned 1 warnings: [299 Elasticsearch-7.10.1-1c34507e66d7db1211f66f3513706fdf548736aa "this request accesses aliases with names reserved for system indices: [.kibana_task_manager, .kibana], but in a future major version, directaccess to system indices and their aliases will not be allowed"]
[...]

PR #11810, opened 2021-12-14
Fix typo in deprecation warnings filter

overhacked:
## Description

`elasticsearch_mute_deprecation_warnings = true` doesn't work for the warning about "system indices" because of a typo that omits a space between "direct" and "access". The warning in question, in full, reads: `299 Elasticsearch-7.15.2-93d5a7f6192e8a1a12e154a2b81bf6fa7309da0c "this request accesses aliases with names reserved for system indices: [.security], but in a future major version, direct access to system indices and their aliases will not be allowed`.

## Motivation and Context

Fixes issue #9822.

## How Has This Been Tested?

Not tested, but confirmed that the Elasticsearch error message has *always had a space in it*, by reviewing the `git blame` from the [deprecation commit](https://github.com/elastic/elasticsearch/commit/91f4b58bf71249cf81562bd6069c64aafa46591b) to HEAD in the elasticsearch repository: https://github.com/elastic/elasticsearch/blame/4e09186b3d3a20e97e44713d4200b22bc3cec332/server/src/main/java/org/elasticsearch/action/admin/indices/alias/get/TransportGetAliasesAction.java#L151-L153

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ElasticsearchFilterDeprecationWarningsInterceptor.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/ElasticsearchFilterDeprecationWarningsInterceptor.java
@@ -30,7 +30,7 @@ import java.util.stream.Collectors;
 public class ElasticsearchFilterDeprecationWarningsInterceptor implements HttpResponseInterceptor {
     private String[] messagesToFilter = {
             "setting was deprecated in Elasticsearch",
-            "but in a future major version, directaccess to system indices and their aliases will not be allowed",
+            "but in a future major version, direct access to system indices and their aliases will not be allowed",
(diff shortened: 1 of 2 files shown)


janheise:
@overhacked thanks for fixing it. There is also the test-class `ElasticsearchFilterDeprecationWarningsInterceptorTest` that has the wrong strings inside. Do you want to fix those, too, so you make a full contribution (otherwise I'll do it). Also, can you please sign the CLA?

overhacked:
>@overhacked thanks for fixing it. There is also the test-class `ElasticsearchFilterDeprecationWarningsInterceptorTest` that has the wrong strings inside. Do you want to fix those, too, so the tests are green and you make a full contribution (if not, I'll do it)? Also, can you please sign the CLA?

@janheise will do both today. Thanks!

overhacked:
@janheise Tests updated & CLA signed. Thanks!

janheise (approved):
thank you for fixing this!

PR #11814, opened 2021-12-14
Update Log4J to 2.16.0

mpfz0r:
From Log4j 2.16.0, support for lookups in log messages has been removed for security reasons.

CVE-2021-44228 has shown the JNDI has significant security issues.
2.16.0 disables JNDI by default.

https://logging.apache.org/log4j/2.x/changes-report.html#a2.16.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
         <json-path.version>2.4.0</json-path.version>
         <kafka.version>2.7.0</kafka.version>
         <kafka09.version>0.9.0.1-6</kafka09.version>
-        <log4j.version>2.15.0</log4j.version>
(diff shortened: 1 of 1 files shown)


philomory:
It's worth noting that [Apache has recently announced](https://logging.apache.org/log4j/2.x/security.html#History) that the `log4j2.formatMsgNoLookups` method is not actually sufficient to fully eliminate the vulnerability, and that the only way to fully eliminate the issue is to upgrade to log4j 2.16.0 (or 2.12.2 if you need to stick with 2.12.x for some reason).

dennisoelkers (approved):

PR #11833, opened 2021-12-15
Update Log4J to 2.16.0 (#11814)

mpfz0r:
From Log4j 2.16.0, support for lookups in log messages has been removed for security reasons.

CVE-2021-44228 has shown the JNDI has significant security issues.
2.16.0 disables JNDI by default.

https://logging.apache.org/log4j/2.x/changes-report.html#a2.16.0
(cherry picked from commit dd24b85cd1bbe1d8504d4670f6b0c394b41cc2d9)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
         <json-path.version>2.4.0</json-path.version>
         <kafka.version>2.7.0</kafka.version>
         <kafka09.version>0.9.0.1-6</kafka09.version>
-        <log4j.version>2.15.0</log4j.version>
+        <log4j.version>2.16.0</log4j.version>
         <metrics.version>4.1.9</metrics.version>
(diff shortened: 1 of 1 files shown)


bernd (approved):

hkotkanen:
Apologies if there would be a better place to ask, but will this fix also be provided for 4.0, seeing as the previous fix was made available for 4.0 and 3.3? And in either case, do you have an ETA for being able to fetch it through `yum`?

bernd:
> Apologies if there would be a better place to ask, but will this fix also be provided for 4.0, seeing as the previous fix was made available for 4.0 and 3.3? And in either case, do you have an ETA for being able to fetch it through `yum`?

@hkotkanen Yes, we are currently preparing releases for 4.2, 4.1, 4.0, and 3.3.

PR #11845, opened 2021-12-16
Cleaning up usage of test helper function `MockAction`.

linuspahl:
## Motivation and Context

`MockAction` is using `jest.fn(() => Promise.resolve())` per default to
mock an action. With this change we are cleaning up the cases where we
provided a payload for `MockAction` unnecessarily.

## Types of changes


mikedklein (approved):
@linuspahl there are some errors with linting but they seem unrelated to the changes you have. All tests ran and passed.

mikedklein (changes requested):
@linuspahl there are some linting errors. I polled the frontend group and we decided it was best to leave it for when you get back. Should be easy fixes the rest looks good and I tested it.

mikedklein (approved):
LGTM and we talked about the linting errors.

PR #11853, opened 2021-12-17
Remove not needed `AppRouter.jsx`.

linuspahl:
## Motivation and Context

Due to an incorrect merge there are currently two `AppRouter` files (`AppRouter.jsx` and `AppRouter.tsx`). In terms of the defined routes and functionality they are identical. 
Only the `AppRouter.jsx` is currently being used. With this PR we are removing the `AppRouter.jsx` because we favour the TypeScript version.

## Types of changes


mikedklein (approved):
Thoroughly tested existing routing and saw no regression from removing the old jsx file.

PR #11863, opened 2021-12-20
Replace removed `internalRuntimeForbidden` for forbiddenapis

bernd:
The `internalRuntimeForbidden` option got removed in forbiddenapis
version 3.0. It got replaced with the `jdk-non-portable` bundled
signature.

Also use the `commons-io.version` property to select the correct
`commons-io-unsafe` bundled signature.


bernd:
This change also removes an annoying error indicator in IntelliJ. :slightly_smiling_face:

thll (approved):
Nice, thank you!

PR #11884, opened 2021-12-28
Perform message timestamp conversion after processing

mpfz0r:
With #11149 we introduced a change that ensured that a message will
have a valid DateTime as the timestamp field.
The conversion or fallback (plus recording errors) was performed when
setting the `timestamp` field.

This caused problems (#11495) with date converters on extractors.
They work by first assiging the non-converted timestamp string to the
message, and then use this string in the date converter.
A fix for this was done in #11750

However, looking at another user's issue with a json extractor
(https://github.com/Graylog2/graylog2-server/issues/11495#issuecomment-996536075)
made me rethink the previous approach.

Our processing works by mutating a Message object and thus passing field
values from one processing step to another.
Enforcing a DateTime object on the timestamp field in the middle of the
processing has the potential to break configurations where multiple
steps are taken to convert a field.
E.g. the json extractor will just assign a tempoarary timestamp string,
which will later be converted using a pipeline rule.

Therefore, we only perform timestamp conversion/fallback under two
circumstances:

 - The processing is completed
 - An explicit call to Message.getTimestamp() is made

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
+++ b/graylog2-server/src/main/java/org/graylog2/plugin/Message.java
@@ -386,6 +386,7 @@ public class Message implements Messages, Indexable {
 
     @Override
     public DateTime getTimestamp() {
+        ensureValidTimestamp();
         return getFieldAs(DateTime.class, FIELD_TIMESTAMP).withZone(UTC);
     }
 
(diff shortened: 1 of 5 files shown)


patrickmann (approved):
Looks good - much cleaner than before. 

Tested regex extractor with date converter and assignment to timestamp field.

PR #11933, opened 2022-01-10
Refactor calculation of widget dimensions.

linuspahl:
## Motivation and Context

The background:
Most widgets are already responsive. They automatically adjust to the parents height, without the need to know its exact height / width in px. However some of them only adapt to e.g. the parent height on render. Plotly graphs also adapt when resizing the window. E.g. graphs and the world map are not adapting, when the parent is just being repainted, without being rerendered. This can happen e.g. when editing a widget and other elements like the search bar change their height, which affects the widget container height. You can also test this by creating a world map widget and changing the browser width.

The widgets are currently being rerendered, e.g. when we resize them, because props like `widgetPositions` change. In this case the `FullSizeContainer` in the `AggregationBuilder` is providing the new height and width for each widget. We are currently also calculating the widget dimensions in the `WidgetFrame` component and maintaining the information with the `widgetDimensions` state in the `WidgetGrid`. I could not find a case were we use these calculated dimension.

With this PR we are:
- removing the calculation of widget dimension in the `WidgetFrame`, because we do not need the information at this point.
- using a `ResizeObserver` for the calculation of widget dimensions inside the `FullSizeConainer`. Before this change the dimension were only being recalculated on a render, now they are being recalculated, whenever the dimensions change.

When these changes stand the test we can create a generic component for the related logic in the `FullSizeContainer` and use it to replace the `SizeMe` component, because this solution is more performant.

## Types of changes


dennisoelkers (approved):
Good change! Looks way cleaner and works great!

PR #11941, opened 2022-01-11
Allow Index Set selection on stream in Cloud

### LINKED ISSUE #988: Output cannot validate configuration settings before it is first invoked
When you configure an Output Plugin and enter the configuration, this configuration is only evaluated when the Output is actually invoked the first time (based on a message assigned to a stream). This is because constructor `dfPlugin (@NotNullable @Assisted Stream stream, @NotNullable @Assisted Configuration configuration)` is not called before first invocation and I see no other means for the plugin to verify this configuration.

This means an operator (or script) cannot verify the proper setup of an Output (notably the checking of credentials and target system) except for *actually inserting a message into a stream* which might pose unforeseen side-effects in production environments.

ousmaneo:
These changes enable selecting Index Sets when using Stream form in Cloud. 

fix https://github.com/Graylog2/graylog-plugin-cloud/issues/988

## Types of changes

## Checklist:


edmundoa (approved):
Looks good to me :ship:

PR #11947, opened 2022-01-12
Replace `react-sizeme` with own implementation which has a slightly better performance.

linuspahl:
## Motivation and Context

With https://github.com/Graylog2/graylog2-server/pull/11933 we implemented a hook to get the current dimensions of an element by using the `ResizeObserver` interface.

With this PR we are creating the common component `ElementDimensions` which implements this hook and which replaces the `SizeMe` component from `react-sizeme`.  `react-sizeme` is not using the `ResizeObserver` interface and has a slightly lower performance.

Since the `useElementDimensions` hook is working with a ref, the `ElementDimensions` component behaves slightly different than the `SizeMe` component. `ElementDimensions` is rendering a `div`, which is not the case when using the `SizeMe` component.

For our current use cases this is fine and in most cases we can just do the following: `const ParentComponent = styled(ElementDimensions)`. When we do not want to render an extra div, we can just use the `useElementDimension` hook instead.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/common/ElementDimensions.tsx
+++ b/graylog2-web-interface/src/components/common/ElementDimensions.tsx
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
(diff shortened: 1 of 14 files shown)


graylog2-web-interface/src/components/common/ElementDimensions.tsx
```diff
@@ -0,0 +1,47 @@
...
+type Props = {
+  children: (dimensions: { width: number, height: number}) => React.ReactNode
+  className?: string,
+  resizeDelay?: number,
+};
+
+/**
+ * Wrapper for the useElementDimensions hook.
+ */
+const ElementDimensions = ({ children, className, resizeDelay }: Props) => {
```
linuspahl, line 31: I thought about naming this component `MeasureDimensions`.

dennisoelkers: I think it is okay this way. It is more declarative and is very symmetrical to `useElementDimensions`.

dennisoelkers (approved):
Great work! Thanks! :tada:

PR #11951, opened 2022-01-12
Pass ChangeUserRequest to user mgmt service

patrickmann:
User profile data is store in the DB. In the case of external users in the cloud we also sync changes back to the Identity Provider (i.e. Okta). However, we only want to do this when changing attributes that are actually managed by the IP.

Pass the `ChangeUserRequest` object to the `UserManagementService`, so it can determine whether a sync is needed.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/users/UserManagementService.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/users/UserManagementService.java
@@ -18,6 +18,9 @@ package org.graylog2.shared.users;
 
 import org.graylog2.plugin.database.ValidationException;
 import org.graylog2.plugin.database.users.User;
+import org.graylog2.rest.models.users.requests.ChangeUserRequest;
+
+import javax.validation.Valid;
 
 /**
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/shared/users/UserManagementService.java
```diff
@@ -18,6 +18,9 @@
 
 import org.graylog2.plugin.database.ValidationException;
 import org.graylog2.plugin.database.users.User;
+import org.graylog2.rest.models.users.requests.ChangeUserRequest;
```
thll, line 21: It feels wrong, that we leak an object from the REST API layer into the service layer. I think we should either create an additional DTO to transport the change request or we could simply move the `ChangeUserRequest` class from the `org.graylog2.rest.models.users.requests` package to `org.graylog2.shared.users`.

thll (changes requested):
Looks good. I only have one comment.

thll (approved):

PR #11953, opened 2022-01-12
Convert ConnectionCounterTest to using Mock Channels

mpfz0r:
This saves us from the trouble of finding
a usable and stable IP address on github action runners.

Refs https://github.com/actions/virtual-environments/issues/3185


mpfz0r:
@dennisoelkers could you verify that this also works on MacOS? I think it should...

mpfz0r:
> @dennisoelkers could you verify that this also works on MacOS? I think it should...

Or maybe it doesn't. We'd need kqueue instead of epoll there :-/

dennisoelkers (approved):
Great improvement, thanks!

PR #11959, opened 2022-01-13
Suggestions size param, other docs count

todvora:
Allow controlling the size param, requesting a specified count of suggestions (with default 10 and max 100). Additionally, the response now contains the `sum_other_docs_count`, indicating how many documents with the requested fields exist out of the suggestion(=aggregation) scope. If the value is zero, the provided suggestions cover all possible values.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
@@ -35,12 +35,15 @@ public abstract class SuggestionResponse {
 
     public abstract Optional<SuggestionError> suggestionError();
 
+    @Nullable
+    public abstract Long sumOtherDocsCount();
 
-    public static SuggestionResponse forSuggestions(final String field, final String input, final List<SuggestionEntry> suggestions) {
-        return new AutoValue_SuggestionResponse(field, input, suggestions, Optional.empty());
+
+    public static SuggestionResponse forSuggestions(final String field, final String input, final List<SuggestionEntry> suggestions, Long sumOtherDocsCount) {
+        return new AutoValue_SuggestionResponse(field, input, suggestions, Optional.empty(), sumOtherDocsCount);
     }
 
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
```diff
@@ -35,12 +35,14 @@
...
-    public static SuggestionResponse forSuggestions(final String field, final String input, final List<SuggestionEntry> suggestions) {
-        return new AutoValue_SuggestionResponse(field, input, suggestions, Optional.empty());
+
+    public static SuggestionResponse forSuggestions(final String field, final String input, final List<SuggestionEntry> suggestions, long sumOtherDocsCount) {
+        return new AutoValue_SuggestionResponse(field, input, suggestions, Optional.empty(), sumOtherDocsCount);
     }
 
     public static SuggestionResponse forError(final String field, final String input, final SuggestionError error) {
-        return new AutoValue_SuggestionResponse(field, input, Collections.emptyList(), Optional.of(error));
+        return new AutoValue_SuggestionResponse(field, input, Collections.emptyList(), Optional.of(error), -1);
```
dennisoelkers, line 46: We could also use a boxed type (`Long`) for the field and return set it to `null`. This way we prevent a consumer to erroneously use the number from this field although it's an error. If we want to the response cleaner, we could even annotate the class with `@JsonInclude(JsonInclude.Include.NON_NULL)` to prevent the field from being serialized in that case.

dennisoelkers (changes requested):
Looks good! One minor remark. Also, I think `yarn-error.log` was committed accidentally.

graylog2-server/src/main/java/org/graylog/plugins/views/search/engine/suggestions/SuggestionResponse.java
todvora: Good point, thanks, implemented it like that!

todvora:
Removed the `yarn-error.log` commited by accident. Thanks for the hint!

dennisoelkers (approved):

Issue #11850, opened 2021-12-17
Update field value suggestions while writing field value in search query.


linuspahl:
## Description:

With https://github.com/Graylog2/graylog2-server/pull/11787 we are now displaying field value suggestions when writing a search query. Currently we only fetch the 10 most important suggestions after writing the field name. E.g. `location:`. 

We should also refetch the suggestions while writing the field name. E.g. `location:Eur` and display the 10 most important suggestions for the provided input.

PR #11974, opened 2022-01-17
Improve fetching of field value suggestions which are being displayed for the search query input.

linuspahl:
## Motivation and Context

Please note, this PR needs be tested together with https://github.com/Graylog2/graylog2-server/pull/11959 or after the PR has been merged.

As described in https://github.com/Graylog2/graylog2-server/issues/11850 we are currently only fetching the 10 most important field value suggestions after writing the field name in a search query.

With this PR we are now fetching 30 suggestions and we are refetching the suggestions when the user continues typing in a more specific field value prefix. We only fetch more suggestions, when we did not fetched all suggestions with the first request.

We are also displaying spelling corrections slightly different than suggestions which are just based on the field value prefix. These corrections include the currently defined field value name in their caption. This is necessary, because otherwise they would be filtered by the ace editor.

Fixes https://github.com/Graylog2/graylog2-server/issues/11850


dennisoelkers (approved):
Looks good, works fine! :tada:

PR #11976, opened 2022-01-17
Remove existing `DateTimeProvider` since it is not providing a huge benefit.

linuspahl:
## Motivation and Context

We added the `DateTimeProvider` to provide the current time range limit for the date time picker.
Since prop drilling is not a big concern in this case we can remove the context. We also want to use the name for the provider and context to provide general date time helper methods in the future.

## Types of changes

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3111


dennisoelkers (approved):
Good work, makes sense!

PR #11994, opened 2022-01-20
Simplify `Timestamp` component and unify format of timestamps displayed in the UI.

linuspahl:
Please note this PR needs to be tested together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/3122

## Motivation and Context

With this PR we are refactoring the `Timestamp` component. With this change we are unifying the way we display most timestamps in the UI.

Before these changes the component was also responsible for displaying a time relatively and based on the browser time zone. Now its main responsibility is displaying times based on the user time zone. It is still possible to define a custom time zone, when needed.

## Types of changes

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3122

### DIFF
--- a/graylog2-web-interface/src/components/alarmcallbacks/AlarmCallbackHistory.jsx
+++ b/graylog2-web-interface/src/components/alarmcallbacks/AlarmCallbackHistory.jsx
@@ -44,8 +44,8 @@ class AlarmCallbackHistory extends React.Component {
       </span>
     );
     const description = (hadError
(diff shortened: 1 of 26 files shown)


edmundoa (approved):
I like the change and I think it makes it easier to write and reason about what happens. The code seems to work and I could not find any missing usage that should be updated (even searched in the cloud plugin). Great work! 👍

PR #12011, opened 2022-01-21
Add group membership info received during AuthN to UserDetails

mpfz0r:
This field can be used to pass group membership information to an AuthNTeamSyncBackend


AntonEbel (approved):
LGTM.

PR #12012, opened 2022-01-21
Routing and navigation for Graylog Security product

mikedklein:
# Security App Routing

- Adds placeholder route for new security app plugin
- Adds placeholder page for new security app plugin

## Description
This provides the navigational adjustments to the core application for
the introduction of the security app plugin.

## Motivation and Context
Needed to get navigational items in place for the new product.

## How Has This Been Tested?
Tested locally and ran unit tests.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/navigation/Navigation.tsx
+++ b/graylog2-web-interface/src/components/navigation/Navigation.tsx
@@ -55,7 +55,7 @@ function pluginMenuItemExists(description: string): boolean {
 
   if (!pluginExports) return false;
 
-  return !pluginExports.find((value) => value.description?.toLowerCase() === description.toLowerCase());
+  return !!pluginExports.find((value) => value.description?.toLowerCase() === description.toLowerCase());
 }
 
 const formatSinglePluginRoute = ({ description, path, permissions }, topLevel = false) => {
@@ -98,8 +98,8 @@ const Navigation = React.memo(({ pathname }: Props) => {
 
   const pluginExports = PluginStore.exports('navigation');
 
-  const enterpriseMenuIsMissing = pluginMenuItemExists(ENTERPRISE_ROUTE_DESCRIPTION);
-  const securityMenuIsMissing = pluginMenuItemExists(SECURITY_ROUTE_DESCRIPTION);
+  const enterpriseMenuIsMissing = !pluginMenuItemExists(ENTERPRISE_ROUTE_DESCRIPTION);
+  const securityMenuIsMissing = !pluginMenuItemExists(SECURITY_ROUTE_DESCRIPTION);
 
   const isPermittedToEnterpriseOrSecurity = isPermitted(permissions, ['licenseinfos:read']);
 
@@ -107,7 +107,7 @@ const Navigation = React.memo(({ pathname }: Props) => {
     // no enterprise plugin menu, so we will add one
     pluginExports.push({
       path: Routes.SYSTEM.ENTERPRISE,
-      description: 'Enterprise',
+      description: ENTERPRISE_ROUTE_DESCRIPTION,
     });
   }
 
@@ -115,7 +115,7 @@ const Navigation = React.memo(({ pathname }: Props) => {
     // no security plugin menu, so we will add one
     pluginExports.push({
       path: Routes.SECURITY,
-      description: 'Analyst Tools',
+      description: SECURITY_ROUTE_DESCRIPTION,
     });
   }
 
--- a/graylog2-web-interface/src/components/navigation/Navigation.tsx
+++ b/graylog2-web-interface/src/components/navigation/Navigation.tsx
@@ -28,7 +28,7 @@ import { IfPermitted } from 'components/common';
 import { isPermitted } from 'util/PermissionsMixin';
 import CurrentUserContext from 'contexts/CurrentUserContext';
 import GlobalThroughput from 'components/throughput/GlobalThroughput';
-import Routes from 'routing/Routes';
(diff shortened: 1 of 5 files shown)


graylog2-web-interface/src/components/navigation/Navigation.tsx
```diff
@@ -88,6 +88,9 @@ const Navigation = React.memo(({ pathname }: Props) => {
   const enterpriseMenuIsMissing = !pluginExports.find((value) => value.description.toLowerCase() === 'enterprise');
   const isPermittedToEnterprise = isPermitted(permissions, ['licenseinfos:read']);
 
+  const securityMenuIsMissing = !pluginExports.find((value) => value.description.toLowerCase() === 'analyst tools');
```
linuspahl, line 91: What do you think about creating a constant which can be reused in the enterprise repo, to ensure we do not change on part without the other? We could do the same for the enterprise menu bindings.

graylog2-web-interface/src/routing/Routes.ts
```diff
@@ -60,6 +60,9 @@ const Routes = {
       show: (notificationId: string) => `/alerts/notifications/${notificationId}`,
     },
   },
+  PROMO: {
+    SECURITY: '/promo/security',
+  },
```
linuspahl, line 65: I think we can leave '/promo' out, I see a small benefit in using the same route for the promo and the non-promo page. You can just link to `/security` and do not need to worry about the security license status.
It would also be similar to the "enterprise" routing.

graylog2-web-interface/src/pages/SecurityPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+
+import * as React from 'react';
+import styled, { css } from 'styled-components';
+
+import { DocumentTitle, IfPermitted, PageHeader } from '../components/common/index';
+import HideOnCloud from '../util/conditional/HideOnCloud';
+import { Col, Row } from '../components/bootstrap/index';
```
linuspahl, line 23: We can slightly simplify the imports by removing `../` and `/index`.

linuspahl (commented):
Looks good! Just added some suggestions.

graylog2-web-interface/src/pages/SecurityPage.tsx
mikedklein: Fixed

graylog2-web-interface/src/routing/Routes.ts
mikedklein: Fixed

graylog2-web-interface/src/components/navigation/Navigation.tsx
mikedklein: Abstracted into the `Routes.ts` file.

graylog2-web-interface/src/routing/Routes.ts
linuspahl: It looks like the route is still `/promo/security`.

graylog2-web-interface/src/routing/Routes.ts
```diff
@@ -337,6 +338,26 @@ const pluginRoute = (routeKey, throwError = true) => {
...
+  return !pluginExports.find((value) => value.description?.toLowerCase() === route);
+}
+
+/**
+ * Exported constants for use in seeing if the plugin and menu item are registered for addon components.
+ *
+ */
+export const enterpriseMenuIsMissing = checkForRoute('analyst tools');
+export const securityMenuIsMissing = checkForRoute('enterprise');
+
```
linuspahl, line 360: Few suggestions regarding the naming:

```
function pluginMenuItemExists(description: string): boolean {
  const pluginMenuItems = PluginStore.exports('navigation');

  if (!pluginMenuItems) return false;

  return !pluginMenuItems.find(value) => value?.toLowerCase() === description.toLowerCase());
}
```

- `checkForRoute(route)` is a little bit misleading. Someone who is not familiar with the context will think this is a helper function for routes like '/the/route'. They will not understand that it is about plugable menu items.
- I implemented `description.toLowerCase()` to ensure we do not have a problem when some is not remembering that we expect a lowercase input.
- It looks like we need to implement `!checkForRoute()` (or now !menuItemExists()) for `enterpriseMenuIsMissing` and `securityMenuIsMissing`. This bug will hopefully be easier to catch with the new name, when reading the code.
- I know I suggested to place this code in the `Routes` file, but after having another look my preference is to only define:
```
const ENTERPRISE_ROUTE_DESCRIPTION = 'Enterprise';
const SECURITY_ROUTE_DESCRIPTION = 'Analyst Tools';
```
in this file and move the helper function `pluginMenuItemExists` to the `Navigation` file, since it is more related to the navigation than the actual routes.

linuspahl (changes requested):

graylog2-web-interface/src/routing/Routes.ts
mikedklein: I couldn't reproduce 
> It looks like we need to implement !checkForRoute() (or now !menuItemExists()) for enterpriseMenuIsMissing and securityMenuIsMissing. This bug will hopefully be easier to catch with the new name, when reading the code.

The rest of these issues are addressed.

graylog2-web-interface/src/routing/Routes.ts
mikedklein: Resolved.

graylog2-web-interface/src/routing/Routes.ts
mikedklein: Ok I think I got this to the correct state now.

linuspahl (approved):
Looks good to me!

PR #12026, opened 2022-01-24
Don't print a stacktrace for missing version files

mpfz0r:
In a cloud dev environment the version info for the cloud
is missing. This makes it hard to see other messages
in the server log.


AntonEbel (approved):
LGTM.

PR #12034, opened 2022-01-25
Add utility class for creating Email instances

thll:
Adds a utility class to create `Email` instances and apply settings from `EmailConfiguration` to it. Before this, we were always applying the same sequence of settings when creating Email instances at various places throughout our code base.

Also:
- adds a missing validation for the hostname
- removes leftover setting from our example `graylog.conf`


mpfz0r (approved):
:+1:

PR #12035, opened 2022-01-25
Allow overwriting `ExpandableList` styles.

linuspahl:
## Motivation and Context

This change for the `ExpandableList` component is necessary to style the component with `styled-components`.

### DIFF
--- a/graylog2-web-interface/src/components/common/ExpandableList.tsx
+++ b/graylog2-web-interface/src/components/common/ExpandableList.tsx
@@ -32,7 +32,7 @@ type Props = {
  */
 const ExpandableList = ({ children, className }: Props) => {
   return (
-    <ul className={`${style.list} ${className}`}>
+    <ul className={className ? `${style.list} ${className}` : style.list}>
       {children}
     </ul>
   );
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/common/ExpandableList.tsx
```diff
@@ -29,16 +30,17 @@ type Props = {
  * of categories. Inside the categories the user has the possibility of doing a selection.
  * The ExpandableList can be used nested.
  */
-const ExpandableList = ({ children }: Props) => {
+const ExpandableList = ({ children, className }: Props) => {
   return (
-    <ul className={style.list}>
+    <ul className={`${style.list} ${className}`}>
```
dennisoelkers, line 35: For the default props, this would result in `undefined` being part of the classes list. Instead we could do it like this:
```suggestion
    <ul className={className ? `${style.list} ${className}` : style.list}>
```

dennisoelkers (changes requested):
Looks good, just a minor suggestion.

dennisoelkers (approved):

PR #12038, opened 2022-01-26
Move NullCodec to server

mpfz0r:
This codec always returns a null Message.
Can be used in situations where we don't need a real codec or a message should be simply dropped after decoding.


thll (changes requested):
If we use the `NullCodec` as a legitimate codec, we have to handle the "dropping message silently" case by acknowledging the message in the message queue somewhere around [here](https://github.com/Graylog2/graylog2-server/blob/e482ad30cc75a8172d92a69340b7e4e02190a6de/graylog2-server/src/main/java/org/graylog2/shared/buffers/processors/DecodingProcessor.java#L177).

thll (approved):
👍

PR #12045, opened 2022-01-27
Add a notification for archiving errors

patrickmann:
Add a notification type for archiving errors. 
This is used in the related Enterprise PR https://github.com/Graylog2/graylog-plugin-enterprise/pull/3148.
See https://github.com/Graylog2/graylog-plugin-enterprise/issues/3127 for details.

### DIFF
--- a/graylog2-web-interface/src/logic/notifications/NotificationsFactory.js
+++ b/graylog2-web-interface/src/logic/notifications/NotificationsFactory.js
@@ -98,7 +98,7 @@ class NotificationsFactory {
           title: 'Elasticsearch cluster unavailable',
           description: (
             <span>
-              Graylog could not successfully connect to the Elasticsearch cluster. If you're using multicast, check that
+              Graylog could not successfully connect to the Elasticsearch cluster. If you are using multicast, check that
               it is working in your network and that Elasticsearch is accessible. Also check that the cluster name setting
               is correct. Read how to fix this in {' '}
               <DocumentationLink page={DocsHelper.PAGES.ES_CLUSTER_UNAVAILABLE}
@@ -306,7 +306,7 @@ class NotificationsFactory {
             </span>
           ),
         };
-      case 'es_version_mismatch':
+      case 'es_version_mismatch': {
         const { initial_version: initialVersion, current_version: currentVersion } = notification.details;
 
         return {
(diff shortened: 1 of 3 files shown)


patrickmann:
I fixed most of the new ESLint warnings except for renaming to .jsx 
I understand we are allowed to merge without fixing all new warnings (on existing code).

patrickmann:
@ousmaneo The JS console is showing this when rendering the new notification:
`Warning: Each child in a list should have a unique "key" prop`
Can you suggest how to fix that?

graylog2-web-interface/src/logic/notifications/NotificationsFactory.js
```diff
@@ -348,6 +374,21 @@ class NotificationsFactory {
...
+      }
+
+      case 'archiving_summary':
+        return {
+          title: 'Some archiving jobs did not complete successfully',
+          description: (
+            <span>
+              Recent failed attempts: {' '}
+              <ul>
+                {notification.details.archiveErrors.map((error) => <li key={error.id}>{error}</li>)}
```
edmundoa, line 386: > The JS console is showing this when rendering the new notification: `Warning: Each child in a list should have a unique "key" prop` Can you suggest how to fix that?

When rendering arrays of elements, React needs to have a unique key to each element, so it can track when to re-render them. In this case, `error` is a string and does not have an `id` property, so it defaults to `undefined`, which is the same as not providing a value for it. Since in this case the array is coming from the backend and will not be changed, I think the easiest way is to use array indices for it:

```suggestion
                {/* eslint-disable-next-line react/no-array-index-key */}
                {notification.details.archiveErrors.map((error, idx) => <li key={idx}>{error}</li>)}
```

Using the index is not ideal and ESLint will complain about it, so I included the comment to disable it on that line. I still think that is better than using a relatively long sentence as key.

edmundoa (changes requested):
The front end part looks good to me, besides the change with the `key`. Thank you for all the linter fixes!

graylog2-web-interface/src/logic/notifications/NotificationsFactory.js
```diff
@@ -348,6 +374,21 @@ class NotificationsFactory {
             </span>
           ),
         };
+      }
+
+      case 'archiving_summary':
+        return {
+          title: 'The following indices could not be archived yet (but will still be retried)',
```
edmundoa, line 381: The new message reads better but also leaves some questions open: Why could those indices not be archived? How many times will Graylog try? Did I lose any messages?

I think we should keep the title short, so I would vote for doing the following:
1. Keep `The following indices could not be archived yet`, or maybe `The following indices could not be automatically archived yet`, to emphasise that this happened in the background
2. Add more information in the description, something like:

```js
<span>
There was an error while archiving some indices during the index rotation process. Graylog will continue trying to archive those indices and will keep all indices until they are successfully archived.
<br />
Please check the following error messages as your assistance (or from a Graylog account manager) may be necessary to fix the issue:
<br />
<ul>(...)</ul>
</span>
```

What do you think?

edmundoa (approved):
I added a comment regarding the notification text. Otherwise looks good to me, so I'm approving it 👍

PR #12046, opened 2022-01-27
Ensure query input editor always references the latest provided data.

linuspahl:
## Description
This PR is fixing two bugs for the query input where an action provided for the `QueryInput` worked with outdated references:
- the `completers`, before this change they references an outdated version of the selected time range and streams.
- the `onExecute` action, before this change this function referenced an outdated an outed version of the `error` prop.

Before this change we tried to configure the editor, by using the relevant editor prop, like
- `onLoad` for the initial configuration
- `completers` for the custom completers
- `commands` for the custom commands

Unfortunately this is not an option when defining data which relies on external data, like `error`, `timeRange` and `streams`. In this case the reference for the external data is not up to date, whenever the ace editor is executing code which implements the mentioned props.

With this PR we are using the editor `ref` prop (again) to ensure we update the editor config and references to external data on every render.

I added tests for the query input and made sure we are testing the actual implementation, but I plan to add more tests in a follow-up PR, also for use cases which are not related to this bugfix.

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/QueryInput.test.tsx
+++ b/graylog2-web-interface/src/views/components/searchbar/QueryInput.test.tsx
@@ -15,20 +15,23 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
-import { mount } from 'wrappedEnzyme';
-import ReactAce from 'react-ace';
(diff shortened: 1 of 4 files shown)


dennisoelkers (approved):
Looks good, fixes the issue for me!

Issue #12060, opened 2022-01-28
Fix Issue with V20211221144300_GeoIpResolverConfigMigration


roberto-graylog:
## Expected Behavior
IF the field **db_vendor_type** is MAXMIND_CITY (old value), it should be updated to MAXMIND; otherwise, no change should be made.

## Current Behavior
The field **db_vendor_type** is updated to MAXMIND on every execution--writing over any user supplied value.

## Possible Solution
Update the migration to include a filter for the field value before updating.

## Steps to Reproduce (for bugs)

1. Update the **Geo-Location Processor** configruation to **IPInfo** vendor type.
2. Restart the server and revisit the configuration--the vendor type will be MaxMind

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kingzacko1:
Looks like this issue was already taken care of. Closing.

PR #12061, opened 2022-01-28
Fixed bug with geo ip resolver config migration

roberto-graylog:
this PR addresses issue [#12060 ]
## Description

IF the field db_vendor_type is MAXMIND_CITY (old value), it should be updated to MAXMIND; otherwise, no change should be made.

## Motivation and Context

This change fixes a bug that could  overwrite user selection for **db_vendor_type**

## How Has This Been Tested?

Selected the alternative option (IPInfo) for db vendor type and restarted the server to confirm the value was not overwritten.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
@@ -50,9 +50,9 @@ public class MigrationsModule extends PluginModule {
         addMigration(V20200722110800_AddBuiltinRoles.class);
         addMigration(GrantsMetaMigration.class);
         addMigration(V20201103145400_LegacyAuthServiceMigration.class);
-        addMigration(V20211221144300_GeoIpResolverConfigMigration.class);
+        addMigration(V202201281638_GeoIpResolverConfigMigration.class);
 
         // Make sure there is always a binder for migration modules
-        Multibinder.newSetBinder(binder(), V20211221144300_GeoIpResolverConfigMigration.class);
+        Multibinder.newSetBinder(binder(), V202201281638_GeoIpResolverConfigMigration.class);
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
@@ -23,6 +23,7 @@ import com.mongodb.client.model.Updates;
 import com.mongodb.client.result.UpdateResult;
 import org.bson.Document;
 import org.bson.conversions.Bson;
+import org.graylog.plugins.map.config.DatabaseType;
 import org.graylog.plugins.map.config.DatabaseVendorType;
 import org.graylog.plugins.map.config.GeoIpResolverConfig;
 import org.graylog2.database.MongoConnection;
@@ -32,9 +33,17 @@ import org.slf4j.LoggerFactory;
 import javax.inject.Inject;
 import java.time.ZonedDateTime;
 
-public class V20211221144300_GeoIpResolverConfigMigration extends Migration {
+/**
+ * The Original migration had a bug which cause the <b>db_vendor_type</b>  field to always be updated--replacing any user supplied values.
+ *
+ * <p>
+ * This update provides a filter when updating the <b>db_vendor_type</b> field such that it is only updated if the current value is
+ * <i>MAXMIND_CITY</i>--this is the old/original  value which we want to replace with the default MAXMIND.
+ * </p>
+ */
+public class V202201281638_GeoIpResolverConfigMigration extends Migration {
 
-    private static final Logger LOG = LoggerFactory.getLogger(V20211221144300_GeoIpResolverConfigMigration.class);
+    private static final Logger LOG = LoggerFactory.getLogger(V202201281638_GeoIpResolverConfigMigration.class);
     private static final String COLLECTION_NAME = "cluster_config";
     public static final String PAYLOAD = "payload";
     private static final String FIELD_DB_VENDOR = PAYLOAD + ".db_vendor_type";
@@ -47,13 +56,13 @@ public class V20211221144300_GeoIpResolverConfigMigration extends Migration {
     private final MongoConnection mongoConnection;
 
     @Inject
-    public V20211221144300_GeoIpResolverConfigMigration(MongoConnection mongoConnection) {
+    public V202201281638_GeoIpResolverConfigMigration(MongoConnection mongoConnection) {
         this.mongoConnection = mongoConnection;
     }
 
     @Override
     public ZonedDateTime createdAt() {
-        return ZonedDateTime.parse("2021-12-21t14:43Z");
+        return ZonedDateTime.parse("2022-01-28t16:38Z");
     }
 
     /**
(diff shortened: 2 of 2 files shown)


danotorrey:
@roberto-graylog This issue only happens when the migration runs each time the server starts _after_ the initial upgrade was successful, correct? 

Another simple solution is to ensure that the migration only runs once and is skipped in subsequent startups. This would ensure that the subsequent upgrade issue does not happen since it would only run once. This can be accomplished by writing a custom `MigrationCompleted` object to the `ClusterConfigService` and check if it's present in subsequent upgrades as is done in [this example](https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboards/V20200204122000_MigrateUntypedViewsToDashboards.java#L70). I think this should probably be done anyhow to avoid extra processing logic/writes to mongo on each server startup and also provides a record that the migration has run.  What do you think?

roberto-graylog:
> @roberto-graylog Great find. This issue only happens when the migration runs each time the server starts _after_ the initial upgrade was successful, correct?
> 
> Another simple solution is to ensure that the migration only runs once and is skipped in subsequent startups. This would ensure that the subsequent upgrade issue does not happen since it would only run once. This can be accomplished by writing a custom `MigrationCompleted` object to the `ClusterConfigService` and check if it's present in subsequent upgrades as is done in [this example](https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboards/V20200204122000_MigrateUntypedViewsToDashboards.java#L70). I think this should probably be done anyhow to avoid extra processing logic/writes to mongo on each server startup and also provides a record that the migration has run. What do you think?

That is a great idea.  I have made the update.

I wonder though, in a later code change, could/should we do this in a parent class:

```Java
@Override public void upgrade(){
   MigrationStatus status = ....
   if(status != null){
     LOG.info(....);
  }else{
   doMigration();

   clusterConfigService.write(MigrationStatus.create(....));
  }
}

protected abstract void doMigration();
```

The status record would need to take the migration class name to identify each migration, so some further complexity.  However, this would assure consistent and out-of-the-box migration status management.

bernd:
> The status record would need to take the migration class name to identify each migration, so some further complexity. However, this would assure consistent and out-of-the-box migration status management.

@roberto-graylog Unfortunately, this wouldn't work with our current usage of migrations. Some migrations expect to run every time instead of only once, so they don't write any migration status to the database. The problem is that we mix up database changes and ensure that some records exist in the database. At one point, we have to refactor the migration system to split it up into "real" migrations only and "default content" delivery.

graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
```diff
@@ -66,6 +84,12 @@ public ZonedDateTime createdAt() {
     @Override
     public void upgrade() {
 
+        MigrationCompletion completion = clusterConfigService.get(MigrationCompletion.class);
+        if (completion != null) {
+            LOG.info("Migration was already completed on '{}'", completion.completionDate());
```
danotorrey, line 89: This works as expected to avoid running the migration multiple times. 👍  Nicely done. 

Can the log statement be changed to `debug` level? Many migrations are being skipped in this way after only running once. The pattern is usually to silently skip them, to avoid many "migration completed" log entries on server start up time, which could be distracting for the user.

I like the idea of capturing the date/time the migration was run, but the migration completion object in Mongo already contains the "last_updated" date, which is the same as the date when the migration has run. If the migration date ever needed to be identified, Mongo could be queried to find it. I think the payload `completion_date` can probably be removed since it seems redundant (and I don't think it's that important to include in the debug log statement). What do you think?
```
{
        "_id" : ObjectId("61f92d22bc3f6cb2a9369c51"),
        "type" : "org.graylog2.migrations.V202201281638_GeoIpResolverConfigMigration.MigrationCompletion",
        "payload" : {
                "completion_date" : ISODate("2022-02-01T12:52:50.437Z")
        },
        "last_updated" : ISODate("2022-02-01T12:52:50.438Z"),
        "last_updated_by" : "e065896b-8a9a-4f45-83f2-e740525ed035"
}
```

How about capturing the original GeoIp configuration and the resulting configuration after running the migration in the MigrationCompleted object instead? That would provide some audit history for which operations the migration actually completed. If that requires too much work (looks like the migration currently only performs the set operations without a query), I also think it's completely fine to write an empty MigrationCompleted marker object (with no fields).

graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
```diff
@@ -17,24 +17,39 @@
...
+ *
+ * <p>
+ * This update writes a completion status record upon initial successful migration, and checks for presence of such record
+ * before stating any subsequent migration--aborts if record found.
+ * </p>
+ */
+public class V202201281638_GeoIpResolverConfigMigration extends Migration {
 
-    private static final Logger LOG = LoggerFactory.getLogger(V20211221144300_GeoIpResolverConfigMigration.class);
+    private static final Logger LOG = LoggerFactory.getLogger(V202201281638_GeoIpResolverConfigMigration.class);
```
danotorrey, line 52: My original suggestion to rename the migration was to cause it to run again for any devs who have run it already, but that was invalid, because the migration was not previously writing can checking the `MigrationCompleted` object. Can the rename be reverted, since it's not needed? That might make the PR changeset more concise if it needs to be reviewed in the future.

danotorrey (changes requested):

graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
roberto-graylog: Reverted name.

graylog2-server/src/main/java/org/graylog2/migrations/V202201281638_GeoIpResolverConfigMigration.java
roberto-graylog: Updated log level to DEBUG and removed fields from migration status entity class.

roberto-graylog:
@janheise .  I merged your change to line ` Multibinder.newSetBinder(binder(), V20201103145400_LegacyAuthServiceMigration.MigrationModule.class);` of the MigrationsModule and added a more detailed comment as suggested by @bernd .  I also moved it to the top of the method such that all future migrations are at the end.

danotorrey (approved):
LGTM! Thanks @roberto-graylog. Looks like the checks have failed due to a transient error. Perhaps they will complete successfully on a second attempt.

Perhaps we can leave this open until tomorrow to see if @janheise has any feedback. Then, I think we can merge it.

graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
```diff
@@ -24,6 +24,12 @@
 public class MigrationsModule extends PluginModule {
     @Override
     protected void configure() {
+
+        // Make sure there is always a binder for migration modules
```
bernd, line 28: Let's remove the first line as suggested in my diff yesterday. I think it's too generic. Also, please add the second comment from my suggestion to the migration class. Thanks!

```diff
--- graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
+++ graylog2-server/src/main/java/org/graylog2/migrations/MigrationsModule.java
@@ -52,7 +52,8 @@ public class MigrationsModule extends PluginModule {
         addMigration(V20201103145400_LegacyAuthServiceMigration.class);
         addMigration(V20211221144300_GeoIpResolverConfigMigration.class);
 
-        // Make sure there is always a binder for migration modules
+        // The LegacyAuthServiceMigration is pluggable. This call ensures that there is always a binder present
+        // even if there are no plugins for this migration registered.
         Multibinder.newSetBinder(binder(), V20201103145400_LegacyAuthServiceMigration.MigrationModule.class);
     }
 }
--- graylog2-server/src/main/java/org/graylog2/migrations/V20201103145400_LegacyAuthServiceMigration.java
+++ graylog2-server/src/main/java/org/graylog2/migrations/V20201103145400_LegacyAuthServiceMigration.java
@@ -58,6 +58,7 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 public class V20201103145400_LegacyAuthServiceMigration extends Migration {
     private static final Logger LOG = LoggerFactory.getLogger(V20201103145400_LegacyAuthServiceMigration.class);
 
+    // This migration is pluggable to allow us to modify the outcome from plugins (e.g. Enterprise)
     public interface MigrationModule {
         void upgrade(Document document, AuthServiceBackendDTO authServiceConfig);
     }
```

PR #12074, opened 2022-02-01
Define more flexible widget export types.

linuspahl:
## Motivation and Context

For some widget export helper functions we are now only typing the data we are really using.
Before this change these functions expected a complete `Widget`.
With the less complex types they are more flexible because we do not need provide a complete widget to satisfy TypeScript.


dennisoelkers (approved):
:+1:

PR #12081, opened 2022-02-02
Allow overriding user timezone in `UserDateTimeProvider`.

linuspahl:
## Motivation and Context

We are using the `UserDateTimeContext` to display times based on the user time zone. In some cases it can be useful to be able to control this timezone. For example if you want to display all times on a page in a specific timezone.

By overriding the timezone in the `UserDateTimeProvider` we are changing the timezone for all displayed times.


dennisoelkers (approved):
:+1:

PR #12084, opened 2022-02-03
Implement safety net to prevent excessive rate of notifications

### LINKED ISSUE #7096: Event/Alert creation safety net
We should find a way to protect users from firing too many alerts.

I've seen a case, where a user created an Event definition with a email Notification that did not have a grace period set.
The definition resulted in 39 million events, where each of them triggered an email to be sent.
This stalled further event processing, because the scheduler was busy sending emails for
over three weeks.

Possible solutions: 
 - Don't allow the grace period to be disabled, but to be only to be set as low as one second.
 - Rate limit notification trigger creation to max  1000/minute?

thll:
Based on the suggestions in #7096 this PR provides the ability to clear the queue of pending notifications for an event.

There are two means of clearing the notification queue:

1. by disabling an event event
2. by triggering clearing of the notification queue manually through the UI

image

Previously, this PR also changed the default grace period from `0` to `1` seconds but after some thoughts and discussions, this change was reverted.

Fixes #7096


boosty:
Pinging @dulanism to check if this will need a change in the docs.

dulanism:
Yeah I think a few updates could be added to the respective page(s). Adding a ticket to our Kanban

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionHandler.java
```diff
@@ -287,9 +288,29 @@ private void deleteJobDefinition(JobDefinitionDto jobDefinition, EventDefinition
...
 
+    /**
+     * Deletes notification triggers for this event.
+     * This mostly servers as a safety net, when a misconfiguration has lead to a high number of outstanding
+     * notifications.
+     * Processing these notifications would unnecessarily block the job scheduler and in some cases consume a lot of
+     * resources.
+     */
+    private void deleteNotificationJobTriggers(EventDefinitionDto eventDefinition) {
+        final DBQuery.Query query = DBQuery.and(
```
mpfz0r, line 303: I think we should only delete runnable triggers here.
Otherwise a currently executed job might get confused.
Also I wonder how the scheduler likes it, if the jobdefinition goes away :thinking: 
But that would be an existing bug.

 sth like this should do it:
`                        DBQuery.is(FIELD_STATUS, JobTriggerStatus.RUNNABLE),`

mpfz0r (changes requested):
Some comments

mpfz0r:
> I'm a bit biased about 1. If I understand this correctly, a grace period means that we won't send out more than one notification for any number of events within this period. We also won't catch up on notifications we have not sent for these events after the grace period has elapsed.

Correct. With the small addition that this only happens for events with the same key.
But that doesn't change the fact that some notifications might be suppressed and lost.

> However, this means that if we have events that usually happen at a low frequency, but two noteworthy events happening within the grace period coincidentally, there will never be a notification triggered for the second event. Making this a default might cause users to overlook important events.

Yeah. I think we could be building something smarter.
First of all, we could be counting the number of suppressed notifications within a grace period.
This counter could be sent as a separate notification afterwards.
e.g. `Warning: 42 similar notifications have been skipped. Please check the event log if you want to make sure they aren't important`

We could also build some kind of rate limit, that only kicks in if we are creating too many notifications. (> 10 per second?)
Those could emit the same warning as mentioned above

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionHandler.java
mpfz0r: Maybe @bernd can answer that quicker, since he has a deeper knowledge about the job scheduler.
Otherwise I can also check the code tomorrow or next week

bernd: > Maybe @bernd can answer that quicker, since he has a deeper knowledge about the job scheduler. Otherwise I can also check the code tomorrow or next week

@thll and I talked about this two weeks ago. I also first said that we should only remove `RUNNABLE` triggers. But as @thll described, it might leave triggers around which are in a retry loop. We also don't skip `RUNNING` triggers when deleting or disabling an event definition.

So we thought it's okay to do it that way.

mpfz0r (approved):

thll:
@mpfz0r I've decided to only focus on the "clear notification queue" aspect in this PR.
I have reverted setting the default grade period to 1 second and added an explicit `clear` action next to the notification counter. Would you be so kind as to have a second look at the PR?

graylog2-server/src/main/java/org/graylog/events/rest/EventDefinitionsResource.java
```diff
@@ -209,6 +209,16 @@ public void unschedule(@ApiParam(name = "definitionId") @PathParam("definitionId
         eventDefinitionHandler.unschedule(definitionId);
     }
 
+    @PUT
+    @Path("{definitionId}/clear-notification-queue")
+    @Consumes(MediaType.WILDCARD)
+    @ApiOperation("Disable event definition")
```
mpfz0r, line 215:
```suggestion
    @ApiOperation("Clear queued notifications for event")
```

mpfz0r (changes requested):
one nitpick. rest looks good

graylog2-server/src/main/java/org/graylog/events/rest/EventDefinitionsResource.java
thll: oops, thanks for catching this!

mpfz0r (approved):

PR #12085, opened 2022-02-03
Only display margin between loading spinner icon and text when text is provided.

linuspahl:
## Motivation and Context

The `Spinner` component has a `margin-right` for the spinner icon, which creates space between the icon and the text.
With this PR we are not setting the margin, when there is no text.

### DIFF
--- a/graylog2-web-interface/src/components/common/Spinner.tsx
+++ b/graylog2-web-interface/src/components/common/Spinner.tsx
@@ -38,7 +38,7 @@ type Props = {
  */
 const Spinner = ({ name, text, delay, ...rest }: Props) => (
   <Delayed delay={delay}>
-    <StyledIcon {...rest} name={name} $displayMargin={!!text} spin />{text}
+    <StyledIcon {...rest} name={name} $displayMargin={!!text?.trim()} spin />{text}
   </Delayed>
 );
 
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/common/Spinner.tsx
```diff
@@ -38,7 +38,7 @@ type Props = {
  */
 const Spinner = ({ name, text, delay, ...rest }: Props) => (
   <Delayed delay={delay}>
-    <StyledIcon {...rest} name={name} spin />{text}
+    <StyledIcon {...rest} name={name} $displayMargin={!!text} spin />{text}
```
dennisoelkers, line 41: Does it make sense to check if `text` is blank as well?

dennisoelkers (changes requested):
Looks good! Just a minor remark.

graylog2-web-interface/src/components/common/Spinner.tsx
linuspahl: Can you describe the context a bit more, what do you mean with blank?

dennisoelkers: When for some reason `text?.trim() === ''`. Not a case that should necessarily happen, but we could catch it here just in case. It would not be visible, but maybe it could lead to unwantend indentation of the next element?

linuspahl: Now I get it. I think it is a good idea, when someone wants the indentation they should style the icon instead.

PR #12089, opened 2022-02-04
Add support for view resolvers for GET view API endpoint

danotorrey:
## Overview
Add support for custom view resolvers for the `GET view` API endpoint. This allows plugins to provide an optional set of view resolvers that will be used when a specific view (dashboard) is requested from the GET Views API endpoint. The endpoint will attempt to find the view in the bound resolvers first, before then using the default Mongo view lookup.

```
public class CustomResolver implements ViewResolver {

    public CustomResolver() {
    }

    @Override
    public Optional<ViewDTO> get(String id) {
        return // <-- Find view from custom source and provide it here.
    }

```

And finally, bind the resolver implementation here.
```
installViewResolver("security-dashboards", CustomResolver.class);
```

## Motivation and Context

See the linked PR below in the PR timeline for details.

## Types of changes

## Checklist:


dennisoelkers (commented):
I like the idea of a pluggable view resolver allowing plugins to contribute their own ways of accessing searches/dashboards. I was wondering if it would be beneficial to avoid using a chained execution of view resolvers. Instead, we could use an addressing scheme for views that looks like this:

```
<resolver>:<id>
```

E.g. it could be `security:overview`. The `<resolver>` part is then used to look up the corresponding resolver and use it to resolve the `<id>` part. If the id does not match the scheme, the default mongodb-resolver is used.

The benefits would be:
  - no id conflicts between resolvers
  - not having to think about error handling (what happens when one resolver in the chain throws an exception? continue? bail out?)
  - better performance due to avoiding multiple lookups

What do you think?

danotorrey:
> E.g. it could be `security:overview`. The `<resolver>` part is then used to look up the corresponding resolver and use it to resolve the `<id>` part. If the id does not match the scheme, the default mongodb-resolver is used.

@dennisoelkers Thanks for the suggestion. I really like this. I was really not pleased about the iteration through the resolvers (it's non-deterministic and does not handle multiple resolvers in a predicable way). I'll try this. I think it should work out.

Issue #11930, opened 2022-01-10
Improve rendering performance for high number of fields


dennisoelkers:
## Expected Behavior

The search should load and be usable for a high number of fields (in the order of n*10^3).

## Current Behavior

In some setups, the search page takes quite long to load if a lof of fields exist.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

boosty:
Related: 
- https://github.com/Graylog2/graylog2-server/issues/11682

dennisoelkers:
Update: General loading of searches/dashboards when many fields are present is improved through #12110, addressing the most pressing issue.
Remaining issues are:

  - selecting fields through our select component is rather slow when opening the select. A possible solution could be to use `react-select-virtualized`, which is a drop-in replacement for `react-select`
  - Expanding a message with a large number of fields can be slow. General performance improvements can help here.

PR #12110, opened 2022-02-09
Improve indexing field names performance for field name completion.

dennisoelkers:
## Motivation and Context

This PR is improving field names indexing performance in the field names completion class, which is happening every time the field types store is refreshed.

Prior to this PR, a `reduce` call was used, subsequently merging objects into each other. For large numbers of fields, this is creating excessive load. A simpler way using `Object.fromEntries` is now being used, which is creating the object from the given array in a linear way.

In addition, a couple of cleanups are performed. These should not be part of backport PRs.

Refs #11930.

## How Has This Been Tested?

As this issue results from users with performance issues for a high number of fields, I ingested messages adding up to a number of 25000 fields being present. After that, loading the initial search page was profiled. Before the change it took about 30s, afterwards less than 4s.

## Types of changes

## Checklist:


linuspahl (approved):
Works very well!

PR #12122, opened 2022-02-11
Use current time per default for the `Timestamp`, `RelativeTime` and `BrowserTime` component when `dateTime` prop is not defined.

linuspahl:
## Motivation and Context

We recently refactored the `Timestamp`, and implemented the `RelativeTime` and `BrowserTime` components to unify the way we display times in the UI.

One part of the refactoring was replacing the usage of the `logic/DateTime` class. The class constructor used the current time by default when no date time has been provided. With this PR we are now also using the current time by default for the mentioned components.

This is also fixing a bug, since the current implementation throws an error (`Date time null is not valid.`) when the date time for these components is not defined. This happened on the system overview page because the `started_at` field of system jobs can be null.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/BrowserTime.test.tsx
+++ b/graylog2-web-interface/src/components/common/BrowserTime.test.tsx
@@ -19,6 +19,12 @@ import { render, screen } from 'wrappedTestingLibrary';
 
 import BrowserTime from './BrowserTime';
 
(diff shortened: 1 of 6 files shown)


edmundoa (approved):
Looks good to me and fixes https://github.com/Graylog2/graylog2-server/issues/12140, great job! 👍

Issue #12131, opened 2022-02-14
Help tooltip for disabled share button can no longer be opened.


linuspahl:
# Description 

In some cases the share button is disabled. When the button is disabled we display an icon for a help text. When a user clicks on the help text, we open a tooltip which describes why the button is disabled.

![image](https://user-images.githubusercontent.com/46300478/153909158-c88d12a2-474e-4471-8f9b-cc4e551575ac.png)

This is the case on the default search page, until the search has been saved. Currently the tooltip can not be displayed.

* Graylog Version: current development version

PR #12150, opened 2022-02-16
Add props onClick, onMouseEnter, onMouseLeave, onFocus, tabIndex, title to Icon component

maxiadlovskii:
## Description
Fixes https://github.com/Graylog2/graylog2-server/issues/12131
Fixes https://github.com/Graylog2/graylog2-server/issues/12166
During investigation I found that Icon component doesn't expect some necessary props.  
So I checked in the code which props we try to use and add them to the Icon  component

## Motivation and Context
As component doesn't expect a lot of props we are not able to handle events when we use Icon component
[](https://github.com/Graylog2/graylog2-server/issues/12131 )

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/common/Icon.tsx
+++ b/graylog2-web-interface/src/components/common/Icon.tsx
@@ -14,11 +14,11 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
+import type { SyntheticEvent } from 'react';
 import React from 'react';
 import { find } from 'lodash';
 import type { IconName } from '@fortawesome/fontawesome-common-types';
 import type { SizeProp } from '@fortawesome/fontawesome-svg-core';
-import type { FontAwesomeIconProps } from '@fortawesome/react-fontawesome';
 
 import deprecationNotice from 'util/deprecationNotice';
 import loadAsync from 'routing/loadAsync';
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/common/Icon.tsx
```diff
@@ -78,7 +79,13 @@ type Props = {
   type?: IconTypes,
   fixedWidth?: boolean,
   inverse?: boolean,
-  style?: React.CSSProperties
+  style?: React.CSSProperties,
+  onClick?: (event: SyntheticEvent) => void,
+  onMouseEnter?: (event: SyntheticEvent) => void,
+  onMouseLeave?: (event: SyntheticEvent) => void,
+  onFocus?: (event: SyntheticEvent) => void,
```
linuspahl, line 86: These type definitions are correct, but we can define slightly more specific type definitions here:
```
  onClick?: (event: React.MouseEvent<SVGSVGElement>) => void,
  onMouseEnter?: (event: React.MouseEvent<SVGSVGElement>) => void,
  onMouseLeave?: (event: React.MouseEvent<SVGSVGElement>) => void,
  onFocus?: (event: React.FocusEvent<SVGSVGElement>) => void,
```

linuspahl (approved):
Looks good to me!

Issue #11109, opened 2021-08-03
Add option to GROK import to skip duplicates


xbb:
Importing a GROK pattern file in Grok patterns while selecting "Replace all existing patterns?" removes
all existing patterns.

## Expected Behavior
Selecting "Replace all existing patterns?" should replace only the patterns with the same name.

## Current Behavior
It deletes all the existing patterns and imports the one in the file.

I'm not sure if this is intended..., also not selecting the option it will fail if pattern with the same name exist.

## Steps to Reproduce (for bugs)
1. Have default GROK patterns installed
2. Import a GROK pattern file selecting "Replace all existing patterns"
   ```
   DATA .*?
   ```

## Your Environment

* Graylog Version: 4.1.2
* Java Version: openjdk 1.8.0_292
* Elasticsearch Version: 6.8.18-1
* MongoDB Version: 3.6.23
* Operating System: CentOS 7

bernd:
@xbb We renamed the title and converted the issue to a feature request for an option to ignore existing patterns on import.

ihe-nu:
![image](https://user-images.githubusercontent.com/74296413/154983410-33025d15-e809-45b4-bca4-91491f9e3812.png)
To me, the text "Replace all existing patterns" could be both:

- purge the database from existing Groks and only take those from this new file
- if a grokpattern from this file already exists replace the one in Graylog with the one from the file
Both variants are very different and it is from my point of view not clear for a user what happens.

I agree on @xbb that this is a bug, at least it is highly misleading. If you are looking with the expectancy of the second possibility you will clear your grok-collection (as it just happened to me) and be frustrated.

For the short therm I'd suggest to make the adjust the text to a more specific description to what really happens. Having both options would be a feature request for the long therm.

tellistone:
This impacted a customer under HS-787535168, caused them to accidentally delete all their patterns.

Renaming the text "Replace all existing patterns" to "drop all existing 
patterns before import" would be much clearer.

PR #12174, opened 2022-02-22
Allow bulk update of Grok patterns

thll:
Adds an option to update Grok patterns on bulk import if they already exist.

This effectively provides three options when performing bulk imports:
1. Abort import if there is a name conflict with any of the patterns already present in graylog. This is the default.
2. If there is a name conflict, update the pattern with that name in graylog with the new pattern provided in the import.
3. Drop all existing patterns in graylog before importing the new set of patterns.

The dialog for importing patterns now shows these options.

Before:
image

After:
image

Fixes https://github.com/Graylog2/graylog2-server/issues/11109

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/grok/GrokPatternService.java
+++ b/graylog2-server/src/main/java/org/graylog2/grok/GrokPatternService.java
@@ -44,7 +44,7 @@ public interface GrokPatternService {
 
     GrokPattern update(GrokPattern pattern) throws ValidationException;
 
-    List<GrokPattern> saveAll(Collection<GrokPattern> patterns, boolean replace) throws ValidationException;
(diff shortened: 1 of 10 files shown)


thll:
@boosty can you please check out the screenshots in the description and give some feedback? I'd like to know if this is going in the right direction.

boosty:
@thll The three options and the wording are really good in my opinion. I think @tellistone will like this, too. Thanks a lot!

kmerz (approved):
Code Looks good for me! :+1: 
- Downloaded the tar.gz
- Imported new grok patterns with "Abort of duplicated"
- Imported them again with "Abort on duplicated" which led to expected error.
- Imported them again with "Replace already existing" which worked as expected
- Imported with "Drop all" and that left only the newly imported as expected

PR #12176, opened 2022-02-22
Enable `noUnusedParameters` and `noUncheckedIndexedAccess` for TypeScript.

dennisoelkers:
## Motivation and Context

This change is making the TypeScript config gradually stricter by enabling:

  - [noUnusedParameters](https://www.typescriptlang.org/tsconfig#noUnusedParameters)
  - [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess)

and fixing the resulting fallout.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/authentication/directoryServices/BackendWizard/BackendWizard.tsx
+++ b/graylog2-web-interface/src/components/authentication/directoryServices/BackendWizard/BackendWizard.tsx
@@ -62,7 +62,7 @@ const SubmitAllError = ({ error, backendId }: { error: FetchError, backendId: st
 
 const _formatBackendValidationErrors = (backendErrors: { [inputNameJSON: string]: string[] }) => {
   const backendErrorStrings = mapValues(backendErrors, (errorArray) => `Server validation error: ${errorArray.join(' ')}`);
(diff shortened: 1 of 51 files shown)


linuspahl (commented):
Looks good to me in general! One suggestion, there are many cases where we need to deine a parameter which is unused. Sometimes we just use `_` as the variable name, sometimes a more meaningful one, like `_theName`.

I prefer to keep this consistent. Although the attribute name can be inferred in [many cases](https://github.com/Graylog2/graylog2-server/pull/12176/files#diff-0dd8d3e714ead2344d3d049d95467ad7851e8080fe8b79d910647a57732d7b44L58):
image 

there are [some](https://github.com/Graylog2/graylog2-server/pull/12176/files#diff-6e40d8972b2f779a35047d026d5bcaf957204beab405c6bba2a3e58d6867281dR49) where it is not the case: 
image

When we agree on always using the more meaningful name like `_theName`, it would make it a little bit easier to understand the code and we do not need to consider if the attribute name can be inferred.

dennisoelkers:
@linuspahl: Good observation! I adjusted related variable names in this PR. I would also suggest using a variable name prefixed with an underscore as a best practice, but I would not consider it an issue e.g. when doing a PR review. In some cases it might be better to leave out the variable name than guessing it.

PR #12179, opened 2022-02-22
Improve Periodical exception handling

mpfz0r:
A periodical implementation could return null
for getLogger() and that would abort a scheduled periodical forever.

Mark getLogger() as @Nonnull and also provide a fallback logger.


thll (approved):
Good catch!

PR #12185, opened 2022-02-23
Ensure that widget keeps id when moved to another dashboard page.

dennisoelkers:
## Motivation and Context

This change is making sure that when a widget is moved to a different dashboard page (without copying it), it will keep its id. This avoids any issues with referential integrity when the widget is referenced somewhere else by id (e.g. reporting).

Nothing changes when the widget is copied instead of being moved. The source widget will stay as is (keeping its id) and the new widget will get a new, random id. This way any references will stay intact, referencing the old widget.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me!

PR #12194, opened 2022-02-24
Support credentials in URLs for MS Edge.

dennisoelkers:
**Note:** This should be backported to `4.1` as well.

## Motivation and Context

Prior to this change, when generating URLs from the frontend that directly link to the API (e.g. for downloading exports, reports, etc.), it was checked if the browser supports embedding crendentials in the URL. This check was written at a time when MS Edge was not based on Chromium yet. Therefore it was assuming that it does not support it.

This has changed, therefore the check for Edge is removed. The whole logic will vanish with the introduction of session cookies, therefore a further revamp is not necessary.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me! Tested it with Microsoft Edge and it works as expected.

PR #12221, opened 2022-03-02
Wait for node balancer recognition period before triggering "HALTING" lifecycle

thll:
During shutdown, at first, set the load balancer status to "DEAD" and then pause for the configured load balancer recognition period. Only after that trigger the lifecycle change to "HALTING".

Previously, we would start with triggering the lifecycle change to "HALTING" prior to pausing for load balancer recognition. This would set the load balancer status to "DEAD" but at the same time, any service listening for the lifecycle change to "HALTING" would already take action and potentially shut down.

While in most cases this shouldn't be a problem, it is e.g. for the server-side of the cloud forwarder which runs as a service and stops accepting messages when server shutdown is detected. With this change, the load balancer status is set now set to "DEAD" ahead of time.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/system/shutdown/GracefulShutdown.java
+++ b/graylog2-server/src/main/java/org/graylog2/system/shutdown/GracefulShutdown.java
@@ -39,7 +39,6 @@ import static org.graylog2.audit.AuditEventTypes.NODE_SHUTDOWN_COMPLETE;
 @Singleton
 public class GracefulShutdown implements Runnable {
     private static final Logger LOG = LoggerFactory.getLogger(GracefulShutdown.class);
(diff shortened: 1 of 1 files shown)


mpfz0r (approved):
LGTM. 

I've tested this with a `grpcurl` and `httpie` loop.
In my setup 
```
while sleep .1; do grpcurl -d '{}' -rpc-header 'api-token: my-token' -plaintext -import-path ~/code/graylog-cloud-proj/graylog-project-repos/graylog-plugin-enterprise/forwarder-commons/src/main/proto -proto services.proto  localhost:13301 org.graylog.cloud.forwarder.ForwarderMessageForwardingService/ForwardMessages  ; ghttp https://localhost:9000/api/system/lbstatus ; done
```

PR #12224, opened 2022-03-03
Hotfix for never-expiring session validity #12220

### LINKED ISSUE #12220: Can't log in with "sessions do not timeout" option
Can't log in if I configure the _sessions do not timeout_ option. When I re-activate the session timeouts again all is working as expected. 

image

## Expected Behavior
I should be able to log in and the session should be valid forever. 

## Current Behavior
Login is not possible

1. Go to user administration
2. Check the session do not timeout checkbox 
3. Logout
4. Try to log-in again.

## Your Environment

* Graylog Version: Graylog 4.3.0-SNAPSHOT+01253df on server (Eclipse Adoptium 17.0.2 on Linux 5.11.0-1028-aws)

todvora:
With the implementation of the cookie auth in #11788 we encountered a bug #12220 that prevents log-in for never expiring sessions. 

Cookies generally don't allow unlimited TTL, they always require a `maxAge` value. The common workaround is to set this value to the very distant future. This PR does exactly that - for never-expiring session configuration it sets the cookie to be valid for 10 years from the moment of login. It's rather a workaround and we should think about removing the non-expiring functionality generally, as we can't implement it correctly.

## Motivation and Context
Fixes #12220 

## How Has This Been Tested?
**TODO** add an integration test!

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/DefaultSessionResponseFactory.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/DefaultSessionResponseFactory.java
@@ -50,7 +50,7 @@ public class DefaultSessionResponseFactory implements SessionResponseFactory {
 
     protected Date getValidUntil(Session session) {
         if(session.getTimeout() < 0) { // means "session never expires", which is not possible in cookie-based auth
-            return new DateTime().plus(Years.years(10)).toDate(); // careful, later we convert the date to seconds as int and it may overflow for too big values
+            return new DateTime(DateTimeZone.UTC ).plus(Years.years(10)).toDate(); // careful, later we convert the date to seconds as int and it may overflow for too big values
         } else {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/rest/models/system/sessions/responses/DefaultSessionResponseFactory.java
```diff
@@ -48,7 +49,11 @@ public SessionResponse forSession(Session session) {
     }
 
     protected Date getValidUntil(Session session) {
-        return new DateTime(session.getLastAccessTime(), DateTimeZone.UTC).plus(session.getTimeout()).toDate();
+        if(session.getTimeout() < 0) { // means "session never expires", which is not possible in cookie-based auth
```
Corlobin, line 52: Maybe add a space between if and `(`
```suggestion
        if (session.getTimeout() < 0) { // means "session never expires", which is not possible in cookie-based auth
```

thll (approved):
Straight-forward and fixes the issue 👍

linuspahl (approved):

PR #12225, opened 2022-03-03
Disable search highlighter for scroll queries

mpfz0r:
Our use cases for scroll queries are:
 - Archinving
 - Alerting
 - CSV log export

Unconditionally activating highlighting makes no sense,
is more expensive and might also cause errors on big documents,
if they exceed `index.highlight.max_analyzed_offset`

This is only needed for ES7 / OpenSearch support.
Our ES6 client never had scroll highlighting enabled.


todvora (approved):
Looks good, thanks!

PR #12227, opened 2022-03-03
Improve fetching of query field value suggestions.

### LINKED ISSUE #12165: Using "Add to query" field option opens value suggestion
When using the "Add to query" field option to modify the query, an overlay with a field value suggestion pops up. It stays in a fixed position when scrolling and only disappears when clicking into the search bar.

Example:

https://user-images.githubusercontent.com/461/154734773-1bbfa5fe-ce13-4bc2-bfae-d448c252c60f.mp4

1. Open message details
2. Select "Add to query" on a message field value
3. Wait and observe the value suggestion popping up

## Your Environment

* Graylog Version: 4.3.0-SNAPSHOT (rev 7aee4be)

linuspahl:
## Motivation and Context

In some cases we display suggestions for field values in the search query, e.g. for query strings like:
`http_method:` and `http_method:POS`

After every query string change we check if we should open the query input autocomplete (we are calling `shouldShowCompletions` of every defined completer class). Opening the autocomplete triggers the actual fetching of completions (we are calling `getCompletions` of every defined completer class).

This PR is improving the logic by:
- not calling `getCompletions` on the initial render
- making sure `shouldShowCompletions` of the `FieldValueCompleter` does not return true for a query string  like `http_method:POST` where `http_method` is the currently selected token.

This change is fixing https://github.com/Graylog2/graylog2-server/issues/12165. In this case the problem occurred when using the field value action "Add to query" (only) while the query string is empty. In this case the ace editor behaves unexpected, because it defines the added field name as the currently selected token. When using the field value action "Add to query" while the query string is not empty the field value will be defined as the current token.

Fixes #12165

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.test.ts
+++ b/graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.test.ts
@@ -356,11 +356,20 @@ describe('FieldValueCompletion', () => {
       expect(result).toEqual(false);
     });
 
-    it('returns true if current token is a keyword and ends with :', async () => {
+    it('returns true if current token is a keyword and ends with ":"', async () => {
       const completer = new FieldValueCompletion();
       const result = completer.shouldShowCompletions(1, [[{ type: 'keyword', value: 'http_method:', index: 0, start: 0 }, null]]);
 
       expect(result).toEqual(true);
+    });
+
+    it('returns false if current token is a keyword that ends with ":" which is already followed by a term', async () => {
+      const completer = new FieldValueCompletion();
+      const result = completer.shouldShowCompletions(1, [[
+        { type: 'keyword', value: 'http_method:', index: 0, start: 0 },
+        { type: 'term', value: 'POST' }, null]]);
+
+      expect(result).toEqual(false);
     });
 
     it('returns true if current token is a keyword in a complex query and ends with :', async () => {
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/views/components/searchbar/QueryInput.tsx
```diff
@@ -156,14 +163,15 @@ const QueryInput = ({
 }: Props) => {
   const { data: queryFields } = useFieldTypes(streams, isNoTimeRangeOverride(timeRange) ? DEFAULT_TIMERANGE : timeRange);
   const { data: allFields } = useFieldTypes([], DEFAULT_TIMERANGE);
+  const isInitialTokenizerUpdate = useRef(true);
```
maxiadlovskii, line 166: If you don't need this ref in your JSX code you can do I other way.

```
const [isInitialTokenizerUpdate, setIsInitialTokenizerUpdate ]= useState(true);
...
const onLoadEditor = useCallback((editor: Editor) => _onLoadEditor(editor, setIsInitialTokenizerUpdate), [ setIsInitialTokenizerUpdate ]);
... 

const _onLoadEditor = (editor, setIsInitialTokenizerUpdate: React.MutableRefObject<boolean>) => {
  if (editor) {
    editor.commands.removeCommands(['find', 'indent', 'outdent']);

    editor.session.on('tokenizerUpdate', (input, { bgTokenizer: { currentLine, lines } }) => {
      setIsInitialTokenizerUpdate( isInitialTokenizerUpdate => {
            if (isInitialTokenizerUpdate.current) {
               return true
            }
      })
        editor.completers.forEach((completer) => {
          if (completer?.shouldShowCompletions(currentLine, lines)) {
            editor.execCommand('startAutocomplete');
          }
        });    
        return isInitialTokenizerUpdate
        });

    editor.renderer.setScrollMargin(6, 5);
    editor.renderer.setPadding(12);
  }
};
```

linuspahl: What would be the benefit? I prefer using a ref in this case to avoid triggering a state update.

Issue #10869, opened 2021-06-16
ES Cluster health not shown correctly


janheise:
## Expected Behavior
I expect that the ES Cluster health changes when I stop ES in the backround.
![ES](https://user-images.githubusercontent.com/1190735/122213821-a9e3b700-cea9-11eb-9fbb-2c92d5c2cbb6.png)

## Current Behavior
I stop the ES instances in docker (see the attached docker-file) in the background until none is left but the health status in graylog never changes.

## Steps to Reproduce (for bugs)

1. use the docker config to start graylog/es/mongo
2. use docker ps to get the id's of the ES instances
3. docker stop <es instance>
4. check the GUI for the health status of ES

[docker-compose.txt](https://github.com/Graylog2/graylog2-server/files/6662762/docker-compose.txt)

## Your Environment

* Graylog Version: 4.1.0-beta1
* Java Version:
* Elasticsearch Version: 7.10.2 with Cluster, 6.8.3 as Single Instance
* MongoDB Version:
* Operating System:
* Browser version:

luk-kaminski:
Seems like back0end. Probably "indicesExist()" method called outside of try-catch block in ClusterAdapters. Will try to have a look tomorrow.

luk-kaminski:
While some minor changes have been made to BE in #12233 PR, the main job needs to be done in FE.
It has to react properly when "{"type":"ApiError","message":"Couldn't read Elasticsearch cluster health"}" response is returned from "api/system/indexer/cluster/health" call.

todvora:
![image](https://user-images.githubusercontent.com/4102775/157042252-b753bdda-d1b5-436b-ace8-f2dbcbdeff0f.png)

This is the current state of FE after the BE fix from @luk-kaminski. In the FE we need to catch the error and display a proper error message.

maxiadlovskii:
<img width="1202" height="167" alt="Image" src="https://github.com/user-attachments/assets/413b0a88-73b4-402c-8c1e-313f9953e2b4" />
it's already fixed

PR #12241, opened 2022-03-08
Move cluster health data to local store. Add is loading state

maxiadlovskii:
#10869 

## Description
Move data for cluster store to local state as there is no usage in other components. Add loading state for the correct handling Loader state. Add error state. 

## Motivation and Context
In case the user has a problem with elastic search, we didn't handle the error state. This PR fixes this problem. 

## Types of changes

## Checklist:


graylog2-web-interface/src/components/indexers/IndexerClusterHealthError.tsx
```diff
@@ -0,0 +1,37 @@
...
+const ESClusterError = styled(Alert)`
+  margin-top: 10px;
+  margin-bottom: 5px;
+`;
+
+const IndexerClusterHealthError = ({ error }: { error: Error }) => {
+  return (
+    <ESClusterError bsStyle="danger">
+      <Icon name="exclamation-triangle" /> &nbsp;{'Could not retrieve Elasticsearch cluster health'}{' '}
+      {`Fetching Elasticsearch cluster health failed: ${error.message}`}
```
linuspahl, line 32: We do not need use a template string here.

graylog2-web-interface/src/components/indexers/IndexerClusterHealthError.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { Alert } from 'components/bootstrap';
+import { Icon } from 'components/common';
+
+const ESClusterError = styled(Alert)`
+  margin-top: 10px;
+  margin-bottom: 5px;
+`;
+
+const IndexerClusterHealthError = ({ error }: { error: Error }) => {
```
linuspahl, line 28: I do not think this is a JS runtime error.

graylog2-web-interface/src/components/indexers/IndexerClusterHealth.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const useLoadHealthAndName = () => {
+  const [state, setState] = useState({ health: undefined, name: undefined, error: undefined, loading: false });
+
+  useEffect(() => {
+    setState({ health: undefined, name: undefined, error: undefined, loading: true });
+
+    Promise.all([
+      IndexerClusterActions.health(),
+      IndexerClusterActions.name(),
```
linuspahl, line 36: Since we fetch the name, let's display it in the UI (when it exists), although this has not been the case before.
It looks like this is only possible, when the cluster is healthy. We could include it in `Elasticsearch cluster "the-name" is green ..`.

linuspahl (commented):
Works very well! One thought I had while looking at the code: Since we are only using the `IndexerClusterActions` in this component we can probably remove the `IndexerClusterStore` and fetch the data with `react-query` instead. This way it will also be easier to refetch the data periodically, like every 5 seconds. Which would be a good idea for this information.

graylog2-web-interface/src/components/indexers/IndexerClusterHealth.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <Row className="content">
+      <Col md={12}>
+        <h2>Elasticsearch cluster</h2>
+
+        <SmallSupportLink>
+          The possible Elasticsearch cluster states and more related information is available in the{' '}
+          <DocumentationLink page={DocsHelper.PAGES.CONFIGURING_ES} text="Graylog documentation" />.
+        </SmallSupportLink>
+        {isSuccess && <IndexerClusterHealthSummary health={health} name={name} />}
+        {loading && <Spinner />}
```
linuspahl, line 62: Can you Improve the styling of the loading indicator, for example by displaying it in a new row?
image

graylog2-web-interface/src/components/indexers/IndexerClusterHealth.tsx
```diff
@@ -0,0 +1,69 @@
...
+import IndexerClusterHealthError from './IndexerClusterHealthError';
+
+const useLoadHealthAndName = () => {
+  const [
+    { data: healthData, isFetching: healthIsFetching, error: healthError, isSuccess: healthIsSuccess },
+    { data: nameData, isFetching: nameIsFetching, error: nameError, isSuccess: nameIsSuccess },
+  ] = useQueries([
+    { queryKey: GET_INDEXER_CLUSTER_HEALTH, queryFn: getIndexerClusterHealth },
+    { queryKey: GET_INDEXER_CLUSTER_NAME, queryFn: getIndexerClusterName },
+  ]);
```
linuspahl, line 38: Now where we use `react-query` let's define a refetch interval, like `{ refetchInterval: 5000, retry: 0 }`

This way we do not have multiple immediate retries, when the cluster is down and we are refetching the status every 5 seconds, which will update the status, when it changes while the page for the cluster health is open.

graylog2-web-interface/src/logic/reactQueryKeys.ts
```diff
@@ -0,0 +1,19 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+
+export const GET_INDEXER_CLUSTER_HEALTH = 'GET_INDEXER_CLUSTER_HEALTH';
+export const GET_INDEXER_CLUSTER_NAME = 'GET_INDEXER_CLUSTER_NAME';
```
linuspahl, line 19: I like the idea to separate the API calls and the query keys in general, but I would suggest to move both into the component where we use `react-query`, to keep it consistent with the other cases where we use `react-query`.

linuspahl (changes requested):

linuspahl (approved):
Looks good to me, fixes the bug described in #10869.

PR #12242, opened 2022-03-08
Using proper time range when search has not results.

### LINKED ISSUE #12236: Improve format of effective time search time range when "all time" range has no results.
## Current Behavior

When a user executes a search with the time range  "All time" and the search result is empty, the backend returns an effective time range like:
```
from: "292278994-08-17T07:12:55.807Z"
to: "-292275055-05-16T16:47:04.192Z"
```

as a result widgets like the bar chart display the following time range: 

image

We should make sure we return a reasonable time range in this case.

Related to https://github.com/Graylog2/graylog2-server/issues/12193.

* Graylog Version: 4.2

luk-kaminski:
## Description
Fixes #12236.
After this change, we start using proper time range in the search response when search has not results.
Aggregations are not used to fetch time range in that situation, but time range from the pivot is used instead.

## Motivation and Context
It has been discovered in #12236.

When we have no results in search response, we still used to obtain time range from aggregations, namely **"timestamp-min"** and **"timestamp-max"**. Those aggregations behaved differently in different versions of ES. The most problematic approach was visible in ES7, when they contained +/- Infinity, resulting in extremely strange time range used:

> from: "292278994-08-17T07:12:55.807Z"
> to: "-292275055-05-16T16:47:04.192Z"

In ES6 everything seemed to work, but I changed the code there to have the same approach used in both places.

## How Has This Been Tested?
1. Go to search.
2. Create a search that will return no results, either by using proper input time range, stream or query text itself.
3. The time range in Message Count widget should match the one from the input, even if there are not results.

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/views/searchtypes/pivots/ESPivotTest.java
+++ b/graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/views/searchtypes/pivots/ESPivotTest.java
@@ -376,12 +376,10 @@ public class ESPivotTest {
     }
 
     @Test
-    public void searchResultForAllMessagesIncludesPivotTimerangeForNoDocuments() throws InvalidRangeParametersException {
+    public void searchResultForAllMessagesIncludesPivotTimerangeForNoDocuments() {
         when(queryResult.getTotal()).thenReturn(0L);
-        final TimeRange pivotRange = mock(TimeRange.class);
+        final TimeRange pivotRange = AbsoluteRange.create(DateTime.parse("1970-01-01T00:00:00.000Z"), DateTime.parse("2020-01-09T15:44:25.408Z"));
         when(query.effectiveTimeRange(pivot)).thenReturn(pivotRange);
-        when(pivotRange.getFrom()).thenReturn(DateTime.parse("1970-01-01T00:00:00.000Z"));
-        when(pivotRange.getTo()).thenReturn(DateTime.parse("2020-01-09T15:44:25.408Z"));
         final SearchType.Result result = this.esPivot.doExtractResult(job, query, pivot, queryResult, aggregations, queryContext);
         final PivotResult pivotResult = (PivotResult) result;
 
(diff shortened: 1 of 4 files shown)


graylog-storage-elasticsearch6/src/test/java/org/graylog/storage/elasticsearch6/views/searchtypes/pivots/ESPivotTest.java
```diff
@@ -376,15 +377,12 @@ public void searchResultForAllMessagesIncludesTimerangeOfDocuments() throws Inva
...
     public void searchResultForAllMessagesIncludesPivotTimerangeForNoDocuments() throws InvalidRangeParametersException {
-        DateTimeUtils.setCurrentMillisFixed(1578584665408L);
-        final long documentCount = 0;
-        when(queryResult.getTotal()).thenReturn(documentCount);
-        final MetricAggregation mockMetricAggregation = createTimestampRangeAggregations(null, null);
-        when(queryResult.getAggregations()).thenReturn(mockMetricAggregation);
-        when(query.effectiveTimeRange(pivot)).thenReturn(RelativeRange.create(0));
-
+        when(queryResult.getTotal()).thenReturn(0L);
+        final TimeRange pivotRange = mock(TimeRange.class);
```
todvora, line 381: Could we avoid the Mock here and use a one of the TimeRange implementations?

graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/views/searchtypes/pivots/ESPivotTest.java
```diff
@@ -384,20 +385,17 @@ public void searchResultForAllMessagesIncludesTimerangeOfDocuments() throws Inva
...
     public void searchResultForAllMessagesIncludesPivotTimerangeForNoDocuments() throws InvalidRangeParametersException {
-        DateTimeUtils.setCurrentMillisFixed(1578584665408L);
-        final long documentCount = 0;
-        returnDocumentCount(queryResult, documentCount);
-        final Aggregations mockMetricAggregation = createTimestampRangeAggregations(0d, 0d);
-        when(queryResult.getAggregations()).thenReturn(mockMetricAggregation);
-        when(query.effectiveTimeRange(pivot)).thenReturn(RelativeRange.create(0));
-
+        returnDocumentCount(queryResult, 0);
+        final TimeRange pivotRange = mock(TimeRange.class);
```
todvora, line 389: Same as above, could we please replace the Mock with real implementation?

todvora (changes requested):

todvora (approved):
Looks good, tested locally. Thx!

PR #12245, opened 2022-03-08
Disable search highlighter for scroll queries (#12225) (4.1 backport)

mpfz0r:
Our use cases for scroll queries are:
 - Archinving
 - Alerting
 - CSV log export

Unconditionally activating highlighting makes no sense,
is more expensive and might also cause errors on big documents,
if they exceed `index.highlight.max_analyzed_offset`

(cherry picked from commit 3bb5207c60bfdd8c0f31886eb499d6a6dddd9a48)


thll (approved):
Looks good. Confirmed that archiving still works with a smoke test.

PR #12247, opened 2022-03-08
Fix a problem with a random rotation strategy order

### LINKED ISSUE #12246: Create Index Set page randomly selects rotation strategy
Every time I open the _Create Index Set_ page, the _Index Rotation Configuration_ gets randomly selected from provided 3 options. 
 
## Expected Behavior
One default strategy should be pre-selected.

## Current Behavior
Every time random strategy is selected

1. Open the System -> Indices page
2. Click on the _Create Index Set_ button
3. Verify which rotation strategy is selected
4. Refresh the page and verify again what strategy is pre-selected now

## Your Environment

* Graylog Version: Graylog 4.3.0-SNAPSHOT+4d97241

luk-kaminski:
## Description
Fixes #12246
Rotation strategies are returned in the same order in /system/indices/rotation/strategies endpoint.
TimeBasedRotationStrategy is the first/default one if nothing is configured in graylog.conf file.
Validator for "enabled_index_rotation_strategies" will not allow to put duplicates in this property.

## Motivation and Context
Every time the Create Index Set page was opened, the Index Rotation Configuration gets randomly selected from provided 3 options. It was caused by the fact that back-end returned the strategies in random order.

## How Has This Been Tested?
   1. Open the System -> Indices page
   2. Click on the Create Index Set button
   3. Verify which rotation strategy is selected
   4. Refresh the page and verify again what strategy is pre-selected now

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RotationStrategyResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RotationStrategyResource.java
@@ -42,6 +42,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -79,7 +80,7 @@ public class RotationStrategyResource extends RestResource {
                 .stream()
                 .filter(this::isEnabledRotationStrategy)
                 .map(this::getRotationStrategyDescription)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(LinkedHashSet::new));
 
         return RotationStrategies.create(strategies.size(), strategies);
     }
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RotationStrategyResource.java
```diff
@@ -79,7 +80,7 @@ public RotationStrategies list() {
                 .stream()
                 .filter(this::isEnabledRotationStrategy)
                 .map(this::getRotationStrategyDescription)
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
todvora, line 83: Nice fix! Could we change the data type to List everywhere, so we signalize that the collection is ordered?

todvora (changes requested):

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RotationStrategyResource.java
luk-kaminski: Yes, I can. I won't propagate this change outside of Resource/AutoValue classes, as I believe that in other places (like ElasticsearchConfiguration), it may be more important to signalize that we do not have duplicates in the collection. Does it sound ok?

todvora (approved):

todvora (approved):
Thanks, looks perfect!

PR #12250, opened 2022-03-09
Add Custom Menu List with using react-window when more then 1000 options

maxiadlovskii:
#11930

## Description
There is a check if the number of the items is more than 1000 would be used Custom Menu List based on react-window List component. 

## Motivation and Context
In some cases, when in the Select component we use a big amount of the options we have performance problems. The opening and scrolling are very slow. That is happening because we have a big amount of the items which are rendered. 

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/CustomMenuList.tsx
```diff
@@ -0,0 +1,57 @@
+/*
```
linuspahl, line 1: This component is so coupled to the `Select` component, let's create a `Select` directory and store this file and the `Select` component in it. The directory can contain an `index.ts` which exports the `Select` component. This way it will still be possible to import it from `component/common/Select`, while we still have a `Select.tsx` file.

graylog2-web-interface/src/components/common/CustomMenuList.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const rowRenderer = ({ key, index, style }) => {
+    return <div data-testid="react-virtualized-list-item" key={key} style={style}>{rows[index]}</div>;
+  };
+
+  return (
+    <List style={{ width: '100%' }}
+          width={300}
+          height={300}
+          rowHeight={30}
```
linuspahl, line 47: It looks like the items have a greater height.

linuspahl (commented):
Works great in general. A few thoughts:
- We can update `react-select` https://github.com/Graylog2/graylog2-server/pull/12548 can you please check if it will help us with the performance. It will probably not fix the the performance issues we are trying to fix, but it is worth a try. 
- We already use `react-window`, can you use it here as well, instead of `react-virtualized`? The implementation will may look pretty similar.
- We can make the width and height of the select options list more flexible, but let's see how this will look like with `react-window`

graylog2-web-interface/src/components/common/Select/CustomMenuList.tsx
```diff
@@ -0,0 +1,116 @@
...
+                           innerProps={{
+                             ...innerProps,
+                             'data-testid': 'react-select-list',
+                           }}>
+        {children}
+      </Components.MenuList>
+    );
+  }
+
+  return <WindowList rows={rows}>{children}</WindowList>;
```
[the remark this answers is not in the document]
linuspahl: @maxiadlovskii It looks like you can still remove `React.Children.toArray(children);` and the variable `rows`.

PR #12252, opened 2022-03-09
Apply test specific ESLint rules also for integration tests.

linuspahl:
## Motivation and Context

We have some ESLint rule overrides for frontend (unit) tests in our `eslint-config-graylog` package. With this PR we are also applying these specific rules for integration tests.

### DIFF
--- a/graylog2-web-interface/packages/eslint-config-graylog/index.js
+++ b/graylog2-web-interface/packages/eslint-config-graylog/index.js
@@ -42,7 +42,10 @@ module.exports = {
       files: ['*.js', '*.jsx'],
     },
     {
-      files: ['*.test.js', '*.test.jsx', '*.test.ts', '*.test.tsx'],
+      files: [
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/packages/eslint-config-graylog/index.js
```diff
@@ -42,7 +42,10 @@ module.exports = {
       files: ['*.js', '*.jsx'],
     },
     {
-      files: ['*.test.js', '*.test.jsx', '*.test.ts', '*.test.tsx'],
+      files: [
+        '*.test.js', '*.test.jsx', '*.test.ts', '*.test.tsx',
+        '*.it.js', '*.it.jsx', '*.it.ts', '*.it.tsx',
```
linuspahl, line 47: Probably only the `*.it.tsx` filename extension is relevant for integration tests, but I added the others for the sake of completeness.

PR #12255, opened 2022-03-10
Fixing bug which resulted in multiple search creations and executions.

linuspahl:
## Motivation and Context

This PR is fixing a bug which caused multiple search creations and executions when opening the search page and when executing a search. This affected the following requests:
- `/views/search/metadata`
- `http://localhost:8080/api/views/search`
- `/api/views/search/[search-id]/execute`

Before this change the deep equal check inside `BindSearchParamsFromQuery` always returned false, because the filter of the existing query was `null` and filter of the possible new query was `undefined`. Of course, only when no query filter has been defined.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.test.ts
+++ b/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.test.ts
@@ -63,6 +63,12 @@ describe('BindSearchParamsFromQuery should', () => {
     expect(QueriesActions.update).not.toHaveBeenCalled();
   });
 
+  it('not update query when query is already up to date', async () => {
+    await bindSearchParamsFromQuery(defaultInput);
+
+    expect(QueriesActions.update).not.toHaveBeenCalled();
+  });
+
   it('update query string with provided query param', async () => {
     const input = {
       ...defaultInput,
--- a/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.test.ts
+++ b/graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.test.ts
@@ -34,10 +34,6 @@ jest.mock('views/stores/QueriesStore', () => ({
 describe('BindSearchParamsFromQuery should', () => {
   const query = Query.builder()
     .id(MOCK_VIEW_QUERY_ID)
-    .filter(Immutable.Map({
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/hooks/BindSearchParamsFromQuery.test.ts
```diff
@@ -63,6 +63,12 @@ describe('BindSearchParamsFromQuery should', () => {
     expect(QueriesActions.update).not.toHaveBeenCalled();
   });
 
+  it('not update query when query is already up to date', async () => {
+    await bindSearchParamsFromQuery(defaultInput);
+
+    expect(QueriesActions.update).not.toHaveBeenCalled();
+  });
+
```
linuspahl, line 71: This test is not testing the described bug, but it still provides a benefit. I would add a filter specific test here, if the `toBuilder` method of the `Query` would still contain filter specific logic.

graylog2-web-interface/src/views/logic/queries/Query.ts
```diff
@@ -149,14 +149,14 @@ export default class Query {
...
-    const builder = Query.builder()
+    return Query.builder()
       .id(id)
       .query(query)
       .timerange(timerange)
-      .searchTypes(searchTypes);
-
-    return filter ? builder.filter(filter) : builder;
+      .searchTypes(searchTypes)
+      .filter(filter);
```
linuspahl, line 159: I am not completely sure why we implemented this filter specific logic previously.

dennisoelkers: The idea was to set the `filter` attribute only if it was specified, leaving the default otherwise.

linuspahl: Do you think we should keep the check?
Another solution could be implementing `return filter !== undefined ? builder.filter(filter) : builder;`. But as far as I can tell the result will be the same.

linuspahl: I updated the implementation as discussed.

dennisoelkers (approved):

dennisoelkers (approved):

dennisoelkers (approved):
Thanks a lot! :pray:

PR #12275, opened 2022-03-14
Fix validation of enterprise params without license

### LINKED ISSUE #12271: Query validation doesn't catch missing license exception
Following screenshot is a result of using a query parameter when no enterprise license is available:

image

## Expected Behavior
Query validation should catch such exception and display a validation error as for any other validation problem

## Current Behavior
The exception is propagated outside the validation and triggers an error banner.

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

todvora:
Adapted query validation logic to catch https://github.com/Graylog2/graylog2-server/issues/12271

## Motivation and Context
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3303
Fixes #12271

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/QueryValidationServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/QueryValidationServiceImpl.java
@@ -21,7 +21,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.lucene.queryparser.classic.ParseException;
 import org.graylog.plugins.views.search.ParameterProvider;
 import org.graylog.plugins.views.search.Query;
+import org.graylog.plugins.views.search.elasticsearch.QueryParam;
 import org.graylog.plugins.views.search.elasticsearch.QueryStringDecorators;
+import org.graylog.plugins.views.search.errors.MissingEnterpriseLicenseException;
 import org.graylog.plugins.views.search.errors.SearchException;
 import org.graylog.plugins.views.search.errors.UnboundParameterError;
 import org.graylog.plugins.views.search.rest.MappedFieldTypeDTO;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/QueryValidationServiceImpl.java
```diff
@@ -66,6 +69,14 @@ public ValidationResponse validate(ValidationRequest req) {
             decoratedQuery(req);
         } catch (SearchException searchException) {
             return ValidationResponse.error(toExplanation(searchException));
+        } catch (MissingEnterpriseLicenseException licenseException) {
```
luk-kaminski, line 72: I've seen that you are testing this change with integration test in #3303 PR of [graylog-plugin-enterprise] project.
Don't you believe that it would be nicer to try to write some unit tests for this class, inside this project?

luk-kaminski: Yes, it will require a lot of mocking, especially if not only your change was to be tested, but the whole class, which has no unit tests at all. It is a slight indication that this class probably does too much (does not have one responsibility) and should be split. It is probably not something that needs to be part of this bug fix, but if you believe it is worth an effort, a new task may be created to clean this code a little bit.

todvora: From my POV, the whole QueryValidationService is just a converter and transforms data from the query parser and query decorators. It doesn't contain much logic on its own. From that reason, I rather unit-tested the query parser itself and integration-tested the whole service, together with the QueryValidationResource. I'll create a task and assign it to you, this will spread the validation knowledge between both of us. Thanks!

luk-kaminski:
Good job!

luk-kaminski (approved):

Issue #12262, opened 2022-03-11
Error when deleting saved search without it being loaded


edmundoa:
Deleting a saved search without loading it first, results in an error page. I could also see a flash of the error when the saved search is loaded but we trigger a redirect which hides the error page.

https://user-images.githubusercontent.com/716185/157871142-7a645623-27bd-4533-bccd-78c7e0bcfafd.mp4

This is the stack trace shown on the page:

```
Cannot read properties of undefined (reading 'filter')


Stack Trace:

TypeError: Cannot read properties of undefined (reading 'filter')
    at cg (http://localhost:9000/assets/app.239563ce538838b8a96e.js:2:1297075)
    at Ji (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:792935)
    at Ba (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:801784)
    at Ls (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:840080)
    at Mc (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:831636)
    at mc (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:831561)
    at sc (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:828591)
    at http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:780339
    at exports.unstable_runWithPriority (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:855639)
    at Ko (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:780048)
    at Qo (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:780284)
    at Go (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:780219)
    at nc (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:825376)
    at ya (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:797279)
    at f.<anonymous> (http://localhost:9000/assets/app.239563ce538838b8a96e.js:2:678447)
    at o.n (http://localhost:9000/assets/vendor.3b152b3d9c1d8273aa7e.js:2:949206)
[...]

PR #12276, opened 2022-03-14
Fallback to current query values

Gaya:
As a solution for the issue mentioned below which happens after deleting a "saved search"

## Description

Fixes the errors that arise when deleting a saved search described in here:

https://github.com/Graylog2/graylog2-server/issues/12262

I am not sure if this is the best way to do this, but it makes sure the component doesn't crash when a currentQuery doesn't exist.

I found that looking at the code that there are actually situtions where the `currentQuery` doesn't exist, and the `DefaultFieldTypesProvider` is only used when `currentQuery` was loaded.

The thing is when currentQuery is set to `undefined` again this will cause the app to break.

Adding default values prevents this crash, and after delete of the saved search the page will reload anyway.

I am not sure if this is a race condition though, so that why I am not sure if this is the right way.

## Motivation and Context

Fixes https://github.com/Graylog2/graylog2-server/issues/12262

## How Has This Been Tested?

Manually by repeating the sets in the issue

## Types of changes

## Checklist:


edmundoa (commented):
Besides an issue with the default time range, I think the changes look good and at least fixed the original issue for me. However, I am not confident enough with what implications that would have, so I think the @Graylog2/search team should take a look and decide if this is the way of solving the issue.

dennisoelkers (approved):

Issue #12257, opened 2022-03-11
Selecting a range from histogram uses wrong TZ when user has no explicit TZ configured


todvora:
![Peek 2022-03-11 09-40](https://user-images.githubusercontent.com/4102775/157837227-01958a16-1636-437b-9c32-e05838c244c2.gif)

![image](https://user-images.githubusercontent.com/4102775/157837249-07ff02ee-0783-4ac2-bee8-186faa86ec9a.png)

Notice the different TZ used for the histogram and the range selector afterwards. 
My user doesn't have any explicit TZ configured

## Expected Behavior
Both the histogram and the range selector should use the same TZ, in this case derived from the browser.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Graylog Version: Graylog 4.3.0-beta.1+a752232

PR #12277, opened 2022-03-14
Always create new time range based on user time zone when selecting a time range in the histogram.

linuspahl:
## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/12257 there is currently a bug when selecting a time range in the histogram, when the current user has no time zone.

We display all times based on the user time zone, when there is no time zone configured we are using the browser time zone. The logic which calculates the new time range used a different fallback for the user time zone.

Fixes: https://github.com/Graylog2/graylog2-server/issues/12257

## Types of changes


maxiadlovskii (approved):

dennisoelkers (approved):
Looks good, works well!

PR #12285, opened 2022-03-15
Expose ES index UUID

patrickmann:
ES assigns a UUID to each index. This information is useful e.g. to resolve situations where an index name is re-used. 

We expose the UUID via `Indices.getIndexId()`


janheise (approved):
LGTM! Thank you.

Issue #12261, opened 2022-03-11
Undeclared parameter error persists


patrickmann:
Error message gets out of sync with the search page.

## Expected Behavior
After deleting the query referring to missing parameters, the corresponding error should disappear.

## Current Behavior
The error persists, even though the condition no longer exist. 
It disappears, if I navigate to another tab and back again to search.

## Steps to Reproduce (for bugs)
1. Define a parameter and use it in the search field
2. Delete the parameter - this causes an _Undeclared parameter_ error to appear
3. Delete the search query - observe that the error persists
![image](https://user-images.githubusercontent.com/12073930/157860374-4e33826e-7613-4a11-98d5-4c947452fb7c.png)

## Your Environment

* Graylog Version: 4.3

PR #12286, opened 2022-03-15
Remove condition which block search button when are undeclared params

maxiadlovskii:
Fixes: #12261 

## Description
In the current implementation, we block submit button when we have at least one undeclared parameter in `SearchMetadataStore`. 
```
const _disableSearch = (undeclaredParameters, parameterBindings, usedParameters) => {
  const bindingsMap = getParameterBindingsAsMap(parameterBindings);
  const missingValues = usedParameters.filter((param) => (param.needsBinding && !param.optional)).map((p) => bindingsMap.get(p.name)).filter((value) => !trim(value));

  return undeclaredParameters.size > 0 || missingValues.size > 0;
};
....
  {
    searchMetadata: SearchMetadataStore,
    executionState: SearchExecutionStateStore,
    configurations: SearchConfigStore,
  },
  ({ searchMetadata, executionState, configurations }) => ({
    isDisabled: _disableSearch(searchMetadata.undeclared, executionState.parameterBindings, searchMetadata.used),
    config: configurations.searchesClusterConfig,
  }),
....
```

When the user removes the parameter from the search input we don't update the `SearchMetadataStore`. That is why the condition `undeclaredParameters.size > 0` would be always `true` and the search button will be disabled. In other words, since we got one undeclared parameter in the metadata - the search button will be always blocked.
I decided to remove condition `undeclaredParameters.size > 0` from the  code. In this case, the user will be able to click the search button, after removing the undeclared param from the input and update the search. That wouldn't be a problem for our logic, because form validation handles the case of using undeclared param in the search input. 

## Types of changes

## Checklist:

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3321

### DIFF
--- a/graylog2-web-interface/src/views/components/WithSearchStatus.tsx
+++ b/graylog2-web-interface/src/views/components/WithSearchStatus.tsx
@@ -28,7 +28,7 @@ const _disableSearch = (undeclaredParameters, parameterBindings, usedParameters)
   const bindingsMap = getParameterBindingsAsMap(parameterBindings);
   const missingValues = usedParameters.filter((param) => (param.needsBinding && !param.optional)).map((p) => bindingsMap.get(p.name)).filter((value) => !trim(value));
 
-  return undeclaredParameters.size > 0 || missingValues.size > 0;
+  return missingValues.size > 0;
 };
 
 type SearchStatusProps = {


linuspahl (approved):
Gave it a try and in my opinion it makes sense. The query validation is now making this obsolete.

dennisoelkers (approved):

PR #12290, opened 2022-03-16
Notification is presented if any active writing index is blocked.

### LINKED ISSUE #10930: Overview page should not report ES Cluster as "Green" when in RO mode after hitting disk high watermark
When running a single node Graylog instance with Elasticsearch running on the same host, if the high watermark for disk usage is triggered, Elasticsearch will set the ```index.blocks.read_only_allow_delete: true``` setting, but when more disk space becomes available, the setting is  not reverted to ```null```, yet Graylog will show the Elasticsearch status as Green, even though messages cannot be written to Elasticsearch.

## Expected Behavior

The System / Overview page should alert an admin to the fact that indicies are in RO mode. Ideally, there would also be a Notification alert raised.

## Current Behavior

Currently, this issue is found by navigating to the Indexer Failure page. This does give a clear message inidcating the problem, and I understand that technically, according to Elasticsearch's design, the cluster is "Green". But this is misleading, as the cluster is essentially non-functional for Graylog's purposes.

## Possible Solution

Maybe some kind of UI element that would be revealed after the high watermark condition is resolved (like a button saying "Enable RW) that would issue the following (or something like it using config vars)

```
curl -XPUT -H "Content-Type: application/json" http://localhost:9200/_cluster/settings -d '{ "transient": { "cluster.routing.allocation.disk.threshold_enabled": false } }'
curl -XPUT -H "Content-Type: application/json" http://localhost:9200/_all/_settings -d '{"index.blocks.read_only_allow_delete": null}'
```
## Context
This issue has caused us to lose hours of data in our log indicies.

## Your Environment

* Graylog Version: 4.0.7
* Java Version: openJDK 1.8.0
* Elasticsearch Version: 7.13
* MongoDB Version: 1:3.6.3
* Operating System: Ubuntu 18.04
* Browser version: Microsoft Edge Version 91.0.864.54

luk-kaminski:
## Description
fixes #10930
Notification is presented if any active writing index is blocked.
Notification contains name of blocked index (i.e. graylog_0) and name of block (i.e. index.blocks.read), so that the admin can easily fix the problem with proper REST call.

## Motivation and Context
In #10930 it was suggested that notification should be presented if index.blocks.read_only_allow_delete is set to true.
This PR generalizes a problem : notification is presented if any active writing index is blocked, with any of index.blocks.* type of blocks. Because of that, admin can react quickly when the block is present and prevent losing important data, which otherwise could not be ingested, without any clear indication that there is a problem.

As presented on the screenshot, notification contains all the necessary information needed to fix the problem with proper REST call.

The solution does not contain any tool for automatic or semi-automatic removal of the blocks.
It may be a good idea to introduce it and improve the comfort of block removal even more, but it definitely requires more discussions about how to do it.

## How Has This Been Tested?
1. Block any active write index with the following (or similar) REST call:
```
http://localhost:9200/unused_0/_settings
{
  "index.blocks.read_only": true
}
```
2. Check if notification is present.
3. Remove all the blocks with  the following (or similar) REST call:
```
http://localhost:9200/unused_0/_settings
{
  "index.blocks.read_only": null
}
```
4. Check if notification is gone.

Screenshot 2022-03-16 at 13-01-31 Graylog - System overview
## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/blocks/JestBlockSettingsParser.java
+++ b/graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/blocks/JestBlockSettingsParser.java
@@ -1,0 +1,43 @@
+package org.graylog.storage.elasticsearch6.blocks;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.searchbox.client.JestResult;
+import org.graylog2.indexer.indices.blocks.IndicesBlockStatus;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+
+public class JestBlockSettingsParser {
+
+    static final Collection<String> POSSIBLE_BLOCKS = Arrays.asList("read_only", "read_only_allow_delete", "read", "write", "metadata");
+    static final String BLOCK_SETTINGS_PREFIX = "index.blocks.";
+
+    public IndicesBlockStatus parseBlockSettings(final JestResult jestResult, final List<String> indices) {
+        final IndicesBlockStatus indicesBlockStatus = new IndicesBlockStatus();
+        if (indices == null || indices.isEmpty() || jestResult == null) {
+            return indicesBlockStatus;
+        }
+        final JsonNode jsonObject = jestResult.getJsonObject();
+        if (jsonObject != null && !jsonObject.isMissingNode()) {
+            for (String indexName : indices) {
+                Collection<String> blocks = new LinkedList<>();
+                for (String possibleBlock : POSSIBLE_BLOCKS) {
+                    final JsonNode blockElem = jsonObject.path(indexName).path("settings").path("index").path("blocks").path(possibleBlock);
+                    if (!blockElem.isMissingNode()) {
+                        if ((blockElem.isBoolean() && blockElem.asBoolean()) ||
+                                (blockElem.isTextual() && blockElem.asText().contains("true"))
+                        ) {
+                            blocks.add(BLOCK_SETTINGS_PREFIX + possibleBlock);
+                        }
+                    }
+                }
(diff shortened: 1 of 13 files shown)


graylog-storage-elasticsearch6/src/main/java/org/graylog/storage/elasticsearch6/blocks/JestBlockSettingsParser.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import io.searchbox.client.JestResult;
+import org.graylog2.indexer.indices.blocks.IndicesBlockStatus;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+
+public class JestBlockSettingsParser {
```
todvora, line 28: Same as the parser for ES6

todvora (approved):
Tested locally, works perfectly, thanks!

dennisoelkers (commented):
I have not looked at the PR in details, but one thing I noticed: Can you add an integration test for it? I think it would be very beneficial to make sure that parsing the result works without mocks and continues to work across versions. Looking at the tests for the flood stage handling _might_ help.

luk-kaminski:
@dennisoelkers -> I've added integration test with commit [7dec793]. Please let me know if it seems to be sufficient or not.

Sirclicksalot:
Nothing really to see here, just wanted to comment because this is the first bug/issue/improvement I've ever submitted to a public GitHub repo that has resulted in a change in a product.

😃

dennisoelkers:
Hey @Sirclicksalot, thanks for your suggestion! :)

linuspahl (approved):
Frontend changes look goos to me and the feature is working as expected!

Issue #12265, opened 2022-03-11
Roles select on create user page is not being cleared after assigning roles.


linuspahl:
## Expected Behavior

Roles select on create user page (`/system/users/new`) should be cleared after assigning roles.

## Current Behavior

The selected roles are still part of the select input after assigning the roles.
![image](https://user-images.githubusercontent.com/46300478/157897683-3f824271-7359-4f5c-9ada-b9766fc702a3.png)
The same problem exists for the collaborator select in the "share entity" modal. Maybe the bug is related to the `Select` component.

## Steps to Reproduce (for bugs)

1. Go to `/system/users/new`
2. select roles
3. assign these roles

## Your Environment

* Graylog Version: 4.3 beta

Gaya:
Fixed in: https://github.com/Graylog2/graylog2-server/pull/12293

Gaya:
Fixed by: https://github.com/Graylog2/graylog2-server/pull/12297

Issue #12259, opened 2022-03-11
Role:assign user interface out of sync


todvora:
The user selector is out of sync with the underlying data structures sometimes:

![user_roles](https://user-images.githubusercontent.com/4102775/157854740-4e2d66b6-36a5-4609-bc83-566a69a5a83c.gif)

## Expected Behavior
After assigning users to the role, they should either disappear from the selector or should stay there and be available for repeated assigning.

## Current Behavior
The users stay in the selector but are not really recognized when I try to assign them again.

## Steps to Reproduce (for bugs)
1. Select an user from the dropdown
2. Assign the user to the role
3. Remove the user from the role
4. Try to assign the aleady selected user again

## Your Environment

* Graylog Version: Graylog 4.3.0-beta.1+a752232 on server
* Browser version:

Gaya:
Fixed in: https://github.com/Graylog2/graylog2-server/pull/12293

Gaya:
Fixed by: https://github.com/Graylog2/graylog2-server/pull/12297

PR #12293, opened 2022-03-16
fix: rerender multi select when value changes

Gaya:
## Description

When clearing values from the `<Select />` component the inner state would not update to the newly provided value(s) when they were passed as a prop in a "controlled" manner.

## Motivation and Context

It makes sure the `<Select />` rerenders on value change. It might not be the most ideal way of solving it, but while digging deeper into the component and `react-select` itself it was hard to spot where it went wrong.

I believe it had something to do with `react-select`'s `StateManager`, since it didn't update its value to the passed prop.

The solution is to add a `key` prop based on the value so `react-select` forces a rerender.

Fixes: https://github.com/Graylog2/graylog2-server/issues/12265
Fixes: https://github.com/Graylog2/graylog2-server/issues/12259

## How Has This Been Tested?

Manually in the browser

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select.tsx
+++ b/graylog2-web-interface/src/components/common/Select.tsx
@@ -566,7 +566,7 @@ class Select<OptionValue> extends React.Component<Props<OptionValue>, State> {
       return <CreatableSelect {...selectProps} />;
     }
 
-    return <ReactSelect {...selectProps} />;
+    return <ReactSelect key={value} {...selectProps} />;
   }
 }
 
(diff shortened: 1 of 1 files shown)


dennisoelkers (commented):
This fixes the issue of the state of the select not being updated. The downside of this is though, that upon every rerender, the select is blurred. This prevents users from being able to enter multiple values in one go, because the input is not focussed anymore after selecting an option.

linuspahl (changes requested):
In my opinion using the `key` attribute to solve this problem is a little bit extreme. This way the `react-select` component and its child components will remount.

It also has some other (related) disadvantages, like the select loosing the focus when changing its value:
select

It looks like another solution could be to ensure the `value` we provide for the `react-select` is never `undefined`: `value: formattedValue ?? '',`

Gaya:
Good points! Going to look into again.

Gaya:
@dennisoelkers @linuspahl forgot to mention I comitted a new fix

Gaya:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/12297

PR #12295, opened 2022-03-16
Disable HTTP Notification submit when url field have error

### LINKED ISSUE #12272: Http notification can be saved with wrong URL
Http notification can be saved with wrong URL.

## Expected Behavior
Saving HTTP Notification should not be possible when wrong URL is used. 

## Current Behavior
Http notification can be saved with wrong URL.
If you put invalid URL in the URL field, you will see an error message, but you still can save the notification.
Screenshot 2022-03-14 at 09-08-46 Graylog - New Notification
If you put non-whitelisted URL in the URL field, you will see an error message, but you still can save the notification. 

## Possible Solution
"Create" button is disabled or clicking this button ends up with an error if wrong URL is provided.

1. Go to Alerts - > Notifications.
2. Click "Create Notification"
3. Choose "HTTP Notification" from Notification Type selector.
4. Enter invalid URL in the URL input that appears.
5. Notice red warning under URL field.
6. Try to save and see that it is possible, despite wrong URL.

## Your Environment

* Graylog Version: Graylog 4.3.0-beta.1+a752232 (Test Day)

ousmaneo:
Fix #12272 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/URLWhiteListInput.tsx
+++ b/graylog2-web-interface/src/components/common/URLWhiteListInput.tsx
@@ -31,7 +31,7 @@ type Props = {
   validationMessage: string,
   validationState: string,
   url: string,
-  setIsSubmitEnabled: (enabled: boolean) => void,
+  setIsSubmitEnabled?: (enabled: boolean) => void,
   labelClassName: string,
   wrapperClassName: string,
   urlType: React.ComponentProps<typeof URLWhiteListFormModal>['urlType'],
@@ -73,7 +73,7 @@ const URLWhiteListInput = ({ label, onChange, validationMessage, validationState
         setIsWhitelisted(result.is_whitelisted);
       });
     }
-  }, [url, validationMessage, validationState]);
+  }, [url, validationMessage, validationState, setIsSubmitEnabled]);
 
   const onUpdate = () => {
     triggerInput(urlInputRef.current.getInputDOMNode());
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/common/URLWhiteListInput.tsx
```diff
@@ -31,12 +31,13 @@ type Props = {
...
   validationState: string,
   url: string,
+  setIsSubmitEnabled?: (enabled: boolean) => void,
   labelClassName: string,
   wrapperClassName: string,
   urlType: React.ComponentProps<typeof URLWhiteListFormModal>['urlType'],
 };
 
-const URLWhiteListInput = ({ label, onChange, validationMessage, validationState, url, labelClassName, wrapperClassName, urlType }: Props) => {
+const URLWhiteListInput = ({ label, onChange, validationMessage, validationState, url, setIsSubmitEnabled, labelClassName, wrapperClassName, urlType }: Props) => {
```
edmundoa, line 40: While this approach works, I think it is breaking some of the semantics of the form: A form input should not decide if the submit button is enabled or disabled. This is also problematic when we expose this as our API, which is in the end what we do when we change the interface of pluggable entities.

I think the input could call a function once the validation state changes (maybe adding an `onValidationChange` prop that consumers can provide, or adding the validation state to the `onChange` callback). From my perspective, that would make it easier to understand, since it's a similar API one uses with another form, and decoupled from a particular implementation. What do you think?

ousmaneo: Hi, thank you for your review. I agree about the semantic issue. I will change that to use `onValidationChange.`

Gaya (approved):
Looks great! Good suggestion @edmundoa

PR #12298, opened 2022-03-17
Async forcemerge request that still blocks as all other index operations

### LINKED ISSUE #12025: Optimizing indices after index rotation blocks master node's ingestion
Optimizing indices after index rotation blocks the master node's ingestion until completed. This is not a scalable behavior.

If the Graylog deployment has a large amount of ingestion, and rotation on P1D, the Graylog master node can stop processing messages for hours. The Graylog master does not communicate this to the load balancer, and as a result its buffers fill up.

## Expected Behavior
Graylog Master node can continue to process message during Elasticsarch index optimisation.

## Current Behavior
Graylog Master node cannot process messages during Elasticsarch index optimisation.

## Possible Solution
Allow the master node to continue ingesting during index rotation. Shift force merge requests to a different thread pool which is not blocking ingestion.

* Graylog Version:
4.2.x
* Elasticsearch Version:
7.10

See customer ticket HS-666683796 for example.

todvora:
This is an attempt to fix issue described in https://github.com/Graylog2/graylog2-server/issues/12025. It assumes, that the forcemerge operations consume all threads from the ES client thread pool, blocking all other ES operations. To prevent that, the forcemerge behind index optimization is triggered as an async call. To keep the app logic and structure, this call is then blocking the caller thread and waits till the forcemerge finishes (SystemJob implementations block generally and their running/blocking time is exposed as a metric). Generally, for each optimize job, it frees one thread from the ES client thread pool yet still blocks one in the app. 

Additionally, under the original issue, I have mentioned several config settings that I believe may add to the problem if kept in the default settings. 

## Motivation and Context
Hopefully it fixes #12025

## How Has This Been Tested?
Manually tested that the index rotation and optimization works. IndicesIT to ensure that nothing else breaks. Performance-wise I was unable to test the benefit and verify our assumptions so far. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java
@@ -27,6 +27,7 @@ import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.flush.FlushRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.forcemerge.ForceMergeRequest;
+import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.forcemerge.ForceMergeResponse;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.open.OpenIndexRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.settings.get.GetSettingsRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.settings.get.GetSettingsResponse;
@@ -86,6 +87,10 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java
```diff
@@ -417,8 +422,13 @@ public void optimizeIndex(String index, int maxNumSegments, Duration timeout) {
                 .indices(index)
                 .maxNumSegments(maxNumSegments)
                 .flush(true);
-
-        client.execute((c, requestOptions) -> c.indices().forcemerge(request, withTimeout(requestOptions, timeout)));
+        try {
+            final FutureActionListener<ForceMergeResponse> future = new FutureActionListener<>();
+            client.execute((c, requestOptions) -> c.indices().forcemergeAsync(request, withTimeout(requestOptions, timeout), future));
+            future.getBlocking(timeout);
```
luk-kaminski, line 428: I'm sorry, I may have not understood everything correctly, but what is the point of introducing an async call if in the next line you block the execution with getBlocking() ? I'm afraid it won't change anything, the thread executing optimizeIndex() method will have to wait for the completion anyway...

luk-kaminski, line 428: I'd say this code block should be given a comment.
One may wonder what is the point of introducing an async call if in the next line you block the execution with getBlocking()? I believe you should state clearly that it is a trick and you hope to save threads from connection pool with this trick.

luk-kaminski, line 428: Maybe we don't have to wait for the completion at all?

todvora:
@bernd this was one of ideas we have discussed with @dennisoelkers. We could have three separate clients, one for ingestion, one for search and one for maintenance. Each type of operation would, in the worst case, only block similar operations. Maintenance (like the index optimization - forcemerge) wouldn't block neither the ingestion nor the search. 

I think technically it's easy to implement, the biggest issue will be how to configure and document three different clients. @dennisoelkers suggested to divide current values (like `elasticsearch_max_total_connections_per_route`) between all three proportionally. So we wouldn't have to expose the configuration. But it could easily cause many unexpected issues elsewhere. 

This PR tries to fix only one very specific problem without touching anything else. I fully agree that it's not a proper system-wide solution and the client separation would be better in the long term. 

Maybe we should abandon this fix, open a new feature request for client separation and fix the issue later, with a proper implementation. Should we take that route?

dennisoelkers:
@todvora: I see two benefits in following up with this fix you created:

- If we can validate that it's actually helping with the symptoms we are seeing, we have a validation for our initial premise of the root cause. The other way around it could falsify it, making the much bigger change potentially unnecessary, saving us lots of work.
- Introducing client classes would be something for 4.4ff and would not help anyone on 4.3 and earlier. This much smaller fix could be backported to current stable versions though, helping users who cannot wait until 4.4.

janheise:
maybe `wait_for_completion=false` is also worth a try? 
(see https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html)

todvora:
I tried to benchmark the sync vs async forcemerge calls and don't see any difference in graylog performance locally. But my setup is one-node-only, which is during the forcemerge heavily utilized and will be the bottleneck in such situation. So I can't really differentiate between blocked client thread and overloaded ES node => both will increase the count of messages waiting for processing in journal.

todvora:
Closing, we agreed on a different solution implemented in https://github.com/Graylog2/graylog2-server/pull/13521

PR #12299, opened 2022-03-17
Provide config option to set upper limit for index retention period

thll:
Adds a new config option `max_index_retention_period`.

Limits the effective retention period that is configurable by a user. The limit will only be enforced if a time-based rotation strategy is chosen.

When a user attempts to safe an index set configuration and has a time-based rotation strategy configured, the effective retention period (rotation period * max indexes to keep) will be calculated and compared with the value of the new setting. If the effective retention period would exceed the configured limit, the index set config will be rejected.

This change also enables validation for index set config updates. Prior to this change, only new index set configurations were validated.
 
This implements the backend for https://github.com/Graylog2/graylog-plugin-cloud/issues/1006.


thll:
I've added a `context` element to the REST response for listing the retention strategies:
image

The new setting is exposed via the `max_index_retention_period` key. If not configured, its value will be `null`. If configured the response will look like this:

```
{
  "total": 4,
  "strategies": [...],
  "context": {
    "max_index_retention_period": "P1D"
  }
}
```

@ousmaneo does that work for you?

ousmaneo:
Thank you @thll, this works for me.

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
```diff
@@ -17,26 +17,65 @@
...
+    @Nullable
+    private Violation validatePrefix(IndexSetConfig newConfig) {
         // Build an example index name with the new prefix and check if this would be managed by an existing index set
         final String indexName = newConfig.indexPrefix() + MongoIndexSet.SEPARATOR + "0";
         if (indexSetRegistry.isManagedIndex(indexName)) {
-            return Optional.of(Violation.create("Index prefix \"" + newConfig.indexPrefix() + "\" would conflict with an existing index set!"));
+            return Violation.create("Index prefix \"" + newConfig.indexPrefix() + "\" would conflict with an existing index set!");
         }
 
         // Check if an existing index set has a more generic index prefix.
```
patrickmann, line 81: This is only half the story. Should be something like:
// Check if the new index set configuration has a more generic index prefix than an existing index set, or vice versa.

graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
```diff
@@ -89,6 +91,24 @@
...
+     * <p>
+     * This setting is used to validate the value a users chooses for {@code max_number_of_indices} when configuring
+     * an index set. However, it is only in effect, when a <em>time-based rotation strategy</em> is chosen, because
+     * otherwise it would be hard to calculate the effective retention period at configuration time.
+     * <p>
+     * If a rotation strategy other than time-based is selected and/or no value is provided for this setting, no upper
+     * limit for index retention will be enforced.
+     */
+    @Parameter(value = MAX_INDEX_RETENTION_PERIOD)
+    private Period maxIndexRetentionPeriod = null;
```
patrickmann, line 105: Add this parameter to graylog.conf (commented out)

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
```diff
@@ -59,14 +60,17 @@
     private final Map<String, Provider<RetentionStrategy>> retentionStrategies;
     private final ObjectMapper objectMapper;
     private final ClusterConfigService clusterConfigService;
```
patrickmann, line 62: Unused variable - we might as well clean it up

patrickmann, line 62: Unused private field - we can remove it

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
thll: Good catch. I didn't write that comment but I'll fix it with your suggestion.

thll:
> Looks good - just some very minor comments. I tested the `system/indices/retention/strategies` API, but not index set creation / update. Easier to do that once the UI is in place.

The UI changes will probably only assist users in choosing the correct values. Functionally it should be almost identical. If users choose invalid values, they will get a validation error. So the feature can be well tested using the UI already.

patrickmann (approved):
Tested index creation / update. Good to go. 
Waiting for 4.4 branch to merge.

PR #12303, opened 2022-03-17
Integration test for getIndexId()

patrickmann:
Adding a simple integration test for getIndexId() to ensure it works against the supported platforms.


janheise (approved):
LGTM!

PR #12305, opened 2022-03-17
Do not hide search results when search refresh hook sets an error, but display overlay.

linuspahl:
## Motivation and Context

Before this change we were hiding the search results, when a search refresh hooks sets an error. This is currently the case for parameter errors, e.g. when a parameter needs to be declared.

This can be confusing for users, but it can also result in disappearing changes, when editing a widget, as described in https://github.com/Graylog2/graylog-plugin-enterprise/issues/3281.

This solution is not fixing all related problems, for example if you remove a parameter on a dashboard, which is being used in a widget query, we display the "Undeclared parameter" warning, but it is still not possible to remove the parameter name from the widget query. In every case it is an improvement.

Before: 
image

After:
image

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/3281

### DIFF
--- a/graylog2-web-interface/src/views/components/SearchResult.test.tsx
+++ b/graylog2-web-interface/src/views/components/SearchResult.test.tsx
@@ -17,7 +17,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { act } from 'react-dom/test-utils';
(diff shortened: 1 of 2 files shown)


dennisoelkers (approved):
Looks great, works fine! :pray:

PR #12308, opened 2022-03-17
Update identifier for query validation parsing error type.

linuspahl:
## Description

This PR is updating the identifier we use in the frontend for the query validation parsing error type.
This is necessary, because we changed the id in the backend recently.

Based on this identifier we display error type specific documentation links.


dennisoelkers (approved):
Looks good, works fine!

PR #12336, opened 2022-03-22
Handle query parsing error for event definition listing

### LINKED ISSUE #12266: It is not possible to filter the event definitions list by id.
## Expected Behavior

The search syntax help of the event definitions filter mentions it should be possible to filter the list by id (`id:the-id`).

image

## Current Behavior

Currently this does not work and all event definitions will be returned, no matter what id a users searches for.

1. go to `/alerts/definitions`
3. search for `id:anything` in the search input of the event definitions overview
4. it will return all event definitions

## Your Environment

* Graylog Version: 4.3

thll:
Catch query parsing error in the backend and display the
error in the frontend.

Previously, a search in the event definitions list for
an invalid id (not a valid mongoDB id) would result
in a 500 error which was also not handled by the frontend.

Fixes #12266


patrickmann (approved):
Fix looks good and works.

The error message is not terribly helpful though. It is always _"invalid hexadecimal representation of an ObjectId"_ - regardless of what value was given for `id`

Issue #12291, opened 2022-03-16
Unit-tests for QueryValidationService


todvora:
The QueryValidationService doesn't have any unit tests, only several integration tests. Maybe we could do some cleanup, limit dependencies and refactor the logic, adding unit tests in the same time.

PR #12338, opened 2022-03-22
Validation code refactored. 

luk-kaminski:
## Description
QueryValidationServiceImpl class split to number of classes, so that each class has one responsibility. Unit tests added.
It is just part of the task, related to graylog2-server only.

## Motivation and Context
#12291 issue shows initial motivation.

Some important points from discussions:
"The only way how I can validate the params is to let the decoration run, ignore the results and catch exceptions it produces. All the information about the params is then propagated back in the exception classes. And since I catch the exceptions in the server2 code, I can't even use the type of the parameters."

"The best solution would be probably to have one shared validation interface, two implementations, one in the server and one in the enterprise code, both providing lists of validation problems. The server would then merge it all together and deliver the validation info."

"If our validation code would be pluggable, we could also provide a parameter validator in enterprise which would take a look at query strings, parameters and bindings which would return validation errors about undeclared parameters, missing bindings, etc."

## Types of changes

## Checklist:


luk-kaminski:
Those changes became incompatible with Tomas's changes. I'm converting this PR to draft, most probably it won't be used as a whole, but rather parts of it will be reused.

luk-kaminski:
Removing this draft PR, incompatible with recent Tomas's changes

PR #12344, opened 2022-03-23
Validate new URL in allow list on first render

### LINKED ISSUE #12267: Instantly run validation when adding URL to whitelist for HTTP Notification
## Expected Behavior
When the title field is empty and the user clicks the `Save` button we should show an error that the title field is required

## Current Behavior
When we add an HTTP Notification, add a new URL to the whitelist and click the `Save` button in the modal directly without adding a title there is no error shown and we can't save changes. 
The bug can be only reproduced if we didn't change the title field. If we would change the title field and clear the input - validation works. 
image

## Possible Solution
Run validation when clicking on the `Save` button

1.  Login into Graylog as an admin
2. Go to `/alerts/notifications/new`
3. Create a new HTTP Notification
4. Type a new URL which not exists in the white list 
5. Click `Add to URL white list `
6. Click `Save` (without changing the form values)
7. You wouldn't see any error message 

## Your Environment

* Graylog Version: 4.3 beta

edmundoa:
Validate new URL on `UrlWhiteListForm` on first render. This component validates entries on the browser, except URLs of regex type, which are validated with an API call each. Since we don't know how many entries users have, and which type they are, I decided to only validate the new URL entry (which must be passed as a prop). I think this gives us a good balance between being helpful when adding an entry, and not firing a lot of network requests to validate each entry in the allow list, which should have already been validated when they were persisted.

I took the opportunity to rewrite and extend our tests for this component, and also decoupled a little bit the validation from the form inputs. Since the logic of this component sometimes have unexpected results, I tried to not go overboard with changes, please let me know if you think we should dial them down so close to the release.

Fixes #12267


graylog2-web-interface/src/components/configurations/UrlWhiteListForm.tsx
```diff
@@ -211,9 +199,29 @@ const UrlWhiteListForm = ({ urls, onUpdate, disabled }: Props) => {
...
+      const valid = firstRender && newEntryId ? await isNewEntryValid() : hasValidationErrors();
+      onUpdate(config, valid);
+    };
+
+    propagateUpdate(!isMounted.current);
+
+    if (!isMounted.current) {
+      isMounted.current = true;
+    }
   }, [config]);
```
edmundoa, line 225: The linter failure here was left on purpose. Changing the dependency array had some negative consequences in the past, as well as increasing the number of calls to `onUpdate()` drastically, so I think changing it right now may do more harm than good before the release.

ousmaneo (approved):
Tested and works!

ousmaneo (changes requested):
Thank you for the changes! I found a small issue on my last test (after approving). When I create a Regex URL and click save. It's still saved as `literal` and on edit too. it seems that the state of the URL type is not properly saved.

ousmaneo (approved):
Tested and works. Thank you for working on this 👍

PR #12348, opened 2022-03-24
Setting render_uri in Dockerfile for integration tests

janheise:
## Description
Hopefully, this attempt solves random timeout problems during the reporting integration tests. Maybe docker is not up correctly (race conditions etc) once in a while so the guessing of the correct render_uri fails and the test fails with a timeout.

## Types of changes

## Checklist:


todvora (commented):
The removal of the timeout is by purpose?

dennisoelkers (changes requested):
I do not think this is a good idea. In most production scenarios `report_render_uri` is not set, defaulting to be based on `http_bind_address`. This is not "guessed" and should be stable. Setting this explicitly for integration tests makes the test scenario differ from most production scenarios, which is not ideal IMO.

janheise:
Ok, I can see that.

PR #12351, opened 2022-03-24
Isolate usage of `react-styleguidist` to separate directory.

dennisoelkers:
## Motivation and Context

Prior to this change, `react-styleguidist` was a dependency of our main web interface module. This lead to outdated dependencies being pulled in transitively repeatedly. With this change, we are moving the `react-styleguidist` dependency to the `docs` subdirectory, isolating it in a separate package.

## Types of changes

## Checklist:


graylog2-web-interface/docs/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@graylog/frontend-documentation",
+  "version": "1.0.0",
+  "description": "Graylog Frontend Documentation",
+  "main": "index.js",
```
linuspahl, line 5: It seems we do not have an `index.js` file, but since this is the default when no `main` entry is defined and there is no good alternative, I do not have a better suggestion.

linuspahl (commented):
Looks good to me, works as as expected. I just saw we should update the documentation regarding the styleguide in `graylog2-webinterface/README.md`.

linuspahl (approved):

PR #12358, opened 2022-03-25
Always true LeaderElectionModule

patrickmann:
## Description
Provide a leader election service that always returns true. 

## Motivation and Context
This is needed when leader election is not required / possible, e.g. the Forwarder.


mpfz0r (approved):
LGTM

PR #12287, opened 2022-03-15
Add support for checking resolved view permissions

danotorrey:
## Description

The `SearchDomain hasReadPermissionFor()` method is responsible for verifying that the current user is authorized to view the search record for the provided `ViewDTO`. Support for external view resolvers was added in https://github.com/Graylog2/graylog2-server/pull/12089. However, when implemented, support for authorizing search access for resolved views was not added, so resolved views were not accessible for non-admin users.

This PR adjusts the `SearchDomain hasReadPermissionFor()` logic to also query search records for resolved views, which subsequently allows permissions checks to be completed for them. 

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3315

## Motivation and Context

This is needed to fix a bug for Graylog Security. See https://github.com/Graylog2/graylog-plugin-enterprise/pull/3315.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #12367, opened 2022-03-28
Proper permissions checking for resolved views

danotorrey:
## Description

Add proper view resolver permissions checks that support view resolver-specific permissions checks and no longer require global `views:read:*` access.

## Motivation and Context

The previous attempt to check resolved view permissions (in https://github.com/Graylog2/graylog2-server/pull/12287) was not complete, because it required users to have global views read permissions (`views:read:*`) in order to access any resolved views. This is not acceptable, because global view access cannot be required to access specific resolved views. Also, a view resolver might require that another particular permission be checked in order to validate access to the view (which completely depends on the view resolver implementation).

## How Has This Been Tested?

See testing comment in linked PR.

## Types of changes

## Checklist:

- [*] My code follows the code style of this project.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/permissions/SearchUser.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/permissions/SearchUser.java
@@ -22,25 +22,34 @@ import org.graylog.plugins.views.search.rest.PermittedStreams;
 import org.graylog.plugins.views.search.rest.ViewsRestPermissions;
 import org.graylog.plugins.views.search.views.ViewDTO;
 import org.graylog.plugins.views.search.views.ViewLike;
+import org.graylog.plugins.views.search.views.ViewResolver;
+import org.graylog.plugins.views.search.views.ViewResolverDecoder;
 import org.graylog2.plugin.database.users.User;
 import org.graylog2.shared.security.RestPermissions;
 import org.joda.time.DateTimeZone;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 
 public class SearchUser implements SearchPermissions, StreamPermissions, ViewPermissions {
+    private static final Logger LOG = LoggerFactory.getLogger(SearchUser.class);
     private final User currentUser;
     private final Predicate<String> isPermitted;
     private final BiPredicate<String, String> isPermittedEntity;
     private final UserStreams userStreams;
+    private final Map<String, ViewResolver> viewResolvers;
 
-    public SearchUser(User currentUser, Predicate<String> isPermitted, BiPredicate<String, String> isPermittedEntity, PermittedStreams permittedStreams) {
+    public SearchUser(User currentUser, Predicate<String> isPermitted, BiPredicate<String, String> isPermittedEntity,
+                      PermittedStreams permittedStreams, Map<String, ViewResolver> viewResolvers) {
         this.currentUser = currentUser;
         this.isPermitted = isPermitted;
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewResolver.java
```diff
@@ -44,4 +46,22 @@
...
+     * @param viewId                  The id of the view.
+     * @param permissionTester        A predicate, which tests a single string permission parameter.
+     *                                For example, is the user authorized to read a view based on having of a
+     *                                particular permission?
+     * @param entityPermissionsTester A bi-predicate which tests two parameters (the permission to test and the id of
+     *                                the entity). For example, is the user authorized to read a view based on having
+     *                                a particular permission and/or having access to a particular entity?
+     */
+    boolean canReadView(String viewId, Predicate<String> permissionTester,
+                        BiPredicate<String, String> entityPermissionsTester);
```
danotorrey, line 66: The addition of this method, allows view resolvers to provide their own specific permissions check as appropriate for the view view resolver. Both the single permission and permission/id predicates are passed as arguments, to allow the view resolver to decide which specific permissions/combination to check.

Passing the permission check predicates in also eliminates the need to inject the security principal into the view resolver implementation to check the permission.

graylog2-server/src/main/java/org/graylog/plugins/views/search/permissions/SearchUser.java
```diff
@@ -53,6 +62,21 @@ public String username() {
...
+        if (decoder.isResolverViewId()) {
+            final ViewResolver viewResolver = viewResolvers.get(decoder.getResolverName());
+            if (viewResolver != null) {
+                return viewResolver.canReadView(viewId, isPermitted, isPermittedEntity);
+            } else {
+                // Resolved view could not be found, so permissions cannot be checked.
+                LOG.info("View resolver [{}] could not be found.", decoder.getResolverName());
+                return false;
+            }
+        }
```
danotorrey, line 77: This implementation adds view resolver specific resolution into the `SearchUser` class. 

The positive: It's centralized, and the same permissions checks will be used everywhere uniformly (whether or not permissions are checked for a database view or a resolved view).

The negative: The conditional view resolver logic feels slightly out of place here. Are the concerns of DB-persisted views and resolved views combined too much here? I could not find a way to separate them. The only alternative I could think of, would be to perform the resolved view permissions checks completely separately from those of the DB views (like up in the resource classes). But that brings its own issues (for one, externalizing/distributing permissions checks into rest logic did not seem like an improvement). Feedback and alternate suggestions welcome.

danotorrey:
Note that unit tests on this PR need more work. But, I am requesting a review on this PR to get feedback on the overall approach sooner. I appreciate everyone's time looking into this. This resolves the 4.3 blocker issue https://github.com/Graylog2/graylog-plugin-enterprise/issues/3336, so we hoping to reach a final solution as quickly as possible.

roberto-graylog (commented):
It looks good to me.

kingzacko1:
I'm not super familiar with views and their permissions, but this approach makes sense to me and looks good.

danotorrey:
I have added tests covering view and resolved view permissions checks in https://github.com/Graylog2/graylog2-server/pull/12367/commits/3923f275910ff5d88f6e916937c0bf6d46b70fa2. I feel much more comfortable with the changes now that they are covered by at least some minimal unit tests.

roberto-graylog (approved):

kingzacko1 (approved):

janheise (approved):
LGTM

danotorrey:
I will merge this PR now for inclusion in the RC (thinking there is some benefit in more testing sooner). But, if anyone sees any issues or changes needed, please let me know and I can open a new PR.

PR #12370, opened 2022-03-29
Reset time based rotation strategy after leader change

thll:
The time-based rotation strategy remembers the last time an index was rotated in order to calculate the next rotation point.

When a node loses the leader state, it will stop being responsible for index rotation. The time-based rotation strategy, however, because it is a singleton, will still reference the last rotation times.

If for any reason, the node becomes leader again after a while, it will be unaware of any previous index rotations and will make a rotation decision based on stale data. If any rotations have happened in between, triggered by the other node, the now new leader will immediately rotate the index.

This is a rare scenario to happen, so I don't think this is critical. But nevertheless good to fix.

The proper way to fix this would be to remove the state from the rotation strategy. This might however introduce slightly different behaviour for some rotation periods. Also, changing the logic of the rotation strategy itself is a delicate undertaking, which is why I opted for a low-risk solution.


mpfz0r (approved):
LGTM

PR #12382, opened 2022-04-01
Use FakeLeaderElectionModule for Migration command

mpfz0r:
Migrations should not need to deal with the leader election at all.


thll (approved):
LGTM

PR #12387, opened 2022-04-04
Support config of sidecars on Darwin and FreeBSD

thll:
Adds missing validation and config options for Darwin and FreeBSD.

They were already supported, but it was not possible to configure them.

Fixes https://github.com/Graylog2/collector-sidecar/issues/377


mpfz0r (approved):
LGTM

PR #12388, opened 2022-04-04
[WIP] Posthog features for core

Gaya:
## Description

Add the following features to core to support PostHog:
- Saving user preference for tracking
- Capture groups on HTML elements so they show up nicely in PostHog
- Expose new PostHog specific settings in `config.js`
- Adds a component which allows showing components only on cloud

## Motivation and Context

To support PostHog in cloud there needed to be a few changes to core. Will not break the current behaviour, only adds to it. Does not depend on cloud PRs to work, but is recommended to be merged around the same time.

See: https://github.com/Graylog2/graylog-plugin-cloud/pull/1009

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:


Gaya:
btw: linter errors are from older files

ousmaneo (approved):
Tested and looks, good 👍

graylog2-web-interface/src/components/users/UserEdit/PreferencesSection.tsx
```diff
@@ -84,6 +85,19 @@ const PreferencesSection = ({ user }: Props) => {
                                value={user.preferences?.[PREFERENCES_THEME_MODE] ?? 'Not configured'}
                                help="Can be changed by using the toggle in the user dropdown" />
 
+            <OnlyOnCloud>
+              <Input id="user-diagnostics-controls"
```
edmundoa, line 89: I have some comments on this. It is not an issue with the PR itself and I believe the UI is still not fully finished, but I still think it is worth writing them now:

I like that it is possible to anonymize usage information with a single checkbox. However, I think we should make a better job at showing cloud users that we are collecting this information and opting out. Additionally, we could also invert that checkbox to default to the anonymized version, so that we only collect the data we need and let cloud users decide if they want to share more data.

edmundoa (approved):

graylog2-web-interface/src/components/users/UserEdit/PreferencesSection.tsx
edmundoa: Yes, I think that is important, and maybe we have a plan for that already. Otherwise people may not even realise we are collecting some data, since the on premise version doesn't do it and the checkbox is not really visible unless you are looking for it.

kroepke: Yes, we are preparing the opt-out path for cloud users. Those need updates to the terms and conditions, which are not finalized yet, which is why we have it like this for now. I think we'll need to iterate a bit on the final state before this goes live. The final version might move the checkbox somewhere else (cloud console for the entire account, perhaps)

Gaya: Let me know if there is anything I can do today or Thurday to finalise this, or I'll just leave the PRs open for further development.

PR #12404, opened 2022-04-06
Always execute search, when pressing enter in search bar form.

linuspahl:
**Please note**, this PR needs to be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/3451

## Motivation and Context

Before this change we were not reexecuting the search on enter press (in the query input or streams select), when the form values did not changed.

This was only the case when clicking on the search submit button. The reason, when there is no change:
- the search submit button did not triggered a form submit, but `SearchActions.refresh`, which triggered the rexecution.
- pressing enter in e.g. the query input triggered a search submit, which updates the query. Updating the current query also triggers a search submit, but only when the query actually changed.

With this PR we are:
- always submitting the search form, when clicking on the search submit button.
- we are triggering a search execution on form submit, even when the query did not changed.

This PR is fixing the behaviour the the search page and widget edit mode.

We should create a backport for this change.

## Types of changes

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3451

### DIFF
--- a/graylog2-web-interface/src/views/components/SearchBar.tsx
+++ b/graylog2-web-interface/src/views/components/SearchBar.tsx
@@ -144,7 +144,11 @@ const SearchBar = ({
   const streams = filtersToStreamSet(queryFilters.get(id, Immutable.Map())).toJS();
   const limitDuration = moment.duration(config.query_time_range_limit).asSeconds() ?? 0;
 
-  const _onSubmit = (values: SearchBarFormValues) => onSubmit(values, currentQuery);
+  const _onSubmit = (values: SearchBarFormValues, ...rest) => {
+    console.log({ rest });
+
+    return onSubmit(values, currentQuery);
+  };
 
   return (
     <WidgetFocusContext.Consumer>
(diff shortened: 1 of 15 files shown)


graylog2-web-interface/src/views/components/SearchBar.tsx
```diff
@@ -95,7 +96,11 @@ const defaultOnSubmit = ({ timerange, streams, queryString }, currentQuery: Quer
     .query(createElasticsearchQueryString(queryString))
     .build();
 
-  return QueriesActions.update(newQuery.id, newQuery);
+  if (!currentQuery.equals(newQuery)) {
+    return QueriesActions.update(newQuery.id, newQuery);
```
linuspahl, line 100: `QueriesActions.update` also contains a check if the query changed. Doing the check here as well still seems to be  a good solution, since we are using the action in other places as well

graylog2-web-interface/src/views/components/SearchBar.tsx
```diff
@@ -95,7 +96,11 @@ const defaultOnSubmit = ({ timerange, streams, queryString }, currentQuery: Quer
...
+  }
+
+  return SearchActions.refresh();
```
linuspahl, line 103: We need to return a promise here, otherwise formik will not update `isSubmitting`. Since we expect the action to return a promise, I created a new method for the store, to actually return a promise.

maxiadlovskii (approved):

PR #12405, opened 2022-04-06
Graylog2/graylog-plugin-enterprise/issues/3419

patrickmann:
See related issue for details

Resolves Graylog2/graylog-plugin-enterprise#3419

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
@@ -32,6 +32,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 import java.util.Collections;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.Set;
 
 public class MigrationHelpers {
@@ -97,7 +98,7 @@ public class MigrationHelpers {
             previousUser = userService.load(userName);
             if (previousUser == null
                     || !previousUser.getRoleIds().containsAll(expectedRoles)
-                    || (isServiceAccount && !previousUser.isServiceAccount())) {
+                    || !Objects.equals(isServiceAccount, previousUser.isServiceAccount())) {
                 final String msg = "Invalid user '" + userName + "', fixing it.";
                 LOG.error(msg);
                 throw new IllegalArgumentException(msg);
--- a/graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
@@ -86,10 +86,18 @@ public class MigrationHelpers {
     @Nullable
     public String ensureUser(String userName, String password, String firstName, String lastName, String email,
                              Set<String> expectedRoles) {
+        return ensureUser(userName, password, firstName, lastName, email, expectedRoles, false);
+    }
+
+    @Nullable
+    public String ensureUser(String userName, String password, String firstName, String lastName, String email,
+                             Set<String> expectedRoles, boolean isServiceAccount) {
         User previousUser = null;
         try {
             previousUser = userService.load(userName);
-            if (previousUser == null || !previousUser.getRoleIds().containsAll(expectedRoles)) {
+            if (previousUser == null
+                    || !previousUser.getRoleIds().containsAll(expectedRoles)
+                    || (isServiceAccount && !previousUser.isServiceAccount())) {
                 final String msg = "Invalid user '" + userName + "', fixing it.";
                 LOG.error(msg);
                 throw new IllegalArgumentException(msg);
@@ -100,6 +108,7 @@ public class MigrationHelpers {
             if (previousUser != null) {
                 fixedUser = previousUser;
                 fixedUser.setRoleIds(expectedRoles);
+                fixedUser.setServiceAccount(isServiceAccount);
             } else {
                 fixedUser = userService.create();
                 fixedUser.setName(userName);
@@ -109,6 +118,7 @@ public class MigrationHelpers {
                 fixedUser.setPermissions(Collections.emptyList());
                 fixedUser.setRoleIds(expectedRoles);
                 fixedUser.setTimeZone(DateTimeZone.UTC);
+                fixedUser.setServiceAccount(isServiceAccount);
             }
             try {
                 return userService.save(fixedUser);
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
@@ -68,6 +68,10 @@ public abstract class CreateUserRequest {
     @Nullable
     public abstract List<String> roles();
 
+    @JsonProperty
+    @Nullable
+    public abstract Boolean isServiceAccount();
+
     @JsonCreator
     public static CreateUserRequest create(@JsonProperty("username") @NotEmpty String username,
                                            @JsonProperty("password") @NotEmpty String password,
@@ -78,8 +82,9 @@ public abstract class CreateUserRequest {
                                            @JsonProperty("timezone") @Nullable String timezone,
                                            @JsonProperty("session_timeout_ms") @Nullable @Min(1) Long sessionTimeoutMs,
                                            @JsonProperty("startpage") @Nullable Startpage startpage,
-                                           @JsonProperty("roles") @Nullable List<String> roles) {
+                                           @JsonProperty("roles") @Nullable List<String> roles,
+                                           @JsonProperty("service_account") @Nullable Boolean isServiceAccount) {
         return new AutoValue_CreateUserRequest(username, password, email, firstName, lastName,
-                                               permissions, timezone, sessionTimeoutMs, startpage, roles);
+                                               permissions, timezone, sessionTimeoutMs, startpage, roles, isServiceAccount);
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
@@ -69,8 +69,7 @@ public abstract class CreateUserRequest {
     public abstract List<String> roles();
 
     @JsonProperty
-    @Nullable
-    public abstract Boolean isServiceAccount();
+    public abstract boolean isServiceAccount();
 
     @JsonCreator
     public static CreateUserRequest create(@JsonProperty("username") @NotEmpty String username,
@@ -85,6 +84,7 @@ public abstract class CreateUserRequest {
                                            @JsonProperty("roles") @Nullable List<String> roles,
                                            @JsonProperty("service_account") @Nullable Boolean isServiceAccount) {
         return new AutoValue_CreateUserRequest(username, password, email, firstName, lastName,
-                                               permissions, timezone, sessionTimeoutMs, startpage, roles, isServiceAccount);
+                                               permissions, timezone, sessionTimeoutMs, startpage, roles,
+                isServiceAccount != null && isServiceAccount);
     }
 }
--- a/graylog2-server/src/main/java/org/graylog2/rest/models/users/responses/UserSummary.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/models/users/responses/UserSummary.java
@@ -108,6 +108,9 @@ public abstract class UserSummary {
     @JsonProperty("account_status")
     public abstract User.AccountStatus accountStatus();
 
+    @JsonProperty("service_account")
+    public abstract boolean isServiceAccount();
+
     @JsonCreator
     public static UserSummary create(@JsonProperty("id") @Nullable String id,
                                      @JsonProperty("username") String username,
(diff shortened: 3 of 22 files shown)


graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
```diff
@@ -86,10 +86,18 @@ public String ensureBuiltinRole(String roleName, String description, Set<String>
...
         User previousUser = null;
         try {
             previousUser = userService.load(userName);
-            if (previousUser == null || !previousUser.getRoleIds().containsAll(expectedRoles)) {
+            if (previousUser == null
+                    || !previousUser.getRoleIds().containsAll(expectedRoles)
+                    || (isServiceAccount && !previousUser.isServiceAccount())) {
                 final String msg = "Invalid user '" + userName + "', fixing it.";
                 LOG.error(msg);
                 throw new IllegalArgumentException(msg);
```
kingzacko1, line 103: I realize this isn't part of your changes, but just for my own curiosity, is there a benefit from throwing an exception and immediately ignoring it in the following lines? Any reason not to remove the try-catch and just handle the logic in lines 108-122 here instead of in the catch?

bernd (changes requested):
I will take a closer look tomorrow. But I think this PR is not complete yet.

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -417,6 +420,17 @@ public void changeUser(@ApiParam(name = "userId", value = "The ID of the user to
...
         }
+
+        if (user.isServiceAccount() && !cr.isServiceAccount()) {
+            // restore permissions that were removed from the service account user
+            final Set<String> selfEditPermissions = new Permissions(new HashSet<>()).userSelfEditPermissions(user.getName());
+            final List<String> userPermissions = user.getPermissions();
+            if (user.getPermissions().addAll(selfEditPermissions)) {
+                user.setPermissions(userPermissions);
+            }
+        }
```
bernd, line 431: The self-edit permissions are added dynamically to a user at runtime in `UserImpl#getPermissions()` and are not stored in the database. `UserImpl#setPermissions()` also removes any self-edit permissions from the user, so the code is actually a no-op and should be removed.

graylog2-server/src/main/java/org/graylog2/shared/users/ChangeUserRequest.java
```diff
@@ -75,7 +79,8 @@ public static ChangeUserRequest create(@JsonProperty("email") @Nullable @Email S
                                            @JsonProperty("timezone") @Nullable String timezone,
                                            @JsonProperty("startpage") @Nullable @Valid Startpage startpage,
                                            @JsonProperty("session_timeout_ms") @Nullable @Min(1) Long sessionTimeoutMs,
-                                           @JsonProperty("roles") @Nullable List<String> roles) {
-        return new AutoValue_ChangeUserRequest(email, firstName, lastName, permissions, timezone, startpage, sessionTimeoutMs, roles);
+                                           @JsonProperty("roles") @Nullable List<String> roles,
+                                           @JsonProperty("service_account") @Nullable Boolean isServiceAccount) {
+        return new AutoValue_ChangeUserRequest(email, firstName, lastName, permissions, timezone, startpage, sessionTimeoutMs, roles, isServiceAccount);
```
bernd, line 84: Let's handle a `null` value for `isServiceAccount` here. (see comment above)

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
```diff
@@ -226,7 +227,7 @@ public void setEmail(final String email) {
 
     @Override
     public List<String> getPermissions() {
-        final Set<String> permissionSet = new HashSet<>(this.permissions.userSelfEditPermissions(getName()));
+        final Set<String> permissionSet = isServiceAccount() ? new HashSet<>() : new HashSet<>(this.permissions.userSelfEditPermissions(getName()));
```
bernd, line 230: Please check all usages of `Permissions#userSelfEditPermissions()` if we need to add more `isServiceAccount` checks. There is at least one usage in `PermissionsResource` where I am not sure if we need to check for a service account or not.

graylog2-server/src/main/java/org/graylog2/users/UserOverviewDTO.java
```diff
@@ -107,6 +108,10 @@
     @JsonProperty(FIELD_ACCOUNT_STATUS)
     public abstract User.AccountStatus accountStatus();
 
+    @Nullable
+    @JsonProperty(FIELD_SERVICE_ACCOUNT)
+    public abstract Boolean serviceAccount();
```
bernd, line 113: Let's make this `boolean` and remove the `@Nullable` annotation. That makes the usage of the property easier because consumers don't have to do `null` checks.

bernd (changes requested):
Some comments below.

graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
```diff
@@ -68,6 +68,10 @@
     @Nullable
     public abstract List<String> roles();
 
+    @JsonProperty
+    @Nullable
+    public abstract Boolean isServiceAccount();
```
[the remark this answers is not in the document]
patrickmann: Darn - just added null checks everywhere. But you are right - KISS.

graylog2-server/src/main/java/org/graylog2/rest/models/users/responses/UserSummary.java
```diff
@@ -127,25 +130,27 @@ public static UserSummary create(@JsonProperty("id") @Nullable String id,
                                      @JsonProperty("session_active") boolean sessionActive,
                                      @JsonProperty("last_activity") @Nullable Date lastActivity,
                                      @JsonProperty("client_address") @Nullable String clientAddress,
-                                     @JsonProperty("account_status") User.AccountStatus accountStatus) {
+                                     @JsonProperty("account_status") User.AccountStatus accountStatus,
+                                     @JsonProperty("service_account") boolean isServiceAccount) {
```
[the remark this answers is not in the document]
patrickmann: It actually did work in testing. But I will change it to be consistent and well defined.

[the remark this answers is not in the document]
bernd: @patrickmann If it works then we can adjust the other auto-values as well to not use `Boolean` but `boolean` in the `@JsonCreator` methods. :slightly_smiling_face:

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
patrickmann: `PermissionsResource` and `Permissions` both have a `readerPermissions()` method that call `userSelfEditPermissions()`. Currently neither class accesses the DB - they operate purely on permissions. The user name is just used to build permission strings. 
`Permissions#readerPermissions` is only used in unit tests. 
`PermissionsResource#readerPermissions` is not currently used in the frontend.

I think it's reasonable to keep these classes agnostic of service accounts.

graylog2-server/src/main/java/org/graylog2/rest/models/users/requests/CreateUserRequest.java
```diff
@@ -78,8 +82,9 @@ public static CreateUserRequest create(@JsonProperty("username") @NotEmpty Strin
                                            @JsonProperty("timezone") @Nullable String timezone,
                                            @JsonProperty("session_timeout_ms") @Nullable @Min(1) Long sessionTimeoutMs,
                                            @JsonProperty("startpage") @Nullable Startpage startpage,
-                                           @JsonProperty("roles") @Nullable List<String> roles) {
+                                           @JsonProperty("roles") @Nullable List<String> roles,
+                                           @JsonProperty("service_account") @Nullable Boolean isServiceAccount) {
         return new AutoValue_CreateUserRequest(username, password, email, firstName, lastName,
-                                               permissions, timezone, sessionTimeoutMs, startpage, roles);
+                                               permissions, timezone, sessionTimeoutMs, startpage, roles, isServiceAccount);
```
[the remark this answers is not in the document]
patrickmann: fixed

graylog2-server/src/main/java/org/graylog2/shared/users/ChangeUserRequest.java
patrickmann: fixed

graylog2-server/src/main/java/org/graylog2/users/UserImpl.java
bernd: Alright, sounds good. Thanks for the investigation. :+1:

graylog2-web-interface/src/components/users/UserCreate/ServiceAccountFormGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+          confirmationModalRef.current.close();
+        };
+
+        return (
+          <>
+            <Input id="service-account-controls"
+                   labelClassName="col-sm-3"
+                   wrapperClassName="col-sm-9"
+                   label="Service Account">
+              <Input label="Service Account"
```
bernd, line 56: The duplicated "Service Account" value looks a bit strange. Can we extend the value right of the check box like we did for the "Session Timeout"? 

image

graylog2-web-interface/src/components/users/UserDetails/SettingsSection.tsx
```diff
@@ -67,12 +67,14 @@ const StartpageValue = ({ type, id }: { type: string | null | undefined, id: str
...
   user: {
     timezone,
+    serviceAccount,
     sessionTimeout,
     startpage,
   },
 }: Props) => (
   <SectionComponent title="Settings">
     <ReadOnlyFormGroup label="Sessions Timeout" value={_sessionTimeout(sessionTimeout)} />
+    <ReadOnlyFormGroup label="Service Account" value={serviceAccount} />
```
bernd, line 77: Regular reader users can set the "Service Account" attribute on their profile and effectively lock themselves out of the web interface. We might want to introduce a separate permission string for this flag or re-use one that only admin users have. (e.g., `users:edit`)

If users don't have the permission to change the value, it should also not show up in the form.

bernd (changes requested):
When a user object has `service_account: true`, the user will not be able to log into the web interface.

I think we should improve the behavior when a service account user tries to log in. The web interface is currently showing this instead of an information message.

image

bernd:
> When a user object has `service_account: true`, the user will not be able to log into the web interface.
> 
> I think we should improve the behavior when a service account user tries to log in. The web interface is currently showing this instead of an information message.

The user is also unable to log out again because the user interface will always show the "preparing" message.

bernd:
> > When a user object has `service_account: true`, the user will not be able to log into the web interface.
> > I think we should improve the behavior when a service account user tries to log in. The web interface is currently showing this instead of an information message.
> 
> The user is also unable to log out again because the user interface will always show the "preparing" message.

I think the backend shouldn't allow the creation of a session for service account users.

graylog2-web-interface/src/components/users/UserCreate/ServiceAccountFormGroup.tsx
```diff
@@ -0,0 +1,79 @@
...
+          confirmationModalRef.current.close();
+        };
+
+        return (
+          <>
+            <Input id="service-account-controls"
+                   labelClassName="col-sm-3"
+                   wrapperClassName="col-sm-9"
+                   label="Service Account">
+              <Input label="User is a Service account"
```
bernd, line 56: Let's keep the spelling of "Service Account" on this page consistent.
```suggestion
              <Input label="User is a Service Account"
```

bernd (changes requested):
Three small comments. Apart from those, it looks good! Thank you!

graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
```diff
@@ -86,10 +86,18 @@ public String ensureBuiltinRole(String roleName, String description, Set<String>
...
+    @Nullable
+    public String ensureUser(String userName, String password, String firstName, String lastName, String email,
+                             Set<String> expectedRoles, boolean isServiceAccount) {
         User previousUser = null;
         try {
             previousUser = userService.load(userName);
-            if (previousUser == null || !previousUser.getRoleIds().containsAll(expectedRoles)) {
+            if (previousUser == null
+                    || !previousUser.getRoleIds().containsAll(expectedRoles)
+                    || (isServiceAccount && !previousUser.isServiceAccount())) {
```
bernd, line 100: Hmm, I think we need to add a check for the opposite case as well. `isServiceAccount` is `false` but `previousUser.isServiceAccount` is true. I guess the following should work?

```suggestion
                    || !Objects.equals(isServiceAccount, previousUser.isServiceAccount())) {
```

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
```diff
@@ -417,9 +419,23 @@ public void changeUser(@ApiParam(name = "userId", value = "The ID of the user to
...
             }
         }
+
+        user.setServiceAccount(cr.isServiceAccount());
+
         userManagementService.update(user, cr);
     }
 
+    private void checkAdminRoleForServiceAccount(ChangeUserRequest cr, User user) {
+        if (user.isServiceAccount() && cr.roles() != null && cr.roles().contains("Admin")) {
```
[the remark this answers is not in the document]
patrickmann: I was hesitant to make that public, but I hate string literals too :)
fixed

graylog2-web-interface/src/components/users/UserCreate/ServiceAccountFormGroup.tsx
patrickmann: fixed

graylog2-server/src/main/java/org/graylog2/rest/resources/system/SessionsResource.java
```diff
@@ -148,6 +152,15 @@ public Response newSession(@Context ContainerRequestContext requestContext,
         }
     }
 
+    private void rejectServiceAccount(JsonNode createRequest) {
+        if (createRequest.has(USERNAME)) {
+            final User user = userService.load(createRequest.get(USERNAME).asText());
+            if (user.isServiceAccount()) {
```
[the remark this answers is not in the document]
patrickmann: I should pay attention to Sonar

bernd (changes requested):
What about this one? https://github.com/Graylog2/graylog2-server/pull/12405/files#r846072753

You resolved the comment but there hasn't been a code change or reply for it.

graylog2-server/src/main/java/org/graylog2/migrations/MigrationHelpers.java
patrickmann: Oops - missed that. I don't think this is a scenario that can happen with the existing calls to ensureUser(). But it is what you'd expect ensureUser() to do.

bernd (changes requested):
Unfortunately, I found another issue.

It looks like assigning or removing a role to/from a user clears the `service_account` flag on the user object.

Steps to reproduce:

1. Start Graylog server
2. Verify that `graylog-sidecar` has `service_account: true`
3. Remove the `Sidecar System` role from the `graylog-sidecar` user
4. Check `service_account` and observe `false`

The problem could be related to the fact that the UI is only sending a `{"roles":["Reader"]}` payload instead of all fields when updating the user.
We have to check if anything else is only sending a partial payload for updating users.

patrickmann:
Good catch - changed `isServiceAccount` back to a `@nullable Boolean` in `ChangeUserRequest`.

> We have to check if anything else is only sending a partial payload for updating users.

We only update the field in `User` via `setServiceAccount()`. There are no other instances where we implicitly overwrite the field.

bernd (approved):
Thak you! Looks good now! :+1:

PR #12415, opened 2022-04-07
QueryValidator interface, unknown fields adapted

todvora:
Introduced a QueryValidation interface, that could be implemented by several validators, each handling a small portion of the complex validation. Adapted UnknownFieldsIdentifier to implement this interface. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/fields/UnknownFieldsIdentifier.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/fields/UnknownFieldsIdentifier.java
@@ -17,20 +17,24 @@
 package org.graylog.plugins.views.search.validation.fields;
 
 import org.graylog.plugins.views.search.rest.MappedFieldTypeDTO;
+import org.graylog.plugins.views.search.validation.ParsedQuery;
 import org.graylog.plugins.views.search.validation.ParsedTerm;
+import org.graylog.plugins.views.search.validation.QueryValidator;
+import org.graylog.plugins.views.search.validation.ValidationMessage;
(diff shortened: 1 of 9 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/fields/UnknownFieldsIdentifier.java
```diff
@@ -17,20 +17,24 @@
...
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 @Singleton
-public class UnknownFieldsIdentifier {
+public class UnknownFieldsIdentifier implements QueryValidator {
```
luk-kaminski, line 37: I'd consider keeping original UnknownFieldsIdentifier and introducing separate UnknownFieldsValidator, which just uses the identifier. The current approach combines two responsibilities in this class (parsing and validating) and forces you to to use parsing in the unit test, making this unit test not a pure "unit test"...

todvora: The query parsing is now moved outside of the scope of this class. The identification of unknown fields is needed only to for the validation, so I see it as one responsibility only. And the scope is so small that I'd rather keep everything inside one class than to split it further into two different. I have refactored the code since your comment. Is it for better of worse from your POV?

luk-kaminski: Ok!

luk-kaminski (approved):
Good job!

PR #12420, opened 2022-04-07
Fix parsing of invalid search operators

todvora:
There was a bug in our detection of invalid operators that used terms, not parsed tokens. Tokens have more knowledge of the context and reflect the quoted text, so they are the correct place to look for invalid operators. 

## Description
Before:
image

After:
image

## How Has This Been Tested?
Added integration test

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/QueryValidationResourceIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/QueryValidationResourceIT.java
@@ -18,7 +18,6 @@ package org.graylog.plugins.views;
 
 import io.restassured.response.ValidatableResponse;
 import io.restassured.specification.RequestSpecification;
-import org.checkerframework.checker.units.qual.A;
 import org.graylog.testing.completebackend.GraylogBackend;
 import org.graylog.testing.containermatrix.annotations.ContainerMatrixTest;
 import org.graylog.testing.containermatrix.annotations.ContainerMatrixTestsConfiguration;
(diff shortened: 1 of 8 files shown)


full-backend-tests/src/test/java/org/graylog/plugins/views/QueryValidationResourceIT.java
```diff
@@ -30,6 +31,7 @@
 import static org.graylog.testing.completebackend.Lifecycle.CLASS;
 import static org.graylog.testing.graylognode.NodeContainerConfig.GELF_HTTP_PORT;
 import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.core.IsEqual.equalTo;
 
 @ContainerMatrixTestsConfiguration(serverLifecycle = CLASS)
```
luk-kaminski, line 37: It is completely unrelated to your fix Tomas, but I am just wondering... is validation dependent on MongoDB or ES versions? If not, maybe we should limit the number of tests to just one Mongo and one ES version? It would speed up execution on Jenkins.

luk-kaminski (approved):
Great job! I've been testing and could not break it this time! :)

full-backend-tests/src/test/java/org/graylog/plugins/views/QueryValidationResourceIT.java
```diff
@@ -18,6 +18,7 @@
 
 import io.restassured.response.ValidatableResponse;
 import io.restassured.specification.RequestSpecification;
+import org.checkerframework.checker.units.qual.A;
```
dennisoelkers, line 21: I guess this import slipped in?

todvora: Indeed! Thanks, fixed!

full-backend-tests/src/test/java/org/graylog/plugins/views/QueryValidationResourceIT.java
todvora: You are right, configured the default versions. Good idea, thanks!

PR #12459, opened 2022-04-12
Log cause of failed preflight check

thll:
A failing preflight check was not logged but only written to STDERR.

This leads to the server stopping without any indication whatsoever of what might have caused it if you look only at the log file.

This PR adds the missing logging.

Steps to reproduce (prior to applying this PR):

* Start the server with a functioning journal
* Make the journal directory unwritable, e.g. with `chmod u-w journal`
* Start graylog again and notice that it stops again with the last line of the log file being something like ```2022-04-12 10:33:09,112 INFO : org.graylog2.bootstrap.preflight.SearchDbPreflightCheck - Connected to (Elastic/Open)Search version <Elasticsearch:7.10.2>```


mpfz0r (approved):
LGTM. Thanks for picking this up :pray:

Issue #12462, opened 2022-04-12
Validation pop-ups don't show when creating inputs


mpfz0r:
With Graylog 4.3 validation messages like the one below are not shown anymore.

![image](https://user-images.githubusercontent.com/3034831/162938718-f913e8b0-5485-43aa-9d14-fe64dc4773b0.png)

I think it got broken with this PR: https://github.com/graylog2/graylog2-server/pull/12189

## Your Environment

* Graylog Version: 4.3

Issue #12314, opened 2022-03-18
Import Grok Pattern upload button does nothing 


ousmaneo:
Selecting a file containing GROK patterns and clicking submit does nothing 
![grok](https://user-images.githubusercontent.com/1381455/158998379-8670f014-1d98-4ccd-89d0-42e11142a764.gif)

## Steps to Reproduce (for bugs)

1. Go to Grok Page 
2. Click on "Import Pattern file"
3. click "Upload"
4.

## Your Environment

* Graylog Version: 4.3-beta2

thll:
My first thought was that this is related to https://github.com/Graylog2/graylog2-server/pull/12174. However, if I revert that PR, the button is still not functional.

ousmaneo:
I did some tests and somehow the submit is not triggered with `formProps={{ onSubmit: this._onSubmit }}` and adding `onSubmitForm={this._onSubmit}` fixed it. I will provide a PR.

PR #12467, opened 2022-04-12
Revert changes in `BootstrapModalForm` and reimplement `type="submit"` for submit button. (backport of #12465 for 4.3)

linuspahl:
_Please note, this PR is a backport of #12465 for 4.3 and should be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/3463_

## Motivation and Context

With https://github.com/Graylog2/graylog2-server/pull/12189 we changed the way the `BootstrapModalForm` is being submitted. Instead of using `type="submit"` for the submit button we called the `onSubmit` action directly. Before the mentioned change, the form submit could influence a parent form, when using formik, by setting its `isSubmitting` attribute. This problem occurred when declaring a parameter, by clicking on the "declare parameter" button in the query validation popover.

Changing the way we submit the form resulted in other problems (https://github.com/Graylog2/graylog2-server/issues/12314 and https://github.com/Graylog2/graylog2-server/issues/12462). Fortunately it is possible to avoid this behaviour by stopping the event propagation.

There is a linter error I did not fixed, to avoid introducing further bugs.

## Types of changes

### DIFF
--- a/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/BootstrapModalForm.jsx
@@ -91,6 +91,7 @@ class BootstrapModalForm extends React.Component {
     const { onSubmitForm } = this.props;
 
     if (typeof onSubmitForm === 'function') {
+      event.preventDefault();
       onSubmitForm(event);
     }
   };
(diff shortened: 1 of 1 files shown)


dennisoelkers (approved):
Looks good, works fine!

PR #12475, opened 2022-04-13
Fix security tools placeholder route error

mikedklein:
Fix for security page error when enterprise plugin isn't installed.

## Description

Addresses an error when the enterprise plugin isn't installed. When a user navigates to the security page without the enterprise plugin installed it will render correctly.

## Motivation and Context
Before this when a user went to the security page it would error because it was trying to use a link from the plugin routes. Now if the enterprise plugin isn't installed it falls back to link to the enterprise information page.

## How Has This Been Tested?

Tested locally with the enterprise plugin removed, enterprise plugin installed but no security license and with enterprise plugin installed with security license active.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/pages/SecurityPage.tsx
+++ b/graylog2-web-interface/src/pages/SecurityPage.tsx
@@ -18,15 +18,20 @@
 import * as React from 'react';
 import styled from 'styled-components';
 
-import { IfPermitted, PageHeader } from 'components/common';
+import { IfPermitted } from 'components/common';
 import { Alert } from 'components/bootstrap';
 import { Link } from 'components/common/router';
 import Routes from 'routing/Routes';
 import AppConfig from 'util/AppConfig';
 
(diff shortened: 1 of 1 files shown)


danotorrey (changes requested):
This tests successfully! Thanks for getting this in so quickly! 

I noticed one very small thing I am wondering if we can adjust:

There are two spaces on both sides of the "our product page" link instead of a single space (hard to see, but can be seen by selecting with the cursor). Is it possible to reduce that to a single space?

image
image

Note: Due to the master package issue, I was only able to test this by cherry-picking the commit over to a `4.3` checkout.

danotorrey:
I am finishing testing this now.

danotorrey (approved):
LGTM and tests successfully

Server only with no license or enterprise plugin:
image

Server with no license and enterprise plugins:
image

kingzacko1 (approved):

PR #12483, opened 2022-04-14
Skeleton of migration of streams in search documents.

luk-kaminski:
Skeleton of migration of streams in search documents, from filter field to own streams field.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/ViewsBindings.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/ViewsBindings.java
@@ -30,6 +30,7 @@ import org.graylog.plugins.views.migrations.V20191204000000_RemoveLegacyViewsPer
 import org.graylog.plugins.views.migrations.V20200204122000_MigrateUntypedViewsToDashboards.V20200204122000_MigrateUntypedViewsToDashboards;
 import org.graylog.plugins.views.migrations.V20200409083200_RemoveRootQueriesFromMigratedDashboards;
 import org.graylog.plugins.views.migrations.V20200730000000_AddGl2MessageIdFieldAliasForEvents;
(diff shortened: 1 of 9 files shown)


todvora:
I'd try to keep the migration as independent and self-contained as possible. If you want to work with raw mongo documents or create own java classes for them, used through the migration - both is ok from my view. I would just try to limit dependencies (and imports) to absolute minimum. 

The API versioning serves to a different purpose and operates on the REST level, not in the services itself. One of the reasons we introduces it was to decouple sevice DTOs from req/res classes, so we can change things independently.

dennisoelkers:
@luk-kaminski: In this case we are only interested in a very small subset of fields in the searches, so I would recommend operating directly on the DB instead of mapping back and forth between duplicated DTOs. Also, could you add some test cases with DB fixtures to show that the migration is working as expected? What happens with searches which cannot be migrated, do they turn out to be unreadable afterwards?

luk-kaminski:
> @luk-kaminski: In this case we are only interested in a very small subset of fields in the searches, so I would recommend operating directly on the DB instead of mapping back and forth between duplicated DTOs. Also, could you add some test cases with DB fixtures to show that the migration is working as expected? What happens with searches which cannot be migrated, do they turn out to be unreadable afterwards?

@dennisoelkers :
1. I'll try to avoid mapping to Java classes, although it may be difficult to create some proper MongoDB update queries on a complex, tree-like JSON structure.
2. Sure, I will add tests with fixtures.
3. I haven't made any decisions what to do with impossible-to-migrate-documents, I've just left TODO comment in this draft PR not to forget about it:
`//TODO: while it seems very unlikely that we will face filters that cannot be converted to streams... what to do with it, anyway?`

PR #12503, opened 2022-04-21
Remove  z index from QueryValidation, Add useIsKeyHeld hook

maxiadlovskii:
## Description
In QueryValidation we have a fixed z-index. Due to that, the user can't see the popover when we use this component in another popover. Removing the z-index fixes this issue

`useIsKeyHeld` hook helps to detect if some key is Held. f.e. 
`const isShiftHeld = useIsKeyHeld('Shift')`

To work with this feature in dev mode you have to change the setting of your Graylog server and add a feature flag in the env variable `GRAYLOG_FEATURE_SEARCH_FILTER=on`

Screenshot 2022-04-27 at 16 07 44

## Motivation and Context
We reuse this component in enterprise to build search filter features. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/util/URLUtils.ts
+++ b/graylog2-web-interface/src/util/URLUtils.ts
@@ -110,10 +110,6 @@ const URLUtils = {
 
     return acceptedProtocols.includes(url.protocol);
   },
-  urlQueryFromObject(object) {
-    return Object.entries(object)
-      .reduce((res, [key, value], i) => `${res}${i === 0 ? '' : '&'}${key}=${value}`, '');
-  },
 };
 
 export default URLUtils;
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/util/URLUtils.ts
```diff
@@ -110,6 +110,10 @@ const URLUtils = {
 
     return acceptedProtocols.includes(url.protocol);
   },
+  urlQueryFromObject(object) {
+    return Object.entries(object)
+      .reduce((res, [key, value], i) => `${res}${i === 0 ? '' : '&'}${key}=${value}`, '');
+  },
```
linuspahl, line 116: We probably do not need to create a new function for this, instead you can use the `PaginatedURL` function.

linuspahl (commented):
One thing I forgot to mention.

linuspahl (approved):
Looks good to me!

PR #12514, opened 2022-04-25
Map query tokens to subqueries

todvora:
The lucene query parser doesn't provide any connection between parsed (sub)queries and tokens used. This PR adds such a connection, which is later used to map key and value tokens, used for validation highlighting in queries. 

## Motivation and Context
Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/3491

## How Has This Been Tested?
It's refactoring, relies on existing tests covering the functionality.

For the query `timestamp:$param$ OR $param$ AND something` the assignments are:

```
Query timestamp:param(TermQuery) contains following tokens: timestamp,:,$param$,OR
Query _default_:param(TermQuery) contains following tokens: $param$,AND
Query _default_:something(TermQuery) contains following tokens: something
```
and after substitution (param=aaaa):

```
Query timestamp:aaaa(TermQuery) contains following tokens: timestamp,:,aaaa,OR
Query _default_:aaaa(TermQuery) contains following tokens: aaaa,AND
Query _default_:something(TermQuery) contains following tokens: something,
```

Note that due to the (not ideal) integration with existing lucene query parser, the tokens contain also AND/OR operators. This isn't problematic now for our usage, just wanted to mention that it's known and expected now.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
@@ -16,12 +16,25 @@
  */
 package org.graylog.plugins.views.search.validation;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.queryparser.classic.ParseException;
 import org.apache.lucene.queryparser.classic.QueryParser;
+import org.apache.lucene.search.Query;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class TokenCollectingQueryParser extends QueryParser {
+
+    private final Set<ImmutableToken> processedTokens = new HashSet<>();
+    private final Map<Query, Collection<ImmutableToken>> tokenLookup = new LinkedHashMap<>();
 
     public TokenCollectingQueryParser(String defaultFieldName, Analyzer analyzer) {
         super(new CollectingQueryParserTokenManager(defaultFieldName, analyzer));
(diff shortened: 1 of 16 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
```diff
@@ -31,4 +44,74 @@ public TokenCollectingQueryParser(String defaultFieldName, Analyzer analyzer) {
...
+
+    /**
+     * This method persists all newly discovered tokens that may be referenced in the query to a lookup for
+     * later processing. This is the only place that binds query and its tokens together.
+     *
+     * This would be also good place to collect all subqueries if needed for any feature.
+     */
+    private Query saveQueryLookupTokens(Query query) {
+        if (!tokenLookup.containsKey(query)) {
+            final List<ImmutableToken> tokens = ((CollectingQueryParserTokenManager) super.token_source).getTokens();
```
luk-kaminski, line 110: Instead of this casting to CollectingQueryParserTokenManager here while using token_source field from the superclass, I'd save the instance of CollectingQueryParserTokenManager as a field in this class, during construction, in line 40.

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
```diff
@@ -31,4 +44,74 @@ public TokenCollectingQueryParser(String defaultFieldName, Analyzer analyzer) {
...
+    }
+
+    /**
+     * This method persists all newly discovered tokens that may be referenced in the query to a lookup for
+     * later processing. This is the only place that binds query and its tokens together.
+     *
+     * This would be also good place to collect all subqueries if needed for any feature.
+     */
+    private Query saveQueryLookupTokens(Query query) {
+        if (!tokenLookup.containsKey(query)) {
```
luk-kaminski, line 109: I am just wondering about identity of the queries in this place...
If we have two queries with the same query string, and they are equal according to equals() method, should we really avoid processing the second one?
You know the details better, but there is a chance that IdentityHashMap would be a better implementation of tokenLookup...

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
```diff
@@ -31,4 +44,74 @@ public TokenCollectingQueryParser(String defaultFieldName, Analyzer analyzer) {
...
+
+    @Override
+    protected Query newTermQuery(Term term, float boost) {
+        return saveQueryLookupTokens(super.newTermQuery(term, boost));
+    }
+
+    @Override
+    protected Query newRangeQuery(String field, String part1, String part2, boolean startInclusive, boolean endInclusive) {
+        return saveQueryLookupTokens(super.newRangeQuery(field, part1, part2, startInclusive, endInclusive));
+    }
```
luk-kaminski, line 100: If Lucene developers decided to refactor their code, then we are doomed... ;)

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
```diff
@@ -16,13 +16,26 @@
...
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class TokenCollectingQueryParser extends QueryParser {
```
luk-kaminski, line 34: With all the changes and bigger responsibility, doesn't this class deserve a unit test?

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
todvora: good idea, implemented, thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
todvora: Good catch! Had to adapt the logic a bit and extend the lucene query parser so it correctly handles identity of queries in boolean queries, but it all seems to work now. Thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/TokenCollectingQueryParser.java
todvora: I think the best approach is to maintain and extend the LuceneQueryParser. This defines clear input and output of the operation. The TokenCollectingQueryParser is just an internal wrapper where I don't see any meaningful unit tests. But if you have any idea for a good test, I'll be happy to implement that.

todvora: I see the technical possibility to unit test the class in such a way. I am rather questioning if such a test makes sense and would be useful. The newSmth methods are useful only in connection with the parser and actual query parsing itself. And from there, it's easy to observe the behaviour, without exactly inspecting internals. We can test those methods with a bit of mocking, but it seems like a test only to have a test.

The LuceneQueryParser is more of an integration test in a unit-test form. The LuceneQueryParser class logic is rather trivial and it mainly integrates the parser and visitor. But in the same time, it tests all those components together, with clean boundaries, inputs and outputs. 

If we manage to write a readable test for the TokenCollectingQueryParser.java, that would answer the question "what is this class purpose", then I am for it. I just worry that it will become a mess of mocks and map.get calls to verify something internal. 

I am lacking the imagination here - so please, if you have the time and will, could you add (in this branch) a test for that class, that you believe would be good? Thank you!

luk-kaminski: Tomas, I completely understand your concerns. I've pushed a unit test and we can decide if it is ok or not.

luk-kaminski (approved):

luk-kaminski:
Geat job, @todvora , thank you! :)

todvora:
Thank you for your contribution @luk-kaminski !

PR #12516, opened 2022-04-25
Always clear query validation explanations.

linuspahl:
_Please note, this PR needs to be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/3534_

## Motivation and Context

Before this change it was possible that old validation explanations were not being removed. This happen because we used the error message as a key for the list items. Because of recent changes this attribute is no longer a uniq one.
We are now using the index as a key, while this can be problematic in some cases, it is not a problem in this case because the list items have no state.

You can reproduce the bug by:
1. paste the following query: `GET or timestamp:"test test test"`
2. open the query validation explanation
3. remove `GET or`

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3534


dennisoelkers (approved):
Looks good, works fine!

PR #11568, opened 2021-11-03
Generate backend API types for frontend from maven run.

dennisoelkers:
## Motivation and Context

This PR is adding functionality to create TypeScript typings for the backend REST API, consumable by the frontend. These typings are generated through the maven build, during the `process-classes` phase.

Typings are accurate currently only for the simpler request/response DTOs. They break if:

  - dynamic polymorphism/subclassing is used
  - a plain `Response` object is returned from a REST method

The intention of this PR is to:
  - improve type-safety for usage of the REST API in the frontend
  - speed up creation of new REST API endpoints due to having to write less boilerplate in the frontend
  - synchronize changes in the backend to frontend usage and being able to detect breakages introduced in the backend

there is no intention to replace all existing REST API usages, especially in places where the generated typings are not ideal. It should be used in places where it makes sense and where we can see a reasonable benefit.

Future plans for this include:

  - adapt this to OpenAPI 3
  - improve typing by streamlining complicated parts of the backend API (e.g. replace usages of `Response` with better types)
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #12525, opened 2022-04-26
Generate backend API types for frontend from maven run.

dennisoelkers:
## Motivation and Context

This PR is adding functionality to create TypeScript typings for the backend REST API, consumable by the frontend. These typings are generated through the maven build, during the `process-classes` phase.

Typings are accurate currently only for the simpler request/response DTOs. They break if:

  - dynamic polymorphism/subclassing is used
  - a plain `Response` object is returned from a REST method

The intention of this PR is to:
  - improve type-safety for usage of the REST API in the frontend
  - speed up creation of new REST API endpoints due to having to write less boilerplate in the frontend
  - synchronize changes in the backend to frontend usage and being able to detect breakages introduced in the backend

There is no intention to replace all existing REST API usages, especially in places where the generated typings are not ideal. It should be used in places where it makes sense and where we can see a reasonable benefit.

Future plans for this include:

  - adapt this to OpenAPI 3
  - improve typing by streamlining complicated parts of the backend API (e.g. replace usages of `Response` with better types)

This PR is different to its predecessor (#11568) which was reverted shortly after (#12518) in the following ways:

The previous PR broke including web interface assets because their compilation happened at a later maven stage than before. `yarn build` was executed at the `package` phase. Resources which are supposed to be included by the default goal need to be created before the `process-resources` phase. Unfortunately, we cannot generate backend APIs before the `compile` phase is completed. Therefore we need to include a second invocation of the `maven-resources-plugin` which is performed at the `compile` phase. Now that there are multiple goals defined for the `compile` phase, their declaration order is defining the execution order, with the default `compile` goal happening first. Because all of the related front end tasks are defined in a single profile, we can order their declaration in a way that is explicit enough.

## Types of changes

## Checklist:


thll (approved):
I only looked at the `pom.xml` changes. The final version of the declaration of goals and assignment to phases looks good to me.

Also tested that the enterprise server built with `mvn package` starts up successfully.

PR #12526, opened 2022-04-26
Fixing usage of multiple reference dates for human readable duration.

dennisoelkers:
## Motivation and Context

This change is fixing a subtle issue when creating a human-readable duration for a time range through the `readableRange` function.  Prior to this change, that function used two different reference dates when calculating the length of a duration by calling `moment()` repeatedly.  One call happened when declaring the `dateAgo` variable, the other when calculating the diff between current time (which is another `moment()` call) and `dateAgo`. When the first and the second call are across a secound boundary, this could lead to the calculation being slightly off.  E.g. when the first call is happening at `12:00:00.999` and the second one at `12:05:01.000`, the function would return `5 minutes 1 second ago`. This also results in test failures happening for timings like that where the test expects the literal string `5 minutes ago` while `5 minutes 1 second ago` is generated.

This change is now calling `moment()` only once and basing all calculations on it, preventing these kinds of overflows.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/logic/queries/TimeRangeToString.ts
+++ b/graylog2-web-interface/src/views/logic/queries/TimeRangeToString.ts
@@ -29,8 +29,9 @@ export const readableRange = (timerange: TimeRange, fieldName: 'range' | 'from' 
     return placeholder;
   }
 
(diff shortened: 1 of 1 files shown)


linuspahl (approved):
Looks good to me!

PR #12537, opened 2022-04-27
Version check only fetches necessary parts of the response.

luk-kaminski:
## Description
Version check only fetches necessary parts of the response, not the whole JSON. Unit tests for version check have been added.

## Motivation and Context
NodeAdapterES7 had no unit tests. I've added them.
NodeAdapterES7 used to grab the whole JSON, despite needing only 1-2 fields. "filter_path" request parameter has been used to limit the size of response.

## How Has This Been Tested?
Tested on Opensearch and Elasticsearch, with debugger and looking in the startup logs for messages like this one:
`Connected to (Elastic/Open)Search version <Elasticsearch:7.10.2>`

## Types of changes

## Checklist:

### DIFF
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/NodeAdapterES7.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/NodeAdapterES7.java
@@ -34,15 +34,19 @@ public class NodeAdapterES7 implements NodeAdapter {
         this.jsonApi = new PlainJsonApi(objectMapper, client);
     }
 
+    NodeAdapterES7(final PlainJsonApi jsonApi) {
+        this.jsonApi = jsonApi;
+    }
+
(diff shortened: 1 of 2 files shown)


todvora:
Looks good, thanks! Just thinking if we should also modernize our NodeIT and migrate it to the container matrix tests structure. What do you think?

luk-kaminski:
> Looks good, thanks! Just thinking if we should also modernize our NodeIT and migrate it to the container matrix tests structure. What do you think?

So far we seemed to use ContainerMatrixTest only inside "full-backend-tests" and "enterprise-full-backend-tests" modules. I'm fine migrating other tests as well, but I'd prefer to do it in another PR and have a team discussion first about the approach - if you agree with it.

todvora:
Sure, let's ask @janheise what would be the best approach or generally if it makes any sense :+1:

todvora (approved):
Looks good, thank you!

Issue #6118, opened 2019-07-10
housekeeping for `system.messages` in MongoDB


jalogisch:
## Context
Graylog does not provide the option or the ability to make some kind of house keeping on the mongoDB collection that is filled continuously. That might end in a huge mongoDB with an enormous `system.messages` collection that is of nearly no use. 

We should add one configuration flag (online available?) that holds the number of records or the time the internal logfile in the database should not go over. 

This will prevent Graylog server issues based on enormous MongoDB Databases.

## Your Environment

* Graylog Version: 3.0

[Z#2485]

jalogisch:
This can be used to regular clean up the collection to keep only a month of data.

```
db.getCollection('system_messages').deleteMany({
    "timestamp":{
            $lt: new Date(new Date().setMonth(new Date().getMonth()-1))
    }
})
```

rubinlinux:
The index_failures collection is a capped collection at 50 meg. Maybe system_messages should be the same? When mongo runs the disk out of space it causes a lot of trouble.

boosty:
@patrickmann I assigned you, since you are already working on it.

PR #12540, opened 2022-04-27
Limit size of system_messages collection

patrickmann:
Mongo DB collection `system_messages` grows indefinitely. There is little point in storing years worth of old system messages. 
We now cap the size of collection, just as we do for e.g. `index_failures` collection.

**Note:** you cannot delete documents from a capped collection. Users will no longer be able trim the collection by deleting a subset of documents.

Resolves Graylog2/graylog2-server#6118

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -25,5 +25,7 @@ The following API endpoints have been removed in 4.4.
 
 ## Behaviour Changes
 
-- The system messages collection in MongoDB will be created as a 50MB capped collection going forward.
-  This happens at creation, so existing system messages collections remain unconstrained.
+- The `system_messages` collection in MongoDB will be created as a 50MB capped collection going forward.
+  This happens at creation, so existing `system_messages` collections remain unconstrained.
+<br>You can manually convert your existing collection to a capped collection by following 
+these [instructions](https://www.mongodb.com/docs/manual/core/capped-collections/#convert-a-collection-to-capped).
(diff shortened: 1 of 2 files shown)


bernd:
> **Note:** you cannot delete documents from a capped collection. Users will no longer be able trim the collection by deleting a subset of documents.

@patrickmann This needs an entry in the `UPGRADING.rst` to make sure users that upgrade are aware of it. Can you please ann an entry?

What's the reasoning behind using 5 MB for the capped collection? The `index_failures` collection is using 50 MB. I think 5 MB is quite small. What do you think?

patrickmann:
@bernd Bumped the cap to 50MB and added an entry in `UPGRADING.md`

thll (changes requested):
The code looks good and works, but `UPGRADING.md` is still for version `4.3`. To make sure that your changes don't get lost, we should wait for #12570 to be merged and add your snippet then.

thll (approved):

patrickmann:
@boosty @bernd Only new customers will benefit from this (unless somebody manually drops the system messages collection so that it is recreated). I went with that approach, since other internal collections are also statically capped.
Is it worth doing this more flexibly via a scheduled job that periodically trims the collection?

boosty:
@patrickmann How about documenting how to convert an existing collection to a capped one?
https://www.mongodb.com/docs/manual/core/capped-collections/#convert-a-collection-to-capped

PR #12545, opened 2022-04-27
Using correct field for pie chart legends.

dennisoelkers:
**Note:** This should be backported to `4.3`.

## Motivation and Context

This PR is fixing an issue where the field name which is used for the legend of a pie chart is taken from the wrong pivot. It happens when there is both at least one row and at least one column pivot defined for a pie chart. In that case, the field name is taken from the row pivot instead of the column pivot. 

This PR is now making the field mapping strategy in the `PlotLegend` component overridable, defaulting to the one which matches the chart data for most visualizations and use the mirrored one for pie charts.

Fixes #12501.

## Types of changes

## Checklist:


linuspahl (approved):

PR #12565, opened 2022-04-28
Introduce a JsonSubTypePropertyDefaultValue annotation

bernd:
The annotation can be used for JsonSubTypes use cases to provide a
default type value for documents that are missing the field.

See the javadoc of the JsonSubTypePropertyDefaultValue annotation for
more details.


thll (approved):
Thanks a lot for this! 🙏

PR #12571, opened 2022-04-29
Replacing usage of views `Select` with common `Select` component.

linuspahl:
## Motivation and Context

Usually we use the `components/common/Select` component every time we need to implement a select. At some point in the past we created a separate `Select` component (`views/components/Select`). Over the time they became so similar, that we are now able to replace the views `Select` component with the common `Select` component.

The main benefit is we only need to maintain one `Select` component.

## Types of changes


dennisoelkers (approved):
Good work!

PR #12600, opened 2022-05-04
Feature/extend search page customization

mikedklein:
## Description

This PR addresses multiple quality of life improvements for the Search component and related systems. It also includes a refactor of the current `SearchPageLayoutContext` components to be renamed to `SearchPagePreferencesContext` to prevent confusion about what these two contexts and their providers are doing.

- Addition of a new SearchPageLayoutContext (along with provider and hook) that allows for control of how the underlying `Search` component is displayed. For now, it only includes configuration options for the `SideBar` and `ViewActionMenu`. However, it has been designed to be an extensible configuration object that can be modified in the future.
- Updated subsequent tests now that the `useSearchPageLayout` hook needs to be used within the provider.
- Updated the `Search` component to hide the sidebar when it is passed the `SideBar.isHidden` state from the provider hook.
- Updated the `ViewActionsMenu` to use the new hook to hide UI elements based on a state enum that is set by the provider.
- Updated all references to s"Analyst Tools" to "Security" in terms of routing and navigation items.

## Motivation and Context

The changes were requested to help facilitate the functionality of the Enterprise product dashboard pages where there are canned dashboards that need certain user interactions removed.

## How Has This Been Tested?
Unit tests and visual tests have been conducted.

## Types of changes

## Checklist:

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3537


graylog2-web-interface/src/views/components/contexts/SearchPageLayoutContext.tsx
```diff
@@ -14,19 +14,21 @@
...
+  BLANK = 'BLANK'
+}
 
-export type SearchPageLayout = {
-  config: SearchPageLayoutConfig,
-  actions: { toggleSidebarPinning: () => void },
-};
+export type LayoutState = {
+  sidebar: { isShown: boolean }
+  viewActionsLayoutOptions: ViewActionsLayoutOptions
```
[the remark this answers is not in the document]
linuspahl: Just an idea, but maybe we can unify the structure of these settings, for example like this (they could be `true` by default): 
```
sidebar: { isShown: boolean },
dashboardActions: {
  save: { isShown: boolean },
  saveAs: { isShown: boolean },
  share: { isShown: boolean },
  actionsDropdown: { isShown: boolean },
},
```

This would make it easier to create a helper component for this logic, similar to `IfPermitted`. The usage could look like this: `<IfLayoutOption options=['dashboardActions.saveAs']>...</IfLayoutOption>`.

Another advantage, it would make it very clear what effect each setting has. One disadvantage is of course that we have to extend it over time and the state itself is larger compared with the current approach.

Alternatively we could use an enum for every option, in this case the usage of the helper component could look like this: `<IfLayoutOption options=['sidebar.COMPLETE']>...</IfLayoutOption>` and `<IfLayoutOption options=['dashboardActions.COMPLETE']>...</IfLayoutOption>`.

I think in every case there is a longterm benefit in unifying the structure of these settings.

graylog2-web-interface/src/views/components/ViewActionsMenu.tsx
```diff
@@ -65,23 +68,35 @@ const ViewActionsMenu = ({ view, isNewView, metadata }) => {
...
+  const showShareButton = viewActionsLayoutOptions === ViewActionsLayoutOptions.FULL_MENU;
+  const showDropDownButton = viewActionsLayoutOptions === ViewActionsLayoutOptions.FULL_MENU;
+
   return (
     <ButtonGroup>
+      {showSaveButton && (
       <Button onClick={() => onSaveView(view)}
               disabled={isNewView || hasUndeclaredParameters || !allowedToEdit}
-              data-testid="dashboard-save-button">
+              title="Save Button">
```
linuspahl, line 81: What du you think about a title like `Save dashboard` and `Save as new dashboard` instead of `Save Button` and `Save As Button`? The title of the component is `ViewActionsMenu`, but we only use it for dashboards.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutContext.tsx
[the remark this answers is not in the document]
simonychuang: Just gonna jump in here and share my two cents too. I can definitely see the upsides to both of these approaches, and in my past experience I have been using the pattern that Linus is describing. However, I can see the points Mike is trying to make here, and I think it would be more advantageous for the component to decide what to render/ not render instead of a state describing what should be rendered. I think the example Mike is giving here is that in the future if another component were to consume that context and the fields we defined right now are not really relevant, it would be better if the component just takes in something like `FULL_MENU` and then decide how to interpret that in a way that's specific to that component. Otherwise we might have to expand the `ViewActionsLayoutOptions` and it could get really big.

Happy to hear what others think about this too.

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export type SearchPreferencesLayout = {
+  config: SearchPagePreferencesConfig,
+  actions: { toggleSidebarPinning: () => void },
+};
+
+const SearchPagePreferencesContext = React.createContext<SearchPreferencesLayout | undefined>(undefined);
+
+export default SearchPagePreferencesContext;
```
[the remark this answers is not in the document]
mikedklein: Fixed.

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesProvider.tsx
```diff
@@ -0,0 +1,59 @@
...
+  <SearchPagePreferencesState>
+    {({ getPreferencesState, setPreferencesState }) => (
+      <SearchPagePreferencesProvider getPreferencesState={getPreferencesState} setPreferencesState={setPreferencesState}>
+        {children}
+      </SearchPagePreferencesProvider>
+    )}
+  </SearchPagePreferencesState>
+);
+
+export default SearchPagePreferencesStateProvider;
```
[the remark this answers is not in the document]
mikedklein: Fixed.

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesState.tsx
```diff
@@ -28,9 +28,9 @@ import { PreferencesActions } from 'stores/users/PreferencesStore';
 import type User from 'logic/users/User';
 
 type Props = {
-  children: (layoutConsumer: {
-    setLayoutState: (stateKey: string, value: boolean) => void,
-    getLayoutState: (stateKey: string, defaultValue: boolean) => boolean,
+  children: (preferencesConsumer: {
+    setPreferencesState: (stateKey: string, value: boolean) => void,
+    getPreferencesState: (stateKey: string, defaultValue: boolean) => boolean,
```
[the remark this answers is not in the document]
mikedklein: Fixed.

graylog2-web-interface/src/views/components/ViewActionsMenu.tsx
mikedklein: Fixed.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutContext.tsx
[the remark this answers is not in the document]
dennisoelkers: Thanks for everyone participating in this discussion, I appreciate the love for the detail and your thoughts about how to make this code sustainable. There are two potential sources of complexity here:

1. Having a complex state (by defining more fine-grained options on how and which elements are displayed) and simpler logic in the components
2. Having a simpler state (by selectining what is essentially a "preset") and more complex logic in the components (by testing which preset is selected and configuring elements according to it)

Both are potential sources for growth. For 1., the number of options could grow, for 2. it would be the number of presets and - in consequence - the logic in the components. I think there could be a middle way though, that leaves all of the pros of both approaches on the table and gets rid of some of the cons:

What we could do is define a larger state as proposed by @linuspahl, but define named default presets like this:

```
export const FULL_MENU = {
  save: { isShown: true },
  saveAs: { isShown: true },
  share: { isShown: true },
  actionsDropdown: { isShown: true },
};

export const SAVE_COPY = { ... };
```
for them which can be imported and used to have a consistent preset. This allows us to have simple logic in the components _and_ makes selecting/defaulting to named presets easy.

Btw: enums in TypeScript are rather clumsy and have a couple of disadvantages. I would always resort to plain constant maps and use union types instead.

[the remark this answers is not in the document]
mikedklein: @dennisoelkers I think I have this way you requested if you could take a look and see if this is how you thought it would be implemented.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
```diff
@@ -14,46 +14,36 @@
...
+  viewActionsLayoutOptions: FULL_MENU,
+} as LayoutState;
 
-    return ({
-      config,
-      actions,
-    });
-  }, [getLayoutState, setLayoutState]);
+function SearchPageLayoutProvider({ children, providerOverrides = defaultState }: SearchPageLayoutProviderProps) {
+  const [state] = React.useState<LayoutState>(providerOverrides);
```
dennisoelkers, line 34: I am not sure what this state variable is for, is it really needed?

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
```diff
@@ -14,46 +14,36 @@
...
+
+  const value = React.useMemo(() => ({ ...state }), [state]);
```
dennisoelkers, line 36: Is this to copy the state before providing it through the context? This is only a shallow copy, so if a consumer manipulates e.g. `state.viewActionsLayoutOptions`, it would still change the original object.

Instead of this, I would suggest solving this through typing the context value as `readonly`.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutContext.tsx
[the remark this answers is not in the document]
dennisoelkers: @mikedklein: Yes, it looks good and is what had in mind. The only (minor) suggestion for improvement is that maybe `viewActions` could be better for the key, as it is shorter and - in the context of a `SearchPageLayoutContext` - eliminates a bit of redundancy.

graylog2-web-interface/src/views/components/ViewActionsMenu.tsx
```diff
@@ -67,21 +81,28 @@ const ViewActionsMenu = ({ view, isNewView, metadata }) => {
 
   return (
     <ButtonGroup>
+      {showSaveButton && (
       <Button onClick={() => onSaveView(view)}
               disabled={isNewView || hasUndeclaredParameters || !allowedToEdit}
-              data-testid="dashboard-save-button">
```
dennisoelkers, line 72: Please be aware that changes like this could also affect our [e2e-tests](https://github.com/Graylog2/e2e-tests). In some cases we introduced identifiers like this to be able to address elements, either in unit or in e2e tests. There are better ways to do this than test ids (like using a title), we just need to coordinate changes. In this case it was only needed for the unit tests afaics, so we are all good. I am working on integrating the e2e-tests into our regular build, so it will be more obvious in the future when the related change in the tests is missing (because the build will fail).

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesProvider.test.tsx
```diff
@@ -86,29 +86,29 @@ describe('SearchPageLayoutProvider', () => {
     return consume;
   };
 
-  it('provides default search page layout with empty preference store', () => {
+  it('provides default search page preference state with empty preference store', () => {
     const consume = renderSUT();
 
     expect(consume.mock.calls[0][0]?.config.sidebar.isPinned).toEqual(false);
```
dennisoelkers, line 92: Not related to your code, but maybe you have some time to fix it: These could be replaced by something like this:
```suggestion
    expect(consume).toHaveBeenCalledWith(expect.objectContaining({ config: expect.objectContaining({ sidebar: expect.objectContaining({ isPinned: false }) }) }));
```
This way we do not have to make assumptions about the mock object.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
```diff
@@ -14,46 +14,36 @@
...
+  sidebar: { isShown: true },
+  viewActionsLayoutOptions: FULL_MENU,
+} as LayoutState;
 
-    return ({
-      config,
-      actions,
-    });
-  }, [getLayoutState, setLayoutState]);
+function SearchPageLayoutProvider({ children, providerOverrides = defaultState }: SearchPageLayoutProviderProps) {
```
dennisoelkers, line 33: Do we actually need this default provider at all? From what I am seeing, we are passing the prop for the provider overrides from the `Search` component or use `defaultState` if that prop is not set. Instead, we could just use `defaultState` as the default value for the context (instead of `undefined`) and leave out the provider in the normal case. If we want to use the `Search` component with a different layout, we can just wrap it in a provider block like this:
```
<SearchPageLayoutContext.Provider value={...}>
  <Search />
</SearchPageLayoutContext.Provider>
```

dennisoelkers (changes requested):
Looks good so far! Added a couple of suggestions inline.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
mikedklein: I had intended for it to be possibly have the ability to be updated by consumer components but I think that is over thinking it and we can it back if it is ever needed.

graylog2-web-interface/src/views/components/ViewActionsMenu.tsx
mikedklein: Got it was worried about this but ran all the tests, didn't think that the e2e tests could be using it.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
mikedklein: @dennisoelkers I went back forth on this, but ultimately I went with this because of the hook. I didn't like the possibility of a component consuming the hook outside the provider so I thought making its usage explicit was the way to go. Otherwise the hook would have to provide a fake piece of context if used outside the provider.

dennisoelkers: What would be the problem with a component consuming the hook outside the provider when the context provides reasonable defaults?

danotorrey (approved):
Tests successfully for me

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutContext.tsx
```diff
@@ -14,19 +14,52 @@
...
+
 import * as React from 'react';
 
-type SearchPageLayoutConfig = {
-  sidebar: {
-    isPinned: boolean,
-  },
+import { singleton } from 'logic/singleton';
+
+export interface ViewActions {
```
dennisoelkers, line 22: One note for the future: If we want to make this interface extensible through plugins, we should move this interface to `src/views/types.ts`. That way it would be decoupled from the component, making it easier to use declaration merging through module augmentation.

For now we can leave it in here though, and do this only when we encounter the actual need for it be extended through plugins.

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesState.tsx
```diff
@@ -28,9 +28,9 @@ import { PreferencesActions } from 'stores/users/PreferencesStore';
 import type User from 'logic/users/User';
 
 type Props = {
-  children: (layoutConsumer: {
-    setLayoutState: (stateKey: string, value: boolean) => void,
-    getLayoutState: (stateKey: string, defaultValue: boolean) => boolean,
+  children: (preferencesConsumer: {
+    setPreferences: (stateKey: string, value: boolean) => void,
```
dennisoelkers, line 32: I think we should name this `setPreference`, because we are sitting a single preference and not all of them. Same for `getPreference(s)`.

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
```diff
@@ -14,46 +14,34 @@
...
-const SearchPageLayoutProvider = ({ getLayoutState, setLayoutState, children }: Props) => {
-  const searchPageLayoutContextValue = useMemo(() => {
-    const config = {
-      sidebar: { isPinned: getLayoutState('sidebarIsPinned', false) },
-    };
-    const actions = { toggleSidebarPinning: () => setLayoutState('sidebarIsPinned', !config.sidebar.isPinned) };
+export const defaultState = {
+  sidebar: { isShown: true },
+  viewActions: FULL_MENU,
+} as LayoutState;
```
dennisoelkers, line 31: I think we can just leave out the cast or use `as const` instead?

graylog2-web-interface/src/views/components/contexts/SearchPageLayoutProvider.tsx
```diff
@@ -14,46 +14,34 @@
...
 
-    return ({
-      config,
-      actions,
-    });
-  }, [getLayoutState, setLayoutState]);
+function SearchPageLayoutProvider({ children, providerOverrides = defaultState }: SearchPageLayoutProviderProps) {
```
dennisoelkers, line 33: Do we actually need this provider? If we set an appropriate default value for the context it becomes unnecessary.

dennisoelkers (commented):
Looks good so far! Added a couple of remaining questions inline for a last round of feedback.

dennisoelkers (approved):
Looks great now! Thanks for the good work and the patience!

PR #12604, opened 2022-05-05
Fixing swagger schema generation for enums.

dennisoelkers:
## Motivation and Context

Prior to this PR, generating the swagger schema failed for enums under certain conditions, because the enum values were provided twice in the generated schema. This results in a map entry conflict during JSON serialization because of the usage of `ImmutableMap.Builder`, which does not allow keys to be set twice.

This change is now making sure that defining the `enum` field for a type schema twice does not lead to an exception, through the use of a plain `HashMap` when providing the JSON value.

Fixes Graylog2/graylog-plugin-enterprise#3529.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
@@ -17,23 +17,10 @@
 package org.graylog2.shared.rest.documentation.generator;
 
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.fasterxml.jackson.databind.JavaType;
-import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonAnyFormatVisitor;
-import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonArrayFormatVisitor;
-import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonBooleanFormatVisitor;
-import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonIntegerFormatVisitor;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog2/shared/rest/documentation/generator/Generator.java
```diff
@@ -827,27 +814,28 @@ public void setDefaultValue(String defaultValue) {
...
-            ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder()
-                    .put("name", name)
-                    .put("description", description)
-                    .put("required", isRequired)
-                    .put("paramType", getKind());
-
-            if (defaultValue != null) {
-                builder = builder.put("defaultValue", defaultValue);
-            }
+            final HashMap<String, Object> result = new HashMap<String, Object>() {{
```
luk-kaminski, line 817: Do we need to create an anonymous subclass of HashMap with custom initialization block inside ? Can't we just invoke all the put methods on the "result" instance, after it gets created?

dennisoelkers: No, both would work similar. It feels a little bit cleaner this way, not having to repeat the variable all over. What would be the benefit of using an instance / the downside of using an anonymous subclass?

luk-kaminski: It is a personal opinion, but for me initializer blocks are much less readable. 
This code also creates additional anonymous class underneath, which does not seem to be needed.

dennisoelkers: Interesting, for me it is the other way around. The initializer block makes it clear to me that initialization of the map is scoped to it. As long as there are no relevant technical issues around it, I would keep it that way.

PR #12528, opened 2022-04-26
Memorize values of every react context.

linuspahl:
## Motivation and Context

With this PR we are fixing all `react/jsx-no-constructed-context-values` linter hints by memorizing the value of every context.

The [linter rule page](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md) describes the benefits of this refactoring pretty well:

> React Context, and all its child nodes and Consumers are rerendered whenever the value prop changes. Because each Javascript object carries its own identity, things like object expressions (`{ foo: 'bar' }`) or function expressions get a new identity on every run through the component. This makes the context think it has gotten a new object and can cause needless rerenders and unintended consequences.
> 
> This can be a pretty large performance hit because not only will it cause the context providers and consumers to rerender with all the elements in its subtree, the processing for the tree scan react does to render the provider and find consumers is also wasted.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #12610, opened 2022-05-05
Avoiding calling `useMemo` outside of component.

dennisoelkers:
## Motivation and Context

This PR is turning the `_column` function into a separate component.  Prior to it, the `_column` function called `useMemo` after #12528 was introduced, which lead to it being called conditionally resulting in different numbers of calls between rerenders, leading to an exception being thrown.

## Types of changes

## Checklist:


linuspahl (approved):
Fixes the described problem, thank you. I had a look at the other changes in https://github.com/Graylog2/graylog2-server/pull/12528 and did not found a similar problem.

PR #12647, opened 2022-05-12
Feature/extend search page customization (#12600)

mikedklein:
# Backport of #12600 

This change is needed to allow for the customization of the `Search` components' layout.
The first iteration allows hiding the sideNav and defining which viewActionMenu buttons are displayed.

* Adds layout configuration context for the `Search` component.
* This refactors the `SearchPageConfigContext` component to SearchPageLayoutContext
and all the SearchPageLayout domain components to SearchPagePreferences. This was done to prevent confusion about the domains these components are acting upon.
* This refactors the `SearchPageLayoutContext` into multiple components
to be more in line with our current approach to managing contexts and hooks.
* Renames "Analyst Tools" back to "Security" for routing purposes.
* Wrap both new contexts in `singleton()`
* Match file name to default export in
* Rename `setPreferencesState` and `getPreferencesState` to
  `setPreference` and `getPreference`
* Rename button titles to be more specific in `ViewActionsMenu.tsx`
        - 'Save Button' -> 'Save dashboard'
        - 'Save As Button' -> 'Save as new dashboard'
* Removed the default provider and instead relies on the hook to
   provide logical defaults for the layout context.
* Updates some tests to use more explicit checks.
* Updates naming conventions to be easier to read.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3579

### DIFF
--- a/graylog2-web-interface/src/hooks/useSearchPageLayout.ts
+++ b/graylog2-web-interface/src/hooks/useSearchPageLayout.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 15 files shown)


mikedklein:
I think we can ignore the linting error in this case since the rule was introduced after the release was cut so and the fixes are not in 4.3.

kingzacko1 (approved):
Functionality LGTM

dennisoelkers (approved):

PR #12668, opened 2022-05-16
Escape field value suggestions before adding them to query.

### LINKED ISSUE #12667: Field value suggestions in query input can lead to invalid query
## Expected Behavior

When a user selects a field value suggestion, the resulting query should be valid.

## Current Behavior

Depending on the value of the suggestion, the query could become invalid. As an example, field values containing windows pathnames could result in invalid queries.

1. Generate documents containing windows pathnames
2. Start a new search
3. Type a query triggering field value suggestions with e.g. windows pathnames
4. Select a query

image

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

**Note:** This needs a backport to `4.3`.

Prior to this change, field value suggestions (existing field values which are suggested in the query input when a user is wring a query) were not escaped when the user selects one to be added to the query string. This could lead to invalid queries being produced (e.g. when the field contains a windows pathname).

This change is now adding escaping to those suggestions being produced by the field value completion. A more general solution in the `SearchBarAutoCompletions` class could be implemented at a later time, but as this change aims towards being integrated into the upcoming release, it tries to fix this issue in the least intrusive way.

Fixes #12667.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.test.ts
+++ b/graylog2-web-interface/src/views/components/searchbar/completions/FieldValueCompletion.test.ts
@@ -17,23 +17,26 @@
 import { StoreMock as MockStore } from 'helpers/mocking';
 import asMock from 'helpers/mocking/AsMock';
 import FieldTypeMapping from 'views/logic/fieldtypes/FieldTypeMapping';
-import FieldType from 'views/logic/fieldtypes/FieldType';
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Fixes the described problem, thanks for taking care of it.

PR #12683, opened 2022-05-17
Create hook to determine if a component is mounted.

linuspahl:
## Motivation and Context

With this PR we are creating a simple hook which determines if a component is mounted. We need this information in a few cases, e.g. to avoid the `Can't perform a React state update on an unmounted component.` warning.

## Types of changes


dennisoelkers (commented):
It looks good! Do we want to replace the similar pattern in `UrlWhitelistForm` with it as well?

zeeklop (approved):

dennisoelkers (approved):
Thanks for the clarification, @linuspahl! Good work!

Issue #12337, opened 2022-03-22
Encoding option for inputs


patrickmann:
## What?

Provide an option to configure inputs for UTF-16.

## Why?

GL expects UTF-8. UTF-16 input results in parsing errors. We don't state this very prominently in our docs and it has led to a number of customer inquiries in the past. If customers use UTF-16, it would be helpful to process the data without requiring them to convert it prior to sending to GL.

tellistone:
Encountered in Enterprise customer issue HS-800074155

This seems like a problematic blindspot. At present UTF-16 input will be ingested without complaint, and then cause problems throughout the platform as it behaves unpredictably with stream and pipeline rules. At no point is a warning/error displayed to user that an incorrect format is being received.

Users can work around this at present by including Logstash in their architecture to act as a converter.

kroepke:
Depending on the input definition, some will have a requirement which encoding is to be used, whereas others implicitly use UTF-8.
Since auto-detection can be very brittle, I'd suggest adding a generic override to inputs so that behavior is consistent everywhere. Note that users need to make sure they understand the consequences, especially for inputs that receive data from multiple sources.

patrickmann:
@kroepke This sentence in the [docs](https://archivedocs.graylog.org/en/latest/pages/pipelines/rules.html#data-types) confuses me:
`The six built-in types in Graylog are string (a UTF-8 string), ...`
It's the job of the codec to transform raw bytes to Java unicode strings by applying the appropriate mapping (e.g. UTF-8). Strings use unicode - there is no charset mapping.
The doc is just trying to say that we use utf-8 mapping to decode strings, right? We don't use any raw message bytes in pipeline processing, do we?

PR #12691, opened 2022-05-18
Release note for input encoding

patrickmann:
Documents Graylog2/graylog2-server#12337


bernd:
@patrickmann Are there any changes to message processing when selecting an input encoding, or is it just important to specify the correct encoding and Java takes care of the rest internally? (probably using UTF-16 internally)

patrickmann:
> Are there any changes to message processing when selecting an input encoding, or is it just important to specify the correct encoding and Java takes care of the rest internally? (probably using UTF-16 internally)

The codecs transform raw byte messages to Java Unicode strings, which we use for all further processing. Encoding should be irrelevant after the decoding stage.

boosty:
Description looks fine to me (unfortunately I can't say anything about the underlying details).

PR #12694, opened 2022-05-18
Bump okta SDK version to 8.2.1

patrickmann:
Rev the okta SDK version to latest 8.2.1.
Amazingly I did not need to fix any breaking changes.

Resolves Graylog2/graylog-plugin-enterprise#3486

See https://github.com/okta/okta-sdk-java/releases

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@
         <netty.version>4.1.60.Final</netty.version>
         <netty-tcnative-boringssl-static.version>2.0.36.Final</netty-tcnative-boringssl-static.version>
         <okhttp.version>3.14.6</okhttp.version>
-        <okta-sdk.version>3.0.1</okta-sdk.version>
+        <okta-sdk.version>8.2.1</okta-sdk.version>
         <opencsv.version>2.3</opencsv.version>
         <os-platform-finder.version>1.2.3</os-platform-finder.version>
         <pkts.version>3.0.5</pkts.version>


danotorrey:
@patrickmann Do you know if this has already been tested with Okta user management for Cloud?

patrickmann:
> Do you know if this has already been tested with Okta user management for Cloud?
@danotorrey I did some testing in Cloud, but would appreciate if you could double-check that group and user sync are working as expected.

danotorrey:
Sorry for the delay @patrickmann. I've started testing this today.

danotorrey:
What do you think about also bumping the `okta-jwt` version to `0.5.1` at the same time? This is used for the Okta login auth part, which is a separate flow from Okta user management. That is easy-enough to test at the same time.

See https://mvnrepository.com/artifact/com.okta.jwt/okta-jwt-verifier
It is defined in two places:
https://github.com/Graylog2/graylog-plugin-enterprise/blob/bb312e6015498b573707f70d3715acbf5897444a/enterprise/pom.xml#L19
https://github.com/Graylog2/graylog-plugin-cloud/blob/0cca7088db2a8c0246eac869caef0ee68874e834/server-plugin/pom.xml#L21

I am testing the user management update now...

patrickmann:
[Here](https://github.com/okta/okta-jwt-verifier-java/releases/latest) are the release notes for _okta-jwt-verifier_ 0.5.1
It's only a patch release over 0.5.0 and includes very few actual bug fixes - none are security issues. So I think we can stick with 0.5.0 for now and avoid retesting.

PR #12705, opened 2022-05-19
Make content pack creation more robust

### LINKED ISSUE #12684: (HS 8267556) Content Pack creation page fails to load entities if any fail to deserialize
If any content pack exportable entity in the DB fails to deserialize when populating the content pack creation page then the entire page will just be empty and have no entities available to export.

## Expected Behavior

The entities that are correctly deserialized should be listed for export and any failing entities should have a descriptive log message as to why they are not included on the page. 

## Current Behavior

If any one entity fails to deserialize when populating the content pack creation list of entities then no entities are displayed for exporting

## Possible Solution

Catch the exception for that failed deserialization, log it, and then continue gathering entities to display on the content pack creation page

Not easily reproducible from the Graylog UI itself. Seems to occur after GL upgrades most often.

1.  Modify an entity in your MongoDB so that it no longer can be deserialized correctly (I renamed the `title` field in a stream from `title` to `not_title`: `db.streams.updateOne({"_id": ObjectId("000000000000000000000001")}, {$rename:{"title":"not_title"}})`
2. Go to System > Content Packs > Create a Content Pack. Should have no options for exporting
image

(yes this breaks other stuff and likely wouldn't ever happen, but it's just a quick way to recreate the issue) 

## Context

Seen in at least two cases by customers and on multiple occasions by Graylog contributors. It seems like it is typically the result of an upgrade where an entity may not have been migrated 100% correctly. 

## Your Environment

My test env where I reproduced the issue: 
* Graylog Version: 4.3
* Java Version: 8
* Elasticsearch Version: OpenSearch1.3
* MongoDB Version: 4.4
* Operating System: Ubuntu
* Browser version: latest Chrome and Firefox

but note this has happened across multiple customer and contributor environments

patrickmann:
Content pack creation should be more robust in the face of serialization issues (see linked issue for details).

As a first step we simply catch and log exceptions, so content pack creation can still proceed; and it is easier to identify the offending entity.

Resolves #12684 
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3616

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/EntityWithExcerptFacade.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/EntityWithExcerptFacade.java
@@ -28,12 +28,16 @@ import org.graylog2.contentpacks.model.entities.NativeEntity;
 import org.graylog2.contentpacks.model.entities.NativeEntityDescriptor;
 import org.graylog2.contentpacks.model.entities.references.ValueReference;
 import org.graylog2.plugin.indexer.searches.timeranges.InvalidRangeParametersException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
 public interface EntityWithExcerptFacade<EntityClass, ExcerptClass> {
+    Logger LOG = LoggerFactory.getLogger(EntityWithExcerptFacade.class);
+
     /**
      * Create an exportable model of a native entity referenced by an {@link EntityDescriptor}
      * including optional constraints.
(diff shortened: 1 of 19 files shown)


graylog2-server/src/main/java/org/graylog2/contentpacks/facades/EntityWithExcerptFacade.java
```diff
@@ -91,6 +95,13 @@ default Optional<NativeEntity<EntityClass>> findExisting(Entity entity, Map<Stri
      */
     void delete(EntityClass nativeEntity);
 
+    /**
+     * Return a unique ID - primarily for troubleshooting
+     * @param nativeEntity The native entity
+     * @return unique ID of the entity
+     */
+    String id(ExcerptClass nativeEntity);
```
bernd, line 103: @patrickmann Adding the non-default `id` method is a backward-incompatible change to the interface. I am not sure if anyone is using the content pack APIs in a 3rd-party plugin, but we should at least add an entry to the `UPGRADING.md` file to let people know about it.

patrickmann: Good point - I will document that. 
I do think we need it, otherwise the error log messages would be pretty useless in helping to pinpoint the cause of a problem.

patrickmann:
Decided against doing this. See related issue for reasoning.

PR #12711, opened 2022-05-20
Updating security policy to remove need for maintenance.

dennisoelkers:
This change is solely updating the security policy in order to remove the overly specific version matrix which would require updates for every release.


bernd (approved):
Thanks! :+1:

PR #12737, opened 2022-05-25
Improve hostname retrieval in EventProcessorEventFactory

bernd:
Previously we looked up the local Graylog node in MongoDB to get the
stored hostname. This can fail when MongoDB isn't reachable, or the node
is overloaded, and its entry in the "nodes" collection in the database is
gone.

This change introduces a Hostname class and a provider to retrieve the
local hostname without talking to the database.


kingzacko1 (approved):
LGTM

Issue #12453, opened 2022-04-11
Certain Users-related API requests slow with large number of users


danotorrey:
## Overview
When a large number of users are present in Graylog, certain user-related API requests seem to be slow and take a long time. The requests take longer when the number of users in Graylog increases.

The problem was initially reported and investigated in the customer support ticket HS842447159.

## Details
In the support ticket, certain actions were reported by the user to be slow (taking more than 20 seconds):
* Viewing the Create new Alert page (of type Email Notification), which loads all users from the database in the process of preparing a dropdown list with all users.
* Sharing certain entities with specific users or team (using the Share button). For example sharing a stream using the Share button on the main streams page.
* Any other request that hits the `GET /api/users/` endpoint or `userService.loadAll()`, which loads all users from the database. 

## Testing
I was able to reproduce the issue by adding 1600 users to my Graylog database in my local development environment. When I attempt to create an email notification alert definition, the page freezes while the `GET /api/users` endpoint for about 60s. 

Directly calling the API endpoint produces the same issue. Note that the endpoint is [deprecated in favor of using pagination](https://github.com/Graylog2/graylog2-server/pull/8520).
https://github.com/Graylog2/graylog2-server/blob/bac3fe9f928b415603b9ed1e3d772de32232c8a1/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java#L213

In my testing, sharing entities was not very slow, but I also don't think I am testing the same setup as the customer. They mentioned in the issue that share operation and the `/api/authz/shares/entities/grn::::stream:000000000000000000000002/prepare` endpoint was returning 1700 available grants. I believe this means that have likely shared the `All Events` stream a large number of times.

## Root Cause
It appears that several areas of the Graylog users API and frontend use the `userService.loadAll()` method, which loads all users from MongoDB instead of just loading a page at a time, or searching for smaller subsets of users. This does not present a problem in smaller setups, but as the number of users increases, this can create a non-linear increase for the amount of time it takes for the requests to complete.

## Possible Solution

Refactor the pages and API request that load API all users from the database to only load a page at a time, to search for smaller subsets of users, or to perform filtering at the database layer.

This will require backend and frontend work, since I believe the same issue exists in both locations. 

### Example 1: Email alert callback
Since the Email Alert definition page loads a full list of users, it looks like backend and frontend code loads all users.
[...]

PR #12743, opened 2022-05-27
Optimize populating of user dropdown

patrickmann:
As described in #12453, creating an email notification can be slow when there are a lot of users. This is due to the amount of work required to build the full set of user objects, particularly computing the set of permissions assigned to each user. Since we are only interested in the list of user emails, we should avoid fetching extraneous data.

The same thing applies to the user creation form, which fetches existing users to ensure uniqueness.

This PR adds the following optional query parameters to the `api/users` endpoint:
- include_sessions (default: true): if false, do not check current session information for users
- include_permissions (default: true): if false, do not return any permissions data

By calling `api/users?include_permissions=false&include_sessions=false` we speed up populating the email dropdown significantly. Testing with 3000 users (each with a single team membership) and 1000 sessions we observed a speed-up of approximately 5x.

**Note**: the API change is backward compatible, since the new parameters are optional with default values that preserve the legacy behavior.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/users/UsersResource.java
@@ -152,6 +152,9 @@ public class UsersResource extends RestResource {
         this.searchQueryParser = new SearchQueryParser(UserOverviewDTO.FIELD_FULL_NAME, SEARCH_FIELD_MAPPING);
     }
 
+    /**
+     * @deprecated
+     */
     @GET
     @Deprecated
     @Path("{username}")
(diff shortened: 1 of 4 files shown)


mnin (approved):
LGTM

danotorrey:
I also added a comment in https://github.com/Graylog2/graylog2-server/issues/12453#issuecomment-1142675544. But, I also think that the improvements made here will still be helpful. We will just need to decide if this PR should be merged first, and a separate should be filed to address the dropdown improvements.

I will review and test the code in this PR tomorrow morning, as well as check back for feedback from @patrickmann and @ousmaneo on the dropdown suggestion.

patrickmann:
Lets get this merged and start a new PR for the filtered dropdown solution suggested in #12453. 
This PR is pretty low-risk and doesn't hurt anything. In certain cases it will still improve the UI responsiveness. So I think it's worth keeping it.

patrickmann:
> I am okay with merging this PR if we have a new issue for the widget change (including the removal of the parameters we are now adding) and a plan for when it will be done.

@Bernd Here is a new issue: #12819

danotorrey:
Thanks @patrickmann. I will finish reviewing today or tomorrow.

danotorrey (approved):

PR #12764, opened 2022-05-31
refactoring of PrometheusExporter for graylog forwarder

kodjo-anipah:
## Description
Graylog Forwarder is using FilteredPrometheusDropwizardMetricsFactory to expose Metrics for Prometheus. There is a newer and better way of doing so implemented in org.graylog.metrics.prometheus.PrometheusExporter. See issues https://github.com/Graylog2/forwarder/issues/51 https://github.com/Graylog2/forwarder/issues/72.

## Motivation and Context
We should refactor it to avoid confusion.

## How Has This Been Tested?
Change has been tested locally. 

-  No api changes on both side
- All forwarder metrics are still being exposed via the same name

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/metrics/prometheus/PrometheusExporter.java
+++ b/graylog2-server/src/main/java/org/graylog/metrics/prometheus/PrometheusExporter.java
@@ -33,6 +33,7 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog/metrics/prometheus/PrometheusExporter.java
```diff
@@ -113,4 +114,15 @@ private Collector createCollector(List<MapperConfig> mapperConfigs) {
                 new CustomMappingSampleBuilder(mapperConfigs)
         );
     }
+
+    public Collector createCollector() {
```
patrickmann, line 118: If this is just for testing then you might annotate it as `@VisibleForTesting`

patrickmann (approved):
Just a commenting suggestion - looks good otherwise.

graylog2-server/src/main/java/org/graylog/metrics/prometheus/PrometheusExporter.java
patrickmann: I see that `createCollector` is going to be called in Forwarder, so I'll resolve this.

PR #12773, opened 2022-05-31
Improve quality of backend API TypeScript types.

dennisoelkers:
## Motivation and Context

This PR is addressing a couple of quirks discovered for the generated backend API types:

  - Using pre-configured object mapper for JSON schema generation.
    Properly snake-cases property names of DTOs.
  - Fixing types for `Headers`, allowing string-arrays (e.g. for `Accept`)
  - Using parameter names from `(Query|Header|Form)Param` annotations.
  - Sanitize variable names generated from query parameters. 
    E.g. `import-strategy` becomes `importStrategy`, while leaving query parameter name intact.
  - Fixing parsing of `additional_properties`, e.g. types for List/Map values.
  - Avoid serializing models which are not referenced (to prevent TS errors for declared, but unused types)

This PR also includes a good example for replacing `fetch`-usages with function stubs in ae92d6b0bc07c5f81273bf6d4c7fc3753a7711ff. It is not that easy in all cases, only if the generated types and the existing code match up. This commit is included not only as an example, but also to create a reference between the existing code and generated types, leading to `tsc` checking the whole module on each run, making regressions visible.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/api/GenerateApiDefinition.java
+++ b/graylog2-server/src/main/java/org/graylog/api/GenerateApiDefinition.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableSet;
 import io.github.classgraph.ClassGraph;
 import io.github.classgraph.ClassInfo;
 import io.github.classgraph.ScanResult;
+import org.graylog2.shared.bindings.providers.ObjectMapperProvider;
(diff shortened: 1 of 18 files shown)


mikedklein (approved):
LGTM and tested locally 🎉

PR #12797, opened 2022-06-02
modifiy interface to differentiate between selecetable and already se…

kodjo-anipah:
## Description

EntitySharesService returns Grantees based on GranteeService implementation.
This filters global and user grantees based on the configuration flags. 
Added a new method getModifiableGrantees and changed getAvailableGrantees to return all visible grants for the user and then filter based on configuration but including already active shares. This is done to avoid major DTO refactoring on frontend and backend while keeping existing validation in place. 

## Motivation and Context

Customer has suggested that not being able to revoke ownership of entities from an individual user while the Share with Users system configuration is disabled is a bug. 
issue: https://github.com/Graylog2/graylog-plugin-enterprise/issues/3561
## How Has This Been Tested?

Tested locally. 
Steps to reproduce: 

1. Share stream with user/global
2. Disable sharing with user/global
3. Stream is still shared with user/global
4. Its not possible to remove active user/global share, because of config flag

After the fix even if flag is disabled active shares are visible to authorised persons to see them and only active shares of that type can be modified/removed. 
## Types of changes

## Checklist:

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3680

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
+++ b/graylog2-server/src/main/java/org/graylog/security/shares/DefaultGranteeService.java
@@ -67,7 +67,7 @@ public class DefaultGranteeService implements GranteeService {
                 .collect(ImmutableSet.toImmutableSet());
     }
 
-    public boolean isAllowedType(UserAndTeamsConfig config, Grantee grantee) {
+    protected boolean isAllowedType(UserAndTeamsConfig config, Grantee grantee) {
         final boolean permittedGlobal = config.sharingWithEveryone() && Grantee.GRANTEE_TYPE_GLOBAL.equals(grantee.type());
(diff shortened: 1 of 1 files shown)


graylog2-server/src/main/java/org/graylog/security/shares/GranteeService.java
```diff
@@ -24,6 +24,8 @@
 public interface GranteeService {
     Set<Grantee> getAvailableGrantees(User sharingUser);
 
+    Set<Grantee> getSelectableGrantees(Set<Grantee> availableGrantees);
```
kodjo-anipah, line 27: maybe rename it to modifiableGrantees which the user can modify, available to get all of them to be able to get activeShares in the [EntitySharesService.java](https://github.com/Graylog2/graylog2-server/pull/12797/files#diff-ba8058310df814ed5be1c2188dbe77ab670f9758bf080eee4689e41bdc10f3b8)

Issue #12818, opened 2022-06-06
Exception in thread "main" java.lang.ArithmeticException: / by zero


Sc0th:
Seeing this after updating container to 4.3.1(same on 4.3)

`2022-06-06 10:41:47,292 INFO : org.graylog2.bootstrap.preflight.SearchDbPreflightCheck - Connected to (Elastic/Open)Search version <Elasticsearch:7.10.2>
Exception in thread "main" java.lang.ArithmeticException: / by zero
        at org.graylog2.shared.system.stats.fs.OshiFsProbe.lambda$fsStats$11(OshiFsProbe.java:116)
`

Mongo 5.0.9
Elastic 7.10.2
official docker images

full log:

`Pod name is graylog-84f99dbddb-24lrx.
Defaulted container "graylog-server" out of: graylog-server, check-graylog-elasticsearch-ready (init)
2022-06-06 10:41:45,772 INFO : org.graylog2.featureflag.ImmutableFeatureFlagsCollector - Following feature flags are used: {}
2022-06-06 10:41:46,516 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: AWS plugins 4.3.1 [org.graylog.aws.AWSPlugin]
2022-06-06 10:41:46,517 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Integrations 4.3.1 [org.graylog.integrations.IntegrationsPlugin]
2022-06-06 10:41:46,519 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Collector 4.3.1 [org.graylog.plugins.collector.CollectorPlugin]
2022-06-06 10:41:46,520 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Threat Intelligence Plugin 4.3.1 [org.graylog.plugins.threatintel.ThreatIntelPlugin]
2022-06-06 10:41:46,520 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Elasticsearch 6 Support 4.3.1+78ee2f1 [org.graylog.storage.elasticsearch6.Elasticsearch6Plugin]
2022-06-06 10:41:46,520 INFO : org.graylog2.bootstrap.CmdLineTool - Loaded plugin: Elasticsearch 7 Support 4.3.1+78ee2f1 [org.graylog.storage.elasticsearch7.Elasticsearch7Plugin]
[...]

PR #12821, opened 2022-06-06
Prevent divide-by-zero exceptions in preflight checks

### LINKED ISSUE #12820: Divide-by-zero exception in preflight-checks blocks server startup
Computing file system stats for the preflight checks may result in a /0 exception.
In particular, `fileStore.getTotalInodes()` may return 0 when running in an LXC container. 

```
Jun 04 10:26:25 graylog graylog-server[2314]: Exception in thread "main" java.lang.ArithmeticException: / by zero
Jun 04 10:26:25 graylog graylog-server[2314]:         at org.graylog2.shared.system.stats.fs.OshiFsProbe.lambda$fsStats$11(OshiFsProbe.java:116)
```

Reported in community forum post [24164](https://community.graylog.org/t/graylog-server-fails-on-startup-after-upgrade-to-4-3/24164)

patrickmann:
Avoid blocking preflight on divide-by-zero exceptions. 

This just handles the divide by zero exceptions, without attempting to address the file system incompatibilities that are causing them. Not ideal, but if we are not even using a particular stat (e.g. percentage of unused inodes) then we should avoid blocking on it, when encountering some unexpected flavor of file system.

See linked issues for details.

Resolves #12820, #12818

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/system/stats/fs/OshiFsProbe.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/system/stats/fs/OshiFsProbe.java
@@ -19,8 +19,11 @@ package org.graylog2.shared.system.stats.fs;
 import com.google.common.collect.ImmutableSet;
 import org.apache.commons.lang3.StringUtils;
 import org.graylog2.Configuration;
+import org.graylog2.bootstrap.preflight.DiskJournalPreflightCheck;
 import org.graylog2.plugin.KafkaJournalConfiguration;
 import org.graylog2.shared.system.stats.OshiService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import oshi.hardware.HWDiskStore;
 import oshi.hardware.HWPartition;
 import oshi.hardware.HardwareAbstractionLayer;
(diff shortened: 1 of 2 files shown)


patrickmann:
I think we need to backport to 4.3, as it has been reported multiple times already.

mpfz0r:
> I think we need to backport to 4.3, as it has been reported multiple times already.

yes, absolutely!

mpfz0r (changes requested):
How about avoiding a division by zero here as well? just in case..

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/system/stats/os/OshiOsProbe.java#L76

patrickmann:
> How about avoiding a division by zero here as well? just in case..
> 
> https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/shared/system/stats/os/OshiOsProbe.java#L76

What should we use as default if we don't know the number of sockets? 1?

mpfz0r (approved):
LGTM :+1:

mpfz0r:
> What should we use as default for if we don't know the number of sockets? We could just set `totalSockets` to 1 in case it is 0 and let the other calculations proceed as is.

yeah, I think that's good enough

Issue #12788, opened 2022-06-01
Null pointer exceptions in GeoIP processor after update to 4.3


H2Cyber:
## Current Behavior
Following an upgrade to Graylog 4.3.1, my server.log in now flooded with the following Warnings :

```
2022-06-01T23:23:58.280+01:00 WARN  [ProcessBufferProcessor] Unable to process message <87c84c80-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:23:58.666+01:00 WARN  [ProcessBufferProcessor] Unable to process message <88030b91-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:23:58.667+01:00 WARN  [ProcessBufferProcessor] Unable to process message <880359b1-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:23:59.761+01:00 WARN  [ProcessBufferProcessor] Unable to process message <88aa2107-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:23:59.769+01:00 WARN  [ProcessBufferProcessor] Unable to process message <88ab8090-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:23:59.778+01:00 WARN  [ProcessBufferProcessor] Unable to process message <88acb916-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:24:00.390+01:00 WARN  [ProcessBufferProcessor] Unable to process message <890a1b66-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:24:00.682+01:00 WARN  [ProcessBufferProcessor] Unable to process message <8936a992-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:24:00.949+01:00 WARN  [ProcessBufferProcessor] Unable to process message <895f6741-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
2022-06-01T23:24:01.286+01:00 WARN  [ProcessBufferProcessor] Unable to process message <8992d351-e1f9-11ec-8ea4-005056afb272>: java.lang.NullPointerException
[...]

PR #12822, opened 2022-06-06
Fix GeoIP Processor NullPointerException Bug

roberto-graylog:
## Description

Reverted commit 424d3c77d0f915c94becac74565563e0dfbac234, which addressed the NullPointerException caused by null Latitude/Longitude returned by MaxMind.  Instead, updated MaxMindIpLocationResolver::doGetGeoIpData to catch Exception (rather than individual exceptions to revert behavior to exact prior behavior.  Updated MaxMindIpLocationResolverTest to conform to new behavior of MaxMindIpLocationResolver::doGetGeoIpData.

## Motivation and Context

While the pr **geoipresolver_error** addresses the NullPointerException caused by MaxMind returning a Null Latitude/Longitude pair, the end result is a `GeoLocationInformation` object with all empty/invalid fields (i.e. 0 or 'N/A').  The end result is that all these empty fields would be added to the message.  

Instead, catching `Exception` will allow any error to be logged, and an empty `GeoLocationInformation` to be thrown.

## Types of changes

## Checklist:


danotorrey (changes requested):
@roberto-graylog The logic in this PR looks good to me and tests successfully. 👍 

Just a couple small requests:
I think this PR should also remove the `GeoLocationInformation.create` method that was added in the previous PR. Looks like it's now no longer referenced other than in the tests. 

https://github.com/Graylog2/graylog2-server/blob/c7ecff1317dbacbcabbbcfd240a255d43119b130/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoLocationInformation.java#L47-L89

I also suggest using a shorter/more concise PR title. I think by default, the PR title will be used for the merge commit title, and if it's too long (more than 50 chars I think), an "..." will appear at the end of commit title at the time of merging. Perhaps a title similar to "Revise MaxMind GeoIP NullPointerException fix" would work.

ryan-carroll-graylog (approved):
Looks good to me. I ran some tests using IPs with location data as well as the IPs without data that were listed in the issue comments, and for IPs without location data the location fields were omitted and the IP was logged.

danotorrey:
@roberto-graylog The code LGTM! Thanks @roberto-graylog Did you see the feedback I had about the PR title in the comment above?

danotorrey:
Fixes https://github.com/Graylog2/graylog2-server/issues/12788

Issue #7470, opened 2020-02-15
Reordering of tabs in dashboards


pascalberger:
## What?

While it is possible to add multiple tabs on a dashboard, tabs can't be rearranged.

## Why?

Having the possibility to rearrange tabs would allow to add tabs to an existing dashboard and still keep them in a logical order.

## Your Environment

* Graylog Version: 3.2.1
* Elasticsearch Version: 6.8.6
* MongoDB Version: 3.6.17
* Operating System: Docker images
* Browser version:

bit-plumber:
I agree.  It's frustrating when you come up with a great new use case but can't arrange it where it logically needs to be in the dashboard.

bit-plumber:
I agree, this is a needed feature.

xaviercho:
This Feature would be great as I'm currently remaking widgets in dashboards.

r-duran:
Any update on this? When I add new pages (tabs) to existing dasboard, I can't re order them as my needs.

JeroenStravers1:
I'll echo the above. I need quite a few tabs in alphabetical order for my use case and will need to add/remove entries. This is currently impossible. This forces me to make a single, gigantic dashboard instead, the tab function might as well not exist.

chalfling:
@dennisoelkers Hello!
I updated from 4.2.9 to 4.3.0 and tab order was a mess. Why can't I reorder this?

HiltiRaceteam:
We have the same problem after the update vom 4.2.9 to 4.3.0
The tab order is random, and if you duplicate a tab the order is recreated randomly.

HungryHowies:
I'm Also having the same issue. with tab order with GL 4.3.1. If I duplicate the tab it will randomly  place it  either in front or behind the  original tab. If I create a new tab it seams I have no control of what order its placed.

agamemnon23:
Since upgrading from 4.2.9 to 4.3.1, I'm also having the issue were all my dashboard tabs have been reordered.
Additionally, adding a new tab or duplicating a tab reorders all of them again.

dennisoelkers:
Hey @chalfling, @HiltiRaceteam, @HungryHowies, @agamemnon23,

this is thanks for reporting this. We will look into it as soon as possible!

H2Cyber:
I am also affected, as the update from 4.2.9 to 4.3.1 reordered all of my dashboard pages randomly. 

[...]

PR #12833, opened 2022-06-07
Use ordered query set for consistent tab ordering.

dennisoelkers:
## Motivation and Context

**Note:** This PR should be backported to `4.3`.

With the introduction of the `SearchDTO` class, which is meant to decouple internal data structures from REST requests/responses, a simple `Set` was used for queries. Unfortunately, the implementation for the `Set` used during deserialization does not preserve insertion order, leading to a different order when this `SearchDTO` is returned back to the caller after saving. The order of the queries set is used in the frontend to order the query tabs, so it relies on a stable ordering.

Therefore, this PR is making the type of the `queries` property more specific, by defining it as `LinkedHashSet`, which does preserve insertion order.

Obviously, the approach of relying on the ordering of elements in the set is not ideal. It does not communicate the fact that insertion order defines tab order and any accidental change to the query order in this set will mess up the tab order as well. Instead, we should prospectively make the order of the `queries` set in the search irrelevant. The tab ordering should be a separate property in the view, providing an explicit ordering of query tabs by referencing their specific id.

Refs #7470.
Refs #12834.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3679

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
@@ -174,7 +174,7 @@ public class SearchSyncIT {
                 .accept("application/vnd.graylog.search.v2+json")
                 .contentType("application/vnd.graylog.search.v2+json")
                 .when()
-                .body(fixture("org/graylog/plugins/views/search-with-three-empty-queries.json"))
+                .body(fixture("org/graylog/plugins/views/search-with-three-empty-queries-v2.json"))
                 .post("/views/search")
                 .then()
                 .statusCode(201)
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
@@ -153,6 +153,7 @@ public class SearchSyncIT {
     @ContainerMatrixTest
     void testThatQueryOrderStaysConsistentInV1() {
         given()
+                .config(sut.withGraylogBackendFailureConfig())
                 .spec(requestSpec)
                 .accept("application/json")
                 .contentType("application/json")
@@ -160,6 +161,7 @@ public class SearchSyncIT {
                 .body(fixture("org/graylog/plugins/views/search-with-three-empty-queries.json"))
                 .post("/views/search")
                 .then()
+                .log().ifStatusCodeMatches(not(201))
                 .statusCode(201)
                 .assertThat()
                 .body("queries*.id", contains("4966dd79-2c7d-4ba9-8f90-c84aea7b5c49",
@@ -170,6 +172,7 @@ public class SearchSyncIT {
     @ContainerMatrixTest
     void testThatQueryOrderStaysConsistentInV2() {
         given()
+                .config(sut.withGraylogBackendFailureConfig())
                 .spec(requestSpec)
                 .accept("application/vnd.graylog.search.v2+json")
                 .contentType("application/vnd.graylog.search.v2+json")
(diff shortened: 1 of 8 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchDTO.java
```diff
@@ -41,7 +42,7 @@ public abstract class SearchDTO {
     public abstract String id();
 
     @JsonProperty
-    public abstract Set<QueryDTO> queries();
+    public abstract ImmutableSet<QueryDTO> queries();
```
luk-kaminski, line 45: ImmutableSet does not guarantee maintaining insertion order. In fact it has some sub classes that will not maintain it at all (i.e. ImmutableSortedSet will sort the entries). 
While in this particular case it works, I'd consider replacing it with LinkedHashSet, to explicitly state that we are interested in maintaining the insertion order of the set.

full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
```diff
@@ -149,6 +150,40 @@ void testRequestStoredSearchWithGlobalOverrideKeepingOnlySingleSearchType() thro
                 .body("results.f1446410-a082-4871-b3bf-d69aa42d0c97.search_types", hasKey("01c76680-377b-4930-86e2-a55fdb867b58"));
     }
 
+    @ContainerMatrixTest
+    void testThatQueryOrderStaysConsistentInV1() {
+        given()
+                .spec(requestSpec)
```
janheise, line 156: `given()
                .config(sut.withGraylogBackendFailureConfig())
                .spec(requestSpec)`
and
`                .then()
                .log().ifStatusCodeMatches(not(201))
                .statusCode(201)
`

would create better logging in case of underlying errors on Jenkins

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchDTO.java
```diff
@@ -41,7 +43,7 @@ public abstract class SearchDTO {
     public abstract String id();
 
     @JsonProperty
-    public abstract Set<QueryDTO> queries();
+    public abstract LinkedHashSet<QueryDTO> queries();
```
janheise, line 46: I understand the usage/creation of the Sets as LinkedHashSets but is it necessary to change the type/interface here?

janheise, line 46: I understand the usage/creation of the Sets as LinkedHashSets but is it necessary to change the type/interface here?

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SearchDTO.java
```diff
@@ -69,7 +71,11 @@ public abstract static class Builder {
         public abstract String id();
 
         @JsonProperty
-        public abstract Builder queries(Set<QueryDTO> queries);
+        public abstract Builder queries(LinkedHashSet<QueryDTO> queries);
```
janheise, line 74: I understand the usage/creation of the Sets as LinkedHashSets but is it necessary to change the type/interface here?

janheise, line 74: I understand the usage/creation of the Sets as LinkedHashSets but is it necessary to change the type/interface here?

dennisoelkers: Yes, this signals to jackson that it should be deserialized as `LinkedHashSet`, otherwise it will be any `Set`-implementation.

full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
dennisoelkers: :white_check_mark:

janheise (approved):
Thank you, LGTM!

PR #12838, opened 2022-06-07
Job scheduler features for archiving

mpfz0r:
# Backend

 - Support Job cancellation
 - Rename JobExecutionContext isRunning() to schedulerIsRunning() to avoid confusion
 - Add JobTriggerStatus to SystemJobSummary

 - Add progress support to the JobScheduler
   To make the progress easily accessible, put the DBJobTriggerService into JobExecutionContexts.
   This feels a bit cludgy, but I couldn't come up with a better idea.

 - Introduce JobResourceHandler / JobResourceHandlerService
   Each JobResourceHandler implementation is responsible on checking the permissions of
   the UserContext and perform the generic job actions accordingly.

   All JobResourceHandlers will be called by the JobResourceHandlerService,
   which can return either JobTriggerDTOs or for compat reasons with the
   existing SystemJobs implementation: SystemJobSummary objects.

 - Duplicate Job type into the JobTriggerDTO.
   This makes it easier to access that info when converting JobTriggers into SystemJobSummary objects.

 - Add last_owner field to the JobTriggerDTO, so we can display which node executed a job.

 - Add lock context support to the LockService
   An existing lock with a context can only be renewed from the for the exact same node and context.
   This can be used to make a lock for a resource non-reentrant.

 - Refactor LockService a bit so it is more convenient to use.

# Frontend

- `ConfirmDialog` can now hide the cancel button, in this case the same function is used for the "X" on the right/top as for the `onConfirm`
- `SystemJob` are considering the `job_status` (with "complete") and the `percent_complete` attribute to decide if the job is done
- `SystemJob` renders an `Acknowledge Job` button with calling a new (not yet existing) API endpoint
- `SystemJob` now have badges to make it easier to see what status they are in

## Screenshots

Screenshot 2022-07-14 at 14 10 32

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3681


graylog2-server/src/main/java/org/graylog/scheduler/JobTriggerDto.java
```diff
@@ -34,7 +34,9 @@
 @AutoValue
 @JsonDeserialize(builder = JobTriggerDto.Builder.class)
 public abstract class JobTriggerDto {
-    private static final String FIELD_ID = "id";
+    public static final String FIELD_ID = "_id";
```
mpfz0r, line 37: note to self: don't change that

graylog2-server/src/main/java/org/graylog2/cluster/lock/LockService.java
```diff
@@ -17,10 +17,30 @@
...
+    /**
+     * Request a lock. If a lock already exists, the lock expiry time will be extended.
+     *
+     * @param resource Unique identifier for the resource that should be guarded by this lock.
+     * @param lockContext  an identifier that will be appended to the callers' node id. This will create the lock owner string.
+     *                       A context can be used for resources that should only allow a single lock to be acquired, even from the same node.
+     *                       If the lockContext is null, only the nodeId will be used.
+     * @return A {@link Lock} object, if a lock was obtained. An empty {@link Optional}, if no lock could be acquired.
+     */
+    Optional<Lock> lock(@Nonnull String resource, @Nullable String lockContext);
```
bernd, line 33: Since we are adding the `extendLock` convenience method below, how about keeping the `lock(String resource)` method to avoid passing a `null` context if it's not needed?

(same for `unlock(String resource)` below)

graylog2-server/src/main/java/org/graylog2/cluster/lock/MongoLockService.java
```diff
@@ -98,23 +99,30 @@ private void ensureTTLIndex(MongoCollection<Document> collection, Bson updatedAt
...
-    public Optional<Lock> lock(@Nonnull String resource) {
+    public Optional<Lock> lock(@Nonnull String resource, @Nullable String lockContext) {
+        if (lockContext == null) {
+            return doLock(resource, nodeId.toString());
+        }
+        return doLock(resource, nodeId.toString() + "-" + lockContext);
+    }
+
+    @Override
+    public Optional<Lock> extendLock(@Nonnull Lock lock) {
```
bernd, line 111: We should probably add a null check here to avoid an NPE if `lock` is null.

graylog2-server/src/main/java/org/graylog2/cluster/lock/MongoLockService.java
```diff
@@ -129,12 +137,25 @@ public Optional<Lock> lock(@Nonnull String resource) {
...
-    public Optional<Lock> unlock(@Nonnull String resource) {
+    public Optional<Lock> unlock(@Nonnull String resource, @Nullable String lockContext) {
+        if (lockContext == null) {
+            return doUnlock(resource, nodeId.toString());
+        }
+        return doUnlock(resource, nodeId.toString() + "-" + lockContext);
+    }
+
+    @Override
+    public Optional<Lock> unlock(@Nonnull Lock lock) {
```
bernd, line 148: Let's add a `null` check here for `lock`.

graylog2-server/src/main/java/org/graylog2/migrations/V20220623125450_AddJobTypeToJobTrigger.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public void upgrade() {
+        final MigrationCompleted migrationCompleted = clusterConfigService.getOrDefault(MigrationCompleted.class, MigrationCompleted.createEmpty());
+
+        // Run this more than once in case old cluster nodes are still writing triggerDTOs with empty types
+        if (migrationCompleted.executionCount() >= 3) {
+            return;
+        }
```
bernd, line 72: I think this running this three times is quite arbitrary. :smile: How about removing the `MigrationCompleted` value and always adding a quick guard query, and only running the migration if there are any triggers that don't have the `job_definition_type` field?

graylog2-server/src/main/java/org/graylog2/cluster/leader/AutomaticLeaderElectionService.java
```diff
@@ -93,7 +98,8 @@ private void runIteration() {
         final boolean wasLeader = isLeader;
 
         try {
-            isLeader = lockService.lock(RESOURCE_NAME).isPresent();
+            optionalLock = lockService.lock(RESOURCE_NAME, null);
+            isLeader = optionalLock.isPresent();
```
bernd, line 102: Are the changes to this class needed? It would be great if we don't have to store the lock as a field if it's unnecessary. I think we don't have to modify this class at all if we add back the `lock(resource)` and `unlock(resource)` methods as suggested in the `LockService` comments. :slightly_smiling_face:

graylog2-server/src/main/java/org/graylog2/cluster/lock/LockService.java
```diff
@@ -17,10 +17,30 @@
...
+     * Extend the expiry time of an existing lock.
+     *
+     * @param existingLock the lock that should be extended.
+     * @return A {@link Lock} object, if the lock could be extended. An empty {@link Optional}, if no lock extension could be acquired.
+     */
+    Optional<Lock> extendLock(@Nonnull Lock existingLock);
+
+
+    Optional<Lock> unlock(@Nonnull String resource, @Nullable String lockContext);
+    Optional<Lock> unlock(Lock lock);
```
bernd, line 45: Please add Javadoc here as well. Thank you!

graylog2-server/src/main/java/org/graylog2/cluster/lock/MongoLockService.java
```diff
@@ -98,23 +99,30 @@ private void ensureTTLIndex(MongoCollection<Document> collection, Bson updatedAt
...
-     * @param resource Unique identifier for the resource that should be guarded by this lock.
-     * @return A {@link Lock} object, if a lock was obtained. An empty {@link Optional}, if no lock could be acquired.
-     */
     @Override
-    public Optional<Lock> lock(@Nonnull String resource) {
+    public Optional<Lock> lock(@Nonnull String resource, @Nullable String lockContext) {
+        if (lockContext == null) {
+            return doLock(resource, nodeId.toString());
+        }
+        return doLock(resource, nodeId.toString() + "-" + lockContext);
```
bernd, line 107: Let's use a small helper method to build the `lockedBy` parameter from the node ID and the `lockContext` to avoid duplication. We can also move the `null` check into that method and shorten the `lock` and `unlock` methods.

bernd (changes requested):
Thank you! :pray: A few comments below.

graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
```diff
@@ -425,6 +438,8 @@ public int forceReleaseOwnedTriggers() {
                 DBQuery.is(FIELD_LOCK_OWNER, nodeId),
                 DBQuery.is(FIELD_STATUS, JobTriggerStatus.RUNNING)
         );
+        // TODO should this also set Once scheduled triggers to RUNNABLE?
+        // TODO should we set triggers to CANCELLED here?
```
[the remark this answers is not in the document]
mpfz0r: Re TODO#1: I was wondering whether aborted (due to the server shutdown) once scheduled jobs should be rescheduled.

Re TODO#2: Would this be a good place to indicate that jobs were aborted on server shutdown and mark them as CANCELLED?

graylog2-server/src/main/java/org/graylog/scheduler/JobExecutionEngine.java
```diff
@@ -83,6 +83,7 @@ public JobExecutionEngine(DBJobTriggerService jobTriggerService,
      * Signal shutdown to the engine.
      */
     public void shutdown() {
+        // TODO this should indicate that running jobs have been aborted
```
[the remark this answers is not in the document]
mpfz0r: let's have a zoom on the whole shutdown topic..

graylog2-server/src/main/java/org/graylog/scheduler/rest/JobResourceHandler.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Pluggable interface for common job scheduler resource tasks.
+ * Implementations are responsible on checking the permissions of the {@link UserContext}
+ */
+public interface JobResourceHandler {
+
+    // TODO should we expose the UserContext into the plugin API like this? Maybe introduce an interface...
```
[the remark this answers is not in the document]
mpfz0r: yep, let's keep it like that :+1:

graylog2-server/src/main/java/org/graylog/scheduler/rest/requests/CreateJobTriggerRequest.java
```diff
@@ -75,6 +76,7 @@ public JobTriggerDto toDto() {
 
     public JobTriggerDto toDto(JobSchedulerClock clock) {
         return JobTriggerDto.builderWithClock(clock)
+                .jobDefinitionType(EventProcessorExecutionJob.TYPE_NAME)
```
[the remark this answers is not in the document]
mpfz0r: ah, thanks. no this can go :+1:

graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
```diff
@@ -464,4 +479,33 @@ public void updateLockedJobTriggers() {
...
+        final DBUpdate.Builder update = DBUpdate.set(JobTriggerDto.FIELD_IS_CANCELLED, true);
+
+        return Optional.ofNullable(db.findAndModify(query, update));
+    }
+
+    public boolean triggerWasCancelled(JobTriggerDto trigger) {
+        final JobTriggerDto reloadedTrigger = db.findOneById(getId(trigger));
+        if (reloadedTrigger == null) {
+            return false;
+        }
```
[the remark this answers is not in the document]
mpfz0r: It could also be that the user doesn't have the right permissions (expressed in the query) to cancel the trigger.
We usually don't expose any information if an object is not found.

graylog2-server/src/main/java/org/graylog2/cluster/lock/MongoLockService.java
mpfz0r: sure. I wonder how much benefit we get from `@Nonnull` though. IntelliJ claims it can never be null :-D

graylog2-server/src/main/java/org/graylog2/migrations/V20220623125450_AddJobTypeToJobTrigger.java
mpfz0r: I agree that it is, but I was afraid that this might be expensive. After all, I've witnessed setups with millions of notification triggers already

graylog2-server/src/main/java/org/graylog2/rest/resources/cluster/ClusterSystemJobResource.java
```diff
@@ -58,28 +66,53 @@
...
-    public Map<String, Optional<Map<String, List<SystemJobSummary>>>> list() throws IOException {
-        return getForAllNodes(RemoteSystemJobResource::list, createRemoteInterfaceProvider(RemoteSystemJobResource.class));
+    public Map<String, Optional<Map<String, List<SystemJobSummary>>>> list(@Context UserContext userContext) throws IOException, NodeNotFoundException {
+        final Map<String, Optional<Map<String, List<SystemJobSummary>>>> forAllNodes = getForAllNodes(RemoteSystemJobResource::list, createRemoteInterfaceProvider(RemoteSystemJobResource.class));
+
+        final List<SystemJobSummary> jobsFromScheduler = jobResourceHandlerService.listJobsAsSystemJobSummary(userContext);
+        final String thisNodeId = serverStatus.getNodeId().toString();
+
+        final ImmutableList.Builder<SystemJobSummary> jobDummys = ImmutableList.builder();
+        // TODO remove stubs
```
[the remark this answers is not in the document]
mpfz0r: will do once @mnin is done with the frontend changes

graylog2-server/src/main/java/org/graylog2/cluster/lock/MongoLockService.java
bernd: I think we only benefit from it in the IDE when we try to pass `null` to it at some point. Some static analysis tools could also use it.

graylog2-server/src/main/java/org/graylog2/cluster/leader/AutomaticLeaderElectionService.java
bernd: @mpfz0r My concern is that we add a mutable field to the class even though it's not needed for the locking API. I think it's good practice to minimize mutable fields in a class.

What is the improvement you see with storing the lock? :slightly_smiling_face:

bernd (changes requested):

graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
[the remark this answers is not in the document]
mpfz0r: ah, sorry. i confused this with the cancelTriggerByQuery() method. 
I could throw an exception, but I'm gonna ignore it in the consumer either way, so I don't think it makes much of a difference.
Your call :-)

graylog2-server/src/main/java/org/graylog2/migrations/V20220623125450_AddJobTypeToJobTrigger.java
mpfz0r: I'm not a mongodb expert, but I think I've read somewhere that count might be expensive, even though the field has an index.
e.g.: https://www.mongodb.com/community/forums/t/why-is-count-so-slow/110389

However, this reminded me to add an index to `FIELD_JOB_DEFINITION_TYPE` as well now :+1:

graylog2-server/src/main/java/org/graylog2/cluster/leader/AutomaticLeaderElectionService.java
mpfz0r: Feels more natural: `here is your lock object` ... `thanks, now have it back please` :-)

graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
```diff
@@ -373,6 +383,8 @@ public boolean releaseTrigger(JobTriggerDto trigger, JobTriggerUpdate triggerUpd
         requireNonNull(trigger, "trigger cannot be null");
         requireNonNull(triggerUpdate, "triggerUpdate cannot be null");
 
+        final boolean wasCancelled = triggerWasCancelled(trigger);
```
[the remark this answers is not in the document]
mpfz0r: makes sense :+1:

graylog2-server/src/main/java/org/graylog2/cluster/leader/AutomaticLeaderElectionService.java
bernd: Okay, fair enough. :slightly_smiling_face:

graylog2-server/src/main/java/org/graylog2/migrations/V20220623125450_AddJobTypeToJobTrigger.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    public ZonedDateTime createdAt() {
+        return ZonedDateTime.parse("2022-06-23T12:54:50Z");
+    }
+
+    @Override
+    public void upgrade() {
+        final MigrationCompleted migrationCompleted = clusterConfigService.getOrDefault(MigrationCompleted.class, MigrationCompleted.createEmpty());
```
bernd, line 69: Let's rename the state class to something other than "migration completed". The current naming implies that we don't run the migration when the state class exists in the database.

bernd (changes requested):
One small comment. I think we are good to go from a backend perspective. :slightly_smiling_face:

bernd (approved):
Thanks for the fixes! The backend code is good to merge from my perspective. I didn't review any frontend code, though.

mpfz0r:
> Thanks for the fixes! The backend code is good to merge from my perspective. I didn't review any frontend code, though.

Thanks for the review! Much appreciated.

graylog2-web-interface/src/components/systemjobs/SystemJob.tsx
```diff
@@ -0,0 +1,130 @@
...
+    color: ${theme.colors.variant.light.default};
+    
+    :hover {
+      color: ${theme.colors.variant.default};
+    }
+  }
+`);
+
+const SystemJob = ({ job }) => {
+  const jobIsOver = job.job_status === 'complete' || job.percent_complete === 100 || job.job_status === 'cancelled' || job.job_status === 'error';
```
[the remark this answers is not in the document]
mnin: done in https://github.com/Graylog2/graylog2-server/pull/12838/commits/b80696705b86b521f5b6778e0f199bb635397b21

PR #12901, opened 2022-06-15
Custom exception when ES batch size too large

patrickmann:
When ES throws an exception due to batch size too large we now generate a custom exception to improve code maintainability.


mpfz0r (approved):
LGTM :+1:

PR #12906, opened 2022-06-16
Added Graylog Schema Enforcement check to GeoIpResolverEngine when selecting field name prefix to add to message

### LINKED ISSUE #12904: geoip resolver strips geoip fieldnames of "_ip" if fieldname containing ip address is source_ip, destination_ip or host_ip
To align with the Graylog Information Model schema we rename source and destination ip address fieldnames as source_ip and destination_ip.

Since version 4.3.x it appears that the geoip resolved fields strip the "_ip" from the resulting fieldnames.

Example: for CheckPoint messages we use a Copy Input extractor to cut the field sourceAddress into the field source_ip.

The expected resulting geoip fields would be:
source_ip_geo_country
source_ip_geo_country_iso

The actual resulting geoip resolved fields look like the below:
source_geo_country
source_geo_country_iso

This is inconsistent with the behaviour if the fieldnames are not source_ip, destination_ip or host_ip. It's also inconsistent with the way that pipeline lookups work if configuring geoip lookups separately. It looks like this might be by design (snippet below), but it causes some overwrite problems other fields named source or destination, or host exist. 

Is this by design, and will this behaviour be permanent?

public class GeoIpResolverEngine {
private static final Logger LOG = LoggerFactory.getLogger(GeoIpResolverEngine.class);

/**
* A mapping of fields (per the Graylog Schema) that to search that contain IP addresses. When the user opts to
* enforce the Graylog Schema ONLY these fields will be checked; otherwise, all message fields will be checked.
* to see if they have valid Geo IP information.
*
* <p>
* The mapping is <b>field name</b> -> <b>new message field prefix</b>, where field_name is the field name expected in the message
* which will be searched in the GeoIP database, and the new message field prefix is the prefix for the new field with the GeoIP data
* that will be inserted into the message.
* </p>
*/
private final Map<String, String> ipAddressFields = new ImmutableMap.Builder<String, String>()
.put("source_ip", "source")
.put("host_ip", "host")
.put("destination_ip", "destination")
.build();

roberto-graylog:
## Description

Closes #12904 

Added Graylog Schema Enforcement check to GeoIpResolverEngine when selecting field name prefix to add to message.

If the user has opted not to enforce the Graylog Schema, we want to scan ALL fields for possible IP addresses, as well as add original field name as the prefix for the Geo message field.  For example, given the field `source_ip`, with Graylog Schema Enforcement Off, the field added for the **Country Name** should be `source_ip_geo_country` (with the original field name as the prefix).  However, with enforcement on, the field name should be `source_geo_country`.

## Motivation and Context

Without checking for schema enforcement on field prefix generation, the user's message fields will be modified in an unexpected/undesired way when enforcement is turned off.

## Testing Steps
1. Turn Graylog Schema Enforcement ON for the Geo-Location Processing and submit a message with IP fields (source_ip, destination_ip, host_ip).  Confirm that the _geo_ fields added to the message have a prefix that exclude the '_ip' token in the input field.  For example, for `source_ip`, the **City Name** should be in the field `source_geo_city`.
2. Turn Schema Enforcement OFF and submit a message with IP fields.  Confirm that the _geo_ fields added to the message have a prefix that is exactly the input field name.  For example, `source_ip` should have a city name field of `source_ip_geo_city`.

## How Has This Been Tested?

Added unit tests and tested manually end-to-end.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java
@@ -92,9 +92,9 @@ public class GeoIpResolverEngine {
                 continue;
             }
 
-            //IF the user has opted NOT to enforce the Graylog Schema, the key will likely not
-            //be in the field map--in such cases use the key (full field name) as the prefix
-            final String prefix = ipAddressFields.getOrDefault(key, key);
+            // IF the user has opted NOT to enforce the Graylog Schema, the key will likely not
+            // be in the field map--in such cases use the key (full field name) as the prefix.
+            final String prefix = enforceGraylogSchema ? ipAddressFields.getOrDefault(key, key) : key;
 
             if (ReservedIpChecker.getInstance().isReservedIpAddress(address.getHostAddress())) {
                 message.addField(prefix + "_reserved_ip", true);
--- a/graylog2-server/src/test/java/org/graylog/plugins/map/geoip/GeoIpResolverEngineTest.java
+++ b/graylog2-server/src/test/java/org/graylog/plugins/map/geoip/GeoIpResolverEngineTest.java
@@ -26,6 +26,7 @@ import org.graylog.plugins.map.config.DatabaseVendorType;
 import org.graylog.plugins.map.config.GeoIpResolverConfig;
 import org.graylog2.plugin.Message;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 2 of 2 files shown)


ryan-carroll-graylog (approved):
Looks good! Tested with this fix and geo fields now contain the full original field name when Graylog Schema Enforcement is turned OFF.

danotorrey (approved):
LGTM and tests successfully. Thanks for the thorough unit tests @roberto-graylog!

PR #12942, opened 2022-06-24
Pin `@types/react` version across the project.

dennisoelkers:
## Motivation and Context

Prior to this change, we depended on the `@types/react` version being pulled in implicitly. This results in the version depending on the set of other dependencies and can change through a dependency update. In order to avoid this, this change is now adding an explicit dependency to it, pinning it to the most recent `17.x` version. Although we are on `16.x`, this is the version which matches our expected typings the most.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me!

PR #12944, opened 2022-06-24
Removing `javascript-natural-sort` dependency.

dennisoelkers:
## Motivation and Context

In the past we were using `javascript-natural-sort` in order to provide a function for natural sorting. This can be achieved through the [internationalization collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) way more efficiently.

We have used a convenience function in the search code for a while now.  This PR is now moving it to the general `logic` directory and replaces all existing usages of `javascript-natural-sort` with it. In addition, it removes the dependency itself.

Any future use case that needs natural sorting should use the `logic/DefaultCompare` module right from the beginning.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3756,Graylog2/graylog-plugin-collector#203

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me!

PR #12961, opened 2022-06-28
Update Jackson to version 2.13

patrickmann:
We want to upgrade to a recent Jackson version, while initially keeping everything else unchanged (as much as possible).
https://github.com/Graylog2/graylog-plugin-enterprise/issues/3674

Notes: Requires a fork of MongoJack 2.10.1 with minimal changes to work with Jackson 2.13.3: https://github.com/graylog-labs/mongojack

**Details**

- If configuration of `ObjectMapper` is modified after first usage, changes may or may not take effect, and configuration calls themselves may fail [[link](https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ObjectMapper.html)]. This behavior appears to have changed since version 2.9. It only affects a few tests which attempt to modify the mapper:
_graylog2-server/src/test/java/org/graylog/events/contentpack/facade/EventDefinitionFacadeTest.java
graylog2-server/src/test/java/org/graylog/events/contentpack/facade/NotificationFacadeTest.java_

- `DateTimeFormatter` now throws an exception for empty string, resulting in a JSON missing node that needs to be handled:
_graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java_

- Jackson 2.11 changed the default `DateTime` formatting for timezone offset to `hh:mm` instead of `hhmm` [[link](https://github.com/FasterXML/jackson-databind/issues/2643)]. We need to explicitly set the legacy format in the `ObjectMapper`.
_graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ObjectMapperProvider.java_
_graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java_
_graylog2-server/src/test/java/org/graylog2/jackson/MongoJodaDateTimeSerializerTest.java
graylog2-server/src/test/java/org/graylog2/jackson/MongoZonedDateTimeSerializerTest.java

- The new Mongo driver version has a slightly different error message for key collisions. This requires a change to a test:
_graylog2-server/src/test/java/org/graylog2/migrations/V2018070614390000_EnforceUniqueGrokPatternsTest.java_

- Jackson no longer includes Joda date/time by default. The Joda module needs to be explicitly registered:
_graylog2-server/src/test/java/org/graylog2/plugin/MessageSummaryTest.java_

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3766


graylog2-server/pom.xml
```diff
@@ -169,8 +169,7 @@
         <dependency>
             <!-- Workaround for https://github.com/FasterXML/jackson-bom/pull/38 -->
             <groupId>com.fasterxml.jackson.module</groupId>
-            <artifactId>jackson-module-scala_2.13</artifactId>
-            <version>2.9.10</version>
+            <artifactId>jackson-module-scala_3</artifactId>
```
mpfz0r, line 172: hmm, we haven't updated to scala 3 yet.
This workaround was only needed for jackson 2.9
I think it can be removed entirely

patrickmann:
Postponed to 5.0, when we expect to drop ES6.

patrickmann:
This was just a test branch, which we no longer need.

Issue #12869, opened 2022-06-09
Field value suggestions should be query-aware


coffee-squirrel:
## What?

Field value suggestions should be "aware" of the surrounding query, such that suggested values are mostly (ideally entirely, but there are probably performance concerns) limited to valid values for the overall query.

## Why?

Valid suggestions for a given field (e.g. `filebeat_log_file_path`) may vary wildly based upon other fields in the query (e.g. `source`).

By limiting suggested values to those that are valid for the surrounding (i.e. not being modified) parts of the query, it'd be possible to get more useful suggestions.  For example: only seeing the log file paths that are associated with a specific source.

## Your Environment

* Graylog Version: 4.3.0

PR #12980, opened 2022-06-30
Filtering query for suggestions

todvora:
This PR adds a possibility to define a filtering query to the suggestion request. This query should help to narrow down possible suggestions related to the query, as requested in https://github.com/Graylog2/graylog2-server/issues/12869

image

## How Has This Been Tested?
Added integration test

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SuggestionResourceIT.java
@@ -31,6 +31,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.graylog.testing.completebackend.Lifecycle.VM;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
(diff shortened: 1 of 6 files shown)


luk-kaminski:
It is a very promising change, @todvora !

What I am concerned about:
1. Front-end will have some hard time preparing proper _filtering_query_. If I understood everything correctly, for this query: `_exists_:destruction_method AND type:<cursor is here> AND NOT _exists_:vehicle`
FE has to know that `AND type: AND` **needs to be cut out** and only `_exists_:destruction_method AND NOT _exists_:vehicle` used for _filtering_query_.
2. Probably multiple occurrences of the same field should not be taken into consideration for _filtering_query_. For query `type:tank AND type:` you rather don't want to use `type:tank` to filter suggestions for the second _type_ occurence, as it will simply limit your suggestion to _tank_ value only...Another problem for FE?
3. Not to mention complexity of query with nested parenthesis and cursor somewhere in the middle...
4. Does it also have to work together with Search Filters ????

Which leads me to a doubt if FE shouldn't just throw all the data at BE (query string with cursor, streams, search filters) and BE should take all the havok of figuring out how to construct _filtering_query_ ...

PR #12982, opened 2022-06-30
Replace remaining usage of legacy `DateTime` class with new `DateTime` utils and `UserDateTimeContext`

linuspahl:
_Please note, this PR needs to be merged together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/3776_

## Motivation and Context

This is a follow-up PR for multiple PR's related to the refactoring of our date time handling. One goal it to replace the `DateTime` class. 
The class is directly calling the `CurrentUserStore`, which is not a good practice since it comes with some disadvantages. We used this class every time we worked with date times, but it is hard to understand that it will always transform the provided date, based on the user timezone. Instead we now implemented:
- `utils/DateTime` which provides functions to transform a date, independently of the timezone of the current user
- `contexts/UserDateTimeContext` which provides functions to convert a date, based on the user timezone.
- `components/common/Timestamp` and similar components to display a timestamp based on the user timezone. These components are already implementing the `UserDateTimeContext`.

With this PR we are replacing the remaining usage of the legacy `DateTime` class. There is only the `AlertMessages` component left in this repo, which will be removed completely soon.

Please note, I have not fixed all linter hints to reduce the complexity of this refactoring.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3776

### DIFF
--- a/graylog2-web-interface/src/components/common/DatePicker.tsx
+++ b/graylog2-web-interface/src/components/common/DatePicker.tsx
@@ -20,9 +20,8 @@ import type { DayModifiers } from 'react-day-picker';
 import DayPicker from 'react-day-picker';
 import styled, { css } from 'styled-components';
 
-import DateTime from 'logic/datetimes/DateTime';
-
 import 'react-day-picker/lib/style.css';
+import { toDateObject, adjustFormat } from 'util/DateTime';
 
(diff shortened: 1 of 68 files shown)


dennisoelkers (approved):
Looks good, great change! Thanks for all the work! :tada:

PR #12983, opened 2022-06-30
Proof of concept: Field-based metadata approach

danotorrey:
## Description
Proof of concept sample implementation of field-based metadata handling. There has been some discussion about the potential for a field-based metadata solution (also extending from the initial implementation idea from the team). This sample was created to explore the idea further and serve as a basis for discussion.

## Field-based metadata solution 
The field-based metadata solution includes the metadata on entities in an additional (optional) metadata field, and the entity fields remain at the document root as they currently exist. This allows backwards compatibility with existing entities, since a default can be applied for existing entities in the builder method. 

Sample object structure from this PR:
```
{
  // Explicit ID at root, as generally required by MongoDB 
  "id": "5d283b47b9d8210001781e9b",

  // Metadata specified as a field for backwards compatibility
  "_metadata": {
    "scope": "illuminate",
    "revision": 2,
    "created_at": "2022-06-13T08:47:12Z",
    "updated_at": "2022-06-29T12:00:28Z"
  },

  // Entity fields
  "title": "my title",
  "description": "my description"
}
```

## Backwards compatibility
Since we will want the solution to be used for entities that exist already, we will want to choose a solutions which is backwards compatible, or has a reasonable migration path. 

### Autovalue objects
Since the existing entities use Autovalue objects, finding a solution that is compatible with them would be helpful. This additional field approach seems to be compatible. 

Note that Autovalue objects cannot extend a parent class, so this example uses an `Entity` interface instead.

## Possible drawbacks
* This implementation example needs to explicitly handle metadata processing using a completely custom-crafted solution. Can MongoJack or other libraries help with this, so we won't have to build it all on our own?
* The interface solution shown here means that every Entity needs to explicitly define an `id` and `metadata` field with Mongo mapping annotations. That is not ideal, since it entails duplication. It would be possibly better to extend a class instead. The downside is that AutoValue does not support this.


danotorrey:
Closing this PR, since we decided on a different approach (see https://github.com/Graylog2/graylog2-server/pull/13032) pending Engineering team feedback. Let's keep the branch around for a bit.

PR #13009, opened 2022-07-06
Allow defining `id` for `QueryInput` textarea.

linuspahl:
## Motivation and Context

With this PR it will be possible to define an `id` for the `textarea` of the `QueryInput`.
This way it will be possible to connect a label with input. For example like this:
```
<label htmlFor="query-input-id">A label</label>
<QueryInput inputId="query-input-id" />
```

Before this change clicking on the label had no effect, in the described case.
The `QueryInput` already has a `name` attribute, but it is being used for the id of the `textarea` container.

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/queryinput/BasicQueryInput.tsx
+++ b/graylog2-web-interface/src/views/components/searchbar/queryinput/BasicQueryInput.tsx
@@ -94,7 +94,11 @@ const BasicQueryInput = forwardRef<StyledAceEditor, Props>((props, ref) => {
     if (editor) {
       editor.renderer.setScrollMargin(6, 5);
       editor.renderer.setPadding(12);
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/views/components/searchbar/queryinput/BasicQueryInput.tsx
```diff
@@ -139,6 +142,7 @@ const BasicQueryInput = forwardRef<StyledAceEditor, Props>((props, ref) => {
                        enableBasicAutocompletion={enableAutocompletion}
                        enableLiveAutocompletion={enableAutocompletion}
                        onBlur={onBlur}
+                       name="just-a-test"
```
maxiadlovskii, line 145: Probably it's a typo

maxiadlovskii (approved):

PR #13018, opened 2022-07-07
Add global Index Sets Defaults system configuration

danotorrey:
## Description

Adds a new Index Set Defaults section to the System > Configurations page where defaults for index set creation can be specified. This can also be configured by API. The new defaults apply to index creation everywhere in Graylog (currently Illuminate installations, and the System > Index Sets > Create Index set flow)

On the first server boot, the configuration values are initialized from the `server.conf` values via the `ElasticsearchConfiguration` class. This allows for both the index defaults configuration, and for system indices (Default, Events, and System Events index sets) to be created with the specified defaults. 

The `@Deprecated` annotation was also removed from the corresponding server configuration values, since we intend to continue using them as initialization defaults.

The following new defaults are also being used:
- Shards: `1`
- Rotation Strategy: `Size: 30GB`

Note that the config UI is hidden for Cloud environments.

The default `graylog.conf` was also updated:
- Comment out default properties, so that the defaults specified in `ElasticsearchConfiguration` will take precedence for users who establish a new configuration file from the template.
- Re-order the properties for consistency.
- Formatting adjustments.

Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/3264 https://github.com/Graylog2/graylog-plugin-enterprise/issues/3319

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3803

## Motivation and Context
The goal is to achieve globally-applicable Index set defaults, which apply uniformly throughout the application. Previously, hard-coded/arbitrary values were used on the Create Index Set page and new Illuminate index creation process. This change standardizes the initialization and use of central defaults. 

## Testing
- Verify that the new System > Configurations > Index Set Defaults are initialized with the correct defaults specified in the `server.conf` file. You might need to delete the configuration from Mongo to test multiple times (`db.cluster_config.deleteOne({"type": /IndexSetsDefaultConfiguration/})`). Also verify that the values can be customized.
- Verify that the Default, Events, and System Events indices are also created with the correct defaults.
- Verify that the System > Index Sets > Create Index page uses the correct defaults. Verify the same for the Illuminate index set.

Index sets to tests:
- Default
- Events
- System Events
- Indexer Failures
- Restored Archives

## How Has This Been Tested?

Testing will be done after the remaining tasks are finished.

A new Index Sets Configuration section has been added: 
image

image

## Types of changes

## Checklist:


boosty:
@danotorrey Thanks for working on this!

> Identify the appropriate default values for both shards and replicas current defaults are 4 and 0 respectively

For me it would be fine to keep the current defaults for now, to not change existing behaviour. But of course it would be good to get more opinions.

danotorrey:
Thanks @boosty. I'll finish this up with current defaults, and we can continue accepting more feedback while the review is in progress.

danotorrey:
It turns out that the `elasticsearch_shards` and `elasticsearch_replicas` server settings are not actually used in the new index creation flow, and are only used in some older migrations.

The defaults for those are hard-coded in the frontend:
https://github.com/Graylog2/graylog2-server/blob/85eb37534ebe8f77a3f78b80f340ed899b57ab4e/graylog2-web-interface/src/pages/IndexSetCreationPage.tsx#L126-L127

I'll work on migrating those hard-coded defaults to use the new in-database configuration.

danotorrey:
Looping in @dennisoelkers for early feedback in case you see anything noteworthy so far.

bernd (commented):
@danotorrey Can you please check the existing index-set-related migrations if there is anything we need to adjust for this? Thank you!

I think we should also add the other hardcoded index set settings to the default configuration while we are at it.

https://github.com/Graylog2/graylog2-server/blob/10cd789ac11957c29c91b497625195750808cb14/graylog2-web-interface/src/pages/IndexSetCreationPage.tsx#L126-L136

boosty:
Just wanted to mention that once merged, we should decide if we are also changing the defaults as discussed here:
https://github.com/Graylog2/graylog-plugin-enterprise/issues/3264

Could make sense to announce and document this together.

danotorrey:
> Just wanted to mention that once merged, we should decide if we are also changing the defaults as discussed here: [Graylog2/graylog-plugin-enterprise#3264](https://github.com/Graylog2/graylog-plugin-enterprise/issues/3264)
> 
> Could make sense to announce and document this together.

@boosty I agree. I think it makes sense to change the defaults in this PR too (and document/announce together), since a consensus appears to have been reached in https://github.com/Graylog2/graylog-plugin-enterprise/issues/3264.

danotorrey:
I am still making progress on this. I now have it working end-to-end. The Configurations page has an Index Defaults section, and the defaults are now used on the Create Index Set page. 

image

These are the steps remaining. I will keep working on this, and hope to have this PR ready for review soon.
- Add support for index rotation and retention defaults (since those are pluggable, there is a bit more work involved).
- Add unit tests.
- Remove deprecated Java properties that are no longer used.

danotorrey:
The frontend is now working end-to-end for all property types (including those that are pluggable [retention and rotation strategies]). 

Now, I need to implement fall-back defaults support for the configuration properties (as [mentioned here](https://github.com/Graylog2/graylog-plugin-enterprise/issues/4091#issuecomment-1255131604)). I need to look at the best way to do that. I'll also need to coordinate any planned changes to the config props with @todvora based on any in-progress work.

I'll also need to do some frontend cleanup and add tests.

todvora:
Hey @danotorrey, I have reverted all my changes, as they would cause only problems. Feel free to continue with your work. When you are done, I'll start with the prefix renaming, so we don't have git collisions everywhere.Thanks!

danotorrey:
@boosty @rich-graylog @waab76 An update on this feature with respect to Graylog `5.0`:

Development of the core functionality (frontend/backend) is now done, which provides the ability to specify database-defaults at runtime VIA API and UI and allow those to be used when creating new index sets in the frontend and through Illuminate. 

Now, we are reviewing the legacy `server.conf` configuration options, and assessing how best to handle/incorporate those with this change set. I am working to complete the review on those today or tomorrow, then will discuss with @bernd and @todvora.

Once we have finished the above-mentioned review, we will need to discuss this with Product to determine the plan for including this in Graylog `5.0`. The first beta will be built tomorrow, and this change will not be done/reviewed/merged in-time for that. So, we will need to discuss the risk/benefit tradeoffs and decide if we ship the complete feature in `5.0`, or defer certain parts to `5.1`. I will summarize here for feedback, or set up a call to discuss within the next days.

danotorrey:
@Graylog2/secdev This is far-enough along to start getting some feedback. The end-to-end implementation and functionality is all in-place.  There are a few tasks left (described above), which I will finish shortly. I'll also get the builds passing. But, any early feedback would be appreciated.

danotorrey:
This PR is now complete and ready for review. I will be adding tests shortly.

kingzacko1:
Found a small bug when testing.
- Change Rotation Strategy to `Index Time`
- Toggle `Rotate Empty Index Sets` and save the config
- Rotation period should be shows as 0 seconds and the configuration is in an error state if you try to edit again. 
- Go back in and update to a valid Rotation Period and save the config
- Edit the config again and `Rotate Empty Index Sets` is no longer true

Looks like the `Rotate Empty Index Sets` toggle has some faulty logic.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetDefaultsResource.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    @PUT
+    @Timed
+    @Consumes(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Update index set defaults configuration")
+    @RequiresPermissions({RestPermissions.CLUSTER_CONFIG_ENTRY_CREATE, RestPermissions.CLUSTER_CONFIG_ENTRY_EDIT})
+    @AuditEvent(type = AuditEventTypes.CLUSTER_CONFIGURATION_UPDATE)
+    public Response update(@ApiParam(name = "body", value = "The payload of the index set defaults configuration", required = true)
+                           @NotNull IndexSetsDefaultConfiguration config) throws IOException {
+        IndexSetValidator.Violation violation = indexSetValidator.validateRefreshInterval(
```
ryan-carroll-graylog, line 71: Playing around with this I think it may be a good idea to beef up the default index config field validation. For instance it's possible to save negative values for many of the numerical fields as well as invalid analyzer types.

image

Some of the bad numerical values are disallowed on the actual Create Index page (by a combo of front end and back end validation) stopping actual index creation there, but the Illuminate installer will still create indices with the bad config values.

image

danotorrey: Good call! I will get this validation in place tomorrow morning on both frontend and backend.

danotorrey:
@kingzacko1 @ryan-carroll-graylog I think I have addressed the recent feedback, so I think this is ready for another look.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/IndexSetDefaultsResource.java
ryan-carroll-graylog: These updates look solid. I think you're right that it makes sense to address the leftover validation deficiencies together in another PR as this is already pretty expansive.  

The list you have covers the other validation I'm tracking. If users can create their own analyzers then then definitely complicates validation of that field, and further supports investigating later in another PR.

ryan-carroll-graylog:
@danotorrey Testing a server start with freshly blown away DBs, and the following fields overridden in my graylog.conf:
```
elasticsearch_analyzer = standard
rotation_strategy = size
elasticsearch_max_docs_per_index = 69
elasticsearch_max_size_per_index = 69
elasticsearch_max_number_of_indices = 69
retention_strategy = close
elasticsearch_shards = 6
elasticsearch_replicas = 6
index_field_type_periodical_full_refresh_interval = 69s
```
I got the following results where some indexes took the configuration file values and some didn't (Yellow looks like the defaults were used instead of the conf values; Green is neither...):
image

I'm not sure how much of this is by design but I thought it merited calling out.

danotorrey:
> some indexes took the configuration file values and some didn't

Thanks seeing that @ryan-carroll-graylog. That is definitely an issue. I found the cause and will fix it.

graylog2-server/src/test/java/org/graylog2/configuration/IndexSetsDefaultConfigurationTest.java
```diff
@@ -0,0 +1,58 @@
...
+                .retentionStrategyConfig(DeletionRetentionStrategyConfig.create(10)).build();
+        final ObjectMapper objectMapper = new ObjectMapperProvider().get();
+        final Map map = objectMapper.convertValue(indexConfig, HashMap.class);
+        objectMapper.convertValue(map, IndexSetsDefaultConfiguration.class);
+    }
+
+    @Test
+    void test() {
+
+    }
```
kingzacko1, line 57: empty test

danotorrey:
Thanks for the additional feedback @kingzacko1! I'll clean those up.

@ryan-carroll-graylog I have been working on getting consistency among the system-created indices. I think I have almost made it through those updates. It was a little tricky, but I think I have it figured out. Should be pushing that up today or tomorrow.

ryan-carroll-graylog:
> Thanks for the additional feedback @kingzacko1! I'll clean those up.
> 
> @ryan-carroll-graylog I have been working on getting consistency among the system-created indices. I think I have almost made it through those updates. It was a little tricky, but I think I have it figured out. Should be pushing that up today or tomorrow.

Ok awesome, I'll keep an eye out for the updates.

graylog2-web-interface/src/components/configurations/IndexSetsDefaultsConfig.tsx
```diff
@@ -0,0 +1,237 @@
...
+                                           pluginExports={PluginStore.exports('indexRetentionConfig')} />
+      </StyledDefList>
+
+      <p>
+        <IfPermitted permissions="indices:changestate">
+          <Button bsStyle="info"
+                  bsSize="xs"
+                  onClick={() => {
+                    setShowModal(true);
+                  }}>Update
```
kingzacko1, line 157: nit: every other configuration button on the `System > Configurations` page says `Edit Configuration`. To keep it consistent, we may want to reuse that wording here too. 
image

graylog2-web-interface/src/components/indices/IndexMaintenanceStrategiesConfiguration.jsx
```diff
@@ -152,36 +152,55 @@ const IndexMaintenanceStrategiesConfiguration = ({
...
+        <Col md={12}>
+          <Input id="strategy-select"
+                 labelClassName="col-sm-3"
+                 wrapperClassName="col-sm-9"
+                 label={selectPlaceholder}>
+            <StyledSelect placeholder={selectPlaceholder}
+                          options={_availableSelectOptions()}
+                          matchProp="label"
+                          value={_activeSelection()}
+                          onChange={_onSelect} />
```
kingzacko1, line 188:
```suggestion
                          onChange={_onSelect}
                          clearable={false} />
```
The Select components can currently be cleared, but the rotation and retention configurations are required. You can save the form with the two selects cleared which does nothing to the existing config. Suggest making them not clearable to avoid this happening.

graylog2-web-interface/src/pages/useIndexDefaults.ts
```diff
@@ -0,0 +1,47 @@
...
+import ApiRoutes from 'routing/ApiRoutes';
+import type { IndexConfig } from 'components/configurations/IndexSetsDefaultsConfig';
+
+const fetchIndexDefaults: () => Promise<IndexConfig> = () => {
+  return fetch('GET', qualifyUrl(`${ApiRoutes.ClusterConfigResource.config().url}/org.graylog2.configuration.IndexSetsDefaultConfiguration`));
+};
+
+const useIndexDefaults = () => {
+  const { data, isLoading } = useQuery(
+    ['scope-permissions'],
```
kingzacko1, line 31: Is this maybe leftover from a copy/paste? I'm not super familiar with `React.useQuery` but I think we want this query key to be unique so maybe something like 'index-defaults' here instead?
```suggestion
    [index-defaults'],
```

graylog2-web-interface/src/components/configurations/IndexSetsDefaultsConfig.tsx
danotorrey: Great catch. Fixed.

graylog2-web-interface/src/pages/useIndexDefaults.ts
danotorrey: Good catch. Definitely a copy/paste error. Fixed.

danotorrey:
@kingzacko1 @ryan-carroll-graylog Thanks for the additional reviews. I think I have addressed all points of feedback. I also just pushed up https://github.com/Graylog2/graylog2-server/pull/13018/commits/854695a6d552552840ab9273f0ffc13d9dd4d19e and https://github.com/Graylog2/graylog-plugin-enterprise/pull/3803/commits/f98c14cf3f1a2e06035d00ad49c23b1c75f78007 in the related `graylog-project-enterprise` PR that ensures all system indexes (listed below) use the appropriate defaults on first server startup. That was a little tricky, but I tried to centralize the configuration settings as much as possible. I definitely would appreciate a look over that commit. 

- Default index set
- Graylog Events
- Graylog System Events
- Graylog Message Failures
- Restored Archives
- Illuminate indices

I still need to update the change log with the system index info, and I think there are some failing tests too, which I will address first thing in the morn.

danotorrey:
@ryan-carroll-graylog @kingzacko1 FYI, I just pushed up https://github.com/Graylog2/graylog2-server/pull/13018/commits/1fbf9fae2282b5c6deb64bcfe062d79cde222d91, which reverts the default index set creation migration back to it's original contents (matching `master`). That migration is unique, since back a long time ago, the default index settings were configurable in the `server.conf` file for a while, then, they were migrated to be configurable in the Cluster Config directly. Then, finally, they were migrated to the `index_sets` collection (VIA `IndexConfig` object). I had changed that migration to read from the `server.conf` directly, but it actually must still read from the Custer Config in the slim case that a user has modified the config there before upgrading to a subsequent version. Very slim case that could happen since the migrations are old, but I figured better to leave that one as it was and avoid possible issues.

The other default index migrations are not affected by this, since they all create new index sets that had not existed prior, so those can read from the `server.conf` (`Elasticsearch` configuration) as they are currently coded.

This should wrap up the changes needed for migrations. I'll update the changelog this afternoon, and unless you see any other issues, I  think this is probably ready for merging and letting it get some time on `master` for testing be fore 5.1. Let me know what you think!

ryan-carroll-graylog (approved):
I've re-run through testing with the latest updates and everything looks great! All the outstanding issues have been fixed and I haven't seen or uncovered anything new, so this is good to go from my end!

Nice work! This ended up being a real beast of an update but the final result is awesome!

kingzacko1 (approved):

danotorrey:
Thanks for all of the time and efforts reviewing this @ryan-carroll-graylog and @kingzacko1! I've just pushed the final update to the change log. I will plan to merge this as soon as it turns green!

boosty:
Thanks @danotorrey and all reviewers! 

This was a quite a significant change, but I think will be totally worth it.

changelog/unreleased/pr-13018.toml
```diff
@@ -0,0 +1,92 @@
...
+        "max_docs_per_index":10
+     },
+     "retention_strategy_class":"org.graylog2.indexer.retention.strategies.ClosingRetentionStrategy",
+     "retention_strategy_config":{
+        "max_number_of_indices":10,
+        "type":"org.graylog2.indexer.retention.strategies.ClosingRetentionStrategyConfig"
+     }
+  }'
+```
+"""
```
boosty, line 92: Thanks for providing the API example.

I just discussed the example values with @bernd. The ones used here can be quite dangerous if people would apply this example 1:1. 

Could we just use the new defaults instead?

danotorrey: Good catch, and good idea. I will fix that.

changelog/unreleased/pr-13018.toml
```diff
@@ -0,0 +1,92 @@
...
+- New indexes created through the Graylog Illuminate system.
+
+Once Graylog 5.1 is installed, these in-database defaults will be established, and the server configuration option
+values described above will no longer be used.
+
+## New Index Default values
+Unless user-specified defaults are specified, the following new defaults will be effective for all new index sets created:
+
+- Shards: 1 (previously 4 in many cases)
+- Rotation Strategy: Index Size - 10GB (previously Index Time [1D] in many cases)
```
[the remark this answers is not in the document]
danotorrey: Thanks for seeing this. Makes sense to me. I will update this too.

danotorrey:
@ryan-carroll-graylog @kingzacko1 When you have some time, can you please review the latest changes? Mostly just updates to the change log and upgrade notes + defaults.

UPGRADING.md
```diff
@@ -3,4 +3,24 @@ Upgrading to Graylog 5.1.x
 
 ## New Functionality
 
+## Index Default Configuration
+Support for configuring system index initialization defaults, and ongoing in-database index set defaults has been added.
+Please see the corresponding change log entry [pr-31018.toml](https://github.com/Graylog2/graylog2-server/blob/master/changelog/5.1.0/issue-1065.toml)
```
kingzacko1, line 8: The link text is **pr-31018.toml** but the link points to **issue-1065.toml**

kingzacko1 (approved):

ryan-carroll-graylog (approved):

danotorrey:
Unless any other feedback is received before, I will plan to merge this PR on Monday.

UPGRADING.md
```diff
@@ -3,4 +3,24 @@ Upgrading to Graylog 5.1.x
 
 ## New Functionality
 
+## Index Default Configuration
+Support for configuring system index initialization defaults, and ongoing in-database index set defaults has been added.
+Please see the corresponding change log entry [pr-13018.toml](https://github.com/Graylog2/graylog2-server/blob/master/changelog/5.1.0/pr-13018.toml)
```
bernd, line 8: Let's not link to a changelog entry that doesn't exist yet. :slightly_smiling_face: We don't know yet if the first version it will appear in is 5.1.

I appreciate the detailed changelog details, but changelog entries shouldn't be too exhaustive or be a replacement for upgrade notes or documentation. Let's put that information into the docs or this UPGRADING document. Also, we currently don't render the user details anywhere.

danotorrey:
> I might have missed the conversation, but didn't we want to prefix the default settings with `default_` and use the JadConfig fallback property to keep backward compatibility?

Thanks for mentioning this @bernd. The config library `fallbackPropertyName` capability only allows us the ability to specify a single fallback property (https://github.com/Graylog2/JadConfig/pull/113). So, we will need to coordinate and change to add `default_` or `index_init_default_` with the more general naming changes that are currently being discussed here https://github.com/Graylog2/graylog2-server/issues/13927. I suggest that we reach a renaming conclusion, then implement that in a single/standalone PR for clarity. This PR is already quite large. What do you think? If you agree, I will create an issue to track that. Looping in @todvora, since I believe he was going to work on the actual renaming change.

UPGRADING.md
danotorrey: Ahh yes, makes sense. Thanks for the guidance on that @bernd. I'll move all of that bulk detail over to `UPGRADING.md`, and coordinate any relevant changes in the docs.

Thanks also for the reminder that the user details are not rendered anywhere right now. I'll keep that in mind. Is there a plan to render those at some point (like to a linkable details page)?

bernd:
> > I might have missed the conversation, but didn't we want to prefix the default settings with `default_` and use the JadConfig fallback property to keep backward compatibility?
> 
> Thanks for mentioning this @bernd. The config library `fallbackPropertyName` capability only allows us the ability to specify a single fallback property ([Graylog2/JadConfig#113](https://github.com/Graylog2/JadConfig/pull/113)). So, we will need to coordinate and change to add `default_` or `index_init_default_` with the more general naming changes that are currently being discussed here #13927. I suggest that we reach a renaming conclusion, then implement that in a single/standalone PR for clarity. This PR is already quite large. What do you think? If you agree, I will create an issue to track that. Looping in @todvora, since I believe he was going to work on the actual renaming change.

Thanks for the update, @danotorrey. I am good with doing the renaming in a separate PR. Thank you!

UPGRADING.md
bernd: > Thanks also for the reminder that the user details are not rendered anywhere right now. I'll keep that in mind. Is there a plan to render those at some point (like to a linkable details page)?

Not sure yet. It all depends on how we are going to render the changelogs in the future.

danotorrey:
> Thanks for the update, @danotorrey. I am good with doing the renaming in a separate PR. Thank you!

Thanks @bernd. Issue opened: https://github.com/Graylog2/graylog2-server/issues/14194

UPGRADING.md
danotorrey: > Let's put that information into the docs or this UPGRADING document.

@bernd This has been done. I have paired down the change log to just the `message`, and moved everything else to the `UPGRADING` doc.

bernd: Thank you!

bernd (approved):
My changelog comments have been fixed. Thanks.

PR #13032, opened 2022-07-11
Core System Entity/Immutable Entities

roberto-graylog:
## Description

Core support for Scoped Entity Support for pitch [Support Unmodifiable Content Pack Entities](https://docs.google.com/document/d/1sTPXCLriIz7X7Sg5SYZYNUkoR3BW2beNGgJjZkYHSP8/edit#)

See corresponding Enterprise changes: https://github.com/Graylog2/graylog-plugin-enterprise/pull/3837

## Motivation and Context

Entities with a defined scope allow mutability to be specified, which can be applied to entities users should not modified (only cloned which they can then modify). 

## How Has This Been Tested?

Unit testing has been performed.

## Types of changes

## Checklist:


danotorrey:
@roberto-graylog @ryan-carroll-graylog I know we had agreed to remove the MyEntity tests, but, I was thinking about that a bit more and having second thoughts. I actually think now that those would be pretty helpful to cover the API and functionality for `ScopedEntityPaginatedDbService`. This came to mind when thinking about the preparation for getting this PR merged. Seems like it would be good to have some coverage of the core components and not just the entities that implement it. What do you think about bringing those back (reverting https://github.com/Graylog2/graylog2-server/pull/13032/commits/b5de6fd41bd241e3c063fcc14d7c91193123f6f1)?

roberto-graylog:
> @roberto-graylog @ryan-carroll-graylog I know we had agreed to remove the MyEntity tests, but, I was thinking about that a bit more and having second thoughts. I actually think now that those would be pretty helpful to cover the API and functionality for `ScopedEntityPaginatedDbService`. This came to mind when thinking about the preparation for getting this PR merged. Seems like it would be good to have some coverage of the core components and not just the entities that implement it. What do you think about bringing those back (reverting [b5de6fd](https://github.com/Graylog2/graylog2-server/commit/b5de6fd41bd241e3c063fcc14d7c91193123f6f1))?

Sure, that sounds find.

ryan-carroll-graylog:
> > @roberto-graylog @ryan-carroll-graylog I know we had agreed to remove the MyEntity tests, but, I was thinking about that a bit more and having second thoughts. I actually think now that those would be pretty helpful to cover the API and functionality for `ScopedEntityPaginatedDbService`. This came to mind when thinking about the preparation for getting this PR merged. Seems like it would be good to have some coverage of the core components and not just the entities that implement it. What do you think about bringing those back (reverting [b5de6fd](https://github.com/Graylog2/graylog2-server/commit/b5de6fd41bd241e3c063fcc14d7c91193123f6f1))?
> 
> Sure, that sounds find.

I like this Idea as well.

danotorrey:
Thanks @ryan-carroll-graylog @roberto-graylog I will work on restoring them now.

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedEntityPaginatedDbService.java
```diff
@@ -0,0 +1,102 @@
...
+
+
+    public final void ensureMutability(E entity) {
+        if (!isMutable(entity)) {
+            throw new IllegalArgumentException("Immutable entity cannot be modified");
+        }
+    }
+
+    @Override
+    public int delete(String id) {
```
bernd, line 97: Should the `#delete` method also be `final` to avoid unwanted mutation in subclasses?

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedEntityPaginatedDbService.java
```diff
@@ -0,0 +1,102 @@
...
+     * @return optional target entity
+     */
+    public final Optional<E> findMutableById(String id) {
+
+        return get(id)
+                .filter(entityScopeService::isMutable);
+    }
+
+    @Override
+    public E save(E entity) {
```
bernd, line 74: Should the `#save` method also be `final` to avoid unwanted mutation in subclasses?

bernd (changes requested):
Thank you! :+1:  First round of review:

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedEntityPaginatedDbService.java
danotorrey: I think so. What do you think @roberto-graylog?

graylog2-server/src/main/java/org/graylog2/database/entities/EntityScopeService.java
```diff
@@ -0,0 +1,73 @@
...
+        Objects.requireNonNull(scopedEntity, "Entity must not be null");
+
+        String scope = scopedEntity.scope();
+
+        // TODO: Should it fail instead?
+        if (scope == null || scope.isEmpty()) {
+            return true;
+        }
+
+        EntityScope entityScope = entityScopes.get(scope);
```
roberto-graylog, line 56: @danotorrey, what do you think about checking upper case and lower case 'scope' here.  Since we have no (good) way of enforsing that EntityScope::getName returns a name in upper case, should we make this lookup case-insensitive?

graylog2-server/src/main/java/org/graylog2/database/entities/EntityScopeService.java
```diff
@@ -0,0 +1,73 @@
...
+        return Collections.unmodifiableList(new ArrayList<>(entityScopes.values()));
+    }
+
+    public boolean isMutable(ScopedEntity scopedEntity) {
+
+        Objects.requireNonNull(scopedEntity, "Entity must not be null");
+
+        String scope = scopedEntity.scope();
+
+        // TODO: Should it fail instead?
```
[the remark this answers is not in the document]
roberto-graylog: Left it as is (option 1).

roberto-graylog:
@danotorrey @ryan-carroll-graylog :  I made the delete/save methods in `ScopedEntityPaginatedDbService` final.  This means you will have to create a delegate method of sorts (rename perhaps?) in your implementations that perform other work before calling `super.save(..)`.  If this proves to be too much of a hustle, we could extract the scope/mutability management in `ScopedEntityPaginatedDbService` to a private/protected final method.  However this would allow a child class to override  the save/delete method and bypass the mutability checks, which is what we are trying to avoid. 

Let me know what you think.

ryan-carroll-graylog:
> @danotorrey @ryan-carroll-graylog : I made the delete/save methods in `ScopedEntityPaginatedDbService` final. This means you will have to create a delegate method of sorts (rename perhaps?) in your implementations that perform other work before calling `super.save(..)`. If this proves to be too much of a hustle, we could extract the scope/mutability management in `ScopedEntityPaginatedDbService` to a private/protected final method. However this would allow a child class to override the save/delete method and bypass the mutability checks, which is what we are trying to avoid.
> 
> Let me know what you think.

Thanks for the heads up. I think it's a good Idea as you have it, and it seems like it should be pretty straightforward to update the implementations based on the ones I've been looking at.

danotorrey:
> I made the delete/save methods in `ScopedEntityPaginatedDbService` final

I like it. Any time a delegate/override is needed, it can probably be named appropriately. This should be more clear to the caller of what that method does (since it probably performs more logic in addition to the super method).

danotorrey:
Note: I have pushed additional changes needed to support setting the `_scope` flag in content packs in https://github.com/Graylog2/graylog2-server/pull/13032/commits/8321be514a2c5c6f54a1819a641756c6f9d429ba. This is needed so that the dependent PRs for specific entity implementations can begin implementing content pack support. 

Feedback on the content packs aspects support is welcome. The first usage can be found in https://github.com/Graylog2/graylog2-server/pull/13058/files#diff-80d98ce3c7615c2557822597f060058df65e5d6a545d634f2ac26f0edd5d5b19

ryan-carroll-graylog (approved):
This has really come together well! I think it's turned out to be a great approach to implement immutable entities! The Scoped Entity and DbService base classes have made it very easy to extend and implement the Immutable Entity concept with other entities, and the Entity Scope service and resource handles the scopes seamlessly.

I've been using this as the base off of which to build out the immutable lookup tables, caches and data adapters, as well as testing out the event definitions and everything seems to be working really well now that we've been through several iterations. At this point I'm good to call these core changes ready to merge.

kingzacko1 (approved):
I really like the approach we've taken to the design and implementation of this feature. Nice work everyone!

danotorrey:
@bernd When you have a chance, can you please take another look to see if the changes you requested were addressed?

danotorrey:
My only suggestion after working with the code in this PR for a while is drop the word `Entity` from the class name `ScopedEntityPaginatedDbService` to shorten it up a bit. It's definitely completely correct as-is, but the shorter version also seems ok to me, since use of an entity is implied with DB services. Thoughts @ryan-carroll-graylog @roberto-graylog @kingzacko1 @bernd?

ryan-carroll-graylog:
> My only suggestion after working with the code in this PR for a while is drop the word `Entity` from the class name `ScopedEntityPaginatedDbService` to shorten it up a bit. It's definitely completely correct as-is, but the shorter version also seems ok to me, since use of an entity is implied with DB services. Thoughts @ryan-carroll-graylog @roberto-graylog @kingzacko1 @bernd?

I do think the shorter the better especially since it's not really adding clarity. No strong preference either way though.

bernd:
> My only suggestion after working with the code in this PR for a while is drop the word `Entity` from the class name `ScopedEntityPaginatedDbService` to shorten it up a bit. It's definitely completely correct as-is, but the shorter version also seems ok to me, since use of an entity is implied with DB services. Thoughts @ryan-carroll-graylog @roberto-graylog @kingzacko1 @bernd?

@danotorrey My suggestion would be `ScopedDbService`. I think the "paginated" word doesn't need to be there. All DB services should offer pagination.

bernd (approved):
Thanks for addressing my review comments. :+1: I didn't test the PR locally, but the overall approach looks good to me now. Thank you! :pray:

danotorrey:
> My suggestion would be `ScopedDbService`

@bernd Sounds good to me.

Issue #12499, opened 2022-04-20
Indices and Index Sets pagination state is lost when navigating to a specific index


danotorrey:
## Expected Behavior

On the Indices and Index Sets page, when clicking on a specific index name to view the Details page for it, then clicking the browser back button, the pagination state in the list of indices is lost.

For example, if on the Indices and Index Sets page, I click to view the second page of indexes, then click to view the details of an index set, then click back, the fact that I was on the second page is lost, and the list displays the first page again.

Note: The pagination footer only shows if there are more than 10 index sets present.
![image](https://user-images.githubusercontent.com/3423655/164316407-55a83242-4927-447a-80ad-c9aaf912234e.png)

## Motivation

This is particularly problematic when there are a large number of indexes, and the user is attempting to manage details on many indexes. The fact that the pagination state is lost when clicking back creates extra work (they must re-navigate to the correct page each time they go back).

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

danotorrey:
This issue was found at the same time the user was reporting https://github.com/Graylog2/graylog2-server/issues/12498

Issue #12160, opened 2022-02-18
Remember how many items to show on page after user defines


tellistone:
## Expected Behavior
There is a button on pages that contain lists of elements, such as streams or dashboards, that allows the user to define how many results are shown per page. See the top right, "Show 10"

![image](https://user-images.githubusercontent.com/85873295/154700105-b5b58808-f627-4cb1-985e-2acc93287356.png)

The value set by the user should be persistant across page changes and applications restarts.

## Current Behavior
It's not persistant, which is fairly annoying.

## Possible Solution
Remember the value defined, store preferance in eg. mongodb.

Graylog 4.2.4

PR #12986, opened 2022-07-01
Fix/#12499 Pagination state saved as query parameter

gally47:
## Description

Now all paginated lists are linkable with a query parameter so the page state will never be lost when we navigate back and forth to it.

## Motivation and Context

https://github.com/Graylog2/graylog2-server/issues/12499

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #13039, opened 2022-07-12
Pagination state saved as query parameter

gally47:
## Description

Now all paginated lists are linkable with a query parameter so the page and pageSize states will not be lost when we navigate back and forth to it.

Note:
- In this PR the previous pagination solution (saved as react state) was replaced by query parameter solution.
- This feature is optional you can turn it off by setting "useQueryParameter" PaginatedList prop to false.
- "useQueryParameter" is set to false for pages with multiple PaginatedList.
- In this PR straightforward eslint issues were fixed the more complicated ones were just disabled.

## Motivation and Context

https://github.com/Graylog2/graylog2-server/issues/12160
https://github.com/Graylog2/graylog2-server/issues/12499

## Types of changes

## Checklist:


mnin:
Hey @gally47,

Looks good so far!

But since some of the same files are introduced here as in PR https://github.com/Graylog2/graylog2-server/pull/12986, do you want to use `fix-issue-12499` as a base branch here as long as it is not merged?

mnin (approved):
LGTM

gally47:
> Hey @gally47,
> 
> Looks good so far!
> 
> But since some of the same files are introduced here as in PR #12986, do you want to use `fix-issue-12499` as a base branch here as long as it is not merged?

Yes `fix-issue-12499` is used as base here, I also merged back these changes to fix the tests after your approval.

gally47:
Screenshot 2022-07-25 at 14 29 05

It will be good if the height of the menu will adjust to the number of users to avoid extra useless space.

mnin:
Some pages that use the `PaginatedList` component no longer work as expected. For example, in the Archives List (enterprise plugin). Both the search and the reset when a search has been performed lead to pages without content or without pagination elements.

graylog2-web-interface/src/components/lookup-tables/LookupTablesOverview.jsx
```diff
@@ -22,41 +22,96 @@ import Routes from 'routing/Routes';
...
+  const empty = { title: 'None' };
+
+  if (!map) {
+    return empty;
+  }
+
+  return map[id] || empty;
+};
+
+const _helpPopover = () => {
```
dennisoelkers, line 40: **Informational:** This pattern is repeated a couple of times in older code like this. The `_helpPopover` function is actually a React component in disguise. Therefore we should just treat it as one and name it `HelpPopover` and render it as a component (`<HelpPopover />`) instead of calling the function.

I am not saying you should change this in this PR, extracting a stateless function like this to the module scope is great already. But maybe if you stumble across patterns like this in the future, you can turn them into components straight-away.

graylog2-web-interface/src/components/users/UsersOverview/UsersOverview.tsx
```diff
@@ -81,27 +76,33 @@ const _loadUsers = (pagination, setLoading, setPaginatedUsers) => {
   });
 };
 
-const _updateListOnUserDelete = (perPage, query, setPagination) => UsersActions.delete.completed.listen(() => setPagination({ page: DEFAULT_PAGINATION.page, perPage, query }));
+const _updateListOnUserDelete = (callback) => UsersActions.delete.completed.listen(() => callback());
```
dennisoelkers, line 79: Can you please type `callback` here?

graylog2-web-interface/src/components/users/UsersOverview/UsersOverview.tsx
```diff
@@ -81,27 +76,33 @@ const _loadUsers = (pagination, setLoading, setPaginatedUsers) => {
...
+  useEffect(() => _loadUsers({ page, perPage, query }, setLoading, setPaginatedUsers), [page, perPage, query]);
+  useEffect(() => _updateListOnUserDelete(resetPage), [resetPage]);
+  useEffect(() => _updateListOnUserSetStatus({ page, perPage, query }, setLoading, setPaginatedUsers), [page, perPage, query]);
 
   if (!users) {
     return <Spinner />;
   }
 
-  const searchFilter = <UsersFilter onSearch={(newQuery) => setPagination({ ...pagination, query: newQuery, page: DEFAULT_PAGINATION.page })} />;
+  const handleSearch = (newQuery) => {
```
dennisoelkers, line 98: Can you please type `newQuery` here?

dennisoelkers (changes requested):
Thanks for the good work, @gally47! The actual change is very helpful, I just included some smaller nitpicks and suggestions.

One note for future PRs: You moved up stateless functions in classes to the top-level. That is great and helpful, but it is unrelated to the scope of this PR. Doing smaller refactorings in PRs is fine, but in this case it is covering up the intent of this PR volume-wise. Ideally you could create two PRs, one is just for the pagination feature, the other would contain cleanups. This would make reviews much easier because a reviewer can concentrate on the core change in an isolated context.

graylog2-web-interface/src/components/common/PaginatedList.tsx
```diff
@@ -52,38 +53,29 @@ const PaginatedList = ({
...
+  const numberPages = pageSize > 0 ? Math.ceil(totalItems / pageSize) : 0;
 
   const _onChangePageSize = (event: React.ChangeEvent<HTMLOptionElement>) => {
     event.preventDefault();
     const newPageSize = Number(event.target.value);
 
-    setCurrentPage(INITIAL_PAGE);
-    setPageSize(newPageSize);
-    onChange(INITIAL_PAGE, newPageSize);
+    if (useQueryParameter) setPageSize(newPageSize);
```
ousmaneo, line 72: I think we should avoid single line if statement just to be safe.
```suggestion
    if (useQueryParameter) { 
      setPageSize(newPageSize);
    }
```

ousmaneo (changes requested):
Thank you for this PR, LGTM!
I added some small nitpick.

dennisoelkers (approved):

ousmaneo (approved):
LGTM! thank you @gally47 !

PR #13060, opened 2022-07-15
Immutable Lookup Tables, Caches, and Data Adapters backend

ryan-carroll-graylog:
Add immutable entities support for Lookup Tables, Caches, and Data Adapters.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3901

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/lookup/db/DBCacheService.java
```diff
@@ -30,24 +31,25 @@
...
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class DBCacheService {
+public class DBCacheService extends ScopedEntityPaginatedDbService<CacheDto> {
+    public static final String COLLECTION_NAME = "lut_caches";
+    // TODO: rename this or get rid of it. Same name (db) as superclass.
```
roberto-graylog, line 43: Yes, remove.

graylog2-server/src/main/java/org/graylog2/lookup/db/DBLookupTableService.java
```diff
@@ -109,17 +114,12 @@ public Collection<LookupTableDto> findByDataAdapterIds(Collection<String> dataAd
...
-
-    public void delete(String idOrName) {
+    @Override
+    public int delete(String idOrName) {
         final Optional<LookupTableDto> lookupTableDto = get(idOrName);
-        lookupTableDto
-                .map(LookupTableDto::id)
-                .map(ObjectId::new)
-                .ifPresent(db::removeById);
+        int numDeleted = super.delete(idOrName);
```
roberto-graylog, line 120: While this is old code, relying on an exception as test (ID or name) is not a good approach.  In this case, we can use ObjectId to test whether idOrName is a valid ObjectId, if so, search by ID, else by name.

roberto-graylog (changes requested):
This looks good to me.  It just needs the latest changes from `core_scoped_db_entities` for a final test.

roberto-graylog (approved):
Looks good to me.

ryan-carroll-graylog:
> Looks good to me.

Thanks for all the feedback.

roberto-graylog (approved):

danotorrey (changes requested):
@ryan-carroll-graylog This LGTM and tests successfully! There is only one small method override that needs to be added in order for the Illuminate installation to automatically inject the scope:

This method override will need to added to all of the facades, so that the injector below will know which entity facades the `ILLUMINATE` scope can be injected for:
```
@Override
    public boolean usesScopedEntities() {
        return true;
    }
```

https://github.com/Graylog2/graylog-plugin-enterprise/blob/923d2d3dbdfde70d7bffd987ea32a1cb84ff7afb/enterprise/src/main/java/org/graylog/plugins/illuminate/installation/ContentPackScopeInjector.java#L46-L50

ryan-carroll-graylog:
> @ryan-carroll-graylog This LGTM and tests successfully! There is only one small method override that needs to be added in order for the Illuminate installation to automatically inject the scope:
> 
> This method override will need to added to all of the facades, so that the injector below will know which entity facades the `ILLUMINATE` scope can be injected for. Without this, the `DEFAULT` scope is used for Illuminate-installed content packs.
> 
> ```
> @Override
>     public boolean usesScopedEntities() {
>         return true;
>     }
> ```
> 
> https://github.com/Graylog2/graylog-plugin-enterprise/blob/923d2d3dbdfde70d7bffd987ea32a1cb84ff7afb/enterprise/src/main/java/org/graylog/plugins/illuminate/installation/ContentPackScopeInjector.java#L46-L50

@danotorrey I was just looking at this. Do these LUT entities need this even though they'll be installed outside the Illuminate content-pack pipeline, in the new code for @kingzacko1 changes?

I suppose it's probably good form to have it anyway for consistency in case we decide to use this elsewhere.

ryan-carroll-graylog:
> > @ryan-carroll-graylog This LGTM and tests successfully! There is only one small method override that needs to be added in order for the Illuminate installation to automatically inject the scope:
> > This method override will need to added to all of the facades, so that the injector below will know which entity facades the `ILLUMINATE` scope can be injected for. Without this, the `DEFAULT` scope is used for Illuminate-installed content packs.
> > ```
> > @Override
> >     public boolean usesScopedEntities() {
> >         return true;
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://github.com/Graylog2/graylog-plugin-enterprise/blob/923d2d3dbdfde70d7bffd987ea32a1cb84ff7afb/enterprise/src/main/java/org/graylog/plugins/illuminate/installation/ContentPackScopeInjector.java#L46-L50
> 
> @danotorrey I was just looking at this. Do these LUT entities need this even though they'll be installed outside the Illuminate content-pack pipeline, in the new code for @kingzacko1 changes?
> 
> I suppose it's probably good form to have it anyway for consistency in case we decide to use this elsewhere.

Just added the overrides of usesScopedEntities() in the facades.

kingzacko1 (approved):
LGTM

danotorrey:
> I was just looking at this. Do these LUT entities need this even though they'll be installed outside the Illuminate content-pack pipeline, in the new code for @kingzacko1 changes?

@ryan-carroll-graylog Thanks @ryan-carroll-graylog Good point, sorry, I had gotten confused about this for a bit there. You are correct about that. They are not strictly needed. I agree though, adding them anyhow just in case sounds good to me.

danotorrey (approved):

PR #13062, opened 2022-07-15
Updated GeoIpResolverEngine to use all the fields in org.graylog.sche…

roberto-graylog:
…ma that have an _ip suffix rather than just the three previously defined IP fields.

## Description

During the updates to the GeoIpResolver engine that enabled support for both IpInfo and MaxMind, only three fields for hard-coded to search when the user opts to enforce the Graylog Schema.  However, there are other fields that may hold IP addresses define in the schema.  

This code change generates a list of all schema fields with a _ip suffix, as the schema package defines the fields in a non-iterable way.

## Motivation and Context

When working with customer reported issues, it was noted they are using schema fields for IP addresses other than the currently defined in the resolver engine.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java
@@ -19,7 +19,6 @@ package org.graylog.plugins.map.geoip;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.net.InetAddresses;
 import org.apache.commons.lang3.StringUtils;
 import org.graylog.plugins.map.config.GeoIpResolverConfig;
@@ -33,7 +32,9 @@ import java.net.InetAddress;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.codahale.metrics.MetricRegistry.name;
 
@@ -51,11 +52,7 @@ public class GeoIpResolverEngine {
      * that will be inserted into the message.
      * </p>
      */
-    private final Map<String, String> ipAddressFields = new ImmutableMap.Builder<String, String>()
-            .put("source_ip", "source")
-            .put("host_ip", "host")
-            .put("destination_ip", "destination")
-            .build();
+    private final Map<String, String> ipAddressFields = buildSchemaIpFieldMap();
 
     private final GeoIpResolver<GeoLocationInformation> ipLocationResolver;
     private final GeoIpResolver<GeoAsnInformation> ipAsnResolver;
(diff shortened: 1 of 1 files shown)


kingzacko1 (approved):

bernd:
> This looks good to me. I am going to ping @bernd in case there's may be a more graceful way to capture these IP fields and then @waab76 and @kroepke so they are aware of the change.

There is currently no existing mechanism to get the list of fields. We could create something with Guice where we register the IP fields so we can inject them. (e.g., with a `@Named` annotation) But that means the bindings also need to be maintained. :smile:

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java
```diff
@@ -51,11 +52,7 @@ public class GeoIpResolverEngine {
      * that will be inserted into the message.
      * </p>
      */
-    private final Map<String, String> ipAddressFields = new ImmutableMap.Builder<String, String>()
-            .put("source_ip", "source")
-            .put("host_ip", "host")
-            .put("destination_ip", "destination")
-            .build();
+    private final Map<String, String> ipAddressFields = buildSchemaIpFieldMap();
```
danotorrey, line 55:
```suggestion
private static final String[] IP_FIELDS = {
            DestinationFields.DESTINATION_IP,
            DestinationFields.DESTINATION_NAT_IP,
            EventFields.EVENT_OBSERVER_IP,
            SourceFields.SOURCE_IP,
            SourceFields.SOURCE_IPV6,
            SourceFields.SOURCE_NAT_IP,
            NetworkFields.NETWORK_FORWARDED_IP,
            AssociatedFields.ASSOCIATED_IP,
            HostFields.HOST_IP,
            HostFields.HOST_IPV6,
            VendorFields.VENDOR_PRIVATE_IP,
            VendorFields.VENDOR_PRIVATE_IPV6,
            VendorFields.VENDOR_PUBLIC_IP,
            VendorFields.VENDOR_PUBLIC_IPV6};
    private static final Map<String, String> ipAddressFields = Stream.of(IP_FIELDS)
            .collect(Collectors.toMap(e -> e, mapFieldNameToPrefix()));
```

danotorrey (approved):
Looks good to me. Adding one small suggestion to define the list of IP fields as a static constant, which might make it a bit more visible when the class is maintained later.

waab76:
When we first decided to add the `Enforce Graylog Schema` option to the Geo-IP processor, it was a deliberate decision not to include coverage for every field that might have an IP address. The focus was on providing geolocation data for IP addresses likely to be outside the user's own network. Also, we wanted to limit the number of lookups in order to improve the processor's performance.

The Graylog Schema doesn't currently support geo fields for the [Vendor entity](https://schema.graylog.org/en/development/schema/entities/vendor.html), the [Event entity](https://schema.graylog.org/en/development/schema/entities/event.html) or the [Associated entity](https://schema.graylog.org/en/development/schema/entities/associated.html).

I think @miwent really ought to weigh in on which of these additional schema fields are likely to contain IPs worth doing geolocation lookups.

miwent:
> 

I recommend adding the fields `event_observer_ip` and `network_forwarded_ip`, `source_nat_ip`, and `destination_nat_ip`. The `associated` fields aren't used in the pipelines, they are generated at index time by field mapping templates with the `copy_to` functionality.

danotorrey:
@miwent @waab76 Thanks for the feedback. So, it sounds like this PR should only add the 4 fields that @miwent mentioned, and the other fields should be removed. Is that correct?

danotorrey (approved):
The new field list LGTM. Thanks @roberto-graylog!

PR #13067, opened 2022-07-18
Detect fresh Graylog installations

mpfz0r:
This introduces a new named boolean that can be used to detect whether a fresh installation
of Graylog is happening.
It can be injected with: `@Named("isFreshInstallation")`

Knowing that a new installation is happening allows us to introduce breaking changes.

### Caveat
The detection code is running together with the preflight checks.
So it won't work if someone enabled `skip_preflight_checks`.

### Other changes
Refactor MongoDBPreflightCheck so it can be used more than once.

Don't allow fresh installations with the disk journal on an NFS volume.
This fixes an open TODO in the `DiskJournalPreflightCheck`


graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/MongoDBPreflightCheck.java
```diff
@@ -43,48 +46,66 @@ public class MongoDBPreflightCheck implements PreflightCheck {
...
+    public boolean dbIsEmpty() {
+        try {
+            final Boolean isEmpty = getDbVersionAndStatus(true).getValue();
+            return isEmpty;
+        } catch (ExecutionException | RetryException e) {
+            return false;
+        }
+    }
+
+    private ImmutablePair<Version, Boolean> getDbVersionAndStatus(boolean runSilent) throws ExecutionException, RetryException {
```
kodjo-anipah, line 77: Not sure if I would not split it, right now org.graylog2.bootstrap.ServerBootstrap#runPreFlightChecks will exectue this register a fresh registration without caring for the verison, afterwards its exectued as part of the preflight check, which only asserts the version but not if its empty

mpfz0r: I agree, this isn't ideal. Let me refactor that.

kodjo-anipah (approved):
LGTM

PR #13087, opened 2022-07-20
Fix field value suggestions when quoting is used.

dennisoelkers:
## Motivation and Context

This PR is fixing an issue with the field value completions not working when the user is starting a field value with quoting, e.g. when the user types `field:""`, no completions appear.

This PR is fixing this and also retains quoting information, so when the user uses quoting, the value returned is not escaped, if the user is not using quoting, escaping is used.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me!

PR #13088, opened 2022-07-20
Adjust server startup to log all throwables to the log file

bernd:
Previously all Errors (Exceptions are handled) haven't been logged to
the log file but only STDERR. Not logging these errors can make
debugging harder if STDERR output and the log files are configured to
be in different places. (e.g., operating system packages with systemd)


mpfz0r (approved):
LGTM

PR #13091, opened 2022-07-20
Create query validation provider

maxiadlovskii:
## Description
We need this provider to be able to add different callbacks for adding parameter success action. In the future we can use this provider to handle other cases

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/searchbar/queryvalidation/QueryValidationContext.ts
```diff
@@ -0,0 +1,25 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+
+import { createContext, useContext } from 'react';
+
+export const QueryValidationContext = createContext(undefined);
+export const QueryValidationProvider = QueryValidationContext.Provider;
+
+export const useQueryValidationContext = <T>() => {
+  return useContext<T>(QueryValidationContext);
```
linuspahl, line 24: Since we only use this context in a plugin we should not store in in the core repo.

graylog2-web-interface/src/views/components/searchbar/queryvalidation/QueryValidationContext.ts
```diff
@@ -0,0 +1,25 @@
...
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+
+import { createContext, useContext } from 'react';
+
+export const QueryValidationContext = createContext(undefined);
+export const QueryValidationProvider = QueryValidationContext.Provider;
+
+export const useQueryValidationContext = <T>() => {
```
linuspahl, line 23: Since we only use this context in a plugin we should store it in the plugin.

linuspahl (changes requested):

graylog2-web-interface/src/views/components/searchbar/queryvalidation/QueryValidationContext.ts
maxiadlovskii: I store it there because I wrap with it QueryValidation component, also that is the reason for a naming.

linuspahl: The rendering and consumption of the context happens in the plugin, right? We usually store code in a plugin if we only execute it in a plugin.

linuspahl:
We can close this PR because we moved the logic into a plugin (https://github.com/Graylog2/graylog-plugin-enterprise/pull/3824).

PR #13092, opened 2022-07-20
Handle slashes in unquoted/escaped field values for completions.

### LINKED ISSUE #13082: Field value suggestions stop after escaped character entered in partial value
## Expected Behavior

If I am working on a search query and am relying on field value suggestions to complete the query, I would expect suggestions to continue to occur if all characters in the partial value are valid (i.e. normal or escaped characters).

## Current Behavior

Currently, all values after an escaped character are treated as the start of a new keyword, with field name suggestions kicking in.

For example:  If I've typed a query `source_type:foo\/b` I would expect to see value suggestions like `foo/bar` and `foo/baz`.  Instead, I begin seeing suggestions for field names containing `b`, such as `job`, `http_request_bytes`, etc.

1. Have a field (e.g. `source_type`) whose values can be suggested, and whose available values contain a field requiring escaping (e.g. `foo/bar`, `foo/baz`).
2. Write a query based upon the prior field and value prefix (e.g. `source_type:foo\/`).
3. Type a character (e.g. `b`) immediately after the escaped character, and note that you see field name, rather than value, suggestions.

## Context

I was writing a query and wanted to drill down into possible values using a prefix that included escaped characters.  There are other ways to find that info, which I did, but it would've been nice if the field value suggestions could've handled this scenario.

## Your Environment

* Graylog Version:  4.3.3

dennisoelkers:
## Motivation and Context

**Note:** This requires #13087 to merged before.

Prior to this PR, getting suggestions for a field value containing a slash did not work. This is related due to the tokenization logic of the lucene mode in the ace editor, which is treating it as a token separator, as it could potentially be part of a regex.

This PR is now doing the following to fix this:

  - copy the existing lucene mode to allow customization
  - including `\` and `/` in `term` tokens
  - extend identifier regex which is used in ace to determine the boundary of the current input which is replaced by the selected completion with `\` and `/`
  - escape completion captions if in unquoted context so matching completions are not filtered out

The drawback of this approach is that we now have to maintain the lucene mode ourselves, but due to the fact that it was not updated in the last years, the additional maintenance overhead will hopefully be small/close to zero.

Fixes #13082.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/SearchBarAutocompletions.ts
+++ b/graylog2-web-interface/src/views/components/searchbar/SearchBarAutocompletions.ts
@@ -47,6 +47,7 @@ export type CompleterContext = Readonly<{
 export interface Completer {
   getCompletions(context: CompleterContext): Array<CompletionResult> | Promise<Array<CompletionResult>>;
   shouldShowCompletions?: (currentLine: number, lines: Array<Array<Line>>) => boolean;
+  identifierRegexps?: RegExp[];
 }
 
 const onCompleterError = (error: Error) => {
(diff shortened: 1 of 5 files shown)


linuspahl (approved):
Looks good to me, could not find a problem!

PR #13093, opened 2022-07-20
Added additional validation of email configuration

### LINKED ISSUE #13085: Add validation for non-null sender email address in notifications
Right now it is possible to set up a email notification with no valid sender email address, either in the configuration or the notification settings. This leads to an exception when sending the validation.

* Graylog Version: 4.3.3

moesterheld:
Adds validation of email configuration in graylog.conf to EventNotificationResource.

## Description
Checks that email transport is enabled and that there is a sender set either in graylog.conf or the notification itself.

closes #13085

## Motivation and Context
If sender isn't set in configuration and notification, a NPE is produced when trying to send the notification.

## Types of changes

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/rest/EventNotificationsResource.java
+++ b/graylog2-server/src/main/java/org/graylog/events/rest/EventNotificationsResource.java
@@ -29,10 +29,12 @@ import org.graylog.events.audit.EventsAuditEventTypes;
 import org.graylog.events.notifications.DBNotificationService;
 import org.graylog.events.notifications.NotificationDto;
 import org.graylog.events.notifications.NotificationResourceHandler;
+import org.graylog.events.notifications.types.EmailEventNotificationConfig;
 import org.graylog.security.UserContext;
 import org.graylog2.alarmcallbacks.EmailAlarmCallback;
 import org.graylog2.audit.jersey.AuditEvent;
 import org.graylog2.audit.jersey.NoAuditEvent;
+import org.graylog2.configuration.EmailConfiguration;
 import org.graylog2.database.PaginatedList;
 import org.graylog2.plugin.alarms.callbacks.AlarmCallback;
 import org.graylog2.plugin.configuration.ConfigurationRequest;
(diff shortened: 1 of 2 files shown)


mpfz0r (changes requested):
This looks good,
but the frontend handles the validation errors poorly,
if you can't see the error on the form, the Update button just seems to be doing nothing.
I think we could involve someone from the frontend to improve this.

Also, I think it makes sense to enforce a valid email address for the sender field.

graylog2-server/src/main/java/org/graylog/events/rest/EventNotificationsResource.java
```diff
@@ -142,13 +152,43 @@ public Response update(@ApiParam(name = "notificationId") @PathParam("notificati
...
+            } else {
+                if (!isNullOrEmpty(emailEventNotificationConfig.sender())) {
+                    try {
+                        InternetAddress email = new InternetAddress(emailEventNotificationConfig.sender());
+                        email.validate();
+                    } catch (AddressException e) {
+                        validationResult.addError("sender", "Invalid email address.");
+                    }
+                }
+                if (!isNullOrEmpty(emailConfiguration.getFromEmail())) {
```
mpfz0r, line 180: I wouldn't validate things from the `server.conf` here. The user might not be able to fix that himself.
We could add a validator to the `EmailConfiguration` class though.

mpfz0r (changes requested):
I'm happy with the frontend solution.
One more comment.

graylog2-server/src/main/java/org/graylog/events/rest/EventNotificationsResource.java
mpfz0r: Or maybe it's okay to validate the serve.conf sender email here, but only if the one from the eventnotification is empty.

moesterheld: That was also my first implementation. 
However, I found out that EmailFactory will always try to set the address from graylog.conf first if it's not empty. If it's not a valid address, sending fails with a AddressException. That's why I changed it to always validate.
The other option would be to change the logic of EmailSender altogether.

mpfz0r: hmm, okay. i think that warrants a validation then. Let's keep it as is :+1:

mpfz0r (approved):

gally47 (approved):
Frontend LGTM

PR #13095, opened 2022-07-20
Immutable content pack support for Event Definitions

danotorrey:
## Overview
Add immutable Illuminate scope support for Event Definition content packs.

These changes were pushed to a new PR, so the full change-set can be reviewed, but this should not be merged. We will want to port the commits up-stream once the team has given feedback in this approach (I wanted to avoid changing many PRs before the approach is certain).

See corresponding enterprise changes in https://github.com/Graylog2/graylog-plugin-enterprise/pull/3858

## Testing
This can be tested with the following Illuminate bundle, which contains a single spotlight pack with an Event Definition within the content pack. 

[test-bundle.zip](https://github.com/Graylog2/graylog2-server/files/9154167/test-bundle.zip)

The bundle sources are here: https://github.com/Graylog2/graylog-project-illuminate/tree/test-day-bundle/dev 

You can adjust the contents of the bundle if needed and rebuild it with the instructions here: 
https://github.com/Graylog2/graylog-project-illuminate/blob/test-day-bundle/dev/bundle-build-instructions.md

## Draft code
This code was put together quickly for feedback on the overall approach. There is lots of room for improvement on the code as need as unit test need to be added. All feedback is welcome, and I will incorporate it into the final PRs.


graylog2-server/src/main/java/org/graylog/events/processor/DBEventDefinitionService.java
```diff
@@ -65,13 +65,23 @@ public EventDefinitionDto saveWithOwnership(EventDefinitionDto eventDefinitionDt
         return dto;
     }
 
-    public int deleteAndUnregister(String id) {
+    public int deleteAndUnregister(String id, boolean overrideMutable) {
```
danotorrey, line 68: @roberto-graylog @ryan-carroll-graylog Do you have any opinion if I should break out a separate method `deleteAndUnregisterMutable` and avoid the boolean parameter here? I was not certain if one is better than the other.

ryan-carroll-graylog: I'm in favor of breaking it out to make it more explicit and less likely to accidently use the wrong way. Plus that's the convention of the superclass.

danotorrey: Thanks for the advice @ryan-carroll-graylog. Will do.

danotorrey: Thanks again @ryan-carroll-graylog. Done in https://github.com/Graylog2/graylog2-server/pull/13095/commits/468e761780205ea17edbaaf5d615c9aeffeb1e29. Let me know if you have any feedback on the implementation.

I like the API better this way (vs the previous boolean approach). It's more expressive and preserves the existing `delete` API method signatures (which we typically want to do in core.).

PR #13099, opened 2022-07-21
skip index rotation on empty index

### LINKED ISSUE #13028: Index Time rotation strategy: Add option to skip rotation on empty index sets
## What?

When using the _Index Time_ rotation strategy, we should add an option to skip rotation when the index set contains 0 messages.

## Why?

Especially with Illuminate we see a lot of index sets being created, but not all of them might be used. When using a time based rotation (e.g. daily), this can lead to huge number of empty indexes / shards being created, which require system resources.

The _Index Message Count_ and _Index Size_ rotation strategies naturally do not have this problem.

## Your Environment

* Graylog Version: 4.3

kodjo-anipah:
## Description

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3869
When using the Index Time rotation strategy, we should add an option to skip rotation when the index set contains 0 messages.

## Motivation and Context

Especially with Illuminate we see a lot of index sets being created, but not all of them might be used. When using a time based rotation (e.g. daily), this can lead to huge number of empty indexes / shards being created, which require system resources.
fixes #13028 
The Index Message Count and Index Size rotation strategies naturally do not have this problem.
## How Has This Been Tested?

Locally by configuring time based rotation for an index set
Case 1. There are messages in any index of the index sets, hence rotation is performed
Case 2. There are no messages in the whole index set, hence rotation is skipped 
Case 3. if rotateEmptyIndexSets is enabled, behaviour stays as before

Screenshot 2022-07-26 at 15 04 57

## Types of changes

## Checklist:


boosty:
Hi @kodjo-anipah, thanks for working on this!

A few points from my side:
1. Is it safe to skip the rotation by just checking the number of messages in the current index? When I wrote https://github.com/Graylog2/graylog2-server/issues/13028, I suggested to only skip when the whole index set is empty, to not confuse the _retention_. This is because the retention is currently based on index count, so it cannot deal with randomly missing indexes within an index set. 
2. I think we should turn on this new option by default. I cannot think about a good reason for why one might want to rotate an unused index set.

graylog2-web-interface/src/components/indices/rotation/TimeBasedRotationStrategyConfiguration.jsx
```diff
@@ -109,6 +133,18 @@ class TimeBasedRotationStrategyConfiguration extends React.Component {
...
                bsStyle={this._validationState()}
                required />
+        <Input id="rotate-empty-index-sets"
+               labelClassName="col-sm-3"
+               wrapperClassName="col-sm-9"
+               label="Empty index sets">
+          <Input id="rotate-empty-index-sets-checkbox"
+                 type="checkbox"
+                 ref={(rotateEmptyIndexSetsRef) => { this.inputs.rotate_empty_index_sets = rotateEmptyIndexSetsRef; }}
+                 label="Rotate empty index sets"
```
boosty, line 143:
```suggestion
                 label="Rotate empty index set"
```

graylog2-web-interface/src/components/indices/rotation/TimeBasedRotationStrategyConfiguration.jsx
```diff
@@ -109,6 +133,18 @@ class TimeBasedRotationStrategyConfiguration extends React.Component {
...
+                 onChange={this._onRotateEmptyIndexSetsUpdate('rotate_empty_index_sets')}
+                 checked={rotateEmptyIndexSets}
+                 help="Apply the rotation strategy on empty index sets." />
```
boosty, line 146:
```suggestion
                 help="Apply the rotation strategy even when all indexes are empty (not recommended)." />
```

graylog2-web-interface/src/components/indices/rotation/TimeBasedRotationStrategyConfiguration.jsx
```diff
@@ -136,14 +136,14 @@ class TimeBasedRotationStrategyConfiguration extends React.Component {
...
-                 ref={(rotateEmptyIndexSetsRef) => { this.inputs.rotate_empty_index_sets = rotateEmptyIndexSetsRef; }}
-                 label="Rotate empty index sets"
-                 onChange={this._onRotateEmptyIndexSetsUpdate('rotate_empty_index_sets')}
-                 checked={rotateEmptyIndexSets}
-                 help="Apply the rotation strategy on empty index sets." />
+                 ref={(rotateEmptyIndexSetRef) => { this.inputs.rotate_empty_index_set = rotateEmptyIndexSetRef; }}
+                 label="Rotate empty index set"
+                 onChange={this._onRotateEmptyIndexSetUpdate('rotate_empty_index_set')}
+                 checked={rotateEmptyIndexSet}
+                 help="Apply the rotation strategy even when all indexes are empty (not recommended)." />
```
boosty, line 146:
```suggestion
                 help="Apply the rotation strategy even when the index set is empty (not recommended)." />
```

Sorry for this new suggestion again, but after reading my previous one again, I think this one fits much better.

gally47 (approved):
Frontend LGTM!

boosty:
What is the default value for this config option? As we are not recommending to rotate empty index sets, the option should be disabled by default.

kodjo-anipah:
> What is the default value for this config option? As we are not recommending to rotate empty index sets, the option should be disabled by default.

Default value is set to false and rotation on empty index sets are disabled.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedRotationStrategy.java
```diff
@@ -58,6 +61,7 @@
 public class TimeBasedRotationStrategy extends AbstractRotationStrategy {
     private static final Logger log = LoggerFactory.getLogger(TimeBasedRotationStrategy.class);
     public static final String NAME = "time";
+    public static final String ELASTICSEARCH_MAX_WRITE_INDEX_AGE_OVERRIDES_CONFIGURED_PERIOD = "(elasticsearch_max_write_index_age overrides configured period)";
```
mpfz0r, line 64: Could we give this constant a shorter name?
How about `OVERRIDE_HINT`

mpfz0r (changes requested):
A few comments

mpfz0r (approved):

PR #13104, opened 2022-07-21
Implement new GRN type `search_filter` in frontend.

linuspahl:
## Motivation and Context

When a user shares a search filter with other users they will appear on their "Shared Entities" overview on the user details page. With this PR we generate the correct route for the search filter details page.


linuspahl:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/13109

PR #13106, opened 2022-07-21
Add service to monitor changes to GeoIP database files

### LINKED ISSUE #13075: GeoIP message processor doesn't reload database files
## Expected Behavior

The GeoIP message processor ([GeoIpResolverEngine](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpResolverEngine.java) class) should reload the database files when they change.

## Current Behavior

The `GeoIpResolverEngine` loads the database files only once at startup and doesn't re-open the database handles when a database file gets updated.

## Possible Solution

Introduce a reload mechanism similar to how [MaxMindDataAdapter#refresh](https://github.com/Graylog2/graylog2-server/blob/c16b8df17df58a0937f4c240bc4aa1b8da450e68/graylog2-server/src/main/java/org/graylog/plugins/map/geoip/MaxmindDataAdapter.java#L121-L147) works.

The code should also handle in-place database file updates correctly and gracefully handle any errors. (see Context below)

1. Configure the "Geo-Location Processor" on System/Configurations with a database file and enable it
2. Check that messages with IP address fields have GeoIP fields
3. Copy a new GeoIP version of the database file to the existing location
4. Observe the exception (see Context below)

## Context

@drewmiranda-gl noticed the missing reload behavior while updating GeoIP databases. The code raised the following exception when a database file got replaced with the `cp` command. We should handle such cases and reload the database handle.

```
2022-07-18T14:19:58.734-05:00 WARN  [MaxMindIpResolver] Error getting ASN for IP Address '/<redacted>'. The MaxMind DB file's data section contains bad data: pointer larger than the database.
com.maxmind.db.InvalidDatabaseException: The MaxMind DB file's data section contains bad data: pointer larger than the database.
	at com.maxmind.db.Decoder.decode(Decoder.java:81) ~[graylog.jar:?]
	at com.maxmind.db.Decoder$1.load(Decoder.java:75) ~[graylog.jar:?]
	at com.maxmind.db.NoCache.get(NoCache.java:19) ~[graylog.jar:?]
	at com.maxmind.db.Decoder.decode(Decoder.java:111) ~[graylog.jar:?]
	at com.maxmind.db.Decoder.decode(Decoder.java:87) ~[graylog.jar:?]
	at com.maxmind.db.Reader.resolveDataPointer(Reader.java:252) ~[graylog.jar:?]
	at com.maxmind.db.Reader.get(Reader.java:150) ~[graylog.jar:?]
	at com.maxmind.geoip2.DatabaseReader.get(DatabaseReader.java:182) ~[graylog.jar:?]
	at com.maxmind.geoip2.DatabaseReader.asn(DatabaseReader.java:261) ~[graylog.jar:?]
	at org.graylog.plugins.map.geoip.MaxMindIpAsnResolver.doGetGeoIpData(MaxMindIpAsnResolver.java:48) [graylog.jar:?]
	at org.graylog.plugins.map.geoip.GeoIpResolver.getGeoIpData(GeoIpResolver.java:64) [graylog.jar:?]
	at org.graylog.plugins.map.geoip.GeoIpResolverEngine.addGeoIpDataIfPresent(GeoIpResolverEngine.java:128) [graylog.jar:?]
	at org.graylog.plugins.map.geoip.GeoIpResolverEngine.filter(GeoIpResolverEngine.java:102) [graylog.jar:?]
	at org.graylog.plugins.map.geoip.processor.GeoIpProcessor.process(GeoIpProcessor.java:94) [graylog.jar:?]
	at org.graylog2.shared.buffers.processors.ProcessBufferProcessor.handleMessage(ProcessBufferProcessor.java:158) [graylog.jar:?]
	at org.graylog2.shared.buffers.processors.ProcessBufferProcessor.dispatchMessage(ProcessBufferProcessor.java:128) [graylog.jar:?]
	at org.graylog2.shared.buffers.processors.ProcessBufferProcessor.onEvent(ProcessBufferProcessor.java:98) [graylog.jar:?]
	at org.graylog2.shared.buffers.processors.ProcessBufferProcessor.onEvent(ProcessBufferProcessor.java:49) [graylog.jar:?]
	at com.lmax.disruptor.WorkProcessor.run(WorkProcessor.java:143) [graylog.jar:?]
	at com.codahale.metrics.InstrumentedThreadFactory$InstrumentedRunnable.run(InstrumentedThreadFactory.java:66) [graylog.jar:?]
	at java.lang.Thread.run(Thread.java:829) [?:?]
```

## Your Environment

* Graylog Version: 4.3.3+86369d3 (4.3 and earlier)

roberto-graylog:
closes #13075 

## Description

This code change adds a service to monitor changes to the GeoIP database files such that the `GeoIpProcessor` can rebuild the `GeoIpResolverEngine` then such changes are detected.

## Motivation and Context

Changing the GeoIP database files on the file system causes errors as `GeoIpProcessor` continues to run with the old files.

End-to-end testing.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
@@ -22,10 +22,19 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
+import javax.annotation.Nullable;
+import javax.validation.constraints.Min;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
 @JsonAutoDetect
 @JsonIgnoreProperties(ignoreUnknown = true)
 @AutoValue
 public abstract class GeoIpResolverConfig {
+
+    private static final int DEFAULT_INTERVAL_MINUTES = 10;
+    public static final String FIELD_REFRESH_INTERVAL_UNIT = "refresh_interval_unit";
+    public static final String FIELD_REFRESH_INTERVAL = "refresh_interval";
 
     @JsonProperty("enabled")
     public abstract boolean enabled();
@@ -42,29 +51,50 @@ public abstract class GeoIpResolverConfig {
     @JsonProperty("asn_db_path")
     public abstract String asnDbPath();
 
+    @JsonProperty(FIELD_REFRESH_INTERVAL_UNIT)
+    @Nullable
+    public abstract TimeUnit refreshIntervalUnit();
+
+    @JsonProperty(FIELD_REFRESH_INTERVAL)
+    @Min(0)
+    public abstract long refreshInterval();
+
+    public Duration refreshIntervalAsDuration() {
+        if (refreshIntervalUnit() == null) {
+            return Duration.ofMinutes(DEFAULT_INTERVAL_MINUTES);
+        }
+        return Duration.ofMillis(refreshIntervalUnit().toMillis(refreshInterval()));
+    }
+
     @JsonCreator
     public static GeoIpResolverConfig create(@JsonProperty("enabled") boolean cityEnabled,
                                              @JsonProperty("enforce_graylog_schema") boolean enforceGraylogSchema,
                                              @JsonProperty("db_vendor_type") DatabaseVendorType databaseVendorType,
                                              @JsonProperty("city_db_path") String cityDbPath,
-                                             @JsonProperty("asn_db_path") String asnDbPath) {
+                                             @JsonProperty("asn_db_path") String asnDbPath,
+                                             @JsonProperty(FIELD_REFRESH_INTERVAL_UNIT) TimeUnit refreshIntervalUnit,
+                                             @JsonProperty(FIELD_REFRESH_INTERVAL) long refreshInterval) {
         return builder()
                 .enabled(cityEnabled)
                 .enforceGraylogSchema(enforceGraylogSchema)
                 .databaseVendorType(databaseVendorType == null ? DatabaseVendorType.MAXMIND : databaseVendorType)
                 .cityDbPath(cityDbPath)
                 .asnDbPath(asnDbPath)
+                .refreshIntervalUnit(refreshIntervalUnit)
+                .refreshInterval(refreshInterval)
                 .build();
     }
 
     public static GeoIpResolverConfig defaultConfig() {
-       return builder()
-               .enabled(false)
-               .databaseVendorType(DatabaseVendorType.MAXMIND)
-               .enforceGraylogSchema(false)
-               .cityDbPath("/etc/graylog/server/GeoLite2-City.mmdb")
-               .asnDbPath("/etc/graylog/server/GeoLite2-ASN.mmdb")
-               .build();
+        return builder()
+                .enabled(false)
+                .databaseVendorType(DatabaseVendorType.MAXMIND)
+                .enforceGraylogSchema(false)
+                .cityDbPath("/etc/graylog/server/GeoLite2-City.mmdb")
+                .asnDbPath("/etc/graylog/server/GeoLite2-ASN.mmdb")
+                .refreshIntervalUnit(TimeUnit.MINUTES)
+                .refreshInterval(DEFAULT_INTERVAL_MINUTES)
+                .build();
     }
 
     public static Builder builder() {
@@ -84,6 +114,10 @@ public abstract class GeoIpResolverConfig {
         public abstract Builder cityDbPath(String dbPath);
 
         public abstract Builder asnDbPath(String asnDBPath);
+
+        public abstract Builder refreshIntervalUnit(TimeUnit unit);
+
+        public abstract Builder refreshInterval(long interval);
 
         public abstract GeoIpResolverConfig build();
     }
--- a/graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
@@ -23,7 +23,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.auto.value.AutoValue;
 
 import javax.annotation.Nullable;
-import javax.validation.constraints.Min;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 
@@ -32,9 +31,10 @@ import java.util.concurrent.TimeUnit;
 @AutoValue
 public abstract class GeoIpResolverConfig {
 
-    private static final int DEFAULT_INTERVAL_MINUTES = 10;
+    private static final Long DEFAULT_INTERVAL = 10L;
     public static final String FIELD_REFRESH_INTERVAL_UNIT = "refresh_interval_unit";
     public static final String FIELD_REFRESH_INTERVAL = "refresh_interval";
+    public static final TimeUnit DEFAULT_INTERVAL_UNIT = TimeUnit.MINUTES;
 
     @JsonProperty("enabled")
     public abstract boolean enabled();
@@ -56,12 +56,11 @@ public abstract class GeoIpResolverConfig {
     public abstract TimeUnit refreshIntervalUnit();
 
     @JsonProperty(FIELD_REFRESH_INTERVAL)
-    @Min(0)
-    public abstract long refreshInterval();
+    public abstract Long refreshInterval();
 
     public Duration refreshIntervalAsDuration() {
         if (refreshIntervalUnit() == null) {
-            return Duration.ofMinutes(DEFAULT_INTERVAL_MINUTES);
+            return Duration.ofMinutes(DEFAULT_INTERVAL);
         }
         return Duration.ofMillis(refreshIntervalUnit().toMillis(refreshInterval()));
     }
@@ -73,15 +72,15 @@ public abstract class GeoIpResolverConfig {
                                              @JsonProperty("city_db_path") String cityDbPath,
                                              @JsonProperty("asn_db_path") String asnDbPath,
                                              @JsonProperty(FIELD_REFRESH_INTERVAL_UNIT) TimeUnit refreshIntervalUnit,
-                                             @JsonProperty(FIELD_REFRESH_INTERVAL) long refreshInterval) {
+                                             @JsonProperty(FIELD_REFRESH_INTERVAL) Long refreshInterval) {
         return builder()
                 .enabled(cityEnabled)
                 .enforceGraylogSchema(enforceGraylogSchema)
                 .databaseVendorType(databaseVendorType == null ? DatabaseVendorType.MAXMIND : databaseVendorType)
                 .cityDbPath(cityDbPath)
                 .asnDbPath(asnDbPath)
-                .refreshIntervalUnit(refreshIntervalUnit)
-                .refreshInterval(refreshInterval)
+                .refreshIntervalUnit(refreshIntervalUnit == null ? DEFAULT_INTERVAL_UNIT : refreshIntervalUnit)
+                .refreshInterval(refreshInterval == null ? DEFAULT_INTERVAL : refreshInterval)
                 .build();
     }
 
@@ -92,8 +91,8 @@ public abstract class GeoIpResolverConfig {
                 .enforceGraylogSchema(false)
                 .cityDbPath("/etc/graylog/server/GeoLite2-City.mmdb")
                 .asnDbPath("/etc/graylog/server/GeoLite2-ASN.mmdb")
-                .refreshIntervalUnit(TimeUnit.MINUTES)
-                .refreshInterval(DEFAULT_INTERVAL_MINUTES)
+                .refreshIntervalUnit(DEFAULT_INTERVAL_UNIT)
+                .refreshInterval(DEFAULT_INTERVAL)
                 .build();
     }
 
@@ -117,7 +116,7 @@ public abstract class GeoIpResolverConfig {
 
         public abstract Builder refreshIntervalUnit(TimeUnit unit);
 
-        public abstract Builder refreshInterval(long interval);
+        public abstract Builder refreshInterval(Long interval);
 
         public abstract GeoIpResolverConfig build();
     }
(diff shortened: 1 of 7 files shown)


roberto-graylog:
@danotorrey , @bernd .  I implemented this as a separate service to try to avoid a trouble some condition wherein the refresh operation was being called multiple times.  I realized the obvious (after the fact) that this was due to mulltiple `GoeIpProcessor`s being created by Guice.  

If you think this is an overkill, I can move it back to the `GeoIpProcessor`.

graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx
```diff
@@ -17,19 +17,22 @@
...
 import { Field, Form, Formik } from 'formik';
 
-import { IfPermitted, Select } from 'components/common';
+import { IfPermitted, Select, TimeUnitInput } from 'components/common';
 import { Button, Col, Input, Modal, Row } from 'components/bootstrap';
 import FormikInput from 'components/common/FormikInput';
 import { DocumentationLink } from 'components/support';
 
 export type GeoVendorType = 'MAXMIND' | 'IPINFO'
+export type TimeUnit = 'SECONDS' | 'MINUTES' | 'HOURS' | 'DAYS'
```
roberto-graylog, line 26: @danotorrey Should we include less/more units--or for that matter, limit at all?

danotorrey:
> I implemented this as a separate service to try to avoid a trouble some condition wherein the refresh operation was being called multiple times. I realized the obvious (after the fact) that this was due to mulltiple `GoeIpProcessor`s being created by Guice.

@roberto-graylog Nice! This approach sounds great to me. Good call. Given those multiple processors, I like this idea of having a single service control that signals the processors to initiate refreshes. Reviewing this now...

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpDbFileChangedEvent.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @JsonProperty
+    public abstract Instant timestamp();
+
+    @JsonProperty
+    @NotEmpty
+    public abstract String nodeId();
+
+    @JsonProperty
+    public abstract Map<DatabaseType, FileInfo.Change> changes();
```
danotorrey, line 44: It looks like these properties are not referenced once set. Are they needed?

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpDbFileChangeMonitorService.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(GeoIpDbFileChangeMonitorService.class.getSimpleName());
+
+    private ScheduledFuture<?> refreshTask;
+
+    private final ScheduledExecutorService scheduler;
+    private final GeoIpResolverConfigValidator geoIpResolverConfigValidator;
+
+    private final ClusterConfigService clusterConfigService;
+    private final ClusterEventBus clusterEventBus;
```
danotorrey, line 58:
```suggestion
    private final EventBus eventBus;
```
I am not certain, but _think_ the `EventBus` should be used here instead of the `ClusterEventBus`. I believe `ClusterEventBus` will broadcast the changed event to all Graylog nodes when a GeoIp the file is changed on just one node. I think we want the event to only be broadcast to all processors running on the same-node, since the file change only applies to the current node.

danotorrey (commented):
@roberto-graylog I was able to review most of the code, but I will need to resume the review and testing on Monday. The approach and implementation are looking good to me so far. I added a couple points for your review/consideration.

graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx
danotorrey: This list sounds good to me.

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpDbFileChangedEvent.java
roberto-graylog: Removed properties.

graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
```diff
@@ -42,29 +51,50 @@ public abstract class GeoIpResolverConfig {
...
+                                             @JsonProperty(FIELD_REFRESH_INTERVAL_UNIT) TimeUnit refreshIntervalUnit,
+                                             @JsonProperty(FIELD_REFRESH_INTERVAL) long refreshInterval) {
         return builder()
                 .enabled(cityEnabled)
                 .enforceGraylogSchema(enforceGraylogSchema)
                 .databaseVendorType(databaseVendorType == null ? DatabaseVendorType.MAXMIND : databaseVendorType)
                 .cityDbPath(cityDbPath)
                 .asnDbPath(asnDbPath)
+                .refreshIntervalUnit(refreshIntervalUnit)
+                .refreshInterval(refreshInterval)
```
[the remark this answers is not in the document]
roberto-graylog: Updated.  I assume you meant:

```Java
 .refreshIntervalUnit(refreshIntervalUnit == null ? TimeUnit.MINUTES : refreshIntervalUnit)
  .refreshInterval(refreshInterval == null ? DEFAULT_INTERVAL_MINUTES : refreshInterval)
```
That is, setting to the actual argument if not null, rather than null.

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpDbFileChangeMonitorService.java
roberto-graylog: Updated to use EventBus.

graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx
```diff
@@ -175,6 +180,17 @@ const GeoIpResolverConfig = ({ config = defaultConfig, updateConfig }: Props) =>
...
                                name="asn_db_path" />
+                  <TimeUnitInput label="Database refresh interval"
+                                 update={(value, unit) => {
+                                   setFieldValue('refresh_interval', value);
+                                   setFieldValue('refresh_interval_unit', unit);
+                                 }}
+                                 help="Interval at which the database files are checked for modifications and refreshed changes are detected on disk."
+                                 value={values.refresh_interval}
+                                 unit={values.refresh_interval_unit || 'MINUTES'}
+                                 defaultEnabled={values.enabled}
```
ryan-carroll-graylog, line 191: I'm having the issue where the time selector is always disabled in the UI but adding this got it working (Disclaimer: not a UI expert)
```suggestion
                                 defaultEnabled={values.enabled}
                                 enabled={values.enabled}
```

roberto-graylog: Thanks.  I had forgotten to test that.  `defaultEnabled` is just, well, the default. So toggling the enabled checkbox requires the other attributes.

graylog2-server/src/main/java/org/graylog/plugins/map/config/GeoIpResolverConfig.java
```diff
@@ -22,11 +22,20 @@
...
 
+import javax.annotation.Nullable;
+import javax.validation.constraints.Min;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
 @JsonAutoDetect
 @JsonIgnoreProperties(ignoreUnknown = true)
 @AutoValue
 public abstract class GeoIpResolverConfig {
```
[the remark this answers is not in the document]
roberto-graylog: Made recommended changes.  It should work now with null interval fields.  Thanks.

ryan-carroll-graylog:
This all looks really good and seems like a smart approach. It looks to be working well in testing: when `GeoIpResolverConfig` is updated the `GeoIpResolverEngine` is rebuilt.

I was just re-reading the original issue and (I might be misreading) it sounds like they want to be able to just copy over the database .mmdb files in place without changing the config. Do you think it would we should also listen for a file change event (similar to what we're doing in `DevelopmentDirectoryObserverService`) in the `GeoIpDbFileChangeMonitorService`, in addition to the current `ClusterConfigChangedEvent`s?

roberto-graylog:
> This all looks really good and seems like a smart approach. It looks to be working well in testing: when `GeoIpResolverConfig` is updated the `GeoIpResolverEngine` is rebuilt.
> 
> I was just re-reading the original issue and (I might be misreading) it sounds like they want to be able to just copy over the database .mmdb files in place without changing the config. Do you think it would we should also listen for a file change event (similar to what we're doing in `DevelopmentDirectoryObserverService`) in the `GeoIpDbFileChangeMonitorService`, in addition to the current `ClusterConfigChangedEvent`s?

Yes, it should be able to handle it.  The method `GeoIpDbFileChangeMonitorService::refreshDatabases`, which gets scheduled per the interval specified in the config, checks for changes in the files--if so, it fires the event `GeoIpDbFileChangedEvent`.

The other event, ClusterConfigChangedEvent, is also received by the monitor so that it can get file changes as well as any schedule changes.

ryan-carroll-graylog:
> > This all looks really good and seems like a smart approach. It looks to be working well in testing: when `GeoIpResolverConfig` is updated the `GeoIpResolverEngine` is rebuilt.
> > I was just re-reading the original issue and (I might be misreading) it sounds like they want to be able to just copy over the database .mmdb files in place without changing the config. Do you think it would we should also listen for a file change event (similar to what we're doing in `DevelopmentDirectoryObserverService`) in the `GeoIpDbFileChangeMonitorService`, in addition to the current `ClusterConfigChangedEvent`s?
> 
> Yes, it should be able to handle it. The method `GeoIpDbFileChangeMonitorService::refreshDatabases`, which gets scheduled per the interval specified in the config, checks for changes in the files--if so, it fires the event `GeoIpDbFileChangedEvent`.
> 
> The other event, ClusterConfigChangedEvent, is also received by the monitor so that it can get file changes as well as any schedule changes.
> 
> If you replace or touch a file (e.g. `touch /etc/graylog/server/artifacts_standard_location.mmdb`) you will note that the resolvers will get updated--if you have DEBUG enabled.

Ah ok perfect, I missed that but I do see now. That makes sense and I think is a good approach so were not firing off a ton of updates unnecessarily, thanks for explaining that.

ryan-carroll-graylog (approved):

graylog2-server/src/main/java/org/graylog/plugins/map/geoip/GeoIpDbFileChangeMonitorService.java
```diff
@@ -0,0 +1,186 @@
...
+ * <p>
+ * If the database files have changed, a {@link GeoIpDbFileChangedEvent} is posted, to which {@link org.graylog.plugins.map.geoip.processor.GeoIpProcessor} subscribes and reloads the {@link GeoIpResolverEngine}.
+ * </p>
+ *
+ * <p>
+ * This service also subscribes to {@link ClusterConfigChangedEvent} to update the database files to be monitored, as well as to update the scheduled task ({@link #refreshDatabases()}) which checks for file changes.
+ * </p>
+ */
+@Singleton
+public final class GeoIpDbFileChangeMonitorService {
```
danotorrey, line 56: The logic in this class looks good and tests successfully, but I think it would be beneficial to convert it to extend AbstractIdleService, which allows it to have `startUp` and `shutDown` lifecycle methods. The main benefit here is that the `registerToEvents()/updateConfiguration()` methods could then be called in the `startUp` method instead of in the constructor. Since the constructors are called in the process on initializing Guice bindings, this isolates the startup logic from the binding init (which prevents any exceptions that might happen in those methods from preventing Guice bindings from completing).

Similar example: https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/messages/StreamCacheService.java#L41-L67

The `eventBus.unregister(this)` could also be called in `shutDown` to properly unregister it.

I am not certain of this is really required, but the future for the update could also be checked and unscheduled in the `shutDown` method too. I guess this is only really helpful if the future is a long-running operation (or one that could hang). Perhaps this is good to have here, since it involves disk I/O.
https://github.com/Graylog2/graylog-plugin-enterprise/blob/56dd7b641b1c7ba9fbef3ec017a13ab0ac0792e1/enterprise/src/main/java/org/graylog/plugins/license/EnterpriseLicenseChecker.java#L108-L110

graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx
```diff
@@ -175,6 +180,18 @@ const GeoIpResolverConfig = ({ config = defaultConfig, updateConfig }: Props) =>
...
+                                 enabled={values.enabled}
+                                 hideCheckbox
+                                 units={['SECONDS', 'MINUTES', 'HOURS', 'DAYS']} />
```
danotorrey, line 194: I think it would also be helpful to display the `Database refresh interval` field on the main config page, so folks can see if it is set from the main Config page (without editing it).

image

https://github.com/Graylog2/graylog2-server/blob/00371b4a4a3d3f84c5241eb0ec78a873dd0e60a3/graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx#L108-L115

danotorrey (commented):
@roberto-graylog I finished up the review today. This tests successfully. Well done! I left two more points for your consideration. But, other than those, I think this is ready.

ryan-carroll-graylog (approved):
Retested with the latest changes and everything looks good!

kingzacko1 (approved):
Tested changing DB files on master branch and was able to get the database pointer error, pulled these changes in and was able to successfully perform geo IP lookups after updating the DB files. Works well.

graylog2-web-interface/src/components/maps/configurations/GeoIpResolverConfig.tsx
kingzacko1: Super nitpicky and not important to fix for this particular PR, but with the addition of the longer config names the spacing in this plugin config UI seems to have strayed from the typical inline formatting of the other plugins. @mikedklein maybe something frontend folks could clean up when there's time.
image

danotorrey (approved):
LGTM! Thanks a lot for this PR @roberto-graylog!

PR #13113, opened 2022-07-22
Reduce the bulk indexing request size for 429 errors too

mpfz0r:
An elastic cluster might not have enough memory available to
accept a full bulk index request. It can reject it with
a HTTP/1.1 429 Too Many Requests exception.

We could handle this the same way we do with Request entity too large
errors. See https://github.com/Graylog2/graylog2-server/pull/7071


graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/testing/ElasticsearchInstanceSmallES7.java
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.storage.elasticsearch7.testing;
+
+import org.graylog2.storage.SearchVersion;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+
+public class ElasticsearchInstanceSmallES7 extends ElasticsearchInstanceES7 {
```
dennisoelkers, line 23: Instead of creating a new subclass, we could also extend the `ElasticsearchInstanceES7` (same for ES6 of course) class to accept a parameter that allow customizing the memory limits. That would avoid a bit of boiler plate and makes it a bit cleaner, I think.

dennisoelkers (commented):
Looks great so far! Just added one suggestion inline.

graylog-storage-elasticsearch7/src/test/java/org/graylog/storage/elasticsearch7/testing/ElasticsearchInstanceSmallES7.java
mpfz0r: Agreed. I didn't like to having to duplicate this entire chain of static methods either.

todvora (approved):
LGTM, thanks! Tested locally and it seems to work correctly.

PR #13121, opened 2022-07-25
Block create/update of view if forbidden search filters are used

luk-kaminski:
## Description
Block create/update of a view if forbidden/hidden search filters are used.

## Motivation and Context
When it comes to our UI, front-end will implement proper blocking.

The problem is that someone can still try to invoke creation/update of a view directly through our REST API.
This person can manipulate Search JSON so that it contains a Search Filter current user has no permissions for.
In this scenario creation/update should fail.

## How Has This Been Tested?
Manually and with unit tests.

## Types of changes

## Checklist:


todvora (approved):
As discussed, I'd suggest to include the verification logic in the SearchUser class and its interfaces, but we can do that in a follow-up PR.

PR #13147, opened 2022-07-27
Scope lut caches data adapters frontend

simonychuang:
## Description

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3901

## Types of changes

## Checklist:


danotorrey:
I believe we will need to hide Illuminate-scoped adapter and caches on the Create/Edit lookup table page. These lookup tables are considered to be read-only, and can come/go with Illuminate installations, so I think we will want to prevent the user from creating new caches/adapters that have the Illuminate scope. Pinging @kingzacko1 in case you have more info on this. I can't remember if this is for-sure the decision we reached.

image

kingzacko1:
@danotorrey that's correct. The list will need to be filtered of any Illuminate caches/data adapters and we may want to also look at a strictly backend check so that calls directly from the API don't complete successfully if they reference an Illuminate scoped entity.

simonychuang:
> For Illuminate-scoped entities, it looks like the Actions are showing as disabled/greyed out for adapters, but are not showing for caches and lookup tables. Is this intentional?
> 
> image
> 
> image
> 
> image

@danotorrey I believe they should be consistent, for the data adapters I made the buttons disabled but still showing because we discussed at the beginning of the cycle that simply hiding the buttons may be a confusing user experience. So I think we should do the same for caches and lookup tables too. @zeeklop @mikedklein what do you think?

zeeklop:
@simonychuang @danotorrey Regarding hiding or disabling the buttons on the list or any other view, I personally lean towards hiding them (in this case not getting mounted on the view) A disabled button can be enabled using the browser dev tools. A button that isn't there can't be added. This is just a personal preference, I'm good with whatever you guys decided. This isn't critical since the API also is checking for the scope permissions.

danotorrey:
@zeeklop @simonychuang I think I had originally suggested disabling the buttons, but, I was checking, and it looks like there might already be a convention in Graylog for hiding buttons (see the Create Stream button below). So, perhaps it might be best to consider hiding the buttons for consistency.

https://github.com/Graylog2/graylog2-server/blob/12e87a3e45727bf14c6b8faef3c9f27b10ad128c/graylog2-web-interface/src/pages/StreamsPage.jsx#L75-L80

graylog2-web-interface/src/components/lookup-tables/Cache.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      expect(baseElement.querySelector('button[alt="edit button"]')).toBeVisible();
+    });
+  });
+
+  it('should not show "edit" button', async () => {
+    const { baseElement } = renderedCache('ILLUMINATE');
+
+    await waitFor(() => {
+      expect(baseElement.querySelector('button[alt="edit button"]')).toBeNull();
+    });
```
linuspahl, line 69: In cases like this one you do not need to use `waitFor`, you can just use `await expect`.
As far as I know, `waitFor(() => expect` is only necessary when waiting for a mock to have been called.

graylog2-web-interface/src/components/lookup-tables/Cache.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  return (
+    <Row className="content">
+      <Col md={12}>
+        <h2>{cache.title} <small>({plugin.displayName})</small></h2>
+        <SummaryContainer>
+          <SummaryRow>
+            <Title>Description:</Title>
+            <Value>{cache.description || <em>No description.</em>}</Value>
```
linuspahl, line 65: In cases like this one we usually use the nullish coalescing operator. In this specific case it will probably never make a difference, but it is a good practice to make sure we would e.g. render the description when it is `0`.

graylog2-web-interface/src/components/lookup-tables/LookupTableView.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { render, waitFor } from 'wrappedTestingLibrary';
+import { BrowserRouter as Router } from 'react-router-dom';
+
+import { TABLE, CACHE, DATA_ADAPTER, mockedUseScopePermissions } from './fixtures';
+import LookupTableView from './LookupTableView';
+
+jest.mock('hooks/useScopePermissions', () => (mockedUseScopePermissions));
+
+const renderedLUT = (scope: string) => {
+  TABLE._scope = scope;
```
linuspahl, line 27: Let's create new object here, to avoid manipulating the reference.

graylog2-web-interface/src/components/lookup-tables/DataAdapter.tsx
```diff
@@ -0,0 +1,121 @@
...
+type Props = {
+  dataAdapter: LookupTableAdapter,
+};
+
+const DataAdapter = ({ dataAdapter }: Props) => {
+  const [lookupKey, setLookupKey] = React.useState(null);
+  const [lookupResult, setLookupResult] = React.useState(null);
+  const { loadingScopePermissions, scopePermissions } = useScopePermissions(dataAdapter);
+
+  const _onChange = (event) => {
```
linuspahl, line 40: You did it in most cases, but we try to always type the function parameters, if they can not be inferred. Current `event` is implicitly `any`.

By typing it we can activate the TS config option `noImplicitAny` at some point.

linuspahl (commented):
It is great that you’ve migrated so many files to TypeScript! I just saw that the changes contain 25x the `any` type.  We try to avoid using `any` to be able to make the TS config stricter at some point.

  I always struggled with this when doing a bigger refactoring, because I thought it is necessary to created perfect and complete types, which can be reused in the future, but in most cases it is already a big improvement to just type what is actually being used the related function.

For example in [this case](https://github.com/Graylog2/graylog2-server/pull/13147/files#diff-581cf79ba87e8c681f740310c0ae39cf412e55187c4e66cfe05e7f266fc8fce0R66), you could just type `metrics` like this: `Array<{ acc: number, nodeId: string }>`, instead of using `any`.

graylog2-web-interface/src/components/lookup-tables/Cache.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <Router><Cache cache={CACHE} /></Router>,
+  );
+};
+
+describe('Cache', () => {
+  it('should show "edit" button', async () => {
+    const { baseElement } = renderedCache('DEFAULT');
+
+    await waitFor(() => {
+      expect(baseElement.querySelector('button[alt="edit button"]')).toBeVisible();
```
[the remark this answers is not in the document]
zeeklop: Thanks Linus, I tried many combinations and all of them were failing to find the element in the DOM. Curious though, if you search for the button in this way `screen(getByRole('button', { name: 'edit' }))` it doesn't find, neither this other way `screen(getByRole('button', { name: /edit/ }))` The key is making the name case-insensitive? ¯\\\_(ツ)\_/¯

graylog2-web-interface/src/components/lookup-tables/Cache.tsx
```diff
@@ -0,0 +1,78 @@
...
+type Props = {
+  cache: LookupTableCache,
+};
+
+const Cache = ({ cache }: Props) => {
+  const history = useHistory();
+  const { loadingScopePermissions, scopePermissions } = useScopePermissions(cache);
+  const plugins = {};
+
+  PluginStore.exports('lookupTableCaches').forEach((p: any) => {
```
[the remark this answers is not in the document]
zeeklop: Excellent, I copy this from the previous implementation. I'll update to use the hook everywhere I missed it.

graylog2-web-interface/src/components/lookup-tables/Cache.tsx
zeeklop: What if the description is an empty string?

kingzacko1:
Looks like a lot of extraneous commits have been picked up in this PR. Once it gets to a good spot, it may be worthwhile to create a new branch off of Ryan's backend branch and then cherry pick all commits from this feature on top of it.

graylog2-web-interface/src/components/lookup-tables/Cache.test.tsx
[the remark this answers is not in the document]
linuspahl: Do you expect the testing library to find the button, because the `name` attribute has the value `edit`?
With the described query it will consider the accessible names, for example the text of the button, the title or the aria-label, because they actually represent what the user is interacting with.

Since the button text is `Edit` we either need to make the query case-insensitive, or use `{ name: 'Edit' }`. I like `{ name: /edit/i }`, because of the flexibility. For example if we add an icon for the button (as a child element) the query `{ name: 'Edit' }` will fail.

Usually I use the `Testing Playground` browser extension when I struggle with finding the best query.

graylog2-web-interface/src/components/lookup-tables/Cache.tsx
linuspahl: Good point, I think ideally the backend should not return an empty string, but in this case I would also just use `||`.

graylog2-web-interface/src/components/lookup-tables/Cache.test.tsx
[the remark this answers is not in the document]
zeeklop: Yes, I thought that `{ name: ... }` was specifically referring to the name attribute of the element. Thank you Linus

ryan-carroll-graylog:
> I believe we will need to hide Illuminate-scoped adapter and caches on the Create/Edit lookup table page. These lookup tables are considered to be read-only, and can come/go with Illuminate installations, so I think we will want to prevent the user from creating new caches/adapters that have the Illuminate scope. Pinging @kingzacko1 in case you have more info on this. I can't remember if this is for-sure the decision we reached.
> 
> image

@simonychuang @zeeklop @danotorrey @kingzacko1 This is looking good! I did notice testing with the latest it looks like Illuminate-scoped Caches and Data Adapters are still selectable when creating/editing a Lookup Table. Is that something we still want to do in this PR?

Also as @kingzacko1 mentioned above is it also worth adding checks in the backend so that saving a Lookup Table that references an Illuminate scoped Cache/Data Adapter fails?

Everything else is looking great! So maybe it's better to get these merged and change the above in other PRs?

zeeklop:
@ryan-carroll-graylog 
> This is looking good! I did notice testing with the latest it looks like Illuminate-scoped Caches and Data Adapters are still selectable when creating/editing a Lookup Table.

> Is that something we should change in other PRs?

This PR got really large, lots of components were modified and new test added. I think it would be better to handle this modification in a different PR to also add some tests and so.

ryan-carroll-graylog:
> @ryan-carroll-graylog
> 
> > This is looking good! I did notice testing with the latest it looks like Illuminate-scoped Caches and Data Adapters are still selectable when creating/editing a Lookup Table.
> 
> > Is that something we should change in other PRs?
> 
> This PR got really large, lots of components were modified and new test added. I think it would be better to handle this modification in a different PR to also add some tests and so.

Yeah I think that's good plan so we can get these merged.

graylog2-web-interface/src/components/common/TimeUnitInput.jsx
```diff
@@ -79,8 +79,10 @@ export const extractDurationAndUnit = (duration, timeUnits) => {
  * value.
  */
 const TimeUnitInput = createReactClass({
+  // eslint-disable-next-line
```
linuspahl, line 82: It looks like you are using `// eslint-disable-next-line` a few times without specifying a rule. Let's always specify the rule to ensure we.are not disabling every rule for this line and also to make it more obvious if we may can fix the rule easily.

graylog2-web-interface/src/components/lookup-tables/fixtures.ts
```diff
@@ -0,0 +1,58 @@
+/*
```
linuspahl, line 1: Usually the fixtures are not part of the `src` directory, they are located at `tests/fixtures`.

graylog2-web-interface/src/components/lookup-tables/fixtures.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const TABLE: LookupTable = {
+  id: '62a9e6bdf3d7456348ef8e53',
+  title: 'Watchlist (Internal - Do not remove)',
+  description: 'A basic description',
+  name: 'watchlist',
+  _scope: null,
+};
+
+export const CACHE: LookupTableCache = {
```
linuspahl, line 32: It is a nitpick, but let's unify the naming with other fixtures by writing it lower case. Since fixtures are always kind of .. fixed we never started writing the names uppercase.

graylog2-web-interface/src/components/lookup-tables/CacheTableEntry.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      scopePermissions: null,
+    });
+
+    renderedCTE('DEFAULT');
+
+    expect(screen.queryByRole('button', { name: /edit/i })).not.toBeInTheDocument();
+    expect(screen.getByText('Loading...')).toBeInTheDocument();
+  });
+
+  it('should show "edit" button', async () => {
```
linuspahl, line 53: It looks like these calls do not have to be `async`.

graylog2-web-interface/src/components/lookup-tables/CacheTableEntry.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    },
+  );
+};
+
+describe('CacheTableEntry', () => {
+  it('should show Loading spinner while loading scope permissions', async () => {
+    asMock(mockedUseScopePermissions).mockReturnValueOnce({
+      loadingScopePermissions: true,
+      scopePermissions: null,
+    });
```
linuspahl, line 45: This seems to work, but I can recommend a slightly cleaner approach:

It was confusing for me that the `fixtures` file contains a complete mock for the implementation of `useScopePermissions`.

Instead I would just mock `useScopePermissions` like this:
`jest.mock('hooks/useScopePermissions')`

and implement
```
beforeEach(() => {
  asMock(useScopePermissions).mockReturnValue(scopePermissions);
})
```
in this case `scopePermissions` is a fixture.

In the test you can then do
```
asMock(useScopePermissions).mockReturnValue({
  loadingScopePermissions: true,
  scopePermissions: null,
});
```
There are a few benefits
- `asMock(useScopePermissions).mockReturnValue` is easier to read
- by replacing `mockReturnValueOnce ` with `mockReturnValue` it behaves more like in the real world, because the state does not change just because another render happened. This way it is in general also easier to test something, because the state / return value is consistent.
- with the `beforeEach` part you can be sure that you reset the return value before every test.

graylog2-web-interface/src/components/lookup-tables/DataAdapter.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      type: 'csvfile',
+      displayName: 'CSV File',
+      summaryComponent: CSVFileAdapterSummary,
+    },
+  ],
+}));
+
+const renderedDataAdapter = (scope: string) => {
+  const auxDataAdapter = { ...DATA_ADAPTER };
+  auxDataAdapter._scope = scope;
```
linuspahl, line 39: It is a nitpick, but you can just write
```
const auxDataAdapter = {
  ...DATA_ADAPTER,
  _scope: scope,
  config: ...
};
```

graylog2-web-interface/src/components/lookup-tables/CacheTableEntry.test.tsx
linuspahl: Alright, I am personally a big fan of using `await findByRole`, since it also fails if the element does not exist and it still works in case the component includes async code in the future. No need to change it for this PR, it is just a personal preference.

graylog2-web-interface/src/components/lookup-tables/DataAdapterTableEntry.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return render(
+    <Router>
+      <DataAdapterTableEntry adapter={auxDataAdapter} error={null} />
+    </Router>,
+    {
+      container: document.body.appendChild(document.createElement('table')),
+    },
+  );
+};
```
linuspahl, line 40: There should be no need to render `Router`, since it is already part of the `DefaultProviders` we use for `render` from `wrappedTestingLibrary`.

linuspahl: I think the following approach is a little bit easier to read:
```
return render(
  <table>
    <DataAdapterTableEntry adapter={auxDataAdapter} error={null} />
  </table>,
);
```

(in this case and in the following cases which are similar)

graylog2-web-interface/src/components/lookup-tables/LookupTableView.tsx
```diff
@@ -0,0 +1,175 @@
...
+                   onChange={handleInputOnChange}
+                   help="Key to look up a value for."
+                   value={lookupKey.value} />
+            <Button type="submit" bsStyle="success" disabled={!lookupKey.valid}>Look up</Button>
+          </fieldset>
+        </form>
+        {lookupResult && (
+          <div style={{ marginTop: '16px' }}>
+            <h4>Lookup result</h4>
+            <pre>{JSON.stringify(lookupResult, null, 2)}</pre>
```
linuspahl, line 167: Maybe let's memorise the result of `JSON.stringify`?

zeeklop: I think this a good call. We can do that once we move to react-query. Right now this value is stored in a state since is the result of a query request where the user can input a key_name to get some data back.

mikedklein (approved):
From a front end POV it looks as though all of @linuspahl requests have been addressed.

linuspahl:
I just noticed the PR contains the old import path for `usePluginEntities` (`views/logic/usePluginEntities`), we adjusted this recently https://github.com/Graylog2/graylog2-server/pull/13270

Let's adjust this before we merge this PR.

PR #13148, opened 2022-07-28
Allow obtaining all triggers for one job definition ID

todvora:
So far our DBJobTriggerService allowed to get only one trigger for one job definition ID. This PR adds a call to obtain all triggers for the job ID.

## Description
Adding new method called `getAllForJob`, returning a list of available triggers. Additionally renaming `getForJob` to `getOneForJob`, returning an `Optional`, either empty or with a value. This method still throws an `IllegalStateException` if there are more triggers, in the same way the original call did.

## Motivation and Context
The reporting migration to our generic scheduler needs more triggers for one job. One report is represented by one job, each delivery configuration is represented by its own trigger.

## How Has This Been Tested?
Added unit tests, adapted existing

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionHandler.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionHandler.java
@@ -334,7 +334,7 @@ public class EventDefinitionHandler {
     }
 
     private Optional<JobTriggerDto> getJobTrigger(JobDefinitionDto jobDefinition) {
-        final List<JobTriggerDto> jobTriggers = jobTriggerService.getForJob(jobDefinition.id());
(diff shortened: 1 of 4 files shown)


todvora:
@bernd @mpfz0r I tried to make the getOne/getAll approach more explicit without actually cutting too deep and changing half of the world. Let me know if this is a good first step or if you have any better suggestions. Thanks!

bernd (approved):
> I tried to make the getOne/getAll approach more explicit without actually cutting too deep and changing half of the world. Let me know if this is a good first step or if you have any better suggestions. Thanks!

@todvora Sounds good! Thank you! :+1:

todvora:
Thanks for your review @bernd !

PR #13151, opened 2022-07-28
Add styles for different sizes of nav component

maxiadlovskii:
## Description
This PR add styles for .nav-sm, .nav-xs and .nav-lg classes of Nav component. Now we can add props bsSize: sm | xs | lg | small | large to Nav component and we will have different sizes for navigation items

## Motivation and Context
According to react-bootstrap documentation https://react-bootstrap-v3.netlify.app/components/navs/  we can add bsStyle one of: "lg", "large", "sm", "small" to change the size of the Navs. That doesn't work and also looks like a mistake in documentation because as bsStyle we use `pills`or  `tabs` to change the style for them.  
Adding bsSize props to `Nav` (like we usually do for other components) component adds an appropriate class to the component. But there are no written styles for these classes. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/bootstrap/Nav.jsx
+++ b/graylog2-web-interface/src/components/bootstrap/Nav.jsx
@@ -63,6 +63,28 @@ const Nav = styled(BootstrapNav)(({ theme }) => css`
         }
       }
     }
+    
+    &.nav-sm {
+      > li > a {
+        padding: 8px 12px;
+        font-size: 12px;
+      }
+    };
+
+    &.nav-xs {
+      > li > a {
+        padding: 2px 6px;
+        font-size: 12px;
+      }
+    };
+
+    &.nav-lg {
+      > li > a {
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/bootstrap/Nav.jsx
```diff
@@ -63,6 +63,28 @@ const Nav = styled(BootstrapNav)(({ theme }) => css`
...
+      }
+    };
+
+    &.nav-lg {
+      > li > a {
+        padding: 14px 18px;
+        font-size: 18px;
+        line-height: 1.3;
+      }
+    };
```
linuspahl, line 87: I thought about using the theme font sizes here, but currently we are not using them for e.g. the button component. I think it is more important to unify the different font-sizes for the different bootstrap sizes, than always using the theme styles.

linuspahl (approved):

PR #13160, opened 2022-07-29
Set `X-Frame-Options` header, disallow embedding by default.

dennisoelkers:
## Motivation and Context

This PR is adding a configuration option `http_allow_embedding`, which is configuring if embedding the web frontend (or any other API request) is allowed or not. It defaults to `false`, disallowing embedding to prevent clickjacking. It can be set to `true`, which would allow embedding "as long as the site including it in a frame is the same as the one serving the page."[*]

This is a short-term solution to prevent clickjacking. It should be replaced at some point with a complete content security policy, particularly because the `X-Frame-Options` header is deprecated by the corresponding CSP directive.

Fixes Graylog2/graylog-plugin-enterprise#3884

[*]: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#directives> "MDN Docs for `X-Frame-Options`"

## Types of changes

## Checklist:

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -7,6 +7,20 @@ Please make sure to create a MongoDB database backup before starting the upgrade
 
 ## Breaking Changes
 
+## Disallowing embedding the frontend by default
+
+To prevent [click-jacking](https://developer.mozilla.org/en-US/docs/Web/Security/Types_of_attacks#click-jacking), we are now preventing the frontend to be embedded in `<frame>`/`<iframe>`/etc. elements by sending the `X-Frame-Options`-header with all HTTP responses. The header value depends on the new configuration setting `http_allow_embedding`. The different combinations are:
+
+| `http_allow_embedding` | `X-Frame-Options`-header value |
+|------------------------|--------------------------------|
+| not set                | `DENY`                         |
+| `false`                | `DENY`                         |
+| `true`                 | `SAMEORIGIN`                   |
+
+If you want to be able to embed the Graylog frontend in another HTML page, you most probably want to set `http_allow_embedding` to `true`. Only do this if you are aware of the implications!
+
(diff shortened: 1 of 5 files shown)


bernd (commented):
@dennisoelkers The MDN docs on [X-Frame-Options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options) say that the option is obsoleted by [Content-Security-Policy=frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors). When looking at the browser compatibility table, the main reason for using `X-Frame-Options` would be IE 8 support.

Should we use the `C-S-P=frame-ancestors` header instead?

dennisoelkers:
@bernd: This is what I am referring to in the PR body:

> This is a short-term solution to prevent clickjacking. It should be replaced at some point with a complete content security policy, particularly because the X-Frame-Options header is deprecated by the corresponding CSP directive.

I think we should use the `X-Frame-Options`-header for now, because its implications are clear. Implementing a CSP would involve a bit more work and testing, but is beneficial and should be done soon to supersede this.

bernd:
> @bernd: This is what I am referring to in the PR body:
> 
> > This is a short-term solution to prevent clickjacking. It should be replaced at some point with a complete content security policy, particularly because the X-Frame-Options header is deprecated by the corresponding CSP directive.
> 
> I think we should use the `X-Frame-Options`-header for now, because its implications are clear. Implementing a CSP would involve a bit more work and testing, but is beneficial and should be done soon to supersede this.

@dennisoelkers Doh! Sorry for not reading properly... :confused: 

Can you open an issue for CSP if we don't have one yet? Thank you!

bernd (approved):
Thank you! :pray:

Issue #13043, opened 2022-07-13
Dots in field names are replaced silently


mpfz0r:
The fact that we don't support dots in field names is nothing new.
This code has been there since ages, silently replacing dots with underscores once a Message gets written to ES/OS:

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/plugin/Message.java#L404-L406

What is problematic, is that we neither have documentation for this behavior, nor do we log a warning if we replace dots.

The fact that field renaming happens very late in the processing can lead to some confusion when writing pipeline rules or extractors.
The ingested Message will show up with underscore fields in the search, but within processing you need to work with dots.

## Possible solutions
 - Add documentation
 - Show a system notification if we detect that changing field names was needed. (This needs to be done efficiently to not cause a performance degradation)
 - ~Change the field names directly when assigning them~ (This will break existing rules and might confuse users who are already accustomed to the existing behavior)

Refs:
https://github.com/Graylog2/graylog2-server/issues/12990
https://github.com/Graylog2/graylog2-server/issues/6588
https://github.com/Graylog2/graylog2-server/pull/5983

https://github.com/Graylog2/graylog2-server/issues/4583
https://github.com/elastic/elasticsearch/issues/15951

coffee-squirrel:
Mentioned on the case, but being able to `debug(to_string($message))` (or equivalent) would've helped greatly in figuring out what was going on.

mpfz0r:
@coffee-squirrel Like this? https://github.com/Graylog2/graylog2-server/pull/13178

mpfz0r:
@coffee-squirrel Yeah, makes sense. I didn't do that in the beginning, because it required some changes to the rule parser.
But it should work now.

OzzyKampha:
Can this get fixed. Since the new versions of elastic have . As standard now

pandel:
Will this ever be solved? The [issue](https://github.com/Graylog2/graylog2-server/issues/4583) was created 5+ years ago and it seems, as if the whole thing hasn't been addressed yet, even though there are many people, which seem to have problems with this "dots aren't allowed" thing.

FlatCodeIq:
Hello,

We need this case to be addressed ASAP, Wazuh Dashboard is unable to read the logs due to the separator "_" instead of ".".

janheise:
[...]

PR #13178, opened 2022-08-02
Support $message expression in pipeline rules

mpfz0r:
This extends the pipeline rule syntax to allow
passing the entire `$message` object as parameters.

A useful example of this would be
 `debug($message);`

Other rule examples that are now possible:
 `has_field("source", $message);`
 `let foo = $message;`

### DIFF
--- a/graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions/debug.txt
+++ b/graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions/debug.txt
@@ -5,10 +5,11 @@ then
     debug($message.somefield);
     debug($message);
 
-    debug($message.nullfield);
 
     let messagevar = $message;
     debug(messagevar);
+
+    debug($message.nullfield);
 
     debug("message converted with to_string: " + to_string($message));
 end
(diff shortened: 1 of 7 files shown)


mpfz0r:
See https://github.com/Graylog2/graylog2-server/issues/13043#issuecomment-1183252047

graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions/debug.txt
```diff
@@ -0,0 +1,14 @@
...
+    debug("moo");
+    debug($message.somefield);
+    debug($message);
+
+    debug($message.nullfield);
+
+    let messagevar = $message;
+    debug(messagevar);
+
+    debug("message converted with to_string: " + to_string($message));
```
moesterheld, line 13: this debug call isn't verified, is it?

moesterheld (changes requested):

graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions/debug.txt
mpfz0r: @moesterheld thanks :+1: fixed

moesterheld (approved):
Looks good.

PR #13185, opened 2022-08-03
Use @Api tags to control visibility API docs in cloud

patrickmann:
Resolves Graylog2/graylog-plugin-cloud#1035

Enable API-Browser in Cloud context, while providing the ability to hide APIs that are not relevant or should not be exposed in Cloud. Visible APIs must be explicitly marked for Cloud by setting a specific `tags` value, e.g.:
`@Api(value = "MyCloudAPI", description = "Does something useful in Cloud", tags = {CLOUD_VISIBLE})`

By default, APIs are hidden in Cloud.

Currently the API Browser is accessible from the help menu.

# Screenshots

Screenshot 2022-08-16 at 11 43 44

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/api/GenerateApiDefinition.java
+++ b/graylog2-server/src/main/java/org/graylog/api/GenerateApiDefinition.java
@@ -87,7 +87,7 @@ public class GenerateApiDefinition {
 
         final ObjectMapper objectMapper = new ObjectMapperProvider().get();
 
-        final Generator generator = new Generator(resourceClasses, Collections.emptyMap(), "/plugins", objectMapper);
+        final Generator generator = new Generator(resourceClasses, Collections.emptyMap(), "/plugins", objectMapper, false);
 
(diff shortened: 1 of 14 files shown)


kodjo-anipah (commented):
Works as expected left one comment and one question regarding: 

> * Should we make the API browser read-only to prevent accidental changes?

Is it know wether POST & UPDATE request should be filtered out?

patrickmann:
> Is it know wether POST & UPDATE request should be filtered out?

@boosty Are you OK with selecting visible APIs at the "top" level (i.e. the list you see when you first open swagger, versus the set of individual APIs grouped beneath each top level entry)? We can be more granular if needed, though then it is also more work to mark each API.

boosty:
@patrickmann My current assumption is that filtering on the top level is a good start, but I can't say for sure that this will hold true in all cases.

I would not filter out POST & UPDATE requests in general. If a user has the permission to perform such a call, the API browser should allow it (as it would be essentially the same as using the UI).

kodjo-anipah (approved):

Issue #13157, opened 2022-07-28
System dropdown menu permissions mismatched


ryan-carroll-graylog:
Some of the user/role permissions for what items get populated in the System dropdown do not line up with the list items.
For instance: In order for "Pipelines" to be displayed in the dropdown to a user, the `inputs:create` permission is required.

The required permissions are set in the file: ` graylog2-server/graylog2-web-interface/src/components/navigation/SystemMenu.jsx`
It appears that there may be several items that need to be updated.

## Expected Behavior

The required permissions would line up with those to view the lined page.
E.g. The "Pipelines" System dropdown permissions would be the same as the pipelines page.

## Possible Solution

Update  ` graylog2-server/graylog2-web-interface/src/components/navigation/SystemMenu.jsx` to use the same permissions as each items page.

## Steps to Reproduce (for bugs)

1. Create a role with the following permissions:
"pipeline:create",
"pipeline:read",
"pipeline:edit",
"pipeline:delete",
"pipeline_rule:create",
"pipeline_rule:read",
"pipeline_rule:edit",
"pipeline_rule:delete",
"pipeline_connection:read",
"pipeline_connection:edit"

2. Create a user and assign them only the new role
3. Click the System dropdown and observe that "Pipelines" is not available

## Context

This was brought up by a customer in a support ticket:
HS-1010204336

ryan-carroll-graylog:
Some open questions for discussion:
- What permissions should be required for each item (just read or do we need to be more specific)?
- How will this affect backwards compatibility with roles that have been created based on current permissions?

PR #13188, opened 2022-08-04
Update permissions to view items in System dropdown to be more relevant

ryan-carroll-graylog:
Updates the user/role permissions for what items get populated in the System dropdown.

## Description

The permissions for the System dropdown now relate more closely to the page permissions.

## Motivation and Context

Some of the user/role permissions for what items get populated in the System dropdown did not line up with the list items.
For instance: In order for "Pipelines" to be displayed in the dropdown to a user, the `inputs:create` permission is required.

See: https://github.com/Graylog2/graylog2-server/issues/13157

## How Has This Been Tested?

Created users with updated permissions and used System dropdown to navigated to affected pages.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I have updated the documentation accordingly.


graylog2-web-interface/src/components/navigation/SystemMenu.jsx
```diff
@@ -149,13 +149,13 @@ const SystemMenu = ({ location }) => {
       <IfPermitted permissions={['dashboards:create', 'inputs:create', 'streams:create']}>
```
danotorrey, line 149: Thanks a lot for noticing and fixing these permissions @ryan-carroll-graylog!

This looks good to me. Just noticed on thing: I think these content pack permission checks might also need to be swapped out with just `contentpack:read`. Perhaps when these individual dashboards/inputs/streams permissions checks were added, these were the only entities supporting content packs. My personal feeling is that checking the content packs read permission here is probably enough, since it's probably not maintainable (or pluggable) to check individual entity permissions here. Those permissions checks should probably happen on the Content Packs pages. I am not sure if that is being done right now, but that might be worth checking as well, just to see if there is any issue with permissions checks not being done (if so, we might want to capture that in another issue).

ryan-carroll-graylog: Ok good call @danotorrey, I hadn't realize that was an option. I just double checked the permissions on the Contents pack and a user with only `contentpack:read` can view the page but they can't install a pack with entities they don't have permissions for or create a Content pack.

I also pared down the Lookup Table permissions to just UPGRADING.md. They can view LUTs, Caches, and Data Adapters but can't make changes without the other permissions.

I also added documentation to UPGRADING.md. Let me know if you think that's specific enough and makes sense.

UPGRADING.md
```diff
@@ -80,6 +80,8 @@ these [instructions](https://www.mongodb.com/docs/manual/core/capped-collections
 UTF-8 encoding. 
 <br>Note that this encoding is applied to all messages received by the input. A single input
 cannot handle multiple log sources with different encodings.
+- The permissions for what gets populated in the System dropdown menu were updated to more closely match the page that they link to.
```
danotorrey, line 83: Suggesting a small adjustment in wording for clarity, and also a link to the PR, in case a user wants to see which permissions/options have changed.
```suggestion
- The permissions for which options are populated in the System dropdown menu were updated to more closely match the page that they link to. See [graylog2-server#13188](https://github.com/Graylog2/graylog2-server/pull/13188) for details.
```

danotorrey (approved):
@ryan-carroll-graylog LGTM! 👍 I left one small suggestion below.

zeeklop (commented):
LGTM

graylog2-web-interface/src/components/navigation/SystemMenu.jsx
danotorrey: LGTM! Thanks @ryan-carroll-graylog!

danotorrey (approved):

graylog2-web-interface/src/components/navigation/SystemMenu.jsx
```diff
@@ -146,16 +146,16 @@ const SystemMenu = ({ location }) => {
...
-      <IfPermitted permissions={['inputs:read']}>
+      <IfPermitted permissions={['grok_pattern:read']}>
         <NavigationLink path={Routes.SYSTEM.GROKPATTERNS} description="Grok Patterns" />
       </IfPermitted>
-      <IfPermitted permissions={['inputs:edit']}>
+      <IfPermitted permissions={['lookuptables:read']}>
         <NavigationLink path={Routes.SYSTEM.LOOKUPTABLES.OVERVIEW} description="Lookup Tables" />
       </IfPermitted>
-      <IfPermitted permissions={['inputs:create']}>
+      <IfPermitted permissions={['pipeline:read', 'pipeline_connection:read']}>
```
bernd, line 158: Hmm, we are changing the permissions from "write" to "read" here. Is that really what we want? Do all the pages work for users that only have `*:read` permissions to the subsystems?

ryan-carroll-graylog: @bernd I originally went with `*:write`, but I tested these out and users with only `*:read` permissions can navigate to the subsystem pages and view the contents but can't make any changes. So I went with the minimal permissions to navigate to the page.

If you think that's not a realistic use case or too permissive I can definitely update the permissions to `*:write` here.

bernd: @ryan-carroll-graylog Okay, thanks for the explanation.

Did you test with a user with only the "Reader" role? Does anything in the menu change for these users with this PR applied?

ryan-carroll-graylog: @bernd No problem. There is one change for a user with only the "Reader" role, because it seems to be tied to `inputs:read`: They can no longer see Grok Patterns. Otherwise they are still unable to see the other updated subsystems in the dropdown.

bernd: Alright, sounds good! Thank you! :+1:

ryan-carroll-graylog: Thanks for taking a look!

kingzacko1 (approved):

PR #13194, opened 2022-08-05
Return correct time range when creating aggregation events during catch-up processing

### LINKED ISSUE #13061: Aggregation sourceMessagesForEvent incorrect when the event processor is catching up on old data?
From the source code, it seems that the time range used to search for messages in method sourceMessagesForEvent of the AggregationEventProcessor is incorrect, when the event processor has been catching up on old data.

## Expected Behavior
This problem concerns aggregation with [series configured](https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L159) and only in the special case where the event was created when the event processor was [catching up on old data](https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorExecutionJob.java#L132). When this is the case, the time range transmitted to createEvents and then put in the created Event is the big time range which contains multiple processingWindowSize chunks.
[sourceMessagesForEvent](https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L143) should search for message summaries in the chunk which triggered the event (rather than the whole timerange). It should feed the messageConsumer only the summaries of the small chunk.

## Current Behavior
According to the code of the [sourceMessagesForEvent](https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L178), it is currently using the time range that is stored in fields timerangeStart and timerangeEnd of the Event. But this time range corresponds to the big time range, because this is what is used to initialize the Event in [eventsFromAggregationResult](https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L286).

## Possible Solution
Here are two potential solutions:
* the correct time range could be computed from the field EventTimeStamp of the Event in the following way: `AbsoluteRange.create(timeStamp.minusSeconds((int) (searchWithinMs / 1000), timeStamp.minusMillis(1))`. (Since, the timeStamp corresponds to the end of the chunk. Also, one millisecond is removed, because the end of the chunk should be excluded)
* the time range could be stored in the AggregationKeyResult and then stored in fields timerangeStart and timerangeEnd of the Event.

Sorry, I don't have steps to reproduce the bugs, since it was found by reading the code.

## Context
In this plugin https://github.com/airbus-cyber/graylog-plugin-correlation-count, we are trying to correlate two streams. So we perform two aggregation searches. And, then we need to retrieve the associated message summaries. At this point we need to have the correct time range from the aggregation results. That is what lead us to having to understand this behaviour.

## Your Environment
* Graylog Version: 4.2.3

patrickmann:
Resolves #13061 

During catch-up processing, aggregation events were being generated with the time range of the extended catch-up window, instead of the time range of the actual result bucket. 

We now add a `from` field to each `AggregationKeyResult` and use that to determine the event time range.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
@@ -31,12 +31,10 @@ import java.util.Optional;
 public abstract class AggregationKeyResult {
     public abstract ImmutableList<String> key();
 
+    // timestamp() is set to the end of the date range.
     public abstract Optional<DateTime> timestamp();
 
     public abstract ImmutableList<AggregationSeriesValue> seriesValues();
-
-    // timestamp() is set to the end of the date range. that's why we don't need to()
-    public abstract Optional<DateTime> from();
 
     public static Builder builder() {
         return Builder.create();
@@ -56,8 +54,6 @@ public abstract class AggregationKeyResult {
         public abstract Builder key(List<String> key);
 
         public abstract Builder seriesValues(List<AggregationSeriesValue> seriesValues);
-
-        public abstract Builder from(@Nullable DateTime from);
 
         public abstract AggregationKeyResult build();
     }
--- a/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
@@ -35,6 +35,8 @@ public abstract class AggregationKeyResult {
 
     public abstract ImmutableList<AggregationSeriesValue> seriesValues();
 
+    public abstract Optional<DateTime> from();
+
     public static Builder builder() {
         return Builder.create();
     }
@@ -42,7 +44,7 @@ public abstract class AggregationKeyResult {
     public abstract Builder toBuilder();
 
     @AutoValue.Builder
-    public static abstract class Builder {
+    public abstract static class Builder {
         @JsonCreator
         public static Builder create() {
             return new AutoValue_AggregationKeyResult.Builder();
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
```diff
@@ -35,14 +35,16 @@ public abstract class AggregationKeyResult {
 
     public abstract ImmutableList<AggregationSeriesValue> seriesValues();
 
+    public abstract Optional<DateTime> from();
```
mpfz0r, line 38: Hmm, I don't see how and where we transfer this to the actual Event

mpfz0r (changes requested):
I don't see how this changes anything.
Weren't we supposed to change the query time here?
https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L178

I think we would need to add the time range here:
https://github.com/Graylog2/graylog2-server/blob/4.2.3/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L286-L287

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationKeyResult.java
patrickmann: We don't :/ Thought I had set the interval, but I was only copying the timestamp from the keyresult.

PR #13196, opened 2022-08-05
Explicitly trigger binders once to avoid injection errors.

dennisoelkers:
## Motivation and Context

Prior to this PR, setting up the injector could fail if running without plugins, due to:

  - No post-validation search normalizer being bound
  - No lookup table service factory annotated with `@SystemEntity` being bound

Guice works in different ways when the type of a multi-binder is supposed to be injected and no implementations are present:

  - When the map/set-binder has never been instantiated, guice throws an injection error
  - When the map/set-binder has been instantiated, but no implementation was bound, no error is raised and an empty collection is injected

So in order to fix the startup failure, this PR is calling just the binder of those two entities once, so instead of throwing an error, empty collections are injected.

## Types of changes

## Checklist:


todvora (approved):
Thank you! Tested locally and the problem is fixed, all good now!

PR #13241, opened 2022-08-15
Create new package for custom ESLint rules.

linuspahl:
## Motivation and Context

With this PR we are adding a new package, which allows us to create custom ESLint rules.
It contains a first rule that suggests implementing the `useUserDateTime` context instead of consuming the `CurrentUserContext`.

Implementing a specific hook instead of consuming the context is a pattern we want to apply more in the future.
Since it is very easy to forget this pattern and consume the context directly, it is great to have an ESLint rule to avoid these mistakes.

I also considered implementing an auto fix suggestion, but the problem is, that running `eslint --fix` would then result in a problem, since we are not automatically adding the necessary import.

Please note I have not yet added tests for the new rule. 

To test this change you need to remove and reinstall the `node_modules`.

### DIFF
--- a/graylog2-web-interface/packages/eslint-plugin-graylog/lib/rules/user-date-time-injection.js
+++ b/graylog2-web-interface/packages/eslint-plugin-graylog/lib/rules/user-date-time-injection.js
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


graylog2-web-interface/packages/eslint-plugin-graylog/lib/rules/user-date-time-injection.js
```diff
@@ -0,0 +1,31 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+/* eslint-disable global-require */
+module.exports = (context) => ({
+  CallExpression: (node) => {
+    const { callee } = node;
+
+    if (callee.name === 'useContext'
+      && callee?.parent?.arguments?.[0].name === 'UserDateTimeContext'
+      && !context.getFilename().endsWith('useUserDateTime.ts')
```
linuspahl, line 24: Maybe the last condition is too specific and we should just disable the rule in the `useUserDateTime` file instead?

simonychuang (approved):
LGTM!

PR #13250, opened 2022-08-17
Inline/delete filters from views and widgets on filter deletion

todvora:
See the enterprise counterpart PR for description

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#3946


luk-kaminski (approved):
Looks fine, thanks!

PR #13256, opened 2022-08-18
Extract StreamMatcherFilter into separate MessageProcessor

### LINKED ISSUE #6464: Split Message Filter Chain into Input and Stream stages
I'd like to process messages as follows:

`Input Handling (including static fields) -> Pipelines -> Stream Rule Processing`

However, because input-handling such as setting static fields, and stream rule processing, both happen as part of the Message Filter Chain, this is currently impossible. The only option I have to delete all my stream rules and replace them with an extra pipeline stage. While this is certainly doable, it's a lot more complicated than just specifying, on the stream, what messages belong in that stream.

## Expected Behavior
The current Message Filter Chain should be split into two components that can be ordered separately; one component would handle the input-oriented functionality such as static fields and input extractors, and the other component would handle the stream-oriented functionality such as stream-rule-based-routing.

## Current Behavior
Input-related message processing and stream-related message processing must occur in the same stage, so pipeline processing cannot go between them.

## Context
I have pipelines that process messages based on the presence of certain fields, and then stream rules route messages into streams based in part on the results of the pipelines. Now, I've added a Cloudwatch input for messages coming from RDS, and the static fields I add (which the processing pipelines require) aren't added until after the processing pipeline is already done.

## Your Environment
* Graylog Version: 3.1.2
* Elasticsearch Version: 6.5.2
* MongoDB Version: 3.0.15
* Operating System: Centos 7.4

mpfz0r:
Matching (routing) Messages into streams used to happen within the
MessageFilterChainProcessor and was run as the last filter.

Making Stream Matching a separate Message Processor gives the users
more flexibility. They can for instance run pipeline and extractors first,
and perform stream assignment last.

Fixes #6464


thll (approved):
Thanks for that improvement. LGTM and works.

PR #13257, opened 2022-08-18
Create `useCurrentUser` hook to receive current user.

linuspahl:
## Motivation and Context

With this PR we are creating a new `useCurrentUser` hook to receive the current user in frontend components.
Before this change we were consuming the `CurrentUserContext` for this purpose.

This is not the first time we are replacing the consumption of a context with implementing a specialised hook. This pattern has a few benefit:
- The main advantage: If we ever change the way we provide the current user, for example if we replace the `CurrentUserContext.Provider` with `react-query`, we only need to change one file instead of every file were we access the current user.
- if we forget to wrap a component with the necessary context provider, the hook throws a more meaningful error message.
- it is simpler to mock and change the return value of an hook in tests

This PR also affects the general mocking of the current user in tests.
Similar to the [UserDateTimeProvider](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/test/DefaultProviders.tsx#L52) we are now always providing a current user when rendering something in tests.

This simplifies many tests, since we often just had to mock the current user, because some child component is consuming the context, while mocking the user is not important for the tested logic.

Here is what changes, in case you need to test for example different user permissions in a test.
Instead of wrapping every component in every test caste with `CurrentUserContext.Provider`, you need to

1. mock the hook `jest.mock('hooks/useCurrentUser');`
2. reset the return value before every test 
```
beforeEach(() => {
  asMock(useCurrentUser).mockReturnValue(adminUser);
})
```
3. change the return value for a specific test case.
```
asMock(useCurrentUser).mockReturnValue(
  adminUser.toBuilder().permissions(Immutable.List(['example-permission'])).build()
);
```

## Types of changes

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3948

### DIFF
--- a/graylog2-web-interface/src/hooks/useCurrentUser.ts
+++ b/graylog2-web-interface/src/hooks/useCurrentUser.ts
@@ -14,7 +14,6 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-
 import { useContext } from 'react';
 
 import CurrentUserContext from 'contexts/CurrentUserContext';
--- a/graylog2-web-interface/src/hooks/useCurrentUser.ts
+++ b/graylog2-web-interface/src/hooks/useCurrentUser.ts
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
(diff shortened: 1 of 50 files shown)


graylog2-web-interface/src/hooks/useCurrentUser.ts
```diff
@@ -0,0 +1,31 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import { useContext } from 'react';
+
+import CurrentUserContext from 'contexts/CurrentUserContext';
```
zeeklop, line 19: At some point, I think it would be a good idea to get rid of the `CurrentUserContext` in favor of the hook.

zeeklop (approved):
Looks good to me.

graylog2-web-interface/src/hooks/useCurrentUser.ts
linuspahl: I have not thought about it, but it is true at some point we no longer need the context.

PR #13258, opened 2022-08-18
Rename 'All Messages' stream to 'Default Stream'

mpfz0r:
Once a user configures other streams, the 'All Messages' stream might
not contain "all messages" any more. This is more logical.

We only do the name change for new installations.
Otherwise we might break existing pipeline rules, which can look up streams by name.

Stream names are also not guaranteed to be unique, btw :frowning_face:


coffee-squirrel:
Might be good to mention this in `UPGRADING.md`, just in case there are pipeline rules (or whatever) targeting the `All messages` name.

mpfz0r:
@coffee-squirrel thanks for the hint. I guess I will need to change those rules as well in the migration :-/

mpfz0r:
change of plans: we will only do that for new installations. It's not worth the effort.

boosty (commented):
@mpfz0r A very useful change. Thanks a lot!

Someone else should review the code, but in addition please make sure that the documentation team gets informed, and the other developers are aware and get a chance to review their code for references to the old name. Thanks!

mpfz0r:
The docs in question are here: https://docs.graylog.org/docs/stream-connections

Since we don't have versioned documents anymore, we probably need to solve this with a footnote
saying that `Default Stream` used to be called `All Messages`

bernd:
> @mpfz0r A very useful change. Thanks a lot!
> 
> Someone else should review the code, but in addition please make sure that the documentation team gets informed, and the other developers are aware and get a chance to review their code for references to the old name. Thanks!

@mpfz0r @boosty Do we want to use "Default Stream" or "Default"? Putting the word stream into the name feels redundant. We don't include "Stream" in our other system streams.

mpfz0r:
> 

I think it makes it a bit clearer in the search context:

image

But I'm not completely against just using Default.
And we are also using `Default Index Set` :-)

boosty:
No strong opinion, but I think @mpfz0r has a point here. So from my side we can keep `Default Stream`.

bernd (approved):

Issue #13259, opened 2022-08-18
When attempting to copy the message, selecting message text immediately gets deselected


drewmiranda-gl:
It's not uncommon at all to need to copy the contents of the message field when viewing results from the serach page (or a message table widget). When selecting text that does NOT have a message summary (as populated via graylog illuminate), the text is immediately deselected and the message row is either expanded or compressed (depending on its starting state).

This can make it difficult and cumbersome to copy text from the message [summary] field.

## Expected Behavior

Selected text should not be deselected, it should remain selected so one can select text and copy it.

## Current Behavior

Messages that don't have a populated message summary (e.g. via illuminate) cause selected text to be deselected.

## Possible Solution

It appears the UX is fully deleting the DIV containing the message so naturally selected text will disappear thus "deselecting" that text.

Ideally this DIV is not deleted.

Jeff D. also suggested a possible "copy message" button when expanding.

## Steps to Reproduce (for bugs)

1. Go to graylog search page
2. Find a message that DOES NOT have a message summary
3. Select text from the message, beneath the message date
4. Observe text is deselected when releasing left click

## Context

This appears to be a pain point among users of graylog causing frustration.

Please feel free to slack/zoom me to discuss further. I can provide a live demonstration :)

## Your Environment

* Graylog Version: 4.3.5
* Java Version: 11.0.16
* Elasticsearch Version: Opensearch 1.3.4
* MongoDB Version: 4.4.15
* Operating System: Ubuntu Server 20.04 LTS
* Browser version: Chrome 104.0.5112.79

drewmiranda-gl:
For context, messages that are getting delected are wrapped in a div and the div is deleted on click:

```html
<div class="MessageFieldRow__MessageWrapper-sc-11840nk-0 iknNhj">example message</div>
```

However, messages with summary, are in a div that is NOT deleted on click

```html
<div title="[{gim_event_type}] {application_name} - - {user_name}" class="MessageSummary__SummaryContainer-sc-1l479fn-0 goFRIA">[logoff] windows - - graylog</div>
```

drewmiranda-gl:
This will be resolved in the next release. Missed the latest release by 1 day!
https://github.com/Graylog2/graylog2-server/pull/12877

PR #13263, opened 2022-08-19
Do not toggle message details in message table widget, when selecting text.

linuspahl:
_Please note this PR should be backported_

## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/13259 we currently toggle the message row details when selecting text in the message table. 

Example (before)

message-table-text-select-before

With this PR we are improving the behaviour, only a simple click will toggle the message row details.

Example (after)

message-table-text-select-after

One thing which may is not optimal, if you select a text and you click on the selected text it requires a second click to actually toggle the message details.

Fixes https://github.com/Graylog2/graylog2-server/issues/13259

## Types of changes


zeeklop (approved):
LGTM

dennisoelkers (approved):
Looks good, works great!

PR #13270, opened 2022-08-22
Move `usePluginEntities` into `hooks` directory.

linuspahl:
## Motivation and Context

The `usePluginEntities` function is a hook which returns and memories the result of `PluginStore.exports` for a defined plugin export. Initially we mainly used it in our `views` code. Since we use this hook everywhere and not only in the views code, it makes sense to move it into the `hooks` directory.

## Types of changes

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3954


grotlue (approved):
Looks good (once it's green 😄 )

Issue #9933, opened 2021-01-19
Permissions: Have Role assigned when User is Created


kyleknighted:
## Expected Behavior

Assign Roles on Creation

## Current Behavior

/system/users/new

- Filled in all the fields, and then clicked Create User with a few Roles
- Logged into the User’s account to see I couldn’t access anything
- Went back to Edit the User and it didn’t have any Roles associated to it
- Went to create another User, same issue
- Then I realized you have to click a separate ‘Assign Role’ button

## Possible Solution

Remove the "Assign Role" button and have the roles assigned to the User when the Create form is submitted

PR #13277, opened 2022-08-23
Assign roles without using the submit button on user create page

grotlue:
## Description
On the user creation page one can add and remove roles. 
This PR will remove the "Assign role" button from this view and assign roles when they are selected instead.

## Motivation and Context
When creating user the actual user creation does only happen when submitting the whole form, so there is no need for an additional button and thus step when adding or removing roles.

Closes https://github.com/Graylog2/graylog2-server/issues/9933

### Note:
To unify the behaviour we should also adjust the role selector on the team creation page.
Additionally the user selector works the same way, this should also be adjusted.

## How Has This Been Tested?
**Environment:** Local

- Created a new user, selected and deselected roles
- Checked the new user after creation

✅ Roles were saved accordingly

### Before:
Monosnap screencast 2022-08-23 12-48-31

### After:
Monosnap screencast 2022-08-23 12-51-42

## Types of changes

## Checklist:


mnin (approved):
LGTM

graylog2-web-interface/src/components/permissions/RolesSelector.tsx
```diff
@@ -59,19 +60,19 @@ const _options = (roles, assignedRolesIds, identifier) => roles
...
 
-const _assignRole = (selectedRoleName, roles, onSubmit, setSelectedRoleName, setIsSubmitting, setError) => {
-  if (!selectedRoleName) {
+const _assignRole = (selectedRoleNames, roles, onSubmit, setselectedRoleNames, setIsSubmitting, setError) => {
+  if (!selectedRoleNames) {
     return;
   }
 
-  const selectedRoleNames = selectedRoleName.split(',');
+  const selectedRoleNamess = selectedRoleNames.split(',');
```
ousmaneo, line 68: We should call this `selectedRolesNameArray` for example to make it more readable.

ousmaneo (changes requested):
Thank you for the changes this looks good to me. I added a small comment inline

ousmaneo (approved):
LGTM!

PR #13283, opened 2022-08-24
Update MongoDB driver

### LINKED ISSUE #13199: Support MongoDB versions greater than 4.4
## What?
Similar to #10720.
Graylog should support current MongoDB versions.

## Why?

Graylog currently only supports MongoDB versions up to 4.4.
In the meantime, MongoDB versions 5.0 and 6.0 have been released.
New installations of MongoDB already default to version 6.0.

Version 5.0 still allows compatibility with 4.4 by setting the appropriate `featureCompatibilityVersion`, however, version 6.0 drops support for 4.4 compatibility.

Note that MongoDB 4.4 is still supported until February 2024:
https://www.mongodb.com/support-policy/lifecycles

MongoDB already only supports MongoDB 5.0+ on Debian 11, which may also be the case for Ubuntu 22.04 when MongoDB starts supporting it:
https://www.mongodb.com/docs/manual/administration/production-notes/#platform-support-matrix

On MongoDB 5.0 with 4.4 compatibility I already see the following warning in the logs after starting up Graylog:
```json
{
    "attr": {
        "clientInfo": {
            "driver": {
                "name": "mongo-java-driver|legacy",
                "version": "3.12.1"
            },
            "os": {
                "architecture": "amd64",
                "name": "Linux",
                "type": "Linux",
                "version": "5.18.16-arch1-1"
            },
            "platform": "Java/Oracle Corporation/1.8.0_345-b01"
        },
        "op": "query"
    },
    "c": "COMMAND",
    "ctx": "conn1",
    "id": 5578800,
    "msg": "Deprecated operation requested. For more details see https://dochub.mongodb.org/core/legacy-opcode-compatibility",
    "s": "W",
    "t": {
        "$date": "2022-08-07T18:53:19.354+00:00"
    }
}
```

## Your Environment

* Graylog Version: 4.3.3+86369d3
* Elasticsearch Version: N/A
* MongoDB Version: 5.0.9
* Operating System: N/A
* Browser version: N/A

thll:
Some WIP to update ~~Mongojack and~~ the MongoDB driver to the most recent version.

**We have decided to limit the scope of this PR to only updating the MongoDB driver. Introducing Mongojack 4 will be part of a separate PR.**

Fixes #13199

* Updates the MongoDB driver to version `4.7`
* Uses a [custom fork](https://github.com/graylog-labs/mongojack/pull/2/files) of Mongojack `2.10` to have it work with the MongoDB driver `4.7`. This way, we can keep most of our existing code.
* ~~Introduces Mongojack `4.7` as a [shaded dependency](https://github.com/Graylog2/graylog-shaded/tree/add/mongojack) so that it doesn't conflict with the legacy version of Mongojack that we are still keeping~~
* ~~Wraps the `JacksonMongoCollection` behind a custom `GraylogMongoCollection` to promote usage of the MongoDB driver API instead of the Mongojack API and make it easier to swap out our shaded Mongojack dependency with a non-shaded dependency once the migration is complete.~~

TODO:
- [x] make the PR builds pass by publishing the forked and the shaded Mongojack maven artifacts
- [ ] ~~migrate more code to use the new `GraylogMongoCollection`~~
- [ ] ~~discuss if the `GraylogMongoCollection` (plus factory) is the right way to interface with Mongojack~~
- [ ] ~~check if the ObjectMapper used with the new Mongojack version needs additional configuration (see `org.graylog2.bindings.providers.MongoJackObjectMapperProvider` for how the legacy Mongojack mapper was configured)~~

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise-integrations#871,Graylog2/graylog-plugin-enterprise#3970,Graylog2/graylog-plugin-collector#206


graylog2-server/src/main/java/org/graylog2/configuration/MongoDbConfiguration.java
```diff
@@ -28,9 +28,6 @@ public class MongoDbConfiguration {
     @Parameter(value = "mongodb_max_connections", validator = PositiveIntegerValidator.class)
     private int maxConnections = 1000;
 
-    @Parameter(value = "mongodb_threads_allowed_to_block_multiplier", validator = PositiveIntegerValidator.class)
```
bernd, line 31: Please add a note about the config option removal to `UPGRADING.md`.

bernd (changes requested):

bernd (approved):

PR #13285, opened 2022-08-24
Replace usage of `CurrentUserStore` with `useCurrentUser` hook and remove no longer needed mocks.

linuspahl:
## Motivation and Context

With this PR we are mainly removing no longer needed `CurrentUserStore` mocks.

There are multiple reasons why we no longer need these mocks.
- We removed the `DateTime` call which implemented the `CurrentUserStore`.
- Instead of consuming the store directly we are now using a hook which provides the current user.
- Before implementing the hook we were consuming the `CurrentUserContext`.

Please note, there are a few more files which use the `CurrentUserStore`. It requires a bigger refactoring to implement the `useCurrentUser` hook for these as well. I will continue the refactoring in a follow-up PR.

## Types of changes


graylog2-web-interface/src/contexts/CurrentUserPreferencesProvider.test.tsx
```diff
@@ -67,30 +72,25 @@ describe('CurrentUserPreferencesProvider', () => {
   });
 
   it('provides default user preferences if the user has none', () => {
-    asMock(CurrentUserStore.getInitialState).mockReturnValue({ currentUser: {} as UserJSON });
+    asMock(useCurrentUser).mockReturnValue({} as User);
```
dennisoelkers, line 75: What about providing a valid user, but with undefined preferences instead?

graylog2-web-interface/src/pages/StreamEditPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { Alert } from 'components/bootstrap';
+import StreamRulesEditor from 'components/streamrules/StreamRulesEditor';
+import { DocumentTitle, PageHeader, Spinner } from 'components/common';
+import withParams from 'routing/withParams';
+import withLocation from 'routing/withLocation';
+import StreamsStore from 'stores/streams/StreamsStore';
+import useCurrentUser from 'hooks/useCurrentUser';
+
+const StreamEditPage = ({ params, location }) => {
```
dennisoelkers, line 28: Can you please type the props here?

graylog2-web-interface/src/contexts/CurrentUserPreferencesProvider.test.tsx
```diff
@@ -52,6 +53,10 @@ describe('CurrentUserPreferencesProvider', () => {
     return consume;
   };
 
+  beforeEach(() => {
+    asMock(useCurrentUser).mockReturnValue(adminUser);
```
dennisoelkers, line 57: I think we can remove this and leave it to the default mock.

graylog2-web-interface/src/pages/EditEventDefinitionPage.test.tsx
```diff
@@ -64,35 +64,29 @@ jest.mock('stores/event-definitions/EventDefinitionsStore', () => ({
...
-    return (
-      <CurrentUserContext.Provider value={currentUser}>
-        {EditEventDefinitionPageWithParams}
-      </CurrentUserContext.Provider>
-    );
-  };
-
-  EditEventDefinitionPageWithPermissions.defaultProps = { permissions: [] };
+  beforeEach(() => {
+    asMock(useCurrentUser).mockReturnValue(adminUser);
```
dennisoelkers, line 72: I think we can remove this and leave it to the default mock.

graylog2-web-interface/src/pages/ViewEventDefinitionPage.test.tsx
```diff
@@ -51,35 +54,32 @@ jest.mock('stores/event-notifications/EventNotificationsStore', () => ({
...
-    return (
-      <CurrentUserContext.Provider value={user}>
-        {EventDefintionPageWithParams}
-      </CurrentUserContext.Provider>
-    );
-  };
-
-  EventDefinitionPageWithPermissions.defaultProps = { permissions: [] };
+  beforeEach(() => {
+    asMock(useCurrentUser).mockReturnValue(adminUser);
```
dennisoelkers, line 58: I think we can remove this and leave it to the default mock.

graylog2-web-interface/src/views/components/contexts/SearchPagePreferencesProvider.test.tsx
```diff
@@ -48,19 +47,28 @@ jest.mock('logic/local-storage/Store', () => ({
...
+jest.mock('views/stores/ViewStore', () => ({
+  ViewStore: {
+    listen: jest.fn(() => () => {}),
+    getInitialState: jest.fn(() => ({})),
+  },
+}));
+
 describe('SearchPagePreferencesProvider', () => {
+  beforeEach(() => {
+    asMock(useCurrentUser).mockReturnValue(adminUser);
```
dennisoelkers, line 59: I think we can remove this and leave it to the default mock.

dennisoelkers (changes requested):
Looks good! Just a couple of minor suggestions inline.

linuspahl:
@dennisoelkers Regarding the usage of `asMock(useCurrentUser).mockReturnValue(adminUser);` in `beforeEach`, I am doing this whenever we are mocking the return value in one or more test cases, to ensure we reset the return value for all following test cases.

dennisoelkers:
@linuspahl: I understand the reasoning, but I think it is not needed in at least a couple of test suites, where we can remove them for now. The benefit is that we have a single mock implementation which we can maintain centrally.

dennisoelkers (approved):

Issue #13267, opened 2022-08-22
Remove legacy alarm callbacks from UI and backend


patrickmann:
We removed legacy alerting management pages UI in #12987 and backend in #12988.
Now we also remove legacy alarm callbacks, which are still available as a _Notification Type_ when creating new notifications. Legacy alarm callbacks have been deprecated since 4.0. 

<img width="465" alt="image" src="https://user-images.githubusercontent.com/12073930/185910892-f9026920-7e42-4a1c-a798-371bc381bc80.png">

Issue #12988, opened 2022-07-01
Remove API code for legacy alerting management pages.


linuspahl:
## Description

A while ago we rewrote the alerting UI. With [this PR](https://github.com/Graylog2/graylog2-server/pull/12987) we removed the related legacy UI code. The next step is removing the related backend code.

Here is an overview of the no longer needed API routes.

```
/streams/${streamId}/alerts/conditions
/streams/${streamId}/alerts/conditions/${alertConditionId}
/streams/${streamId}/alerts/sendDummyAlert
/streams/${streamId}/alerts/conditions/${conditionId}/test
/alerts/conditions/types
/alerts/conditions
/streams/alerts/${alertId}
/streams/${streamId}/alerts
/streams/alerts/paginated
/streams/alerts
```

Please note we still want to support legacy alarm callbacks, until the next major release.

## Your Environment

* Graylog Version: 4.3

boosty:
Sounds like a candidate for removal in 5.0.

mpfz0r:
We can also remove all the stream related alerting logic:
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/streams/StreamService.java#L65-L83

patrickmann:
For those less familiar (like myself): This issue only aims to remove backend code that relates to stream-based alerting. It does not remove legacy alarm callbacks. Eventually we would want to remove those too; but that will require migration code to ensure that 3rd party plugins don't break.

PR #13286, opened 2022-08-24
Remove deprecated legacy alarms/alerts

patrickmann:
Addresses issues Graylog2/graylog2-server#12988 and Graylog2/graylog2-server#13267

Legacy alarms have been deprecated for a while. We are now completely removing them, including the corresponding migration.


patrickmann:
This PR was too aggressive in deleting code - creating a new one that is more surgical

PR #13287, opened 2022-08-24
Show message on edit event definition page if it is immutable

simonychuang:
Fixes a slight issue with the event definitions page where manually navigating to the edit page of an immutable event definition will still display the details like it is editable.

## Description

Rewrite the EditEventDefinition component into Typescript, leveraging the new useScopePermissions hook to check for the _scope of the selected Event Definition. If it is found that it is immutable, simply display an empty page with a short explanation.

## How Has This Been Tested?

Tested on local environment by installing a content pack with immutable event definitions, then manually going to the edit page of an immutable event definition.

Screen Shot 2022-08-24 at 3 14 35 PM

## Types of changes

## Checklist:


graylog2-web-interface/src/hooks/useScopePermissions.ts
```diff
@@ -50,7 +50,7 @@ const useGetPermissionsByScope = (entity: Partial<GenericEntityType>) => {
     },
   );
 
-  const scope = entity._scope ? entity._scope.toUpperCase() : 'DEFAULT';
+  const scope = entity?._scope.toUpperCase() || 'DEFAULT';
```
zeeklop, line 53: I'm not entirely sure, but I think existing entities would not have a `_scope` field. If that's the case, this will throw an error when trying to `toUpperCase` an `undefined` variable.

zeeklop (commented):
All looks good. Just a left a comment and it might be a good idea to add a test for that case

graylog2-web-interface/src/hooks/useScopePermissions.ts
simonychuang: Actually this is the reason why I made the changes to this line. Before it would throw an error when trying to access `_scope`, but using the optional chaining it would result in `undefined` if the `_scope` doesn't exist, and therefore will return 'DEFAULT'

zeeklop: So the actual error was assuming the entity will always be there to initialize the hook.

This works because the evaluation stops at the entity value. This is undefined, so the rest isn't executed, returning the `'DEFAULT'` value that we want.

zeeklop (approved):
Approved

zeeklop (approved):
Looks good!

graylog2-web-interface/src/hooks/useScopePermissions.ts
mnin: Create a PR https://github.com/Graylog2/graylog2-server/pull/13365 to fix https://github.com/Graylog2/graylog2-server/issues/13361

PR #13296, opened 2022-08-25
Improve explanation of search filter feature in search bar.

linuspahl:
## Motivation and Context

With this PR we are adding a more detailed description of the search filter feature for the explanation we display in the search bar.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/3974

### DIFF
--- a/graylog2-web-interface/src/components/common/HoverForHelp.tsx
+++ b/graylog2-web-interface/src/components/common/HoverForHelp.tsx
@@ -87,7 +87,7 @@ HoverForHelp.propTypes = {
   pullRight: PropTypes.bool,
   title: PropTypes.string,
   testId: PropTypes.string,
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/components/common/HoverForHelp.tsx
```diff
@@ -86,6 +87,7 @@ HoverForHelp.propTypes = {
   pullRight: PropTypes.bool,
   title: PropTypes.string,
   testId: PropTypes.string,
+  trigger: PropTypes.arrayOf(PropTypes.string),
```
maxiadlovskii, line 90: As far as I understand OverlayTrigger>['trigger'] might be just one single `string`

maxiadlovskii (approved):

PR #13327, opened 2022-08-30
Always provide the correct time range type during builder initialization

todvora:
Our AbsoluteRange builder needs an explicit type set in order to work properly. We can avoid that by predefining the type during the builder initialization, as we are doing in other generic types.

## Motivation and Context
```
AbsoluteRange range = AbsoluteRange.builder()
                .from(DateTime.now().minus(1000 * 60 * 50))
                .to(DateTime.now())
                .build();
```
This snippet would throw an IllegalStateException, because the type is not set. But the type is already predefined by the class itself and we can simply provide it from inside.

## How Has This Been Tested?
Added unit test

## Types of changes

## Checklist:


luk-kaminski (approved):
LGTM, thx!

dennisoelkers (approved):
Looks great now! Thanks! :pray:

janheise (approved):
LGTM!

dennisoelkers (approved):
Thanks a lot! :pray:

Issue #13238, opened 2022-08-15
Sidecar - Select Hosts -> Apply Configuration -> Filter by configuration is broken


HenryTheSir:
## Expected Behavior

Go to /system/sidecars/administration , select some hosts and a collector to which configs should be applied. Dropdown "Configure" Search for the configuration and apply the configuration
## Current Behavior

Same as above but dropdown "Configure" does not search correctly (always empty dropdown), scroll to thousands of configurations till the correct one is found
## Steps to Reproduce (for bugs)

1. Go to /system/sidecars/administration
2. Select some Hosts
3. Select Collector to configure
4. Dropdown 'Configure' Try to search -> empty result set

## Your Environment

* Graylog Version: 4.3.4
* Java Version: 11.x
* Elasticsearch Version: 7.10
* MongoDB Version: x
* Operating System: RHEL
* Browser version: Firefox ESR

coffee-squirrel:
#9133

PR #13339, opened 2022-08-31
Fix 13238 filter by the unique configurationNames instead of configurationIds

gally47:
## Description
As described in this issue https://github.com/Graylog2/graylog2-server/issues/13238 searching by configuration name in /system/sidecars/administration is not working when items are selected. this is because the search/filter logic works with IDs but in this case there is no reason to not work with names since they are unique.

## Motivation and Context

https://github.com/Graylog2/graylog2-server/issues/13238

## Types of changes

## Checklist:


mpfz0r (approved):
LGTM :+1:

PR #13347, opened 2022-09-01
Improve performance of ContainerMatrixTestEngine

bernd:
The "#discover()" method was running reflection on every call to find all
classes that are annotated with the "ContainerMatrixTestsConfiguration"
annotation. Since the "#discover()" method is called quite often during
test setup, we saw a long "pause" before the test execution started when
running "mvn test".

We now run the reflection in a "static" block and re-use the result for
each "#discover()" invocation.

The change improves running "mvn -Dskip.web.build clean test" in the
graylog2-server repository as follows:

Before: 05:10 min
After:  02:23 min


bernd:
I am checking the build error.

dennisoelkers (approved):

todvora:
Nice! Thanks @bernd!

PR #12613, opened 2022-05-06
Beta version of stream-aware field list, based on search engine queri…

luk-kaminski:
…es, combining ad-hoc and periodical-storage variant

## Description
See https://github.com/Graylog2/graylog2-server/issues/12572 for description.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #13363, opened 2022-09-05
Stream-aware field types per index

luk-kaminski:
Upgrade of #12613 with different approach.

## Description
1. Pollers responsible for creating and updating documents in "index_field_types" will now for each field add a new property called "streams", holding IDs of streams where this field was used.
2. MappedFieldTypesServiceImpl will go through "index_field_types" documents in order to provide "stream aware field list".
3. Pollers are based on terms aggregation on a "stream" field with "exists" query on the field. 

## Known restrictions
If field is present in more than **search.max_buckets** streams (by default 10k/65k), we are in trouble... Aggregation from point 3 won't return all streams and field may be missing in some suggestions later on.
Changing **search.max_buckets** to a low value may cause problems. We are working to document that changing this setting is highly discouraged.

## Types of changes

## Checklist:


dennisoelkers (changes requested):
Looks good, works fine! Just added some minor suggestions inline.

dennisoelkers (approved):
Looks good, works great! Thanks a lot! :pray:

Issue #13361, opened 2022-09-05
Broken "create" page for lookup tables


thll:
The "create" page for lookup tables (`/system/lookuptables/create`) cannot be rendered. It fails with the following error: 

`Loading component failed: Cannot read properties of undefined (reading 'toUpperCase')`

![image](https://user-images.githubusercontent.com/886541/188445033-37dc1028-9189-4dfb-93fc-2e8ac913287f.png)

## Your Environment

* Graylog Version: `master`
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

mnin:
Fixed in https://github.com/Graylog2/graylog-plugin-cloud/pull/1051

PR #13365, opened 2022-09-05
fix: use scope permissions toUpperCase call on undefined

mnin:
## Description

Fix the bug https://github.com/Graylog2/graylog2-server/issues/13361 which is visible on `/system/lookuptables/create`.

## Types of changes


ousmaneo (approved):
LGTM !

PR #13367, opened 2022-09-05
Support tags for automatic sidecar configuration

thll:
WIP


mpfz0r (approved):

graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/models/NodeDetails.java
```diff
@@ -50,12 +50,17 @@ public abstract class NodeDetails {
     @Nullable
     public abstract CollectorStatusList statusList();
 
+    @JsonProperty("tags")
+    @Nullable
+    public abstract List<String> tags();
```
bernd, line 55: Shouldn't this be a Set to prevent duplicates?

mpfz0r: wouldn't hurt yeah. can I fix this in a separate PR? I'm waiting for this one to be merged

PR #13405, opened 2022-09-09
Removing now unneeded package resolutions.

dennisoelkers:
## Motivation and Context

This change is removing package version resolutions which have become obsolete due to other package updates.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/package.json
+++ b/graylog2-web-interface/package.json
@@ -60,6 +60,7 @@
     "core-js": "3",
     "crossfilter": "1.3.x",
     "dompurify": "^2.0.11",
+    "glob-parent": "6.0.2",
     "graylog-web-plugin": "file:packages/graylog-web-plugin",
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/package.json
```diff
@@ -60,6 +60,7 @@
     "core-js": "3",
     "crossfilter": "1.3.x",
     "dompurify": "^2.0.11",
+    "glob-parent": "6.0.2",
```
linuspahl, line 63: Just because I am curious, why are we now listing `glob-parent` as a dependency?

dennisoelkers: That is actually a mistake. It should actually be bumped in the lockfile only, I missed this in `git diff`. Thanks!

linuspahl: Alright!

dennisoelkers:
Outdated.

PR #13425, opened 2022-09-12
Small general font improvements

linuspahl:
## Motivation and Context

With this PR we are mainly increasing the body and navigation font size and adjusting the default font size for bootstrap elements.

image

### DIFF
--- a/graylog2-web-interface/src/components/common/PageHeader.tsx
+++ b/graylog2-web-interface/src/components/common/PageHeader.tsx
@@ -35,7 +35,6 @@ const LifecycleIndicator = styled.span(({ theme }) => css`
 const H1 = styled.h1(({ theme }) => css`
   margin-bottom: 0.2em;
   font-family: ${theme.fonts.family.navigation};
-  font-size: ${theme.fonts.size.huge};
 `);
 
 const ActionsSM = styled.div`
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/theme/constants.ts
```diff
@@ -20,11 +20,12 @@ export type ThemeMode = 'teint' | 'noir';
 export type PreferencesThemeMode = 'themeMode';
 
 const PREFERENCES_THEME_MODE: PreferencesThemeMode = 'themeMode';
-const ROOT_FONT_SIZE = 14;
+const ROOT_FONT_SIZE = 14; // This value is also being maintained as @font-size-base in bootstrap-config.js.
```
maxiadlovskii, line 23: Why @font-size-base and  this one are different? Is it should be like that?

linuspahl: Currently it is the equivalent to body font size.
https://github.com/Graylog2/graylog2-server/pull/13425/files#diff-ea1ddcace101558dd56650177440aef11fa139013d8183c04f8e26ec1b02e291R76

But I will remove the comment, I realised it does not make sense anymore.

linuspahl: Unfortunately the value can not be defined in `rem`.

linuspahl:
After using these changes in a follow-up PR, I realised I can simplify the font-size relations and pushed another commit.

linuspahl:
I also realised that using a smaller scale looks better and pushed another change. If more comes to my mind I will implement it in a follow-up PR.

maxiadlovskii (approved):

PR #13486, opened 2022-09-19
Implement new navigation and headline font.

linuspahl:
## Motivation and Context

This this PR we are installing the font DM Sans and define it for the navigation and headlines.


dennisoelkers (approved):
:pray:

PR #13490, opened 2022-09-19
Implement shared component for modal submit and cancel button.

linuspahl:
## Motivation and Context

With this PR we are implementing a shared component (`ModalSubmit`) for modals, which maintains the submit and cancel button. This main benefit is that we can change for example the styling in one place.

Please note, I create a `FormSubmit` and a `ModalSubmit` component. The `FormSubmit` component will be used for forms in the future as well. Currently the `ModalSubmit` component is completely based on the `FormSubmit` component and does not contain any logic. I could have just implement the `FormSubmit` component for modals, but I think it is a good idea to have this abstraction.

## Types of changes


graylog2-web-interface/src/components/common/FormSubmit.tsx
```diff
@@ -0,0 +1,77 @@
...
+  leftCol?: React.ReactNode,
+  onCancel: () => void,
+  onSubmit?: () => void,
+  submitButtonText: string,
+  submitIcon?: IconName,
+  submitButtonType?: 'submit' | 'button',
+  submitLoadingText?: string,
+}
+
+const FormSubmit = ({
```
maxiadlovskii, line 38: Maybe it would be helpful to add `bsSize` props as well

linuspahl: I think so too, let me add it in a follow-up PR.

maxiadlovskii (approved):

Issue #13116, opened 2022-07-22
Show Title of dashboard on page


drewmiranda-gl:
## What?

When working with dashboards, I run into the issue where there isn't any visible indication of what dashboard you are looking at

The page title does show the dashboard title, but it would be very helpful from a UX perspective to see this a the top of the page.

## Why?

Dashboard Title: When working across with multiple dashboards at once (either multiple tabs in the same browser window, or multiple browser windows), it's difficult to keep track of which dashboard is which.

## Your Environment

* Graylog Version: Graylog 4.3.3+86369d3
* Elasticsearch Version: Opensearch 1.3.4
* MongoDB Version: 4.4.15
* Operating System: Ubuntu Server 20.04
* Browser version: Chrome 103

H2Cyber:
Another nice to have: being able to copy an entire page from one dashboard to another.

dennisoelkers:
Hey @drewmiranda-gl,

those are valid points, thank you for submitting this! We are tracking the tab ordering issue in #7470 already though, can you please reduce this issue to be only about displaying the dashboard title?

drewmiranda-gl:
> 

Done!

PR #13491, opened 2022-09-19
Add ViewHeader for saved search and dashboards

maxiadlovskii:
## Description
This PR adds
- Title for saved search and dashboards
- Edit metadata component into header

## Motivation and Context
fix: #13116 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/views/ViewHeader.tsx
+++ b/graylog2-web-interface/src/views/components/views/ViewHeader.tsx
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+
+import React, { useCallback, useMemo, useState } from 'react';
+import styled, { css } from 'styled-components';
+
+import { useStore } from 'stores/connect';
+import { ViewStore } from 'views/stores/ViewStore';
+import { Icon } from 'components/common';
+import { Row } from 'components/bootstrap';
+import ViewPropertiesModal from 'views/components/views/DashboardPropertiesModal';
+import onSaveView from 'views/logic/views/OnSaveViewAction';
+
+const StyledList = styled.dl(() => css`
+  display: flex;
+  gap: 8px;
+  margin: 0;
+  flex-wrap: wrap;
+  align-items: center;
+  dd {
+    font-style: italic;
+  }
+  dt {
+    text-transform: capitalize;
+  }
+`);
+
+const Content = styled.div(({ theme }) => css`
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: nowrap;
+  align-items: center;
+  padding-bottom: ${theme.spacings.sm}
+`);
+const EditButton = styled.div(() => css`
+  width: 25px;
+`);
+
+const ViewHeader = () => {
+  const { view } = useStore(ViewStore);
+  const isSavedView = useMemo(() => view?.id && view?.title, [view]);
+  const [showMetadataEdit, setShowMetadataEdit] = useState<boolean>(false);
+  const toggleMetadataEdit = useCallback(() => setShowMetadataEdit((cur) => !cur), [setShowMetadataEdit]);
+
+  return isSavedView ? (
+    <Row>
+      <Content>
+        <StyledList>
+          <dt>{view.type.toLocaleLowerCase()}:</dt>
+          <dd>{view.title}</dd>
+        </StyledList>
+        <EditButton onClick={toggleMetadataEdit}
+                    role="button"
+                    title={`Edit ${view.type.toLocaleLowerCase()} ${view.title} metadata`}
+                    tabIndex={0}>
+          <Icon name="pen-to-square" />
+        </EditButton>
+        {showMetadataEdit && (
+          <ViewPropertiesModal show
+                               view={view}
+                               title={`Editing saved ${view.type.toLocaleLowerCase()}`}
+                               onClose={toggleMetadataEdit}
+                               onSave={onSaveView} />
+        )}
+      </Content>
+    </Row>
+  ) : null;
+};
+
+export default ViewHeader;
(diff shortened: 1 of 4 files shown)


graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const Content = styled.div(({ theme }) => css`
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: nowrap;
+  align-items: center;
+  padding-bottom: ${theme.spacings.sm}
+`);
+const EditButton = styled.div(() => css`
+  width: 25px;
```
linuspahl, line 50: I do not think the width has any effect here.

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const [showMetadataEdit, setShowMetadataEdit] = useState<boolean>(false);
+  const toggleMetadataEdit = useCallback(() => setShowMetadataEdit((cur) => !cur), [setShowMetadataEdit]);
+
+  return isSavedView ? (
+    <Row>
+      <Content>
+        <StyledList>
+          <dt>{view.type.toLocaleLowerCase()}:</dt>
+          <dd>{view.title}</dd>
+        </StyledList>
```
linuspahl, line 65: The breadcrumb works as expected, just a few thoughts
- I would not display the edit icon in top right corner, it look like it opens a mode where you can edit for example the widget layout.
- Let's call the view type "Dashboards" instead of "Dashboard"
- We can consider making "Dashboards" a link. The link to the dashboards page is not far away, but it is still a breadcrumb and I can imagine some users will expect a link.
- Currently the bold text feels heavy
image

we can probably define a more subtle style
image

linuspahl (changes requested):

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
maxiadlovskii: Look like without it Icon is placed a bit outside the container... Or it looks like that just for me?

linuspahl: Not as far as I can tell:
image

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+font-size: 0.50rem;
+`;
+
+const ViewHeader = () => {
+  const { view } = useStore(ViewStore);
+  const isSavedView = useMemo(() => view?.id && view?.title, [view]);
+  const [showMetadataEdit, setShowMetadataEdit] = useState<boolean>(false);
+  const toggleMetadataEdit = useCallback(() => setShowMetadataEdit((cur) => !cur), [setShowMetadataEdit]);
+
+  return isSavedView ? (
```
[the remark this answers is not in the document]
dennisoelkers: Using the ternary operator is fine. If you want a short-circuit condition at the beginning of a component, using `if` is better, but if you want to decide what to render, the ternary operator is good too.

dennisoelkers (commented):
The PR looks good and works fine. Some open questions:

1. Should we show the breadcrumbs even for unsaved searches/dashboards? This way the layout would always be the same.
2. Do we want to show the edit button only when the user is hovering over the title? This would make it a bit cleaner.

maxiadlovskii:
@dennisoelkers 
It doesn't make sense to show breadcrumbs for unsaved searches. They don't have their name and there is nothing to show. Except smth like `Dashboard > Unsaved dashboard`. But I don't know if it makes sense for the users. 

I added a show on hover functionality.  Please take a look should we go with that

linuspahl:
Regarding the question when to display the breadcrumb, I do not have a strong preference, but I can imagine it will be a little bit easier for the user to orientate, when display it always. My suggestion is to try it out and see how it feels.

I would probably differentiate the "Untitled Search" title slightly by displaying it italic.

maxiadlovskii:
@dennisoelkers @linuspahl I added breadcrumbs to the unsaved view. Please take a look if you like that

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,109 @@
...
+font-size: 0.50rem;
+`;
+
+const ViewHeader = () => {
+  const { view } = useStore(ViewStore);
+  const isSavedView = view?.id && view?.title;
+  const [showMetadataEdit, setShowMetadataEdit] = useState<boolean>(false);
+  const toggleMetadataEdit = useCallback(() => setShowMetadataEdit((cur) => !cur), [setShowMetadataEdit]);
+
+  const typeText = useMemo<string>(() => view.type.toLocaleLowerCase(), [view]);
```
linuspahl, line 77: Do you see a benefit in using `useMemo` here?

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,109 @@
...
+`);
+const EditButton = styled.div(({ theme }) => css`
+  color: ${theme.colors.gray[60]};
+  font-size: ${theme.fonts.size.tiny}
+`);
+const TitleWrapper = styled.span`
+  display: flex;
+  gap: 4px;
+  align-items: center;
+  cursor: pointer;
```
linuspahl, line 58: I saw that we display a pointer cursor for the view title, but it has no effect when you click on it. Let's use a default cursor instead and only display cursor pointer for the edit icon.

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  return (
+    <Row>
+      <Content>
+        <Link to={links[view.type].link}>
+          {links[view.type].label}
+        </Link>
+        <StyledIcon name="chevron-right" />
+        <TitleWrapper>
+          <span>{view.title || <i>{`Unsaved ${typeText}`}</i>}</span>
```
linuspahl, line 87: Let's use `Unsaved Search` instead of `Unsaved search` for the default search title, to unify it with the title we display in the sidebar.

linuspahl (changes requested):
@maxiadlovskii I like displaying the breadcrumb for the default search.

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
maxiadlovskii: Just only one - do not run `.toLocaleLowerCase()` every time. But maybe it's not so heavy operation. I will remove it from here.

graylog2-web-interface/src/views/components/views/ViewHeader.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  return (
+    <Row>
+      <Content>
+        <Link to={links[view.type].link}>
+          {links[view.type].label}
+        </Link>
+        <StyledIcon name="chevron-right" />
+        <TitleWrapper>
+          <span>{view.title || <i className="unsaved-title">{`Unsaved ${typeText}`}</i>}</span>
```
linuspahl, line 90: We have a helper function we can use here, called `ViewTitle`, which we also use in similar places.

linuspahl (changes requested):
Looks good, just added a comment for one last thing I noticed.

drewmiranda-gl:
🤩

PR #13500, opened 2022-09-20
fixing the directions, Set.difference is used

### LINKED ISSUE #13013: Bug in ViewsResouce.validateIntegrity()
## Current Behavior

`Sets.difference()` works different as expected (checks the sets the wrong way around) so the error message makes no sense, when the `!xxx.containsAll()` happens. 

```        
if(!searchQueries.containsAll(stateQueries)) {
            final Sets.SetView<String> diff = Sets.difference(searchQueries, stateQueries);
            throw new BadRequestException("Search queries do not correspond to view/state queries, missing query IDs: " + diff);
        }
```
and 
```       
if(!searchTypes.containsAll(stateTypes)) {
            final Sets.SetView<String> diff = Sets.difference(searchTypes, stateTypes);
            throw new BadRequestException("Search types do not correspond to view/search types, missing searches: " + diff);
        }
```

## Possible Solution

Found by a customer working with an existing dashboard: I actually want to propose that we think about the Exception. In the past, this was not validated and the view was usable, even if the config was invalid. Now, with the exception, customers with a wrong state in the db can no longer update this view.
We should either convert this to a warning or put in the code that fixes the problem before saving.

Issue #8257 as a reference why we introduced the check.

## Your Environment

* Graylog Version: change from 4.2 to 4.3
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

janheise:
## Description
The views resource checks for missing elements but the parameters of Sets.difference was used wrong way around.

***Note: This needs to be backported to 4.3 to help solving a customer issue.***

fixes #13013 

## Types of changes

## Checklist:


todvora (changes requested):
Could we add a test to verify and preserve the behaviour? Thanks!

todvora (approved):

PR #13514, opened 2022-09-21
Invalidate registration cache only for affected sidecars

thll:
When a configuration is saved, and its tags have changed, we used to invalidate the registration cache for all sidecars.

We now check which sidecars are potentially affected by the change by looking at their tags, and invalidate the cache only for those.

There is still the case where a change to the tags of a configuration doesn't result in a change of the assignment. E.g. if a sidecar has the tags `["apache-logs", "dns-logs"]` and we change the tags of a configuration from `["apache-logs"]` to `["dns-logs"]`. In that case, we wouldn't need to invalidate the cache because the config will still be assigned after the change. But in favour of keeping things simple, we don't try to prevent this.


thll:
Needs adjustment after https://github.com/Graylog2/graylog2-server/pull/13504

mpfz0r (changes requested):
we could add this optimization also to `createConfiguration()`

mpfz0r:
other than that looks great :+1:

mpfz0r (approved):

PR #13523, opened 2022-09-22
Type property shouldn't be serialized twice in parameters and their b…

luk-kaminski:
…inds

## Description
Type property shouldn't be serialized twice in parameters and their bindings.

## Motivation and Context
Parameters are currently serialized like that:
{
   **"type":"value-parameter-v1",**
   "name":"param",
   "title":null,
   "description":null,
   "optional":false,
   "binding":{
      **"type":"value",**
      "value":"message",
      **"type":"value"**
   },
   **"type":"value-parameter-v1",**
   "data_type":"any",
   "default_value":null
}
Type property needs to be present only once!

## How Has This Been Tested?
Unit test has been added (in corresponding enterprise module).

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/shared/utilities/StringUtils.java
```diff
@@ -20,9 +20,23 @@
...
 
-    private StringUtils() { }
+    private StringUtils() {
+    }
 
     public static String f(String format, Object... args) {
         return String.format(Locale.ENGLISH, format, args);
     }
+
+    public static boolean containsOnce(final String mainString, final String subString) {
```
todvora, line 30: Why do you need this method? Only usages I could find are in the test of this util class.

todvora (changes requested):

graylog2-server/src/main/java/org/graylog2/shared/utilities/StringUtils.java
luk-kaminski: Good catch! It is not needed anymore after changes made in the enterprise part.
Thank you!

todvora (approved):
Looks good, thanks!

PR #13541, opened 2022-09-26
Update to jackson 2.13

patrickmann:
Upgrade to the current Jackson version 2.13.4, while keeping everything else unchanged (as much as possible).
https://github.com/Graylog2/graylog-plugin-enterprise/issues/3674

**Details**

- If configuration of `ObjectMapper` is modified after first usage, changes may or may not take effect, and configuration calls themselves may fail [[link](https://fasterxml.github.io/jackson-databind/javadoc/2.13/com/fasterxml/jackson/databind/ObjectMapper.html)]. This behavior appears to have changed since version 2.9. It only affects a few tests which attempt to modify the mapper:
_graylog2-server/src/test/java/org/graylog/events/contentpack/facade/EventDefinitionFacadeTest.java
graylog2-server/src/test/java/org/graylog/events/contentpack/facade/NotificationFacadeTest.java_

- `DateTimeFormatter` now throws an exception for empty string, resulting in a JSON missing node that needs to be handled:
_graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java_

- Jackson 2.11 changed the default `DateTime` formatting for timezone offset to `hh:mm` instead of `hhmm` [[link](https://github.com/FasterXML/jackson-databind/issues/2643)]. We need to explicitly set the legacy format in the `ObjectMapper`.
_graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ObjectMapperProvider.java
graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java
graylog2-server/src/test/java/org/graylog2/jackson/MongoJodaDateTimeSerializerTest.java
graylog2-server/src/test/java/org/graylog2/jackson/MongoZonedDateTimeSerializerTest.java_

- Jackson no longer includes Joda date/time by default. The Joda module needs to be explicitly registered:
_graylog2-server/src/test/java/org/graylog2/plugin/MessageSummaryTest.java_

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4099,Graylog2/graylog-plugin-enterprise-integrations#902

### DIFF
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -181,8 +181,8 @@
         <dependency>
             <!-- Workaround for https://github.com/FasterXML/jackson-bom/pull/38 -->
             <groupId>com.fasterxml.jackson.module</groupId>
-            <artifactId>jackson-module-scala_2.13</artifactId>
-            <version>2.9.10</version>
+            <artifactId>jackson-module-scala_3</artifactId>
+            <version>2.13.4</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -179,6 +179,11 @@
             <artifactId>jackson-module-jsonSchema</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-scala_3</artifactId>
+            <version>2.13.4</version>
+        </dependency>
+        <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
             <artifactId>jackson-jaxrs-json-provider</artifactId>
         </dependency>
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -179,12 +179,6 @@
             <artifactId>jackson-module-jsonSchema</artifactId>
         </dependency>
         <dependency>
-            <!-- Workaround for https://github.com/FasterXML/jackson-bom/pull/38 -->
-            <groupId>com.fasterxml.jackson.module</groupId>
-            <artifactId>jackson-module-scala_3</artifactId>
-            <version>2.13.4</version>
-        </dependency>
-        <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
             <artifactId>jackson-jaxrs-json-provider</artifactId>
         </dependency>
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -179,11 +179,6 @@
             <artifactId>jackson-module-jsonSchema</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.module</groupId>
-            <artifactId>jackson-module-scala_3</artifactId>
-            <version>2.13.4</version>
-        </dependency>
-        <dependency>
             <groupId>com.fasterxml.jackson.jaxrs</groupId>
             <artifactId>jackson-jaxrs-json-provider</artifactId>
         </dependency>
--- a/graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java
+++ b/graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java
@@ -18,7 +18,6 @@ package org.graylog.plugins.views.migrations;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.util.StdDateFormat;
 import com.mongodb.BasicDBObject;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoDatabase;
@@ -74,8 +73,7 @@ public class V20200204122000_MigrateUntypedViewsToDashboardsTest {
     @Mock
     private ClusterConfigService clusterConfigService;
 
-    private final ObjectMapper objectMapper = new ObjectMapperProvider().get()
-            .setDateFormat(new StdDateFormat().withColonInTimeZone(false));
+    private final ObjectMapper objectMapper = new ObjectMapperProvider().get();
 
     private Migration migration;
     private MongoCollection<Document> viewsCollection;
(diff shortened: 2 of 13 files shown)


thll:
> * `DateTimeFormatter` now throws an exception for empty string, resulting in a JSON missing node that needs to be handled:
>   _graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java_

I don't think that `DateTimeFormatter` is the culprit here, but rather `ObjectMapper.readTree(input)` not returning `null` for empty input anymore: https://github.com/FasterXML/jackson-databind/issues/2211.

graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java
```diff
@@ -73,7 +74,8 @@ public class V20200204122000_MigrateUntypedViewsToDashboardsTest {
     @Mock
     private ClusterConfigService clusterConfigService;
 
-    private final ObjectMapper objectMapper = new ObjectMapperProvider().get();
+    private final ObjectMapper objectMapper = new ObjectMapperProvider().get()
+            .setDateFormat(new StdDateFormat().withColonInTimeZone(false));
```
thll, line 78: Not needed.

graylog2-server/src/test/java/org/graylog2/jackson/MongoJodaDateTimeSerializerTest.java
```diff
@@ -26,7 +27,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class MongoJodaDateTimeSerializerTest {
-    private final ObjectMapper objectMapper = new ObjectMapperProvider().get();
+    private final ObjectMapper objectMapper = new ObjectMapperProvider().get()
+            .setDateFormat(new StdDateFormat().withColonInTimeZone(false));
```
thll, line 31: Not needed.

graylog2-server/src/test/java/org/graylog2/jackson/MongoZonedDateTimeSerializerTest.java
```diff
@@ -27,7 +28,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class MongoZonedDateTimeSerializerTest {
-    private final ObjectMapper objectMapper = new ObjectMapperProvider().get();
+    private final ObjectMapper objectMapper = new ObjectMapperProvider().get()
+            .setDateFormat(new StdDateFormat().withColonInTimeZone(false));
```
thll, line 32: Not needed.

thll (changes requested):

graylog2-server/pom.xml
```diff
@@ -181,8 +181,8 @@
         <dependency>
             <!-- Workaround for https://github.com/FasterXML/jackson-bom/pull/38 -->
             <groupId>com.fasterxml.jackson.module</groupId>
-            <artifactId>jackson-module-scala_2.13</artifactId>
-            <version>2.9.10</version>
+            <artifactId>jackson-module-scala_3</artifactId>
+            <version>2.13.4</version>
         </dependency>
```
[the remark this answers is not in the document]
patrickmann: We don't. I was just updating versions without looking any deeper. 
Removed it.

graylog2-server/src/test/java/org/graylog/plugins/views/migrations/V20200204122000_MigrateUntypedViewsToDashboardsTest.java
patrickmann: I must have been testing this before I put the setting in ObjectMapperProvider.
Removed.

graylog2-server/src/test/java/org/graylog2/plugin/MessageSummaryTest.java
```diff
@@ -91,7 +91,7 @@ public void testGetFields() throws Exception {
 
     @Test
     public void testJSONSerialization() throws Exception {
-        final ObjectMapper mapper = new ObjectMapper();
+        final ObjectMapper mapper = new ObjectMapper().registerModule(new JodaModule());
```
mpfz0r, line 94: hmm, the test passes without this change. but wouldn't it be better if all our tests used the objectmapperprovider instead?

mpfz0r:
> > * `DateTimeFormatter` now throws an exception for empty string, resulting in a JSON missing node that needs to be handled:
> >   _graylog2-server/src/main/java/org/graylog/plugins/beats/Beats2Codec.java_
> 
> I don't think that `DateTimeFormatter` is the culprit here, but rather `ObjectMapper.readTree(input)` not returning `null` for empty input anymore: [FasterXML/jackson-databind#2211](https://github.com/FasterXML/jackson-databind/issues/2211).

Hah, so they changed it from throwing an exception to returning null and then to returning an empty node object?

graylog2-server/src/test/java/org/graylog2/plugin/MessageSummaryTest.java
mpfz0r: > the test passes without this change
that was a mistake. it doesn't pass. still using the objectmapper is better :+1: 
it actually exposes an error in the test now. it should serialize as `stream_ids` which is correct

graylog2-server/pom.xml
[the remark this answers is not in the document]
patrickmann: Getting a dependency error running the integration tests, so I am reverting the reversion of scala 3.
> com.fasterxml.jackson.databind.JsonMappingException: Scala module 2.10.5 requires Jackson Databind version >= 2.10.0 and < 2.11.0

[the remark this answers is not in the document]
bernd: @patrickmann Is this a runtime dependency, or can we put it into the test scope?

The stack trace in the test output seems to be rooted in the rest-assured dependency, a testing library.

The `jackson-module-scala_3` dependency should also be present in the jackson-bom now. That means we don't need the hardcoded `<version>` tag for it.

bernd (changes requested):
See: https://github.com/Graylog2/graylog2-server/pull/13541#discussion_r992434942

mpfz0r (approved):
LGTM :+1:

bernd (changes requested):
The jackson-bom got bumped to `2.13.4.20221013` to fix a jackson-databind security issue: https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.13#micro-patches. Let's update to that version. :slightly_smiling_face:

mpfz0r:
Jackson, a gift that keeps on giving :-/

thll (approved):

bernd (approved):

PR #2443, opened 2016-07-01
Remove deprecated MongoDB metrics reporter

joschi:
The integrated MongoDB metrics reporter has been deprecated for quite some time now and should be removed.

If users really depend on functionality like this, they can use https://github.com/Graylog2/graylog-plugin-metrics-reporter/tree/master/metrics-reporter-mongodb as replacement.

PR #13553, opened 2022-09-27
Remove unused and dysfunctional MetricsHistoryResource

bernd:
This resource was missed during the metrics history removal in #2443.

Refs #2443


dennisoelkers (approved):
:pray:

Issue #13228, opened 2022-08-11
Alerts are only generated if another one is being created


Filisimus:
Hi guys,

first I would like to provide some background information. We want to set up a honeypot and as a part of it there is a CMS vulnerable to the Log4Shell exploit. If someone tries to use this exploit a system message is being sent and Graylog should send an email notification to certain people.

So the expected behaviour is that Graylog gets the exploit notification and pretty much immediately sends the email. But instead the following happens (using timestamps for better understanding):

00:00:00 exploit is being used and message with timetamp is saved in logfile
00:00:00 filebeat sends message with timestamp '00:00:00' to Graylog
00:00:00 message can be found via search in Graylog (also with timestamp 00:00:00)

(At this point the event should be triggered and an email being sent but nothing happens)

00:05:00 exploit is being used again
00:05:00 filebeat sends message with timestamp '00:05:00' to Graylog
00:05:00 message can be found via search in Graylog (also with timestamp 00:05:00)
And now the alert for the '00:00:00' appears and the email, which should have been sent at 00:00:00, is being sent now

I can reproduce this easily. I also tried to use different settings for the filter in events but no success. It is probably worth mentioning that this is not related to a single alarm - we have multiple different alerts and this bevahiour also occurs with mixed alarms, e. g. alarm 1 should be triggered but is only triggered after alarm 2 comes up. Lets take a look at 3 screenshots which should explain this better:

![image](https://user-images.githubusercontent.com/55972194/184122046-f4ca310e-dd60-4d6f-bcce-1fdd7edb31dc.png)

Here you can see that the email was received 13:03 but the timestamp is 12:49.

![image](https://user-images.githubusercontent.com/55972194/184122483-712d1484-c9b5-4398-8fd0-d55ea09fcb66.png)

Here is the next email I received, arrival was 13:08 but the timestamp is 13:03

Here is the overview for the alerts. You will also notice that the alert for 13:08:08 should have been a MySQL message but in the email it says Log4J exploit:

![image](https://user-images.githubusercontent.com/55972194/184125068-070b35c3-1ec2-4caa-9532-2f2fe4dacd3a.png)

Also the email delivery is not the issue - if an alert is being generated the email is being sent immediately. 

![image](https://user-images.githubusercontent.com/55972194/184123028-7d201c42-431e-4b2b-a56a-b2becc7ac736.png)

Here you have my filter settings. 

TLDR:
[...]

PR #13556, opened 2022-09-27
fix alert system on low message input

### LINKED ISSUE #6770: Alerts not Triggering
We have a 4 node cluster setup and alerts are not triggering unless we have a sustained message  processing of >20 per second.

## Expected Behavior
When an event comes in an alert is triggered as soon as the message comes in.  In our case an email alert when a message contains "Started Network Time Service".

## Current Behavior
When the message comes in we may or may not receive an alert depending on how chatty our systems are being to our Graylog instance.  It can range from a minute later to whenever graylog gets busy enough to send it.  We've had the alert come in weeks later.  Our cluster is up and running but not production.   When setting up the Event Definition we can see the message in the Filter Preview but no alert is ever generated until our graylog server gets busy.  When we enable the Random HTTP Message Generator then the alerts will come through in less than a minute.

## Possible Solution
Fix what ever process logic was added in to version 3.1 to alerts.  We can't have an alert missed in the middle of the night when our systems are less chatty.  That just doesn't make sense for an alert system.  Continuously spamming our Graylog server is not ideal or helpful.

1. Setup 4 node cluster
2. Forward a servers's Syslog to Graylog
3. Create an Event Definition and Alert
4. Trigger the Event and watch the Alert is never created
5. Enable the Random HTTP Message Generator and the alert will come through shortly after.

## Context
We need proper alerting for events that come in when they come in.

## Your Environment
4 Node Centos 7.5 Cluster

* Graylog Version: 3.1.3
* Elasticsearch Version: 6.8.3-1
* MongoDB Version: 4.0.12-1
* Operating System:  CentOS Linux release 7.5.1804
* Browser version: Any
[server.txt](https://github.com/Graylog2/graylog2-server/files/3838157/server.txt)

I've attached our server config (from the master)

There are posts related to this [here](https://community.graylog.org/t/alerts-events-do-not-trigger/11792/48) and [here](https://github.com/Graylog2/graylog2-server/issues/6453)

AntonEbel:
## Description

In our current implementation of the alert system, there is a check that is responsible for ensuring that all nodes have indexed their messages by a certain time. Only if this check is successful, we create events. However, there are two scenarios in which the alert system does not create events in time. 

- The first one is when we have multiple nodes and the message rate decreases. In this case we always select the post indexing time of the node with the oldest post indexing time and use it for the previously mentioned check. The danger here is that one node not receiving messages for some time would be enough to prevent the alert system from creating events, since the post-indexing time of that node would always be used at a low message rate.

- The second scenario is when a message arrives that leads to an event, and no other message arrives for some time afterwards. In this case, the event would be created too late. Like in this issue #13228.  The following image illustrates this. 
AlertSystem

This PR fixes the above two cases, but uses a different approach. The general approach is to determine if there are busy nodes in the cluster, and if there are, wait for them to index the messages until a certain time.

close #6770

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4194


graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorDependencyCheck.java
```diff
@@ -62,14 +65,10 @@ public boolean canProcessTimerange(DateTime maxTimestamp, Set<String> processorD
...
+     * @param timeRange the timestamp to check
      * @return true if messages up to the given latestTimestamp have already been indexed, false otherwise
      */
-    public boolean hasMessagesIndexedUpTo(DateTime latestTimestamp) {
-        // If there is no timestamp returned, there is no node in the cluster that is processing any messages.
-        // In that case we return false because there is nothing to process.
-        return processingStatusService.earliestPostIndexingTimestamp()
-                .map(latestPostIndexTimestamp -> latestPostIndexTimestamp.isAfter(latestTimestamp) || latestPostIndexTimestamp.isEqual(latestTimestamp))
-                .orElse(false);
+    public boolean hasMessagesIndexedUpTo(TimeRange timeRange) {
```
mpfz0r, line 71: Change this signature would need another PR for the new SigmaEventProcessor.
But do we need that? (see below)

graylog2-server/src/main/java/org/graylog2/system/processing/DBProcessingStatusService.java
```diff
@@ -195,4 +140,51 @@ ProcessingStatusDto save(ProcessingStatusRecorder processingStatusRecorder, Date
...
+            int behindNodes = 0;
+            while (statusCursor.hasNext()) {
+                activeNodes++;
+                ProcessingStatusDto nodeProcessingStatus = statusCursor.next();
+                DateTime postIndexing = nodeProcessingStatus.receiveTimes().postIndexing();
+                if (postIndexingBehind(timeRange.getTo(), postIndexing)) {
+                    if (isBusy(nodeProcessingStatus)) {
+                        return ProcessingNodesState.SOME_BUSY;
+                    }
+                    if (postIndexing.isBefore(timeRange.getFrom())) {
```
mpfz0r, line 159: I don't understand why we are comparing the beginning of the timerange here.
Maybe add some comments to the code?

graylog2-server/src/main/java/org/graylog2/system/processing/DBProcessingStatusService.java
```diff
@@ -195,4 +140,51 @@ ProcessingStatusDto save(ProcessingStatusRecorder processingStatusRecorder, Date
...
+        }
+
+        return ProcessingNodesState.ALL_UP_TO_DATE;
+    }
+
+    private boolean isBusy(ProcessingStatusDto nodeProcessingStatus) {
+        return nodeProcessingStatus.inputJournal().uncommittedEntries() > 0L;
+    }
+
+    private boolean postIndexingBehind(DateTime timeRangeTo, DateTime postIndexing) {
```
mpfz0r, line 176: I would just inline this method. I found it easier to understand if I see that `.isAfter` call directly

mpfz0r (changes requested):
First round of review comments

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorDependencyCheck.java
AntonEbel: https://github.com/Graylog2/graylog-plugin-enterprise/pull/4194

mpfz0r (approved):

PR #13604, opened 2022-10-04
fix: Fixes issues active page update and deleting all rows in a page

zeeklop:
Fix small bugs on `PaginatedList.tsx`

## Description

- Update internal `currentPage` when `activePage` changes.
- Change to immediate previous page when the current page doesn't have any items to show.
- Fix issue when changing page size and the number of pages would stay the same. 

## Motivation and Context

While working with the component `PaginatedList` I noticed a few inconsistencies when parameters would change from the parent component. This small fix is to make this component to act in a more predictable way.

## How Has This Been Tested?

I've tested the use of this component manually and with unit tests.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/common/PaginatedList.tsx
```diff
@@ -60,38 +59,39 @@ const PaginatedList = ({
...
+  const [{ currentPage, currentPageSize }, setPagination] = React.useState({
+    currentPage: useQueryParameter ? page : Math.max(activePage, INITIAL_PAGE),
+    currentPageSize: (useQueryParameter && showPageSizeSelect) ? queryParamPageSize : propPageSize,
+  });
+
+  const numberPages = React.useMemo(() => (
+    currentPageSize > 0 ? Math.ceil(totalItems / currentPageSize) : 0
+  ), [currentPageSize, totalItems]);
 
   const _onChangePageSize = (event: React.ChangeEvent<HTMLOptionElement>) => {
```
linuspahl, line 72: Let's use `useCallback` here as well?

linuspahl (changes requested):
I found a problem which also exists in the previous version, but maybe we can fix it in this PR?

When you change the values for `DEFAULT_PAGE_SIZES`, for example from `[10, 50, 100]` to `[5, 10, 50, 100]`, it selects 5 as the page size by default, but it still lists more items. I tested it with the dashboards list.

graylog2-web-interface/src/components/common/PaginatedList.tsx
zeeklop: sounds good. I'll take a look

zeeklop:
> 

Sure thing. I think I have an idea why this is happening. I'll take a look. Thanks

zeeklop:
> I found a problem which also exists in the previous version, but maybe we can fix it in this PR?
> 
> When you change the values for `DEFAULT_PAGE_SIZES`, for example from `[10, 50, 100]` to `[5, 10, 50, 100]`, it selects 5 as the page size by default, but it still lists more items. I tested it with the dashboards list.

@linuspahl 
After further analysis, I found out the issue is with the hook implementation. I tried a few different solutions for this issue. The easiest one is to not check for the existence of the page size inside the hook. Without this check, users would be able to set the parameter to any number they want, which in some cases could crash the application.

In my opinion, this issue, should be addressed in a different PR since it is not related to the paginated list component. I can open a new issue to work on it.

linuspahl (approved):
Looks good to me.

Issue #13101, opened 2022-07-21
Suggestions/auto-completions for query input are not themed for dark mode


dennisoelkers:
## Current Behavior

![image](https://user-images.githubusercontent.com/41929/180222445-c496414f-6f40-4684-83c3-bbcac9b2acdf.png)

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

PR #13612, opened 2022-10-05
Fix auto completions for dark mode

mikedklein:
Issue #13101 Fix auto-complete/suggestions to be consistent in dark mode.

## Description
- Updated `GlobalThemeStyles.js` to override default styles and use our theme object
- Updated opacify util to function properly
- Removed unused type from Opacify (looked incorrect as it was)

## Motivation and Context
Inconsistent UI in dark mode for our ACE editor auto-complete/suggestions

## How Has This Been Tested?
Tested locally

image
image (1)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/theme/utils/opacify.ts
+++ b/graylog2-web-interface/src/theme/utils/opacify.ts
@@ -16,11 +16,9 @@
  */
 import chroma from 'chroma-js';
 
-export type Opacify = {
-  (string, number): string,
-};
+export type Opacify = (color: string, amount:number) => string;
 
-function opacify(color: string, amount: number): string {
+const opacify: Opacify = (color, amount) => {
   /**
    * Increases the opacity of a color. Its range for the amount is between 0 to 1.
    *
@@ -33,9 +31,14 @@ function opacify(color: string, amount: number): string {
   }
 
   const parsedAlpha = chroma(color).alpha();
-  const newAlpha = (parsedAlpha * 100 + amount * 100) / 100;
+
+  /*
+   * NOTE: the reason for the over complication of the math is to account for rounding issues.
+   * example: without the multiplier and divisor 1 - .7 returns 0.30000000000000004
+  */
+  const newAlpha = ((parsedAlpha * 100) - (amount * 100)) / 100;
 
   return chroma(color).alpha(newAlpha).css();
-}
+};
 
 export default opacify;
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/theme/utils/index.ts
```diff
@@ -20,15 +20,13 @@ import type { ColorLevel } from './colorLevel';
...
 import contrastingColor from './contrastingColor';
-import type { Opacify } from './opacify';
 import opacify from './opacify';
 import type { ReadableColor } from './readableColor';
 import readableColor from './readableColor';
 
 export type Utils = {
   colorLevel: ColorLevel,
   contrastingColor: ContrastingColor,
-  opacify: Opacify,
```
dennisoelkers, line 31: Are we removing this on purpose? This would break when `GlobalThemeStyles.js` is migrated to TS. Maybe we should fix the typings instead?

graylog2-web-interface/src/theme/utils/opacify.ts
```diff
@@ -32,10 +28,7 @@ function opacify(color: string, amount: number): string {
     return color;
   }
 
-  const parsedAlpha = chroma(color).alpha();
-  const newAlpha = (parsedAlpha * 100 + amount * 100) / 100;
-
-  return chroma(color).alpha(newAlpha).css();
+  return chroma(color).alpha(amount).css();
```
dennisoelkers, line 31: This is now just setting the alpha for the color instead of increasing it. We should either fix the comment or adjust it to actually increast it.

dennisoelkers (changes requested):
Looks good, works fine! Added some questions/suggestions inline.

graylog2-web-interface/src/theme/utils/opacify.ts
mikedklein: I was pretty confused on this one because it seemingly did nothing with that calculation, no matter what color I tried it on. I will adjust the comment I think it makes more sense to set the alpha, especially when you are working with hex colors.

dennisoelkers: The calculation is overly convoluted, it could be simplified to:
```suggestion
  const newAlpha = parsedAlpha + amount;
```
So it is just adding the provided delta to the current alpha of the color. I am not sure if that is more useful. Did you check other usages of the function throughout the code? From what I can see, it is not used?

dennisoelkers (approved):
Looks good, works great - thanks a lot!

PR #13627, opened 2022-10-07
Migrating local to global pivot limits in aggregation widgets.

dennisoelkers:
**Note:** This requires #13610 to be merged before.

## Motivation and Context

This PR is based on #13610, which replaces our aggregation strategy so that it uses a global pivot limit instead of one configured for each pivot field. To support this, this PR is now changing the data model in a corresponding way and adds a migration to change objects stored in MongoDB to the new format.

We cannot migrate local pivot limits to global ones 1:1. The assumption
is that in most cases users want a global limit and have picked their
local limits to support this, e.g. 

  - most users leaving the default of 15 for all pivots
  - some users increasing the limit to see more results
  - _assumption_: no users need exactly a limit of n for field 1 and m for field 2

Therefore, we are picking new limits by retrieving existing local limits and finding their maximum, using this for the global limit.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/views/search/views/widgets/aggregation/PivotDTO.java
```diff
@@ -23,13 +23,15 @@
 import com.google.auto.value.AutoValue;
 import org.graylog.autovalue.WithBeanGetter;
 
+import java.util.OptionalInt;
```
janheise, line 26: obsolete import

graylog2-server/src/main/java/org/graylog2/migrations/V20220929145442_MigratePivotLimitsInViews.java
```diff
@@ -0,0 +1,149 @@
...
+                update,
+                new UpdateOptions().upsert(false)
+        );
+    }
+
+    private Optional<Integer> extractMaxLimit(List<Document> pivots) {
+        return pivots.stream()
+                .filter(pivot -> "values".equals(pivot.get("type")))
+                .map(pivot -> pivot.get("config", Document.class))
+                .map(pivotConfig -> pivotConfig != null ? pivotConfig.getInteger("limit") : null)
```
janheise, line 141: could be first filtered for null, too - instead of the 'pivotConfig != null`. The second filter is still needed (not that it get's deleted by accident if this suggestion is implemented)

janheise (approved):

PR #13629, opened 2022-10-07
generate cluster id in migration instead of periodical

AntonEbel:
## Description

Make `ClusterIdGeneratorPeriodical` a migration, as it has to run earlier in our bootstrap process to fix an issue in the cloud environment.

## Motivation and Context

close Graylog2/graylog-plugin-cloud#1054


moesterheld (approved):
approved by @kodjo-anipah

PR #13640, opened 2022-10-10
Fixing latest aggregation if field is not present in time range.

### LINKED ISSUE #13593: Aggregation "Latest Value" without value breaks dashboard
If you add an aggregation with "Latest Value" and choose a field, you see the latest value (as intended). However, if no value for this field exists in the current selected time interval, ALL dashboard widgets error out with "While retrieving data for this widget, the following error(s) occurred: 0."
I think, only the affected widget should error out...

## Error in all widgets
image

## What it looks if all goes well
image

## Settings
image

dennisoelkers:
## Motivation and Context

This PR is fixing the `latest` aggregation for cases where no document with the field is present in the time range.

Code-wise, it checks if the list of hits returned has at least one element before trying to extract it.

Fixes #13593.

## Types of changes

## Checklist:


todvora (approved):
LGTM, thanks!

PR #13656, opened 2022-10-11
Cleanup theme font sizes.

linuspahl:
## Motivation and Context

We are using a type scale for our font sizes, based on https://type-scale.com/.
In the past we tried out a few different variations. With this PR we are cleaning up some errors that slipped in.


grotlue (approved):
Looks good! 👍

PR #13677, opened 2022-10-13
Improve page subarea navigation.

linuspahl:
## Motivation and Context

With this PR we are improving the way users navigate to page subareas.
Before we often displayed three button in the top right corner.

image

Using buttons for this case is not ideal, instead we are now displaying a tab navigation above the page header.

image

The placement and tab style helps the user to understand the hierarchy and relation of these subareas. It is also easier to see the active area.

Please note, usually the active state work well. For example, when a navigation item links to `/alerts/definitions` and the current path is `/alerts/definitions/new`, the nav item will be active. There are a few cases where this does not work perfectly. For example for on the lookup table create page.

Please also note, when testing this PR the primary color for the active nav item will look different. We will change the color in a follow-up PR.

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/4178


maxiadlovskii (approved):

dennisoelkers (approved):
Looks good, works great!

PR #13681, opened 2022-10-13
Define new colors for search query syntax highlighting.

linuspahl:
## Motivation and Context

With this PR we are adjusting the search query syntax highlighting colors. We tried to:
- color the operators in a way that they clearly separate the resulting blocks
- color the field values in a way that have more focus and it is easier to extract them / understand which messages the searchtargeting.
- color the field names in a way that they are more subtle.

Before (4.3)
image

After 
image


maxiadlovskii (approved):

linuspahl:
We decided to close this PR because the contrast of these colors is not good enough, we will stick with the current colors for now.

PR #13689, opened 2022-10-14
Removing `c3` mock from tests.

dennisoelkers:
## Motivation and Context

While debugging an issue where a module was not loaded in tests, I found out that the mock for the `c3` module is specified in a way, that it replaces all modules imported in tests which _contain_ the string `c3` at any place in their import specifier will be replaced with an empty module.

Before fixing the mock specification, I quickly checked if the mock is still required at all, and it looks like it is not. Therefore, instead of fixing up the mock specification, the mock is removed completely.

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me!

PR #13707, opened 2022-10-17
default retention time & max value via config

kodjo-anipah:
## Description

Added following config flags: default_archive_retention_time (to set retention time on new installations, 0 or no setting means feature is off)  & max_archive_retention_time (limit max retention time) 

Graylog2/graylog-plugin-enterprise#4113

## Motivation and Context

With https://github.com/Graylog2/graylog-plugin-enterprise/issues/3514 we introduced a feature to set a retentionTime on archive configurations (per index set), to allow an automatic clean up of old archives.

What would be useful in addition is to support setting a default and a max retention value (in the server config), so that a Graylog administrator could set these.

The retentionTime value entered by the user in the UI (or via the API) should be validated against a configured max value.
## How Has This Been Tested?

Locally with max set, default set and later on overwritting it via ui
## Types of changes

## Checklist:

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -137,3 +137,7 @@ The Page permissions remain unchanged but this could affect the workflow for use
 On new Graylog installations, the default archiving configuration will now 
 store archives under the `data_dir` instead of `/tmp/graylog-archives`. 
 (The `data_dir` is configured in graylog.conf and defaults to `/var/lib/graylog-server`)
+
+### Configuring archive retention Time and max value
+It is now possible to configure default archive retention time and a limit via config flags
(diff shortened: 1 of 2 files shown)


patrickmann (commented):
Add the settings in a commented out section of `graylog.conf`, so users can be aware of them and what units to use to specify the duration.

kodjo-anipah:
> Add the settings in a commented out section of `graylog.conf`, so users can be aware of them and what units to use to specify the duration.

@patrickmann  I can do that, not sure if we should for enterprise features thought

patrickmann:
> @patrickmann I can do that, not sure if we should for enterprise features thought

Good point - probably not

PR #13709, opened 2022-10-17
Preventing index field type polling running twice for the same index.

dennisoelkers:
## Motivation and Context

Prior to this change, when polling the field types of an index set took longer than 1 second, a second polling job was submitted to the executor, running in parallel. If it took longer than 2s, a third one is started, and so on.

In order to prevent this, a very simple locking mechanism is implemented, where a running polling job adds the current index set's id to a concurrent hash set, removing it when finished (or failing).

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
@@ -218,6 +218,9 @@ public class IndexFieldTypePollerPeriodical extends Periodical {
     }
 
     private boolean needsFullRefresh() {
+        if (fullRefreshInterval.toSeconds() == 0) {
+            return false;
+        }
         Instant nextFullRefresh = lastFullRefresh.plusSeconds(fullRefreshInterval.toSeconds());
         return !Instant.now().isBefore(nextFullRefresh);
     }
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
@@ -16,9 +16,11 @@
  */
 package org.graylog2.indexer.fieldtypes;
 
+import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodical.java
```diff
@@ -164,20 +168,30 @@ private void poll(Collection<IndexSetConfig> indexSetConfigs) {
...
                 .filter(IndexSetConfig::isWritable)
-                .filter(config -> {
+                .forEach(config -> {
                     final Instant previousPoll = lastPoll.getOrDefault(config.id(), Instant.MIN);
                     final Instant nextPoll = previousPoll.plusSeconds(
                             config.fieldTypeRefreshInterval().getStandardSeconds());
-                    return !Instant.now().isBefore(nextPoll);
-                })
-                .forEach(this::poll);
+                    if (!Instant.now().isBefore(nextPoll)) {
```
luk-kaminski, line 175: Won't `Instant.now().isAfter()` be slightly more readable?

dennisoelkers: I had the same thought. I think the intention was that it should also start when when `now == nextPoll`.

graylog2-server/src/test/java/org/graylog2/indexer/fieldtypes/IndexFieldTypePollerPeriodicalTest.java
```diff
@@ -62,14 +81,55 @@ void setUp() throws Exception {
...
+        final ReentrantLock lock = new ReentrantLock();
+        lock.lock();
+        when(indexFieldTypePoller.pollIndex(anyString(), anyString()))
+                .thenAnswer((Answer<Optional<IndexFieldTypesDTO>>) invocationOnMock -> {
+                    lock.lock();
+                    return Optional.empty();
+                });
+
+        periodical.doRun();
+        periodical.doRun();
```
luk-kaminski, line 130: Those 2 invocations does not seem to be run in different threads, they are run sequentially in the test's thread.
Therefore, in my opinion, lock usage has completely no impact on the test behavior.

dennisoelkers: Good catch! I overlooked that we are using a single thread executor. Fixed that!

luk-kaminski (approved):
Great job, thanks a lot!

PR #13735, opened 2022-10-19
Fail fast if proxied requests would miss auth info

### LINKED ISSUE #13721: Proxied resources don't properly handle trusted header authentication
## Expected Behavior

A call to a proxied resource, like `/api/cluster/metrics/multiple` should properly handle trusted header authentication.
 
## Current Behavior

When the client session expires but trusted headers are used, the call to `/api/cluster/metrics/multiple` is authenticated, but the cluster-internal subsequent calls to `/api/system/metrics/multiple` don't provide any authentication data. They are therefore rejected as _unauthorized_ by the receiving nodes.

This leads to the following effect: When Trusted Header Authentication is enabled, and the session of a user expires, the frontend is still usable. But the global throughput counter drops to zero. Only after the page is reloaded and a new session cookie is included in the requests the throughput counter becomes functional again.

## Possible Solution

- We could try to fix the frontend to not send any requests to any resources other than those for session creation.
- Also, we could fail early in the proxied resource if the original request was authenticated, but we are unable to add authentication information to the federated requests [here](https://github.com/Graylog2/graylog2-server/blob/875c7c2a31b8c11c527be417b2cf0d2072333f83/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java#L68)

1.  Set up a Graylog cluster of at least two nodes
1.  Set up a local Graylog user with a short session expiration (e.g. use 30 seconds)
1. Set up Trusted Header Authentication and make sure to set the header to the Graylog user for every request. Use a proxy or a browser plugin to do that.
1. Open the UI which should log you in automatically. Wait for 30 seconds until the session times out.
1. Notice errors similar to the following in the `server.log`:
```
2022-10-18 15:41:43,596 WARN : org.graylog2.shared.rest.resources.ProxiedResource - Unable to call http://127.0.0.1:9001/api/system/metrics/multiple on node <fd4d8b06-ad11-46e7-ae0b-78a611c56311>, result: Unauthorized
2022-10-18 15:41:43,597 WARN : org.graylog2.shared.rest.resources.ProxiedResource - Unable to call http://127.0.0.1:9000/api/system/metrics/multiple on node <9c0c55fa-548f-4a44-9b8f-1138b3c5c4c3>, result: Unauthorized
```

## Context

[HS-1127509680]

## Your Environment

* Graylog Version: `master`
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

thll:
If a request to a proxied resource is authenticated, but the authentication information cannot be extracted from the headers of the original request, we would send unauthenticated requests to the cluster nodes. In most, if not all cases, these requests will then be rejected by the receiving nodes. 

A scenario, where this will happen, is the "Trusted Header Authentication". See #13721.

With this change, we fail the original requests with a `401 Unauthorized`. Otherwise, the request would look successful from the client's perspective, even though it wasn't.

Fixes #13721.


dennisoelkers (approved):
Looks good and works fine!

PR #13742, opened 2022-10-20
Upgrade notes updated, new property added to sample config file

luk-kaminski:
## Description
Upgrade notes updated, new property added to sample config file


dennisoelkers (commented):
Thanks, looks good! Maybe you can add some helpful advice how to decide if this should be turned on or off by a user. Something along the lines of:

> If all of your streams go to dedicated index sets, turning this off will decrease load on ES/OS.
> If multiple streams go to the same index sets and you want precise field types and suggestions, you should turn this on.

dennisoelkers (approved):
Very good!

PR #13746, opened 2022-10-20
Fix search form width bug

### LINKED ISSUE #13752: SearchFrom UI width bug in Alerts page
Styling problem with the SearchFrom component in /alerts

image

gally47:
## Description
Before one shared prop **queryWidth** was used for both the input width and its container width which caused a bug in /alert page's search form. Now this PR fixes that by adding a different prop for the container's width: **queryContainerWidth**.

## Motivation and Context

fixes #13752

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/sidecars/administration/CollectorConfigurationModal.tsx
+++ b/graylog2-web-interface/src/components/sidecars/administration/CollectorConfigurationModal.tsx
@@ -214,7 +214,11 @@ const CollectorConfigurationModal = ({
         </ModalTitle>
       </Modal.Header>
       <Modal.Body>
-        <SearchForm query={searchQuery} onQueryChange={(q) => setSearchQuery(q)} topMargin={0} queryWidth="100%" />
+        <SearchForm query={searchQuery}
+                    queryWidth="100%"
+                    queryContainerWidth="100%"
+                    onQueryChange={(q) => setSearchQuery(q)}
+                    topMargin={0} />
         <ConfigurationContainer>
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/sidecars/administration/CollectorConfigurationModal.tsx
```diff
@@ -214,7 +214,11 @@ const CollectorConfigurationModal = ({
         </ModalTitle>
       </Modal.Header>
       <Modal.Body>
-        <SearchForm query={searchQuery} onQueryChange={(q) => setSearchQuery(q)} topMargin={0} queryWidth="100%" />
+        <SearchForm query={searchQuery}
```
ousmaneo, line 217: I think here we should use styled-component instead of introducing new props to the `SeachForm` component: 
- We can create a StyledSearchForm as below
```
const StyledSearchForm = styled(SearchForm)`
  .form-group, .query {
    width: 100% !important;
  }
`;
```
- then use it here:
```suggestion
        <StyledSearchForm query={searchQuery}
```

ousmaneo: Then we can remove the passed props below here and also from the `SearchForm` component.

ousmaneo (commented):
Hi thank you for the PR, I made some suggestions inline.
Can you please link the original issue in the description, we can create it if it does not exist already.

ousmaneo (approved):
LGTM!

PR #13762, opened 2022-10-24
Fix flaky test for index field type poller periodical.

dennisoelkers:
## Description

In #13709, a test was added that was supposed to check if the index field type poller periodical is started while another one was already running, it should skip execution when the current index set is already polled. The test was prone to a race condition on very slow environments, when:

  - The periodical was run for the first time and submitted a job
  - The job was executed stuck in indexing
  - The periodical was run once again and submitted a job
  - The second job has not executed yet
  - The first job is released before the second job could check if the
    index set is locked
  - The second job is executing and does a full check again

In this case, the assertion if the index field types are retrieved only once is violated and test fails.

In order to address this, this PR does add two waits:

  - Waiting for the second job to complete before releasing the first one
  - Waiting for the first job to complete before doing the final assertion

## Types of changes

## Checklist:


luk-kaminski (approved):
Looks good, thx!

PR #13778, opened 2022-10-25
Bump required MongoDB version in pre-flight check to 5.0

### LINKED ISSUE #13660: Bump required MongoDB and ES/OpenSearch versions in pre-flight check
In the upcoming Graylog 5.0, we bump the minimum required versions for MongoDB, Elasticsearch, and OpenSearch.

The pre-flight checks need to be adjusted accordingly.

AntonEbel:
Bump only MongoDB version because Elastic Search/Open Search versions are up to date.
closes #13660 

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4241


mpfz0r (approved):

mpfz0r:
@bernd is smd-dev prepared for this change? which mongdb are we running there?

bernd:
> @bernd is smd-dev prepared for this change? which mongdb are we running there?

@mpfz0r Yes, it's running 5.0 already.

coffee-squirrel:
Will MongoDB 5.0 only be supported under Graylog 5.0+?

Currently, based upon support case 1000676139 and docs, our understanding is MongoDB 4.4 is the latest supported for Graylog 4.x.  Just a bit concerned that we'd need to do a MongoDB 4.4 to 5.0 upgrade immediately prior to the Graylog 5.0 upgrade.

kroepke:
@coffee-squirrel Hi! I think that information is simply not updated, our development setups are all running with MongoDB 5.0 (see https://github.com/Graylog2/graylog-project/blob/master/runner/docker/docker-compose.yml#L109-L117) and there have been no problems.
I'll check up on the docs, I think the requirement for up to 4.4 wasn't updated and/or was written too defensively. In general, the drivers support connecting to newer versions, but some newer features might not work. So far we've kept the compatibility level on the MongoDB 3.6 baseline, but that was becoming quite a drag when writing queries.

The recommendation will be: Update to MongoDB 5.0 with Graylog 4.3, then update Graylog to 5.0, then make the decision whether to go to MongoDB 6.0.
The driver version in Graylog went from [3.12.1 in 4.3](https://github.com/Graylog2/graylog2-server/blob/4.3.8/pom.xml#L138) to the latest [4.17.1 in 5.0](https://github.com/Graylog2/graylog2-server/blob/52236ab2bf45d6fe384ef3d71f41282e3149bbf4/pom.xml#L140), if you are interested check out the [MongoDB compatibility matrix](https://www.mongodb.com/docs/drivers/java/sync/current/compatibility/) as well.

We've tested both single node and replica set upgrades from MongoDB 3.6 all the way to 6.0 with Graylog 4.3.8, those all worked fine.

Just for posterity: Graylog 5.0 will fail its preflight check and will not start with MongoDB 4.4 or earlier, if that happens and you cannot directly upgrade MongoDB, simply switch back to Graylog 4.3.8 and start that again, nothing will have touched the database at that point. Then prepare the MongoDB update to at least 5.0 and do the Graylog 4.3->5.0 update again.

PR #13782, opened 2022-10-25
Allow `Show Top Values` action for enumerable compound fields.

dennisoelkers:
## Description

**Note:** This should be backported to `4.3`.

## Motivation and Context

Prior to this PR, the `Show top values` action was disabled when the field it is triggered on is a compound field. This is too strict, because if all of the field types composing this compound field type are enumerable, then it is also possible to show the top values.

This PR is now changing the predicate to check if the field type is enumerable instead of if it is a compound type.

## Types of changes

## Checklist:


linuspahl (approved):

Issue #13773, opened 2022-10-25
Improve styling of active state in main navigation


linuspahl:
## What?

With https://github.com/Graylog2/graylog2-server/pull/13498 we fixed the navigation active state, but its styling is not ideal, especially in dark mode.

<img width="112" alt="image" src="https://user-images.githubusercontent.com/46300478/197712539-8907f3f5-f14d-4539-b11f-9c2a0635e6bb.png">

We should try out different ways to highlight the active page. One option is to use the red color we use for the logo and display a bottom border for the active entry.

PR #13785, opened 2022-10-25
Improve navigation active state indicator styles.

linuspahl:
## Motivation and Context

Currently the active state indicator in the navigation is not ideal, especially with the dark mode version. With this PR we are changing the way we display the active and hover state.

Before:

image

image

After:

image

image

Fixes: https://github.com/Graylog2/graylog2-server/issues/13773


dennisoelkers (approved):
Looks good, works great!

PR #13792, opened 2022-10-26
Removed pivot limits from contentpack

janheise:
## Description

This PR removes the `limits` from the pivots in the given contentpack. The structure changed because of changes in the pivots dto. 
If you start with an empty MongoDB, this leads to errors.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #13796, opened 2022-10-26
Revert "Migrating local to global pivot limits in aggregation widgets" #13627

bernd:
This breaks an empty Graylog. (see #13792 for a fix) But also the SearchAggregationsIT fails locally and doesn't run on CI.

This reverts commit 265e7d62d8ee29f3da8c773a6d29fe071f90b421.


mpfz0r (approved):

bernd:
Tested locally. (`mvn test` and a Graylog setup)

PR #13809, opened 2022-10-27
Prototype of possible solution of Simple Scripting API task

luk-kaminski:
## Description
After discussion with the team, we have clarified slightly the idea behind the Simple Scripting API.

This draft PR represents a prototype/alpha version of my concept of how to solve the problem (and may reveal all my further misunderstandings).
It is functional and can be used from "REST API browser".

Imho, it seems that the most important elements of the task are:
1. Simpler format for Aggregation (the Graylog aggregation, not ES/OS one).
2. Mapper from simple format to Search.
3. Simpler format for Result.
4. Mapper from PivotResult to new, simple format.

There is a chance that the rest could stay untouched, as suggested by Dennis.

Example request (may be the subject of change):
```
{
  "timerange": {
    "type": "relative",
    "range" : 0
  },
  "streams": ["000000000000000000000001"],
  "query": "*",
  "aggregation": {
    "group_by": [
      {
        "limit": 5,
        "field_name": "source"
      },
      {
        "limit": 5,
        "field_name": "http_method"
      }
    ],
    "metrics": [
      {
        "function_name": "avg",
        "field_name": "took_ms"
      },
      {
        "function_name": "min",
        "field_name": "took_ms"
      },
      {
        "function_name": "max",
        "field_name": "took_ms"
      }
    ]
  }
}
```

Example response (may be the subject of change):
```
{
  "schema": [
    {
      "name": "Grouping on : source",
      "type": "string"
    },
    {
      "name": "Grouping on : http_method",
      "type": "string"
    },
    {
      "name": "Metric : avg",
      "type": "numeric"
    },
    {
      "name": "Metric : min",
      "type": "numeric"
    },
    {
      "name": "Metric : max",
      "type": "numeric"
    }
  ],
  "data": {
    "rows": [
      [
        "example.org",
        "GET",
        70.61851214257906,
        36,
        5450
      ],
      [
        "example.org",
        "DELETE",
        93.49547309359858,
        46,
        5850
      ],
      [
        "example.org",
        "POST",
        168.57841166568411,
        122,
        5850
      ],
      [
        "example.org",
        "PUT",
        118.8289678066446,
        71,
        5850
      ],
      [
        "(Empty Value)",
        "-",
        "Infinity",
        "Infinity",
        "-Infinity"
      ]
    ]
  }
}
```

Many elements of the input have reasonable defaults, so that you can invoke the endpoint with a relatively small input object, i.e.:
```
{
  "aggregation": {
    "group_by": [
      {
        "field_name": "source"
      },
      {
        "field_name": "http_method"
      }
    ],
    "metrics": [
      {
        "function_name": "avg",
        "field_name": "took_ms"
      },
      {
        "function_name": "max",
        "field_name": "took_ms",
        "sort" : "Descending"
      },
      {
        "function_name": "min",
        "field_name": "took_ms",
        "sort" : "Ascending"

      }

    ]
  }
}
```


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/SearchRequestSpecToSearchMapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+        if (searchRequestSpec.streams() != null && !searchRequestSpec.streams().isEmpty()) {
+            query = query.addStreamsToFilter(new HashSet<>(searchRequestSpec.streams()));
+        }
+
+        final Search search = Search.builder()
+                .queries(ImmutableSet.of(query))
+                .build();
+
+        search.addStreamsToQueriesWithoutStreams(() -> Set.of(Stream.DEFAULT_STREAM_ID));
```
dennisoelkers, line 61: What is the reason for using the default stream here for queries without streams? I think we should instead use all of the streams the user has access to, similar to how we do it in other place. At this point we also do not know if the user has even access to the default stream.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/ScriptingApiResource.java
```diff
@@ -0,0 +1,119 @@
...
+        postAuditEvent(searchJob);
+
+        //Step 3: take complex response and try to map it to simpler, tabular form
+        final SearchJobDTO searchJobDTO = SearchJobDTO.fromSearchJob(searchJob);
+        final QueryResult queryResult = searchJobDTO.results().get(SearchRequestSpecToSearchMapper.QUERY_ID);
+        if (queryResult != null) {
+            final SearchType.Result aggregationResult = queryResult.searchTypes().get(AggregationSpecToPivotMapper.PIVOT_ID);
+            if (aggregationResult instanceof PivotResult pivotResult) {
+                final TabularResponse tabularResponse = responseCreator.mapToResponse(searchRequestSpec, pivotResult);
+                return Response.ok(tabularResponse).build();
```
dennisoelkers, line 106: We can just return `tabularResponse` in here, that makes it a bit easier to understand from the method signature what's returned. Below, we can throw a `NotFoundException`.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/request/SearchRequestSpec.java
```diff
@@ -0,0 +1,43 @@
...
+                                @JsonProperty("timerange") TimeRange timerange,
+                                @JsonProperty("aggregation") @Valid @NotNull AggregationSpec aggregationSpec) {
+
+    public static final RelativeRange DEFAULT_TIMERANGE = RelativeRange.create(24 * 60 * 60);
+
+    public SearchRequestSpec {
+        if (Strings.isNullOrEmpty(queryString)) {
+            queryString = "*";
+        }
+        if (timerange == null) {
```
dennisoelkers, line 39: Can we do the same for `streams` and assign an empty collection if it is missing? That also makes handling it later easier.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/request/SearchRequestSpec.java
```diff
@@ -0,0 +1,43 @@
...
+import com.google.common.base.Strings;
+import org.graylog2.plugin.indexer.searches.timeranges.RelativeRange;
+import org.graylog2.plugin.indexer.searches.timeranges.TimeRange;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import java.util.List;
+
+public record SearchRequestSpec(@JsonProperty("query") String queryString,
+                                @JsonProperty("streams") List<String> streams,
```
dennisoelkers, line 29: Is there a reason for this to be a `List` instead of a `Set`?

dennisoelkers (changes requested):
Looks good in general, added some suggestions for improvements inline.

luk-kaminski:
> Looks good in general, added some suggestions for improvements inline.

Thank you for your feedback, I'll soon push the changes. :)

dennisoelkers (approved):

PR #13833, opened 2022-10-31
System events

### LINKED ISSUE #12555: System notification should inform via Email alerts to admin role users.
The system notification is shown in System/Overview informs admin role users 
## What?

The specific/admin group of users should get notified of those trigger notifications via Email

Here are the sample notification:
image

## Why?

It will not require a login each time to take necessary actions on such System notifications. The notified user can take action according to the urgency of the notification.

##Context
HS-887267106/HS-880186036

## Your Environment

* Graylog Version: ALL
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

patrickmann:
Resolves #12555 

Creates an event for every system notification and persists it in the default system event stream.

This is planned for 5.0.1 - do not merge prior to 5.0.0 split.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4303


graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessorConfig.java
```diff
@@ -0,0 +1,71 @@
...
+    @Override
+    public ValidationResult validate() {
+        return new ValidationResult();
+    }
+
+    @Override
+    public EventProcessorConfigEntity toContentPackEntity(EntityDescriptorIds entityDescriptorIds) {
+        // TODO: Don't forget to add the correct Jackson subtype mapping, so that the content pack import is successful:
+        //  e.g. registerJacksonSubtype(SigmaEventProcessorConfigEntity.class,
+        //                SigmaEventProcessorConfigEntity.TYPE_NAME);
```
mpfz0r, line 63:
```suggestion
    // Don't export this into content packs
```

mpfz0r: Hmm, on a second thought, we might want to export them, but we need some special handling,
because the entity itself should always exist, we would only update the notification settings via content packs...

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedDbService.java
```diff
@@ -74,16 +90,23 @@ public final void ensureValidScope(E entity) {
...
+            throw new IllegalArgumentException("Non-deletable entity cannot be deleted");
+        }
+    }
+
     @Override
     public final int delete(String id) {
-        ensureMutability(get(id).orElseThrow(() -> new IllegalArgumentException("Entity not found")));
+        final E entity = get(id).orElseThrow(() -> new IllegalArgumentException("Entity not found"));
+        ensureDeletability(entity);
+        ensureMutability(entity);
```
mpfz0r, line 109: are we sure we want to define it this way? or do we maybe want to be able to delete non-mutable entitys?

changelog/unreleased/issue-12555.toml
```diff
@@ -0,0 +1,5 @@
+type = "a"
+message = "Generate an event in the `System Event` stream for every system notification"
+
```
mpfz0r, line 3: This could have some more details in the `details.user` section. Like what can be done with it

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionDto.java
```diff
@@ -114,6 +115,10 @@ public abstract class EventDefinitionDto extends ScopedEntity implements EventDe
     @JsonProperty(FIELD_STORAGE)
     public abstract ImmutableList<EventStorageHandler.Config> storage();
 
+    @Override
+    @JsonProperty(FIELD_IS_SYSTEM_EVENT)
+    public abstract boolean isSystemEvent();
```
mpfz0r, line 120: are we actually using that field?

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessor.java
```diff
@@ -0,0 +1,67 @@
...
+        this.eventDefinition = eventDefinition;
+    }
+
+    @Override
+    public void createEvents(EventFactory eventFactory, EventProcessorParameters processorParameters, EventConsumer<List<EventWithContext>> eventsConsumer) throws EventProcessorException {
+        SystemNotificationEventProcessorParameters eventParameters = (SystemNotificationEventProcessorParameters) processorParameters;
+        LOG.debug("Creating system event for notification: {}", eventParameters.notificationType());
+
+        final ImmutableList.Builder<EventWithContext> eventsWithContext = ImmutableList.builder();
+        final Event event = eventFactory.createEvent(eventDefinition, eventParameters.timestamp(), eventParameters.notificationType().name());
```
mpfz0r, line 58: We should add more details from the systemnotification to the event. 
We are missing the actual message from the Notification for example.

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedDbService.java
patrickmann: There is a separate `deleteImmutable` method. I assume immutable entity should only be deleted via that method.

changelog/unreleased/issue-12555.toml
patrickmann: Good point - added more detail

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessorConfig.java
patrickmann: Per offline discussion we won't export this into content packs at this time.

ousmaneo (changes requested):
I have found some console error other than that it works.

graylog2-web-interface/src/components/event-definitions/event-definition-form/EventConditionForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+    errors: {
+      config: any,
+      title: string,
+    }
+  },
+  currentUser: User,
+  onChange: (name: string, value: string | number) => void,
+}
+
+const EventConditionForm = ({ eventDefinition, validation, currentUser, onChange }: Props) => {
```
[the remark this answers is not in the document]
mnin: Solved this by passing this both props too.

graylog2-server/src/main/java/org/graylog2/migrations/V20190705071400_AddEventIndexSetsMigration.java
```diff
@@ -96,32 +110,35 @@ public void upgrade() {
...
         } catch (NotFoundException ignored) {
             createEventsStream(streamId, streamTitle, streamDescription, eventsIndexSet);
         }
     }
 
-    private void checkIndexPrefixConflicts(String indexPrefix, String configKey) {
+    private void checkIndexPrefixConflicts(String indexPrefix, String configKey, String templateType) {
         final DBQuery.Query query = DBQuery.and(
-                DBQuery.notEquals(IndexSetConfig.FIELD_INDEX_TEMPLATE_TYPE, Optional.of(EVENT_TEMPLATE_TYPE)),
+                DBQuery.notEquals(IndexSetConfig.FIELD_INDEX_TEMPLATE_TYPE, Optional.of(templateType)),
```
mpfz0r, line 141: This only works for new installations. on upgrades, this will fail because the system event index already exists with `template_type` `events`

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
```diff
@@ -0,0 +1,96 @@
...
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.events.processor.systemnotification;
+
+import com.google.common.collect.ImmutableMap;
+import org.graylog2.indexer.EventsIndexMapping7;
+
+public class SystemNotificationEventsIndexMapping7 extends EventsIndexMapping7 {
```
mpfz0r, line 22:
```suggestion
public class SystemEventsIndexMapping extends EventsIndexMapping7 {
```
I would drop the `Notification` from the name, because it will be used for system events.
Also the `7` is unnecessary, because we don't support ES 6 anymore.

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    protected ImmutableMap<String, Object> fieldProperties() {
```
mpfz0r, line 24: you are overriding fieldProperties() but are not adding the alias from `EventsIndexMapping7`
This should happen with a `.putAll(super.fieldProperties())`

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
```diff
@@ -0,0 +1,96 @@
...
+                        .build())
+                .put("fields", map()
+                        .put("type", "object")
+                        .put("dynamic", true)
+                        .build())
+                .put("group_by_fields", map()
+                        .put("type", "object")
+                        .put("dynamic", true)
+                        .build())
+                /* TODO: Enable the typed fields once we decided if that's the way to go
```
mpfz0r, line 71: I think we can remove that TODO from the copy here. One is enough.

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedDbService.java
mpfz0r: Ah, okay. 
Should we change the name from `deleteImmutable` to `forceDelete` or something? Because it can now
also delete deletable :-D

graylog2-server/src/main/java/org/graylog2/migrations/V20190705071400_AddEventIndexSetsMigration.java
```diff
@@ -206,4 +223,27 @@ private void createEventsStream(String streamId, String streamTitle, String stre
...
+                            .description("Reserved event definition for system notification events")
+                            .alert(false)
+                            .priority(1)
+                            .keySpec(ImmutableList.of())
+                            .notificationSettings(EventNotificationSettings.builder()
+                                    .gracePeriodMs(0) // Defaults to 0 in the UI
+                                    .backlogSize(0) // Defaults to 0 in the UI
+                                    .build())
+                            // Empty notifications list by default. The user will specify later in the UI.
+                            .notifications(ImmutableList.<EventNotificationHandler.Config>builder().build())
```
mpfz0r, line 241: do we need that? it should be empty by default
```suggestion

```

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessorParameters.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public abstract Builder toBuilder();
+
+    public static Builder builder() {
+        return Builder.create();
+    }
+
+    @AutoValue.Builder
+    public abstract static class Builder implements EventProcessorParametersWithTimerange.Builder<Builder> {
```
mpfz0r, line 68:
```suggestion
    public abstract static class Builder implements EventProcessorParameter.Builder<Builder> {
```
see above

graylog2-server/src/main/java/org/graylog2/migrations/V20190705071400_AddEventIndexSetsMigration.java
patrickmann: I now check for the upgrade case and delete the indexset (plus associated ranges and field types) and stream if necessary. Then we can create a clean new index set and stream regardless.

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
patrickmann: done

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
patrickmann: done

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventsIndexMapping7.java
patrickmann: done

graylog2-server/src/main/java/org/graylog2/migrations/V20190705071400_AddEventIndexSetsMigration.java
patrickmann: done

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessorParameters.java
```diff
@@ -0,0 +1,97 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+@AutoValue
+@JsonTypeName(SystemNotificationEventProcessorConfig.TYPE_NAME)
+@JsonDeserialize(builder = SystemNotificationEventProcessorParameters.Builder.class)
+public abstract class SystemNotificationEventProcessorParameters implements EventProcessorParametersWithTimerange {
```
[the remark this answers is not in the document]
patrickmann: done

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessorParameters.java
patrickmann: done

graylog2-server/src/main/resources/org/graylog2/freemarker/templates/PLAINTEXT/archiving_summary.ftl
```diff
@@ -0,0 +1,10 @@
+<#if _title>Indices could not be archived yet</#if>
+
+<#if _description><span>
```
mpfz0r, line 3:
```suggestion
<#if _description>
```

graylog2-server/pom.xml
```diff
@@ -722,6 +722,13 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
+            <groupId>net.htmlparser.jericho</groupId>
+            <artifactId>jericho-html</artifactId>
+            <version>3.4</version>
+        </dependency>
+
```
mpfz0r, line 731:
```suggestion
```
we don't need that anymore, right?

patrickmann: Indeed - forgot to remove that

graylog2-server/src/main/java/org/graylog/events/processor/systemnotification/SystemNotificationEventProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+            message += ": " + eventParameters.notificationMessage();
+        }
+        final Event event = eventFactory.createEvent(eventDefinition, eventParameters.timestamp(), message);
+        for (Map.Entry<String, Object> entry: eventParameters.notificationDetails().entrySet()) {
+            event.setField(entry.getKey(), FieldValue.builder().dataType(FieldValueType.STRING).value(entry.getValue().toString()).build());
+        }
+
+        final ImmutableList.Builder<EventWithContext> eventsWithContext = ImmutableList.builder();
+        eventsWithContext.add(EventWithContext.create(event));
+        eventsConsumer.accept(eventsWithContext.build());
```
mpfz0r, line 71:
```suggestion
        eventsConsumer.accept(List.of(EventWithContext.create(event)))
```

graylog2-web-interface/src/logic/notifications/NotificationsFactory.jsx
```diff
@@ -14,406 +14,32 @@
...
-            </span>
-          ),
+          values: {
+            AUTHENTICATION_BACKEND: Routes.SYSTEM.AUTHENTICATION.BACKENDS.show(authServiceId),
+          },
         };
       }
 
-      case 'archiving_summary':
+      case 'no_input_running': {
```
mpfz0r, line 33:
```suggestion
      case 'no_input_running':
      case 'input_failure_shutdown':
      case 'input_failed_to_start': {
```

@ousmaneo there are a few missing here

graylog2-server/src/main/java/org/graylog2/migrations/V20190705071400_AddEventIndexSetsMigration.java
```diff
@@ -84,41 +105,68 @@ public void upgrade() {
...
+        Optional<IndexSetConfig> optSystemEventConfig =
+                getEventsIndexSetConfig(elasticsearchConfiguration.getDefaultSystemEventsIndexPrefix(), EVENT_TEMPLATE_TYPE);
+        if (optSystemEventConfig.isPresent()) {
+            String id = optSystemEventConfig.get().id();
+            try {
+                for (Stream stream: streamService.loadAllWithIndexSet(id)) {
+                    streamService.destroy(stream);
+                }
+                final IndexSet indexSet = indexSetRegistry.get(id)
+                        .orElseThrow(() -> new javax.ws.rs.NotFoundException("Index set <" + id + "> not found."));
```
mpfz0r, line 139:
```suggestion
                        .orElseThrow(() -> new IllegalStateException("Index set <" + id + "> not found."));
```
I wouldn't use that exception here. It's typically only used in Resources.

graylog2-server/src/main/java/org/graylog2/notifications/NotificationServiceImpl.java
```diff
@@ -126,11 +153,29 @@ public boolean publishIfFirst(Notification notification) {
         try {
             save(notification);
             auditEventSender.success(AuditActor.system(nodeId), SYSTEM_NOTIFICATION_CREATE, notification.asMap());
-        } catch(ValidationException e) {
+
+            final EventDefinitionDto systemEventDefinition =
```
mpfz0r, line 157: this would be nicer if it's extracted into a separate method. e.g. `createSystemEvent(notifcation)`

mpfz0r (changes requested):
a few more comments. but we are almost there :sweat_smile:

mpfz0r (approved):
LGTM :+1:

graylog2-server/src/main/java/org/graylog2/database/entities/ScopedDbService.java
mpfz0r: I think we should rename `deleteImmutable()` to `forceDelete()` and fix the comment

mpfz0r (changes requested):
Sorry, forgot one thing

mpfz0r (approved):

graylog2-web-interface/src/components/event-definitions/event-definitions/EventDefinitionEntry.tsx
```diff
@@ -82,6 +75,26 @@ const EventDefinitionEntry = ({
...
+
+    const plugin = getConditionPlugin(eventDefinition.config.type);
+
+    let suffix = <span>{plugin?.displayName ?? eventDefinition.config.type}</span>;
+
+    if (!isScheduled) {
+      suffix = (<span>{suffix} <Label bsStyle="warning">disabled</Label></span>);
+    }
+
+    return suffix;
```
grotlue, line 95: I think it's good to avoid lets if possible. As these two versions of `suffix` don't share anything except for the span tag, we could rewrite it as:

```suggestion
    if(isScheduled) {
      return (<span>{plugin?.displayName ?? eventDefinition.config.type}</span>);
    }
    
    return(<span>{suffix} <Label bsStyle="warning">disabled</Label></span>);
```

graylog2-web-interface/src/components/notifications/Notification.tsx
```diff
@@ -0,0 +1,116 @@
...
+const NotificationTimestamp = styled.span(({ theme }) => css`
+  margin-left: 3px;
+  font-size: ${theme.fonts.size.small};
+`);
+
+const _sanitizeDescription = (description) => {
+  return DOMPurify.sanitize(description);
+};
+
+/* eslint-disable react/no-danger */
```
grotlue, line 63: Can we only disable it on one specific line, or at least place this above the return?

graylog2-web-interface/src/components/notifications/Notification.tsx
```diff
@@ -0,0 +1,116 @@
...
+      <NotificationHead>
+        <Icon name="bolt" />{' '}
+        <span />
+        {message.title}{' '}
+
+        <NotificationTimestamp>
+          (triggered <RelativeTime dateTime={notification.timestamp} />)
+        </NotificationTimestamp>
+      </NotificationHead>
+      <div dangerouslySetInnerHTML={{ __html: _sanitizeDescription(message?.description) }}
```
grotlue, line 101: Is it possible to infer from the Notification if it's a System Event? I think it would be nice to only use the htmlMessage for System Notifications, as they are generated by Graylog.

grotlue (changes requested):
I think the general approach looks good, just a few suggestions and questions.
However when testing I found an error in the console. For me it happened when opening the System Events Stream and stopping all inputs. When the log came in, the console errored the following:
Screenshot 2022-12-19 at 17 04 42

Besides that I tested in Chrome by triggering two System Events and everything seems to work fine:
Screenshot 2022-12-19 at 16 33 46
Screenshot 2022-12-19 at 16 33 22
Screenshot 2022-12-19 at 16 49 39
Screenshot 2022-12-19 at 16 49 54

graylog2-web-interface/src/components/event-definitions/event-definition-form/EventConditionForm.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+import { defaultCompare as naturalSort } from 'logic/DefaultCompare';
+import { Select } from 'components/common';
+import { Clearfix, Col, ControlLabel, FormGroup, HelpBlock, Row } from 'components/bootstrap';
+import { HelpPanel } from 'components/event-definitions/common/HelpPanel';
+import type User from 'logic/users/User';
+
+import styles from './EventConditionForm.css';
+
+import type { EventDefinition } from '../event-definitions-types';
```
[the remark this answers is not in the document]
ousmaneo: Hi, That is because there is already a `type.d` file that defines the type for the Graylog plugin system

graylog2-web-interface/src/components/event-definitions/event-definition-form/FieldsForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+  eventDefinition: EventDefinition,
+  validation: {
+    errors: {
+      title: string,
+    }
+  },
+  onChange: (name: string, value: any) => void,
+}
+
+const FieldsForm = ({ currentUser, eventDefinition, validation, onChange }: Props) => {
```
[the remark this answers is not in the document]
ousmaneo: Good catch! Thank you !

graylog2-web-interface/src/components/notifications/Notification.tsx
ousmaneo: Notifications displayed by this component and the ones in system Events are unrelated. Notifications data were hard coded in HTML before now we are getting them from the backend.

graylog2-web-interface/src/components/event-definitions/event-definitions/EventDefinitionEntry.tsx
ousmaneo: > I think the general approach looks good, just a few suggestions and questions. However when testing I found an error in the console. For me it happened when opening the System Events Stream and stopping all inputs. When the log came in, the console errored the following: Screenshot 2022-12-19 at 17 04 42

I checked this issue but I think it's not directly related to this PR.

graylog2-web-interface/src/components/event-definitions/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const SYSTEM_EVENT_DEFINITION_TYPE = 'system-notifications-v1';
```
grotlue, line 1: I think this is missing the license header to make the build green.

grotlue (approved):
Thanks for the changes! The build is red, but this is just due to a missing license header.
LGTM 👍

ousmaneo (approved):

PR #13844, opened 2022-11-01
Failing the container startup early in IT tests if an error message occurs

janheise:
## Description

When a container fails during startup, the existing logic would wait for the success message until it times out, effectively wasting time. This PR is now adding logic which also checks for error indicators in the container's log and fails immediately when one of them is seen.

/nocl

## Types of changes

## Checklist:


graylog2-server/src/test/java/org/graylog/testing/graylognode/WaitForSuccessOrFailureStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+                for(final String failure: failures) {
+                    if (outputFrame.getUtf8String().matches("(?s)" + failure)) {
+                        throw new ContainerLaunchException("Container startup failed. Was looking for: '" + failure + "'");
+                    }
+                }
+                return outputFrame.getUtf8String().matches("(?s)" + success);
+            };
+            try {
+                waitingConsumer.waitUntil(waitPredicate, startupTimeout.getSeconds(), TimeUnit.SECONDS, 1);
+            } catch (TimeoutException e) {
```
dennisoelkers, line 65: We can get rid of this nested `try-catch`-block if we just pull up the catch-clause by one level.

dennisoelkers (changes requested):
Looks good, just one inline annotation.

dennisoelkers (approved):

PR #13854, opened 2022-11-01
Task/input list component

zeeklop:
Adds a new abstraction to use `Creatable` component from `react-select` as an input that returns a list of values.

/nocl

## Description

Adds a new component `InputList` that is an abstraction of `Creatable` component from `react-select`. This component allows us to use it as a regular input that returns a list of values instead of a single value.

## Motivation and Context

When creating a list on a form field, the most common approach is to make the user follow a format, to then parse the string and create the list. This component makes use of the `Creatable` provided by `react-select` to create a better UX for the users when they need to enter a list of values. Furthermore, this component profusely returns an event as if it were a native HTML Element, to avoid having to handle the `onChange` event in a custom method and instead handle it with the rest of the inputs in a form.

## How Has This Been Tested?

Manually and Unit tests

## Types of changes

## Checklist:


mikedklein (approved):
Tested functionality of new component, works as expected.

PR #13855, opened 2022-11-01
Configuring pivot limits globally instead of per-pivot in frontend.

### LINKED ISSUE #8111: Allow defining total amount of column pivots for an aggregation
The configuration option "Number of values" of a column pivot currently relates to each row pivot. Imagine the following scenario:

* The field timestamp is configured as a row
* The field source is configured as a column with the option "Number of values" with an amount of 2
* The two sources with the highest amount of messages for timestamp X are A and B
* The two sources with the highest amount of messages for timestamp Y are B and C

In this case e.g. a line chart will display three categories (A, B and C). It would be helpful to be able to configure to total amount of categories / column pivots, which is based on the amount of all results for a category / column pivot.

* Graylog Version: 3.2

dennisoelkers:
## Motivation and Context

This PR is providing the frontend changes to enable the user to configure a _global_ pivot limit instead of a per-pivot one. It requires the migration from #13806 to have run before.

Pivot limits are now configured a) for rows and b) for columns and will be applied to consecutive values pivots (not date fields, which are based on calendaric units/auto intervals).

Fixes #8111

## Types of changes

## Checklist:


janheise (approved):
LGTM!

chalfling:
This is really annoying. I was able to create a Data Table with several "Group By Rows" and limiting results from each row like "limit by 1 result", but now I can't.

Example:
event(limit 10) | srcip(limit 1) | total
Logon  | 192.168.0.1 | 500
Logoff | 192.168.1.1 | 200
XML Attack | 192.168.2.1 | 100
XSS Attack | 192.168.3.1 | 50
SQL Injection | 192.168.4.1 | 10

Then a was able to see top 10 events with The top1 Source.

Now is something like:
event | srcip | total (row limit 10)
Logon  | 192.168.0.1 | 500
Logon  | 192.168.0.2 | 450
Logon  | 192.168.0.3 | 400
Logon  | 192.168.0.4 | 350
Logon  | 192.168.0.5 | 300
Logon  | 192.168.0.6 | 250
Logon  | 192.168.0.7 | 200
Logon  | 192.168.0.8 | 150
Logon  | 192.168.0.9 | 100
Logon  | 192.168.0.10 | 50

For me this is a Breaking Change =(

dennisoelkers:
Hey @chalfling,

thanks for reporting this. We do consider this a breaking change too. It has been requested by many users, because for most use cases not applying a global limit to multi-field pivots returns unexpected results (because bucketing is applied on multiple, consecutive levels). For your use case this is not ideal, but this is also the first one we are seeing in practice that makes use of the previous behavior.

We will take a look into this and improve it. Please take note that most developers are on vacation until the new year, but we will come up with something to make this work again for you.

chalfling:
> Hey @chalfling,
> 
> thanks for reporting this. We do consider this a breaking change too. It has been requested by many users, because for most use cases not applying a global limit to multi-field pivots returns unexpected results (because bucketing is applied on multiple, consecutive levels). For your use case this is not ideal, but this is also the first one we are seeing in practice that makes use of the previous behavior.
> 
> We will take a look into this and improve it. Please take note that most developers are on vacation until the new year, but we will come up with something to make this work again for you.

Thank you for your fast reply @dennisoelkers !

On Kibana I'm able to limit by global AND by pivots. I thinking this is the best approach.

Happy Christmas and Happy New Year!

dennisoelkers:
Hey @chalfling,

we are in the middle of working on https://github.com/Graylog2/graylog2-server/pull/14446 addressing the issue you raised. The idea is to roll back some of the changes we did (changing the strategy from nesting aggregations to generating tuples back to nesting aggregations again), so your use case should work again as expected. In addition, we are now also adding the capability to define multiple fields per group by, so we are still extending the product's capabilities to allow generating proper top-n/bottom-n results with limits being applied over multiple fields.

Once this is tested and merged, we plan to include it in one of the next 5.0.x releases too, effectively rolling back the breaking change and doing a less intrusive change instead.

Does this make sense to you?

chalfling:
> Hey @chalfling,
> 
> we are in the middle of working on #14446 addressing the issue you raised. The idea is to roll back some of the changes we did (changing the strategy from nesting aggregations to generating tuples back to nesting aggregations again), so your use case should work again as expected. In addition, we are now also adding the capability to define multiple fields per group by (generating tuples), so we are still extending the product's capabilities to allow generating proper top-n/bottom-n results with limits being applied over multiple fields, therefore hopefully supporting all use cases.
> 
> Once this is tested and merged, we plan to include it in one of the next 5.0.x releases too, effectively rolling back the breaking change and doing a less intrusive change instead.
> 
> Does this make sense to you?

Hey @dennisoelkers ! Thank you for your reply.

I think this is the best solution. Thank you very much =)

dennisoelkers:
Hey @chalfling,

we have merged the PR for 5.0 (https://github.com/Graylog2/graylog2-server/pull/14485), so it will be part of the `5.0.3` release happening next wednesday (February 1st). Thanks for your feedback and patience!

chalfling:
> Hey @chalfling,
> 
> we have merged the PR for 5.0 (#14485), so it will be part of the `5.0.3` release happening next wednesday (February 1st). Thanks for your feedback and patience!

Hello @dennisoelkers thank you very much! =)

PR #13513, opened 2022-09-21
Unify placement and styling of form buttons.

linuspahl:
## Motivation and Context

With this PR we are starting to unify the styling and placement of our form buttons. Currently their styling, placement and naming varies.

With these changes we are creating a `FormSubmit` and a `ModalSubmit` component.
The `FormSubmit` component should be used for forms on pages, while the `ModalSubmit` should be used for modals which contain a submit button.

By using these components in every suitable form and modal, we are able to change the form and modal submit and cancel button in one place.

With this PR we are also:
- defining more meaningful names for the submit buttons. Instead of `Save` or `Submit` we are using for example `Create stream`
- unifying the naming of the cancel button
- unifying the placement of submit buttons in forms on pages:

image

Next steps:
- Implementing the `FormSubmit` and `ModalSubmit` components for more forms and modals.
- Describing the best practices in our UI documentation

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/4089

PR #13871, opened 2022-11-03
Extract updated value from target before it is synthesized.

### LINKED ISSUE #13867: Checkboxes in sidecar system configuration are broken
## Expected Behavior

Sidecar configuration on `System/Configurations` page should be editable.

## Current Behavior

Clicking any of the checkboxes results in

```
Something went wrong:

Cannot read properties of null (reading 'type')
````

https://user-images.githubusercontent.com/886541/199548410-d3ac6595-9244-4bd2-bccb-3b7a6dd150cc.mp4

## Your Environment

* Graylog Version: `master`

dennisoelkers:
## Motivation and Context

In #13513, the `setState`-call of the `_onUpdate` function was extended to use a callback instead of a value. Extracting the new value from the event was also moved into the `setState`-callback. Unfortunately, at this point the event is already synthesized for performance reasons and target value extraction fails.

This PR is now moving target value extraction out of the scope of the `setState`-call to fix this.

Fixes #13867.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/configurations/SidecarConfig.jsx
+++ b/graylog2-web-interface/src/components/configurations/SidecarConfig.jsx
@@ -87,14 +87,12 @@ const SidecarConfig = createReactClass({
 
   _onUpdate(field) {
     return (value) => {
+      const newValue = typeof value === 'object' ? getValueFromInput(value.target) : value;
+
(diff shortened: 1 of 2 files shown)


linuspahl (approved):
Fixes the described problem!

PR #13878, opened 2022-11-03
Disallow exporting non-exportable Event Definitions in Content Packs

ryan-carroll-graylog:
Enables controlling what Event Definition config types are able to be exported to a Content Pack.
Sigma Event Definition config specific changes are in: https://github.com/Graylog2/graylog-plugin-enterprise/pull/4304

## Description

With the introduction of Sigma Rule Event Definitions, there are some there are some types of Event Definitions that should not be able to be exported to Content Packs. This change creates an `isContentPackExportable()` function in the  `EventProcessorConfig` interface, which defaults to true but allows sub-classes to override to false. This is then checked in the `EventDefinitionFacade` when providing the list of exportable Event Definitions to the User.

## Motivation and Context

Sigma Rule Event Definitions are not meant to be exportable to Content Packs as exporting Rules will be done using the Sigma Rule Entities themselves, and exporting a Content Pack with Sigma Rule Event Definitions causes the export to fail.

Addresses: https://github.com/Graylog2/graylog-plugin-enterprise/issues/4292

## How Has This Been Tested?

Locally with development environment and unit testing.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/contentpack/facade/EventDefinitionFacade.java
+++ b/graylog2-server/src/main/java/org/graylog/events/contentpack/facade/EventDefinitionFacade.java
@@ -172,6 +172,7 @@ public class EventDefinitionFacade implements EntityFacade<EventDefinitionDto> {
     @Override
     public Set<EntityExcerpt> listEntityExcerpts() {
         return eventDefinitionService.streamAll()
+                .filter(ed -> ed.config().isContentPackExportable())
                 .map(this::createExcerpt)
                 .collect(Collectors.toSet());
     }
--- a/graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
@@ -73,6 +73,16 @@ public interface EventProcessorConfig extends ContentPackable<EventProcessorConf
         return Collections.emptySet();
     }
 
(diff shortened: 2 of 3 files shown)


ryan-carroll-graylog:
@danotorrey @kingzacko1 let me know what you think of this solution to the issue.

I also considered simply checking for the `sigma-v1` config type in `EventDefinitionFacade::listEntityExcerpts` e.g.:
```
    public Set<EntityExcerpt> listEntityExcerpts() {
        return eventDefinitionService.streamAll()
                .filter(ed -> ed.config().type().equals("sigma-v1")
                .map(this::createExcerpt)
                .collect(Collectors.toSet());
    }
```
But it would require checking for a hard coded String since the `SigmaEventProcessorConfig` class can't be referenced directly from `graylog2-server`, which seemed fragile and would introduce Sigma specific code to the repo.

Let me know if you think the current approach is too heavy handed or you have ideas for a better way to filter out the Sigma Event Definitions.

kingzacko1 (approved):
I agree we'd rather not include the actual event type name in the filtering here. I think this approach makes sense for what we want to accomplish, seems minimally intrusive, looks good, and tests successfully.

danotorrey:
Thanks for this @ryan-carroll-graylog! I'll review after lunch.

danotorrey (approved):
Nice solution! Thanks @ryan-carroll-graylog!

PR #13926, opened 2022-11-09
Prefer  endpoint when */* accept header is used

### LINKED ISSUE #13755: Inconsistent results from the Hello World API endpoint
See customer issue: https://github.com/Graylog2/graylog-cloud/issues/3267
## Description
When requesting the `graylog-server/api` endpoint with cURL (and a default Accept header of `*/*`), the Graylog server will sometimes respond with the appropriate hello world response:

```
{
  "cluster_id": "cluster-id",
  "node_id": "node-id",
  "version": "4.3.7+154",
  "tagline": "Manage your logs in the dark and have lasers going and make it look like you're from space!"
}
```

And sometimes, it will issue a temporary redirect (407) back to the root path `graylog-server/`.

Each time the Graylog server is restarted, it seems to pick one or the other, and responds accordingly until the Graylog server is rebooted. 

The problem is that, it seems users expect the API endpoint to be consistent in it's response.

**Workaround:**
When explicitly providing the `Accept: application/json` header, the endpoint always returns JSON. This is a reliable solution, but I think the inconsistency issue without the header is confusing for users, and we had once instance of this with a paying customer already.
 
`curl localhost:9000/api -H "Accept: application/json"`

## Possible Solution

See the mapping for the request here:
https://github.com/Graylog2/graylog2-server/blob/f35df42e165ac570b8b27de3f8eeac85e74ed610/graylog2-server/src/main/java/org/graylog2/rest/resources/HelloWorldResource.java

It explicitly maps to accept `application/json`, so that is why the above workaround is successful. 

It appears that when the `*/*` header is used, one of the two methods in `HelloWorldResource` is picked at random. The order is probably determined at startup. 

1.
2.
3.
4.

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

luk-kaminski:
## Description
Hopefully fixes #13755
/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/HelloWorldResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/HelloWorldResource.java
@@ -43,6 +43,8 @@ public class HelloWorldResource extends RestResource {
     private final NodeId nodeId;
     private final ClusterConfigService clusterConfigService;
 
+    private static final String LOWEST_PRIORITY = ";q=0";
+
     @Inject
     public HelloWorldResource(NodeId nodeId,
                               ClusterConfigService clusterConfigService) {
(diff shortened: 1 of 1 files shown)


danotorrey (approved):
LGTM and tests successfully! I received a consistent JSON API response each time when supplying the `*/*` or no content type. Thanks a lot for the fix. Do you think this should be merged and included in Graylog 5.0? Seems that it only provides benefits, since it resolves the random behavior, so I am in favor of merging it.

luk-kaminski:
> LGTM and tests successfully! I received a consistent JSON API response each time when supplying the `*/*` or no content type. Thanks a lot for the fix. Do you think this should be merged and included in Graylog 5.0? Seems that it only provides benefits, since it resolves the random behavior, so I am in favor of merging it.

As Bernd has said in the dev channel, we can now again merge bugfixes.
So I'll go with it and merge it. :)

PR #13885, opened 2022-11-04
Fix Pipeline rule description not saved

gally47:
## Motivation and Context

fixes #13820 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #13928, opened 2022-11-09
Test to prevent pipeline rule description bug (13885)

gally47:
In this PR we added a test to prevent the issue #13885.

## Motivation and Context

#13885

## Types of changes

## Checklist:


mnin (approved):
LGTM can we convert the files to typescript too?

gally47:
> LGTM can we convert the files to typescript too?

Okay will do that

mnin (changes requested):
LGTM

Please convert `RuleForm.jsx` also to TypeScript, thank you!

mnin (changes requested):
Something went wrong on the CI please fix this first:

https://jenkins.ci.torch.sh/blue/organizations/jenkins/Graylog-Snapshots%2Fgraylog2-server/detail/PR-13928/7/pipeline#step-133-log-743

Screenshot 2022-11-25 at 12 23 10

graylog2-web-interface/src/components/rules/RuleForm.tsx
```diff
@@ -102,6 +106,8 @@ const RuleForm = ({ create }) => {
         <PipelinesUsingRule create={create} />
 
         <Input id="rule-source-editor" label="Rule source" help="Rule source, see quick reference for more information." error={errorMessage}>
+          {/* TODO: Figure out issue with props */}
```
mnin, line 109: @gally47 any chance to fix this issue too?

gally47: No straightforward fix so far, SourceCodeEditor is not recognising the prop 'id' when used in a typescript file, when it clearly exists. I will create a separate Issue/PR for this case to investigate more, what do you think @mnin ?

mnin: I took a closer look, the same problem occurs among others here: https://github.com/Graylog2/graylog2-server/blob/master/graylog2-web-interface/src/components/sidecars/configuration-forms/ConfigurationForm.tsx#L319. This is out of scope for me.

changelog/unreleased/pr-13928.toml
```diff
@@ -0,0 +1,5 @@
+type = "fixed"
+message = "Test to prevent pipeline rule description bug (13885)"
```
grotlue, line 2: This point was made by @linuspahl in another PR and I think was a good reminder: We only need to add an entry to the changelog if the changes made in the PR affect the user - in this one it is not the case: https://graylogdocumentation.atlassian.net/wiki/spaces/GENERAL/pages/2518908931/Changelog+Entries#When-to-include-a-changelog-entry%3F

I think it would be good to remove this as I imagine someone reading the changelog would like to only see information important to them.

You could add this PR to the changelog of the original fix though if you want to track it.

graylog2-web-interface/src/components/rules/RuleForm.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import * as React from 'react';
+import { render } from 'wrappedTestingLibrary';
+
+import RuleForm from './RuleForm';
+import { PipelineRulesProvider } from './RuleContext';
+
+describe('RuleForm', () => {
+  it('should show the previous Description value in the field', () => {
```
grotlue, line 24:
```suggestion
  it('should show the previous description value in the field', () => {
```

grotlue: Maybe I am not understanding the test correctly but I think it does not test the bug fixed in the reference ticket. If I got it right the bug was that the description was not saved correctly caused by the field not updated properly.

I'd change this test:
- remove the description to be rendered by default
- use a user event in the input field to change the description
- mock `handleSavePipelineRule` (part of `PipelineRulesContext`) and check that it was called with the correct attributes

This test would be closer to the user behaviour and is thus (a bit) more independent from the implementation details.

WDYT?

grotlue (changes requested):
I like the added types 🎉 

I think we should remove the changelog or add it to the existing one and I think the test is not testing the bug. Let's chat if I didn't understand the bug correctly :)

changelog/unreleased/pr-13928.toml
gally47: Yep I will remove it.

graylog2-web-interface/src/components/rules/RuleForm.test.tsx
gally47: Basically the problem was that the field was always empty, I agree with your suggestion to make the test as closer to user behaviour as possible it is better to mock the `handleDescription` and make sure it is always called with the correct `description` value.

graylog2-web-interface/src/components/rules/RuleForm.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <RuleForm create={false} />
+      </PipelineRulesContext.Provider>,
+    );
+
+    const descriptionInput = getByLabelText('Description');
+
+    expect(descriptionInput).toHaveValue('');
+
+    userEvent.paste(descriptionInput, ruleToUpdate.description);
+    const createRuleButton = getByTitle('Update rule & close');
```
grotlue, line 62: nit: I think for buttons `getByRole` is nicer.

Does it also work with:
```suggestion
    const createRuleButton = getByRole('button', { name: 'Update rule & close' });
```
?

grotlue (approved):
LGTM thanks a lot for adding this! 👍

graylog2-web-interface/src/components/rules/RuleForm.test.tsx
gally47: Yep that should work

Issue #8592, opened 2020-07-22
Nodes with too long GC pauses with G1GC


mikkolehtisalo:
Graylog generates a lot of GC pause alerts like

> Nodes with too long GC pauses (triggered 2 days ago)
> There are Graylog nodes on which the garbage collector runs too long. Garbage collection runs should be as short as possible. Please check whether those nodes are healthy. (Node: 7ee19339-0bae-45fe-9325-56166bcf088e, GC duration: 1699 ms, GC threshold: 1000 ms)

after switching to G1 GC. I believe the alerts are in error. GC probably does take a long time with G1GC, but as the world isn't stopped it isn't an issue. This alert should probably be modernized for use with newer JVMs with newer garbage collectors.

zino7825:
I've same experienced after upgrade java from 8 to 11.

mikkolehtisalo:
For now it is possible set `gc_warning_threshold = 5s` as workaround.

boosty:
https://github.com/Graylog2/graylog2-server/pull/13929#issuecomment-1326126342

Issue #5759, opened 2019-03-11
clone_message creates endless loop


stockf:
In Graylog 3.0 using a rule that contains the clone_message() function creates an endless loop like described in #4722.
#5284 is related but I believe that this is a bug in the code not in the documentation.

## Expected Behavior

The debug output ist displayed once -> The rule is evaluated just once

## Current Behavior

The debug output gets repeated over and over until Graylog becomes completely unusable

## Possible Solution

It seems that clone_message() makes the cloned message immediately enter the same stream again and is therefore evaluated in an infinity loop.

## Steps to Reproduce (for bugs)

1. Create the following rule:
```
rule "clone"
when
    true
then
    debug("cloning message");
    clone_message();
end
```
2. Connect it to a stream over a pipeline
3. Then send a message to that stream or use the Simulator.

## Context

We are trying to forward a long message to a different stream, then shorten the message field to display in the same stream. Since we want the shortened field only in the original message and not in the one forwarded, we forward a cloned message.

## Your Environment

* Graylog Version: Graylog 3.0.0
* Elasticsearch Version: 6.5.4
* MongoDB Version: 4.0.4
* Operating System: OEL 6

kroepke:
Currently the interpreter will create an exact copy of the message, but that message will receive a new ID.
That means that from the interpreters view that message has not been processed on the streams it is routed to yet, so it will flow through the pipeline as newly seen message.
The intention for `clone_message` is that the rule then proceeds to change the clone message, leading to another path through the pipelines.

sonnylaskar:
What is the solution to this issue?
What workaround can be done to send a message from one stream to another via pipeline?

stockf:
As far as I remember, one workaround is to add a field e.g. "do_not_process" before calling `clone_message` and adding an if condition at the beginning that checks for the field. Like that (untested):

```
rule "clone only once"
when
    $message.do_not_process == false
then
    set_field("do_not_process", true);
    debug("cloning message");
    clone_message();
end
```

We endet up not using `clone_message` at all.

sonnylaskar:
That didn't work. On adding clone_message, the throughput on that rule increased exponentially leading to 100% heap size usage and Graylog crashed.

@graylog Team, this seems to be a serious bug.

stockf:
[...]

Issue #4394, opened 2017-12-04
UI pagination not working as expected with default search


shakalandy:
## Expected Behavior
- make a search
- browse to page 2 of search result
- get page 2 of initial search result

## Current Behavior
- make a search
- browse to page 2 of initial search result
- get page -10 due to the default "realtive" search of "5 minutes". Clicking on "page 2" results on getting the page 2 of a later search

## Possible Solution
In most searches you expect to get the page 2 result of your initial search.
Maybe it would be possible to change to generate the links on the different pages with "absolute" time ranges instead of relative ones?

## Steps to Reproduce (for bugs)

1. Make a "search in the last 5 minutes"
2. Receive some new messages in the meantime
3. click to page 2 of the initial search
4. get another page 2, not the one from your initial search

## Context
Our developers expect that the graylog search works as other search mechanisms out there.
This would mean, that you really get page 2 of your initial search result.
In Graylog this is only possible if you search with an "absolute" interval before.
So i am not that sure if this is really considered as a bug in your point of view.

## Your Environment

* Graylog Version: graylog-server-2.3.2-1.noarch
* Elasticsearch Version: elasticsearch-5.6.3-1.noarch
* MongoDB Version: mongodb-server-2.6.12-3.el7.x86_64
* Operating System: RHEL 7
* Browser version: Chrome 62

bernd:
@shakalandy This is not as easy to solve as it seems, unfortunately. Even when using absolute time ranges for the pagination, the data can change. For example, a log shipper was disconnected for some time and buffered the logs. If it re-connects and empties its buffer there will be new messages with an older timestamp. In that case the pagination would be wrong even with absolute time ranges.

If our backend storage would implement consistent views over search results, we could use this. But Elasticsearch cannot do this right now. (it's a pretty hard problem, especially in a distributed data store)

Another way to solve this would be to materialize every search into some storage (e.g. writing every message in the result to disk) and paginate over that. That wouldn't ensure that the user will see all data, but the view would be consistent.
This would be quite expensive in regard to resources and not that easy to implement, though.

I am afraid for now we cannot really do anything about it. Sorry.

mdollinger:
Sorry, i do not understand the consideration about some "log shipper"-corner case.

To me the case ist very clear:
- To execute a search is one thing
- To navigate in the search results is another thing
And navigating between pages must not result in another search being executed.
So the way it is implemented now seems to me like a big conceptual insufficiency which should stepwise be refactured

[...]

PR #13929, opened 2022-11-09
Enforce usage of G1GC

mpfz0r:
There are some cases where Java falls back to using the Serial GC instead of G1GC:
Less then 2GB of RAM or only one CPU.
This can be the case in some docker environments.
We always want to use G1GC, because it ensures that GC pauses are always short.

Also drop NewRatio=1 which was a tuning parameter used for the previously used GC.
The default from Java is better here.
ResizeTLAB is also on by default now, so no need to specify that option.

https://github.com/openjdk/jdk/blob/3121898c33fa3cc5a049977f8677105a84c3e50c/src/hotspot/share/runtime/os.cpp#L1673

/nocl

### DIFF
--- a/bin/graylogctl
+++ b/bin/graylogctl
@@ -51,13 +51,7 @@ GRAYLOG_CONF=${GRAYLOG_CONF:=/etc/graylog/server/server.conf}
 GRAYLOG_PID=${GRAYLOG_PID:=/tmp/graylog.pid}
 LOG_FILE=${LOG_FILE:=log/graylog-server.log}
 LOG4J=${LOG4J:=}
-DEFAULT_JAVA_OPTS="-Dlog4j2.formatMsgNoLookups=true -Djdk.tls.acknowledgeCloseNotify=true -Xms1g -Xmx1g -XX:NewRatio=1 -server -XX:+ResizeTLAB -XX:-OmitStackTraceInFastThrow"
-if $JAVA_CMD -XX:+PrintFlagsFinal 2>&1 |grep -q UseParNewGC; then
-	DEFAULT_JAVA_OPTS="${DEFAULT_JAVA_OPTS} -XX:+UseParNewGC"
-fi
-if $JAVA_CMD -XX:+PrintFlagsFinal 2>&1 |grep -q UseConcMarkSweepGC; then
-	DEFAULT_JAVA_OPTS="${DEFAULT_JAVA_OPTS} -XX:+UseConcMarkSweepGC -XX:+CMSConcurrentMTEnabled -XX:+CMSClassUnloadingEnabled"
-fi
+DEFAULT_JAVA_OPTS="-Dlog4j2.formatMsgNoLookups=true -Djdk.tls.acknowledgeCloseNotify=true -Xms1g -Xmx1g -XX:+UseG1GC -server -XX:-OmitStackTraceInFastThrow"
 
 JAVA_OPTS="${JAVA_OPTS:="$DEFAULT_JAVA_OPTS"}"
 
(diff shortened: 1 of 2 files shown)


boosty:
@mpfz0r @bernd @kroepke Should we relax (or remove) our GC alert accordingly (current default for `gc_warning_threshold` is `1s`? 

https://github.com/Graylog2/graylog2-server/issues/8592

As the user reported, with G1GC there might be longer GC runs, but if it does not stop the world, it is not necessarily a problem.

bernd:
> @mpfz0r @bernd @kroepke Should we relax (or remove) our GC alert accordingly (current default for `gc_warning_threshold` is `1s`)?
> 
> #8592
> 
> As the user reported, with G1GC there might be longer GC runs, but if it does not stop the world, it is not necessarily a problem.

@boosty 
I have never seen the alert on setups with Java 17. But I also don't have very busy setups. :smile: I also don't remember an issue that mentioned the GC check notification.

@etgraylog Did you see the error mentioned in #8592 during your benchmarks?

So I am okay with removing the GC check in an upcoming release. If we see people having issues with it when 5.0 is out, we can adjust the default value of the `gc_warning_threshold` for the stable release.

mpfz0r:
@bernd I've seen it, but only when I accidentally created an endless message loop https://github.com/Graylog2/graylog2-server/issues/5759

I agree that we should remove it.
But what would be nice, if we could export the GC metrics somehow..

etgraylog:
> @etgraylog Did you see the error mentioned in https://github.com/Graylog2/graylog2-server/issues/8592 during your benchmarks?

Yes I did see it before @bernd, I thought it was just due to JVM heap min/max needing adjusted. Recent [benchmarks](https://graylogdocumentation.atlassian.net/wiki/spaces/~879882445/pages/2661548033/Graylog+4.3.9+vs+5.0.0-13.rc.2) with RC2 did not generate a single instance of it though. And in these benchmarks, I used a custom `prometheus-exporter-mapping-custom.yml` to export more JVM metrics for reference (will use it in future benchmarks too).
The G1 GC in OpenJDK 17 seems comparable to CMS GC in Java v8, no big increase in count or elapsed-time. Not sure if we should decrease computational-resource utilization by Input, increase Graylog API usage (to use more computational-resources, maybe open more web-browser tabs to Graylog pages), and observe behavior to further compare CMS & G1 GC behavior?

mpfz0r:
> And in these benchmarks, I used a custom prometheus-exporter-mapping-custom.yml to export more JVM metrics for reference

cool. would you mind sharing these. I think we could add them to our default prometheus mapping.

bernd:
> And then there are a few problems specifically with OpenJDK v17. In Graylog 5.x, there are 15 `jvm.memory.pools.CodeHeap` metrics that have single-quotes which I have not been able to define a `match_pattern` for.

Oof, I think we should update our Prometheus exporter to strip these quotes. :smile:

etgraylog:
> @etgraylog cool! would you mind sharing these. I think we could add them to our default prometheus mapping.

I created issue [#4393](https://github.com/Graylog2/graylog-plugin-enterprise/issues/4393) to consider adding them to our Prometheus Exporter Core Mappings 👍 

> > And then there are a few problems specifically with OpenJDK v17. In Graylog 5.x, there are 15 `jvm.memory.pools.CodeHeap` metrics that have single-quotes which I have not been able to define a `match_pattern` for.
> 
> Oof, I think we should update our Prometheus exporter to strip these quotes. 😄

I created issue [#4394](https://github.com/Graylog2/graylog-plugin-enterprise/issues/4394) to make note of the quotes 👍

PR #13931, opened 2022-11-09
Add Replay Search link to expanded Event/Alert description

### LINKED ISSUE #9438: Play-Button for Events
## What?
I'd like to have a Play-Button for Events and Alerts as it exists for the dashboards.

## Why?
It would speed up my investigations a lot. At the moment I'm copy&paste everything together from start or use one predefined view.

## Details
Every Event has a few values in its definition:
* Stream(s)
* Search query
* time (time to rescan + time to consider) 
* optional: group by some field name

If I hit the play-button next to the event I'd like to have a stream-view with those properties already in place. If there is a value in "group by", I'd like to have the box from "Show top values" for that field as well.

## Your Environment

* Graylog Version: 3.3.8

kingzacko1:
## Description

Users can quickly navigate to a search page populated with the criteria that triggered a specific event from the Events/Alerts page.
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4321

## Motivation and Context

Currently, in order to start an investigation into an Event/Alert, users must recreate the search from the event definition and event details. These changes allow users to more quickly start an investigation into an alert by handling that search recreation programmatically. 

Closes #9438
## How Has This Been Tested?

Tested locally in dev environment with search-based event types.
## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -213,6 +213,8 @@ private void filterSearch(EventFactory eventFactory, AggregationEventProcessorPa
                 final Message msg = resultMessage.getMessage();
                 final Event event = eventFactory.createEvent(eventDefinition, msg.getTimestamp(), eventDefinition.title());
                 event.setOriginContext(EventOriginContext.elasticsearchMessage(resultMessage.getIndex(), msg.getId()));
+                event.setTimerangeStart(parameters.timerange().getFrom());
+                event.setTimerangeEnd(parameters.timerange().getTo());
```
kingzacko1, line 217: Is there any reason we wouldn't want to have the timerange of the EventProcessorParameters added to the event here? It makes life much easier when trying to generate a search query than trying to create some arbitrary range around the `timestamp` value of the event message. Open to feedback.

graylog2-web-interface/src/routing/Routes.ts
```diff
@@ -185,6 +186,8 @@ const Routes = {
     };
 
     if (rangeType && timeRange) {
+      queryParams.rangetype = rangeType;
```
dennisoelkers, line 189: I think this is not necessary. We do deduce the type of the time range passed through query parameters by the presence of the corresponding fields (`relative` vs. `from`/`to` vs. `keyword`).

graylog2-web-interface/src/components/events/events/EventDetails.jsx
```diff
@@ -94,13 +113,16 @@ class EventDetails extends React.Component {
               &emsp;
               ({plugin.displayName || event.event_definition_type})
             </dd>
+            <dd>
+              {this.renderReplaySearchLink(event)}
```
dennisoelkers, line 117: I think we could wrap this in a container with an `Actions` title. Currently, the `Replay search` link is right below the link to the event definition, which makes it a bit hard to distinguish.

dennisoelkers (changes requested):
Looks great, works fine! Thank you very much! Just added two suggestions/questions inline.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
bernd: @kingzacko1 We use the timerange start and end values for events we create via aggregations because a single event represents aggregated values for a timerange. For filtered events, there is no timerange because the event describes a single message.

If we add the parameters timerange here, we change the semantics of the fields. For aggregation events, the timerange is also not always the timerange from the parameters. (see the comment [here](https://github.com/Graylog2/graylog2-server/blob/48b9e8bf5b2ec121bf022a36e3af4586ea516fd8/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L283)) /cc @mpfz0r 

This PR couples the events closer to OpenSearch. (e.g., by adding the query to the event DTOs) How would we handle events that are not created via OpenSearch queries? For example, we are thinking about systems events right now that are made from events in the system and not from messages in OpenSearch.
What do we put into the `query` for Sigma rules queries? Are these always a single query string?

Is there a design document or pitch for this feature? The issue is relatively sparse in detail. If we didn't do that, I think we should take a step back and talk with product management about how the feature should work in general and for events that are not created by OpenSearch searches and what

bernd (changes requested):

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
kingzacko1: > We use the timerange start and end values for events we create via aggregations because a single event represents aggregated values for a timerange. For filtered events, there is no timerange because the event describes a single message.
>If we add the parameters timerange here, we change the semantics of the fields. For aggregation events, the timerange is also not always the timerange from the parameters. (see the comment [here](https://github.com/Graylog2/graylog2-server/blob/48b9e8bf5b2ec121bf022a36e3af4586ea516fd8/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L283)) /cc @mpfz0r

That makes sense. Prior to the last commit I was actually just searching for the single message in the `origin_context` field, but without time range information the search page would default to 5 minutes and if the message was older than that you wouldn't be able to find it. I was going to fashion a time range around the `timestamp` field, but having a time range that the event processor ran resolved this issue in a way that seemed to make more sense. It also didn't give the users any insight to the actual query that was used to find the message, just the message itself. But for filter events I think that would suffice. Not sure if it is easy to see changes for all commits except the most recent, but if you look at that it doesn't have a query on the EventDTO and relies on either the existing `origin_context` or aggregation `timerange_start/end` fields.

> This PR couples the events closer to OpenSearch. (e.g., by adding the query to the event DTOs) How would we handle events that are not created via OpenSearch queries? For example, we are thinking about systems events right now that are made from events in the system and not from messages in OpenSearch.
What do we put into the query for Sigma rules queries? Are these always a single query string?

I was not considering system events that do not involve OpenSearch queries. Actually didn't even know those were a thing. Would it be possible to create a level of abstraction for events that come from OpenSearch queries and ones that do not? Then one could have the `query` field and the other would not. As Sigma rule queries work now, we generate a single string that is queried. See the dependent PR in the enterprise repo for how that gets set.

>Is there a design document or pitch for this feature? The issue is relatively sparse in detail. If we didn't do that, I think we should take a step back and talk with product management about how the feature should work in general and for events that are not created by OpenSearch searches and what

[Here is the pitch.](https://graylogdocumentation.atlassian.net/wiki/spaces/~6316c47e9794410874c6fb26/pages/2610724902/Sigma+Events+-+Log+search+replay) It was supposed to only be for Sigma Events, but since they use the existing Events framework it seemed to make sense to enable a replay for all events that were generated from an OpenSearch query.
@bernd

graylog2-web-interface/src/routing/Routes.ts
kingzacko1: That logic may need some work then, because I tried leaving it out first and I couldn't get `absolute` time ranges to work without the inclusion of `rangetype`. `relative` works fine (I used it in a similar feature for Sigma rules) but `absolute` defaults to the relative 5 minutes ago search if `&rangetype=absolute` is not provided as a parameter.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
bernd: I think we don't need to add the  `message_replay_info` field to the event index mapping. Or do you think we need to search for any information in the object? It will be available the message JSON we get from OpenSearch.

kingzacko1: That makes sense to me to avoid changing what the existing timerange fields have been being used for.

kingzacko1: @bernd I don't think we'll need to search the new `message_replay_info` object for anything. Maybe there is some use case for it, but as far as this feature is concerned we just need it to be returned with the message when searching for the events.

mpfz0r: Adding a `message_replay_info` sounds like a good idea.

Not sure if you already looked at the existing code, that builds the backlog of messages for notifications.
It kind of is a replay of the search already.
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L140

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java#L175-L176

But the search relies on the values from the event definition, which can be changed, so having and using the historic info
from the time the event was created, is better. 
We also need to store the configured streams within the `message_replay_info` then.

kingzacko1: > We also need to store the configured streams within the `message_replay_info` then.

Good idea. I will add that as well!

danotorrey:
@kingzacko1 I am currently hitting this error when trying to run this: 
```
SigmaEventProcessor.java:142:45
java: cannot find symbol
  symbol:   variable SigmaQueryBuilder
  location: class org.graylog.plugins.securityapp.sigma.processor.SigmaEventProcessor
```

kingzacko1:
> @kingzacko1 I am currently hitting this error when trying to run this:
> 
> ```
> SigmaEventProcessor.java:142:45
> java: cannot find symbol
>   symbol:   variable SigmaQueryBuilder
>   location: class org.graylog.plugins.securityapp.sigma.processor.SigmaEventProcessor
> ```

Looks like I somehow missed an import of the SigmaQueryBuilder class. Must've accidentally deleted it right before committing the latest changes. Fixing and testing now, will have fix up shortly @danotorrey

graylog2-web-interface/src/components/events/events/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  'key': string,
+  'fields': Object[],
+  'group_by_fields': Object[],
+  'source_streams': string[],
+  'replay_info': EventReplayInfo,
+};
+
+export type EventDefinitionContext = {
+  'id': string,
+  'title': string,
```
linuspahl, line 41: The attribute types can also be defined without `''`

graylog2-web-interface/src/components/events/events/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  'event_definition_type': string,
+  'priority': string,
+  'timestamp': string,
+  'timerange_start': string,
+  'timerange_end': string,
+  'key': string,
+  'fields': Object[],
+  'group_by_fields': Object[],
+  'source_streams': string[],
+  'replay_info': EventReplayInfo,
```
linuspahl, line 36: It looks like `replay_info` can also be `undefined`.

graylog2-web-interface/src/components/events/events/EventDetails.tsx
```diff
@@ -0,0 +1,150 @@
...
+      range = { from: `${replayInfo.timerange_start}`, to: `${replayInfo.timerange_end}` };
+    }
+
+    if (replayInfo.streams) {
+      streams = replayInfo.streams;
+    }
+
+    const replaySearchUrl = Routes.search_with_query(replayInfo.query, rangeType, range, streams);
+
+    return <Link to={replaySearchUrl}>Replay Search</Link>;
```
linuspahl, line 91: We have a "Replay Search" button for dashboards widgets. 
image

It looks like we can reuse the component or at least some logic and styling like the icon, by adjusting it slightly.

bernd (commented):
My change request was implemented, but I currently don't have time for another review. Thanks!

ryan-carroll-graylog (approved):
This looks awesome! Will be a really useful feature. Everything looks solid to me.

Testing this all works as expected with various Sigma and Filter and Aggregation events, and correlation events remain unchanged. The replay button on search widgets also appears to be functioning correctly.

linuspahl (approved):

graylog2-web-interface/src/routing/Routes.ts
kingzacko1: @dennisoelkers would you mind checking this out again and seeing if this is working as you thought?

graylog2-web-interface/src/components/events/events/EventDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const getConditionPlugin = (type: string): EventDefinitionType => {
+    if (type === undefined) {
+      return null;
+    }
+
+    return eventDefinitionTypes.find((edt) => edt.type === type);
+  };
+
+  const renderEventFields = (eventFields: Object[]) => {
```
dennisoelkers, line 47: Can we declare this outside of the component? This way it gets redefined on every render, effectively thrashing React's rerendering optimizations. Also, we can probably treat it as a proper React component (like `EventFields` or sth).

graylog2-web-interface/src/components/events/events/EventDetails.tsx
```diff
@@ -0,0 +1,153 @@
...
+          );
+        })}
+      </ul>
+    );
+  };
+
+  const currentUser = useCurrentUser();
+  const plugin = getConditionPlugin(event.event_definition_type);
+
+  const renderLinkToEventDefinition = () => {
```
dennisoelkers, line 66: This should also be a React component defined outside of the scope of `EventDetails`.

dennisoelkers (changes requested):
Good work! Looks good so far, added a couple of suggestions for improvements/code cleanups inline. Thanks! :pray:

graylog2-server/src/main/java/org/graylog/events/event/Event.java
```diff
@@ -96,6 +97,10 @@ public interface Event extends Indexable {
 
     void setGroupByFields(Map<String, String> fields);
 
+    void setReplayInfo(EventReplayInfo replayInfo);
+
+    EventReplayInfo getReplayInfo();
```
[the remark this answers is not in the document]
kingzacko1: Should all of the other fields that are `Optional<>` in the DTO then be made optional here as well? I just followed the pattern of the other fields that can be null in `EventDto`: `timerangeStart`, `timerangeEnd`, and `originContext`.

kingzacko1:
@dennisoelkers My latest commit addresses all feedback with the exception of the one I left unresolved with comment. I am not so familiar with `useMemo` but took a stab at using it for the `getConditionPlugin` logic. Let me know what you think!

graylog2-server/src/main/java/org/graylog/events/event/EventReplayInfo.java
```diff
@@ -31,16 +31,13 @@
 @AutoValue
 @JsonDeserialize(builder = EventReplayInfo.Builder.class)
 public abstract class EventReplayInfo {
-    public static final String FIELD_TIMERANGE_START = "timerange_start";
-    public static final String FIELD_TIMERANGE_END = "timerange_end";
+    public static final String FIELD_TIME_RANGE = "timeRange";
```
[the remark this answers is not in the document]
kingzacko1: Whoops, fixed! Thanks for catching that!

graylog2-server/src/main/java/org/graylog/events/event/EventReplayInfo.java
```diff
@@ -0,0 +1,78 @@
...
+@AutoValue
+@JsonDeserialize(builder = EventReplayInfo.Builder.class)
+public abstract class EventReplayInfo {
+    public static final String FIELD_TIMERANGE_START = "timerange_start";
+    public static final String FIELD_TIMERANGE_END = "timerange_end";
+    public static final String FIELD_QUERY = "query";
+    public static final String FIELD_STREAMS = "streams";
+
+    @JsonProperty(FIELD_TIMERANGE_START)
+    public abstract DateTime timerangeStart();
```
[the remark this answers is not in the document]
kingzacko1: Do you think we should revert this back to use `start` and `end` timestamps instead then? @bernd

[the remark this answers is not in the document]
bernd: @kingzacko1 If there are no strong reasons to use the absolute range object, I suggest going back to `start` and `end`, yes. That also keeps consistency with the other time range in the event object.

PR #13937, opened 2022-11-10
Always serialize joda Durations as milliseconds

mpfz0r:
Add a  Joda DurationSerializer that ignores the `SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS` setting and always serializes `org.joda.time.Duration` to milliseconds. This became necessary with the Jackson 2.13 update, which forced us to disable `WRITE_DURATIONS_AS_TIMESTAMPS` to have `java.time.Duration` objects serialized as strings.

#Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/4320

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/jackson/JodaDurationCompatSerializer.java
+++ b/graylog2-server/src/main/java/org/graylog2/jackson/JodaDurationCompatSerializer.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog2.jackson;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.datatype.joda.ser.DurationSerializer;
+import org.joda.time.Duration;
+
+import java.io.IOException;
+
+/**
+ * A Joda DurationSerializer that ignores the {@link com.fasterxml.jackson.databind.SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS}
+ * setting and always serializes Durations to milliseconds.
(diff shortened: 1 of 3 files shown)


patrickmann:
Looks good and fixes the problem.
Looking around for other instances of `joda.time.Duration` that end up in the DB I found these:
- `@JsonProperty(FIELD_CREATION_DURATION)`
These are also serialized to milliseconds.

bernd:
@mpfz0r @patrickmann Thanks for the fix! Good work! :pray: 

Can you elaborate on how this worked before the Jackson update? Did we deliberately serialize joda-time and java.time durations differently in the old version, or was that by accident?

mpfz0r:
> Can you elaborate on how this worked before the Jackson update?

The jackson update changed the default for `java.time.Duration`. We had to disable `WRITE_DURATIONS_AS_TIMESTAMPS` to make it backwards compatibile. 
joda Durations always serialized into milliseconds, so disabling `WRITE_DURATIONS_AS_TIMESTAMPS` changed the behaviour for them.

We couldn't have both at the same time anymore.

We could've done it the other way around, and writing a custom serializer for java time Durations instead.
Or change all our code to move away from joda Durations, but that's more involved, and probably not a good idea in this phase of the release.

bernd:
> > Can you elaborate on how this worked before the Jackson update?
> 
> The jackson update changed the default for `java.time.Duration`. We had to disable `WRITE_DURATIONS_AS_TIMESTAMPS` to make it backwards compatibile. joda Durations always serialized into milliseconds, so disabling `WRITE_DURATIONS_AS_TIMESTAMPS` changed the behaviour for them.
> 

Got it. Thank you! :pray:

patrickmann (approved):

bernd (approved):
LGTM. Tested in my test environment. The licenses page works and also the field refresh interval on the indices page.

patrickmann:
> Can you elaborate on how this worked before the Jackson update?

In Jackson 2.9
- the [Java Duration deserializer](https://github.com/kezhuw/jackson-modules-java8/blob/b8ccc50a245825621fd455850b2125cb9d8f4f87/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java) honors `WRITE_DATES_AS_TIMESTAMPS` and ignores `WRITE_DURATIONS_AS_TIMESTAMPS`
- Joda Duration deserializer [honors](https://github.com/FasterXML/jackson-datatype-joda/blob/31a02d03967b97f0f81ba809ca680eda2f66a42c/src/main/java/com/fasterxml/jackson/datatype/joda/ser/DurationSerializer.java#L36) `WRITE_DURATIONS_AS_TIMESTAMPS`

Since the default setting for WRITE_DURATIONS_AS_TIMESTAMPS is enabled and GL 4.3 explicitly disables `WRITE_DATES_AS_TIMESTAMPS` we end up deserializing Java Durations to string and Joda Durations to milliseconds.

In Jackson 2.13 Java Duration deserializer has been corrected to use WRITE_DURATIONS_AS_TIMESTAMPS. When this serialization feature was disabled in GL 5.0 to be compatible with serialization of Java Durations in GL 4.3 we broke compatibility with previously serialized values of Joda Duration.

bernd:
@patrickmann Awesome, thanks for the detailed explanation! :pray:

PR #13939, opened 2022-11-10
Conditionally hide edit button for permission config

### LINKED ISSUE #12090: Hide configure buttons on System/Configurations for reader users
A user with the reader role can go to System/Configurations and see what is configured there. For what I could see, most buttons to edit configuration are hidden except these two:

Screenshot 2022-02-04 at 14 44 44
Screenshot 2022-02-04 at 14 44 50

This is only a presentation issue, since the back end authorizes both requests and no changes are allowed.

## Expected Behavior
Reader users should not see any buttons to configure features in System/Configurations

## Current Behavior
Two features show buttons to configure them, even for reader users.

1. Log in with a reader user
2. Go to System/Configurations
3. See the "configure" buttons

## Your Environment
* Graylog Version: Graylog-4.3.0-SNAPSHOT (da5961aa7a)

thll:
Hide the edit button for the permission config on the configuration page if the user does not have the required edit permission.

Fixes #12090

### DIFF
--- a/changelog/unreleased/issue-12090.toml
+++ b/changelog/unreleased/issue-12090.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Hide edit button for permissions config if the user doesn't have edit permissions."
+
+issues = ["12090"]
+pulls = ["13939"]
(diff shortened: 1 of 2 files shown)


ousmaneo (approved):
LGTM!

PR #13943, opened 2022-11-10
Events: change default for `search_within_ms` and `execute_every_ms` to 5 mins

thll:
Changes the default values for `search_within_ms` and `execute_every_ms` from one minute to five minutes.

The previous default of one minute leads to a rather high frequency of event generation runs. A default of five minutes seems to be a good default for most users to start with.

Fixes https://github.com/Graylog2/graylog-plugin-enterprise/issues/3740

### DIFF
--- a/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterAggregationForm.jsx
+++ b/graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterAggregationForm.jsx
@@ -34,8 +34,8 @@ const initialFilterConfig = {
   query: '',
   query_parameters: [],
   streams: [],
-  search_within_ms: 60 * 1000,
-  execute_every_ms: 60 * 1000,
+  search_within_ms: 5 * 60 * 1000,
+  execute_every_ms: 5 * 60 * 1000,
 };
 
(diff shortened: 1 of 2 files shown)


graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterAggregationForm.jsx
```diff
@@ -115,12 +115,6 @@ class FilterAggregationForm extends React.Component {
     this.setState(stateChange);
   };
 
-  handleChange = (event) => {
```
thll, line 118: @gally47: I had to change this method to make the linter check pass. Could you please double-check if the method is really not used anymore? I don't know our frontend well enough to rule out that there is any magic going on that would rely on this method being present.

mpfz0r (approved):

graylog2-web-interface/src/components/event-definitions/event-definition-types/FilterAggregationForm.jsx
thll: @mnin already checked this for me 🙏

mnin (approved):
LGTM

PR #13960, opened 2022-11-14
Change default values for `http_thread_pool_size` and `proxied_requests_thread_pool_size`

thll:
## Description

The default value for `http_thread_pool_size` is 16, which is pretty low.

16 concurrent requests to `api/cluster/metrics/multiple` on a node will exhaust resources because each request triggers (among others) a subsequent request to the node itself.

As a quick fix, this PR sets the default pool size to 64. This can only really serve as a band-aid to avoid users running into that situation easily because it moves the breaking point to ~64 concurrent requests. We are currently investigating a more sustainable solution.

We are also increasing the default value for `proxied_requests_thread_pool_size` from 32 to 64. The previous advisory of setting it to `http_thread_pool_size * average_cluster_size` has been removed from the example `graylog.conf` though, because that feels a bit excessive.

We also add previously missing (prometheus) metrics for the proxied requests thread pool:
- `gl_http_proxied_requests_executor{type="running"}`
- `gl_http_proxied_requests_executor{type="duration"}`
- `gl_http_proxied_requests_executor{type="idle"}`
- `gl_http_proxied_requests_executor_total{type="completed"}`
- `gl_http_proxied_requests_executor_total{type="submitted"}`

## Motivation and Context

Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/4325

## How Has This Been Tested?

Before the change:
```
ab -c16 -n1600 -p <(echo '{"metrics":["org.graylog2.throughput.input.1-sec-rate"]}') -T application/json -H 'X-Requested-By: ab' -A admin:admin http://127.0.0.1:9000/api/cluster/metrics/multiple
...
Concurrency Level:      16
Time taken for tests:   100.178 seconds
Complete requests:      1600
Failed requests:        123
...
```

After the change:
```
ab -c16 -n1600 -p <(echo '{"metrics":["org.graylog2.throughput.input.1-sec-rate"]}') -T application/json -H 'X-Requested-By: ab' -A admin:admin http://127.0.0.1:9000/api/cluster/metrics/multiple
...
Concurrency Level:      16
Time taken for tests:   7.233 seconds
Complete requests:      1600
Failed requests:        0
...
```

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ProxiedRequestsExecutorService.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ProxiedRequestsExecutorService.java
@@ -16,6 +16,8 @@
  */
 package org.graylog2.shared.bindings.providers;
 
+import com.codahale.metrics.InstrumentedExecutorService;
+import com.codahale.metrics.MetricRegistry;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.graylog2.plugin.Tools;
 import org.graylog2.shared.rest.resources.ProxiedResource;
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/shared/bindings/providers/ProxiedRequestsExecutorService.java
```diff
@@ -27,24 +29,33 @@
...
-                .setUncaughtExceptionHandler(new Tools.LogUncaughtExceptionHandler(LoggerFactory.getLogger(ProxiedResource.class.getName())))
-                .build()
-        );
+                .setUncaughtExceptionHandler(
+                        new Tools.LogUncaughtExceptionHandler(LoggerFactory.getLogger(ProxiedResource.class.getName())))
+                .build();
+        return new InstrumentedExecutorService(
+                Executors.newFixedThreadPool(proxiedRequestsMaxThreads, threadFactory),
+                metricRegistry,
+                name(this.getClass()));
```
bernd, line 59: Let's use a name for the metric in addition to the class name. (like we do in [JerseyService](https://github.com/Graylog2/graylog2-server/blob/dba006b663632f6265882328d3134c7cb03c3456/graylog2-server/src/main/java/org/graylog2/shared/initializers/JerseyService.java#L325)) Like `http-proxied-requests-executor`? What do you think?
```suggestion
                name(this.getClass(), "http-proxied-requests-executor"));
```
If we do that, we also need to update the Prometheus exporter mapping file.

bernd (changes requested):

PR #13969, opened 2022-11-15
Fix extraction of column metrics from other bucket.

dennisoelkers:
## Motivation and Context

This PR is fixing an issue when the following occurs:

  - an aggregation has at least one row and column pivot
  - there are documents in the result which are missing one of the pivot fields

In these cases, the `other` bucket is improperly used to extract column pivot values, resulting in a class cast exception being thrown. Instead, it should be used only for extracting metrics for the missing bucket. This is fixed in this PR.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,10 @@ jobs:
       matrix:
         java-version: [ 17 ]
     steps:
+      - name: Freeing up more disk space
+        run: |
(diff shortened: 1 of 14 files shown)


luk-kaminski:
@dennisoelkers - I have tested more, and there are still bugs when you mix rows and columns and have more than 1 metric.
I may merge this PR, as it limits the problem, but probably more work is needed here... :(

PR #13980, opened 2022-11-16
Add support for a bundled JVM in bin/graylogctl

bernd:
This change let's `graylogctl` use a bundled JVM when no `JAVA_HOME` is defined and a bundled JVM is present.


mpfz0r (approved):
LGTM

PR #13989, opened 2022-11-17
Added a function that prints a human readable byte count

janheise:
## Description

A test in the enterprise plugin failed because the used OSHI.format function is prone to locale-problems. Fixing the test would mean setting/resetting the locale which I find rather unfortunate. As we have replaced the `String.format()`because of the same problem everywhere in our code, I thought a replacement is a better idea. 

/nocl

## Types of changes

## Checklist:


thll (commented):
Looks good, but what do you think about adding a small test to perform a sanity check?

thll (approved):
Nice, thank you!

Issue #13998, opened 2022-11-18
Form validation errors lead to error notifications instead of highlights


kroepke:
## Expected Behavior
Form validation errors should be highlighted around the fields while editing/sending the form.

## Current Behavior
Validation errors lead to toastr-style notifications which flood the entire screen.
![image](https://user-images.githubusercontent.com/52014/202689503-495cc415-4b4f-495c-9e08-029b58b1f3f9.png)

## Possible Solution
n/a

## Steps to Reproduce (for bugs)
1. Create a data adapter and start typing the title.
2. Pause long enough so it fires the validation requests
3. Observe stacking of error notifications as shown above.

## Context
The form also no longer shows which field triggers the highlight, making this not only annoying but also hard to use.

## Your Environment

* Graylog Version: 5.0.0-beta.5 but also affects earlier versions

PR #14004, opened 2022-11-18
Fix toast errors on data adapter HTTP JSONPath when editing the title

mnin:
Fix mainly toast error notifications when editing the HTTP JSONPath data adapter without valid lookup url.

Fix: https://github.com/Graylog2/graylog2-server/issues/13998

Fix also one warning:

Screenshot 2022-11-18 at 16 30 09

And introduce spacing between the title and the form:

Before:

Screenshot 2022-11-18 at 17 03 53

After:

Screenshot 2022-11-18 at 17 14 49

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/HTTPJSONPathDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/HTTPJSONPathDataAdapter.java
@@ -284,7 +284,7 @@ public class HTTPJSONPathDataAdapter extends LookupDataAdapter {
         public Config defaultConfiguration() {
             return Config.builder()
                     .type(NAME)
-                    .url("")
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/adapters/HTTPJSONPathDataAdapter.java
```diff
@@ -284,7 +284,7 @@ public Descriptor() {
         public Config defaultConfiguration() {
             return Config.builder()
                     .type(NAME)
-                    .url("")
+                    .url("https://example.com/api/lookup?key=${key}")
```
dennisoelkers, line 287: I think we should not use a valid domain for the initial value, to avoid leaking something when a user accidentally uses the default.

dennisoelkers (changes requested):

PR #1146, opened 2015-05-11
Fix setting receive buffer size for UDP-based transports (#1143)

joschi:
This PR fixes the allocation of the receive buffer for UDP-based transports and increases the default receive buffer size for UDP-based transports to 64K.

While the receive buffer size for UDP-based transports could be configured for each transport, the effective size for the receive buffer was fixed to 8192 bytes by using a `FixedReceiveBufferSizePredictor` with a hard-coded value.

By changing the default receive buffer size for UDP-based transports to 65535 bytes and using the configured receive buffer size as value for the `FixedReceiveBufferSizePredictor`, the maximum UDP packet size that can be processed by UDP-based transports is now finally configurable.

As the practical maximum size for a UDP packet is 65507 bytes, the default of 65535 bytes for the receive buffer should be fine (at the cost of increased memory allocations).

Fixes #1143

PR #14005, opened 2022-11-18
Only allocate the maximum needed size for one UDP packet

### LINKED ISSUE #13306: CPU performance issues with Graylog 4.2.12 + Java 17
## Expected Behavior

The performance of Graylog 4.2.12 running on Java 17 (suggested by installation docs) should be comparable to what observed running on Java 8.

## Current Behavior

For a more extensive discussion see the community thread: [High CPU usage after upgrade from Graylog 4.1 + Java 8 to Graylog 4.2 + Java 17 ](https://community.graylog.org/t/high-cpu-usage-after-upgrade-from-graylog-4-1-java-8-to-graylog-4-2-java-17/25432)

After upgrading an installation from Graylog 4.1.14 + Java 8 to 4.2.12 to Java 17 I observed a noticeable (~4x) increase in CPU usage. Graylog and Java were the only updated parts of the installation.

Replacing Java 17 with Java 8 brought back CPU usage to the levels observed before the upgrade.

This was previously observed in a very similar way running Graylog 4.1.11 + Java 11. The performance drop vanished when switching to Java 8. The issue was curiously absent on 4.1.1 + Java 11. See community thread: [Graylog 4.1.x high CPU usage after updating for log4j](https://community.graylog.org/t/graylog-4-1-x-high-cpu-usage-after-updating-for-log4j/22294).

## Possible Solution

I suspect, though I lack the expertise to support my claim, that some change introduced between 4.1.1 and 4.1.9 influences performance when running on Java > 8, and that this change is still in effect in 4.2.12.

Apart from the obvious reproduction and fix, which would be ideal if this is confirmed to be a non-local issue, clarify the supported and suggested version of Java in the docs, and that it may have an effect on the final performance.

1. Run a load test on a Graylog 4.2.12 + Java 17 installation, observing CPU usage
2. Run the same on a Graylog 4.2.12 + Java 8

## Context

Normal operation of a Graylog cluster mainly ingesting UDP and TCP RFC5424 Syslog, running pipelines to enrich messages with fields (GROK, JSON, regular expressions) and a few alerts.

## Your Environment

* Graylog Version: 4.2.12
* Java Version: 17 and 8
* Elasticsearch Version: OSS 7.10.2
* MongoDB Version: 4.2
* Operating System: Debian 10 buster
* Browser version: Not relevant

mpfz0r:
Don't allocate the size of the entire UDP socket buffer for each packet!

This has been wrong for a long time [1], but only came up as a regression when using Java > 9.

On newer Java versions Netty does not use direct buffers via "Unsafe" to allocate memory. It falls back to ByteBuffer.allocateDirect which has inferior performance[2].

This behavior could be restored by adding these JVM options:
```
-Dio.netty.tryReflectionSetAccessible=true
--add-opens=java.base/java.nio=ALL-UNNAMED
--add-opens=java.base/sun.nio.ch=ALL-UNNAMED
```
But since fixing the receive buffer size to 64k gives us the same performance boost [3], we can avoid using these hacks, and not having to rely on Unsafe is also a good idea.

Refs
 https://github.com/netty/netty/pull/7650

 https://github.com/netty/netty/blame/e8df52e442629214e0355528c00e873e213f0139/common/src/main/java/io/netty/util/internal/PlatformDependent0.java#L979-L980

Fixes #13306

[1] https://github.com/Graylog2/graylog2-server/pull/1146
[2] In my test case I only received about `300 msg/sec`.
[3] With this change, I was able to receive ~ `35000 msg/sec` (utilizing only one netty thread)

### DIFF
--- a/changelog/unreleased/issue-13306.toml
+++ b/changelog/unreleased/issue-13306.toml
@@ -1,0 +1,11 @@
+type = "fixed"
+message = "Fix performance regression on UDP inputs with newer Java versions"
+
+issues = ["13306"]
+pulls = ["14005"]
+
+contributors = ["@giangi"]
+
+details.user = """
+UDP inputs became slow when running Graylog with a Java version > 8.
+"""
--- a/graylog2-server/src/main/java/org/graylog2/inputs/transports/UdpTransport.java
+++ b/graylog2-server/src/main/java/org/graylog2/inputs/transports/UdpTransport.java
@@ -88,7 +88,7 @@ public class UdpTransport extends NettyTransport {
         return new Bootstrap()
                 .group(eventLoopGroup)
                 .channelFactory(new DatagramChannelFactory(transportType))
-                .option(ChannelOption.RCVBUF_ALLOCATOR, new FixedRecvByteBufAllocator(getRecvBufferSize()))
+                .option(ChannelOption.RCVBUF_ALLOCATOR, new FixedRecvByteBufAllocator(65535)) // Maximum possible UDP packet size
                 .option(ChannelOption.SO_RCVBUF, getRecvBufferSize())
                 .option(UnixChannelOption.SO_REUSEPORT, true)
                 .handler(getChannelInitializer(getChannelHandlers(input)))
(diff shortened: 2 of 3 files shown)


bernd (approved):
Good work! Thank you! :pray:

PR #14017, opened 2022-11-21
Fixing sort by limiting options and reordering keys

### LINKED ISSUE #13957: Multi-sort on aggregations does not work
Multi-sort on aggregations does not work.

Screenshot 2022-11-14 at 14-41-31 Graylog - Unsaved Search

## Expected Behavior
If there are multiple sorts configured for aggregation, they are all used. Secondary sorting should be used inside single grouping only.

## Current Behavior
It seems that only the first sort is used.

The screenshot shows them.

## Your Environment

* Graylog Version: 5.0.0
* Java Version: 17

dennisoelkers:
## Motivation and Context

This PR is addressing the issues related to (multi-)sorting in aggregations as reported in #13957. Sorting options are now limited to:

  - one pivot field
  - one or multiple metrics

This is related to the way bucketing now works. We cannot produce a meaningful multi-sort with multiple fields and/or combinations of fields and metrics and do value returning better results in terms of bucketing and limiting over a more flexible multi-sort.

In addition, we are fixing issues with choosing any other field than the first pivot for sorting. In these cases, we will now reorder the pivots so the one containing the field being sorted on is first. During the extraction of results, we will shuffle values back to the original order.

Fixes #13957.

## Types of changes

## Checklist:


graylog2-server/src/test/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/ValuesBucketOrderingTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.graylog.plugins.views.search.searchtypes.pivot.PivotSort;
+import org.graylog.plugins.views.search.searchtypes.pivot.SortSpec;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class ValuesBucketOrderingTest {
```
luk-kaminski, line 28: I'd add a test to check if nothing happens when using sorts different than PivotSorts.

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/ValuesBucketOrdering.java
```diff
@@ -0,0 +1,87 @@
...
+                    }
+                    if (s2Index == -1) {
+                        return -1;
+                    }
+                    return Integer.compare(s1Index, s2Index);
+                })
+                .collect(Collectors.toList());
+    }
+
+    public static Function<List<String>, List<String>> reorderKeysFunction(List<BucketSpec> bucketSpecs, List<SortSpec> sorts) {
```
luk-kaminski, line 72: This method seems to be not unit-tested at all...

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/ValuesBucketOrdering.java
```diff
@@ -0,0 +1,87 @@
...
+                    if (s1Index == s2Index) {
+                        return 0;
+                    }
+                    if (s1Index == -1) {
+                        return 1;
+                    }
+                    if (s2Index == -1) {
+                        return -1;
+                    }
+                    return Integer.compare(s1Index, s2Index);
```
luk-kaminski, line 67: Can we consider moving this Comparator to a separate class and testing it separately?

luk-kaminski (approved):
Great job! Thanks!
I could not break anything while testing!
I've left some comments about unit-testing, it is up to you to what degree you would like to use them.

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/ValuesBucketOrdering.java
dennisoelkers: Thanks, added tests!

graylog2-server/src/main/java/org/graylog/plugins/views/search/searchtypes/pivot/buckets/ValuesBucketOrdering.java
dennisoelkers: Good idea, did that!

linuspahl (approved):
Looks good to me!

PR #14021, opened 2022-11-21
Do not implement link for `LinkContainer` children, when child is disabled.

linuspahl:
## Motivation and Context

When wrapping a `Button` or `MenuItem` with `LinkContainer`, the `LinkContainer` defines an href for the child. Previously it was still possible to click on the link, even when the child (e.g. the `Button`) is disabled. 

With this PR we no longer deine the `href` for the children, if it is disabled.

/nocl

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/router.tsx
+++ b/graylog2-web-interface/src/components/common/router.tsx
@@ -83,7 +83,7 @@ const LinkContainer = ({ children, onClick, to: toProp, relativeActive, ...rest 
     if (!disabled) {
       history.push(to);
     }
-  }, [childrenOnClick, onClick, to]);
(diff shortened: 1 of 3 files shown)


graylog2-web-interface/src/components/common/router.tsx
```diff
@@ -80,10 +80,18 @@ const LinkContainer = ({ children, onClick, to: toProp, relativeActive, ...rest
       onClick();
     }
 
-    history.push(to);
+    if (!disabled) {
+      history.push(to);
+    }
   }, [childrenOnClick, onClick, to]);
```
maxiadlovskii, line 86: Add dependency disabled

maxiadlovskii (approved):

PR #14028, opened 2022-11-22
Unify paddings of search bar, dashboard search bar and widget edit search bar.

linuspahl:
## Motivation and Context

With this PR we are unifying the search bar paddings and fine-tuning the paddings of elements like the stream select, query input and the time range.

Before:
image

After:
image


maxiadlovskii (approved):
Looks good. Just one idea. Maybe we can take this padding from a theme? In this case, we would be 100% sure that we have the same values everywhere

linuspahl:
@maxiadlovskii I like the idea, but I would not store it in the theme. I created a `Constants` file in the search bar directory for it.

Issue #13973, opened 2022-11-15
Lookup table: Button disabled when using colons for looking up keys (version 5.0.0-beta.4-1)


grotlue:
# Expected Behavior
I am not sure if colons just like that should be allowed but I would expect that I am able to test a lookup table with IP address data using an IP address as a key.

## Current Behavior
When attempting to look up an IP address the button is disabled, as the input field does not seem to be valid when it includes colons.

![lookuptables-ip-](https://user-images.githubusercontent.com/3169354/201947121-8d52e5e7-9a21-4703-80bf-33dbbbac444a.gif)

## Steps to Reproduce (for bugs)

1. Run graylog 5.0.0-beta.4-1 
2. Create a data adapter and cache.
3. Create a lookup table
4. In the lookup table or data adapter attempt to insert an IP address in the `key` field in the `Test lookup` section

## Your Environment

* Graylog Version: 5.0.0-beta.4-1
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

PR #14029, opened 2022-11-22
Allow non-word-characters in lookup tables key field

gally47:
## Description
Before we were only allowing word-characters in lookup tables key field, in this PR we fixed that.

## Motivation and Context

#13973

## Types of changes

## Checklist:


mnin (changes requested):
Please add a test to cover this bug. As there is already a test file for this component existing, it should be easy to just add one for this. To make sure the bug doesn't get reimplemented.

If this does consume too much time, let's create a follow up issue.

mnin (approved):
LGTM

PR #14054, opened 2022-11-24
Ensure Guice binder creation for the JobResourceHandler interface

bernd:
We don't bind any JobResourceHandler classes in Graylog Open, so we have to make sure we create the binding.

This PR changes the RestApiBindings to extend PluginModule instead of Graylog2Module to get access to the `jobResourceHandlerBinder()` method.

Fixes a Jersey injection error when navigating to "System/Overview" in a pure-open Graylog instance.

/nocl

*Notice:* This needs a backport to the 5.0 branch.


mpfz0r (approved):
Works with open and enterprise :+1:

PR #14060, opened 2022-11-25
Scripting api: error extraction from QueryResult

todvora:
## Description
Extracting errors from search results

## Motivation and Context
Provide better error messages to users

## How Has This Been Tested?
Added integration test

image

/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/SearchJobToTabularResponseMapper.java
```diff
@@ -0,0 +1,118 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.NotFoundException;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class SearchJobToTabularResponseMapper {
```
luk-kaminski, line 42: 1. It does not need to happen in this PR, but we may want to unit test this class.
2. I am not 100% sure if we want to to use "javax.ws.rs.*" exceptions in this mapper, or should it rather be unaware/independent of this package, throwing more general exceptions, that will be mapped to proper exceptions by the resource class - just to consider.

luk-kaminski (approved):
Great job, thanks!
I've left some comments, but nothing they may be even considered even after merging this to master, as further improvements.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/SearchJobToTabularResponseMapper.java
todvora: yeah, custom exceptions and their mappers would be the cleanest solution. Let me do that here when we are already refactoring error handling.

todvora: added custom exception that's re-thrown in the resource. Primitive but cleaner than before.

PR #14067, opened 2022-11-28
Flush remaining messages to the indexer during shutdown

thll:
## Description

Adds an extra step during the shutdown routine of a node to flush out messages to the indexer.
We try to make sure that a blocked indexer is not inhibiting shutdown. Therefore the flush is attempted from a separate thread which we interrupt after a timeout.

## Motivation and Context

When a node shuts down, it may still have messages waiting to be flushed out in the `BlockingBatchedESOutput` singleton.
The `BatchedElasticSearchOutputFlushThread` periodical will be stopped at some point, so the messages will never be written to the index and in turn not be committed to the journal. However, if other outputs are configured, the messages will have been flushed to those outputs already. When the node restarts, these unflushed messages will be reread from the journal and duplicated in outputs other than the ES output.

To reproduce this
1. Temporarily configure `output_flush_interval = 60` in `graylog.conf`
2. Assign a STDOUT output to the "All messages" stream.
3. Send in some messages and observe them being printed to STDOUT
4. Restart the node
5. Observer the messages being printed to STDOUT again
6. Search for the messages and observe them being stored only once

## How Has This Been Tested?

See the steps to reproduce above.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/outputs/BlockingBatchedESOutput.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/BlockingBatchedESOutput.java
@@ -22,6 +22,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.google.common.collect.Maps;
 import org.graylog2.indexer.IndexSet;
+import org.graylog2.indexer.cluster.Cluster;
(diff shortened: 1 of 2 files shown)


mpfz0r (approved):
LGTM :+1:

PR #14063, opened 2022-11-25
Fix indexer connectivity check by catching the correct exception type

thll:
The connectivity check for ES [Edit: *and OpenSearch*] was handling the wrong exception so that all calls to `#isConnected` would result in an exception when the ES cluster was not reachable. ~~This only affected Elasticsearch, not OpenSearch.~~

This PR fixes that by catching the correct exception type.

To see the effect, sever the connection to ES while the server is running.

Before the change, the server log will be filled with a lot messages from the `IndexFieldTypePollerPeriodical` like these: 

```
2022-11-25 17:23:26,792 ERROR: org.graylog2.indexer.fieldtypes.IndexFieldTypePollerPeriodical - Uncaught exception in Periodical
org.graylog.shaded.elasticsearch7.org.elasticsearch.ElasticsearchException: An error occurred: 
	at org.graylog.storage.elasticsearch7.ElasticsearchClient.exceptionFrom(ElasticsearchClient.java:155) ~[classes/:?]
	at org.graylog.storage.elasticsearch7.ElasticsearchClient.execute(ElasticsearchClient.java:112) ~[classes/:?]
	at org.graylog.storage.elasticsearch7.ElasticsearchClient.execute(ElasticsearchClient.java:105) ~[classes/:?]
	at org.graylog.storage.elasticsearch7.ClusterAdapterES7.isConnected(ClusterAdapterES7.java:169) ~[classes/:?]
	at org.graylog2.indexer.cluster.Cluster.isConnected(Cluster.java:115) ~[classes/:?]
	at org.graylog2.indexer.fieldtypes.IndexFieldTypePollerPeriodical.doRun(IndexFieldTypePollerPeriodical.java:106) ~[classes/:?]
	at org.graylog2.plugin.periodical.Periodical.run(Periodical.java:94) [classes/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) [?:?]
	at java.util.concurrent.FutureTask.runAndReset$$$capture(FutureTask.java:305) [?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
[...]

PR #14075, opened 2022-11-28
Backport: Fix indexer connectivity check (#14063)

thll:
Backport of #14063


janheise:
LGTM!  but I'm not approving yet because i think we should fix the unrelated test with https://github.com/Graylog2/graylog2-server/pull/14078

janheise (approved):

PR #14077, opened 2022-11-28
Fixing width of login notifications.

dennisoelkers:
## Motivation and Context

Prior to this PR, public notifications shown on login were displayed too narrow. This PR is fixing this.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/login/LoginChrome.tsx
+++ b/graylog2-web-interface/src/components/login/LoginChrome.tsx
@@ -54,6 +54,7 @@ const NotificationsContainer = styled.div`
   position: fixed;
   top: 0;
   margin-top: 5px;
+  width: 75%;
 `;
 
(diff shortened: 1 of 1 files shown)


graylog2-web-interface/src/components/login/LoginChrome.tsx
```diff
@@ -54,6 +54,7 @@ const NotificationsContainer = styled.div`
   position: fixed;
   top: 0;
   margin-top: 5px;
+  width: 75%;
```
linuspahl, line 57: I just saw that when we define the `width` this way, we depend on the width of the left col (for the form) to always be `33%`. Once the left column width changes, it can look like this:
image

Instead we can use `width: 100%` and `position: absolute`:
image

linuspahl (changes requested):

PR #14069, opened 2022-11-28
Allow selecting which columns should be displayed for streams overview.

linuspahl:
Please note, this PR is based on https://github.com/Graylog2/graylog2-server/pull/14066 which needs to be merged first.

## Motivation and Context

With this PR we are implementing the option to select which columns you want to display for the streams overview.

image

## Screenshots (if appropriate):

Fixes https://github.com/Graylog2/graylog2-server/issues/13853

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

/nocl

PR #14084, opened 2022-11-29
Improve dynamic column width for `EntityDataTable`.

linuspahl:
_Please note, this PR is based on https://github.com/Graylog2/graylog2-server/pull/14069 which needs to be merged first._

## Motivation and Context

Before this PR the width of a column in the `EntityDataTable` was dynamic and depended on the column content. This means the column width can change when using the pagination.

To be able to define a width which is relative to other columns but stays the same when the content changes, we are now calculating the width for each column.

When using the `EntityDataTable`, you can change the width of a column, by adjusting the related column renderer. Column renderers can have
- either a `width` defined as a fraction like `2`. If no width is defined we are using a default of `1`.
  The width defines which fraction of the assignable space the column should claim. It works similar to the css attribute `flex`.
  Optionally you can also define a `minWidth` to override the default min width for flexible columns. This can be helpful to ensure a column has enough space, no matter how large the table width is.
- or a `staticWidth` in px. Useful when the cells contain e.g. just an icon and its width never changes.

/nocl


graylog2-web-interface/src/components/common/EntityDataTable/types.ts
```diff
@@ -25,12 +25,14 @@ export type Column = {
...
+// A column render should have either a `width` and optionally a `minWidth` or only a `staticWidth`.
 export type ColumnRenderer<Entity extends { id: string }> = {
-  maxWidth?: string,
   renderCell?: (entity: Entity, column: Column) => React.ReactNode,
   renderHeader?: (column: Column) => React.ReactNode,
   textAlign?: string,
-  width?: string,
+  minWidth?: number, // px
+  width?: number, // fraction of unassigned table width, similar to CSS unit fr.
+  staticWidth?: number, // px
```
linuspahl, line 35: I tried to define the type in a way that you can only I tried to define the type in a way that you can only deine a `staticWidth` or a `width` and optionally a `minWidth`, but it is problematic because we are merging the custom and default column definition. Especially the custom column definitions provided by e.g. the `StreamsOverview` are often not complete.

maxiadlovskii (approved):

PR #14088, opened 2022-11-29
Change bulk indexing retry failure log level to warn

### LINKED ISSUE #14086: Bulk indexing retries should be logged as a warning, not error
## Expected Behavior

Bulk index retries should be logged as a warning.

## Current Behavior

Currently an error is being logged on bulk index retries:
https://github.com/Graylog2/graylog2-server/blob/bb71c4542ce575cee1ee011bf1e41ede7c933b90/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java#L87

While the retries might indicate an issue with the search backend, it might be a temporary condition. And since the retries are not limited, no data should be lost (yet).

## Possible Solution

Change the log level to `warn`.

## Context

Noticed failing bulk index requests when OpenSearch responds with 429 errors, which can sometimes happen due to temporary memory pressure, for example.

## Your Environment

* Graylog Version: Graylog Cloud 4.3.10 (195)
* Elasticsearch Version: OpenSearch 1.2 (AWS OpenSearch Service)

danotorrey:
Reduce the log level for bulk indexing retry attempts from `error` to `warning`, since they sometimes indicate a temporary condition, which will resolve with time.

Closes #14086

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/issue-14086.toml
+++ b/changelog/unreleased/issue-14086.toml
@@ -1,0 +1,15 @@
+type = "c"
+message = "Changed bulk indexing retry failure log-level from error to warning."
+
+issues = ["14086"]
+pulls = ["14088"]
+
+details.user = """
+The log-level for Opeansearch bulk indexing retry attempts has been changed from `ERROR` to `WARN`.
+
+While bulk indexing retries might indicate an issue with the Opensearch backend, it also might be a temporary condition
+that would resolve on its own (for example, temporary high memory pressure). Warn is a better log-level for such a case.
+
+Example message:
+> WARN Caught exception during bulk indexing: [specific error], retrying (attempt #1).
+"""
--- a/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/messages/Messages.java
@@ -84,7 +84,7 @@ public class Messages {
                     @Override
                     public <V> void onRetry(Attempt<V> attempt) {
                         if (attempt.hasException()) {
-                            LOG.error("Caught exception during bulk indexing: {}, retrying (attempt #{}).", attempt.getExceptionCause(), attempt.getAttemptNumber());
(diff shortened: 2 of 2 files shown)


ryan-carroll-graylog (approved):

kingzacko1 (approved):

dennisoelkers (approved):

changelog/unreleased/issue-14086.toml
```diff
@@ -0,0 +1,15 @@
+type = "c"
+message = "Changed bulk indexing retry failure log-level from error to warning."
+
+issues = ["14086"]
+pulls = ["14088"]
+
+details.user = """
+The log-level for Opeansearch bulk indexing retry attempts has been changed from `ERROR` to `WARN`.
```
boosty, line 8: Small typo in Opeansearch.

Should we still mention Elasticsearch as well? Otherwise it sounds as if we only applied this change in the case of OpenSearch.

boosty (commented):
Thanks for the quick change! Just left a comment regarding the change log snippet.

changelog/unreleased/issue-14086.toml
danotorrey: Thanks @boosty. I think mentioning both can't hurt. I'll open a small PR to fix the typo and mention both.

boosty:
Hi @danotorrey, the 5.0 GA build was already started, I think it's a bit too late now. 

There might be more cases where we want to change the log level, so from my side it would be ok to collect a few of them and release them together in 5.1. These changes are not urgent, and they need to be communicated properly in the changelog (like you did here), as users might process the server log, so it is a kind of interface change.

danotorrey:
> There might be more cases where we want to change the log level, so from my side it would be ok to collect a few of them and release them together in 5.1. These changes are not urgent, and they need to be communicated properly in the changelog (like you did here), as users might process the server log, so it is a kind of interface change.

@boosty Good point, makes sense.

danotorrey:
Thanks @boosty. This all makes sense. 

> Could you explain what you mean by that? This change is in master already, so should be part of `5.1`, which is fine. There is nothing to be backported I think.

Thanks. Oops, yes of course. That is correct.

PR #14090, opened 2022-11-30
Fix collector templates on Windows

mpfz0r:
For some reaseon Freemarker needs
escaped backslashes, when used within a variable.

/nocl


thll (approved):
LGTM and works

PR #14091, opened 2022-11-30
Fix collector templates on Windows

mpfz0r:
For some reaseon Freemarker needs
escaped backslashes, when used within a variable.

(cherry picked from commit 9c6a3d8c35d56f546516d90468c2b099bd82c296)

/nocl


thll (approved):
LGTM

PR #14096, opened 2022-11-30
Adding bulk action to assign multiple streams to index set.

dennisoelkers:
## Motivation and Context

This PR is adding a bulk action to the REST API which allows adding multiple streams to a single index set. This allows an easy migration of streams to a newly created index set.

The action is used like this:

`PUT /api/streams/indexSet/{id}`

with a body that contains an array of stream ids.

Alternatively, we can use the TypeScript stubs:

```
import { Streams } from '@graylog/server-api';

[...]

await Streams.assignToIndexSet(indexSetId, [streamId1, streamId2, ...]);
```

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -491,14 +480,61 @@ public Response cloneStream(@ApiParam(name = "streamId", required = true) @PathP
...
+    @ApiOperation(value = "Assign multiple streams to index set")
+    @ApiResponses(value = {
+            @ApiResponse(code = 404, message = "Index set not found.")
+    })
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @AuditEvent(type = AuditEventTypes.STREAM_UPDATE)
+    public Response assignToIndexSet(@ApiParam(name = "indexSetId", required = true) @PathParam("indexSetId") String indexSetId,
+                                    @ApiParam(name = "JSON body", required = true) @Valid @NotNull List<String> streamIds) {
+        checkPermission(RestPermissions.INDEXSETS_READ, indexSetId);
```
janheise, line 503: INDEXSETS_EDIT?

dennisoelkers: No, the index set itself is not manipulated. Also, this would result in different resources than when creating/updating a single stream (which was the default way to assign a stream to an index set before).

PR #14097, opened 2022-11-30
Support more than 10 shutdown hooks in graceful shutdown service

thll:
The GracefulShutdownService used a SynchronousQueue with a fixed capacity. If all threads were busy, the executor would reject additional tasks.

We fix this by using an unbounded blocking queue for the executor. We still limit the number of concurrently run hooks but allow additional hooks to be queued for execution.

To reproduce the problem, assign > 10 outputs to an active stream and then shut down the server. Before aplpying the fix, you should see a `RejectedExecutionException` in the server log. After applying the fix, shutdown should proceed as planned.


AntonEbel (approved):
LGTM.

PR #14098, opened 2022-11-30
integration test for scripting api messages

todvora:
The heading says it all :-)

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
@@ -54,7 +54,6 @@ import java.util.Map;
 
 import static io.restassured.RestAssured.given;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.withPrecision;
 import static org.assertj.core.api.Java6Assertions.within;
 import static org.graylog.testing.completebackend.Lifecycle.CLASS;
 import static org.hamcrest.CoreMatchers.not;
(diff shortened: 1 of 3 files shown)


luk-kaminski (approved):
LGTM, thx!

full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
```diff
@@ -610,6 +612,31 @@ void testErrorHandling() {
...
+                        }
+                        """)
+                .post("/search/messages")
+                .then()
+                .log().ifStatusCodeMatches(not(200))
+                .statusCode(200);
+
+        validateSchema(validatableResponse, "field: source", "string", "source");
+        validateSchema(validatableResponse, "field: facility", "string", "facility");
+        validateSchema(validatableResponse, "field: level", "numeric", "level");
```
luk-kaminski, line 632: You could consider replacing "string" with "ResponseEntryDataType.STRING.toString()" to avoid magic-string usage.

todvora: IDK, I probably prefer the hardcoded value, as it holds our expectation in the test without the dependency. If we change the value of ResponseEntryDataType.STRING, our test fails and we'll know that we have changed the contract - it's not "string" anymore. That wouldn't happen if we reuse the same enum value on both ends.

luk-kaminski: The question is what we want to test here: that the value is "string" or that the value has a string from the proper enum. It is up to you to decide, I do not have a strong opinion. :)

todvora: I'd keep the hardcoded value. That's what our users will see and expect.  What we use internally is from my perspective not important.

PR #14110, opened 2022-12-01
Prevent changing unmodifiable Event Definition condition types

ryan-carroll-graylog:
Prevent updating unmodifiable Event Definition condition types.

## Description

Updates the Event Definitions Edit page's Condition form to disallow changing some Event Definition Processor Config types.

## Motivation and Context

Graylog 5.0 Introduces some features that make use of Event Definitions, but have ties to other entities in the application and should not have their Processor Configs changed.

See: https://github.com/Graylog2/graylog-plugin-enterprise/issues/4290

## How Has This Been Tested?

Locally in development environment and unit testing.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
```diff
@@ -74,12 +74,15 @@ default Set<String> requiredPermissions() {
...
+     * as opposed to being used for background functionality.
      *
-     * @return whether the config type can be exported in a Content Pack
+     * i.e. Should this be able to be exported in a Content Pack or modified on the Event Definitions Page
+     *
+     * @return whether the config type should be presented to users
      */
     @JsonIgnore
-    default boolean isContentPackExportable() {
+    default boolean isUserPresentable() {
```
ryan-carroll-graylog, line 85: This is the check that was added to prevent users from exporting Sigma configs in Content Packs. I repurposed it to be more general since it seems like they really cover the same use case: whether or not this event processor type is meant to just serve a background purpose. But I'm open to suggestions if anyone sees issues with this.

Also open to suggestions on the naming/wording.

danotorrey:
I am around and working on Monday, so I should have some time to take a look at this then.

graylog2-server/src/main/java/org/graylog/events/processor/EventProcessorConfig.java
kingzacko1: This works well for this particular use case, but I wonder if there might ever be a case in the future where we want the ability to have `isContentPackExportable` and `isUserPresentable` be different. I can't think of a concrete example but we may want to add a separate method instead of combining the two.

ryan-carroll-graylog: Ok I think that makes sense. It certainly wont hurt anything to separate out the logic so I'll go ahead and separate out into two functions.

simonychuang (approved):
Frontend changes look good, we should get approval from someone on the core team since this is going into that repo though.

danotorrey (approved):
LGTM 👍

ryan-carroll-graylog:
> Frontend changes look good, we should get approval from someone on the core team since this is going into that repo though.

Good call @simonychuang , 

> Frontend changes look good, we should get approval from someone on the core team since this is going into that repo though.

Good call @simonychuang . @linuspahl @dennisoelkers would either of you mind taking a look at this as it affects the core Event Definition code base? The intent is that these changes will only affect Sigma Event Definitions and not other ED processor types, but it would be great if you could take a look and verify that there aren't any issues with this.

kingzacko1:
@ryan-carroll-graylog My last commit just converted the frontend component that was failing checks to be a functional TypeScript component. @zeeklop @simonychuang could one of y'all look at that and make sure it looks okay? I've done a few components now but always like another set of eyes.

ryan-carroll-graylog:
> @ryan-carroll-graylog My last commit just converted the frontend component that was failing checks to be a functional TypeScript component. @zeeklop @simonychuang could one of y'all look at that and make sure it looks okay? I've done a few components now but always like another set of eyes.

Right on, thanks @kingzacko1

kingzacko1 (approved):

graylog2-web-interface/src/components/event-definitions/types.d.ts
```diff
@@ -15,8 +15,12 @@
...
  */
 interface EventDefinitionType {
-  type: string;
-  displayName: string;
+  type: string,
+  displayName: string,
+  sortOrder: number,
+  description: string,
+  defaultConfig: any,
+  formComponent: any,
```
linuspahl, line 23: Regarding `any`, here we can type it like this:
```
  defaultConfig: EventDefinition['config'],
  formComponent: React.ComponentType<React.ComponentProps<{
    eventDefinition: EventDefinition,
    currentUser: UserJSON,
    validation: { errors: { [key: string]: Array<string> } },
    onChange: (name: string, newConfig: EventDefinition['config']) => void,
    action: string,
  }>
```

graylog2-web-interface/src/components/event-definitions/event-definition-form/EventConditionForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+    return sortedEventDefinitionTypes()
+      .map((type) => ({ label: type.displayName, value: type.type }));
+  };
+
+  const handleEventDefinitionTypeChange = (nextType) => {
+    const conditionPlugin = getConditionPlugin(nextType);
+    const defaultConfig = conditionPlugin?.defaultConfig || {};
+
+    onChange('config', { ...defaultConfig, type: nextType });
+  };
```
linuspahl, line 75: Regarding the types, we can type it like this:

```
  const handleEventDefinitionTypeChange = (nextType: string) => {
    const conditionPlugin = getConditionPlugin(nextType);
    const defaultConfig = conditionPlugin?.defaultConfig || {} as EventDefinition['config'];

    onChange('config', { ...defaultConfig, type: nextType });
  };
```

graylog2-web-interface/src/components/event-definitions/event-definition-form/EventConditionForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const disabledSelect = () => {
+    return !formattedEventDefinitionTypes().some((edt) => eventDefinition.config.type === edt.value) && action === 'edit';
+  };
+
+  const eventDefinitionType = getConditionPlugin(eventDefinition.config.type);
+
+  const eventDefinitionTypeComponent = eventDefinitionType?.formComponent
+    ? React.createElement(eventDefinitionType.formComponent, {
```
linuspahl, line 98: For some reason I had to type `createElement` and ended up typing it like this, because I could not figure out a better solution:
`React.createElement<React.ComponentProps<any>>`

linuspahl (commented):
Frontend looks good to me and I could not find a problem with the existing functionality. I just added a suggestion regarding the type definitions.

ryan-carroll-graylog:
Thanks for taking a look and for all the feedback @linuspahl! I'm working on incorporating your suggestions.

ryan-carroll-graylog:
I've added @linuspahl 's suggestions and had to do some minor refactoring to do some minor refactoring to the front end because of some changes made in another PR. 

@zeeklop @kingzacko1 , when y'all have a sec can you take a look at the front end piece (mostly just `EventConditionForm.tsx`
) to make sure my refactoring makes sense

zeeklop (commented):
LGTM

kingzacko1:
> @zeeklop @kingzacko1 , when y'all have a sec can you take a look at the front end piece (mostly just `EventConditionForm.tsx` ) to make sure my refactoring makes sense

Still looks good and functions as expected!

PR #14123, opened 2022-12-02
Expose batchSize read-only in API

### LINKED ISSUE #9461: API `System / Configuration` does not include `output_batch_size`
## Expected Behavior
Settings that are fundamental to tune Graylog should be visible from the API. The `output_batch_size` is not be visible via API. 

## Current Behavior

image

## Possible Solution
Include the `output_batch_size` at the above visible read-only cluster configuration API endpoint.

## Context
This setting that will be changed in bigger environments can't be checked via API. That makes it impossible to check if the settings are made right - what is mostly for docker based installations important.

## Your Environment

* Graylog Version: 4.0

patrickmann:
Add `output_batch_size` to the values returned in the `system/configuration` API

Resolves #9461

### DIFF
--- a/changelog/unreleased/issue-9461.toml
+++ b/changelog/unreleased/issue-9461.toml
@@ -1,0 +1,5 @@
+type = "added"
+message = "Added output_batch_size to the values exposed by the system/configuration API call."
+
+issues = ["9461"]
+pulls = ["14123"]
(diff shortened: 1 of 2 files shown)


moesterheld (approved):
works as expected.

PR #14177, opened 2022-12-09
Avoiding `GET` APIs for cluster/deflector health.

### LINKED ISSUE #14164: "An HTTP line is larger than 4096 bytes." against the search provider health and name endpoints with large number of indexes
When polling endpoints like `/api/system/indexer/cluster/name` or `/api/system/indexer/cluster/health`, it's possible that a URI longer than 4096 bytes can be generated if installation has a large number of indexes, even if we request by index prefix wildcards.

`2022-11-29 21:23:14,090 ERROR: [org.graylog.storage](http://org.graylog.storage/).elasticsearch7.ClusterAdapterES7 - An error occurred: (method [HEAD], host [http://redacted:9200], URI [/log_dead_*,log_dih_*,log_b2bi_*,graylog_*,log_genosonar-test_*,gl-events_*,gl-failures_*,gl-system-events_*,
<redacted, full output in HS ticket>
?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open%2Cclosed&allow_no_indices=false], status line [HTTP/1.0 400 Bad Request])`

Worked with customer to adjust Opensearch config variable _http.max_initial_line_length_ to 8kb as a temporary workaround. 

[HS-1291154535](https://app.hubspot.com/contacts/8267556/ticket/1291154535)

* Graylog Version: 4.3.9
* Java Version: 11
* Elasticsearch Version: Opensearch 1.3.6
* MongoDB Version: 4.4

dennisoelkers:
## Motivation and Context

Prior to this change, retrieving the cluster/deflector health relied on Elasticsearch/Opensearch endpoints taking a list of indices as a query parameter and we passed e.g. the list of all existing index wildcards to this. This can lead to a request being generated that has a query string that is too big, if a high number of index sets exist, as seen in #14164.

Unfortunately, there are no endpoints for this taking the list of indices outside of a query parameter. Therefore this PR is now doing two things to fix this:

  - General cluster health is retrieved with the generic cluster health endpoint without passing index names. This returns a health status (red, yellow, green) in the same way, but is not specific to the indices used by Graylog.
  - Using the CAT API to resolve and retrieve the health for the individual deflector aliases configured. We do retrieve the complete list of aliases and indices and filter out the ones we are interested in. Due to the compact nature of the CAT API this should work well for a highly utilized ES/OS instance as well.

Fixes #14164.

## Types of changes

## Checklist:


graylog-storage-opensearch2/src/main/java/org/graylog/storage/opensearch2/ClusterAdapterOS2.java
```diff
@@ -292,10 +290,45 @@ private Optional<ClusterHealthResponse> clusterHealth(Collection<String> indices
...
+                    }
+                    if (status1.equals(HealthStatus.Green)) {
+                        return status2.equals(HealthStatus.Green) ? 0 : 1;
+                    }
+                    if (status2.equals(HealthStatus.Green)) {
+                        return -1;
+                    }
+                    if (status2.equals(HealthStatus.Red)) {
+                        return 1;
+                    }
```
janheise, line 329: Intuitively, I would have thought that `compareTo()` would work, too?

janheise: to clarify, because the extract seems to small (I thought I selected the full comparison): just compare the enum in min(left, right) - because the ordering inside the enum is, what we expect

janheise (approved):

PR #14183, opened 2022-12-12
Checking if the sum of size and from does not exceed ES/OS limits

luk-kaminski:
## Description
Checking if the sum of size and from does not exceed ES/OS limits.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/ScriptingApiResource.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/ScriptingApiResource.java
@@ -66,6 +66,8 @@ import static org.graylog2.shared.utilities.StringUtils.splitByComma;
 @RequiresAuthentication
 public class ScriptingApiResource extends RestResource implements PluginRestResource {
 
+    private final int MAX_RESULT_WINDOW = 10_000; //default "index.max_result_window" for both ES and OS
+
     private final SearchExecutor searchExecutor;
     private final EventBus serverEventBus;
     private final SearchRequestSpecToSearchMapper searchCreator;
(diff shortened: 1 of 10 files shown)


todvora (approved):
LGTM, thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/ScriptingApiResource.java
```diff
@@ -96,6 +98,11 @@ public ScriptingApiResource(final SearchExecutor searchExecutor,
     @NoAuditEvent("Creating audit event manually in method body.")
     public TabularResponse executeQuery(@ApiParam(name = "queryRequestSpec") @Valid MessagesRequestSpec messagesRequestSpec,
                                         @Context SearchUser searchUser) {
+
+        if (messagesRequestSpec.from() + messagesRequestSpec.size() > MAX_RESULT_WINDOW) {
```
dennisoelkers, line 102: Wouldn't this also throw an error and abort searches for users which have raised the maximum result window?

luk-kaminski: Yes, it would. 
The question is if we want to take care of this unusual/unlikely possibility and perform additional query to check this setting for an index we query (adding some performance overhead) or are we ok with the current, simple approach? It is hard for me to imagine people getting thousands of messages with scripting API, so I am a little bit hesitant to add complexity and perf. overhead just to make it extremely bulletproof...

dennisoelkers: What would be the benefit of the current approach? Either way the user would get back an error/exception when the overall message size exceeds `n` (defaulting to 10000). If we are worried about the default message coming back from ES/OS is not understandable, wouldn't it be better to just catch it and return a better one instead of doing a pre-flight check?

luk-kaminski: The benefit is we don't reach ES/OS at all in this case.
But if you believe it is not worth it, or risky, we can throw this PR away, it is nothing big anyway...

dennisoelkers (approved):
Looks good, works great! :pray:

PR #14186, opened 2022-12-12
Start global JSON path input on leader node only

### LINKED ISSUE #14074: JSON path input should run only on leader node when global is enabled
## Expected Behavior

The "JSON path value from HTTP API" input should only run on the leader node if "global" is enabled.

## Current Behavior

Currently, the "JSON path value from HTTP API" input runs on all nodes when "global" is enabled.

## Possible Solution
Override the `onlyOnePerCluster` method in `JsonPathInput` and return `true`.

* Graylog Version: 5.0.0

thll:
## Description

The `JSON path value from HTTP API` input will now only run on the leader node,
if the `Global` option has been selected in the input configuration.
Previously, the input was started on all nodes in the cluster.

## Motivation and Context

Fixes #14074 

## How Has This Been Tested?

Smoke test in a two-node cluster

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,3 +4,9 @@ Upgrading to Graylog 5.1.x
 ## New Functionality
 
 ## Breaking Changes
+
+## Behaviour Changes
+
+The `JSON path value from HTTP API` input will now only run on the leader node,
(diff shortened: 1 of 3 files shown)


AntonEbel (approved):
LGTM

PR #14199, opened 2022-12-13
Make email socket timeouts configurable

thll:
## Description

Adds two config options to configure socket connection and read timeouts for email sending:

| Option                                      | Action | Description                                                                                |
|---------------------------------------------|--------|--------------------------------------------------------------------------------------------|
| `transport_email_socket_connection_timeout` | added  | Connection timeout for establishing a connection to the email server. Default: 10 seconds. |
| `transport_email_socket_timeout`            | added  | Read timeout while communicating with the email server. Default: 10 seconds.               |

## Motivation and Context

The previous default was 60 seconds and was not configurable. Waiting this long, and therefore potentially blocking a request from the UI to the backend, may cause a load balancer between the UI and the backend to abort the call and return a `504` to the browser.

## How Has This Been Tested?

Connection timeout: configured Graylog to use `example.com:81` as the email server
Read timeout: ran `ncat -l 9026` and configured `localhost:9026` as the email server

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/shared/email/EmailFactory.java
```diff
@@ -119,6 +120,9 @@ private <T extends Email> T applyConfig(T email) throws EmailException {
             email.setFrom(configuration.getFromEmail());
         }
 
+        email.setSocketConnectionTimeout(Ints.saturatedCast(configuration.getSocketConnectionTimeout().toMillis()));
+        email.setSocketTimeout(Ints.saturatedCast(configuration.getSocketTimeout().toMillis()));
+
```
moesterheld, line 125: can this be a problem for the user that the validation is always done using longs, however this is set to a integer? 
I know that durations larger than Integer.MAX_VALUE won't be seen in reality, so I'm fine with that in any way.

moesterheld (approved):
lgtm, just one small comment

graylog2-server/src/main/java/org/graylog2/shared/email/EmailFactory.java
thll: I only added the validation to fail fast and not wait until the first email is sent, but I thought it was ok to silently limit the timeout here to ~24 hours. It doesn't make much sense anyway to configure such a high value. It looks like we are doing this at a couple of places already when dealing with durations.

Issue #14142, opened 2022-12-05
IP Address search using ip slash notation shows a validation warning


drewmiranda-gl:
IP Address search using ip slash notation shows a validation warning:

> Invalid value type: Type of source_ip is ip, cannot use value 10.0.0.0/8

![image](https://user-images.githubusercontent.com/107503402/205696235-c9750298-da9a-4d5e-8c58-43150262314f.png)

## Expected Behavior

No validation warnings are shown if the search query is valid.

## Current Behavior

A validation warning is shown even though the query works and is valid.

## Steps to Reproduce (for bugs)

Search using a search query that contains ip slash notation, such as: `source_ip:"10.0.0.0/8"`

## Context

Warning is confusing and incorrect.

## Your Environment

* Graylog Version: Graylog 5.0.0
* Java Version: 17.0.5
* Elasticsearch Version: Opensearch 1.3.4
* MongoDB Version: 5.0.14
* Operating System: Ubuntun Server 20.04 LTS
* Browser version: 108.0.5359.94

PR #14204, opened 2022-12-13
Query validation can handle CIDR format of IP

todvora:
Fixes https://github.com/Graylog2/graylog2-server/issues/14142

## Description
CIDR format of the IP requires some special handling in the query validation 

## How Has This Been Tested?
Added unit tests 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/FieldTypeValidationImpl.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/FieldTypeValidationImpl.java
@@ -17,12 +17,10 @@
 package org.graylog.plugins.views.search.validation;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.net.InetAddresses;
 import org.apache.commons.net.util.SubnetUtils;
 import org.apache.commons.validator.routines.InetAddressValidator;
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/validation/FieldTypeValidationImpl.java
```diff
@@ -89,6 +90,24 @@ private static boolean isDate(final String dateCandidate) {
...
+    }
+
+    private static boolean isValidCidrIp(String cidrIp) {
+        try {
+            new SubnetUtils(cidrIp);
+            return true;
+        } catch (IllegalArgumentException e) {
+            return false;
+        }
+    }
```
luk-kaminski, line 109: I'd consider moving it to separate class, as the responsibility of this class has grown to multiple different types of validation. But it is not a must...

todvora: I think the class is still reasonably short and focused on one thing, I'd keep the method there now. If it grows significantly in the future, or we'll have the need to reuse the IP validation elsewhere, then we can always extract it.

PR #14221, opened 2022-12-15
Disable GracefulShutdownService test which is flaky on Jenkins

thll:
It's not a critical test so I think it's OK to disable it.
Could probably be fixed by adjusting the timeouts but that would potentially increase test runtime even more and is also guesswork.

/nocl

### DIFF
--- a/graylog2-server/src/test/java/org/graylog2/system/shutdown/GracefulShutdownServiceTest.java
+++ b/graylog2-server/src/test/java/org/graylog2/system/shutdown/GracefulShutdownServiceTest.java
@@ -18,9 +18,9 @@ package org.graylog2.system.shutdown;
 
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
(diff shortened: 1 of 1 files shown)


graylog2-server/src/test/java/org/graylog2/system/shutdown/GracefulShutdownServiceTest.java
```diff
@@ -142,6 +143,7 @@ public void failRegisterAndUnregisterDuringShutdown() {
                 .isInstanceOf(IllegalStateException.class);
     }
 
+    @Disabled("Disabled because it is flaky on Jenkins. Could maybe be fixed by adjusting timeouts.")
```
bernd, line 146: How about using a JUnit environment variable condition to let the test run outside of Jenkins?

```
@DisabledIfEnvironmentVariable(named = "CI", matches = "true", disabledReason = "Flaky on Jenkins")
```

bernd (changes requested):

graylog2-server/src/test/java/org/graylog2/system/shutdown/GracefulShutdownServiceTest.java
thll: Good idea. Thank you!

PR #14225, opened 2022-12-15
Stop output only once when deleting globally

thll:
When an output was deleted, we would send an OutputDeletedEvent. However, we would also remove the output from all streams, emitting a StreamsChangedEvent. The OutputRegistry would handle both events to stop the output. This lead to concurrent stopping of the output, which in the case of journaled outputs, could cause issues when cleaning up the journal.

We can fix this by not emitting the OutputDeletedEvent because the OutputRegistry can determine by the StreamsChangedEvent if an output must be deleted.


mpfz0r (approved):
LGTM

Issue #13821, opened 2022-10-28
Field autocompletion does not work well together with "backspace"


luk-kaminski:
## Steps to Reproduce:
1. In the search bar type "http_meth". Note that "http_method" appears as a suggetion.
2. Continue typing, using wrong characters, i.e. "http_methxyz". Suggestions disappear, which is fine.
3. Using "backspace", keep removing the wrong characters. No matter if you go back to "http_meth" or even delete more, reaching "http_me", there will be no suggestions.
4. The only way for suggestions to appear back is to use at least one correct character after using backspace.

luk-kaminski:
[Open Defect in TestQuality](https://web.testquality.com/site/graylog/project/14870?defectId=12350)
- Test [Query Input has auto-completion (showing up after starting to type)](https://web.testquality.com/site/graylog/project/14870/runs/171283?resultId=3015010) in Run [Łukasz testing Query input](https://web.testquality.com/site/graylog/project/14870/runs/171283) (linked to Cycle [Query Input](https://web.testquality.com/site/graylog/project/14870/cycles/27707)).

PR #14233, opened 2022-12-16
Define shouldShowCompletions method for FieldNameCompletion

maxiadlovskii:
## Description
When the user types smth in query input the react-ace fills the completions list using the appropriate method. When a user types the next character in some cases logic doesn't runs this method again and uses the previous filtrated list. When the user presses backspace this method never runs. That is why nothing is shown in the list. 
This PR adds the `shouldShowCompletions` method for `FieldNameCompletion`. It defines cases when we have to show completions in case of editing field name

## Motivation and Context
fix: https://github.com/Graylog2/graylog2-server/issues/13821

## Types of changes

## Checklist:

/nocl


linuspahl:
Looks good to me, since we paired on this logic it would be good if a third person can have a look.

PR #14235, opened 2022-12-16
Display saved search title as anchor instead of button.

linuspahl:
## Motivation and Context

Previously we were rendering the title of a saved search in the saved searches overview as a button.
With this PR we are rendering the title as a link instead.

It has a few benefits, it reflect the actual intention of the element, it unifies the title element with similar cases like the dashboards overview and users able to see the URL of the saved search and can copy it.

/nocl


dennisoelkers (approved):
:pray:

PR #14240, opened 2022-12-16
use separate inputs endpoint for stream rules

AntonEbel:
As discussed in this [issue](https://github.com/Graylog2/graylog-plugin-cloud/issues/1066), we have decided to also provide forwarder inputs for the stream rule "match input". This PR is the first step to make this possible.  

## Types of changes


grotlue (approved):
Frontend-wise it looks good for me!

For testing this I just tested it on on premise to check that it still works as before (shows inputs).

If I understand correctly there is another PR for this feature in Enteprise and Cloud, right? So I didn't test those parts in scope of this PR.
Screenshot 2023-01-13 at 10 29 30

kodjo-anipah (approved):
works as expected, unit test would be nice

graylog2-web-interface/src/stores/inputs/StreamRulesInputsStore.ts
```diff
@@ -0,0 +1,84 @@
...
+          (error) => {
+            UserNotification.error(`Fetching Stream Rule Inputs List failed with status: ${error}`,
+              'Could not retrieve Stream Rule Inputs');
+          },
+        );
+
+      StreamRulesInputsActions.list.promise(promise);
+    },
+  }),
+);
```
linuspahl, line 84: I just saw the new store by coincidence. Is there a benefit in using a store in this case? We started using `@tanstack/react-query` instead, to be able to get rid of reflux at some point.

PR #14262, opened 2022-12-20
Sort streams case-insensitive in API.

dennisoelkers:
## Motivation and Context

Prior to this PR, sorting streams by title/description/... sorted them lexicographically, but case-sensitive. This is most probably not what users are expecting.

This change is now adding functionality which is lower-casing the sort field before applying the sort to it.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/streams/PaginatedStreamService.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/PaginatedStreamService.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.streams;
 
+import com.google.common.collect.ImmutableList;
 import com.mongodb.client.AggregateIterable;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.model.Aggregates;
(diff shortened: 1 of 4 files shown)


janheise (approved):
LGTM!

graylog2-server/src/main/java/org/graylog2/streams/PaginatedStreamService.java
```diff
@@ -57,20 +58,28 @@ public long count() {
...
+                    ))
+                    .add(Aggregates.set(new Field<>("index_set_title", doc("$first", "$index_set.title"))));
+        }
+
+        if (isStringField(sortField)) {
+            pipelineBuilder.add(Aggregates.set(new Field<>("lower" + sortField, doc("$toLower", "$" + sortField))))
+                    .add(Aggregates.sort(getSortBuilder(order, "lower" + sortField)))
+                    .add(Aggregates.project(Projections.exclude("index_set", "lower" + sortField)));
+        } else {
+            pipelineBuilder.add(Aggregates.sort(getSortBuilder(order, sortField)));
```
janheise, line 79: maybe a `Projections.exclude("index_set"` missing here?

dennisoelkers: I decided to drop it, because we do not really need to strip it out (it will just not be used when creating the `StreamDTO`) and it adds yet another branch (because we would need another `if (sortField.equals("index_set_title") {...}`. I could remove it from the previous `Projections.exclude` too though, for consistency.

janheise (approved):

linuspahl (approved):

PR #14265, opened 2022-12-20
Added test to aggregate messages by streams

todvora:
Some of the scripting api integration tests are flaky. The assumption is that there may be a problem/async/timing issue between stream and stream rule creation and message ingestion. In the moment the message is ingested, the stream could be not fully initialized yet. 

This test verifies that all messages are routed to correct streams by aggregating by stream ID. 

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
@@ -37,6 +37,7 @@ import org.hamcrest.Matchers;
 import org.joda.time.DateTime;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.shadow.com.univocity.parsers.csv.Csv;
 import org.junit.jupiter.params.shadow.com.univocity.parsers.csv.CsvParser;
 import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
```diff
@@ -119,6 +121,36 @@ public void beforeAll() {
...
+                             }
+                           ]
+                        }
+                         """)
+                .post("/search/aggregate")
+                .then()
+                .log().ifStatusCodeMatches(not(200))
+                .statusCode(200);
+
+        validatableResponse.assertThat().body("datarows", Matchers.hasSize(3));
```
luk-kaminski, line 148: I'm not sure if you shouldn't verify the whole response with one assertion.
If first assertion fails, you will know smth is wrong, but you will not have the full picture, you will just learn that you miss one stream. But if you believe it is enough, we can leave it as it is.

luk-kaminski (approved):

full-backend-tests/src/test/java/org/graylog/plugins/views/ScriptingApiResourceIT.java
dennisoelkers: I like to add a `.log().ifValidationFails()` for those cases, so I will get the response body when an assertion fails.

todvora: Thanks, added logging Dennis suggested :+1:

PR #14270, opened 2022-12-20
Honor query parameter in auth backend pagination

### LINKED ISSUE #13023: filtering / searching Authentication Backends is broken
https://community.graylog.org/t/search-backends-via-the-api/24703

filtering authentication backends should work

## Expected Behavior
when you filter authentication backends, only backends that have a title or description that matches your search should appear.  You should also be able to query for specific backends via the API using the `query` param (according to the swagger doc)

## Current Behavior
when you filter authentication backends, nothing happens.  The "search" button starts spinning and never stops....but no filtering occurs.  Any attempt to use the API and provide a `query` param when GETing all backends fails with a 415

1. login to the GL admin UI
2. go to Authentication
3. create some backend services
4. try to search for anything you created

## Context
using salt to manage our GL instance and I'd like to GET an authentication backend using the `query` param via the API to avoid nested for/each loops

## Your Environment
* Graylog Version: 4.2.7
* Java Version: 11
* Elasticsearch Version:
* MongoDB Version: 
* Operating System: ubuntu
* Browser version: chrome latest

patrickmann:
Resolves #13023

We provide a filter field in the auth backend form, but did not evaluate the contents.
Apply the filter query, the same way it is handled in other forms.


kodjo-anipah (approved):
LGTM

PR #14275, opened 2022-12-21
Fix unit test to avoid issues with day boundary or DST

patrickmann:
Unit test for date conversion should not use `now()`, as this may cause problems depending on when the test is executed.

/nocl


thll (approved):
LGTM

Issue #14214, opened 2022-12-14
Autocomplete doesn't work for fields that are typed as numbers (long, int), warning thrown in server.log


drewmiranda-gl:
When typing in the searchbox following a field that has a type of long or int, no autocomplete values are returned.

An ElasticsearchClient exception is logged as WARN in the graylog `server.log`

```
WARN  [ElasticsearchClient] Failed to parse exception
com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `org.graylog.storage.elasticsearch7.errors.Cause$Builder`, problem: Missing required properties: indexUuid index
 at [Source: (ByteArrayInputStream); line: 1, column: 112] (through reference chain: org.graylog.storage.elasticsearch7.errors.AutoValue_ResponseError$Builder["error"]->org.graylog.storage.elasticsearch7.errors.AutoValue_ElasticsearchError$Builder["root_cause"]->java.util.ArrayList[0])
	at com.fasterxml.jackson.databind.exc.ValueInstantiationException.from(ValueInstantiationException.java:47) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.DeserializationContext.instantiationException(DeserializationContext.java:2047) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.DeserializationContext.handleInstantiationProblem(DeserializationContext.java:1400) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.wrapInstantiationProblem(BeanDeserializerBase.java:1866) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.BuilderBasedDeserializer.finishBuild(BuilderBasedDeserializer.java:202) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.BuilderBasedDeserializer.deserialize(BuilderBasedDeserializer.java:217) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer._deserializeFromArray(CollectionDeserializer.java:355) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:244) ~[graylog.jar:?]
	at com.fasterxml.jackson.databind.deser.std.CollectionDeserializer.deserialize(CollectionDeserializer.java:28) ~[graylog.jar:?]
[...]

PR #14279, opened 2022-12-21
Provide suggestions for numerical fields

todvora:
This PR extends the current prefix query by its scripted alternative for non-textual fields. The benefit is that now we can provide prefix suggestions for numerical fields. The drawback could be lower performance of such solution, but we have to compare the performance to the situation where we deliver no suggestions, only exception.

Additionally I have added a filter for timerange, that we haven't used before. This should limit suggestions even further, compensating the decreased performance of scripting filter for non-text fields.

## Motivation and Context
Fixes  https://github.com/Graylog2/graylog2-server/issues/14214

## How Has This Been Tested?
Added integration test

image

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/SuggestionsResource.java
```diff
@@ -101,14 +112,19 @@ private Set<String> adaptStreams(Set<String> streams, SearchUser searchUser) {
...
 
     private ImmutableSet<String> loadAllAllowedStreamsForUser(SearchUser searchUser) {
         return permittedStreams.load(searchUser);
     }
+
+    private SuggestionFieldType getFieldType(Set<String> streams, TimeRange timerange, final String fieldName) {
+        final Set<MappedFieldTypeDTO> fieldTypes = mappedFieldTypesService.fieldTypesByStreamIds(streams, timerange);
+        return fieldTypes.stream().filter(f -> f.name().equals(fieldName))
+                .map(field -> field.type().type())
+                .findFirst()
```
luk-kaminski, line 126: I am not sure if "findFirst" is a good solution here.
If a field has different types in different indices, mainly if we mix textual and other types, I'd say we should go for OTHER or even new (MIXED?) enum value to trigger scripted variant of the query.

luk-kaminski: You are right.
Maybe MappedFieldTypesServiceImpl should return Map<String, MappedFieldTypeDTO> instead of Set<MappedFieldTypeDTO>, to precisely announce that it prepares only one DTO per field name... But that's a question for different task...

todvora: I think the MappedFieldTypesServiceImpl and its types would deserve some refactoring. And agree that this is a goal for another task.

dennisoelkers (approved):

PR #14284, opened 2022-12-21
Concatenate query strings of queries/search types properly when exporting.

### LINKED ISSUE #14268: Blank CSV dashboard export
[HS 1317645421]

## Overview
When attempting to export a dashboard (or the individual widgets within the dashboard) to CSV format, the resulting export is blank (a zero byte file is produced). The dashboard does show results before attempting the export (so we would expect the export to also include the results).

When using the Replay Search feature (to run the search again from the Search page), the query produces results. 

## Possible Cause

From some additional investigation by @ryan-carroll-graylog:

I was able to reproduce this with the provided dashboard. It looks like when exporting from the dashboard search widget, the query that gets executed on the search backend it getting changed from event_code:4625 OR EventID:4625 to event_code:4625 OR EventID:4625 AND event_code:4625 OR EventID:4625.

I traced the fork in the execution between exporting form the dashboard vs exporting from the search replay to here:

https://github.com/Graylog2/graylog2-server/blob/237fa479a272d8d045461d741285c166de73b258/graylog2-server/src/main/java/org/graylog/plugins/views/search/export/CommandFactory.java#L134

Where it looks like the extra `AND event_code:4625 OR EventID:4625` is being added for the dashboard export, vs here for the search page:

https://github.com/Graylog2/graylog2-server/blob/237fa479a272d8d045461d741285c166de73b258/graylog2-server/src/main/java/org/graylog/plugins/views/search/export/CommandFactory.java#L136

We get a different searchType provided for the different export locations witch looks to be what's driving this but I still don't understand what the SearchTypes mean or are for.

The user has provided a content pack, which can reproduce the issue. 

## Environment

* Graylog Version: 4.3.9
* OpenSearch Version: 2.4.1
* MongoDB Version: 5.0.14

dennisoelkers:
**Note:** This needs to be backported to `4.3` and `5.0`.

## Motivation and Context

This PR is fixing an issue related to exporting a search type. When both the search type and the query contain query strings, they are being concatenated, by simply combining them with an `AND`. For simple query strings this works, but it changes the logic for more complicated ones (e.g. when query string1 is `foo OR bar` and the second is also `foo OR bar`, the resulting query string `foo OR bar AND foo bar` has a different meaning, due to the stronger binding of the logical AND.

With this PR, concatenating two query strings wraps them in braces too, so `foo OR bar` concatenated to itself ends up as `(foo OR bar) AND (foo or BAR)`, which returns the same, correct results.

Fixes #14268.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4502

## Types of changes

## Checklist:


danotorrey (approved):
LGTM and tests successfully with instance of problematic dashboard. Exports are now successful.

@dennisoelkers Looks like there are some test failures in `AggregationEventProcessorTest`. Perhaps those tests need to be updated with the new grouping logic.

ryan-carroll-graylog (approved):
Looks good and tests successfully with Dashboard that was previously causing export issues.

PR #14286, opened 2022-12-21
Escape Event content in HTML email notifications

### LINKED ISSUE #11183: Html escape in JMTE
Some log message may contain chars like `<xxx>` which may broken or dismiss in the html email notification.
How to deal with them?

mpfz0r:
Event notifications can be send as HTML emails.

We didn't escape the event content that was rendered within those emails.

This PR fixes that.
Special thanks to @zhu for providing an initial PR.

Fixes #11183


graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
```diff
@@ -174,7 +176,7 @@ private void bindSingletons() {
         bind(ClusterStatsModule.class).asEagerSingleton();
         bind(ClusterConfigService.class).to(ClusterConfigServiceImpl.class).asEagerSingleton();
         bind(GrokPatternRegistry.class).in(Scopes.SINGLETON);
-        bind(Engine.class).toInstance(Engine.createEngine());
+        bind(Engine.class).annotatedWith(Names.named("HtmlSafe")).toProvider(HtmlSafeJmteEngineProvider.class).asEagerSingleton();
```
patrickmann, line 179: Why not bind the plain Engine as a singleton too? Looks like we create a new instance with each instance of EmailSender.

patrickmann (commented):
Looks good - just have a question about bindings.

graylog2-server/src/main/java/org/graylog2/bindings/ServerBindings.java
mpfz0r: make sense. thank you :+1:

patrickmann (approved):

PR #14298, opened 2022-12-22
Remove GC warnings

### LINKED ISSUE #8592: Nodes with too long GC pauses with G1GC

thll:
Now that we [enforce usage of G1GC](https://github.com/Graylog2/graylog2-server/pull/13929), garbage collection should be much less problematic for a Graylog node. 

Graylog would emit a warning notification when the duration of a GC run exceeded a configured threshold. With G1GC, the actual pause during the GC run should be relatively short, so the overall duration of the GC run is less meaningful. 

This PR removes these notifications.

Fixes #8592


bernd (commented):
Please add the removal of the config options to `UPGRADING.md`. Thanks!

mpfz0r:
> Please add the removal of the config options to `UPGRADING.md`. Thanks!

we should also mention the API change in the exposed configuration response

thll:
> > Please add the removal of the config options to `UPGRADING.md`. Thanks!
> 
> we should also mention the API change in the exposed configuration response

Right, I forgot to mention this as an open question. At first, I only deprecated the option in the exposed configuration and always returned a dummy value of `-1`. But then I thought for this specific API it would be OK to make a breaking change, which is what we now have. @mpfz0r, @bernd, do you agree?

mpfz0r (approved):

bernd:
> > > Please add the removal of the config options to `UPGRADING.md`. Thanks!
> > 
> > 
> > we should also mention the API change in the exposed configuration response
> 
> Right, I forgot to mention this as an open question. At first, I only deprecated the option in the exposed configuration and always returned a dummy value of `-1`. But then I thought for this specific API it would be OK to make a breaking change, which is what we now have. @mpfz0r, @bernd, do you agree?

@thll If we don't backport this to 5.0 or earlier, I am okay with removing the value from the API response.

PR #14304, opened 2022-12-23
Escape Event content in HTML email notifications (#14286) (5.0)

mpfz0r:
* replace Jmte Engine instance with JmteEngineProvider

plugins can add NamedRenderer to engine
add HtmlSafeRenderer to engine by default

* Add EmailJmteEngineProvider, always escape values

* Apply html escape to new event notification emails

The legacy email notification was plaintext only
and does not need to be escaped.

Also simplify escaping a bit and don't escape the email Subject.

* Add EmailSenderTest

* add changelog

* remove unneded multibinder

* revert test notification title change

* keep singleton binding for regular template engine

Co-authored-by: 朱穆穆 <mumu.zhu@feat.com>
(cherry picked from commit 29a4c31682274c252b7a95a355b2361966611471)

### DIFF
--- a/changelog/unreleased/issue-11183.toml
+++ b/changelog/unreleased/issue-11183.toml
@@ -1,0 +1,8 @@
+type = "fixed"
+message = "Properly escape event content for HTML email notifications"
+
+issues = ["11183"]
+pulls = ["14286"]
+
+contributors = ["zhu"]
+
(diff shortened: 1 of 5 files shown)


patrickmann (approved):
Simple backport from 5.1 without any changes

PR #14337, opened 2023-01-02
Add reply-to to email notification config

### LINKED ISSUE #14254: Make Mail Header 'Reply-To' configurable in Mail Notifications
It would be nice if it would be configurable to set an Reply-To Mail Header in Mail Notifications, so that the sender mail can remain an "server e-mail" but answers to the mail alerts will be sent to the professional team which can assist in interpreting the alarm.

## What?
Configuring to set a Reply-To Mail Header in Mail Notification so that replied to the mail alerts are sent to a different email. See more here: [https://en.wikipedia.org/wiki/Email#Message_header](https://en.wikipedia.org/wiki/Email%23Message_header)

## Why?
This is in the event described above where another team needs to notified so that their team can work the alarm instead of it going to a generic email address.

## Your Environment
Issue is related directly to Graylog.

* Graylog Version: 4.3.9
* Elasticsearch Version: n/a
* MongoDB Version: n/a
* Operating System: n/a
* Browser version: n/a

patrickmann:
Resolves #14254 

Adds support for a separate reply-to email address in notifications by extending `EmailEventNotificationConfig`.


graylog2-server/src/main/java/org/graylog/events/contentpack/entities/EmailEventNotificationConfigEntity.java
```diff
@@ -85,6 +89,9 @@ public static Builder create() {
         @JsonProperty(FIELD_SENDER)
         public abstract Builder sender(ValueReference sender);
 
+        @JsonProperty(FIELD_REPLY_TO)
+        public abstract Builder replyTo(ValueReference sender);
```
thll, line 93: Because `replyTo` is not optional, we need to add a default in the `create()` method in line `82` in order to still support the installation of older content packs which don't have the property set.

thll (changes requested):

graylog2-server/src/main/java/org/graylog/events/contentpack/entities/EmailEventNotificationConfigEntity.java
patrickmann: Good catch - added that

graylog2-server/src/main/java/org/graylog/events/notifications/types/EmailEventNotificationConfig.java
```diff
@@ -137,6 +141,7 @@ public static Builder create() {
             return new AutoValue_EmailEventNotificationConfig.Builder()
                     .type(TYPE_NAME)
                     .sender(DEFAULT_SENDER)
+                    .replyTo(DEFAULT_SENDER)
```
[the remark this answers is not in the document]
patrickmann: I went with the a simple empty String default.

thll (approved):
LGTM and works.

thll:
@ousmaneo, this is currently assigned to you. Feel free to merge if you think it's ready.

ousmaneo:
Thank you ! for your review @thll .

PR #14343, opened 2023-01-03
Add global user session timeout configuration

### LINKED ISSUE #11379: Add configurable global user timeout configuration setting
## What?

Currently, the session timeout is configurable for each user on the Users > Edit page in the Sessions Timeout field. 

image

A customer is requesting the ability to edit the timeout globally for all users at once.

## Why?

A configurable global timeout would prevent the work required to edit and maintain the session timeout for individual users.

Ticket reference: HS-579202485

patrickmann:
Resolves #11379 

A new parameter `global_session_timeout_interval` is added to `server.conf`.
When this is present, you can no longer configure a timeout per user; and it overrides any previously configured user-specific timeout value!

Re-enable individual session timeouts by deleting the global setting from `server.conf`.


bernd:
@patrickmann @boosty  Is there a reason that this setting is a `server.conf` setting instead of a global runtime setting in the database?

Server config file settings have the drawback that they need to be in sync on all Graylog nodes and that changing the setting requires a server restart. The global user session timeout is a good candidate for a runtime rather than a server config file setting.

patrickmann:
> The global user session timeout is a good candidate for a runtime rather than a server config file setting

Makes sense - I'll fix that

bernd:
> > The global user session timeout is a good candidate for a runtime rather than a server config file setting
> 
> Makes sense - I'll fix that

Thank you! :pray: 

Please talk with a frontend developer about where users would set the global session timeout. I probably fits somewhere in "System / Users and Teams".

gally47:
Frontend is ready for review. in `System/Configuration` I added a panel for User Configuration to enable the feature and assign a timeout value, in `System/Users and Teams` I made sure the user session timeout is disabled when the global timeout is enabled.

patrickmann:
@gally47 No input fields are visible in the system config form when there is no existing configuration entry in the `cluster_config` DB collection.
I would expect it to always be present - just disabled and with no / default value when it has never been set.

gally47:
> @gally47 No input fields are visible in the system config form when there is no existing configuration entry in the `cluster_config` DB collection. I would expect it to always be present - just disabled and with no / default value when it has never been set.

Reverted the commit because we will have a default value in the backend.

graylog2-web-interface/src/components/configurations/UserConfig.tsx
```diff
@@ -0,0 +1,152 @@
...
+                                         name="enable_global_session_timeout"
+                                         id="enable_global_session_timeout"
+                                         label={(
+                                           <LabelSpan>Enable global session timeout</LabelSpan>
+                                         )} />
+                            <InputDescription help="If enabled, it will be set for all the users." />
+                          </Col>
+                          <Col sm={12}>
+                            <fieldset>
+                              <ISODurationInput id="global_session_timeout_interval"
```
ousmaneo, line 118: The validation is triggered when the field is disabled as you can see below:
image

The input has a default value `0` and when I remove it to enter a value a `P` appear. I think we should fix this with a proper default value.

graylog2-web-interface/src/components/users/TimeoutInput.tsx
```diff
@@ -92,7 +94,8 @@ const TimeoutInput = ({ value: propsValue, onChange }: Props) => {
                help="When checked, sessions never time out due to inactivity."
                formGroupClassName="no-bm"
                onChange={_onClick}
-               checked={sessionTimeoutNever} />
+               checked={sessionTimeoutNever}
+               disabled={isGlobalTimeoutEnabled} />
```
ousmaneo, line 98: I think we should use the `isGlobalTimeoutEnabled` as a condition to display the help message or hide the timeout fields instead of disabling them.

graylog2-web-interface/src/components/users/UserCreate/TimeoutFormGroup.tsx
```diff
@@ -15,17 +15,31 @@
...
+  isGlobalTimeoutEnabled: boolean;
+};
+
+const TimeoutFormGroup = ({ isGlobalTimeoutEnabled }: Props) => (
   <Field name="session_timeout_ms">
     {({ field: { name, value, onChange } }) => (
       <TimeoutInput value={value}
-                    onChange={(newValue) => onChange({ target: { name, value: newValue } })} />
+                    onChange={(newValue) => onChange({ target: { name, value: newValue } })}
+                    isGlobalTimeoutEnabled={isGlobalTimeoutEnabled} />
```
ousmaneo, line 32: This will not be needed when we make the above suggestion.

ousmaneo (changes requested):
I tested it and the feature works ! I made some comments and suggestion about some findings

graylog2-web-interface/src/components/users/UserCreate/UserCreate.tsx
```diff
@@ -187,7 +196,12 @@ const UserCreate = () => {
               </div>
               <div>
                 <Headline>Settings</Headline>
-                <TimeoutFormGroup />
+                {isGlobalTimeoutEnabled ? (
+                  <GlobalTimeoutMessage label="Sessions Timeout"
```
ousmaneo, line 200: The timeout disabled box is not centered vertically compared to the other fields. I think we should fix it. see below 
image

ousmaneo (changes requested):
The frontend looks good! I added a small comment.

graylog2-web-interface/src/components/users/UserCreate/UserCreate.tsx
gally47: yes there was an extra padding top there, should be fixed now.

ousmaneo (changes requested):
LGTM !
I noticed that we display the global timeout on the user overview page. We should also indicate that this is set globally.

user-overview

ousmaneo (approved):
LGTM! Thank you!

PR #14353, opened 2023-01-04
Revert introduction of a separate index mapping for system events

### LINKED ISSUE #14345: Revert introduction of a separate mapping for System Events
#13833 introduced a new mapping for System Events. This required deleting the existing index and stream and creating new instances, referencing the new template. It turns out that the mappings are identical and this is all redundant. 

Preserve the existing System Events index and stream. Remove all redundant code.

patrickmann:
Resolves #14345 

Remove all unneeded code for a separate system event index mapping. System events should use the same mapping that regular events do.


patrickmann:
> We should provide this mongo command before merging it

Will do - thx for providing that

danotorrey:
@simonychuang and I just hit an issue that I think is related:

When starting up the Graylog server in a dev environment, the following error was happening: 
```
2023-01-11 20:00:25,795 ERROR: org.graylog2.bootstrap.ServerBootstrap - Exception while running migrations
java.lang.IllegalStateException: Index prefix conflict: a non-events index-set with prefix <gl-system-events> already exists. Configure a different <default_system_events_index_prefix> value in the server config file.
	at org.graylog2.migrations.V20190705071400_AddEventIndexSetsMigration.checkIndexPrefixConflicts(V20190705071400_AddEventIndexSetsMigration.java:139) ~[classes/:?]
	at org.graylog2.migrations.V20190705071400_AddEventIndexSetsMigration.ensureEventsStreamAndIndexSet(V20190705071400_AddEventIndexSetsMigration.java:120) ~[classes/:?]
```

It turned out that the dev instance had a System Events index set with a `system_events` index_template_type. The issue was resolved by setting the template type to events as @marco mentioned above. 

```
db.index_sets.updateOne({'title':/System Events/}, {$set: {index_template_type: "events"}})
```

I think this might affect anyone who started a new dev database instance between the time https://github.com/Graylog2/graylog2-server/pull/13833 and https://github.com/Graylog2/graylog2-server/pull/14353 were merged. Just wanted to mention this in-case anyone else hits the issue.

PR #14355, opened 2023-01-04
Saved searches can be sorted by 3 additional fields

### LINKED ISSUE #14140: Sort for paginated saved searches API endpoint not working as expected.
## Description

When defining a sort when fetching the paginated saved searches the resulting list does not have the expected sort.
It looks like it only works with 'created_at' but not 'title', 'description', 'summary' and 'owner'.

## Your Environment

* Graylog Version: 5.0

luk-kaminski:
## Description
Saved searches can be sorted by 3 additional fields: owner, description and summary.
Proper indexes prepared in Mongo DB for sortable fields.
Fixes #14140.

## Motivation and Context
See #14140.

## How Has This Been Tested?
Manually and with additional integration test.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
@@ -16,6 +16,7 @@
  */
 package org.graylog.plugins.views.search.views;
 
+import com.mongodb.BasicDBObject;
 import com.mongodb.DuplicateKeyException;
 import org.bson.types.ObjectId;
 import org.graylog.security.entities.EntityOwnershipService;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
```diff
@@ -62,6 +63,11 @@ protected ViewService(MongoConnection mongoConnection,
         this.viewRequirementsFactory = viewRequirementsFactory;
         this.entityOwnerShipService = entityOwnerShipService;
         this.viewSummaryService = viewSummaryService;
+        for (String sortField : ViewDTO.SORT_FIELDS) {
```
todvora, line 66: Would it make sense to run this code inside a migration, only once?

todvora (changes requested):

graylog2-server/src/main/java/org/graylog/plugins/views/search/views/ViewService.java
luk-kaminski: Hmm, I feel it should be run on every start-up.
If the index is there, Mongo won't duplicate it, so there is nothing to worry about.
If it is not there (i.e. smb has dropped it), we recreate it.

todvora: I checked again and we are doing the same in other services. Let's keep it as it is. Thanks!

todvora (approved):

PR #14371, opened 2023-01-05
Separate concerns in `PaginatedList` to simplify state management.

linuspahl:
## Motivation and Context

With this PR we are simplifying the state management in the `PaginatedList` component. The component can maintain the state for the active page and page size in two different ways:
1. It can have its own react state
2. Or It can use the URL query params to maintain the state

Before this PR the component contained all necessary logic for both use cases. Even when the state was based on the URL query params it still maintained its own state. This resulted in a comparable complex state management.

The `PaginatedList` component now contains the stateless `ListBase` component which receives the active page and page size. 
In case 1 we render a wrapper component which maintains a state for the active page and page size and passes it to the `ListBase` component.
In case 2 we render a wrapper component which extracts the page and page size from the URL and passes it to the `ListBase` component

With this approach the change in https://github.com/Graylog2/graylog2-server/pull/14141 is no longer necessary.

One thought I had while working on the state management, in case 1 we currently often maintain two states, one outside of the `PaginatedList` and one which is part of the component. In my opinion we can remove the internal `PaginatedList` state to simplify the sate management even more.

## Types of changes

/nocl

### DIFF
--- a/graylog2-web-interface/src/components/common/PaginatedList.test.tsx
+++ b/graylog2-web-interface/src/components/common/PaginatedList.test.tsx
@@ -33,14 +33,25 @@ jest.mock('react-router-dom', () => ({
 
 describe('PaginatedList', () => {
   it('should display Pagination', () => {
-    const { getByText } = render(<PaginatedList totalItems={100} onChange={() => {}}><div>The list</div></PaginatedList>);
+    const { getByText } = render(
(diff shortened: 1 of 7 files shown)


linuspahl:
@zeeklop Thanks for having the look! The streams sorting issue has been fixed with https://github.com/Graylog2/graylog2-server/pull/14447.

PR #14389, opened 2023-01-09
Update the node_id field on Notification.addNode()

mpfz0r:
Otherwise Notification.getNodeId() is returning nil

Improve SystemNotificationRenderService render exception message.

I stumbled over these while testing with an `An input has failed to start` notification.

Relates to #14399

/nocl


patrickmann (approved):
Looks good - thanks for contributing the fix.

PR #14409, opened 2023-01-10
Reduce logging and add metric for stream warning

danotorrey:
## Description

Implements the following:
1) Reduce logging level for warning from `warn` to `trace`, to not show the warning unless `trace` logging is explicitly enabled:
> Couldn't remove default stream [stream-id] from message [message-id]

2) Add a metric, so that the condition can be monitored, if desired:

### Graylog metric (counter)
`counter org.graylog2.streams.StreamRouterEngine.cannotRemoveDefault`

image

### Prometheus metric
`stream_router_cannot_remove_default`

image

## Motivation and Context

Closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/3481

This warning is logged when a previously executed message processor (or Illuminate) has already removed the default stream from the message (Illuminate does this by default for all messages currently). Once the message exits the first processor, it then enters the Graylog MessageFilterChain and is offered up for matching on all streams. If any stream (including an Illuminate stream) has a stream rule defined that matches the message, then the warning would trigger. This could result in a high volume of server log messages (one per message), which can cause increased disk IO and potentially other contention issues.

## Workaround

One workaround is to uncheck the `Remove matches from 'Default Stream' option for any stream that is matching messages that have completed Illuminate processing. Or, remove the matching stream rule (which could be present on any Illuminate or non-Illuminate stream).

Another workaround is to turn down the logging level in the affected class with this API command:
```
curl -I -X PUT http://<the-api-token>:token@localhost:9000/api/system/loggers/org.graylog2.streams.StreamRouterEngine/level/error \
-H 'X-Requested-By: graylog-api-user' -v 
```

## How Has This Been Tested?

Manual testing to make sure the warning no longer shows unless trace is enabled. Manual testing also verified that the metric is working in Graylog > Nodes > Metrics and in Prometheus when the export is set up. 

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/pr-14409.toml
+++ b/changelog/unreleased/pr-14409.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Reduce logging and add metric for [Couldn't remove default stream] warning."
+
+issues = ["graylog-plugin-enterprise#3481"]
+pulls = ["14409"]
--- a/changelog/unreleased/pr-14409.toml
+++ b/changelog/unreleased/pr-14409.toml
@@ -1,5 +1,5 @@
 type = "fixed"
-message = "Reduce logging and add metric for [Couldn't remove default stream] warning."
+message = "Reduce logging and add metric for \"Couldn't remove default stream\" warning."
 
 issues = ["graylog-plugin-enterprise#3481"]
 pulls = ["14409"]
--- a/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
+++ b/graylog2-server/src/main/java/org/graylog2/streams/StreamRouterEngine.java
@@ -229,12 +229,12 @@ public class StreamRouterEngine {
                         LOG.trace("Successfully removed default stream <{}> from message <{}>", defaultStream.getId(), message.getId());
                     }
                 } else {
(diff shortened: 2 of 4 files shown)


danotorrey:
Due to the potentially severe impact of this issue on production environment, I suggest backporting to `5.0` and `4.3`.

danotorrey:
Due to the potentially severe impact of this issue on production environment, I suggest backporting to 5.0 and 4.3. FYI @rich-graylog @waab76

changelog/unreleased/pr-14409.toml
```diff
@@ -0,0 +1,5 @@
+type = "fixed"
+message = "Reduce logging and add metric for [Couldn't remove default stream] warning."
```
bernd, line 2: Is there are specific reason for the brackets usage instead of quotes?

```suggestion
message = "Reduce logging and add metric for \"Couldn't remove default stream\" warning."
```

bernd (commented):
One comment, apart from that it LGTM. :+1:

ryan-carroll-graylog (approved):
Looks good! Warn log message is cleared up and metric is reporting correctly.

PR #14418, opened 2023-01-11
Fix line break in field select options.

linuspahl:
## Motivation and Context

This PR is fixing a styling problem in the fields select in the aggregations builder.

Before:

image

After:

image

The styling / scrolling behaviour is now similar to 4.3.

/nocl


dennisoelkers (approved):
Looks good now! :pray:

PR #14424, opened 2023-01-12
Time Based Size Optimizing Index Rotation / Retention

mpfz0r:
Implement a new Rotation Strategy: `TimeBasedSizeOptimizing`

The new index rotation strategy aims to make it easier to achieve good resource usage,
i.e. shards that are neither too small nor too big, while avoiding excessive number
of shards. This has been difficult to date, especially when faced with variable ingest
rate.
The Time Based Size Optimizing Rotation Strategy tries to rotate the index daily.'
It can however skip the rotation to achieve optimal sized indices by keeping the shard size between 20 and 50 GB.
The optimization can delay the rotation within the range of the configured retention min/max lifetime.
If an index is older than the range between min/max, it will be rotated regardless of its current size.

New configuration parameters introduced:
`time_size_optimizing_rotation_min_shard_size=20g`: index should be at least this big before rotating.
`time_size_optimizing_rotation_max_shard_size=50g`: rotate if index exceeds this size limit.
`time_size_optimizing_rotation_min_lifetime=30d`: index should be at least this old before rotating.
`time_size_optimizing_rotation_max_lifetime=40d`: rotate if index exceeds this age limit.

`time_size_optimizing_rotation_period`: defaults to `1d` and should not be tweaked for regular use.

Refs https://github.com/Graylog2/graylog-plugin-enterprise/issues/4447


graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
```diff
@@ -95,6 +90,22 @@ private void retainTimeBased(IndexSet indexSet, TimeBasedSizeOptimizingStrategyC
...
 
+    private DateTime getClosingDateBestEffort(String indexName) {
+        if (!indices.isClosed(indexName)) {
+            throw new IllegalStateException(f("Invalid request for closing date - index %s is not closed", indexName));
+        }
+        Optional<DateTime> closingDate = indices.indexClosingDate(indexName);
+        if (closingDate.isPresent()) {
+            return closingDate.get();
+        }
+        final IndexRangeStats indexRangeStats = indices.indexRangeStatsOfIndex(indexName);
```
mpfz0r, line 98: I'm not sure if we should pull in the timestamps from the log data itself.
So far this is strictly based on the time from the rotation and I think we shouldn't mix them.

For an alternative idea: If we don't have the `closingDate` can we use the `creationDate` and compare it against
the `indexLifeTimeHard()` instead?

What we could also do, is to make use of the fact that the creationDate of a newer index is kind of like the closingDate of the current one. e.g

```
index_4   creationDate  2022-12-05-...
index_3   creationDate  2022-12-04-...
index_2   creationDate  2022-12-03-...
```
the closingDate of `index_2` is most likely the creationDate of `index_3`

patrickmann: Agreed - we shouldn't mingle message timestamps with index dates. Checking `creationDate` against `lifetimeHard` is a good idea.

Evaluating dates on other indices is a bit messy, due to having to parse and sort index names. I don't think it's worth adding that complexity, just for the transition period when there are still indices that have no `closingDate`. Instead I propose forcing retention as a fallback, when we cannot even find a `creationDate`.

ousmaneo (approved):

changelog/unreleased/pr-14424.toml
```diff
@@ -0,0 +1,24 @@
...
+of shards. This has been difficult to date, especially when faced with variable ingest
+rate.
+"""
+
+details.ops = """
+`time_size_optimizing` is the new default rotation strategy, as it provides the best results
+with a minimum of user input. The default settings are based on ES/OS sharding recommendations and generally
+do not need to be changed.
+The following configuration file settings are available, if you wish to modify the behavior
+for your requirements:
```
boosty, line 19: I think we should mention the default values here.

changelog/unreleased/pr-14424.toml
```diff
@@ -0,0 +1,24 @@
...
+`time_size_optimizing_rotation_min_shard_size`: index should be at least this big before rotating.
+`time_size_optimizing_rotation_max_shard_size`: rotate if index exceeds this size limit.
+`time_size_optimizing_rotation_min_lifetime`: index should be at least this old before rotating.
+`time_size_optimizing_rotation_max_lifetime`: rotate if index exceeds this age limit.
```
boosty, line 23: I think the two last ones affect the retention, not the rotation, so the current naming can be misleading. 

As we write in some other places: 

> "time-size-optimizing" **tries to rotate daily**, while focussing on optimal sized shards. 

So as a user I would wonder if it still tries to rotate daily when `time_size_optimizing_rotation_min_lifetime` is set to e.g. 30 days.

boosty (commented):
Great work! 

I just added a few comments regarding the config options.

changelog/unreleased/pr-14424.toml
```diff
@@ -0,0 +1,24 @@
...
+rate.
+"""
+
+details.ops = """
+`time_size_optimizing` is the new default rotation strategy, as it provides the best results
+with a minimum of user input. The default settings are based on ES/OS sharding recommendations and generally
+do not need to be changed.
+The following configuration file settings are available, if you wish to modify the behavior
+for your requirements:
+`time_size_optimizing_rotation_min_shard_size`: index should be at least this big before rotating.
```
boosty, line 20: s/index/shards

changelog/unreleased/pr-14424.toml
```diff
@@ -0,0 +1,24 @@
...
+`time_size_optimizing_rotation_max_shard_size`: rotate if index exceeds this size limit.
```
boosty, line 21: s/index/shards

changelog/unreleased/pr-14424.toml
mpfz0r: can do :+1:

changelog/unreleased/pr-14424.toml
mpfz0r: > I think the two last ones affect the retention, not the rotation, so the current naming can be misleading.
yes and no. 
The rotation needs to know the time range for the retention, so it knows for how long it can keep an index open.
While it would be more correct to put this setting into the retention, it is in fact implemented in the rotation strategy, because otherwise we would have to touch each existing retention.

So from a code perspective the rotation name is better.
From a user perspective, maybe not. OTOH, this defines only the default, which again can now be
tweaked in the system/configurations UI. So I think most users won't be using this setting. (cloud will be though)

boosty:
> `time_size_optimizing_rotation_period`: defaults to `1d` and should not be tweaked for regular use.

What happens if a shard gets bigger than `time_size_optimizing_rotation_max_shard_size` in less than a day? Would the strategy notice with a default of `1d`?

mpfz0r:
> What happens if a shard gets bigger than `time_size_optimizing_rotation_max_shard_size` in less than a day? Would the strategy notice with a default of `1d`

It would. The rotation is checked by a periodical every 10 seconds

PR #14446, opened 2023-01-16
Supporting nested and tupled groupings in aggregations.

dennisoelkers:
**Note:** This needs a backport to `5.0`.

## Motivation and Context

For 5.0, we changed the aggregation strategy from strictly nesting aggregations to generating tuples when multiple fields are defined for grouping. This improved results for users wanting to generate top-n/bottom-n tabulations. On the other hand, there are valid use cases for nested aggregations, as seen in e.g.
https://github.com/Graylog2/graylog2-server/pull/13855#issuecomment-1363912347.

Therefore, this PR is now changing back the general strategy for aggregations back to nesting individual groupings, but in addition allows to define multiple fields per grouping, generating n-tuples in results. This allows us to support both classes of use cases and avoids the breaking change for `5.0` that the original change resulted in.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/Constants.ts
```diff
@@ -31,6 +32,12 @@ export const TIMESTAMP_FIELD = 'timestamp';
 export const MESSAGE_FIELD = 'message';
 export const SOURCE_FIELD = 'source';
 export const MISSING_BUCKET_NAME = '(Empty Value)';
+export const DEFAULT_LIMIT = 15;
```
linuspahl, line 35: Maybe let's call this variable `DEFAULT_PIVOT_LIMIT`.

janheise (approved):
LGTM!

maxiadlovskii (approved):
Looks good. One suggestion, we can add a cross icon as a delete icon for fields. Or we can make the current icon smaller. We already have a trash-can icon for deleting the whole group. Having the same icon, almost in the same place, might not be very comfortable. But that is just my personal preference.

PR #14450, opened 2023-01-16
Implement smart index rotation strategy

mpfz0r:
/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
@@ -42,6 +42,8 @@ public class SmartRotationStrategy extends AbstractRotationStrategy {
 
     private final Indices indices;
 
+    // TODO: move this into server.conf or maybe into IndexSetsDefaultConfiguration
+    // also see elasticsearch_max_size_per_index
     public static final Size MAX_INDEX_SIZE = Size.gigabytes(50);
     public static final Size MIN_INDEX_SIZE = Size.gigabytes(20);
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
@@ -16,20 +16,34 @@
  */
 package org.graylog2.indexer.rotation.strategies;
 
+import com.github.joschi.jadconfig.util.Size;
 import org.graylog2.audit.AuditEventSender;
 import org.graylog2.configuration.ElasticsearchConfiguration;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.plugin.indexer.rotation.RotationStrategyConfig;
 import org.graylog2.plugin.system.NodeId;
+import org.graylog2.shared.utilities.StringUtils;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
 import javax.inject.Inject;
+import java.time.Instant;
+import java.time.Period;
+
+import static org.graylog2.shared.utilities.StringUtils.f;
 
 public class SmartRotationStrategy extends AbstractRotationStrategy {
+    private static final Logger LOG = LoggerFactory.getLogger(SmartRotationStrategy.class);
     public static final String NAME = "smart";
+    public static final Period ROTATION_PERIOD = Period.ofDays(1);
 
     private final Indices indices;
+
+    public static final Size MAX_INDEX_SIZE = Size.gigabytes(50);
+    public static final Size MIN_INDEX_SIZE = Size.gigabytes(20);
 
     @Inject
     public SmartRotationStrategy(Indices indices,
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
```diff
@@ -50,14 +66,83 @@ public RotationStrategyConfig defaultConfiguration() {
...
-        return null;
+        final DateTime creationDate = indices.indexCreationDate(index).orElseThrow(()-> new IllegalStateException("No index creation date"));
+        final Long sizeInBytes = indices.getStoreSizeInBytes(index).orElseThrow(() -> new IllegalStateException("No index size"));
+
+        Period leeWay;
+        if (indexSet.getConfig().rotationStrategy() instanceof SmartRotationStrategyConfig rotationConfig) {
+            leeWay = rotationConfig.indexLifetimeHard().minus(rotationConfig.indexLifetimeSoft());
+        } else {
+            throw new IllegalStateException(f("Unsupported RotationStrategyConfig type <%s>", indexSet.getConfig().rotationStrategy()));
+        }
```
patrickmann, line 80: Do we need this check? It shouldn't be possible to create an instance of `SmartRotationStrategy` with a mismatched config.
I do see that other rotation strategy classes have the same check, so we should do it just for consistency.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
```diff
@@ -50,14 +66,83 @@ public RotationStrategyConfig defaultConfiguration() {
...
+            return new Result(true,
+                    f("Index creation date <%s> exceeds optimization leeway <%s>",
+                            creationDate, leeWay));
+        }
+
+        if (indexIsOldEnough(creationDate) && !indexSubceedsSizeLimit(sizeInBytes)) {
+            return new Result(true,
+                    f("Index is old enough (<%s>) and has a reasonable size (<%s>) for rotation",
+                            creationDate, StringUtils.humanReadableByteCount(sizeInBytes)));
+        }
```
patrickmann, line 97: We could optimize by only computing time-based limits when we are on a `ROTATION_PERIOD` boundary.
Not sure if the small amount of time-saving justifies the extra complexity though.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
```diff
@@ -50,14 +66,83 @@ public RotationStrategyConfig defaultConfiguration() {
...
     @Override
     public String getStrategyName() {
         return NAME;
     }
+    static class Result implements AbstractRotationStrategy.Result {
+        private final boolean shouldRotate;
+        private final String message;
+
+        private Result(boolean shouldRotate, String message) {
+            this.shouldRotate = shouldRotate;
```
patrickmann, line 133: This fully generic version of `Result` could be broken out and used across all rotation strategies. It seems unnecessarily complex to have custom implementations in each; which mostly to nothing more than provide string formatting of the message.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
```diff
@@ -50,14 +66,83 @@ public RotationStrategyConfig defaultConfiguration() {
...
+    private boolean indexExceedsLeeWay(DateTime creationDate, Period leeWay) {
+        final Instant now = Instant.now();
+        final Instant rotationPlusLeeWay = now.minus(ROTATION_PERIOD.plus(leeWay));
+
+        return creationDate.isAfter(rotationPlusLeeWay.toEpochMilli());
+    }
+
+    private boolean indexIsOldEnough(DateTime creationDate) {
+        final Instant now = Instant.now();
+        final Instant aDayAgo = now.minus(ROTATION_PERIOD);
```
patrickmann, line 111: Since the period can be a multiple or fraction of a day, the name `aDayAgo` is not quite accurate.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
```diff
@@ -50,14 +66,83 @@ public RotationStrategyConfig defaultConfiguration() {
...
+        }
+
+        return new Result(false, "No reason to rotate found");
+    }
+
+    private boolean indexExceedsLeeWay(DateTime creationDate, Period leeWay) {
+        final Instant now = Instant.now();
+        final Instant rotationPlusLeeWay = now.minus(ROTATION_PERIOD.plus(leeWay));
+
+        return creationDate.isAfter(rotationPlusLeeWay.toEpochMilli());
```
patrickmann, line 106: I think this should be `isBefore`. We want to know if the current index is older than the allowed max.

patrickmann (changes requested):

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
mpfz0r: I think it doesn't hurt to have it. What else would you do in this case? Just return silently and do nothing?

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
mpfz0r: > Do we want to snap to a grid of size ROTATION_PERIOD to ensure predictable times for rotation

yes, I didn't implement that yet. It's less trivial than I thought :-D

Let's postpone this and do it in the next PR.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
mpfz0r: Agreed. I didn't want the change to be too big. Will do that in a separate PR.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategy.java
mpfz0r: ouch. how did I miss that :roll_eyes:

graylog2-server/src/test/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategyTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertThat(result.shouldRotate()).isEqualTo(true);
+        assertThat(result.getDescription()).contains("Index is old enough");
+    }
+
+    @Test
+    void shouldRotateWhenBeyondLeeWay() {
+        final Period leeWay = smartRotationStrategyConfig.indexLifetimeHard().minus(smartRotationStrategyConfig.indexLifetimeSoft());
+        // TODO only works with multiple of days
+        final java.time.Duration leeWayDuration = java.time.Duration.ofDays(leeWay.getDays() + ROTATION_PERIOD.getDays());
+        final DateTime creationDate = now.minus(leeWayDuration.toMillis());
```
patrickmann: `creationDate` is right at the limit. The test fails if there isn't any elapsed time between computing this date and checking the rotation criteria. Make the creationdate a little bit earlier.

patrickmann (changes requested):

graylog2-server/src/test/java/org/graylog2/indexer/rotation/strategies/SmartRotationStrategyTest.java
patrickmann: Already fixed - nice

patrickmann (approved):

PR #14457, opened 2023-01-17
Store the time when an index gets closed (rotated)

mpfz0r:
For the new index rotation and retention strategy feature, we need to know when an index was created and when it was closed.

While the creation date is already available, the "closing" date is not.

A good place to store this information is the _meta field of each index:
 https://www.elastic.co/guide/en/elasticsearch/reference/7.10/mapping-meta-field.html#mapping-meta-field

```
A mapping type can have custom meta data associated with it. These are
not used at all by Elasticsearch, but can be used to store
application-specific metadata.
```


patrickmann (approved):
Looks good - postponed testing until we have integrated with the other changes

mpfz0r:
Frontend tests are failing. Haven't touched any frontend code :shrug:

PR #14462, opened 2023-01-18
Implement async responses and timeouts for proxied requests

bernd:
Before this change, it was easy to saturate the server HTTP thread pool ("http_thread_pool_size") on busy servers with lots of proxied cluster requests. (e.g., cluster metrics) These cluster requests happen periodically when users are using the Graylog UI.

To mitigate these situations, we introduce a timeout for proxied HTTP requests. The default timeout can be configured with the "proxied_requests_default_call_timeout" setting in the server config file. We can override the default timeout in "ProxiedResource" subclasses for specific requests. (e.g. some metric calls can use a lower timeout)

This change also adds the "ProxiedResource#processAsync" method to schedule the request processing in the proxied resource thread pool and free up the regular HTTP thread pool.

The "ClusterMetricsResource" is adjusted to process cluster metric requests asynchronously and with a low timeout.

Additional changes:

- Extend the WARN messages for proxied requests to show the call duration. That should make it easier to debug issues.
- Introduce "#getForAllNodes" and "#requestOnAllNodes" methods in the "ProxiedResource" class that take a timeout duration to override the default timeout.
- Ensure the error body is closed in "ProxiedResource#doNodeApiCall".


thll (approved):
Very nice improvement 👍

mpfz0r:
This introduced a possible session takeover, because the Subject was not cleaned up when the http worker thread
is returned back into the pool. The next request on this thread would then have the session of this previous async request.

Thanks to @dennisoelkers for investigating and to @thll for the final hint :handshake: 

https://github.com/Graylog2/graylog2-server/pull/14904

bernd:
@dennisoelkers @mpfz0r @thll Great job folks, thank you! :pray:

PR #14473, opened 2023-01-19
Add TimeBasedSizeOptimizingRotationAndRetentionTest

mpfz0r:
This test creates indices, forwards the clock,
calls the rotations and retentions and asserts that the list of open indices gets modified accordingly.

Use the JobSchedulerClock which can be used forward the time. Extend the JobSchedulerClock to support java.time Instant and ZonedDateTime.

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
@@ -79,6 +79,8 @@ public abstract class AbstractIndexRetentionStrategy implements RetentionStrateg
                 .filter(indexName -> !hasCurrentWriteAlias(indexSet, deflectorIndices, indexName))
                 .filter(indexName -> {
                     DateTime closingDate = indices.indexClosingDate(indexName)
+                            // TODO we might encounter older indices that don't have a closing date yet. How do we handle this?
+                            // TODO maybe use the creationDate as a fallback?
                             .orElseThrow(() -> new IllegalStateException(f("Index %s has no closing date - retention failed", indexName)));
                     return closingDate.isBefore(cutoff + 1);
                 })
--- a/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
@@ -16,6 +16,7 @@
  */
 package org.graylog2.indexer.retention.strategies;
 
+import org.graylog.scheduler.clock.JobSchedulerClock;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.indices.Indices;
 import org.graylog2.indexer.rotation.strategies.TimeBasedSizeOptimizingStrategyConfig;
@@ -27,7 +28,6 @@ import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -46,11 +46,14 @@ public abstract class AbstractIndexRetentionStrategy implements RetentionStrateg
 
     private final Indices indices;
     private final ActivityWriter activityWriter;
+    private JobSchedulerClock clock;
 
     public AbstractIndexRetentionStrategy(Indices indices,
-                                          ActivityWriter activityWriter) {
+                                          ActivityWriter activityWriter,
+                                          JobSchedulerClock clock) {
         this.indices = requireNonNull(indices);
         this.activityWriter = requireNonNull(activityWriter);
+        this.clock = clock;
     }
 
     protected abstract Optional<Integer> getMaxNumberOfIndices(IndexSet indexSet);
@@ -60,8 +63,7 @@ public abstract class AbstractIndexRetentionStrategy implements RetentionStrateg
     public void retain(IndexSet indexSet) {
         if (indexSet.getConfig().rotationStrategy() instanceof TimeBasedSizeOptimizingStrategyConfig smartConfig) {
             retainTimeBased(indexSet, smartConfig);
-        }
-        else {
+        } else {
             retainCountBased(indexSet);
         }
     }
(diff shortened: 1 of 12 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
```diff
@@ -70,14 +72,15 @@ private void retainTimeBased(IndexSet indexSet, TimeBasedSizeOptimizingStrategyC
...
+        final long cutoff = clock.instantNow().minus(smartConfig.indexLifetimeSoft()).toEpochMilli();
         final int removeCount = (int)deflectorIndices.keySet()
                 .stream()
                 .filter(indexName -> !indices.isReopened(indexName))
+                .filter(indexName -> !hasCurrentWriteAlias(indexSet, deflectorIndices, indexName))
                 .filter(indexName -> {
                     DateTime closingDate = indices.indexClosingDate(indexName)
                             .orElseThrow(() -> new IllegalStateException(f("Index %s has no closing date - retention failed", indexName)));
-                    return closingDate.isBefore(cutoff);
+                    return closingDate.isBefore(cutoff + 1);
```
patrickmann, line 83: 1 ms change - what difference does that make when retention periods are measured in days?

mpfz0r: It make the comparison inclusive instead of exclusive. doesn't matter that much, but it felt nicer while testing it.

mpfz0r:
`SuggestionResourceIT.testSuggestionsAreLimitedToStream()` test failure. has nothing to do with my change. ignoring it :shrug:

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
```diff
@@ -100,21 +105,25 @@ protected Result shouldRotate(final String index, IndexSet indexSet) {
...
 
-        return creationDate.isBefore(rotationLimit.toEpochMilli());
+        final Duration between = Duration.between(Instant.ofEpochMilli(creationDate.getMillis()), now);
+        return between.compareTo(Duration.ofDays(ROTATION_PERIOD.getDays())) >= 0;
+        //return creationDate.isBefore(rotationLimit.toEpochMilli());
     }
 
     private boolean indexExceedsSizeLimit(long size) {
-        return size > MAX_INDEX_SIZE.toBytes();
+        return size >= MAX_INDEX_SIZE.toBytes();
```
patrickmann, line 126: The name `indexExceedsSizeLimit` is a bit misleading, if we allow equality. Also the log messages that rely on this method explicitly say "exceeded".
1 byte up or down is insignificant for what is typically measured in GB. Why is this change needed?

mpfz0r: to accommodate the test. but yeah, I can revert that

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
```diff
@@ -100,21 +105,25 @@ protected Result shouldRotate(final String index, IndexSet indexSet) {
...
 
     private boolean indexExceedsLeeWay(DateTime creationDate, Period leeWay) {
-        final Instant now = Instant.now();
+        final Instant now = clock.instantNow();
         final Instant leewayLimit = now.minus(ROTATION_PERIOD.plus(leeWay));
 
-        return creationDate.isBefore(leewayLimit.toEpochMilli());
+        final Duration between = Duration.between(Instant.ofEpochMilli(creationDate.getMillis()), now);
+        return between.compareTo(Duration.ofDays(ROTATION_PERIOD.plus(leeWay).getDays())) >= 0;
+        //return creationDate.isBefore(leewayLimit.toEpochMilli());
```
patrickmann, line 113: `leewayLimit` is now unused.
Can you add a comment briefly explaining the reasoning? It's hard to understand why date are compared this way. In fact I would break it out into a separate method, given that we use it twice. Then it can have a nice descriptive name.

graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
```diff
@@ -100,21 +105,25 @@ protected Result shouldRotate(final String index, IndexSet indexSet) {
...
 
     private boolean indexIsOldEnough(DateTime creationDate) {
-        final Instant now = Instant.now();
+        final Instant now = clock.instantNow();
         final Instant rotationLimit = now.minus(ROTATION_PERIOD);
 
-        return creationDate.isBefore(rotationLimit.toEpochMilli());
+        final Duration between = Duration.between(Instant.ofEpochMilli(creationDate.getMillis()), now);
+        return between.compareTo(Duration.ofDays(ROTATION_PERIOD.getDays())) >= 0;
+        //return creationDate.isBefore(rotationLimit.toEpochMilli());
```
patrickmann, line 122: `rotationLimit` is now unused

graylog2-server/src/test/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingRotationAndRetentionTest.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    private String getBetween(TestIndex i) {
+        if (i.getClosingDate() == null) {
+            return "null";
+        }
+        return org.joda.time.Period.fieldDifference(clock.nowUTC().toLocalDateTime(), i.getClosingDate().toLocalDateTime()).toString();
+    }
+
+    class TestIndexSet extends org.graylog2.indexer.TestIndexSet {
```
patrickmann, line 218: TestIndexSet and TestIndex could be useful for other tests. Maybe break out those into `indexer.indices.util`?

mpfz0r: hmm, not sure about that. 
it all relies on a lot of mocks (line 107)

patrickmann (approved):

PR #14476, opened 2023-01-20
Bulk removal for streams and views

luk-kaminski:
## Description
Bulk Removal - main implementation.
Shown on the examples of Streams, Views and Event Definitions.

Main concepts:
1. Bulk delete is a POST request, where client provides proper BulkDeleteRequest (with list of entities) in order to create and execute the task of bulk removal.
2. Current implementation is a sequential implementation, which in fact performs in loop removal of each entities. Preparing a bulk removal that works with single request to MongoDB is very hard because of many reasons (permission checking, audit logs etc.)
3. Each resource will have its own endpoint for bulk removal ("/events/definitions/bulk_delete", "/streams/bulk_delete" etc.), but the framework behind bulk removal should be general and easily reusable.
4. Audit log part may sometimes require some customizations (please see StreamResource example), as we have strange, hidden dependency to enterprise classes defining message and attribute format in a text-based form (i.e. GraylogServerEventFormatter.DESCRIPTORS). For most cases, default approach should work fine.
/nocl

## How Has This Been Tested?
For now only with unit tests.
Integration tests will be added later.

Screenshot 2023-01-20 at 11-42-09 Graylog REST API browser

## Types of changes

## Checklist:


janheise (commented):
Works great for me but I am wondering if the functionality should move one level lower into the services and not work on/inside the resource so that other parts of graylog might reuse it? @dennisoelkers @todvora @luk-kaminski what do you think?

luk-kaminski:
> Works great for me but I am wondering if the functionality should move one level lower into the services and not work on/inside the resource so that other parts of graylog might reuse it? @dennisoelkers @todvora @luk-kaminski what do you think?

@janheise I'd say that a lot of our resource classes have a tendency to perform actions that should be performed in the service class. We can definitely meet and discuss if we want to move code between layers, but it is imho a big thing.

moesterheld:
Could the model `Request` and `Response` classes be made more generic as for example `BulkOperationRequest` and `BulkOperationResponse`? 
As a use case, the event definitions could also be enabled and disabled in bulk. I think a reusable request/response could be used for deletion and these cases as well.

janheise:
After discussing it further, I see my questions above settled

luk-kaminski:
> Could the model `Request` and `Response` classes be made more generic as for example `BulkOperationRequest` and `BulkOperationResponse`? As a use case, the event definitions could also be enabled and disabled in bulk. I think a reusable request/response could be used for deletion and these cases as well.

Thanks @moesterheld !
I'll try to go further and modify not only model classes, but also classes responsible for the bulk execution.

janheise (approved):
LGTM!

PR #14477, opened 2023-01-20
Make the rotation period and sizes customizable

mpfz0r:
While the default should be good enough,
add a new config option:

`time_size_optimizing_rotation_period = 1d`
`time_size_optimizing_rotation_min_size = 20gb`
`time_size_optimizing_rotation_max_size = 50gb`

We also convert `java.time.Period` back to `org.joda.time.Period` since it allows smaller durations
which is helpful for testing and needs less conversions

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
@@ -19,6 +19,7 @@ package org.graylog2.configuration;
 import com.github.joschi.jadconfig.Parameter;
 import com.github.joschi.jadconfig.converters.StringListConverter;
 import com.github.joschi.jadconfig.util.Duration;
+import com.github.joschi.jadconfig.util.Size;
 import com.github.joschi.jadconfig.validators.PositiveDurationValidator;
 import com.github.joschi.jadconfig.validators.PositiveIntegerValidator;
 import com.github.joschi.jadconfig.validators.PositiveLongValidator;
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
```diff
@@ -96,6 +99,16 @@ public class ElasticsearchConfiguration {
     @Parameter(value = "elasticsearch_max_number_of_indices", required = true, validator = PositiveIntegerValidator.class)
     private int maxNumberOfIndices = 20;
 
+    // TimeBasedSizeOptimizingStrategy Rotation
+    @Parameter(value = "time_size_optimizing_rotation_period")
+    private Period timeSizeOptimizingRotationPeriod = Period.days(1);
```
patrickmann, line 104: Do we need to make sure rotation period is a multiple of full days, or an integer division of a day? Can anything bad happen if this is changed to e.g. 7 hours?

graylog2-server/src/main/java/org/graylog2/indexer/IndexSetValidator.java
```diff
@@ -104,31 +105,23 @@ private Violation validatePrefix(IndexSetConfig newConfig) {
     @Nullable
     public Violation validateRotation(RotationStrategyConfig rotationStrategyConfig) {
         if ((rotationStrategyConfig instanceof TimeBasedSizeOptimizingStrategyConfig smartConfig)) {
```
patrickmann, line 107: "smartConfig" no longer matches the naming scheme

graylog2-server/src/main/java/org/graylog2/configuration/ElasticsearchConfiguration.java
mpfz0r: No, I'm currently testing it with minutes `PT1M`

patrickmann (approved):
LGTM - I didn't do in-depth testing, since the changes are functionally equivalent.

PR #14499, opened 2023-01-24
Replacing reflux state management in views with Redux.

dennisoelkers:
**Important:** When testing this PR/after merging it, the following steps need to be done:

```
yarn install
yarn build
```
As this PR is bringing in new dependencies which need to be part of the `vendor` bundle.

## Description

This PR is replacing the state management for the views code with Redux/[Redux Toolkit (RTK)](https://github.com/reduxjs/redux-toolkit).  

This massive refactoring was performed in the views code, because it is the only part of our current code which needs complex state handling. All other code using Reflux can be replaced with either `useQuery` or (in really complex cases) with `useReducer`.

The change itself is very big. Some upfront PRs were created to keep it smaller, but this PR contains the core of the refactoring which can not be broken down to smaller pieces due to the interdependencies between the different stores, unfortunately.

When it comes to state usage in the views code, the following best practices should be followed for future feature work:

  - The Redux context is established in the `SearchPage` component and is wrapping only views functionality
  - All stores/actions in the views code which depended upon the `ViewStore` have been removed, including the `ViewStore` itself
  - The store state is separated into multiple slices:
    * The `viewSlice` contains the `View` including the associated `Search`
    * The `searchExecutionSlice` contains the context of the current execution (widget filter, execution state) and the last search result
    * The `searchMetadataSlice` contains the metadata (parsing information, declared/undeclared parameters) as well as the result loading state for the current search
  - If a component needs to consume a part of the state that is related to the currently loaded `View`, it should use `useAppSelector` and a pre-defined selector
  - When no appropriate selector is defined already, it should be defined in one of the `views/logic/slices/*Selectors` modules. Pick the one which is the most accurate for the scope of your selector (i.e. all widget-related selectors should go into `widgetsSelector`)
  - When you were calling an action (`SomethingAction.whatever(...)`) in the past, you are now dispatching it to the central views-reducer (e.g. `dispatch(updateWidgetConfig(...))`). The `dispatch` function is returned by the `useAppDispatch` hook.

## Motivation and Context

The benefit of using Redux instead of Reflux for our use case is:

  - No globals/singletons.
  - A single dispatcher for actions, preventing the need for the state to trickle down:
    * Avoids race conditions of all kinds
    * Improves performance due to eliminating unnecessary rerenders for intermediate states
  - Better testing due to not needing to mock stores/actions
  - No imports with side effects
  - Our version of Reflux is very old and outdated

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4606
/nocl

## How Has This Been Tested?

The biggest issue with this PR is its size. Smaller PRs preparing this refactoring were created before, but replacing the core state management with stores depending on each other requires one big leap forward. While a line-by-line review might take too much time, testing this as thorough as possible and supporting this through E2E-tests covering the core functionality should provide the confidence to merge this early enough in the development cycle for us to be able to fix any uncovered bugs early enough before the next release.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/contexts/WidgetFocusProvider.tsx
```diff
@@ -70,58 +70,81 @@ const _updateQueryParams = (
...
+    } as FocusContextState;
 
     if (!isEqual(focusedWidget, nextFocusedWidget)) {
-      if (focusUriParams.id && !widgets.has(focusUriParams.id)) {
+      if (focusUriParams.id && !widgetIds.includes(focusUriParams.id)) {
         return;
       }
 
       setFocusedWidget(nextFocusedWidget);
       const filter = nextFocusedWidget?.id ? [nextFocusedWidget.id] : null;
```
linuspahl, line 102: Let's change `null` to `undefined`, to avoid a state update?
image

graylog2-web-interface/src/views/logic/slices/searchExecutionSlice.ts
```diff
@@ -0,0 +1,177 @@
...
+  const globalOverride = selectGlobalOverride(getState()) ?? GlobalOverride.empty();
+  const newGlobalOverride = globalOverride.toBuilder().timerange(timerange).build();
+
+  return dispatch(searchExecutionSlice.actions.updateGlobalOverride(newGlobalOverride));
+};
+
+export const setGlobalOverride = (queryString: string, timerange: TimeRange) => (dispatch: AppDispatch, getState: () => RootState) => {
+  const globalOverride = selectGlobalOverride(getState()) ?? GlobalOverride.empty();
+  const newGlobalOverride = globalOverride.toBuilder()
+    .query({ type: 'elasticsearch', query_string: queryString })
```
linuspahl, line 140: Here and in line `125` we can use `createElasticsearchQueryString`.

linuspahl (commented):
It looks like we can remove `views/actions/SearchActions` as well.

linuspahl (approved):

maxiadlovskii (approved):

PR #14503, opened 2023-01-24
Retention improvements

mpfz0r:
After the index rotation, we asynchronously start the
`SetIndexReadOnlyJob`, which will flush the index, set the index read only and
then set the closing Date.

This means that there is a time frame, where we can have rotated indices
that still haven't received their closing Date.

The retention strategy would falsely assume that it is missing and
fallback to using the creation date.

Avoid this by fetching the `IndicesBlockStatus` and use it to filter
only indices that are read only ("index.blocks.write")

Add a test for the backwards compatible retention without a closing date.

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
@@ -19,6 +19,7 @@ package org.graylog2.indexer.retention.strategies;
 import org.graylog.scheduler.clock.JobSchedulerClock;
 import org.graylog2.indexer.IndexSet;
 import org.graylog2.indexer.indices.Indices;
+import org.graylog2.indexer.indices.blocks.IndicesBlockStatus;
 import org.graylog2.indexer.rotation.strategies.TimeBasedSizeOptimizingStrategyConfig;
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
```diff
@@ -92,6 +94,10 @@ private void retainTimeBased(IndexSet indexSet, TimeBasedSizeOptimizingStrategyC
         }
     }
 
+    private boolean indexIsReadOnly(IndicesBlockStatus blockStatus, String index) {
+        return Optional.ofNullable(blockStatus.getIndexBlocks(index)).map(b -> b.contains("index.blocks.write")).orElse(false);
```
patrickmann, line 98: What about `index.blocks.read_only`?

mpfz0r: We don't use that. It would also prevent us from writing the closeDate into the metadata field

https://github.com/Graylog2/graylog2-server/blob/master/graylog-storage-opensearch2/src/main/java/org/graylog/storage/opensearch2/IndicesAdapterOS2.java#L216-L225

patrickmann (approved):

PR #14512, opened 2023-01-25
Refresh index before accessing stats for rotation decision

### LINKED ISSUE #11689: Count based index rotation can be delayed
We configured a count based index rotation (>2000) on an index that is not queried on a regular basis.

While the index is constantly written with new data, it isn't queried.
This leads to a situation where the rotation was delayed by up to 13 hours.

The problem comes from the fact that the index stats API is giving us stale results:
```
2021-11-24 16:03:37,145 DEBUG: org.graylog2.indexer.rotation.strategies.AbstractRotationStrategy - Rotation strategy result: Number of messages in <graylog_129> (117) is lower than the limit (2,000). Not doing anything. 
2021-11-24 16:03:47,150 DEBUG: org.graylog2.indexer.rotation.strategies.AbstractRotationStrategy - Rotation strategy result: Number of messages in <graylog_129> (117) is lower than the limit (2,000). Not doing anything. 
2021-11-24 16:03:57,144 DEBUG: org.graylog2.indexer.rotation.strategies.AbstractRotationStrategy - Rotation strategy result: Number of messages in <graylog_129> (117) is lower than the limit (2,000). Not doing anything. 
2021-11-24 16:04:07,145 DEBUG: org.graylog2.indexer.rotation.strategies.AbstractRotationStrategy - Rotation strategy result: Number of messages in <graylog_129> (117) is lower than the limit (2,000). Not doing anything. 
```
Only after manually issuing a flush:
`curl  -XPOST  http://localhost:9200/graylog_129/_flush`

It will see the actual document count and start rotating:

```
2021-11-24 16:04:47,156 DEBUG: org.graylog2.indexer.rotation.strategies.AbstractRotationStrategy - Rotation strategy result: Number of messages in <graylog_129> (2,975) is higher than the limit (2,000). Pointing deflector to new index now! 
```

This probably comes from this change in 7.0:

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_skipped_background_refresh_on_search_idle_shards
```
Shards belonging to an index that does not have an explicit index.refresh_interval configured
will no longer refresh in the background once the shard becomes "search idle", 
ie the shard hasn’t seen any search traffic for index.search.idle.after seconds (defaults to 30s). 
Searches that access a search idle shard will be "parked" until the next refresh happens.
Indexing requests with wait_for_refresh will also trigger a background refresh.
```

## Solutions
 - Configure `index.refresh_interval` or `index.search.idle.after`
 - Call `_flush` before running a `_stats`query   (this might be too expensive, because the index rotation runs every 10s)

## Versions
* Elasticsearch Version: 7.10
* OpenSearch 2.4.1

mpfz0r:
The new optimization in ES7 onwards prevents us from getting correct message counts on indices:

https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_skipped_background_refresh_on_search_idle_shards

Fixes #11689

### DIFF
--- a/changelog/unreleased/issue-11689.toml
+++ b/changelog/unreleased/issue-11689.toml
@@ -1,0 +1,8 @@
+type = "fixed"
+message = "Fix count based index rotation on idle indices."
+
+issues = ["11689"]
+pulls = ["14512"]
+
+contributors = [""]
+
--- a/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java
+++ b/graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/IndicesAdapterES7.java
@@ -29,6 +29,7 @@ import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.flush.FlushRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.forcemerge.ForceMergeRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.open.OpenIndexRequest;
+import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.refresh.RefreshRequest;
 import org.graylog.shaded.elasticsearch7.org.elasticsearch.action.admin.indices.settings.get.GetSettingsRequest;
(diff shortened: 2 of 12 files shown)


mpfz0r:
> The code looks good, the change makes sense. I have not tested it though, so consider this a semi-approval :)

Thanks. I've noticed through some failing integration tests, that the increased refresh interval of 30s might pose a problem.

While I could lower this to 1s, I think it's safer to just issue an refresh request when needed.

patrickmann (approved):
Tested with count-based index. Looks good.

PR #14541, opened 2023-01-27
Display entity data table bulk actions as dropdown.

linuspahl:
## Motivation and Context

With this PR we are changing the way we display bulk actions for the entity data table.

Before:
image

After:
image

The new layout has a few benefits:
- it scales better, there are cases where we might will have over 10 bulk actions
- it looks cleaner, especially with many bulk actions
- it is easier to style the surrounding layout, because the bulk actions button width will not change a lot
- we unify the UI with similar cases in the application


linuspahl:
Closing in favour of https://github.com/Graylog2/graylog2-server/pull/14767

PR #14550, opened 2023-01-27
Add basic infrastructure to allow tracing instrumentation

bernd:
WIP

### DIFF
--- a/graylog-project-parent/pom.xml
+++ b/graylog-project-parent/pom.xml
@@ -585,6 +585,25 @@
                 <version>${unboundid-ldap.version}</version>
             </dependency>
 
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/system/jobs/SystemJobManager.java
```diff
@@ -106,12 +109,18 @@ public void run() {
         return job.getId();
     }
 
+    @WithSpan
+    protected void executeJob(SystemJob job) {
```
mpfz0r, line 113:
```suggestion
    private void executeJob(SystemJob job) {
```

mpfz0r (approved):

PR #14558, opened 2023-01-29
Optimum index size should be based on shards

mpfz0r:
The optimal index size needs to take into account how many shards have been configured.

Thus we change the size definition from per index to per shard:

`time_size_optimizing_rotation_min_size` -> `time_size_optimizing_rotation_min_shard_size`

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
@@ -86,7 +86,6 @@ public class TimeBasedSizeOptimizingStrategy extends AbstractRotationStrategy {
 
         final int shards = indexSet.getConfig().shards();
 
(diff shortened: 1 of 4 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingStrategy.java
```diff
@@ -76,32 +76,36 @@ protected Result shouldRotate(final String index, IndexSet indexSet) {
...
         }
 
         if (indices.numberOfMessages(index) == 0) {
             return createResult(false, "Index is empty");
         }
 
-        if (indexExceedsSizeLimit(sizeInBytes)) {
+        final int shards = indexSet.getConfig().shards();
+
+        //final Size maxIndexSize = getTotalIndexSizeLimit(shards, maxShardSize);
```
patrickmann, line 89: Obsolete comment

patrickmann (approved):
Looks good - only noticed one old comment that can be deleted

Issue #14293, opened 2022-12-22
Persist organisation state of entity data tables


linuspahl:
## Description

We recently implemented the `EntityDataTable` component for the streams, dashboards and saved searches overview. The component offers sorting and selecting the attributes which should be displayed in the list. This state is currently not being persisted and gets lost after leaving the page. To be improve the user experience, the state should be persisted.

## Your Environment

* Graylog Version: 5.0

PR #14573, opened 2023-01-31
Entity list/user preferences stored in Mongo

luk-kaminski:
## Description
Entity list/user preferences stored in Mongo.
BE part of #14293

## Motivation and Context
Replace temporary in-memory implementation with proper, Mongo-based implementation.
Limit the number of properties that FE really needs to store in a permanent storage.
/nocl

## How Has This Been Tested?
Manually and with additional tests.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/entities/preferences/EntityListPreferencesResource.java
```diff
@@ -63,7 +63,7 @@ public EntityListPreferencesResource(final EntityListPreferencesService entityLi
     @Timed
     @ApiOperation(value = "Create or update user preferences for certain entity list")
     @Consumes(MediaType.APPLICATION_JSON)
-    @NoAuditEvent("Storage details have not yet been decided") //TODO
+    @NoAuditEvent("Audit logs are not stored for entity list preferences") //TODO: make sure it is true
```
todvora, line 66: Should we resolve this TODO before merging?

todvora (approved):
Looks good! Thanks!

graylog2-server/src/main/java/org/graylog2/rest/resources/entities/preferences/EntityListPreferencesResource.java
luk-kaminski: _TLDR - I do not see any big risk, but we can wait till the evening to merge after release, to be on the safe side._

I personally think that we should not store audit log every time a user changes visible columns or page size in any entity list. I'd love to consult it with Kay (he is back in a few days). If he had different opinion, I'll simply add it.
For now nobody uses this endpoints, there is no FE ready. 
It would be nice to have it merged so that FE works on the final solution when developing FE part.
I suggest merging after release and quickly adding audit logging if needed.

PR #14605, opened 2023-02-06
Handle deprecated short zone IDs correctly

patrickmann:
Java has deprecated short time zone IDs like EST (except for GMT and UTC).

We need to explicitly map these to the corresonding offset, to ensure backwards compatibility.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/scheduler/schedule/CronJobSchedule.java
+++ b/graylog2-server/src/main/java/org/graylog/scheduler/schedule/CronJobSchedule.java
@@ -25,7 +25,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.google.auto.value.AutoValue;
-import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import org.graylog.scheduler.JobSchedule;
 import org.graylog.scheduler.clock.JobSchedulerClock;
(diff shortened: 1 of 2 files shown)


moesterheld:
lgtm, however there is another use of timezones might try to set the short ids in org.graylog2.plugin.utilities.date.NaturalDateParser (lines 61 and 66). Have you checked if this could potentially still be used with the short ids from the UI somewhere?

patrickmann:
> org.graylog2.plugin.utilities.date.NaturalDateParser (lines 61 and 66). Have you checked if this could potentially still be used with the short ids from the UI somewhere?

That gets used in a lot of places. I couldn't definitively trace anything back to a UI timezone input. But I'm not 100% sure. And it might still happen via new code in the future. Best to also update this occurrence.

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/NaturalDateParser.java
```diff
@@ -58,7 +58,7 @@ public NaturalDateParser(final String timeZone, final Locale locale) throws Ille
 
         this.locale = locale;
         this.timeZone = TimeZone.getTimeZone(timeZone);
-        this.zoneId = ZoneId.of(timeZone);
+        this.zoneId = ZoneId.of(timeZone, ZoneId.SHORT_IDS);
```
moesterheld, line 61: unfortunately, this alone won't do as the `isValidTimeZone` (line 55) will already reject the short ids

moesterheld (changes requested):

graylog2-server/src/main/java/org/graylog2/plugin/utilities/date/NaturalDateParser.java
patrickmann: Good catch. In that case I think we should not make any change to `NaturalDateParser` at all. Just throw the exception if we encounter a short timezone ID going forward. We should move away from those deprecated IDs anyway. And if we really do need to support them for specific scenarios, we can handle that in the scenario-specific code rather than in a central utility class.

moesterheld (approved):
lgtm

PR #14616, opened 2023-02-07
allow disabling of retention strategies

kodjo-anipah:
## Description

Introduce a new configuration setting `disabled_retention_strategies` that can be used to disable certain
retention strategies.

## Motivation and Context

fixes https://github.com/Graylog2/graylog-plugin-cloud/issues/1081
## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
```diff
@@ -112,6 +117,20 @@ private RetentionStrategyDescription getRetentionStrategyDescription(@ApiParam(n
...
+
+    private void removeDeactivatedStrategiesActions(JsonSchema schema) {
+        Map<String, JsonSchema> properties = ((ObjectSchema) schema).getProperties();
+        JsonSchema indexAction = properties.get("index_action");
+
+        if (Objects.nonNull(indexAction)) {
+            Set<String> actionEnums = ((StringSchema) indexAction).getEnums();
+            Set<String> disabledRetentionStrategies = elasticsearchConfiguration.getDisabledRetentionStrategies();
+            disabledRetentionStrategies.stream().map(s -> s.toUpperCase(Locale.ENGLISH)).toList().forEach(actionEnums::remove);
+        }
```
mpfz0r, line 134: Glad you found a solution :+1: 
However, this only works for hiding it in the frontend. the API would still accept every `index_action`.
I think we need to add a pluggable validation method to the `RetentionStrategy` interface.
And implement the validation in the `ArchiveRetentionStrategy`

graylog2-server/src/main/java/org/graylog2/indexer/retention/RetentionStrategyBindings.java
```diff
@@ -16,16 +16,43 @@
...
-        addRetentionStrategy(ClosingRetentionStrategy.class);
-        addRetentionStrategy(NoopRetentionStrategy.class);
+        List<Class<? extends AbstractIndexCountBasedRetentionStrategy>> retentionStrategies = new LinkedList<>(Arrays.asList(DeletionRetentionStrategy.class, ClosingRetentionStrategy.class, NoopRetentionStrategy.class));
+        Set<String> disabledRetentionStrategies = configuration.getDisabledRetentionStrategies();
+
+        for (String disabledStrategy : disabledRetentionStrategies) {
+            switch (disabledStrategy) {
+                case DeletionRetentionStrategy.NAME -> retentionStrategies.remove(DeletionRetentionStrategy.class);
+                case ClosingRetentionStrategy.NAME -> retentionStrategies.remove((ClosingRetentionStrategy.class));
+                case NoopRetentionStrategy.NAME -> retentionStrategies.remove((NoopRetentionStrategy.class));
```
mpfz0r, line 48: would be nice if we could also disable the archiving retention this way, but I guess that's not worth the effort.
Maybe add a TODO here?

kodjo-anipah: image
It works just that the configuration page will definitly break when there are indexSet already configured to use archiving 
image
@ousmaneo  maybe you can help, so that we change the page that the configuration is no longer available? There is actually a nice notification triggered https://user-images.githubusercontent.com/98284293/218089232-dddc045b-69ef-468c-b623-065e68401a3e.png

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indices/RetentionStrategyResource.java
kodjo-anipah: added validation to the config interface

graylog2-server/src/main/java/org/graylog2/configuration/validators/RetentionStrategyValidator.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.common.collect.Sets;
+import org.graylog2.indexer.retention.strategies.ClosingRetentionStrategy;
+import org.graylog2.indexer.retention.strategies.DeletionRetentionStrategy;
+import org.graylog2.indexer.retention.strategies.NoopRetentionStrategy;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class RetentionStrategyValidator implements Validator<Set<String>> {
+    public static final String ARCHIVE_RETENTION_STRATEGY = "archive";
```
kodjo-anipah, line 30: @mpfz0r  kinda ugly but I was out of ideas

graylog2-server/src/main/java/org/graylog2/indexer/retention/RetentionStrategyBindings.java
mpfz0r: >  It works just that the configuration page will definitly break when there are indexSet already configured to use archiving
 

isn't that true for any of those disabled retentions that are already enabled somewhere? How are we supposed to handle that in the backend? 
We could migrate them to one of the remaining retentions, but it feels a bit hackish.
I'm not sure what's a good approach here. Has @boosty given any input on this yet?

boosty: Prior to rolling this change out to Cloud, we will check if there are unsupported options currently being used. 

I would like to discuss that part in https://github.com/Graylog2/graylog-plugin-cloud/issues/1081.

graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/ClosingRetentionStrategyConfig.java
```diff
@@ -48,4 +51,13 @@ public static ClosingRetentionStrategyConfig create(@JsonProperty("max_number_of
     public static ClosingRetentionStrategyConfig createDefault() {
         return create(DEFAULT_MAX_NUMBER_OF_INDICES);
     }
+
+    @Override
+    public void validate(ElasticsearchConfiguration elasticsearchConfiguration) throws IllegalArgumentException {
```
mpfz0r, line 56: do we need to do this for every retentionconfig? can we even create one if the binding is missing? :thinking:

graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/DeletionRetentionStrategyConfig.java
```diff
@@ -48,4 +51,12 @@ public static DeletionRetentionStrategyConfig create(@JsonProperty("max_number_o
     public static DeletionRetentionStrategyConfig createDefault() {
         return create(DEFAULT_MAX_NUMBER_OF_INDICES);
     }
+
+    @Override
+    public void validate(ElasticsearchConfiguration elasticsearchConfiguration) throws IllegalArgumentException {
```
mpfz0r, line 56: same here

graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/NoopRetentionStrategyConfig.java
```diff
@@ -48,4 +51,12 @@ public static NoopRetentionStrategyConfig create(@JsonProperty("max_number_of_in
     public static NoopRetentionStrategyConfig createDefault() {
         return create(DEFAULT_MAX_NUMBER_OF_INDICES);
     }
+
+    @Override
+    public void validate(ElasticsearchConfiguration elasticsearchConfiguration) throws IllegalArgumentException {
```
mpfz0r, line 56: and here

graylog2-server/src/main/java/org/graylog2/plugin/indexer/retention/RetentionStrategyConfig.java
```diff
@@ -29,4 +30,8 @@ public interface RetentionStrategyConfig {
 
     @JsonProperty(MAX_NUMBER_OF_INDEXES_FIELD)
     int maxNumberOfIndices();
+
+    default void validate(ElasticsearchConfiguration elasticsearchConfiguration) throws IllegalArgumentException {
```
mpfz0r, line 34: I think valditation methods should not be implemented by throwing an exception.
Better return something like 
https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/plugin/rest/ValidationResult.java

graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/ClosingRetentionStrategyConfig.java
mpfz0r: nevermind. let's keep it as is

kodjo-anipah: Just to prevent people from still setting wrong action types via api. Strategy itself wont even get bound  but the configuration will still be valid otherwise

mpfz0r (approved):
LGTM

grotlue (changes requested):
Code looks alright in general, just some small suggestions.

I tested and it seems to work but I think the UI could be a bit better.
This is how it looked for me for the creation and edit form:

Screenshot 2023-03-02 at 16 59 15

My suggestions:
- Show the proper name of the retention strategy instead of `org.....`
- Mark the field as invalid in this case
- For the creation form, if the default value for the retention strategy is disabled take the next one as the default value of the select

grotlue (approved):
Looks good now! The texts are shown properly and the disabled item disappears once it is deselected 👍 

https://user-images.githubusercontent.com/3169354/223382127-3f56f14d-7aeb-437d-925a-3d9e34157680.mp4

PR #14618, opened 2023-02-07
Add test for Mongo WriteConflict Error

mpfz0r:
If multiple nodes are running job schedulers,
they can trigger a warning in MongoDB since version 5.0.

I think this is because of this change, but I'm not 100% sure:
https://github.com/mongodb/mongo/commit/d295b6646fcc815e73ad3085b212ad14c8c6de01

This test runs nextRunnableTrigger() with 10 concurrent Threads and triggers the Warning in MongoDB.

Refs https://github.com/Graylog2/graylog-plugin-enterprise/issues/4664


mpfz0r:
finally fixed by MongoDB no need for this PR anymore

PR #14639, opened 2023-02-09
Fix Migration compareTo and hashCode implementation

### LINKED ISSUE #13622: Some migrations not being run
When Graylog starts up not all migrations are being run. 

Upon further inspection it looks like when the migrations are run in `ServerBootstrap.runMigrations()` the Guava `ImmutableSortedSet.copyOf` function is filtering out one of the migrations from the original set.
https://github.com/Graylog2/graylog2-server/blob/0b9ccf6c11d5893821b797a7d29f32048c7e6af1/graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java#L305

This was noticed because V202206061427_MigrateDefaultSummaryTemplateContentPack was not being run, but it appears that if you switch the order whichever migration was added last won't run.

If `ImmutableSet` is used in place of `ImmutableSortedSet`, all lmigrations run.

## Expected Behavior

All registered migrations will run at startup.

## Current Behavior

Not all registered migrations run at startup.

## Possible Solution

Use Guava `ImmutableSet` instead of `ImmutableSortedSet`.

1. Start up server with breakpoints in loop of ServerBootstrap.runMigrations()
2. Verify that not all migrations in original set are not run

## Context

Summary Templates are not appearing in search results because the migration that loads them into the DB never runs

## Your Environment

Dev environment with latest master. 

* Graylog Version: 5.0 Alpha
* Java Version: 17
* Elasticsearch Version: 
* MongoDB Version:
* Operating System:
* Browser version:

mpfz0r:
The hashCode implementation was always returning the same value, because it took the hash of the static comparator. The compareTo was returning 0 for classes with identical createdAt() times. Make this deterministic by also sorting by class name.

Fixes #13622

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/Migration.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/Migration.java
@@ -22,7 +22,7 @@ import java.util.Comparator;
 import java.util.Objects;
 
 public abstract class Migration implements Comparable<Migration> {
-    private static final Comparator<Migration> COMPARATOR = Comparator.comparingLong(migration -> migration.createdAt().toEpochSecond());
+    private static final Comparator<Migration> COMPARATOR = Comparator.comparing(migration -> migration.createdAt().toString() + migration.getClass().getName());
 
     public abstract ZonedDateTime createdAt();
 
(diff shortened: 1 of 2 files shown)


moesterheld (approved):
looks good to me

PR #14331, opened 2022-12-30
Showing widget placeholders allowing inline widget creation.

dennisoelkers:
## Motivation and Context

This PR is adding a feature that adds widget placeholders in empty slots on the widget grid, allowing the user to create widgets in these spots. This removes the need for the user to create a widget and then move it to the desired spot.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4510

## Screenshots (if appropriate):
inline-widget-creation

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

PR #14644, opened 2023-02-10
Make `WidgetPosition` singleton, implement `equals`.

### LINKED ISSUE #14643: Initial moving/resizing of widgets on dashboards does not have an effect.
## Expected Behavior

When creating a new dashboard, moving/resizing a widget for the first time after creating it should be reflected in the widget's position/dimensions afterwards.

## Current Behavior

When creating a new dashboard, adding a widget and resizing/moving it, it will snap back to its original dimensions/position afterwards when doing it the first time. This only happens with the enterprise plugin installed.

1. Create new dashboard
2. Add e.g. message count widget
3. Move or resize it - no effect
4. Move or resize it again - this time it will work

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

This PR is adding `equals` to the `WidgetPosition` class and makes it effectively a singleton (importing it from plugins gets you the same class and not a new version of the class). This makes equality checks for invididual widget positions as well as the widget position map deterministic, eliminating subtle bugs like e.g. #14643. As this was introduced by #14331 which is yet unreleased, no changelog entry and no backports are needed.

The relationship to #14643 is that without this PR, on dashboards widget positions might be created by code in one of the plugins. This causes those widget positions to be based on a different class (as the same module is included in two different bundles), leading to equality checks to return `false`. The result of this is that upon the initial movement, even if the widget positions are structurally equal, the equality check returning `false` leads to an additional update from `onSyncLayout` with stale positions (the previous ones), so the change is immediately overwritten. After this change, the equality checks works as expected and returns `true`, preventing the additional update with stale data.

Fixes #14643.

/nocl

## Types of changes

## Checklist:


linuspahl (approved):
Looks good to me, fixes the described bug.

PR #14648, opened 2023-02-10
Refactor: Migrate ConfigurationForm to typescript and functional component

grotlue:
This PR migrates the `ConfigurationForm` component to a functional component and typescript.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4814

/nocl

## Description
- Migrates `ConfigurationForm`  to typescript and convert it into a functional component
  - Did not change a lot of existing functionality, only what was really necessary for converting it
- Adjusted all components using the `ConfigurationForm` and calling its methods
- The `DecoratorSummary` component also uses it but no adjustments were necessary here, it should be tested though
 
## Motivation and Context
`ConfigurationForm` is still using old code (e.g. class component) and is not typed. This makes adjusting it tough, as types would help a lot in working with a generic component like this. 

## How Has This Been Tested?
I have checked the InputForm, but the other parts using this also need to be tested. I could use some help with that.

## Types of changes

## Checklist:


graylog2-web-interface/src/components/configurationforms/ConfigurationForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+type Props = {
+  cancelAction: () => void,
+  configFields?: {
+    [key: string]: ConfigurationField,
+  },
+  children: React.ReactNode | null,
+  helpBlock: React.ReactNode | null,
+  includeTitleField: boolean,
+  submitAction: (data: any) => void,
```
linuspahl, line 35: I think in this case we can define a generic type which extends an object and which can be defined when using the component. This way we could type `data` and  the `values` prop and  avoid using `any`.

`type Props<T extends object>`
...
`<ConfigurationForm<{...}> `

graylog2-web-interface/src/components/configurationforms/ConfigurationForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+  title: string | React.ReactNode | null,
```
linuspahl, line 36: Regarding the optional props, instead of 
```
children: React.ReactNode | null,
helpBlock: React.ReactNode | null,
title: string | React.ReactNode | null,
```
can't we just say write it like this?
```
children?: React.ReactNode,
helpBlock?: React.ReactNode,
title?: React.ReactNode,
```

linuspahl (changes requested):
Found a problem when creating an input and added some suggestions, but in general it looks good to me!

grotlue:
The typescript errors are being fixed in https://github.com/Graylog2/graylog-plugin-enterprise/pull/4814. Running both PRs locally there should be no errors anymore.

ousmaneo (approved):
Tested it and works !

PR #14649, opened 2023-02-10
Ensure users are unique

### LINKED ISSUE #12963: Add unique index for users.username
## What?

Add a unique index for the `users.username` field. 

## Why?

At least two instances have occurred when customers were upgrading to Graylog `4.3.2` where duplicate users (with the same username) were found and resulted in failed migrations.

For example for the `graylog-sidecar` user:
```
2022-06-03T10:22:18.961-05:00 ERROR [UserServiceImpl] There was more than one matching user for username graylog-sidecar. This should never happen.
2022-06-03T10:22:18.961-05:00 WARN  [ServerBootstrap] Exception while running migrations
java.lang.RuntimeException: There was more than one matching user for username graylog-sidecar. This should never happen.
        at org.graylog2.users.UserServiceImpl.load(UserServiceImpl.java:175) ~[graylog.jar:?]
        at org.graylog2.migrations.MigrationHelpers.ensureUser(MigrationHelpers.java:98) ~[graylog.jar:?]
        at org.graylog.plugins.sidecar.migrations.V20180323150000_AddSidecarUser.upgrade(V20180323150000_AddSidecarUser.java:60) ~[graylog.jar:?]
        at org.graylog2.bootstrap.ServerBootstrap.lambda$runMigrations$0(ServerBootstrap.java:263) ~[graylog.jar:?]
        at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:422) ~[graylog.jar:?]
        at com.google.common.collect.RegularImmutableSortedSet.forEach(RegularImmutableSortedSet.java:8
```

This also happened for another user during a `4.2.9`-`4.3.2` upgrade for the Forwarder system user, which is added in this migration:

https://github.com/Graylog2/graylog-plugin-enterprise/blob/5ea3908413513f3a344d936a0d86780572b7776e/enterprise/src/main/java/org/graylog/plugins/forwarder/migrations/V20200918000000_AddForwarderUser.java#L40-L47

## Your Environment

* Graylog Version: 4.3.2
* Elasticsearch Version:
* MongoDB Version: 4.2.20
* Operating System:
* Browser version:

patrickmann:
Resolves #12963 

We are not able to recover if a duplicate user has been inadvertently added. It is unclear how this happens; but it results in GL failing to start due to migration errors. In versions prior to 4.3 we didn't notice, because GL would start in spite of failing migrations. 

Handle this more robustly:
- rename all but one of the duplicate users
- add a unique index to block future creation of duplicate users


changelog/unreleased/issue-12963.toml
```diff
@@ -0,0 +1,4 @@
+type = "a"
+message = "Adds a unique `username` index to the `user` collection to prevent creation of duplicate users."
```
mpfz0r, line 2: I think we should mention that we also rename duplicate users

mpfz0r (approved):

PR #14659, opened 2023-02-13
prevent creation of loops with `clone_message()` calls

### LINKED ISSUE #5759: clone_message creates endless loop

moesterheld:
## Description
Functions can now retrieve the rule in which they are called from the `evaluationContext`.
The `clone_message()` function uses this to check if it was invoked more than once by the same rule for the same message.

An additional boolean parameter `preventLoops` is added to the function.
If this is set to `true`, `clone_message` will not clone a message more than once.
To provide backward compatibility, this parameter is set to `false` by default. However, if it is not explicitly set to `false`, `clone_message` will stop creating clones after the 100th clone and log an error. Existing edge cases where the same message was intentionally cloned more than 100 times therefore would have to set the parameter explicitly to `false`. 

## Motivation and Context
fixes #5759

## How Has This Been Tested?
see `MessageCreationLoopPreventionTest` for tests of the different cases.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change) -- only for hypothetical cases

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/messages/CloneMessage.java
```diff
@@ -26,26 +28,56 @@
...
+                return null;
+            }
+
+        }
 
         final Message clonedMessage = new Message(currentMessage.getMessage(), currentMessage.getSource(), currentMessage.getTimestamp());
         clonedMessage.addFields(currentMessage.getFields());
         clonedMessage.addStreams(currentMessage.getStreams());
+        clonedMessage.addField(CLONE_SOURCE, rule);
+        clonedMessage.addField(CLONE_NUMBER, ++cloneNumber);
```
bernd, line 80: How about using the metadata fields in `Message` instead of setting actual fields in the message?

https://github.com/Graylog2/graylog2-server/blob/ed3a4da30364e786df619a3067c1854346752791/graylog2-server/src/main/java/org/graylog2/plugin/Message.java#L1011-L1064

Otherwise, those fields would end up being indexed in OpenSearch. Or do I miss the removal of the fields somewhere?

moesterheld: thanks for the comment. I'll change that.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/messages/CloneMessage.java
```diff
@@ -56,7 +90,7 @@ public Message evaluate(FunctionArgs args, EvaluationContext context) {
     public FunctionDescriptor<Message> descriptor() {
         return FunctionDescriptor.<Message>builder()
                 .name(NAME)
-                .params(ImmutableList.of(messageParam))
+                .params(ImmutableList.of(loopDetectionParam, messageParam))
```
thll, line 93:
```suggestion
                .params(ImmutableList.of(messageParam, loopDetectionParam))
```

I think we need to add the new parameter at the end. Otherwise, existing usages of the rule will break with an error like this:

```
2023-02-22 11:43:13,724 WARN : org.graylog.plugins.pipelineprocessor.processors.ConfigurationStateUpdater - Ignoring non parseable rule <clone/63f5ec78b3a3156cd887fe6c> with errors <[Expected type Boolean for argument preventLoops but found Message in call to function clone_message in line 5 pos 4]>

File: graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/messages/CloneMessage.java [hunk 3]
```diff
@@ -33,19 +38,48 @@ public class CloneMessage extends AbstractFunction<Message> {
...
+        if (ObjectUtils.allNotNull(cloneSource, rule) && Objects.equals(cloneSource, rule)) { // message was cloned by the same rule before
+
+            if (preventLoops.orElse(false)) {
+                return null;
+            }
+
+            if (preventLoops.isEmpty() && cloneNumber >= MAX_CLONES) {
+                LOG.error("Message was cloned more than {} times by rule '{}'. Not allowing any more clones to prevent a potential endless loop. If this was intentional, please explicitly set the 'preventLoops' parameter to 'false'.",
+                        MAX_CLONES, cloneSource.name());
+                return null;
```
thll, line 71: As we consider this an error, I think it should show up in our "Indexing and Processing Errors". I think we can achieve this with something along those lines:

```suggestion
               throw new IllegalStateException(
                        StringUtils.f("Message was cloned more than %d times by rule '%s'. Not allowing any more " +
                                        "clones to prevent a potential endless loop. If this was intentional, please " +
                                        "explicitly set the 'preventLoops' parameter to 'false'.",
                                MAX_CLONES, cloneSource.name()));
```

thll (changes requested):
Looks good in general and works well, but I think we could improve the error handling a little in case we forcefully break the loop. Sorry that I didn't notice this before.

thll (approved):
LGTM

Issue #13852, opened 2022-11-01
Implement filters for streams table.


linuspahl:
_Please note, this needs to be implemented after https://github.com/Graylog2/graylog2-server/issues/13848_

## Description

To make it easier to find a specific stream we should offer some filters for the streams table. For example: Author, Capability, Created At, related index set, status (paused / running), etc.

We should implement it in a way that it will be easy to implement it for similar cases in the future.

PR #14690, opened 2023-02-16
Implement basic filters for streams overview

linuspahl:
## Motivation and Context

With this PR we are creating the foundation to offer filters for entity data tables. As an example we implemented the option to filter the streams table by status.

The next steps are: 
- Implement filtering by index set. The filter is already supported by the API, but we want to create a generic backend service which offers the suggestions (index sets) for the UI. 
- Implement filtering by creation date. We need to allow defining a range (like from `2021-01-01` to `2021-01-02`). The API does not support that currently.
- persist the active filters and search query in the URL, like we do it for `perPage`.

Even with the future filter types the current process will not change
- when creating a filter users will open the create dropdown, select an attribute like index set and see the list of available options. After selecting an option we will close the dropdown and display the new filter.
- when changing a filter, users still can click on the filter value. While we currently toggle the value on click for boolean attributes, we can show the same dialog on click for the index set and creation date, which we display for the initial filter creation. After the selection we can close the dropdown / dialog and update the filter.

filter example

Fixes https://github.com/Graylog2/graylog2-server/issues/13852

## Types of changes

## Checklist:


dennisoelkers (approved):
Looks good, works fine!

Issue #14555, opened 2023-01-27
Excessive WARN logging of pipeline rule output


damianharouff:
We have recently noted that excessive WARN logging can negatively impact the customer experience, ref: https://github.com/Graylog2/graylog-plugin-enterprise/issues/3481

Today we observed WARN output for a pipeline rule that was emitting approximately ~400 messages per second, and for quite some time:
![image](https://user-images.githubusercontent.com/2953243/215161631-4e1e4e10-8d98-4784-8503-b48676a71c33.png)

This was specifically noted in a Graylog Cloud installation, where customer does not otherwise have visibility this is happening.

In terms of pipeline rule outcome adding information to the message, this does not seem to have negatively impacted customer's data, so this message was more or less emitting into the void without consequence otherwise.

HS-1395765433

boosty:
This specific log message was added in https://github.com/Graylog2/graylog2-server/commit/c1104df9a42d81bfdde9b1a33ec869385551d256.

@kroepke @bernd Do you think it's worth keeping it at the WARN level? 

Due to the high volume of log lines this can cause, maybe it would be better to reduce the log level, and instead improve the hints during rule editing?

bernd:
@boosty @kroepke If the log message causes problems, we should either change the log level for the class in the config file or use a different log level in the code. Just keep in mind that this will hide pipeline rule issues.

I don't think we can improve the editor here because we cannot predict what a function call returns at runtime. Using `when lookup_value("table", $message.field) == "yolo"` is perfectly fine when editing the rule. But at runtime the `lookup_value` function might return `null`.

boosty:
> If the log message causes problems, we should either change the log level for the class in the config file or use a different log level in the code. Just keep in mind that this will hide pipeline rule issues.

@mpfz0r brought up the idea of introducing a rate limit / throttling for such log messages. I think this could be a reasonable solution for this particular issue, but also for other processing related messages that could appear in large amounts. So we would not hide these warnings completely, but reduce them.

> I don't think we can improve the editor here because we cannot predict what a function call returns at runtime. 

Right, but I was more thinking in ways of guiding the user with hints like "this function could return null, here is a code example on how to test for this". 

Can be a lot of work to get this right though, and should therefore be tracked as a separate issue.

boosty:
Making these warnings visible in Cloud should also be a separate issue.

patrickmann:
Resolved by #14705 and #14709

PR #14705, opened 2023-02-17
Show rule name when logging execution errors

patrickmann:
When pipeline rules log warnings or errors, it is hard to map log entries to rules.
Add current rule to the evaluation context of the pipeline processor; and include the rule name when logging errors.

See #14555 for details.

/nocl


kodjo-anipah (approved):
LGTM
`````
2023-02-22 09:06:46,995 WARN : org.graylog.plugins.pipelineprocessor.ast.expressions.EqualityExpression - Rule <random> left expression evaluated to null, returning false: lookup_set_value(key: "key", lookup_table: "lookup_table", value: "value")

PR #14707, opened 2023-02-17
Fixing up handling of title/metadata changes for views.

dennisoelkers:
## Motivation and Context

This PR is fixing up two things:

1. The "Edit Metadata" action now does not require the current view to be saved before.
   It changes only the view in memory and does not try to update the
   version in the database, making it more consistent with other actions as well.

2. The save functionality is now keeping its own version of the title,
   propagating it back when the user clicks save or save as. This reduces the 
   amount of state held in the `SearchActionsMenu` component.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchForm.tsx
+++ b/graylog2-web-interface/src/views/components/searchbar/saved-search/SavedSearchForm.tsx
@@ -14,7 +14,8 @@
  * along with this program. If not, see
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
-import React from 'react';
+import * as React from 'react';
+import { useCallback, useState } from 'react';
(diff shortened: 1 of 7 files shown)


linuspahl (changes requested):
Two problems I noticed

1. It looks like it is no longer possible to save a search

image

2. It is now possible to add an unsaved view to "favorites", after editing the meta data.

image

image

Some additional ideas:
1. We could also communicate in the breadcrumb title, that the search is still unsaved.

image

In this case we can also display the breadcrumb title for unsaved and untitled searches like this:

image

Currently it is not italic:
image

2. We can now always display the edit icon next to the breadcrumb title, also for an unsaved and untitled search.

image

linuspahl (approved):
Looks good to me, fixes the described problems.

PR #14708, opened 2023-02-17
Integrate page size select in entity data table.

linuspahl:
## Motivation and Context

With this PR we are changing the way we display the page size select for entity data tables.

Before:
image

After:
image

We are now grouping the elements which allow configuring the table layout. Changing the element of the page size select from an input to a button also effects all other places where we use the page size select component. We could create a new component for the entity data table, but I see a benefit in unifying the styles.

/nocl

### DIFF
--- a/graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
+++ b/graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
@@ -20,7 +20,7 @@ import { useMemo, useState, useCallback, useRef } from 'react';
 import type * as Immutable from 'immutable';
 import { merge } from 'lodash';
 
-import { Button, Table, ButtonToolbar } from 'components/bootstrap';
+import { Button, Table, ButtonGroup, ButtonToolbar } from 'components/bootstrap';
 import { isPermitted, isAnyPermitted } from 'util/PermissionsMixin';
 import useCurrentUser from 'hooks/useCurrentUser';
 import StringUtils from 'util/StringUtils';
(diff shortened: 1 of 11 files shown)


graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
```diff
@@ -127,6 +134,10 @@ type Props<Entity extends EntityBase> = {
   onColumnsChange: (columnIds: Array<string>) => void,
   /** Function to handle sort changes */
   onSortChange: (newSort: Sort) => void,
+  /** Function to handle page size changes */
+  onPageSizeChange?: (newPageSize: number) => void,
+  /** Active page size */
+  pageSize?: number
```
linuspahl, line 140: Instead of implementing the page size select in the entity data table I thought about providing the whole select component as a prop, but we would have to do this every time we use the entity data table and we would have to repeat ourselfs.

graylog2-web-interface/src/components/common/PageSizeSelect.tsx
```diff
@@ -18,61 +18,66 @@ import * as React from 'react';
...
+    return (
+      <Container className={className}>
+        Show
+        {select}
+      </Container>
+    );
+  }
+
+  return select;
+};
```
linuspahl, line 67: Maybe it is cleaner to create a separate component to render the `PageSizeSelect` with a label?

graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
dennisoelkers: I would assume that pagination is a central responsibility for the entity data table, so presenting (optional!) props for this is totally fine.

PR #14709, opened 2023-02-17
Use rate-limited logging in pipeline processor

patrickmann:
The pipeline processor may generate an excessive number of log messages since it runs on every message.
Introduce rate-limited logging to mitigate this.
See #14555 for details.

/nocl


kodjo-anipah (commented):
Its working but other logs would still end up polluting the logs I guess we should extend the scope 
image

patrickmann:
>I guess we should extend the scope

I considered that, but it's a pretty open-ended thing. Maybe we do it as needed?

kodjo-anipah:
> > I guess we should extend the scope
> 
> I considered that, but it's a pretty open-ended thing. Maybe we do it as needed?

Makes sense, can you resolve the conflict?

kodjo-anipah (approved):
LGTM

PR #14727, opened 2023-02-20
Rename a few time size optimizing rotation/retention settings

mpfz0r:
And adopt the new default mentioned in UPGRADING.md

### DIFF
--- a/changelog/unreleased/pr-14424.toml
+++ b/changelog/unreleased/pr-14424.toml
@@ -17,8 +17,8 @@ with a minimum of user input. The default settings are based on ES/OS sharding r
 do not need to be changed.
 The following configuration file settings are available, if you wish to modify the behavior
 for your requirements:
-`time_size_optimizing_rotation_min_shard_size=20g`: shard should be at least this big before rotating.
-`time_size_optimizing_rotation_max_shard_size=50g`: rotate if shard exceeds this size limit.
-`time_size_optimizing_retention_min_lifetime=30d`: index should be at least this old before rotating.
-`time_size_optimizing_retention_max_lifetime=40d`: rotate if index exceeds this age limit.
+`time_size_optimizing_rotation_min_shard_size=20g`: Shard should be at least this big before rotating.
+`time_size_optimizing_rotation_max_shard_size=50g`: Rotate if shard exceeds this size limit.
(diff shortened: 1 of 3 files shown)


changelog/unreleased/pr-14424.toml
```diff
@@ -17,8 +17,8 @@ with a minimum of user input. The default settings are based on ES/OS sharding r
...
 The following configuration file settings are available, if you wish to modify the behavior
 for your requirements:
-`time_size_optimizing_rotation_min_shard_size`: index should be at least this big before rotating.
-`time_size_optimizing_rotation_max_shard_size`: rotate if index exceeds this size limit.
-`time_size_optimizing_rotation_min_lifetime`: index should be at least this old before rotating.
-`time_size_optimizing_rotation_max_lifetime`: rotate if index exceeds this age limit.
+`time_size_optimizing_rotation_min_shard_size=20g`: shard should be at least this big before rotating.
+`time_size_optimizing_rotation_max_shard_size=50g`: rotate if shard exceeds this size limit.
+`time_size_optimizing_retention_min_lifetime=30d`: index should be at least this old before rotating.
+`time_size_optimizing_retention_max_lifetime=40d`: rotate if index exceeds this age limit.
```
boosty, line 23: The descriptions do not match the new config option name (retention vs. rotation).

Would the following suggestion describe the behaviour correctly?

```suggestion
`time_size_optimizing_retention_min_lifetime=30d`: minimum index age for retention configuration to be applied.
`time_size_optimizing_retention_max_lifetime=40d`: maximum index age for retention configuration to be applied.
```

boosty (changes requested):
Hi @mpfz0r, thanks for the PR!

Upgrade notes look good 👍 

I left a comment regarding the lifetime descriptions.

After thinking about it again, I would actually be ok with keeping the original option names (also helps to keep it in sync with the code), as long as the description is very clear.

I hope my comment with a suggestion for the description makes sense. If not let please let me know.

changelog/unreleased/pr-14424.toml
boosty: I think the description in the UI is actually very good:

Screenshot 2023-02-21 at 10 09 42

Maybe we should just use the same wording here.

boosty (approved):
Thanks a lot!

danotorrey (approved):
LGTM 👍 thanks @mpfz0r!

PR #14759, opened 2023-02-22
New server startup: Use correct Time Size Optimizing rotation strategy for Default index set

### LINKED ISSUE #14758: Default index set is created with incorrect default retention strategy
## Current Behavior

When starting a new Graylog server instance (without any pre-existing Mongo or Opensearch instance), the Default index set is created with an incorrect rotation strategy. The `Message Count` strategy is used instead of the correct new `Index Time Size Optimizing` value as was introduced in https://github.com/Graylog2/graylog2-server/pull/14424.

Index Sets page from a freshly started Graylog instance:
image
 
## Possible Solution

The source of the issue seems to be a very old migration that established the initial in-database index settings (previously were defined in the server.conf file only).

https://github.com/Graylog2/graylog2-server/blob/4284bd41fe32ffc88b54001426d2834e33ab9894/graylog2-server/src/main/java/org/graylog2/migrations/V20151210140600_ElasticsearchConfigMigration.java#L119

## Context

Discovered this when verifying which defaults need to be updated in the docs as part of recent index sets work (#13018).

## Your Environment

* Graylog Version: 5.1.0-SNAPSHOT
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

danotorrey:
## Description

In #14758, I noticed that the `Default index set` is not using the correct new default `Time Size Optimizing` rotation strategy for newly started Graylog server instances. This PR fixes it.

The creation of the `Default index set` is done in some very legacy migrations. In those early days, index set settings were stored `server.conf` settings, then later, they were migrated to cluster config settings. Finally, they were migrated to the `IndexSetConfig` DTO objects that back the current Index Set page implementation. In order to fix this issue, I had to adjust one of those early migrations and add awareness for this new default. Perhaps at some point, we could remove those migrations, but for the scope of resolving this issue, I thought the approach taken was best. Feedback is welcome.

No need for a change log, since the new Time Size Optimizing change log is already sufficient (see https://github.com/Graylog2/graylog2-server/pull/14424).
/nocl

## Motivation and Context

Fixes #14758. 

## How Has This Been Tested?

Starting a brand new Graylog instance with a blank a Mongo and Opensearch. Then, view the Index Sets page and verified that the correct Time Size Optimizing rotation strategy is used.

image

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/migrations/MaintenanceStrategiesHelper.java
+++ b/graylog2-server/src/main/java/org/graylog2/migrations/MaintenanceStrategiesHelper.java
@@ -30,31 +30,22 @@ import org.graylog2.indexer.rotation.strategies.TimeBasedRotationStrategy;
 import org.graylog2.indexer.rotation.strategies.TimeBasedRotationStrategyConfig;
 import org.graylog2.indexer.rotation.strategies.TimeBasedSizeOptimizingStrategy;
 import org.graylog2.indexer.rotation.strategies.TimeBasedSizeOptimizingStrategyConfig;
-import org.graylog2.plugin.cluster.ClusterConfigService;
-import org.graylog2.plugin.indexer.retention.RetentionStrategy;
 import org.graylog2.plugin.indexer.retention.RetentionStrategyConfig;
-import org.graylog2.plugin.indexer.rotation.RotationStrategy;
 import org.graylog2.plugin.indexer.rotation.RotationStrategyConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
-import javax.inject.Provider;
-import java.util.Map;
 
+/**
+ * Provides ability to read rotation and retention strategies from the Graylog server configuration file.
+ */
 public class MaintenanceStrategiesHelper {
     private static final Logger LOG = LoggerFactory.getLogger(MaintenanceStrategiesHelper.class);
-    private final Map<String, Provider<RotationStrategy>> rotationStrategies;
-    private final Map<String, Provider<RetentionStrategy>> retentionStrategies;
-    private final ClusterConfigService clusterConfigService;
     private final ElasticsearchConfiguration elasticsearchConfiguration;
 
-
     @Inject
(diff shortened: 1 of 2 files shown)


danotorrey:
Looks like I had an incorrect base when creating this PR. I will force-push up a rebase shortly.

graylog2-server/src/main/java/org/graylog2/migrations/MaintenanceStrategiesHelper.java
```diff
@@ -78,16 +69,13 @@ public ImmutablePair<String, RotationStrategyConfig> readRotationConfigFromServe
...
             }
             default -> {
-                LOG.warn("Unknown retention strategy [{}]. Defaulting to [{}]",
-                        elasticsearchConfiguration.getRotationStrategy(), MessageCountRotationStrategy.NAME);
-                return ImmutablePair.of(MessageCountRotationStrategy.class.getCanonicalName(),
-                        MessageCountRotationStrategyConfig.create(elasticsearchConfiguration.getMaxDocsPerIndex()));
+                LOG.warn("Unknown rotation strategy [{}]. Defaulting to [{}].",
+                        elasticsearchConfiguration.getRotationStrategy(), TimeBasedSizeOptimizingStrategy.NAME);
+                return ImmutablePair.of(TimeBasedSizeOptimizingStrategy.class.getCanonicalName(),
+                        buildDefaultTimeSizeStrategy());
```
danotorrey, line 78: This changes the fall-back for unknown/invalid rotation strategies in the `server.conf` to use the new Time Size Optimizing strategy.

ryan-carroll-graylog (approved):
Looks good and tests successfully with migration as well as handling invalid server.conf values. Agree that this is the best approach for resolving this for the scope of the issue.

mpfz0r:
Ah, thanks @danotorrey you beat me to it.
I've started working on this before I got ill.
My approach was to fix it like this,
 https://github.com/Graylog2/graylog2-server/pull/14778
 
 I agree that we should merge all those old legacy migrations into a new one. This is insane :-D

danotorrey:
Sorry about that @mpfz0r! I like your approach better (since it uses that `createDefault` method and is a bit more direct). If you are happy with it, we can go with your solution. Just let me know, and I can test that out on Monday.

danotorrey:
Closed in favor of https://github.com/Graylog2/graylog2-server/pull/14778.

PR #14761, opened 2023-02-23
Report unused disable directives when linting.

dennisoelkers:
## Motivation and Context

This change is making use of the `--report-unused-disable-directives` option for ESLint, which reports unused disable directives (and removes them when run with `--fix`).

This avoids stale disable directives when the rule set is changing and cleans them up automatically.

/nocl

## Types of changes

## Checklist:


linuspahl:
Do you have an idea, why it is not always removing the complete line for the removed disable directives?

It looks like this seems to result in new linter hints in some cases:
- https://github.com/Graylog2/graylog2-server/pull/14761/files#diff-f4a323d4fe3c9a248e338932bf38884684c87bb12dbd31e5b9b376458bae3ed2R210
Also related:
- https://github.com/Graylog2/graylog2-server/pull/14761/files#diff-f52b7e2a5537bdc0d4e3162d0996e49880523b00fecd16a3b5ede008e798062dR120

dennisoelkers:
@linuspahl:

> Do you have an idea, why it is not always removing the complete line for the removed disable directives?

> It looks like this seems to result in new linter hints in some cases

I noticed that too, it is probably a limitation of the AST-based approach. In general, I do expect that this will be improved over time through ESLint.

linuspahl (approved):
Looks good to me!

PR #14775, opened 2023-02-24
Prevent unneeded search execution when opening new search.

dennisoelkers:
## Motivation and Context

This change is preventing one unneeded search execution when starting a new search. It is caused by the widget focus provider which compares the current focus state (which is initially `undefined`) with the next one when syncing with query parameters. At this point, the next one can also be `{ editing: false, focusing: false, id: undefined }`, what is semantically identical, but not equal to `undefined`, causing a state change.

Therefore, this change is now making sure that if the current focus state is `undefined`, the initial object `{ editing: false, focusing: false, id: undefined }` will be used for the equality check instead.

/nocl

## Types of changes

## Checklist:


linuspahl (approved):
Works as expected!

PR #14776, opened 2023-02-24
Actively wait for stream router refresh when creating streams in integration tests

todvora:
Stream creation is an async operation. Newly created stream and its rules have to be processed by the stream router first, which happens in a separated thread. Only after this refresh, messages are routed correctly to streams.

In our tests we were creating streams and sending messages into them right after the creation. Depending on delays and concurrency, this could lead to unpredictable results of message routing. 

Now we are actively waiting for the refresh of the stream router engine (by monitoring its fingerprint), when creating a new stream. The creation operation returns the stream ID only then, when the engine is really changed. Because this waiting and verification happens already during the stream creation, all our existing integration tests are now resilient to this problem.    

/nocl
)

## Checklist:


graylog2-server/src/main/java/org/graylog2/streams/RouterEngineInfo.java
```diff
@@ -0,0 +1,20 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog2.streams;
+
+public record RouterEngineInfo(String fingerprint) {
```
luk-kaminski, line 19: If this engine info contains only one string field, why to introduce a new class/record for it?

graylog2-server/src/test/java/org/graylog/testing/completebackend/apis/Streams.java
```diff
@@ -17,17 +17,25 @@
...
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.notNullValue;
 
 public final class Streams implements GraylogRestApi {
```
luk-kaminski, line 37: It might be a stupid suggestion, as I have not thought about it deeply, but is there a way to provide StreamRouter as a dependency inside this class, instead of adding a new endpoint to the REST resource, which will be visible to anyone, but useful only in tests?

todvora: It's a good point and I am currently rewriting that part. It will get its own "debug" resource, hidden behind an environment variable. The resource is needed, because in the integration test, the API is the only way how to access the running graylog instance.

luk-kaminski (approved):

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
```diff
@@ -74,6 +74,7 @@
 import org.graylog2.streams.PaginatedStreamService;
 import org.graylog2.streams.StreamDTO;
 import org.graylog2.streams.StreamImpl;
+import org.graylog2.streams.StreamRouter;
```
dennisoelkers, line 77: I think this is a remainder?

dennisoelkers (changes requested):
Looks great! Just a tiny remark on an unused import, left over from the previous approach.

graylog2-server/src/main/java/org/graylog2/rest/resources/streams/StreamResource.java
todvora: Removed, thanks!

dennisoelkers (approved):

PR #14778, opened 2023-02-24
Apply new index set defaults to default index set

mpfz0r:
The newly introduced index set defaults with #13018 were not applied to the default index set.

/nocl

Fixes https://github.com/Graylog2/graylog2-server/issues/14758


danotorrey (approved):
LGTM and tests successfully.

PR #14782, opened 2023-02-24
Fix page size handling for dashboards and streams overview.

linuspahl:
## Motivation and Context

Before this change there was a problem with the pagination on the streams and dashboards overview.
To reproduce the problem you need:
- 15 entities
- an initial page size of 10, when you open the page
- change the page size to 50
- change the page size to 10
 
Now you will no longer see the pagination. We mainly make sure `usePaginationQueryParameter` is returning the correct page size and we also no longer sync the page size with the URL since it is saved as a user setting.

Longterm we should change the way we maintain the currently selected page and page size.

## Types of changes

/nocl


dennisoelkers (approved):
Looks good, works fine!

PR #14791, opened 2023-02-27
Fix parsing date math expressions in time range queries

todvora:
Our query validation failed for queries containing [date math expressions](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#date-math):

image

## Description

To fix this issue, several smaller fixes are needed. 

First problem is the lucene query parser and its handling of range queries. The parser creates Automatons and uses them in the visitor pattern. Automation is a state machine for possible valid inputs. But its implementation is not working for our range queries, because the automation doesn't understand the range content and creates wrong states. To fix this problem, I have added a BypassAutomationRangeQuery. This inherits all TermRangeQuery functionality but overrides the visit method to ignore the automation when visited by our visitor. 

Then our TermCollectingQueryVisitor can access the start/end tokens and add them to the parsing process. 

Finally, the FieldTypeValidation is has to be able to recognize date math expressions and validate them properly. The expressions may be validated by simple `[+-/](\d+)?[yMwdhHms]` regex with some additional logic for the initial time anchor (either `now` or a valid date). 
 
Given all the changes, the query validation is now fully supporting date math expressions and should prevent warnings like in the screenshot above.

## How Has This Been Tested?
Added unit and integration tests for all changes.

## Types of changes

## Checklist:


luk-kaminski (approved):
Looks good!

PR #14515, opened 2023-01-25
Server changes to support Azure multi-node capability

danotorrey:
## Description

1) Suppress `INFO`-level Azure SDK log messages.
2) Other misc changes needed to support:
https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/980
/nocl

## Motivation and Context

While implementing https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/980, it was discovered that when using the Azure SDK, many varied SDK-specific `INFO`-level log messages appear, which are generally not helpful for Graylog users using the Azure Event Logs input (Graylog input log messages are enough to confirm successful operation and error conditions).

For example:
```
2023-01-25 11:40:57,030 INFO : com.azure.messaging.eventhubs.EventHubClientBuilder - connectionId[MF_95c194_1674668457016]: Emitting a single connection.
2023-01-25 11:40:57,043 INFO : com.azure.messaging.eventhubs.EventHubClientBuilder - connectionId[MF_ee24b9_1674668457043]: Emitting a single connection.
2023-01-25 11:40:57,202 INFO : com.azure.messaging.eventhubs.implementation.EventHubConnectionProcessor - namespace[secdev.servicebus.windows.net] entityPath[test]: Setting next AMQP channel.
2023-01-25 11:40:57,202 INFO : com.azure.messaging.eventhubs.implementation.EventHubConnectionProcessor - namespace[secdev.servicebus.windows.net] entityPath[test]: Setting next AMQP channel.
2023-01-25 11:40:57,205 INFO : com.azure.messaging.eventhubs.implementation.EventHubConnectionProcessor - namespace[secdev.servicebus.windows.net] entityPath[test]: Next AMQP channel received, updating 0 current subscribers
2023-01-25 11:40:57,205 INFO : com.azure.messaging.eventhubs.implementation.EventHubConnectionProcessor - namespace[secdev.servicebus.windows.net] entityPath[test]: Next AMQP channel received, updating 0 current subscribers
```

This PR suppresses all `com.azure` log messages be default in Graylog. For advanced troubleshooting cases when the Azure SDK info messages are needed again, they can be re-enabled at any time in Graylog by removing this line from the `log4j2.xml` file or by executing the following command to restore the default logging level of `info` for the package. 

```
curl -I -X PUT http://<the-api-token>:token@localhost:9000/api/system/loggers/com.azure/level/info \
-H 'X-Requested-By: graylog-api-user' -v 
```

See corresponding PRs for docker images and OS packages:
[...]

PR #14796, opened 2023-02-27
Mention new InputService#allByType method in UPGRADING.md

danotorrey:
A new method `InputService#allByType(String type)` was added in https://github.com/Graylog2/graylog2-server/pull/14515. This PR updates the `UPGRADING.md` document accordingly, so that users will know they need to update classes that implement `InputService`.

/nocl


danotorrey:
See backport in https://github.com/Graylog2/graylog2-server/pull/14797

thll:
@danotorrey what do you think about adding a default implementation instead, as in https://github.com/Graylog2/graylog2-server/pull/14804?

We wouldn't have to add release notes, and more importantly, we wouldn't introduce a breaking change in a bugfix release.

danotorrey:
Sounds like a great approach, thank you for filing the PR!

danotorrey:
See comment in https://github.com/Graylog2/graylog2-server/pull/14797#issuecomment-1448151197. This PR is no loner relevant, since a backwards-compatible solution was implemented in https://github.com/Graylog2/graylog2-server/pull/14804 and backported in https://github.com/Graylog2/graylog2-server/pull/14814.

PR #14797, opened 2023-02-27
Mention new InputService#allByType method in UPGRADING.md (5.0)

danotorrey:
A new method `InputService#allByType(String type)` was added in https://github.com/Graylog2/graylog2-server/pull/14671 for `5.0`. This PR updates the `UPGRADING.md` document accordingly, so that users will know they need to update classes that implement `InputService`.

/nocl

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -107,10 +107,11 @@ The following Java Code API deprecations have been made in 5.0.
 
 The following Java Code API changes have been made in 5.0.
 
-| File                                                                                                   | Description                                              |
-|--------------------------------------------------------------------------------------------------------|----------------------------------------------------------|
-| `PaginatedPipelineService.java` | Concrete implementation has been changed to an interface |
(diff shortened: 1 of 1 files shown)


ryan-carroll-graylog (approved):

bernd (changes requested):
This is a breaking API change for inputs. We cannot do that. Please find a way to make this backwards compatible. Thank you!

See @thll's suggested solution: https://github.com/Graylog2/graylog2-server/pull/14804/files#diff-d44d87c881b73e71b3693072a6f3d798bb76ad65c0bc3658837f8035c465198b

danotorrey:
Thanks again for flagging this @bernd. We definitely want to make breaking API changes in bug fix releases. I'll use extra care on this in the future. 

We've merged the fix to `master` and a `5.0` backport is in review now https://github.com/Graylog2/graylog2-server/pull/14814 and will be merged today. I've verified that the Forwarder can be built successfully with this (and submodule versions at `5.0`). 

I'm closing this PR now since it's not relevant anymore.

PR #14804, opened 2023-02-28
Support encrypted values in new method InputService#allByType

thll:
A new method `InputService#allByType` has been added in https://github.com/Graylog2/graylog2-server/pull/14515 after implementing support for encrypted values has been started in #14459. Thus, the new method was not covered in  #14459 and now doesn't return encrypted values as expected.

This PR fixes this.

Also, I added a default implementation to the `InputSerfvice` interface so that other implementations don't break.

/nocl


danotorrey:
@thll This looks like a great approach. I will review shortly this morning.

danotorrey (approved):
LGTM and tests successfully. Thank you @thll!

AntonEbel (approved):
LGTM.

PR #14837, opened 2023-03-01
Improve column rendering in entity data table (handle camel case attributes, allow defining column render for attribute types)

linuspahl:
## Motivation and Context

This PR is improving the rendering of columns in the entity data table. We are now
- Handing entities with camel case attributes correctly. The column meta data is defined in snake case, but some entities like saved searches and dashboards have camel case attributes. We are now transforming the camel case attributes to snake case to fix the mapping.
- It is now possible to define a column renderer for attributes types. Columns have a type like `DATE`, since we often want to render e.g. date times the same way, it makes sense to define the column renderers per type. Otherwise we would have to define the same logic for each date attribute.

## Types of changes

## Checklist:

/nocl

### DIFF
--- a/graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
+++ b/graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
@@ -34,7 +34,7 @@ import { PageSizeSelect } from 'components/common';
 import BulkActionsRow from './BulkActionsRow';
 import TableHead from './TableHead';
 import TableRow from './TableRow';
-import type { ColumnRenderers, Column, EntityBase } from './types';
+import type { ColumnRenderers, Column, EntityBase, ColumnRenderersByAttribute } from './types';
 
 const ScrollContainer = styled.div`
   overflow-x: auto;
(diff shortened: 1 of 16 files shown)


graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
```diff
@@ -98,12 +98,13 @@ const filterVisibleColumns = (
...
   columns,
-  columnRenderers,
+  columnRenderersByAttribute,
   displayBulkSelectCol,
 }: {
   columns: Array<Column>,
-  columnRenderers: ColumnRenderers<Entity>,
+  columnRenderersByAttribute: ColumnRenderersByAttribute<Entity>,
   displayBulkSelectCol: boolean
+  entityAttributesAreCamelCase: boolean
```
maxiadlovskii, line 107: The name of this prop sounds like our attributes already in camel case. But then, if this attr is `true`, we are converting columns to camel case. Maybe make sense to rename it to smth like `shouldEntityAttributesBeCamelCase` or `setEntityAttributesToCamelCase`

And, just another idea, to avoid props drilling, we can create a kind of `<EntityDataTable />` Context and pick props like `entityAttributesAreCamelCase` in, f.e. `<TableCell />`, from the Context. But that is up to you, currently hard to say would we need a lot of props from  `<EntityDataTable />` in other parts of the table

linuspahl: `The name of this prop sounds like our attributes already in camel case` - When the attribute is true we actually expect, that the entity attributes are camel case. The logic behind it is, that the column / attribute ids in the column definitions / metadata are always snake case. If `entityAttributesAreCamelCase` is true we transform the column ids to camel case to be able to make a connection with the entity attributes.

linuspahl: I like the idea with the contexts.

linuspahl: I improved the props description for `entityAttributesAreCamelCase`. I hope it makes it more clear.

maxiadlovskii (approved):

PR #14841, opened 2023-03-02
Revert "Apply new index set defaults to default index set (#14778)"

mpfz0r:
The new field type polling interval breaks integration tests. Revert that change until we have that fixed.

Thanks @dennisoelkers for finding this!

This reverts commit 5d41c2a38f0ede7957bb8d564b05251e0fc81094.

/nocl


dennisoelkers:
Not required anymore, superseded by #14842.

PR #14847, opened 2023-03-02
Fix and improve the @HideOnCloud annotation

mpfz0r:
@HideOnCloud only used to work on the class level.

implement it as a ContainerRequestFilter, which is easier and more robust

Filter out the in-memory log for cloud hosting.

/nocl


mpfz0r:
maven is having issues since yesterday thus the failing github build :shrug:

thll (approved):
LGTM

PR #14853, opened 2023-03-03
Add grok pattern validation and unique index

### LINKED ISSUE #11342: New Grok Pattern errors with no relevant information
## Expected Behavior

The actual error is that the title cannot have spaces in it, but this is not communicated to the User in any way in the UI.

Highlight the problem field
Provide an error message with the relevant information
Communicate the solution more effectively 

## Current Behavior

Passing `{"id":"","pattern":"%{BASE10NUM}","name":"THIS IS NAME"}` to the API returns

```{"type":"ValidationApiError","message":"Validation failed!","validation_errors":{"_":[{"error":"Invalid pattern GrokPattern{id=null, name=THIS IS NAME, pattern=%{BASE10NUM}, contentPack=null}\nIllegal repetition near index 0\n%{THIS IS NAME}\n^"}]}}```

image

## Possible Solution

Provide better error message

1. Navigate to `/system/grokpatterns`
2. Click `Create pattern` button
3. Set a title to anything with spaces in it
4. Create a valid pattern
5. Click submit

## Context

Trying to accomplish Testing Day tasks and discovered this error.

### PR DESCRIPTION
**Pull request by patrickmann**
Resolves #11342

- Improve error message when attempting to create a pattern with invalid name.
- Add unique index. This was already prepared in 2018 and we have been enforcing it in code, so it should be safe to do so now.
- Refactor duplicate code

### CODE REVIEW

File: graylog2-server/src/main/java/org/graylog2/grok/MongoDbGrokPatternService.java [hunk 1]
```diff
@@ -50,11 +47,10 @@
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static org.graylog2.grok.GrokPatternService.ImportStrategy.ABORT_ON_CONFLICT;
 import static org.graylog2.grok.GrokPatternService.ImportStrategy.DROP_ALL_EXISTING;
```
moesterheld, line 52: I think this could be made a singleton to avoid creating the index every time the class is constructed (i.e. during pagination)

```suggestion
@Singleton
```

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternServiceImpl.java
```diff
@@ -0,0 +1,80 @@
...
+import io.krakens.grok.api.exception.GrokException;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+abstract class GrokPatternServiceImpl implements GrokPatternService {
```
moesterheld, line 31: maybe out of scope, but `saveAll` validation also behaves differently in the two implementing classes. Perhaps this could also be moved here?

moesterheld (commented):
Works as expected, just two small comments. 

Also, some tests could have been implemented.

graylog2-server/src/main/java/org/graylog2/grok/MongoDbGrokPatternService.java
patrickmann: Mongo doesn't recreate an existing index. But it's still a good idea to avoid an unnecessary roundtrip.

graylog2-server/src/main/java/org/graylog2/grok/GrokPatternServiceImpl.java
patrickmann: `validateAll` is already moved to the parent. Which part of `saveAll` do you mean?

moesterheld: The additional validation of the collections in lines 163-183 in `MongoDbGrokPatternService` and lines 128-135 `InMemoryGrokPatternService` behave differently, but should be consistent in my eyes. Everything else except the actual persisting of the patterns is identical and could be moved to the abstract class.

moesterheld (approved):
lgtm

PR #14866, opened 2023-03-06
Introduce flag to mark inputs as not compatible with forwarder

thll:
Adds a boolean `isForwarderCompatible` method to mark inputs as not being compatible with Forwarders.

Prerequisite for https://github.com/Graylog2/graylog-plugin-enterprise/pull/4818

/nocl


AntonEbel (approved):
LGTM

Issue #11634, opened 2021-11-15
Graylog 4.2 + Java 11 = WARNING: sun.reflect.Reflection.getCallerClass is not supported.


chi-miel:
## Expected Behavior
Unsupported classes should not be used.

## Current Behavior
Running Graylog 4.2 with Java 11 generates following Warning

```
WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by retrofit2.Platform (file:/usr/share/graylog-server/graylog.jar) to constr>
WARNING: Please consider reporting this to the maintainers of retrofit2.Platform
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

## Steps to Reproduce (for bugs)
1. Upgrade Graylog to 4.2 with Java 11
2. Start service
3. Check service status

## Your Environment

* Graylog Version: 4.2
* Java Version: openjdk version "11.0.13" 2021-10-19 LTS
* Operating System: CentOS Linux release 8.4.2105

chi-miel:
Seem same problem : 
https://github.com/Graylog2/graylog2-server/issues/9802
https://github.com/Graylog2/graylog2-server/issues/7223

bernd:
@chi-miel This is just a warning message and shouldn't cause any problems. In this case this is related to the retrofit library. The warning will disappear when the library fixes the reflective access.

maktas:
The warning message is still being shown in Graylog 5 with default java version in that package:
openjdk 17.0.6

Aug 24 11:30:03 *** systemd[1]: Started Graylog server.
Aug 24 11:30:04 *** graylog-server[11733]: WARNING: sun.reflect.Reflection.getCallerClass is not supported. >

Issue #9802, opened 2020-12-15
Graylog 4.0 + Java 11 = WARNING: sun.reflect.Reflection.getCallerClass is not supported.


hulkk:
## Expected Behavior

Unsupported classes should not be used.

## Current Behavior

Running Graylog 4.0 with Java 11 generates following Warning

`WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.`

## Steps to Reproduce (for bugs)

1. Install Graylog 4.0.1 with Java 11
2. Start service
3. Check service status

## Your Environment

* Graylog Version: 4.0.1
* Java Version: openjdk version "11.0.9" 2020-10-20 LTS

Seb35:
This is the same warning as in #7223 (just different Graylog version).

shivamtiwari18:
Using Graylog 5  + Java 11 = WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
generate same error.

How to reslove this issue.

loveyang2012:
> Using Graylog 5 + Java 11 = WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance. generate same error.
> 
> How to reslove this issue.

hahah

bernd:
> Using Graylog 5 + Java 11 = WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance. generate same error.
> 
> How to reslove this issue.

We are working on a fix here: https://github.com/Graylog2/graylog2-server/pull/14884

Issue #7223, opened 2020-01-19
WARNING: An illegal reflective access operation has occurred


alias454:
When using OpenJDK 11 on CentOS 7, I see warnings about illegal reflective access operations that show up after startup.

Output from systemctl -l status graylog-server:
```
● graylog-server.service - Graylog server
   Loaded: loaded (/usr/lib/systemd/system/graylog-server.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2020-01-15 00:35:06 CST; 3 days ago
     Docs: http://docs.graylog.org/
 Main PID: 17354 (graylog-server)
   CGroup: /system.slice/graylog-server.service
           ├─17354 /bin/sh /usr/share/graylog-server/bin/graylog-server
           └─17356 /usr/bin/java -Xms2g -Xmx2g -XX:NewRatio=1 -server -XX:+ResizeTLAB -XX:+CMSConcurrentMTEnabled -XX:+CMSClassUnloadingEnabled -XX:-OmitStackTraceInFastThrow -jar -Dlog4j.configurationFile=file:///etc/graylog/server/log4j2.xml -Djava.library.path=/usr/share/graylog-server/lib/sigar -Dgraylog2.installation_source=rpm /usr/share/graylog-server/graylog.jar server -f /etc/graylog/server/server.conf -np

Jan 15 00:35:06 graylog systemd[1]: Started Graylog server.
Jan 15 00:35:07 graylog graylog-server[17354]: WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.
Jan 15 00:35:12 graylog graylog-server[17354]: WARNING: An illegal reflective access operation has occurred
Jan 15 00:35:12 graylog graylog-server[17354]: WARNING: Illegal reflective access by com.google.inject.assistedinject.FactoryProvider2$MethodHandleWrapper (file:/usr/share/graylog-server/graylog.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
Jan 15 00:35:12 graylog graylog-server[17354]: WARNING: Please consider reporting this to the maintainers of com.google.inject.assistedinject.FactoryProvider2$MethodHandleWrapper
Jan 15 00:35:12 graylog graylog-server[17354]: WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
Jan 15 00:35:12 graylog graylog-server[17354]: WARNING: All illegal access operations will be denied in a future release
```

## Context
[...]

PR #14884, opened 2023-03-08
Fix sun.reflect.Reflection.getCallerClass warning with Java >= 9

bernd:
Enable the Multi-Release flag for the ManifestResourceTransformer in the shade plugin. Without this, the JVM cannot take advantage of the class files for newer JVM versions and hence logs the warning because slower compatibility clases will be used at runtime.

Fixes the following warning on server startup:

    WARNING: sun.reflect.Reflection.getCallerClass is not supported.
    This will impact performance.

See: https://openjdk.org/jeps/238

Refs #7223
Refs #9802
Refs #11634

### DIFF
--- a/changelog/unreleased/pr-14884.toml
+++ b/changelog/unreleased/pr-14884.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Fix `WARNING: sun.reflect.Reflection.getCallerClass is not supported.` on server startup."
+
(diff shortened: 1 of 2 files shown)


mpfz0r (approved):
Works for me. 
I also compared the jar files. the new one has 
```
Multi-Release: true
```
in `META-INF/MANIFEST.MF`  :+1:

PR #14890, opened 2023-03-09
Limit search execution to current (dashboard) page.

### LINKED ISSUE #6867: All view searches executed upon change
## Expected Behavior

A change to a query/tab, or introducing a new tab (duplicate or `+`) should only initiate searches/queries required to fulfill what is being displayed within that tab.

## Current Behavior

Any change initiating a view search (e.g. add or duplicate a tab, change a widget's configuration, change the search query or time range) invokes `/api/views/search/ID_HERE/execute`, which results in all tabs' searches being executed.  When dealing with relatively large searches a trivial change can have a huge performance impact.

1. Have a Graylog environment with non-trivial amounts of data
2. Create a view with 10+ queries/tabs interacting with all (or a fair amount) of the aforementioned data in various ways.
3. Create a new query/tab within the view that has the default 5 minute search window, etc.
4. Note the time taken is significantly longer than if you had done a "legacy" search in that same 5 minute search window, since all searches required for all queries/tabs of the view are executed.

## Context

We've been investigating an issue and created some view queries/tabs to easily switch between certain timeframes.  I noticed the time taken continued to grow as queries/tabs were added, and eventually even a default query/tab in the view took over 30 seconds to display.

This is related to #6849.

## Your Environment

* Graylog Version: 3.1.3
* MongoDB Version: 4.0.6
* Operating System: RHEL
* Browser version: Firefox

dennisoelkers:
## Motivation and Context

This PR is adding a performance optimization so that only the widgets for the current dashboard page are being executed. For searches, this does not make any difference, as there is always a single page only. For dashboards with lots of widgets (or long time ranges), this can make a big difference regarding search execution times, as widgets which are not currently visible will not be executed.

To achieve this, the following is done in this PR:

  - A parameter `keep_queries` is added to the global override, which
    can be used to select queries which are executed only
  - In the frontend, the currently active query id will always be put
    into the `keep_queries` parameter in the global override

Fixes #6867.

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/issue-6867.toml
+++ b/changelog/unreleased/issue-6867.toml
@@ -1,0 +1,5 @@
+type = "f" # One of: a(dded), c(hanged), d(eprecated), r(emoved), f(ixed), s(ecurity)
+message = "Only executed widgets of current page ond dashboards."
+
+issues = ["6867"]
+pulls = ["14890"]
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/SearchSyncIT.java
@@ -148,6 +148,20 @@ public class SearchSyncIT {
     }
 
     @ContainerMatrixTest
+    void testRequestStoredSearchWithGlobalOverrideKeepingOnlySingleQuery() throws ExecutionException, RetryException {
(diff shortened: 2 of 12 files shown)


linuspahl (approved):
Looks good to me, could not find a problem!

PR #14892, opened 2023-03-09
Execute a single rule in isolation

### LINKED ISSUE #14891: Simulate a single rule
Provide option to simulate a single rule within the context of the rule editor.

## What?

In addition to the Simulate page which simulates everything (codecs, extractors, pipelines, multiple stages and rules), it would be nice to be able to quickly simulate just the rule I am currently working on. This would speed up development of a new rule, when I don't want to be bothered with everything else.

This would not need to be a high-fidelity simulation. Just something to support rapid development of a first working version of a rule. 

## Your Environment

* Graylog Version: 5.1

patrickmann:
Resolves #14891

Leverage existing code in PipelineInterpreter to simulate a single rule in isolation, i.e. 
- no codecs
- no pipelines and stages
- no stream input
- no changes to state

We just take in a message string and run the rule on it. 

Introduces a new API, which takes in a rule and an input message. It returns the processed message.
```
POST /system/pipelines/rule/simulate/{messageString}
      Body: RuleSource
```

Screenshot 2023-03-15 at 16 20 36


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/RuleResource.java
```diff
@@ -173,14 +190,51 @@ public RuleSource parse(@ApiParam(name = "rule", required = true) @NotNull RuleS
...
     }
 
+    @ApiOperation(value = "Simulate a single processing rule")
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/simulate/{messageString}")
+    @NoAuditEvent("only used to test a rule, no changes made in the system")
+    public Message simulate(
+            @ApiParam(name = "rule", required = true) @NotNull RuleSource ruleSource,
+            @ApiParam(name = "messageString", required = true) @PathParam("messageString") @NotBlank String messageString
```
bernd, line 200: We cannot pass the message string as a path parameter. That doesn't work for large values. And there might be escaping issues as well when passing in multi-line strings or structured data like JSON.

Please pass it in the request body. We can create a new request entity that contains the rule source and the message.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/SimulateRuleRequest.java
```diff
@@ -0,0 +1,38 @@
...
+public abstract class SimulateRuleRequest {
+    @JsonProperty
+    public abstract String message();
+
+    @JsonProperty
+    public abstract RuleSource ruleSource();
+
+    @JsonCreator
+    public static SimulateRuleRequest create (@JsonProperty("message") String message,
+                                              @JsonProperty("ruleSource") RuleSource ruleSource) {
```
bernd, line 35: Our API convention is to use snake case instead of camelcase.
```suggestion
                                              @JsonProperty("rule_source") RuleSource ruleSource) {
```

patrickmann: fixed

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/PipelineInterpreter.java
```diff
@@ -368,7 +368,7 @@ private void evaluateStage(Stage stage,
         interpreterListener.exitStage(stage);
     }
 
-    private boolean executeRuleActions(Rule rule,
+    public boolean executeRuleActions(Rule rule,
```
AntonEbel, line 371: I think this method can remain `private`.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/RuleResource.java
```diff
@@ -173,14 +189,50 @@ public RuleSource parse(@ApiParam(name = "rule", required = true) @NotNull RuleS
...
+                .build();
+        stage.setPipeline(pipeline);
+
+        PipelineInterpreter pipelineInterpreter = new PipelineInterpreter(
+                new NoopMessageQueueAcknowledger(), MetricRegistryFactory.create(), configurationStateUpdater);
+        final PipelineInterpreterTracer pipelineInterpreterTracer = new PipelineInterpreterTracer();
+
+        Message message = new Message(request.message(), "127.0.0.1", DateTime.now(DateTimeZone.UTC));
+        final Stream stream = streamService.load(Stream.DEFAULT_STREAM_ID);
+        message.addStream(stream);
```
AntonEbel, line 220: It also works fine without a stream, is there a reason to add a stream here?

AntonEbel (changes requested):

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/RuleResource.java
patrickmann: You're right - removed it.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/processors/PipelineInterpreter.java
patrickmann: fixed

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/RuleResource.java
```diff
@@ -173,14 +188,50 @@ public RuleSource parse(@ApiParam(name = "rule", required = true) @NotNull RuleS
...
+                .build();
+        stage.setPipeline(pipeline);
+
+        PipelineInterpreter pipelineInterpreter = new PipelineInterpreter(
+                new NoopMessageQueueAcknowledger(), MetricRegistryFactory.create(), configurationStateUpdater);
+        final PipelineInterpreterTracer pipelineInterpreterTracer = new PipelineInterpreterTracer();
+
+        Message message = new Message(request.message(), "127.0.0.1", DateTime.now(DateTimeZone.UTC));
+//        final Stream stream = streamService.load(Stream.DEFAULT_STREAM_ID);
+//        message.addStream(stream);
```
AntonEbel, line 219: Is there a reason to comment out this code instead of removing it?

AntonEbel (changes requested):

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rest/RuleResource.java
patrickmann: Forgot to remove it after testing. Gone now.

AntonEbel (approved):
Backend LGTM

graylog2-web-interface/src/components/search/MessageFields.jsx
```diff
@@ -46,7 +46,9 @@ class MessageFields extends React.Component {
 
   render() {
     const { message } = this.props;
-    const formattedFields = message.formatted_fields;
+    // eslint-disable-next-line no-unused-vars
+    const { _id, ...formatted_fields } = message.fields;
```
grotlue, line 50: Why do we need the `_id`? I think we can use `_` and eslint might not complain, but not sure.

grotlue (approved):
Works and looks good! 👍

PR #14300, opened 2022-12-22
Fix overflow of more actions dropdown in streams overview.

linuspahl:
## Motivation and Context

The new layout for the streams overview scroll horizontally when there is not enough space. This resulted in a problem with the more actions dropdown, under some circumstances

image

With this PR we are using a portal for the dropdown to fix the overflow problem
image

/nocl

PR #14897, opened 2023-03-09
Feat/4794 investigations add evidence

zeeklop:
Add evidence to an investigation

## Description

With this feature, we are trying to make it easier for the user to add evidence (messages, searches, dashboards or events) to the investigation they are currently working on and have open.

## How Has This Been Tested?

Unit test

## Types of changes

## Checklist:


simonychuang (approved):
LGTM

graylog2-web-interface/src/components/events/events/EventDetails.tsx
```diff
@@ -74,6 +94,15 @@ const EventDetails = ({ event, eventDefinitionContext }: Props) => {
                   Replay search
                 </ReplaySearchButton>
               </dd>
+              <AddEvidence id={event.id}
+                           type="events"
+                           investigationSelected={(val: boolean) => setDisableAddEvidence(!val)}>
```
linuspahl, line 99: Instead of calling `investigationSelected` and maintaining a state, we could just provide this information for the children:
`<AddEvidence>{({ investigationSelected }) => <dd>}</AddEvidence>`

For cases where we do not need this information, we could still support the current structure, by checking if `children` is a function in `AddEvidence`.
`typeof children === 'function' ? children({ investigationSelected }) : children`

What do you think?

graylog2-web-interface/src/views/components/dashboard/DashboardsOverview/DashboardActions.tsx
```diff
@@ -52,6 +53,11 @@ const DashboardActions = ({ dashboard, refetchDashboards }: Props) => {
 
   return (
     <>
+      <InvestigationsMenu id={dashboard.id}
+                          type="dashboards"
+                          bsSize="xsmall"
+                          title={<span><Icon name="puzzle-piece" size="xs" /> Investigations</span>}
+                          pullRight />
```
linuspahl, line 60: With this component we have an overflow problem in the dashboards overview:
image

I had a similar problem in the past: https://github.com/Graylog2/graylog2-server/pull/14300 and fixed it by creating a `OverlayDropdown` component. 

Alternatively we could also create a `More Actions` dropdown and move `Delete` and "Add to investigation" into it, like we have it on other pages like the streams overview.

graylog2-web-interface/src/components/security/investigations/AddEvidence.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+describe('AddEvidence', () => {
+  it('should render button', () => {
+    render(<AddToInvestigation id="dashboard-id" type="dashboards" />);
+
+    expect(screen.getByText('Add to investigation')).toBeInTheDocument();
+  });
+
+  it('should not fail and render null if no plugin is registered', () => {
+    mockExports.mockReturnValueOnce(undefined);
```
linuspahl, line 131: Since we use `mockReturnValueOnce` it does not really reflect the actual behaviour. I personally prefer the following approach:

(Optionally) Not defining any return value in the initial mock
```
jest.mock('graylog-web-plugin/plugin', () => ({
  PluginStore: { exports: jest.fn() },
}));
```

Using `beforeEach` instead to mock the return value:
```
  beforeEach(() => {
    asMock(PluginStore.exports).mockReturnValue(mockExports());
  });
```

In the test we can now use `mockReturnValue`:
```
asMock(PluginStore.exports).mockReturnValue(undefined);
```

In this test the benefit is neglectable, but in general this has the benefits,
- That mocking in a test never affects the following tests, because we always reset the mock.
- We can use `mockReturnValue` and can be sure our test specific mock is consistent, even after a rerender.
- It is slightly easier to read, since you do not need to look up what module e.g. `mockUseInvestigationDrawer` is relating to.

The mocking could still be simple if we use default values for `mockExports`:
```
const mockExports = (AddEvidence = MockSecAddEvidence, useInvestigationDrawer = mockUseInvestigationDrawer) =>

...
asMock(PluginStore.exports).mockReturnValue(mockExports(undefined, { selectedInvestigationId: undefined });
```

graylog2-web-interface/src/components/events/events/EventDetails.tsx
zeeklop: I really like this pattern. I have seen it, but never implemented it. It makes more sense and it is cleaner. I'll make this change. Thanks! Excellent suggestion.

graylog2-web-interface/src/views/components/dashboard/DashboardsOverview/DashboardActions.tsx
zeeklop: OK. I'll do that, Add a `More Actions` menu with the delete and add to investigation options

zeeklop: Looking into this issue. Looks like the root of the problem is the `EntityDataList` component. Setting `oveflow-x: 'auto'` for the wrapper element was also affecting `overflow-y`

Setting `overflow-x: 'clip'` prevents the table overflow the container div and it doesn't affect `overflow-y`, but there is no scroll option in the X axis.

I observed that the issue is at a very small break point, so we might need to create another PR to handle the overflow in the different break points.

One Idea could be to collapse all action under a dropdown using the `ellipsis` icon for the trigger. That would make the Actions column very small, preventing the overflow of the table.

graylog2-web-interface/src/components/security/investigations/AddEvidence.test.tsx
zeeklop: In this case, I'll have to disagree. The intention is to test the plugin functionality and how the proxy element handles the absence of the plugin.

So the absence of the plugin is the exception. In my mind, it makes more sense to have one case that uses `mockReturnValueOnce` to test the exception and for the rest of the tests, use the default behavior with the plugin present.

graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
```diff
@@ -221,6 +222,22 @@ const EntityDataTable = <Entity extends EntityBase>({
...
+      scrollContainer.style.overflow = scrollContainer.style.overflow !== 'visible' ? 'visible' : 'auto';
+    };
+
+    document.addEventListener('click', handleOverflow, false);
+
+    return () => {
+      document.removeEventListener('click', handleOverflow, false);
+    };
+  }, []);
+  */
```
linuspahl, line 239: Let's remove the comment.

graylog2-web-interface/src/views/components/dashboard/DashboardsOverview/DashboardActions.tsx
linuspahl: Unfortunately we can not rely on the breakpoint being small. The `EntityDataTable` is also being used on other pages and their content (the columns we display) is configurable. I think we need the horizontal scrolling.

My suggestion is to keep `overflow-x` and to use the `OverlayDropdown` button instead of the `DropdownButton`, since it displays the dropdown in a portal.

graylog2-web-interface/src/components/security/investigations/AddEvidence.test.tsx
linuspahl: I think we have the same idea of the general process.

In my suggestion the plugin would be present by default, since it will always be defined for each test with `beforeEach`. 

I just did not defined it in the initial `jest.mock`, since this will always be overwritten by `beforeEach`.
In my suggestion the `jest.mock` is only necessary to be able to mock the module.

My main point is, when we use `mockReturnValueOnce`, `PluginStore.exports` will only be `undefined` (in this test case) on the first render. When the component rerenders we call `PluginStore.exports` again and the plugin will be present again, although this would not be the case in production.

In every case this is just a general pattern I wanted to talk about and it is neglectable for this case and I have no problem with merging the current state.

linuspahl: I just saw that we are using `usePluginEntities`, which memorises the return value of `PluginStore.exports`, which means my suggestion ist even less relevant for this specific case.

graylog2-web-interface/src/components/common/EntityDataTable/EntityDataTable.tsx
zeeklop: Thanks, I forgot to get that out when I was trying things to fix the overflow issue

PR #14904, opened 2023-03-10
Unbinding subject from thread context at beginning of filter chain.

dennisoelkers:
## Motivation and Context

This change is explicitly unbinding the subject at the _beginning_ of the filter chain. Previously this happened through a `ContainerResponseFilter`, which is not always called.

/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityBinding.java
```diff
@@ -59,8 +59,5 @@ public void configure(ResourceInfo resourceInfo, FeatureContext context) {
             LOG.debug("Resource method {}#{} requires an authorization checks.", resourceClass.getCanonicalName(), resourceMethod.getName());
             context.register(new ShiroAuthorizationFilter(requiresPermissions));
         }
-
-        // TODO this is the wrong approach, we should have an Environment and proper request wrapping
-        context.register((ContainerResponseFilter) (requestContext, responseContext) -> ThreadContext.unbindSubject());
```
mpfz0r, line 64: hmm, i guess it can't hurt to keep this as well, no?

dennisoelkers: I do not think it makes sense to keep it when there are requests where it is just not executed, in theses cases it provides a false sense of security.

mpfz0r: fair enough

mpfz0r (approved):
LGTM. I think we should try to do this like suggested here, if possible.
https://shiro.apache.org/subject.html#Subject-ThreadAssociation

PR #14914, opened 2023-03-13
Retrieve subject initially when constructing `SearchUser`.

dennisoelkers:
## Motivation and Context

Prior to this PR, in the constructed `SearchUser` instance, the subject was resolved for every call. This lead to the `SearchUser` being unusable in other threads than the one of the request, because for resolving the subject the `ThreadContext` is required.

This PR is now resolving the subject upon construction and keeps the reference to it in the `SearchUser` instance. This has the potential to change the behavior in a subtle way, because due to a single reference being kept instead of resolving the subject for each call, the subject might have changed before between calls and does not do anymore. The behavior now is potentially more consistent for short-running tasks, because the subject is consistent for a single request.

/nocl

## Types of changes

## Checklist:


janheise (approved):
LGTM!

PR #14928, opened 2023-03-14
Endpoint providing entities titles

luk-kaminski:
## Description
Endpoint providing entities titles.

Underneath, a significant refactoring has been started.
@CollectionName annotation has been replaced with more complex @DBEntity annotation.
A new catalog of entities has been created, built automatically from the data present in @DBEntity annotations on model classes.
All the entities that have been annotated with @CollectionName before are now annotated with @DBEntity and are accessible through the catalog.

/jenkins-pr-deps Graylog2/graylog-plugin-collector#225
/nocl

## Motivation and Context
Frontend needs an easy way to get titles of many different entities with one call.
Sending many requests, with a lot of unnecessary data about each entity, is not acceptable.

## How Has This Been Tested?
Manually and with new automatic tests.

obraz

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/entities/titles/EntityTitleResource.java
bernd, line 1: I think this new system can be part of the system catalog refactoring to make it independent from content packs.

I suggest adding the resources to the `CatalogResource` under `/system/catalog/entities/titles` or keeping this separate resource but putting it under the `/system/catalog` API namespace. (`/system/catalog/entities/titles`)
That way, we don't introduce a separate namespace for related functionality. Once we move out the catalog features from the content packs namespace, we can move this resource without having to change the API endpoint.

What do you think?

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorImpl.java
```diff
@@ -30,10 +30,15 @@
...
+import static org.graylog2.database.DbEntity.NO_TITLE;
+import static org.graylog2.shared.security.RestPermissions.DEFLECTOR_READ;
+
 @AutoValue
 @WithBeanGetter
 @JsonAutoDetect
-@CollectionName("decorators")
+@DbEntity(collection = "decorators",
+          titleField = NO_TITLE,
+          readPermission = DEFLECTOR_READ)
```
bernd, line 41: The `DEFLECTOR_READ` permission is related to index management, not to decorators. :slightly_smiling_face:
I guess you meant the `DECORATORS_READ` permission here.

graylog2-server/src/main/java/org/graylog2/rest/resources/entities/titles/EntityTitleService.java
```diff
@@ -0,0 +1,137 @@
...
+                }
+        );
+
+        return new EntitiesTitleResponse(titles, notPermitted);
+    }
+
+    private boolean checkCanReadTitle(final Subject subject,
+                                      final String readPermission,
+                                      final String idAsString) {
+        return DbEntity.NO_PERMISSION.equals(readPermission) ||
```
bernd, line 134: This makes all `DbEntity`s without an explicit permission configuration readable for all users, right?
I think it should be the other way around, not permitted by default.

What do you think?

bernd (changes requested):
Thanks for starting this! :pray: 

First round of feedback and questions.

graylog2-server/src/main/java/org/graylog2/decorators/DecoratorImpl.java
luk-kaminski: Yes! Thanks for spotting this !!!

graylog2-server/src/main/java/org/graylog2/rest/resources/entities/titles/EntityTitleResource.java
luk-kaminski: I have no strong opinion.
For me "EntityTitleResource" does not need to be part of the `catalog` package and the fact that it uses catalog underneath... I treated it as an implementation detail. So I thought about it in a functional way - if an API user wants to get some titles, he has a top-level resource just for that and does not need to search for it in any other resource.
But I also understand that it kind of pollutes the resource list with a very "technical" endpoint, so if you prefer to move it to a new `CatalogResource`, I am also fine with than.

graylog2-server/src/main/java/org/graylog2/rest/resources/entities/titles/EntityTitleService.java
bernd: @luk-kaminski Hmm, good question. :thinking: How about a property on the annotation? (e.g., `@DbEntity(requiresReadPermission = false)` and default is `true`)

thll (changes requested):

graylog2-server/src/main/java/org/graylog2/database/dbcatalog/DbEntitiesScanner.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @SuppressWarnings("unused")
+    DbEntitiesScanner(String[] packagesToScan) {
+        this.packagesToScan = packagesToScan;
+    }
+
+    @Override
+    public DbEntitiesCatalog get() {
+        final ConfigurationBuilder configuration = new ConfigurationBuilder()
+                .forPackages(packagesToScan)
```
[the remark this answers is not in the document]
luk-kaminski: 3 entities from graylog-plugin-collector, which have been given new annotation in related PR https://github.com/Graylog2/graylog-plugin-collector/pull/225, seem to be missing.

graylog2-server/src/main/java/org/graylog2/alerts/AlertImpl.java
```diff
@@ -30,9 +30,11 @@
 import javax.annotation.Nullable;
 import java.util.Map;
 
+import static org.graylog2.alerts.AlertImpl.FIELD_DESCRIPTION;
+
 @AutoValue
 @JsonAutoDetect
-@CollectionName("alerts")
+@DbEntity(collection = "alerts", titleField = FIELD_DESCRIPTION)
```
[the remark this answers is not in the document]
luk-kaminski: Good catch. It seems that we don't use it anymore.
It seems that alert is now simply an Event with `alert=true`.
I'll try to remove this class and Alert interface.

luk-kaminski:
> @luk-kaminski: that's a cool feature. I really like how little effort it will take in the future to add new entities to the catalogue! +1

Thanks! And thank you for the in-depth review!!!

graylog2-server/src/main/java/org/graylog2/rest/resources/entities/titles/EntityTitleService.java
luk-kaminski: @bernd - this would require handling a situation when someone states that permission is required, but does not provide this permission... or permission is not required, but provides this permission. 

I am seriously considering another constant (FORBIDDEN_ACCESS = "you:shall:not:pass!" ? ;) ), used as default.

graylog2-server/src/main/java/org/graylog2/security/AccessTokenImpl.java
```diff
@@ -18,15 +18,18 @@
...
 import org.graylog2.plugin.database.validators.Validator;
 import org.joda.time.DateTime;
 
 import java.util.Map;
 
-@CollectionName(AccessTokenImpl.COLLECTION_NAME)
+import static org.graylog2.security.AccessTokenImpl.COLLECTION_NAME;
+import static org.graylog2.security.AccessTokenImpl.NAME;
+
+@DbEntity(collection = COLLECTION_NAME, titleField = NAME)
```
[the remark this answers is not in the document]
luk-kaminski: I've changed the approach so that by default you are forbidden to work with the entity (before "missing" readPermission meant "you can access").
It should solve the case.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/contentpacks/CatalogResource.java
```diff
@@ -78,4 +86,14 @@ public CatalogResolveResponse resolveEntities(
 
         return CatalogResolveResponse.create(entities);
     }
+
+    @POST
+    @Timed
+    @ApiOperation(value = "Get titles of provided entities")
+    @NoAuditEvent("This endpoint does not change any data")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/entity_titles")
```
bernd, line 95: Let's make the naming a bit more RESTful. :slightly_smiling_face: 
```suggestion
    @Path("/entities/titles")
```

graylog2-server/src/main/java/org/graylog2/alerts/Alert.java
bernd, line 1: Let's remove the alerting interface and implementations in a separate PR. It needs a separate changelog entry for the removal and a note in `UPGRADING.md`.

Please check if you can find a deprecation notice of this interface in previous upgrade notes. I think we left it in to not break third-party plugins. There are also more alert-related classes in the `org.graylog2.alerts` package. So if we decide to remove the classes, we should remove all of them.

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/database/DbEntity.java
```diff
@@ -34,5 +34,5 @@
     String NO_TITLE = "";
 
     String ALL_ALOWED = "";
-    String NOBODY_ALOWED = "youshall:notpass";
+    String NOBODY_ALOWED = "__no_access__";
```
bernd, line 37: Sorry for being the party :hankey:. :smile:

luk-kaminski: I will come back as Gandalf the White! ;)

graylog2-server/src/main/java/org/graylog2/database/dbcatalog/DbEntitiesScanner.java
```diff
@@ -33,21 +35,29 @@ public class DbEntitiesScanner implements Provider<DbEntitiesCatalog> {
...
     }
 
     @Override
     public DbEntitiesCatalog get() {
         final ConfigurationBuilder configuration = new ConfigurationBuilder()
                 .forPackages(packagesToScan)
                 .setScanners(Scanners.TypesAnnotated);
+        if (chainingClassLoader != null) {
+            configuration.setClassLoaders(new ClassLoader[]{chainingClassLoader});
+        }
```
thll: Apparently, the classloader needs to be provided with each package to scan. Otherwise, it won't pick up the correct jars. The following seems to work for me:

```suggestion
        final ConfigurationBuilder configuration = new ConfigurationBuilder();

        if (chainingClassLoader != null) {
            configuration.setClassLoaders(new ClassLoader[]{chainingClassLoader});
            Stream.of(packagesToScan).forEach(pkg -> {
                configuration.forPackage(pkg, chainingClassLoader);
            });
        } else {
            configuration.forPackages(packagesToScan);
        }

        configuration.setScanners(Scanners.TypesAnnotated);
```

graylog2-server/src/main/java/org/graylog2/database/dbcatalog/DbEntitiesScanner.java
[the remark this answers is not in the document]
thll: It seems to be a classloader issue. I've added a separate comment for the current version of the code.

graylog2-server/src/main/java/org/graylog2/database/dbcatalog/DbEntitiesScanner.java
luk-kaminski: It works! Thank you!

thll (approved):
LGTM. Didn't do a thorough test again but all my comments have been addressed.

bernd (commented):
My comments have been fixed. Thank you for working through this! :slightly_smiling_face:

PR #14965, opened 2023-03-17
Optimize service bindings in PersistenceServiceBindings

moesterheld:
## Description

make sure all bindings which use `createIndex()` in the constructor are threadsafe and make them a singleton.

Also use `asEagerSingleton()` and annotation for all singletons.

resolves https://github.com/Graylog2/graylog-plugin-enterprise/issues/4862

## Motivation and Context
`createIndex` makes unnecessary calls to mongoDB.
Eager instantiation helps in finding binding bugs, since they will show up immediately at startup.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/ranges/MongoIndexRangeService.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/ranges/MongoIndexRangeService.java
@@ -48,6 +48,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.inject.Inject;
+import javax.inject.Singleton;
 import java.util.Iterator;
 import java.util.SortedSet;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/ranges/MongoIndexRangeService.java
```diff
@@ -79,25 +81,25 @@ public MongoIndexRangeService(MongoConnection mongoConnection,
         this.auditEventSender = auditEventSender;
         this.nodeId = nodeId;
```
patrickmann, line 82: Don't we need a separate instance for each node?

patrickmann (commented):
Needs CL or /nocl

graylog2-server/src/main/java/org/graylog2/indexer/ranges/MongoIndexRangeService.java
moesterheld: Each node should have it's own vm and within this vm, nodeid will be unique, shouldn't it?

patrickmann (approved):

PR #14967, opened 2023-03-17
Notify when aggregation search throws an exception

### LINKED ISSUE #14746: Event fails silently on huge dataset from search provider
Event definitions fail if search for the event is larger than the permitted dataset from the search provider associated with Graylog.

Via HS-1439970994:

`Aggregation search query <streams-query-1> returned an error: Elasticsearch exception [type=circuit_breaking_exception, reason=[parent] Data too large, data for [<http_request>] would be [16623428320/15.4gb], which is larger than the limit of [16320875724/15.1gb], real usage: [16623425576/15.4gb], new bytes reserved: [2744/2.6kb], usages [request=0/0b, fielddata=0/0b, in_flight_requests=2744/2.6kb, accounting=0/0b]]. ElasticsearchException[Unable to perform search query: ]; nested: ElasticsearchStatusException[Elasticsearch exception [type=circuit_breaking_exception, reason=[parent] Data too large, data for [<http_request>] would be [16623428320/15.4gb], which is larger than the limit of [16320875724/15.1gb], real usage: [16623425576/15.4gb], new bytes reserved: [2744/2.6kb], usages [request=0/0b, fielddata=0/0b, in_flight_requests=2744/2.6kb, accounting=0/0b]]];`

While the issue is what it says on the tin (returned data is too large, and indeed that's a large dataset), the expectation by customer is that this is not a silent failure, and appears in something like the All Events stream, or a red notification banner in System -> Overview. 

- GL: 4.3
- Opensearch 1.3.7
- MongoDB 5.0
- JDK 11

patrickmann:
resolves #14746 

An event definition with aggregation will fail silently (with just a log message) when encountering an ES/OS error.
We now also generate a notification to alert the admin that events are getting lost.


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
```diff
@@ -144,9 +149,18 @@ public AggregationResult doSearch() throws EventProcessorException {
             });
 
             // If we have only EmptyParameterErrors, just return an empty Result
-            if (! (errors.stream().filter(e -> !(e instanceof EmptyParameterError)).count() > 1)) {
+            if (errors.stream().allMatch(e -> e instanceof EmptyParameterError)) {
                 return AggregationResult.empty();
             }
+
+            Notification systemNotification = notificationService.buildNow()
+                    .addType(Notification.Type.GENERIC)
```
mpfz0r, line 157: If we do this, we should introduce a new type, so it does not prevent other GENERIC notifications to show up.
Also, I'd use the new notification key with the event definition id.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
```diff
@@ -144,9 +149,18 @@ public AggregationResult doSearch() throws EventProcessorException {
             });
 
             // If we have only EmptyParameterErrors, just return an empty Result
-            if (! (errors.stream().filter(e -> !(e instanceof EmptyParameterError)).count() > 1)) {
+            if (errors.stream().allMatch(e -> e instanceof EmptyParameterError)) {
```
mpfz0r, line 152: we might consider backporting this fix by itself

changelog/unreleased/issue-14746.toml
```diff
@@ -0,0 +1,10 @@
+type = "c"
+message = "Generate a system notification when an aggregation search errors out unexpectedly."
+
+details.user = """
+An event definition with aggregation will fail silently (with just a log message) when encountering an ES/OS error.
+We now also generate a notification to alert the admin that events are getting lost.
```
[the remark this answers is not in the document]
patrickmann: Improved wording for accuracy

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
patrickmann: Agreed. That is an inconvenient limitation we will need to eliminate sometime.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
patrickmann: I'll check with triage

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
```diff
@@ -144,9 +149,19 @@ public AggregationResult doSearch() throws EventProcessorException {
...
                 return AggregationResult.empty();
             }
+
+            Notification systemNotification = notificationService.buildNow()
+                    .addType(Notification.Type.SEARCH_ERROR)
+                    .addSeverity(Notification.Severity.NORMAL)
+                    .addTimestamp(DateTime.now(DateTimeZone.UTC))
+                    .addKey(searchJob.getId())
```
mpfz0r, line 160: I think it would be better to use the eventdefintion id here. 
Otherwise we might get a new notification for each failed attempt, which might be a lot.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java
```diff
@@ -144,9 +149,19 @@ public AggregationResult doSearch() throws EventProcessorException {
...
+                    .addDetail("title", "Aggregation search failed")
```
mpfz0r, line 161: We should also add the title/id of the event definition here

mpfz0r (approved):

PR #14974, opened 2023-03-20
Fix permission checks for role "Event Notification Creator"

moesterheld:
## Description
Users with role "Event Notification Creator" got an error when trying to create a email notification.
This was caused by the user selection dropdown in that notification type's form checking if the user has the permission `users:list` and returning undefined. The permission check was moved into the form and prevents the whole field from showing when the user does not have the permission.

Additionally, the permission was added to the role "User inspector", which only had `users:read` previously to be able to assign a sensible role which actually has this permission.

## Motivation and Context
closes https://github.com/Graylog2/graylog-plugin-enterprise/issues/4886

## Types of changes

## Checklist:


patrickmann (approved):
Looks good and works

PR #14975, opened 2023-03-20
Use singleton scope to bind some services

thll:
Some services ensure that proper indexes exist in MongoDB in their constructor.

If the services are not bound as singletons, we issue a bunch of unnecessary index create calls e.g. for REST API requests.

Together with https://github.com/Graylog2/graylog-plugin-enterprise/pull/4888 this is the effect:

Before:
image

After:
image

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
+++ b/graylog2-server/src/main/java/org/graylog/security/DBGrantService.java
@@ -30,6 +30,7 @@ import org.mongojack.DBQuery;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
+import javax.inject.Singleton;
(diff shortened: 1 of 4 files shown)


mpfz0r (commented):
I think @moesterheld beat you to it

https://github.com/Graylog2/graylog2-server/pull/14965

thll:
> I think @moesterheld beat you to it
> 
> #14965

This is so not fair 😂

moesterheld:
I didn't take care of the DBGrantService

thll:
> I didn't take care of the DBGrantService

Your change results in the same optimized trace, though. I think it's the binding of `UserService` as a singleton which has the biggest effect, so I'm fine with leaving this PR be.

PR #14979, opened 2023-03-20
Telemetry integration to the UI

### LINKED ISSUE #14900: Add telemetry for generic usage metrics collections in Graylog
Add the following features to core to support telemetry:

- Saving user preference for telemetry data metrics
- Add tools to configure and send metric
- Instrument metric collection for events. (ex. pageview)

ousmaneo:
This PR implements anonymous usage data collection on the UI part. 

fix #14900 
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4943
## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryResponseFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                                          int nodesCount,
+                                          Map<String, SystemOverviewResponse> nodes,
+                                          long averageLastMonthTraffic,
+                                          long usersCount) {
+        Map<String, Object> clusterInfo = new LinkedHashMap<>();
+        clusterInfo.put("cluster_id", clusterId);
+        clusterInfo.put(f("%s.cluster_creation_date", CLUSTER), clusterCreationDate);
+        clusterInfo.put(f("%s.nodes_count", CLUSTER), nodesCount);
+        clusterInfo.put(f("%s.average_last_month_traffic", CLUSTER), averageLastMonthTraffic);
+        clusterInfo.put(f("%s.users_count", CLUSTER), usersCount);
```
kroepke, line 94: I think this does not need the `%s`, `CLUSTER` nesting, we will only ever have a single cluster here. If we ever have multi-cluster setups that users interact with, that will need a different representation anyay.
Remove the cluster id here makes the reported properties much simpler, too.

graylog2-web-interface/src/logic/telemetry/TelemetryProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+        const {
+          cluster: { cluster_id: clusterId, ...clusterDetails },
+          current_user: { user, ...userDetails },
+          license,
+          plugin,
+          search_cluster: searchCluster,
+          user_telemetry_settings: { telemetry_permission_asked: isPermissionAsked },
+        } = telemetryData as TelemetryDataType;
+
+        posthog.group('company', clusterId, {
```
kroepke, line 74: could we rename this to `'cluster'` before merging, please?

graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryResource.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @PUT
+    @Path("user/settings")
+    @ApiOperation("Update a user's telemetry settings.")
+    @ApiResponses({@ApiResponse(code = 404, message = "Current user not found.")})
+    @NoAuditEvent("Audit event is sent manually.")
+    public void saveTelemetryUserSettings(@ApiParam(name = "JSON body", value = "The telemetry settings to assign to the user.", required = true)
+                                          @Valid @NotNull TelemetryUserSettings telemetryUserSettings) {
+
+        User currentUser = getCurrentUserOrThrow();
```
moesterheld, line 100: This is the endpoint which is also called when editing another user's settings, so either this needs to be changed to accept a user id or disabled in the UI.
Right now, changing the value for any user in the UI will result in changing my own setting.

moesterheld: Oh yeah, same for `getTelemetrySettings`

graylog2-web-interface/src/components/indices/IndexSetsComponent.jsx
```diff
@@ -64,7 +66,14 @@ const IndexSetsComponent = createReactClass({
 
   _onSetDefault(indexSet) {
     return () => {
-      IndexSetsActions.setDefault(indexSet).then(() => this.loadData(this.currentPageNo, this.currentPageSize));
+      IndexSetsActions.setDefault(indexSet).then(() => {
+        this.loadData(this.currentPageNo, this.currentPageSize);
```
grotlue, line 70: I think we should not make tracking dependent on any calls or actions. If I understand correctly, we want to track user behavior, so I'd track on click immediately, as the user clicks anyways, no matter if calls fail or not.

graylog2-web-interface/src/components/inputs/CreateInputControl.jsx
```diff
@@ -94,6 +102,11 @@ const CreateInputControl = createReactClass({
   _createInput(data) {
     InputsActions.create(data).then(() => {
       this.setState(this.getInitialState());
+
+      this.props.sendTelemetry('submit_form', {
```
grotlue, line 106: Same here, the tracking should not depend on the action.

graylog2-web-interface/src/components/inputs/InputStateControl.jsx
```diff
@@ -66,14 +68,28 @@ const InputStateControl = createReactClass({
     this.setState({ loading: true });
 
     InputStatesStore.start(this.props.input)
-      .finally(() => this.setState({ loading: false }));
+      .finally(() => {
+        this.setState({ loading: false })
+
+        this.props.sendTelemetry('click', {
```
grotlue, line 74: Same here, we should track the click as soon as the user clicks.

graylog2-web-interface/src/components/inputs/InputStateControl.jsx
```diff
@@ -66,14 +68,28 @@ const InputStateControl = createReactClass({
...
 
   _stopInput() {
     this.setState({ loading: true });
 
     InputStatesStore.stop(this.props.input)
-      .finally(() => this.setState({ loading: false }));
+      .finally(() => {
+        this.setState({ loading: false })
+
+        this.props.sendTelemetry('click', {
```
grotlue, line 88: Dependent tracking.

graylog2-web-interface/src/components/outputs/OutputsComponent.jsx
```diff
@@ -100,6 +106,11 @@ const OutputsComponent = createReactClass({
     StreamsStore.addOutput(this.props.streamId, outputId, (response) => {
       this._handleUpdate();
 
+      this.props.sendTelemetry('submit_form', {
```
grotlue, line 109: Dependent tracking.

graylog2-web-interface/src/components/outputs/OutputsComponent.jsx
```diff
@@ -110,6 +121,11 @@ const OutputsComponent = createReactClass({
         UserNotification.success('Output was terminated.', 'Success');
         this._handleUpdate();
 
+        this.props.sendTelemetry('submit_form', {
```
grotlue, line 124: Dependent tracking.

graylog2-web-interface/src/components/roles/RoleEdit/UsersSection.tsx
```diff
@@ -59,6 +61,11 @@ const UsersSection = ({ role: { id, name }, role }: Props) => {
     .then((result) => {
       setPaginatedUsers(result);
 
+      sendTelemetry('click', {
```
grotlue, line 64: Dependent tracking.

graylog2-web-interface/src/logic/telemetry/TelemetryInfoText.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Icon, ExternalLink } from 'components/common';
+import { Alert } from 'components/bootstrap';
+
+type Props = {
+  showProfile?: boolean
+}
+
+const TelemetryInfoText = ({ showProfile }: Props) => {
+  return (
+    <Alert bsStyle="info">
```
grotlue, line 28: I have some suggestions for improvements in the copy:

```
We would like to collect anonymous usage data to help us prioritize improvements and make Graylog better in the future.

We do not collect personal data, sensitive information, or content such as logs in your instances.

Learn more on our [Privacy Policy](https://www.graylog.com/).

You can turn data collection off or on at any time in the user profile.
```

graylog2-web-interface/src/logic/telemetry/TelemetryInfoText.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <Icon name="info-circle" /> We would like to collect anonymously usage data to help us prioritize improvements
+      and make Graylog better in the future.
+      <br />
+      We <b>do not</b> collect any personal data, any sensitive information or content such as logs that are in your
+      instances.
+      <br />
+      Learn more on our <ExternalLink href="https://www.graylog.com/">Privacy Policy</ExternalLink>.
```
grotlue, line 35: We need to update the link to link to a proper Privacy Policy.

graylog2-web-interface/src/logic/telemetry/TelemetryInit.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    asMock(TelemetrySettingsStore.getInitialState).mockReturnValue({
+      telemetrySettings: {
+        telemetry_permission_asked: false,
+        telemetry_enabled: true,
+      },
+    });
+
+    const { result } = renderHook(() => usePostHog(), { wrapper: Wrapper });
+
+    expect(result.current).toBeTruthy();
```
grotlue, line 74: Can we make this expect more explicit?

graylog2-web-interface/src/pages/IndexSetCreationPage.tsx
```diff
@@ -61,6 +63,11 @@ const IndexSetCreationPage = ({ retentionStrategies, rotationStrategies, retenti
     copy.creation_date = adjustFormat(new Date(), 'internal');
 
     return IndexSetsActions.create(copy).then(() => {
+      sendTelemetry('submit_form', {
```
grotlue, line 66: Dependent tracking.

graylog2-web-interface/src/threatintel/components/ThreatIntelPluginConfig.jsx
```diff
@@ -92,7 +94,14 @@ const ThreatIntelPluginConfig = createReactClass({
   },
 
   _saveConfig() {
-    this.props.updateConfig(this.state.config).then(() => {
+    const { updateConfig, sendTelemetry } = this.props;
+
+    updateConfig(this.state.config).then(() => {
```
grotlue, line 99: Dependent tracking.

graylog2-web-interface/src/util/StringUtils.ts
```diff
@@ -50,6 +50,34 @@ const StringUtils = {
...
+    if (typeof reactNode === 'string') {
+      return reactNode;
+    }
+
+    const { children } = (reactNode as any)?.props || {};
+
+    if (Array.isArray(reactNode)) {
+      const joinedNodes = [];
+
+      reactNode.forEach((node) => {
```
grotlue, line 63: Why not use `map` instead?

graylog2-web-interface/src/views/components/CreateNewWidgetModal.tsx
```diff
@@ -82,7 +82,9 @@ const CreateNewWidgetModal = ({ onCancel, position }: Props) => {
 
   return (
     <Modal onHide={onCancel}
-           show>
+           show
+           data-app-section=""
```
grotlue, line 86: Can we just not set the attribute at all?

grotlue (changes requested):
The general approach looks good! 
Code-wise the main point I found is that some of our tracking depends on API calls, which I think we shouldn't do, as we are tracking clicks and e.g. not only successful submits. I tried to comment all lines in the code, but not sure if I found them all.

I tested the pageview tracking, info modal setting and the tracking on search page. I have not tested all the other sub pages as I understood that the tracking is not complete. If there are specific pages I should have a look at, let me know which :)

# Testing Results 

## Info modal
- ✅ Telemetry info modal is shown on first load
- ❌ “Privacy Policy” link still links to [graylog.org](http://graylog.org/)
- ✅ The modal does not appear again after closing it (when logging out and back in)
- ✅ The modal is shown to each user independently

## Tracking
- ✅Telemetry can be turned off in the user profile (except for the admin user)
- ❌ Tracking is still sent after turning telemetry off in the user profile
- ✅ Does not track anything when logged out (on login page)
- ✅ Identifies the user when logging in 
- ✅ Page view event is tracked on page load

### Modals and popovers
Tested those on search page

- ❌ When closing a modal or popover we have no information if it was closed via cancel or save/confirm. The event_info only contains a constructer, no other values.

### Search Page
- ❌ When resetting the search stream filter it is tracked as “change_input_value”. The event_info only has a constructor, and no info about the value that was selected. Having this will enable us to differentiate.
- ⚠️ No tracking inside of the “Search Time Range” form
- ⚠️ No tracking for the search button
- ⚠️ No tracking for the widgets
- ⚠️ No tracking for search filters

graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryResource.java
AntonEbel: Good catch. I spoke to Ousmane, we are adjusting the UI so that a user can only change his own telemetry settings.

ousmaneo: Hi @moesterheld, now we are not allowing a user to change the setting for other. You can only change your own if your profile is not read-only.

grotlue (changes requested):
✅ Tracking is now not fired when I turn it off in user profile

I have one small text suggestions.

Besides that I think we should make sure we distinguish between modal/popover cancel & "submit" actions for closing as they can tell a lot about user behaviour.

The rest looks good! Thanks a lot!

graylog2-server/src/test/java/org/graylog2/telemetry/rest/TelemetryServiceTest.java
```diff
@@ -0,0 +1,166 @@
...
+        mockUserTelemetryEnabled(false);
+
+        Map<String, Object> response = telemetryService.getTelemetryResponse(user, Map.of());
+
+        assertThat(response).containsOnlyKeys(USER_TELEMETRY_SETTINGS);
+    }
+
+    @Test
+    void test_no_telemetry_user_settings() {
+        TelemetryService telemetryService = createTelemetryService(true);
```
moesterheld, line 105: Why do telemetry data have to be collected when the permission has not been asked?

moesterheld (approved):
two small comments, otherwise backend lgtm

graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryResponseFactory.java
AntonEbel: As discussed, the cluster node data is no longer flattened.

graylog2-web-interface/src/util/StringUtils.ts
gally47: Because the output array isn't guaranteed to have the same length as the input array.

grotlue:
> > ✅ Tracking is now not fired when I turn it off in user profile
> > I have one small text suggestion.
> > Besides that I think we should make sure we distinguish between modal/popover cancel & "submit" actions for closing as they can tell a lot about user behaviour.
> > The rest looks good! Thanks a lot!
> 
> @grotlue I am not sure if we should distinguish between modal/popover cancel & submit actions for closing, but since it can't be done globally in the modal component, I suggest we add a new PR if it's needed.

I think we should definitely distinguish between those actions. If it's too much effort for this PR we can do this in another one, but I'd really think that we create an issue for this and add this soon. We have a lot of modals and popovers and knowing when a user cancels an action (e.g. closing modal via cancel button) will give us a lot of insights.

graylog2-web-interface/src/views/components/searchbar/SearchButton.tsx
```diff
@@ -53,19 +54,32 @@ type Props = {
...
 };
 
-const onButtonClick = (e: MouseEvent, disabled: Boolean) => {
+const onButtonClick = (e: MouseEvent, disabled: Boolean, triggerTelemetry: () => void) => {
   if (disabled) {
     e.preventDefault();
     QueryValidationActions.displayValidationErrors();
   }
+
+  triggerTelemetry();
```
grotlue, line 63: We can call `sendTelemetry` here directly, no need for an extra function.

graylog2-web-interface/src/views/components/widgets/WidgetActionsMenu.tsx
```diff
@@ -151,12 +152,48 @@ const WidgetActionsMenu = ({
...
+      appSection: 'search-widget',
+      eventElement: 'widget-delete-button',
+    });
+
+    return dispatch(_onDelete(widget, view, title));
+  }, [dispatch, sendTelemetry, title, view, widget]);
+  const focusWidget = useCallback(() => {
+    sendTelemetry('click', {
+      appSection: 'search-widget',
+      eventElement: 'widget-delete-button',
```
grotlue, line 192: Is this the correct eventElement?

grotlue (approved):
Frontend part looks good.

I have two small comments, I think the 2nd one should be looked at, but after that it's ready to go.

As said I think we need an issue for improving the modal tracking.

Besides that, I think it would be great if we have a documentation about our tracking: What events we track, what data do we send with which event, how to name elements. This way we can make sure that other engineers can add tracking properly to new features in the future.

Thanks a lot for working on this topic! Excited to see it go live!

graylog2-web-interface/src/views/components/searchbar/SearchButton.tsx
ousmaneo: Actually, we can't because the function is defined outside of the component so we need to pass the function or move it inside the scope of the component.

ousmaneo:
> Besides that, I think it would be great if we have a documentation about our tracking: What events we track, what data do we send with which event, how to name elements. This way we can make sure that other engineers can add tracking properly to new features in the future.


Thank you for your review, I will prepare the documentation for the telemetry.

graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryService.java
```diff
@@ -0,0 +1,208 @@
...
+        this.elasticsearchVersion = elasticsearchVersion;
+        this.telemetryResponseFactory = telemetryResponseFactory;
+        this.dbTelemetryUserSettingsService = dbTelemetryUserSettingsService;
+        eventBus.register(this);
+    }
+
+    public Map<String, Object> getTelemetryResponse(User currentUser, Map<String, SystemOverviewResponse> systemOverviewResponses) {
+        TelemetryUserSettings telemetryUserSettings = getTelemetryUserSettings(currentUser);
+        if (isTelemetryEnabled && telemetryUserSettings.telemetryEnabled()) {
+            String clusterId = getClusterId();
```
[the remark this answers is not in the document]
AntonEbel: fixed

graylog2-server/src/test/java/org/graylog2/telemetry/rest/TelemetryServiceTest.java
AntonEbel: We do not check permission here, as a user can only check his own telemetry settings.

graylog2-web-interface/src/logic/telemetry/TelemetryProvider.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    if (posthog) {
+      setGroup();
+    }
+  }, [posthog, isTelemetryDataLoaded, telemetryData]);
+
+  const TelemetryContextValue = useMemo(() => {
+    const sendTelemetry = (eventType: TelemetryEventType, event: TelemetryEvent) => {
+      if (posthog) {
+        posthog.capture(eventType, event);
```
dennisoelkers, line 66: Can we wrap this call with a `try {} catch { console.warn(...); }`? It is now used all over the product and we should make sure that adding telemetry does not interfere with other functionality.

dennisoelkers, line 66: What happens when the user has telemetry disabled? We are skipping the group configuration in the `useEffect` block in that case, but should we also just return a no-op function here instead of calling `posthog.capture()`?

dennisoelkers (changes requested):
The code looks good in general. Due to the size I was not able to look at the complete code and focussed on the core parts in general, adding two major questions/suggestions in there.

graylog2-web-interface/src/logic/telemetry/TelemetryProvider.tsx
ousmaneo: Thank you, @dennisoelkers, for having a look and for the suggestions.

> What happens when the user has telemetry disabled? We are skipping the group configuration in the useEffect block in that case, but should we also just return a no-op function here instead of calling posthog.capture()?


When telemetry is disabled, the init is not called, and the PostHog context should be empty, resulting in `usePostog` returning `posthog` with the value of `undefined`, therefor calling `posthog.capture` is not called.

graylog2-server/src/main/java/org/graylog2/telemetry/rest/TelemetryResource.java
ousmaneo: Thank you @moesterheld . I fixed the view part

dennisoelkers (approved):
Looks good to me!

PR #15003, opened 2023-03-22
Datanode periodical for NodeService pings

todvora:
Added missing node ping for datanode 

Additionally there is a small refactoring included, that removes the need to query Node from the DB before we can trigger markAsAlive. This saves one mongodb query every second on each node. The newly created markAsAlive method also prepares a bit of infrastructure for the usage of the `$currentDate`. This will help us to use the mongodb time and avoid out-of-sync clock of different nodes. But the full implementation will have to happen in a separate PR. It's too invasive and requires backwards compatibility changes.

/nocl

## Motivation and Context
Needed to keep overview over active nodes connected to the graylog server.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/data-node/src/main/java/org/graylog/datanode/management/OpensearchProcess.java
+++ b/data-node/src/main/java/org/graylog/datanode/management/OpensearchProcess.java
@@ -29,6 +29,7 @@ public interface OpensearchProcess extends ManagableProcess {
 
     Object nodeName();
 
+    boolean isLeaderNode();
     void setLeaderNode(boolean isManagerNode);
 
     String opensearchVersion();
--- a/data-node/src/main/java/org/graylog/datanode/periodicals/NodePingPeriodical.java
+++ b/data-node/src/main/java/org/graylog/datanode/periodicals/NodePingPeriodical.java
@@ -18,7 +18,6 @@ package org.graylog.datanode.periodicals;
 
 import org.graylog.datanode.Configuration;
 import org.graylog.datanode.management.OpensearchProcess;
-import org.graylog2.cluster.Node;
 import org.graylog2.cluster.NodeNotFoundException;
 import org.graylog2.cluster.NodeService;
 import org.graylog2.plugin.Tools;
@@ -30,22 +29,32 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nonnull;
 import javax.inject.Inject;
 import java.net.URI;
+import java.util.function.Supplier;
 
 public class NodePingPeriodical extends Periodical {
 
     private static final Logger LOG = LoggerFactory.getLogger(NodePingPeriodical.class);
-
     private final NodeService nodeService;
     private final NodeId nodeId;
-    private final Configuration configuration;
-    private final OpensearchProcess managedOpenSearch;
+    private final URI httpPublishUri;
+    private final Supplier<Boolean> isLeader;
+
 
     @Inject
     public NodePingPeriodical(NodeService nodeService, NodeId nodeId, Configuration configuration, OpensearchProcess managedOpenSearch) {
+        this(nodeService, nodeId, configuration.getHttpPublishUri(), managedOpenSearch::isLeaderNode);
+    }
+
+    NodePingPeriodical(
+            NodeService nodeService,
+            NodeId nodeId,
+            URI httpPublishUri,
+            Supplier<Boolean> isLeader
+    ) {
         this.nodeService = nodeService;
         this.nodeId = nodeId;
-        this.configuration = configuration;
-        this.managedOpenSearch = managedOpenSearch;
+        this.httpPublishUri = httpPublishUri;
+        this.isLeader = isLeader;
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


data-node/src/main/java/org/graylog/datanode/periodicals/NodePingPeriodical.java
```diff
@@ -0,0 +1,103 @@
...
+            final Node node = nodeService.byNodeId(nodeId);
+            nodeService.markAsAlive(node, isLeader, httpPublishUri);
+        } catch (NodeNotFoundException e) {
+            LOG.warn("Did not find meta info of this node. Re-registering.");
+            nodeService.registerServer(nodeId.getNodeId(),
+                    isLeader,
+                    httpPublishUri,
+                    Tools.getLocalCanonicalHostname());
+        }
+    }
```
luk-kaminski, line 102: What about a small unit test for this periodical?

data-node/src/main/java/org/graylog/datanode/management/OpensearchProcess.java
```diff
@@ -29,6 +29,7 @@ public interface OpensearchProcess extends ManagableProcess {
 
     Object nodeName();
 
+    boolean isLeaderNode();
```
luk-kaminski, line 32: It is really minor, but OpensearchProcessImpl class now misses @Override over "isLeaderNode" method signature.

data-node/src/main/java/org/graylog/datanode/periodicals/NodePingPeriodical.java
luk-kaminski: Also, this code seems a little bit duplicated in NodePingThread. Maybe it could be reused?

todvora: yeah, it's a copy of that code. But the params are all different. I don't think it makes sense to abstract such a small piece of code. Let's do it if we'll have a third usage and see some pattern there.

data-node/src/main/java/org/graylog/datanode/management/OpensearchProcess.java
todvora: fixed, thx!

data-node/src/main/java/org/graylog/datanode/periodicals/NodePingPeriodical.java
todvora: Added two unit tests, good point! Thanks!

data-node/src/test/java/org/graylog/datanode/periodicals/NodePingPeriodicalTest.java
```diff
@@ -0,0 +1,90 @@
...
+        task.doRun();
+
+        final ArgumentCaptor<NodeId> nodeIdCaptor = ArgumentCaptor.forClass(NodeId.class);
+        final ArgumentCaptor<Boolean> isLeaderCaptor = ArgumentCaptor.forClass(Boolean.class);
+        final ArgumentCaptor<URI> uriCaptor = ArgumentCaptor.forClass(URI.class);
+        Mockito.verify(nodeService).markAsAlive(nodeIdCaptor.capture(), isLeaderCaptor.capture(), uriCaptor.capture());
+
+        Assertions.assertThat(nodeIdCaptor.getValue().getNodeId()).isEqualTo("5ca1ab1e-0000-4000-a000-000000000000");
+        Assertions.assertThat(isLeaderCaptor.getValue()).isTrue();
+        Assertions.assertThat(uriCaptor.getValue().toString()).isEqualTo("http://localhost:8999");
```
luk-kaminski, line 56: I believe you could get rid of ArgumentCaptor and just use smth like this:
`verify(nodeService).markAsAlive(eq("5ca1ab1e-0000-4000-a000-000000000000"), eq(true), eq("http://localhost:8999"))`

What do you think?

todvora: You are right, that will be cleaner and equally valid. Thanks!

data-node/src/test/java/org/graylog/datanode/periodicals/NodePingPeriodicalTest.java
```diff
@@ -0,0 +1,90 @@
...
+        final ArgumentCaptor<String> nodeIdCaptor = ArgumentCaptor.forClass(String.class);
+        final ArgumentCaptor<Boolean> isLeaderCaptor = ArgumentCaptor.forClass(Boolean.class);
+        final ArgumentCaptor<URI> uriCaptor = ArgumentCaptor.forClass(URI.class);
+        final ArgumentCaptor<String> hostnameCapture = ArgumentCaptor.forClass(String.class);
+        Mockito.verify(nodeService).registerServer(nodeIdCaptor.capture(), isLeaderCaptor.capture(), uriCaptor.capture(), hostnameCapture.capture());
+
+        Assertions.assertThat(nodeIdCaptor.getValue()).isEqualTo("5ca1ab1e-0000-4000-a000-000000000000");
+        Assertions.assertThat(isLeaderCaptor.getValue()).isTrue();
+        Assertions.assertThat(uriCaptor.getValue().toString()).isEqualTo("http://localhost:8999");
+        Assertions.assertThat(hostnameCapture.getValue()).isNotBlank();
```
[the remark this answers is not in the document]
todvora: Done, thx!

luk-kaminski (approved):
Great job, thanks!

danotorrey:
@todvora Just wanted to make sure this is on our radar: The Forwarder also has a custom implementation of the `NodeService` interface:

https://github.com/Graylog2/forwarder/blob/928c9a881099733d9b0e7f0faff59747fe6adb14/forwarder/src/main/java/org/graylog2/cluster/CloudForwarderNodeService.java#L16

I was testing the Forwarder with the latest `master` code and discovered that it appears this change is triggering the following error in the Forwarder due to the API change in `NodeService`:

> forwarder/forwarder/src/main/java/org/graylog2/cluster/CloudForwarderNodeService.java:15:8
java: org.graylog2.cluster.CloudForwarderNodeService is not abstract and does not override abstract method markAsAlive(org.graylog2.plugin.system.NodeId,boolean,java.net.URI) in org.graylog2.cluster.NodeService

image

It looks like we might need to make the API change backwards-compatible, or add a corresponding PR for the Forwarder for the API change.

todvora:
Thanks for catching that @danotorrey ! I'll prepare a PR for the forwarder.

PR #15009, opened 2023-03-22
Backport 4.3: Fix off by one error in aggregation search

patrickmann:
One line fix for an off by one error in aggregation search.

This is a backport of a fix that was unrelated to, but included with #14746.


mpfz0r (approved):
Thank you :+1:

Issue #14574, opened 2023-01-31
TTL for Lookup Table Entries


ChristopherKB:
A time to live feature for lookup table entries would be a very useful addition.
## What?
The ability to assign a time to live, or duration setting for lookup table entries would allow us to automatically delete values that are no longer useful.

## Why?
Some data has a useful life span. Things like policy violations on watchlists or entries on threat intelligence feeds often become less valuable or accurate over time. The ability to automatically delete these entries would enable new types of uses for lookup tables. 

A specific example would be Indicators of Compromise. An IoC often appears and disappears over time. Since IoC lists are constantly updated, a TTL would allow for the automatic deletion of old entries so that lookup tables don't grow forever. Watchlists in particular would benefit from this feature.

Another example would be a check against a list of users who have had password changes in the past 72 hours. If those users get locked out of their account, the presence of their username on a lookup table would enable the analysts to treat these lockouts differently than a user who has not changed their password recently.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

mpfz0r:
@ChristopherKB  but you can control that via the cache on the lookup table. Or am I missing something here?

<img width="1881" alt="image" src="https://user-images.githubusercontent.com/3034831/216277617-9d383944-0a56-4460-bab9-fbc9b151ac48.png">

ChristopherKB:
You're both correct. The cache does allow you to expire a cached entry, but
you cannot actually remove it from the data adapter, whether in Mongo, CSV
or DSV. Adding this capability will allow both new functionality for time
limited conditions and also help to keep large data sets like Threat
Intelligence IoC's from growing out of control with outdated data.

Thanks.

drewmiranda-gl:
There are definitely some use cases i can think of where this would be super useful. 

* Wanting a pipeline rule to run once a day, so you add a value set with TTL 1 day (or 12 hours or whatever), easy way to do a "run this but only once per X"
* Caching/adding data to a lookup table that you don't need/want to be long lived. For example dynamic lookup tables, you may have a bunch of log data where some of the logs contain a value you want associated with the logs and others don't (think something like a username for a web session), while you can use dynamic lookup tables to scrape this via pipelines and add to a dynamic lookup table, that data will get stale really quickly and its not a great idea to have lookup data like that live very long

ryan-carroll-graylog:
[...]

PR #15014, opened 2023-03-22
Adds ability to add a TTL to Data Adapter entries

ryan-carroll-graylog:
Adds structure and a lookup table function to add a TTL to Data Adapter Entries.

Also see changes in: https://github.com/Graylog2/graylog-plugin-enterprise/pull/4854

/nocl
## Description

- Adds `lookup_assign_ttl` lookup table function
- Updates LUT/Data Adapter interfaces with `assignTtl` functions to allow inheriting Data Adapters to implement TTL functionality 
 
## Motivation and Context

Issue: https://github.com/Graylog2/graylog2-server/issues/14574

## How Has This Been Tested?

Locally in dev environment and unit tests.

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/lookup/LookupAssignTtl.java
```diff
@@ -0,0 +1,82 @@
...
+import org.graylog2.plugin.lookup.LookupResult;
+
+import java.util.List;
+
+import static org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor.object;
+import static org.graylog.plugins.pipelineprocessor.ast.functions.ParameterDescriptor.string;
+
+public class LookupAssignTtl extends AbstractFunction<Object> {
+
+    public static final String NAME = "lookup_assign_ttl";
```
kingzacko1, line 35: Would a name like `lookup_set_value_with_ttl` follow our naming convention a little more? It does the same thing as `lookup_set_value` but just with the addition of a TTL. On that note, we also have a `lookup_set_string_list` pipeline function that we may also want to add a TTL to as well.

ryan-carroll-graylog: With this pipeline function we're only setting the TTL for an existing entry with the specified key (see `MongoDBDataAdapterService.assignTtl())`, so I think the naming makes sense but open to changing it if you think otherwise.

Definitely agree that adding an optional parameter to `lookup_set_value` and `lookup_set_string_list` makes sense in the long run so opening a ticket for that as discussed in check-in. For now they could at least achieve the same thing by doing something like:
```
lookup_set_value("table", "key", "val");
lookup_assign_ttl("table", "key", 500);
```
Which I think is not too bad for the time being.

danotorrey (approved):
@ryan-carroll-graylog This looks good to me and tests successfully. Aside from the potential addition of the `ttl` field to the other functions, this all LGTM!

kingzacko1 (approved):

PR #15016, opened 2023-03-22
Add CIDR Range lookup support to CSV file data adapters

kingzacko1:
Adds the capability to match lookup key values to CIDR-formatted entries

## Description

Adds a `CIDR Lookup` configuration option to CSV File data adapters that when enabled will change how lookups are performed. Instead of being an exact text match it will check an IP address against the CIDR addresses in the CSV file.

## Motivation and Context

Will enable customers to enrich event data by CIDR

## How Has This Been Tested?

Local dev environment 
## Types of changes

## Checklist:

- [in progress] I have updated the documentation accordingly.

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
+++ b/graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
@@ -37,10 +37,13 @@ import org.graylog2.plugin.lookup.LookupDataAdapter;
 import org.graylog2.plugin.lookup.LookupDataAdapterConfiguration;
 import org.graylog2.plugin.lookup.LookupResult;
 import org.graylog2.plugin.utilities.FileInfo;
+import org.graylog2.utilities.IpSubnet;
+import org.graylog2.utilities.ReservedIpChecker;
 import org.joda.time.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotEmpty;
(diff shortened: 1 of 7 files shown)


kingzacko1:
@simonychuang @zeeklop the conversion from JSX to functional TSX components was fairly straightforward, but if you guys wouldn't mind checking it out pretty closely I would appreciate it!

graylog2-server/src/main/java/org/graylog2/lookup/adapters/CSVFileDataAdapter.java
```diff
@@ -245,6 +252,27 @@ public LookupResult doGet(Object key) {
...
+            Optional<IpSubnet> subnet = ReservedIpChecker.stringToSubnet(range);
+            if (subnet.isEmpty()) {
+                LOG.warn("CIDR range '{}' in data adapter '{}' is not a valid subnet, skipping this key in lookup.", entry, name);
+            } else {
+                try {
+                    if (subnet.get().contains(String.valueOf(ip))) {
+                        return LookupResult.single(entry.getValue());
+                    }
+                } catch (UnknownHostException e) {
+                    LOG.warn("Attempted to do a CIDR range lookup on invalid IP '{}'", ip);
```
bernd, line 267: We should consider changing the log level for both statements to DEBUG because WARN can result in **lots** of server log messages. :smile: Misconfigured lookups or unexpected field values can become a problem with high message rates.

kingzacko1: Good call! Have dropped level to debug in latest commit

ryan-carroll-graylog (approved):
Looks good and tests successfully with old and new csv adapters!

I can't remember if this was already discussed, but this opens up the possibility of the same "key" mapping to multiple values if cidr ranges overlap. I'm not sure there's a good way to validate for that without a lot of nested looping, but it maybe worth documenting if we're not already?

danotorrey:
I got a bit delayed today troubleshooting an issue preventing the frontend from running on my machine, but I should be able to solve that in the morning and review/test out this PR then!

danotorrey (approved):
LGTM and tests successfully. Very well done @kingzacko1!

PR #15018, opened 2023-03-23
Using database time for node heartbeats.

dennisoelkers:
## Motivation and Context

This PR is a refactoring of the queries used for:

  - registering new nodes
  - sending the heartbeat/marking nodes as alive
  - finding alive/dead nodes

The improvement is that the DB time is used for the timestamps instead of the node's clock. This is achieved by using `$currentDate` for registering/updating a node entry and a `$where` JS fragment using `new Date()` for finding dead/alive nodes.

The benefit of this is that there is no need for a very exact clock synchronization of the noeds anymore. In the past, if a node clock was late by more than 2 seconds (or the non-default ping timeout), then a node would always be marked as dead.
## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
@@ -18,6 +18,7 @@ package org.graylog2.cluster;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.bson.types.BSONTimestamp;
 import org.bson.types.ObjectId;
 import org.graylog2.database.DbEntity;
 import org.graylog2.database.PersistedImpl;
@@ -58,7 +59,11 @@ public class NodeImpl extends PersistedImpl implements Node {
 
     @Override
     public DateTime getLastSeen() {
-        return new DateTime(((Integer) fields.getOrDefault("last_seen", 0)) * 1000L, DateTimeZone.UTC);
+        final Object rawLastSeen = fields.get("last_seen");
+        if (rawLastSeen instanceof BSONTimestamp) {
+            return new DateTime(((BSONTimestamp) rawLastSeen).getTime() * 1000L, DateTimeZone.UTC);
+        }
+        return new DateTime(((Integer) rawLastSeen) * 1000L, DateTimeZone.UTC);
     }
 
     @Override
--- a/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
@@ -60,6 +60,9 @@ public class NodeImpl extends PersistedImpl implements Node {
     @Override
     public DateTime getLastSeen() {
         final Object rawLastSeen = fields.get("last_seen");
+        if (rawLastSeen == null) {
+            throw new IllegalStateException("Last seen timestamp of node is unexpectedly null!");
+        }
         if (rawLastSeen instanceof BSONTimestamp) {
             return new DateTime(((BSONTimestamp) rawLastSeen).getTime() * 1000L, DateTimeZone.UTC);
         }
--- a/graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
@@ -94,8 +94,6 @@ public class NodeServiceImpl extends PersistedServiceImpl implements NodeService
 
     @Override
     public Map<String, Node> allActive(Node.Type type) {
-        Map<String, Node> nodes = Maps.newHashMap();
-
         final BasicDBObject query = new BasicDBObject(Map.of(
                 "$and", List.of(
                         Map.of("type", type.toString()),
--- a/graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
@@ -164,7 +164,7 @@ public class NodeServiceImpl extends PersistedServiceImpl implements NodeService
                 )
         ));
 
-        return query(NodeImpl.class, query).isEmpty();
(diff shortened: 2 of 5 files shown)


graylog2-server/src/main/java/org/graylog2/cluster/NodeImpl.java
```diff
@@ -58,7 +59,11 @@ public String getTransportAddress() {
 
     @Override
     public DateTime getLastSeen() {
-        return new DateTime(((Integer) fields.getOrDefault("last_seen", 0)) * 1000L, DateTimeZone.UTC);
+        final Object rawLastSeen = fields.get("last_seen");
+        if (rawLastSeen instanceof BSONTimestamp) {
+            return new DateTime(((BSONTimestamp) rawLastSeen).getTime() * 1000L, DateTimeZone.UTC);
+        }
+        return new DateTime(((Integer) rawLastSeen) * 1000L, DateTimeZone.UTC);
```
luk-kaminski, line 66: After eliminating "getOrDefault" call, shouldn't you add a null-check here?

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
```diff
@@ -51,29 +54,25 @@ public Node.Type type() {
...
+        final Map<String, Object> fields = Map.of(
+                "$set", Map.of(
+                        "node_id", nodeId,
+                        "type", type().toString(),
+                        "is_leader", isLeader,
+                        "transport_address", httpPublishUri.toString(),
+                        "hostname", hostname
+                ),
+                "$currentDate", lastSeenField
+        );
```
luk-kaminski, line 67: I'd consider using` com.mongodb.client.model.Updates`, namely its set() and currentTimestamp() methods to make the code cleaner, get rid of special mongo-magic strings and `lastSeenField` constant.

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
```diff
@@ -153,24 +147,38 @@ public void markAsAlive(NodeId node, boolean isLeader, URI restTransportAddress)
...
+                "$and", List.of(
+                        recentHeartbeat(),
+                        Map.of(
+                                "type", Node.Type.SERVER.toString(),
+                                "is_leader", true
+                        )
+                )
+        ));
+
+        return !query(NodeImpl.class, query).isEmpty();
```
luk-kaminski, line 182: If we only care it the result is empty or not, maybe we should replace `query()` with `count()`, to not fetch unnecessary data?

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
```diff
@@ -153,24 +147,38 @@ public void markAsAlive(NodeId node, boolean isLeader, URI restTransportAddress)
...
+                        recentHeartbeat(),
+                        Map.of(
+                                "type", Node.Type.SERVER.toString(),
+                                "node_id", new BasicDBObject("$ne", nodeId.getNodeId()),
+                                "is_leader", true
+                        )
+                )
+        ));
+
+        return query(NodeImpl.class, query).isEmpty();
```
luk-kaminski, line 167: If we only care it the result is empty or not, maybe we should replace `query()` with `count()`, to not fetch unnecessary data?

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
```diff
@@ -51,29 +54,25 @@ public Node.Type type() {
...
+                "$currentDate", lastSeenField
+        );
+
+        final WriteResult result = this.collection(NodeImpl.class).update(
+                new BasicDBObject("node_id", nodeId),
+                new BasicDBObject(fields),
+                true,
+                false
+        );
+        return result.getN() == 1;
```
luk-kaminski, line 75: As far as I can see, we ignore the returned value in all the usages of `registerServer` method.
The question is - is it ok? 
If not, then maybe we should throw an exception when registration was not successful?
Maybe this method should no return boolean anyway?

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
```diff
@@ -153,24 +147,38 @@ public void markAsAlive(NodeId node, boolean isLeader, URI restTransportAddress)
...
     @Override
     public boolean isAnyLeaderPresent() {
-        BasicDBObject query = new BasicDBObject();
-        query.put("type", Node.Type.SERVER.toString());
-        query.put("last_seen", new BasicDBObject("$gte", Tools.getUTCTimestamp() - pingTimeout));
-        query.put("is_leader", true);
-
-        return query(NodeImpl.class, query).size() > 0;
+        final BasicDBObject query = new BasicDBObject(Map.of(
+                "$and", List.of(
```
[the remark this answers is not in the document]
dennisoelkers: I would love to, but the `Filters`/`Updates` query builders generate `Bson`, while our methods expect `DBObject`s, so I was considering this to be part of a future, more fundamental refactoring of the `NodeService` implementation.

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
dennisoelkers: :+1:

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
dennisoelkers: :+1:

graylog2-server/src/main/java/org/graylog2/cluster/NodeServiceImpl.java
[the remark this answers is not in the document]
luk-kaminski: Yeah, I have not spotted that, sorry!
It probably invalidates many of my comments.
Refactoring of `NodeService` will require... refactoring of `PersistedService` ... :(

todvora (approved):

luk-kaminski (approved):
Looks good, thx!

Issue #14735, opened 2023-02-21
Redesign Graylog configuration page to have a sidebar menu.


ousmaneo:
## What?

The current Graylog configuration page shows the configuration options on a single-page grid, forcing users to search for a configuration option.

We propose redesigning the Graylog configuration page with a sidebar menu navigation similar to the search page sidebar. 

This will allow showing a single config information and easy navigation between the configuration option.

Update:
After a discussion with @dennisoelkers  the suggestion is to turn the page into a more linear form with sections contributed by each component, but not as a grid, with separate boxes and individual edit buttons/forms (see user Edit form as an example).
It should be one big form where settings are editable in place. A sidebar would also be great to show the general structure of the page and be able to jump to individual sections.
We are fetching each configuration for each section on the backend side. Instead, we should bundle up the configuration keys and retrieve the complete configuration in one go.

grotlue:
I had a look at the current situation and the suggestions mentioned in this and I think the best solution would be:

- Have a sidebar
  - so the user can easily see what they can configure
  - so the user can find the section/option they want to configure fast
- Show only the active section of the sidebar as a whole form so the user can edit the options directly (ne extra click needed)
- With this layout that there is no need for a new API request fetching all configuration data which would mean we would fetch a lot of data we don't need
- We can either save the available sections in the frontend and fetch data for each section or have one call just returning available configuration sections and another returning the form for each section

Please find more details on what advantages/disadvantages I discovered for each option as well as user scenarios for this page in this overview:
[canvas_configurations-page-230310_1135.pdf](https://github.com/Graylog2/graylog2-server/files/10941594/canvas_configurations-page-230310_1135.pdf)

What do you think @kroepke @dennisoelkers @ousmaneo?

What would be the advantages in having one big form?

grotlue:
I had a first look and tried adding a sidebar. This is the result and what I would go with. There is a sidebar with entries for each section and for plugins, we can have a sub-sidebar with entries for each plugin.
On top of that, I'd remove the modal functionality, meaning instead of showing the configuration data and an edit button which opens a modal we show the configuration data as a form which can be edited right away.



https://user-images.githubusercontent.com/3169354/224938784-8307a74a-2f10-4d71-993b-8f400bccff55.mp4

ousmaneo:
> 
> What would be the advantages in having one big form?

Hi @grotlue, My first thought was to have something similar to what you have in the screenshot above.
[...]

PR #15025, opened 2023-03-23
Configuration Form Makeover

grotlue:
Part of: https://github.com/Graylog2/graylog2-server/issues/14735

## Description
This PR restructures the configuration page, migrates related components and improves its performance.

- Shows configurations in sections adding a sidebar for navigation
- For each section, the logic for fetching the configuration was moved into it's component, making sure we only fetch the data when the section is active
  - As a side effect, these components are now independent, and can be easily reused anywhere else
- Migrates configurations components to typescript
- Unifies headlines of configuration components
- Cleans up the configuration page component
- Adds enum types for configuration keys
- The plugin section has an own subnav, the content was moved into an own component

### Remarks
This PR does not include everything that could have been improved on this page. We could further improve the design, as well as restructure the plugin configurations in the same way to only fetch the data when the plugin section is active.
However, this would blow this PR out of proportion and due to time constraints I'd suggest to cut the scope here. 
We can still add further improvements afterwards.

## Motivation and Context
Up until now the configuration page was filled with content, making it hard to navigate. Additionally we would always make one call for each configuration section.
The assumption is that the main use case for this page is to change a specific configuration, thus it makes sense to structure it into sections which can easily be found through the nav.

Screenshot 2023-03-23 at 17 32 37
Screenshot 2023-03-23 at 17 32 46

## Types of changes

## Checklist:


gally47 (changes requested):
To keep it consistent with the other configs can we rename the buttons to `Cancel` and `Update configuration`

Screenshot 2023-04-05 at 16 59 20

graylog2-web-interface/src/pages/ConfigurationsPage.tsx
```diff
@@ -15,187 +15,163 @@
...
+
+      <ConfigletRow className="content">
+        <Col md={2}>
+          <Nav bsStyle="pills" stacked activeKey={activeSectionKey} onSelect={handleNavSelect}>
+            {configurationSections.map(({ hide, name, showCaret }, index) => (
+              !hide && (
+              <NavItem key={`nav-${name}`} eventKey={index + 1} title={name}>
+                {name}
+                {showCaret && (isSectionActive(name)
+                  ? <Icon name="caret-right" style={{ marginLeft: 5, verticalAlign: 'middle' }} />
```
gally47, line 161: Since we agreed on avoiding inline styling, can we convert these to styled components?

gally47 (changes requested):
Tested it. Looking good! just one more thing

gally47 (approved):
LGTM!

PR #15041, opened 2023-03-24
Suggestions are now based on DbEntityCatalog. Title as default column…

luk-kaminski:
… for suggestions.

## Description
After introduction of DbEntitiesCatalog, hardcoded map of permissions can be replaced with that catalog.
/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/database/suggestions/MongoEntitySuggestionService.java
+++ b/graylog2-server/src/main/java/org/graylog2/database/suggestions/MongoEntitySuggestionService.java
@@ -26,6 +26,7 @@ import org.apache.shiro.subject.Subject;
 import org.bson.BsonDocument;
 import org.bson.Document;
 import org.bson.conversions.Bson;
+import org.graylog2.database.DbEntity;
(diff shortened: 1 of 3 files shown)


graylog2-server/src/main/java/org/graylog2/database/suggestions/MongoEntitySuggestionService.java
```diff
@@ -101,11 +98,15 @@ public boolean hasAllPermission(final Subject subject) {
 
     public boolean hasReadPermissionForWholeCollection(final Subject subject,
                                                        final String collection) {
-        final String permission = mongoCollectionToPermission.get(collection);
-        if (permission == null) {
+        final DbEntityCatalogEntry dbEntityCatalogEntry = catalog.getByCollectionName(collection).orElse(null);
```
todvora, line 101: I think you could rewrite this with a several map calls and falls as default, without ifs and null checks?

todvora (approved):
LGTM, just a small detail if you want to make the code slightly more readable.

todvora (approved):
LGTM, thanks!

PR #15088, opened 2023-03-29
Run Container Matrix Tests with a given Version on the Cmdline

janheise:
## Description

Adds property `graylog.matrix.tests.search.version.override` so you can run the matrix tests with `mvn verify -Dgraylog.matrix.tests.version.override=OpenSearch:2.6.0`  for a given version.

It excludes all incompatible tests (e.g. if you test for OpenSearch2, it removes all tests for Elasticsearch7 and OpenSearch1.3) and also reduces the matrix to only include the version set on the commandline.

## Motivation and Context

Can be used to test if new OpenSearch versions that are released are still compatible without having to change our code to include the new version.

Can be used to split the IT tests into several runs to reduce runtime

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/test/java/org/graylog/testing/containermatrix/ContainerMatrixTestEngine.java
+++ b/graylog2-server/src/test/java/org/graylog/testing/containermatrix/ContainerMatrixTestEngine.java
@@ -100,29 +100,20 @@ public class ContainerMatrixTestEngine extends ContainerMatrixHierarchicalTestEn
     }
 
     /**
-     * Property has to be in the form of Elasticsearch|OpenSearch-x.x.x
+     * Property has to be in the form of Elasticsearch|OpenSearch:x.x.x
      *
      * @return
      */
-    private static String getSearchVersionOverride() {
-        return System.getProperty("graylog.matrix.tests.search.version.override");
-    }
-
-    public static boolean searchVersionOverrideExists() {
-        return getSearchVersionOverride() != null;
-    }
-
-    private SearchVersion getSearchVersionFromProperty() {
-        final var split = getSearchVersionOverride().split("-");
-        final var distro = SearchVersion.Distribution.valueOfLabel(split[0]);
-        if(SearchVersion.Distribution.ELASTICSEARCH.equals(distro)) {
-            return SearchVersion.elasticsearch(split[1]);
+    public static Optional<SearchVersion> getSearchVersionOverride() {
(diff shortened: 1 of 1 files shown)


graylog2-server/src/test/java/org/graylog/testing/containermatrix/ContainerMatrixTestEngine.java
```diff
@@ -98,12 +99,57 @@ private Set<Class<? extends PluginJarsProvider>> getPluginJarsProvider(Set<Class
...
     }
 
-    private Set<SearchServer> getSearchServerVersions(Set<Class<?>> annotatedClasses) {
+    /**
+     * Property has to be in the form of Elasticsearch|OpenSearch-x.x.x
+     *
+     * @return
+     */
+    private static String getVersionOverride() {
+        return System.getProperty("graylog.matrix.tests.version.override");
```
bernd, line 108: How about making the property name "search version" specific? If we want to do the same for MongoDB at some point, we would have a naming problem. :smile: 

```suggestion
        return System.getProperty("graylog.matrix.tests.search.version.override");
```

graylog2-server/src/main/java/org/graylog2/storage/SearchVersion.java
```diff
@@ -43,6 +43,15 @@ public enum Distribution {
         public String toString() {
             return this.printName;
         }
+
+        public static Distribution valueOfLabel(String label) {
```
todvora, line 47: This part seems very similar to SearchVersion#decode. Could we reuse it here?

graylog2-server/src/test/java/org/graylog/testing/containermatrix/ContainerMatrixTestEngine.java
```diff
@@ -98,12 +99,57 @@ private Set<Class<? extends PluginJarsProvider>> getPluginJarsProvider(Set<Class
...
     }
 
-    private Set<SearchServer> getSearchServerVersions(Set<Class<?>> annotatedClasses) {
+    /**
+     * Property has to be in the form of Elasticsearch|OpenSearch-x.x.x
+     *
+     * @return
+     */
+    private static String getSearchVersionOverride() {
+        return System.getProperty("graylog.matrix.tests.search.version.override");
```
todvora, line 108: What about doing all the splitting, parsing and comparisions here and return Optional<SearchVersion> that could be then used elsewhere?

todvora (approved):
LGTM

PR #15097, opened 2023-03-29
Move index creation to constructor of AccessTokenServiceImpl

### LINKED ISSUE #15012: Access token mongodb service creates index excessively
## Expected Behavior
The database index should only be created once, to ensure there's a unique index on the collection.

## Current Behavior
Every time a token is a saved, including when its access time is updated, the index gets recreated.
This leads to at least an additional table lock, and extra work on index creation in the background. On systems with lots of sidecars this creates performance problems.

## Possible Solution
In https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/security/AccessTokenServiceImpl.java#L142 move the `createIndex` call to the constructor.

Code pointer above.

## Context
Each access token use will save it to the database, leading to excessive queries.

## Your Environment

* Graylog Version: reported in 4.3 but applicable to all versions

patrickmann:
Resolves #15012 

Avoid performance issues by creating the index only once. See linked issue for details.

### DIFF
--- a/changelog/unreleased/issue-15012.toml
+++ b/changelog/unreleased/issue-15012.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Improve performance by avoiding unnecessary attempts to create index for access tokens."
+
+issues = ["15012"]
+pulls = [""]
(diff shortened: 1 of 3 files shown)


moesterheld (changes requested):
Looks good to me. 
But please remove `src/test/resources/org/graylog2/security/accessTokensMultipleIdenticalTokens.json` as this isn't used anymore.

moesterheld (approved):
thank you

PR #12911, opened 2022-06-17
Updating to `react-router` v6.

dennisoelkers:
## Motivation and Context

This PR is preparing the update to `react-router` v6. Notable changes
include:

  - `Switch` has become `Routes`
  - Only `Route` components are allowed below `Routes`
  - `Redirect` access should be avoided, instead we can use `useNavigation`
  - `history` access should be avoided, instead we can use `useHistory`/`withHistory`
  - Turning browser router into data router
  
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4398,Graylog2/graylog-plugin-enterprise-integrations#1034,Graylog2/graylog-plugin-integrations#1329
/nocl

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #15098, opened 2023-03-29
fix inputs ui route in cloud

AntonEbel:
Fix for inputs route in cloud environment. The #12911 PR disabled the enabled route again.  

/nocl


kodjo-anipah (approved):
LGTM

PR #15111, opened 2023-03-31
Cache last access time of authentication tokens

### LINKED ISSUE #15013: Make updating access token last access time optional
## What?
Access tokens get updated with their last update time on every use.
For busy systems, especially with systems having many sidecars, this can lead to excessive write load on the database.

Either limit the frequency the https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/security/realm/AccessTokenAuthenticator.java#L76 touches the access token (e.g. through an in-memory cache) or make the behavior optional for all or some access tokens.

## Why?
The access time is intended to inform users about whether an access token is in current use or has gone stale. So updating it on every use is beneficial, but when it's being used by sidecars, then depending on the update intervals and number of them, maintaining the access time can be very costly.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

patrickmann:
Resolves #15013 
Introduce a cache for last access time. Items expire after a set duration, ensuring that they stay reasonably fresh without being refreshed more often than is useful.

See linked issue for details.

### DIFF
--- a/changelog/unreleased/issue-15013.toml
+++ b/changelog/unreleased/issue-15013.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Avoid excessive DB requests for maintaining token last access time."
+
+issues = ["15013"]
+pulls = [""]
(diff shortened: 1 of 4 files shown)


kodjo-anipah (approved):
LGTM, tested locally

PR #15131, opened 2023-04-03
Add regex matching to remove_field function

### LINKED ISSUE #5653: Allow remove_field() function to take a regular expression or glob
The `remove_field()` pipeline function currently only allows to remove fields that match an exact name. This works if I know which fields my message will have, but not for more dynamic data, like in DNS responses that can contain multiple answers to a request.

## Expected Behavior

I'd like to be able to remove all fields of a message that match a regular expression or glob pattern.

## Current Behavior

Currently, I can only remove fields if I know the name of each field.

## Context

2019-02-07-121406_201x988_scrot

In this example, I'd like to call something like `remove_fields('dns_authorities_*')`.

patrickmann:
Resolves #5653 

`remove_field` performs regex match on the provided argument, so you can supply a field name or a regex pattern. 
Any field matching the pattern will be removed.


kodjo-anipah (approved):
LGTM

PR #15143, opened 2023-04-04
Add pipeline function lookup_has_value

### LINKED ISSUE #9173: Lookup Table Pipeline Function `lookup_has_value`
## What?
To check if a value is contained in a lookup data adapter we need a function which returns a boolean, so we can use it in the `when` block of a pipeline rule 

## Why?
If we want to route a message based on a lookup table we have to add the "All messages" stream as a default value:

```
rule "route table"
when
  has_field("source")
then
  let streamToRoute = lookup_value(
    lookup_table: "test-stream-lookup",
    key: $message.source,
    default: "All messages"
  );
  route_to_stream(
    name: to_string(streamToRoute),
    remove_from_default: streamToRoute != "All messages" // Prevent to remove a message from "All messages" if we route to "All messages"
  );
end
```

with a `lookup_has_value` function we could simplify this pipeline rule to:
```
rule "route table"
when
  has_field("source") && lookup_has_value("test-stream-lookup", $message.source)
then
  let streamToRoute = lookup_value(
    lookup_table: "test-stream-lookup",
    key: $message.source,
  );
  route_to_stream(
    name: to_string(streamToRoute),
    remove_from_default: true // or false
  );
end
```

Which would be far more readable

## Your Environment
* Graylog Version: 4.0.0 pre beta

patrickmann:
Resolves #9173 

`lookup_has_value(<table>,<key>)` will return `true`, if `table` contains `key`; otherwise `false`.


moesterheld (approved):
LGTM

I added a test which I used while reviewing. Please take a look at it before merging.

PR #15158, opened 2023-04-05
Fix OpenSearch permissions in data node tarball

bernd:
We tried to use fileSet/fileMode to make the OpenSearch config permissions work when running OpenSearch with a different user.

That didn't work because the fileMode changed all files and removed the executable bit from all executable scripts.

Now we use the Maven antrun plugin to set the permissions on the OpenSearch files right after extraction.


todvora (approved):
The permissions look correct and executables are still executable. Thanks!

PR #15163, opened 2023-04-06
Refactor ProxiedResource and fix custom Timeouts

mpfz0r:
Timeouts longer than 10s did not work when using the RemoteInterfaceProvider. This is because the timeout needs to be passed to the OKHttpClient as well.

- Inline the createRemoteInterfaceProvider method into all ProxiedResource methods. This allows us to use the timeout value for both, the RemoteInterfaceProvider and the RetroFit timeouts. As a side effect, this makes the method signatures smaller and easier to use.

- Change a few deprecated getForAllNodes to requestOnAllNodes. Provide a stripCallResult() method that makes that easier.

- Change requestOnMaster to requestOnLeader

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/4979
/nocl


mpfz0r:
Ignore the failing frontend tests :shrug:

PR #15192, opened 2023-04-12
Setting up Preflight UI component structure.

dennisoelkers:
## Motivation and Context

This PR is setting up the component structure for the preflight UI, including theming, component library (mantine) and basic bootstrapping logic (`App` component et al).

/nocl

## Types of changes

## Checklist:


graylog2-web-interface/src/preflight/theme/ThemeWrapper.tsx
```diff
@@ -0,0 +1,103 @@
...
+          fontSize: theme.fonts.size.h6.rem,
+        },
+      },
+    },
+    spacing: {
+      xs: theme.spacings.px.xs,
+      sm: theme.spacings.px.sm,
+      lg: theme.spacings.px.lg,
+      xl: theme.spacings.px.xl,
+    },
```
linuspahl, line 68: It looks like this version of mantine now expects the spacings to have a unit.

Instead we can use:
```
xs: theme.spacings.xs,
sm: theme.spacings.sm,
lg: theme.spacings.lg,
xl: theme.spacings.xl,
```

linuspahl: I am not sure why we do not see a TS error here.
Having a look at the type definition, TS expects the object values to be a string:

image

But we actually define a number:
image

PR #15183, opened 2023-04-11
Fixed some issues with OshiOsProbe

panxt:
## Motivation and Context
1.Fixed when using OSHI library to get cpu load is always 0.
2.Fixed getting CPU frequency could be negative because CPU frequency could exceed Int max value.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #15195, opened 2023-04-12
Fixed calculation and expression of CPU load in OshiProbe to be consistent in OsStats

mpfz0r:
## Motivation and Context
Fixed calculation and expression of CPU load in OshiProbe to be consistent in OsStats.
 It's related to https://github.com/Graylog2/graylog2-server/pull/15183

## Types of changes

## Checklist:


panxt:
> I am not a fan of adding a second of sleep to call these APIs, but since we don't call these from anywhere automatically, it's okay for now.
> 
> We should probably remove these metrics at some point because we don't really use them. There are better tools to get system metrics.
Thanks for the correction, looking forward to new tools.

PR #15196, opened 2023-04-12
Merge data-node and data-node-distribution modules

bernd:
The data-node-distribution module only contained the data-node assembly. There is currently no reason to have two modules.

Refs https://github.com/Graylog2/graylog-project-internal/pull/95

/nocl


data-node/pom.xml
```diff
@@ -611,6 +621,31 @@
...
+                    </execution>
+                </executions>
+                <configuration>
+                    <attach>true</attach>
+                    <appendAssemblyId>true</appendAssemblyId>
+                    <descriptors>
+                        <descriptor>src/main/assembly/datanode.xml</descriptor>
+                    </descriptors>
+                    <!-- to make it easier to collect assemblies, we put them into the _parent's_ build directory -->
+                    <!-- FIXME: Use a proper output directory. parent.parent.build.directory is a hack… -->
```
janheise, line 643: Is this comment still valid? I can't see the usage of `parent.parent.build.directory`?

bernd: Good point. I will remove the comment. :+1:

PR #15222, opened 2023-04-14
feat: system notification on sidecar inactive

### LINKED ISSUE #6697: create event if sidecar gets inactive / active
## Expected Behavior
When a sidecar agent is reporting in or stopped reporting it's present graylog should write that as event that the alerting can act on the presence or missing of sidecars.

## Context
Some environments need to get notified when sidecars report in or are taken offline. Mostly in Multi-Team/Multi-Timezone environments where one team is responsible for the logging while other for the operation.

## Your Environment

* Graylog Version: 3.1

[z#842863]

moesterheld:
create a message in the system notification stream when a sidecar becomes inactive

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/5627

## Description
when a sidecar gets the status `UNKNOWN`, a message is generated in the system notification stream.
this is disabled by default and can be enabled by setting the configuration value `system_event_excluded_types=`

## Motivation and Context
resolves #6697 

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/Configuration.java
```diff
@@ -210,6 +212,9 @@ public class Configuration extends BaseConfiguration {
     @Parameter(value = "lock_service_lock_ttl", converter = JavaDurationConverter.class)
     private java.time.Duration lockServiceLockTTL = MongoLockService.MIN_LOCK_TTL;
 
+    @Parameter(value = "system_event_excluded_types", converter = TrimmedStringSetConverter.class)
+    private Set<String> systemEventExcludedTypes = Sets.newHashSet(Notification.Type.SIDECAR_STATUS_UNKNOWN.name());
```
mpfz0r, line 216: we usually document these settings and their defaults in  `misc/graylog.conf`

graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/SidecarService.java
```diff
@@ -211,6 +218,20 @@ public int markExpired(Period period, String message) {
                                     .nodeDetails(nodeDetailsToSave)
                                     .build();
                             save(toSave);
+
+                            Notification notification = notificationService.buildNow();
```
mpfz0r, line 222: this could be extracted into a separate method to increase readability

mpfz0r (changes requested):
Left a few comments. 
I wonder if we can find a better name and wording for "status unknown"
Sounds very abstract.  Maybe something like "Lost contact to sidecar" or the likes? :thinking:

moesterheld:
> Left a few comments. I wonder if we can find a better name and wording for "status unknown" Sounds very abstract. Maybe something like "Lost contact to sidecar" or the likes? 🤔

"Unknown" is also the status shown for the sidecars in the UI, so I think it is the one the user can relate to the best. 

There would also be a Status FAILING (currently unused), which also isn't really correct I would say. The method setting the state to UNKNOWN is called markExpired.

misc/graylog.conf
```diff
@@ -706,3 +706,7 @@ mongodb_max_connections = 1000
 # Use with caution - skipping failing migrations may result in an inconsistent DB state.
 # Default: false
 #ignore_migration_failures = false
+
+# Comma-separated list of notifcation types which should not emit a system event.
+# Default: SIDECAR_STATUS_UNKNOWN which will create a new type whenever the status of a sidecar becomes "Unknown"
```
mpfz0r, line 711:
```suggestion
# Default: SIDECAR_STATUS_UNKNOWN which will create a new event whenever the status of a sidecar becomes "Unknown"
```

misc/graylog.conf
```diff
@@ -706,3 +706,7 @@ mongodb_max_connections = 1000
 # Use with caution - skipping failing migrations may result in an inconsistent DB state.
 # Default: false
 #ignore_migration_failures = false
+
+# Comma-separated list of notifcation types which should not emit a system event.
+# Default: SIDECAR_STATUS_UNKNOWN which will create a new event whenever the status of a sidecar becomes "Unknown"
```
mpfz0r, line 711:
```suggestion
# Default: SIDECAR_STATUS_UNKNOWN which would create a new event whenever the status of a sidecar becomes "Unknown"
```

Sorry, you clicked faster then I was able to correct my suggestion  :-D

mpfz0r (approved):
LGTM. 
Since we are only merging this after 5.1 we might get the chance to improve the system events system to make this a bit less hacky

mpfz0r:
build currently fails with `[2023-08-16T08:58:29.022Z] [ERROR] /home/jenkins/workspace/apshots_graylog2-server_PR-15222/.repos/graylog-plugin-enterprise/enterprise/src/main/java/org/graylog/plugins/cloud/command/NullNotificationServiceImpl.java:[23,7] error: NullNotificationServiceImpl is not abstract and does not override abstract method createSystemEvent(Notification) in NotificationService`

mpfz0r:
@moesterheld sorry this took so long :sweat_smile:  feel free to merge when it's green

PR #15228, opened 2023-04-14
Add MongoJack 4

thll:
Adds the current version of [MongoJack](https://github.com/mongojack/mongojack) as a shaded dependency.

The intended usage is, for now, to use MongoJack to provide a codec for the MongoDB driver and only use the official MongoDB API in our code.

/nocl


janheise (approved):
works for me

PR #15242, opened 2023-04-17
The docker runtime detection stopped working with cgroups v2

mpfz0r:
Add other modes of detection

https://stackoverflow.com/questions/68816329/how-to-get-docker-container-id-from-within-the-container-with-cgroup-v2


bernd (approved):
Thanks! The change fixes the `SystemStatsIT` test on my machine that uses Docker with cgroups v2.

PR #15243, opened 2023-04-17
Use telemetry configuration to set connect-src value in CSP header

### LINKED ISSUE #15238: Update CSP to allow telemetry url using connect-src
## Expected Behavior
XHR requests made to the configured telemetry URL should be allowed in the Content Security Policy.

## Current Behavior
Currently, the browser will fall back to `default-src` and block the request.

> `.js:1 Refused to connect to 'https://telemetry.graylog.cloud/decide/?v=3&ip=1&_=1681721588616&ver=1.52.0' because it violates the following Content Security Policy directive: "default-src 'self'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.`

## Possible Solution
Using the configured telemetry URL (already available in the config), dynamically update the returned CSP header to allow XHR requests to it, using `connect-src`.

1. Go to a production instance.
2. Check the browser console for the CSP errors above.

## Your Environment

* Graylog Version:  5.1.0-SNAPSHOT+eacf8b2

patrickmann:
Resolves #15238
/nocl fixing an unreleased feature, which already has a CL

Telemetry needs to connect to an external URL. Dynamically modify the CSP header, when telemetry is enabled.

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/testing/fullbackend/FiltersIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/testing/fullbackend/FiltersIT.java
@@ -36,6 +36,7 @@ public class FiltersIT {
 
     @ContainerMatrixTest
     void cspDocumentationBrowser() {
+        String expected = CSP.CSP_SWAGGER + "connect-src https://telemetry.graylog.cloud;";
         given()
                 .spec(api.requestSpecification())
                 .when()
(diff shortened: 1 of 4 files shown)


ousmaneo (approved):
Tested and Works!

patrickmann:
Back to draft status: we also need to consider auth backends in the `connect-src` setting.

patrickmann:
Superseded by #15283

PR #15283, opened 2023-04-19
Introduce service to manage connect-src allow list in CSP

### LINKED ISSUE #15238: Update CSP to allow telemetry url using connect-src

patrickmann:
Resolves #15238

The `connect-src` value in CSP is populated dynamically from multiple sources:
- telemetry service
- authentication backends

Since auth backends are created and deleted on the fly, without a server restart, we need to update the CSP accordingly.
This PR introduces a new service for that purpose.

Tested locally with Okta and OpenLDAP authentication.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#5053

### DIFF
--- a/changelog/unreleased/issue-15238.toml
+++ b/changelog/unreleased/issue-15238.toml
@@ -1,0 +1,5 @@
+type = "c"
+message = "Include URLs for authentication service(s) in the CSP header connect-src."
+
+issues = ["15238"]
+pulls = ["15283", "Graylog2/graylog-plugin-enterprise#5053"]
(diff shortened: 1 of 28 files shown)


kodjo-anipah (approved):
LGTM

PR #15305, opened 2023-04-21
CSP listener for auth backend events

### LINKED ISSUE #15371: Reimplement generating `connect-src` directives for auth backends.
## What?

For 5.1, we reverted the [PR](https://github.com/Graylog2/graylog2-server/pull/15283) which introduced generating the `connect-src` sources for configured auth backends dynamically. We should reimplement it for 5.2 to further tighten down our CSP.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

patrickmann:
Resolves #15371

CSP `connect-src` should be updated on the fly, when a new auth backend is created. Since the existing `AuthServiceEventListener` only runs in Enterprise, we introduce a dedicated CSP event listener in core.

Note that the CSP-relevant event (when a backend is created or updated) is broadcast on the cluster event bus. This ensures that an auth backend can be created on one node, and a user can login on another, without requiring a server restart.

## How Tested
- locally with Okta and LDAP on 1 and 2 nodes
- HTTP notifications (doesn't need CSP entry)
- Lookup tables (Graynoise)
- Cloud environment 

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#5066


bernd (changes requested):
A few comments inline. But I found a problem as well:

I created an LDAP authentication service and changed the hostname of the LDAP server a few times. The problem is that the CSP header doesn't change. Only after a restart of the server can I see the changed CSP header value.

bernd:
@patrickmann What's the reason not to include the port number in the CSP `connect-src` values? Let's say the port of a URL is `8080`, shouldn't we include the port in that case as well? What happens if we don't include the port?

Another question: Should we even include the hosts for non-HTTP(s) services like LDAP? If not, we might want to use a more specific method name for `hostAllowList()` to better communicate the use case.

graylog2-server/src/main/java/org/graylog2/shared/rest/resources/csp/CSPEventListener.java
```diff
@@ -0,0 +1,49 @@
...
+        eventBus.register(this);
+    }
+
+    @Subscribe
+    public void handleBackendCreated(AuthServiceBackendSavedEvent event) {
+        try {
+            LOG.info("Updating CSP for authentication service <{}>", event.authServiceId());
+            cspService.buildConnectSrc();
+        } catch (Exception e) {
+            LOG.warn("Failed to update CSP for authentication service <{}>", event.authServiceId());
```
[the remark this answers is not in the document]
patrickmann: done

graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackendConfig.java
```diff
@@ -40,7 +41,7 @@ default void validate(ValidationResult result) {
     }
 
     @JsonIgnore
-    List<String> hostAllowList();
+    Optional<List<String>> hostAllowList();
```
bernd, line 44: As mentioned [here](https://github.com/Graylog2/graylog2-server/pull/15305#issuecomment-1519746783), we should give this method a more specific name to communicate the use case. Let's also add javadoc to the method to describe its use.

How about something like `externalHTTPHosts()`?

bernd (changes requested):
Two comments. Could you reproduce the issue I mentioned [here](https://github.com/Graylog2/graylog2-server/pull/15305#pullrequestreview-1395591534)

graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackendConfig.java
patrickmann: I like the name - refactored it.

bernd:
> Re LDAP host change not reflected in CSP: I couldn't repro. LDAP is now ignored so I tested again with AD. I see a host change reflected immediately in the CSP header.
> 

@patrickmann It looks like we have different behavior depending on the resource used. I set the URL for an Okta authentication service to `http://bootvalue`. That gives me the following CSP header:

```
$ http --verbose :9001/api/system | grep Content-Security
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; img-src 'self' data:;connect-src 'self' https://telemetry.graylog.cloud bootvalue

$ http --verbose :9001/config.js | grep Content-Security
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; img-src 'self' data:;connect-src 'self' https://telemetry.graylog.cloud bootvalue
```
That looks correct. Both resources return `bootvalue` for `connect-src`.

Now I change the Okta URL to `http://changedvalue`.

The DEBUG log shows:

```
2023-04-25 13:49:38,549 DEBUG: org.graylog2.shared.rest.resources.csp.CSPServiceImpl - Updated CSP: 'self' https://telemetry.graylog.cloud changedvalue
```

The two resources now return different values:

```
$ http --verbose :9001/api/system | grep Content-Security
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; img-src 'self' data:;connect-src 'self' https://telemetry.graylog.cloud changedvalue

$ http --verbose :9001/config.js | grep Content-Security
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; img-src 'self' data:;connect-src 'self' https://telemetry.graylog.cloud bootvalue
```

The `/api/system` endpoint returns the updated value, and the `/config.js` endpoint returns the previous value. The `/config.js` value only changes when I restart the server.

graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackendConfig.java
```diff
@@ -40,7 +41,7 @@ default void validate(ValidationResult result) {
     }
 
     @JsonIgnore
-    List<String> hostAllowList();
+    Optional<List<String>> externalHTTPHosts();
```
bernd, line 44: Let's add some javadoc here to help developers to know what this is used for. We can mention CSP as an example use case.

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackendConfig.java
```diff
@@ -104,8 +105,8 @@ public void validate(ValidationResult result) {
     }
 
     @Override
-    public List<String> hostAllowList() {
-        return servers().stream().map(HostAndPort::host).toList();
+    public Optional<List<String>> externalHTTPHosts() {
+        return Optional.of(servers().stream().map(HostAndPort::toString).toList());
```
bernd, line 109: Active Directory is just a slightly different LDAP implementation. So we should return `Optional.empty()` here.

bernd (changes requested):

graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackendConfig.java
patrickmann: done

graylog2-server/src/main/java/org/graylog/security/authservice/AuthServiceBackendConfig.java
patrickmann: done

bernd (approved):

PR #15318, opened 2023-04-24
Move AWS plugin into graylog2-server

thll:
Moves the AWS plugin from https://github.com/Graylog2/graylog-plugin-aws into `graylog2-server`.


kodjo-anipah (approved):
LGTM, tested locally cloud and none cloud setup. 
Launched inputs, and AWS Plugin configuration is still working for non cloud.

Issue #15320, opened 2023-04-24
Add server.conf configuration value for initial Field type refresh interval value


danotorrey:
https://github.com/Graylog2/graylog2-server/pull/13018 added system wide index set defaults functionality. As part of this change, it was intended that every index set default configuration setting can also be specified in the `server.conf` file, so that the values properly initialize on the first server startup. 

Initially, the `Field type refresh interval` was incorrectly mapped to the `index_field_type_periodical_full_refresh_interval` setting which has a different purpose. 

This was fixed in https://github.com/Graylog2/graylog2-server/pull/14842. However, the initial value is hard-coded in application code. I think we should add a `server.conf` property for this value, so it can be customized for newly initializations of Graylog server clusters.

If this is done, we should also update the `UPGRADING.md` documentation, which still references the incorrect property: 
https://github.com/Graylog2/graylog2-server/blob/a4509254592e29e542e8793bef466f95e12e8d32/UPGRADING.md#new-graylog-cluster-index-set-initialization-defaults
## Your Environment

* Graylog Version: 5.1.0-SNAPSHOT

danotorrey:
Note, I have added a PR for removing the incorrect property from `UPGRADING.md` for the 5.1 release (to eliminate the potential for confusion for referencing the wrong property). I suggest that we look at adding the corresponding configurable default cluster initialization value for `Field type refresh interval` to the `server.conf` in a future release, since we are so close to the 5.1 release date. Seems prudent to not change such foundational index set settings so close to a release.

PR #15321, opened 2023-04-24
Index set defaults: Remove incorrect Field type refresh interval property from upgrading instructions

danotorrey:
## Description

The `Field type refresh interval` index set default setting was incorrectly mapped to the `index_field_type_periodical_full_refresh_interval` `server.conf` value in the initial implementation of Global Index Set defaults (see https://github.com/Graylog2/graylog2-server/pull/13018). This mismapping was fixed in https://github.com/Graylog2/graylog2-server/pull/14842, where the defaults code was changed to reference the correct hard-coded default in `IndexSetConfig. DEFAULT_FIELD_TYPE_REFRESH_INTERVAL`. 

## Motivation and Context

There is currently no `server.conf` value for the `IndexSetConfig.DEFAULT_FIELD_TYPE_REFRESH_INTERVAL` constant. I think we are too late in the 5.1 release cycle to be adding new `server.conf` properties, so I have opened https://github.com/Graylog2/graylog2-server/issues/15320 to request adding that, and I am filing this PR to remove the incorrect property from the `UPGRADING.md` documentation for `5.1` so that it won't create confusion for users. I am guessing most users won't initialize index settings on first boot, and probably even fewer would be interested in customizing the refresh interval. So, probably good to at least remove the incorrect property from the `UPGRADING.md` docs to eliminate the potential for confusion.

/nocl

## How Has This Been Tested?

N/A

## Types of changes

- [x] Documentation fix

## Checklist:

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -18,7 +18,6 @@ values. Please see the sample [graylog.conf](https://github.com/Graylog2/graylog
 - `elasticsearch_replicas`
 - `disable_index_optimization`
 - `index_optimization_max_num_segments`
-- `index_field_type_periodical_full_refresh_interval`
 - `rotation_strategy`
 - `elasticsearch_max_docs_per_index`
 - `elasticsearch_max_size_per_index`


danotorrey:
I noticed this while working on 5.1 docs for Index Set Defaults.

danotorrey:
Also see comment in original issue for additional motivation for this change: https://github.com/Graylog2/graylog2-server/issues/15320#issuecomment-1520850009

ryan-carroll-graylog (approved):

kingzacko1 (approved):

PR #15329, opened 2023-04-25
Let cookie auth take precedence over basic auth.

### LINKED ISSUE #6831: HTTP authentication impossible in front of graylog

dennisoelkers:
## Motivation and Context

This change is making sure that cookie auth has precedence over basic auth during the authentication process of the REST API. This allows us to put a proxy in front of Graylog that does basic auth separately, without the need of user databases being in sync with Graylog's user database.

What this does not solve is API access through tokens as basic auth headers. This needs a separate fix and will most probably require an additional, separate header being user for the token.

Fixes #6831.

## How Has This Been Tested?

Caddy, configured with basic auth was used to reverse proxy the Graylog frontend/API. General functionality tests were performed with it.

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/pr-15329.toml
+++ b/changelog/unreleased/pr-15329.toml
@@ -1,0 +1,6 @@
+type = "f"
+message = "Let cookie auth take precedence over basic auth. "
+
+issues = ["6831"]
+pulls = ["15329"]
+
--- a/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/security/ShiroSecurityContextFilter.java
@@ -137,12 +137,12 @@ public class ShiroSecurityContextFilter implements ContainerRequestFilter {
         if ("token".equalsIgnoreCase(credential)) {
             return new AccessTokenAuthToken(userName, host);
         }
-        if (!Strings.isNullOrEmpty(userName) && !Strings.isNullOrEmpty(credential)) {
-            return new UsernamePasswordToken(userName, credential, host);
-        }
         if (cookies.containsKey(SESSION_COOKIE_NAME)) {
             final Cookie authenticationCookie = cookies.get(SESSION_COOKIE_NAME);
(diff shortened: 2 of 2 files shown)


mpfz0r (approved):
Works and LGTM

PR #15332, opened 2023-04-25
Fix wrong conversion of default timeout for proxied resources

thll:
There is a conversion error when reading the `proxied_requests_default_call_timeout` setting. For some calls, the actual timeout being used was higher by a factor of 1000.

I noticed this when looking at requests like the following:

```
http://localhost:8080/api/cluster/aaa02d2d-1c2c-426b-9dca-daa2a38c828a/metrics/namespace/org
```

/nocl


bernd (approved):
Good catch! Thanks! :pray:

mpfz0r (approved):
Ooops, thanks for catching this

Issue #15266, opened 2023-04-18
Ensure user identify is done before sending pageview Telemetry


ousmaneo:
Some `pageview` telemetry events do not have valid user info.

AntonEbel:
@ousmaneo @kroepke Has this issue been resolved by improving telemetry performance and can we close it?

AntonEbel:
Synced with Ousmane, issue can be closed.

PR #15342, opened 2023-04-25
telemetry performance improvement: save and use telemetry node data from db

AntonEbel:
This PR improves the performance when collecting node telemetry data and tries to fix the following issue: #15266 
/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/configuration/TelemetryConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/TelemetryConfiguration.java
@@ -72,7 +72,7 @@ public class TelemetryConfiguration {
         @Override
         public void validate(final String name, final Duration value) throws ValidationException {
             if (value != null && value.compareTo(Duration.ofMinutes(10)) < 0) {
-                throw new ValidationException("Parameter " + name + " should be at least 10 minute (found " + value + ")");
+                throw new ValidationException("Parameter " + name + " should be at least 10 minutes (found " + value + ")");
             }
         }
     }
--- a/graylog2-server/src/main/java/org/graylog2/configuration/TelemetryConfiguration.java
+++ b/graylog2-server/src/main/java/org/graylog2/configuration/TelemetryConfiguration.java
@@ -17,13 +17,18 @@
 package org.graylog2.configuration;
 
 import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.ValidationException;
+import com.github.joschi.jadconfig.Validator;
+import org.graylog2.configuration.converters.JavaDurationConverter;
 
 import javax.annotation.Nullable;
+import java.time.Duration;
 import java.util.Map;
 
 public class TelemetryConfiguration {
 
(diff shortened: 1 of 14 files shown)


ousmaneo (approved):
LGTM ! I tested It and it's getting the telemetry settings is faster  `59ms` against `144ms` before.

graylog2-server/src/main/java/org/graylog2/telemetry/cluster/TelemetryClusterInfoPeriodical.java
```diff
@@ -0,0 +1,124 @@
...
+import java.time.Duration;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
+
+import static org.graylog2.configuration.TelemetryConfiguration.TELEMETRY_CLUSTER_INFO_TTL;
+import static org.graylog2.telemetry.cluster.db.DBTelemetryClusterInfo.FIELD_NODE_ID;
+
+public class TelemetryClusterInfoPeriodical extends Periodical {
+
```
moesterheld, line 44: would it make sense to override `leaderOnly()` with `true`?

graylog2-server/src/main/java/org/graylog2/configuration/TelemetryConfiguration.java
```diff
@@ -48,11 +56,24 @@ public boolean isTelemetryEnabled() {
...
                 "host", getTelemetryApiHost(),
                 "enabled", isTelemetryEnabled()
         );
     }
+
+    public static class Minimum10MinuteValidator implements Validator<Duration> {
+        @Override
+        public void validate(final String name, final Duration value) throws ValidationException {
+            if (value != null && value.compareTo(Duration.ofMinutes(10)) < 0) {
+                throw new ValidationException("Parameter " + name + " should be at least 10 minute (found " + value + ")");
```
moesterheld, line 75:
```suggestion
                throw new ValidationException("Parameter " + name + " should be at least 10 minutes (found " + value + ")");
```

moesterheld (approved):
LGTM. Just two small comments.

graylog2-server/src/main/java/org/graylog2/telemetry/cluster/TelemetryClusterInfoPeriodical.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public int getInitialDelaySeconds() {
+        return 0;
+    }
+
+    @Override
+    public int getPeriodSeconds() {
+        return Ints.saturatedCast(telemetryClusterInfoTtl.minusMinutes(1).toSeconds());
+    }
```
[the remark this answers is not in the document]
AntonEbel: We create a mongodb ttl index depending on the telemetry_cluster_info_ttl configuration property. Before the document is removed due to the ttl index, live nodes update this document one minute earlier. For dead nodes, the document is removed because of the ttl index. The one minute is just to make sure that the node has enough time to update the document.

PR #15360, opened 2023-04-26
Fix padding for form input descriptions.

linuspahl:
## Motivation and Context

With https://github.com/Graylog2/graylog2-server/pull/14756 we changed the padding for input descriptions.
It looked good for forms with the class `form-horizontal`, but it looked bad in other cases.

For example for the index set form:
image

With this change we fix the padding for all cases:
image

/nocl


dennisoelkers (approved):
Looks great, works fine!

PR #15373, opened 2023-04-27
Use wildcard for CSP img-src

### LINKED ISSUE #15344: World Map Visualizatin Does Not Load Map Image Tiles - 5.1
## Expected Behavior
When viewing a World Map widget, the map images should load as you zoom in and out.

## Current Behavior
When viewing a World Map widget, the map images do not load.  The images are accessible via a direct call, however.

image

## Possible Solution
Cookie policy?

1. Load any previous World Map widget or create a new one.

## Context
Without the images, the World Map widget provides no value.

## Your Environment

* Graylog Version: 5.1-beta
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System: Windows
* Browser version: Firefox/Chrome/Edge

patrickmann:
/nocl change to an unreleased feature
Resolves #15344 

Use wildcard for the image source in CSP, while we implement a more configurable solution for the next release.


dennisoelkers (approved):
Looks good, fixes the issue! :pray:

PR #15385, opened 2023-04-28
Fix GELF output blocking and output processing race condition

### LINKED ISSUE #9917: Message processing gets blocked - thread dump included
## Expected Behavior

Message processing should never get blocked, there should be some kind of method for self healing, e.g. timeout. And if self healing does not work, there should be error handling which would at least log the problem.

## Current Behavior

In some cases message processing gets blocked, please see screenshot and thread dump below. No events in user interface or logs.

Last time this kind of behaviour was related to GELF TCP outputs #9064
This time the GELF TCP outputs are checked, tested and restarted and it should not be the cause this time. Also Graylog service and the server has been restarted.

Screenshot 2021-01-15 at 17 17 46 - output 0 copy

Thread dump of node 060bfd4c / redacted-graylog018.redacted Taken at Fri Jan 15 2021 17:17:55 GMT+0200

[thread-dump-2021-01-15.txt](https://github.com/Graylog2/graylog2-server/files/5821821/thread-dump-2021-01-15.txt)

## Your Environment

* Graylog Version: 3.3.8
* Java Version: openjdk version "1.8.0_272"

bernd:
Switch from GelfTransport#send() to GelfTransport#trySend() to avoid blocking output threads. We retry the send operation as long as the output is running.
This fixes an issue where removing an output from a stream or deleting it doesn't unblock processing. (with a slow output target)

A race condition in OutputRegistry was contributing to the problem.

When the OutputRegistry didn't have a cached output instance, we loaded the output configuration from the database and started the output. The problem is that each message holds a stream object that might be outdated because the output could have been removed since adding the stream object to the message object.

So instead of just starting an output, we need to check if the output is still assigned to the stream.

Fixes #9917
Fixes Graylog2/graylog-plugin-enterprise#4623

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
+++ b/graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
@@ -27,6 +27,7 @@ import com.google.common.collect.Sets;
 import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.google.common.util.concurrent.UncheckedExecutionException;
+import com.swrve.ratelimitedlogger.RateLimitedLog;
 import org.graylog2.database.NotFoundException;
 import org.graylog2.notifications.Notification;
 import org.graylog2.notifications.NotificationService;
@@ -46,6 +47,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import java.time.Duration;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Callable;
@@ -57,6 +59,10 @@ import java.util.stream.Collectors;
 @Singleton
 public class OutputRegistry {
     private static final Logger LOG = LoggerFactory.getLogger(OutputRegistry.class);
+    private static final Logger RATE_LIMITED_LOG = RateLimitedLog.withRateLimit(LOG)
+            .maxRate(1)
+            .every(Duration.ofSeconds(5))
+            .build();
 
     private final Cache<String, MessageOutput> runningMessageOutputs;
     private final MessageOutput defaultMessageOutput;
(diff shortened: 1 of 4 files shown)


thll (approved):
Looks good. Great find!

For testing, I reproduced the issue with the output inhibiting message throughput on connection issues even after it was unassigned from the stream. This went away after applying this PR.

graylog2-server/src/main/java/org/graylog2/outputs/OutputRegistry.java
```diff
@@ -179,8 +187,17 @@ public Callable<MessageOutput> loadForIdAndStream(final String id, final Stream
...
             @Override
             public MessageOutput call() throws Exception {
-                final Output output = outputService.load(id);
-                return launchOutput(output, stream);
+                // Check if the output is still assigned to the given stream before loading and starting it.
+                // The stream assignment of the output could have been removed while the message object went
+                // through processing and output buffer processing handling.
+                // Without this check, we would start the output again after it has been stopped by removing it
+                // from a stream.
+                final Stream dbStream = streamService.load(stream.getId());
```
luk-kaminski, line 195: I'm asking just for curiosity - have you checked if this additional communication with Mongo has some noticeable implications on performance?

bernd: @luk-kaminski The database query to load the stream and the existing one below to load the output from the database are only executed if the output cannot be found in the cache. That means the query adds a bit of overhead when loading the cache entry.

When the server runs into the described race condition and message objects still have the output in their streams list, we might have a spike in database requests because of the additional query executed per message. But that only happens until all message objects created before removing the output from the stream are processed.

So yes, in these situations, there can be a performance hit. The only alternative I see to avoid the performance hit is to maintain an in-memory cache of all streams and their outputs to avoid the database calls. That requires careful cache management, though.

luk-kaminski: Thanks for explanation, @bernd.
If performance decrease only happens in those rare situation, I vote with my both hands against any complications and for leaving the current solution! :)

PR #15400, opened 2023-05-02
Fix concurrency issue in SystemJobManager

todvora:
The SystemJobManager is checking and preventing unwanted concurrent tasks. But the implementation was not bulletproof and it was possible to run more tasks even if the concurrency was disabled for them.

## Description
The `jobs` map is a ConcurrentHashMap, but this is not enough to prevent two tasks starting in the same time. The concurrency check and putting a task into this map happens not atomically and any delay between these two operation would allow starting another task.

The easiest and reliable solution is to mark the whole `submitWithDelay` as _synchronized_. 

## How Has This Been Tested?
Addapted the existing concurrency test to actually start tasks concurrently, set to 100 repetitions of the test, verified before and after the _synchronized_ fix.

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/pr-15400.toml
+++ b/changelog/unreleased/pr-15400.toml
@@ -1,0 +1,4 @@
+type = "f"
+message = "Fixed possible unwanted concurrency in SystemJobManager"
+
+pulls = ["15400"]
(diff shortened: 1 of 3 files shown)


dennisoelkers (approved):
Looks good to me! :pray:

PR #15401, opened 2023-05-02
AMQP improvements: retry initial connection; add failure recorder, add password encryption

### LINKED ISSUE #12447: There are no retries when AMQP Input fails to start after Graylog host reboot
## Expected Behavior
Graylog should retry starting the AMQP input when it fails to start.

## Current Behavior
Graylog doesn't retry starting the input and logs the following exception:
```
Could not launch AMQP consumer.
```

## Reasons & Possible Solution
The reason for that error is in the code. When Graylog fails to start the AMQP consumer it gets to the following line of code:
https://github.com/Graylog2/graylog2-server/blob/18bb1682aa1bd36d6e10ed50ac1029ef8f38f429/graylog2-server/src/main/java/org/graylog2/inputs/transports/AmqpTransport.java#L179

It breaks deeper in the AmqpConsumer class. When a connection is created and it fails in the beginning - it throws an exception that is not well served by the try-catch block here:
https://github.com/Graylog2/graylog2-server/blob/18bb1682aa1bd36d6e10ed50ac1029ef8f38f429/graylog2-server/src/main/java/org/graylog2/inputs/transports/AmqpConsumer.java#L169-L173

It then propagates to the following block in AmqpTransport class:
https://github.com/Graylog2/graylog2-server/blob/18bb1682aa1bd36d6e10ed50ac1029ef8f38f429/graylog2-server/src/main/java/org/graylog2/inputs/transports/AmqpTransport.java#L175-L180

Which then propagates further and is logged. The AMQP library used does not take care of failures when creating connection instances. I would suggest retrying to create connection when it throws TimeoutException instead of just logging failure. This probably could be done at the AmqpConsumer class level.

1. Run Graylog server.
2. Run RabbitMQ server, I'm not sure if having constant flow of messages is needed for reproducing the issue.
3. Set up AMQP GELF Input in Graylog, pointing to the RabbitMQ instance. Run it. It should work.
4. Stop Graylog and RabbitMQ.
5. Start Graylog, wait for it to fully start.
6. Start RabbitMQ.
7. Check whether the input has started.
8. Check Graylog logs to see that it failed to start.

## Context
During deployment we need to reboot all of our VMs. We noticed that sometimes (not always) Graylog AMQP inputs fail to start, leaving a huge amount messages in RabbitMQ.
We did some empiric tests to check different scenarios:
- When RabbitMQ is rebooted - Graylog retries connecting to RabbitMQ until it works again.
- When Graylog and RabbitMQ are rebooted:
   -  If Graylog stands up first - it fails to start the input and never retries
   - If RabbitMQ stands up first - everything is alright, there is no need for retries

## Your Environment
Graylog Docker - MongoDB Docker - ES Docker - RabbitMQ Docker.
Graylog, MongoDB and ES are managed by SystemD, RabbitMQ is running in Nomad.

* Graylog Image Version: 3.3.16
* Java Version: The version included in the image
* Elasticsearch Version: 6.8.23
* MongoDB Image Version: 4.0.4
* Operating System: Ubuntu 18.04.6 LTS

AntonEbel:
Improvements:
- retry initial connection
- add configuration value for connection recovery interval
- encrypt password
- use failure recorder

Fixes #12447

## How to test:
Start rabbitmq and login with user:guest password:guest.  You can use the following docker compose file:
```
version: "3.7"
services:
  rabbitmq:
    hostname: 'rabbitmq'
    image: rabbitmq:3.8.14-management
    container_name: 'rabbitmq'
    ports:
      - 5672:5672
      - 15672:15672
    volumes:
      - '<path>:/var/lib/rabbitmq/mnesia/'
volumes:
  rabbitmq:
```
Create a queue with the following binding and use it in Graylog:
image

For testing connection recovery interval value you can close the connection here:
image


moesterheld (approved):
works as expected. lgtm

PR #15410, opened 2023-05-03
Set ESLint rule `arrow-body-style` to `as-needed`.

linuspahl:
## Motivation and Context

With this PR we are changing the configuration of the ESLint rule `arrow-body-style` from `off` to `['error', 'as-needed']`.

https://eslint.org/docs/latest/rules/arrow-body-style

This way the linter will prefer
```
const exampleFn = () => 'example-return-value';
```
over
```
const exampleFn = () => {
  return 'example-return-value';
}
```

/nocl


dennisoelkers (approved):
Could not check every line, but general changes look good, tests are green and interface builds fine.

PR #15411, opened 2023-05-03
Move zstd dependency into the server

mpfz0r:
We only had this as an enterprise dependency.
The server could use it through a transititive dependency of kafka-client. This however was version 1.4.5-6

which has an issue on M1 Macs: https://github.com/luben/zstd-jni/issues/153

/nocl


janheise (approved):
Fixes it!

PR #15419, opened 2023-05-04
Add CLOUD capability to allow loading plugins conditionally

thll:
By adding this capability when the server starts up in "cloud" mode, we can include plugins which are meant only for running in the cloud.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise-integrations#1054,Graylog2/graylog-plugin-enterprise#5128
/nocl


mpfz0r (approved):
LGTM

PR #15421, opened 2023-05-04
Refactor scripting api for advanced field types

todvora:
This PR refactors how scripting api deals with field types (string, number, geo...). It also adds support for advanced field types - it automatically translates stream/node/input IDs to human readable titles. 

By default, for stream, input and node we'll output the name of the entity. This is a breaking change which we agreed upon and accept. The original ID can be returned by using the `.id` suffix.

The drawback of post-processing the IDs and translating them to titles is the non-uniqueness of the titles. We don't require unique stream names. So if the user aggregates by the streams (or streams.name), we could return more stream names as keys which are also non-unique, causing some confusion if the aggregation actually works correctly. 

## How Has This Been Tested?
Existing set of unit and integration tests, added new unit and integration tests.

image
image

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/response/ResponseEntryDataType.java
```diff
@@ -17,38 +17,46 @@
...
+                .filter(dataType -> dataType.correspondingFieldTypes.stream().anyMatch(corresponding -> isMatch(corresponding, type)))
                 .findFirst()
                 .orElse(UNKNOWN);
     }
 
+    private static boolean isMatch(FieldTypes.Type corresponding, FieldTypes.Type type) {
+        // FieldTypes.Type is not just type, it includes also properties. Properties may differ in the received and corresponding
+        // type, which leads to wrong detection of a match. So we have to use the type itself, which is reliable for now.
+        // TODO: can we solve that in a more elegant way?
+        return corresponding.type().equals(type.type());
```
todvora, line 57: @dennisoelkers This feels like a hack. The problem is that FieldTypes.Type is actually not an enum but a real class and the content may differ. Do you have any idea if we can do better here?

graylog2-server/src/test/java/org/graylog/plugins/views/search/rest/scriptingapi/request/RequestedFieldTest.java
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.plugins.views.search.rest.scriptingapi.request;
+
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class RequestedFieldTest {
```
luk-kaminski, line 22: What happens if we:
1. Provide a string that starts with ".", i.e. ".streams.smth".
2. Provide a string with multiple dots, i.e. "streams.outputs.title"
?
Can the test verify and document expected behavior in those 2 scenarios?

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
```diff
@@ -96,6 +114,30 @@ private static List<List<Object>> getDatarows(final AggregationRequestSpec searc
...
+            final Object decorated = decorate(decorators, field, value, searchUser);
+            result.add(decorated);
+        }
+        return result.stream();
+    }
+
+    private Object decorate(Set<FieldDecorator> decorators, RequestedField field, Object val, SearchUser searchUser) {
+        final Optional<Object> decorated = decorators.stream()
+                .filter(d -> d.accept(field))
+                .findFirst()
```
luk-kaminski, line 131: Is there any chance that at any point we will have conflicting decorators, so that single field is accepted by multiple decorators?
If yes, is it ok that we pick the first one and silently ignore the rest?
If no, can we somehow make it more explicit that we do not expect more than one decorator for one field?

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
```diff
@@ -96,6 +114,30 @@ private static List<List<Object>> getDatarows(final AggregationRequestSpec searc
...
+                .map(d -> d.decorate(field, val, searchUser));
+
+        if (decorated.isEmpty() && field.hasDecorator()) {
+            throw new IllegalArgumentException(StringUtils.f("Unsupported decorator '%s' on field '%s'", field.decorator(), field.name()));
```
luk-kaminski, line 135: Do the API user need to know/be aware of the concept of decorator at all?
If they use "stream.abracadabra" in the API, they should rather receive more general error message, like "cannot provide 'abracadabra' property for stream", it would be nice if it could be followed by "valid properties : id, name, title, whatever", automatically taken from the list of decorators.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/request/RequestedField.java
```diff
@@ -0,0 +1,58 @@
...
+                .splitToList(value);
+
+        if (parts.size() == 1) {
+            return new RequestedField(parts.get(0), null);
+        } else {
+            return new RequestedField(parts.get(0), parts.get(1));
+        }
+    }
+
+    @JsonCreator
```
luk-kaminski, line 41: @JsonCreator annotation here look like a mistake...

graylog2-server/src/test/java/org/graylog/plugins/views/search/rest/scriptingapi/request/RequestedFieldTest.java
todvora: This cases are handled on another level, namely in the TabularResponseCreator implementations. I will add these tests there.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
todvora: I think providing the list of the properties would destroy the simplicity of the accept method and would require way more complicated logic. But the phrasing should be changed, will do that. Thanks1

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/request/RequestedField.java
todvora: Indeed, a leftover from solution that I have dropped along the way. Removed, thanks!

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
luk-kaminski: There is one problem with this approach: the wrong decorator setup will be identified in the unknown moment in the future, when someone invokes a call for certain field and decorator.
I'd prefer to know it straight away, during the startup.

I just wonder if we can simplify "accept" logic, so that wrong setup is identified when decorators are bound in the container?
We can talk more about it if you want.

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
```diff
@@ -17,29 +17,45 @@
...
 import org.graylog.plugins.views.search.rest.scriptingapi.request.AggregationRequestSpec;
+import org.graylog.plugins.views.search.rest.scriptingapi.request.RequestedField;
 import org.graylog.plugins.views.search.rest.scriptingapi.response.Metadata;
 import org.graylog.plugins.views.search.rest.scriptingapi.response.TabularResponse;
+import org.graylog.plugins.views.search.rest.scriptingapi.response.decorators.CachingDecorator;
+import org.graylog.plugins.views.search.rest.scriptingapi.response.decorators.FieldDecorator;
 import org.graylog.plugins.views.search.searchtypes.pivot.Pivot;
 import org.graylog.plugins.views.search.searchtypes.pivot.PivotResult;
 import org.graylog.plugins.views.search.searchtypes.pivot.SeriesSpec;
+import org.graylog2.shared.utilities.StringUtils;
```
luk-kaminski, line 35: Seems like unused import

graylog2-server/src/main/java/org/graylog/plugins/views/search/rest/scriptingapi/mapping/AggregationTabularResponseCreator.java
```diff
@@ -17,29 +17,45 @@
 package org.graylog.plugins.views.search.rest.scriptingapi.mapping;
 
 import com.google.common.collect.ImmutableList;
+import org.apache.shiro.subject.Subject;
```
luk-kaminski, line 20: Seems like unused import

luk-kaminski (approved):
Looks good, thx!

Issue #15370, opened 2023-04-27
Allow defining additional sources for CSP directives


dennisoelkers:
## What?

Currently our content security policy is rather lax. In order to tighten it down, specifying `self` is not sufficient for some directives, due to 3rd party dependencies making use of external sources, which we cannot infer. Therefore, we need to be able to define additional sources in the code. This could happen e.g. through a resource file declaring source patterns and their types (`img`, `connect`, `style`, `script`, ...). These declarations would be merged with our general policy and returned as part of the CSP header.

Ideally it would be maintained in a single (resource) file which is easy to parse, so we can also make use of in `webpack-dev-server` to return CSP headers in development which are as close as possible to the ones in production.

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

patrickmann:
@dennisoelkers Will this format of settings work for FE?

> default.default-src='self'
> default.style-src='self' 'unsafe-inline'
> default.script-src='self' 'unsafe-eval'
> default.connect-src=*
> default.img-src=*
> 
> swagger.style-src='self' 'unsafe-inline'
> swagger.script-src='self' 'unsafe-eval' 'unsafe-inline'
> swagger.connect-src=*
> swagger.img-src=*
>

patrickmann:
Completed with final merge of #15439

PR #15439, opened 2023-05-05
Read static CSP configuration values from file

patrickmann:
See #15370 for details

Initially you could specify a CSP string within the `@CSP` annotation. Combining this with a config file and dynamically updated auth hosts was messy. Now you specify a CSP "group" in the annotation, which corresponds to a group of values from the config file. 

Values are now stored in memory as sets, to avoid duplicates and simplify processing. Any dynamic update is applied to that data structure, so there is no ambiguity or on-the-fly parsing and concatenating of values.


graylog2-server/src/main/resources/org/graylog2/security/csp.config
```diff
@@ -0,0 +1,34 @@
...
+# value is not ignored, and is treated as part of the value.
+#
+# Example:
+# default.img-src=my_imgsrc.com
+# default.connect-src=my_connectsrc.com:9999
+
+default.default-src='self'
+default.style-src='self' 'unsafe-inline'
+default.script-src='strict-dynamic' 'nonce-{nonce}' 'unsafe-eval'
+default.img-src=data:*
```
dennisoelkers, line 30: This should be `data:` instead and we should also include `'self'` and the exception for OSM here:
```suggestion
default.img-src=data: 'self' https://*.tile.openstreetmap.org
```

graylog2-server/src/main/resources/org/graylog2/security/csp.config
```diff
@@ -0,0 +1,34 @@
...
+
+swagger.style-src='self' 'unsafe-inline'
+swagger.script-src='self' 'unsafe-eval' 'unsafe-inline'
+swagger.img-src=*
```
dennisoelkers, line 34: Is swagger actually loading images from somewhere else than `'self'`?

dennisoelkers (changes requested):
Looks good so far, added on suggestion and a question inline.

graylog2-server/src/main/resources/org/graylog2/security/csp.config
patrickmann: updated

graylog2-server/src/main/resources/org/graylog2/security/csp.config
patrickmann: Doesn't appear to - works fine with `'self'`

bernd:
@patrickmann @dennisoelkers I noticed we are implementing a custom property file parser for the CSP config file. What's the reason for configuring the CSP feature using a property-style file and a custom parser instead of a JSON or YAML file?

patrickmann:
> What's the reason for configuring the CSP feature using a property-style file and a custom parser instead of a JSON or YAML file?

A Java property file seemed the simplest solution. Do we have a coding convention not to use these?

Allowing multiple entries for the same attribute is a convenience feature; but it also means it is no longer a valid property file.

bernd:
> A Java property file seemed the simplest solution. Do we have a coding convention not to use these?
> 
> Allowing multiple entries for the same attribute is a convenience feature; but it also means it is no longer a valid property file.

We don't have a coding convention for not using property files. It's about having to implement a custom parser. :slightly_smiling_face: If we would use JSON/YAML for that, we wouldn't have to implement and maintain a custom property file parser. Or we don't use the convenience feature and a regular property parser.

dennisoelkers (approved):
Looks great, works fine! Thanks a lot!

PR #15470, opened 2023-05-10
Rule Builder UI

### LINKED ISSUE #15490: Pipeline Rule UI
## What?

Creating a UI layer to allow building pipeline rules visually rather than just using source code.

## Why?

Graylog’s pipeline rules are written in a custom language that can be difficult to learn. It is hard to correctly express what you want it to do, especially if you have little experience writing code, and it includes a bit of boilerplate you need to write, even for simple, everyday processing tasks

## Your Environment

* Graylog Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

kodjo-anipah:
## Description
closes #15490 
Implements the RuleBuilder UI

- [x] Add modal for choosing between Rule Builder and Editor when clicking on "Create Rule" button
- [x] Change placeholder for Rule Editor Simulator Input (JSON format supported)
- [x] Add hint to Rule Editor that this rule cannot be converted into Rule Builder
- [x] Create a new page for Rule Builder
	- Title
	- Description
	- no Rule Reference
	- Rule Simulator at the bottom
- [x] Button to convert Rule Builder to text
	- On click it opens a modal that shows the result and a button to copy to the clipboard
	- Optional: Maybe add a second button to create a new rule from it directly
- [x] Rule Builder component
	- Conditions on the left, actions on the right
	- select for functions, default nothing is selected 
	- Using a dictionary for supported functions to dynamically generate form fields based on parameters
- [x] Rule Block component (can be used for actions and conditions
	- default: nothing is selected
	- takes functions information (e.g. name, description, fields as props)
	- UI for negating the whole Rule Block (only if condition)
	- Buttons at the bottom right to save/cancel the rule
		- on save: validate the rule, save the current text version in the state
		- show validation result on each rule block
	- Saved rules
		- have an edit and delete button
		- don't show any editable fields
		- Otherwise, look the same and show information about rule and parameters
	- On delete
		- show confirmation
		- validate the whole rule, regenerate text version
	- On edit show initial view with editable fields
- [ ] Add telemetry
- [ ] GIM Name suggestion
	- Only for actions that create or rename fields (using backend flag, e.g. `gim-field`) we show a text field with autocomplete, using an API to suggest gim field names

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rulebuilder/parser/ActionParser.java
```diff
@@ -47,32 +46,42 @@ String generateAction(RuleBuilderStep step, boolean generateSimulatorFields) {
...
-        FunctionDescriptor<?> function = actions.get(step.function()).descriptor();
+        final RuleFragment ruleFragment = actions.get(step.function());
+        FunctionDescriptor<?> function = ruleFragment.descriptor();
+
         String syntax = "  ";
         if (Objects.nonNull(step.outputvariable())) {
-            if (function.returnType().equals(Void.class)) {
+            if (function.returnType().equals(Void.class)) { // cannot set output variables for functions returning void
                 throw new IllegalArgumentException("Function " + step.function() + " does not return a value.");
+            } else if (ruleFragment.isFunction()) { // only set output variables to result of functions
```
kodjo-anipah, line 56: maybe for debugging we should throw an error here when we have an output that is not the result of a function, since this shouldn't be possible or?

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rulebuilder/parser/ActionParser.java
```diff
@@ -47,32 +46,42 @@ String generateAction(RuleBuilderStep step, boolean generateSimulatorFields) {
             throw new IllegalArgumentException("Function " + step.function() + " not available as action for rule builder.");
         }
 
-        FunctionDescriptor<?> function = actions.get(step.function()).descriptor();
+        final RuleFragment ruleFragment = actions.get(step.function());
+        FunctionDescriptor<?> function = ruleFragment.descriptor();
+
         String syntax = "  ";
         if (Objects.nonNull(step.outputvariable())) {
```
kodjo-anipah, line 53: We could create a set of validators that we need to run and separate that out of the parsing, what do you think?

moesterheld: yes, good idea. we need to pull out validation.

ousmaneo (changes requested):
I started testing and reviewing the Rule builder. Here are my first findings below.
I will focus on code review and continue testing later.

ousmaneo (changes requested):
Hi I add some comment for some smaller issue I found.

ousmaneo (approved):
LGTM !

PR #15479, opened 2023-05-11
EntityTitleService uses SearchUser instead of Subject

todvora:
EntityTitleService used Subject, which is quite complex and heavy-weight object. With the SearchUser and especially with its Permission interfaces we can test the code more easily and don't need that many mocks. 

Extracted interface for cleaner access to the service.

/nocl

## Description
Purely internal refacoring

## How Has This Been Tested?
Existing unit and integration tests

## Types of changes

## Checklist:


luk-kaminski:
Wouldn't it be easier to store Subject as a field in SearchUser, with some getter?
With this approach one could use the service, no matter if having Subject object or SearchUser object wrapping Subject object.
Now no one can user the service with Subject anymore.

todvora:
@luk-kaminski The SearchUser is an abstraction that helps us to get rid of the Subject dependency. We don't want to use Subject in our implementations because it's too powerful and broad, needs to be mocked (fragile) and doesn't guide you into correct implementations and usages. 

But the service itself doesn't accept SearchUser, it accepts TitlePermissions, very narrow and specific interface. So you could always call the service with the original subject like this, using the Subject instance:

```java
service.getTitles((perm, id -> getSubject().isPermitted(perm + ":" + id), request);
```

The goal here is not to replace Subject with SearchUser but to narrow down the needed dependency to very basic and limited TitlePermissions interface which happens to be also implemented by the SearchUser. 

This limits the error surface, communicates better what's the real dependency, helps testing and makes code more cleaner and resilient.

graylog2-server/src/test/java/org/graylog2/rest/resources/system/contentpacks/CatalogResourceTest.java
```diff
@@ -139,9 +140,9 @@ public void getTitles() {
         );
         doReturn(expectedResponse)
                 .when(entityTitleService)
-                .getTitles(any(), eq(request));
+                .getTitles(eq(request), any());
```
luk-kaminski, line 143: It is very minor, but I would replace `any()` with `eq(<test search user instance>)`

luk-kaminski (approved):
LGTM, thx!

PR #15503, opened 2023-05-15
Handle "429 too many requests exception" in S3 archiving

patrickmann:
Resolves Graylog2/graylog-plugin-enterprise#5136

See issue for details


graylog-storage-opensearch2/src/test/java/org/graylog/storage/opensearch2/OpenSearchExceptionTest.java
```diff
@@ -0,0 +1,149 @@
...
+        };
+        when(response.getEntity()).thenReturn(httpEntity);
+
+        ResponseException responseException = new ResponseException(response);
+        RestStatus restStatus = RestStatus.BAD_REQUEST;
+        OpenSearchStatusException statusException = new OpenSearchStatusException(
+                "status msg", restStatus, responseException);
+
+        OpenSearchInstance openSearchInstance = OpenSearchInstance.create();
+        final OpenSearchClient openSearchClient = openSearchInstance.openSearchClient();
```
mpfz0r, line 143:
```suggestion
        final OpenSearchClient openSearchClient = new OpenSearchClient(mock(RestHighLevelClient.class), true, new ObjectMapper());```
        
If we are going to throw the exception ourselves anyway, we don't need a real opensearch instance, which takes minutes to start, but can also just mock the client.

mpfz0r: also we should've called
`openSearchInstance.close()` to shut down the test instance properly

mpfz0r (changes requested):
test can be made faster

graylog-storage-opensearch2/src/test/java/org/graylog/storage/opensearch2/OpenSearchExceptionTest.java
patrickmann: Good point. Removed the instance.

PR #15504, opened 2023-05-15
Sliding expiration for authentication cookie.

### LINKED ISSUE #15502: Session Cookies do not get updated on Client Action - Session timesout in the frontend even if user is active
## Expected Behavior

Configured Session Timeouts (e.g. 15 Minutes) logout users if they are more than 15 Minutes inactive in the graylog webinterface. 
## Current Behavior

User gets logged out nearly random because the session cookie gets only set/updated on full page refreshs. Interactions e.g. with search or other part of the admin area will update the mongodb "last_access_time" value, but the cookie expire date is depending on the last full page refresh/login. 

## Possible Solution

Different options here:
- Do not send an expire for the cookies or expire at browser close and let the backend handle complete session management.
- Call Session Api periodically to update the cookie. But keep in mind to set the X-Graylog-No-Session-Extension Header to not extend the session while calling the session endpoint :-) 

1. Login to Graylog Webinterface with open Network Analyze Dev Console
2. Observe that the cookie gets set on first login / full page refresh when the Sessions endpoint is called with an expire of session timeout + x minutes?
3. All Actions after this will nearly never call the session endpoint again so the Cookie expires
4. Frontend looses the session cookie
5. user has to relogin even though the backend session would still be valid

## Context

Active users should not get "logged out", this is especially in regard of the search resetting on full page refresh/logout very time consuming as an analyst has to redo all his work if he didn't hit save on the view just before the logout triggers. 

## Your Environment

* Graylog Version: 5.0.4
* Java Version: 17
* Opensearch Version: 1.3.8
* MongoDB Version: 5.x
* Operating System: Windows 
* Browser version: Firefox Current ESR

dennisoelkers:
**Note:** This requires a backport to `5.0` and `5.1`.

## Motivation and Context

Prior to this PR, the `authentication` cookie set when authenticating via `POST /api/system/sessions` had a max-age defined by the session lifetime of the user. This max-age was unfortunately not extended by subsequent request, leading to early logout (due to the browser invalidating the cookie) in the frontend, although the session was still valid, unless the user performed a full-page refresh.

This PR is changing this by adding a sliding expiration mechanism extending the max-age of the cookie on every request that does not disable session extension.

Fixes #15502.

## Types of changes

## Checklist:


thll (approved):
LGTM

HenryTheSir:
Hi @dennisoelkers ,

can we get an backport for this? Thanks in advance!

best regards

PR #15514, opened 2023-05-16
Manage zstd-jni and commons-lang3 dependencies in graylog-parent

bernd:
Move zstd-jni from graylog-project-parent to graylog-parent because some transitive plugin dependencies are using it.

Add version property for the latest commons-lang3 dependency.

/nocl

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,6 +104,7 @@
         <classgraph.version>4.8.157</classgraph.version>
         <commons-codec.version>1.14</commons-codec.version>
         <commons-email.version>1.5</commons-email.version>
+        <commons-lang3.version>3.12.0</commons-lang3.version>
         <commons-validator.version>1.6</commons-validator.version>
         <commons-io.version>2.6</commons-io.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -282,6 +283,18 @@
                 <version>${netty-tcnative-boringssl-static.version}</version>
                 <classifier>linux-x86_64</classifier>
             </dependency>
+            <!-- Ensure consistent commons-lang3 versions across dependencies and transitive dependencies -->
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
```
mpfz0r, line 289: I was running into that as well. 
Funny enough, we don't even have an explicit dependency on commons-lang3 anywhere

thll (approved):
👍  I was also about to add this.

pom.xml
bernd: I added an explicit commons-lang3 dependency to the server.

PR #15520, opened 2023-05-16
Update to assertj 3.24.2

bernd:
Remove assertj-json dependency and create a custom JsonPathAssert class that works with the newer assertj version. (assertj-json is unmaintained)

Adjust tests to work with the new assertj version.

/nocl

/jenkins-pr-deps https://github.com/Graylog2/graylog-plugin-enterprise/pull/5187,https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/1055


graylog2-server/src/test/java/org/graylog/plugins/views/search/export/SimpleMessageChunkTest.java
```diff
@@ -45,8 +48,8 @@ void valuesInOrderContainsMissingFieldsAsNull() {
                 linkedHashSetOf("timestamp", "source"),
                 msg1, msg2);
 
-        assertThat(sut.getAllValuesInOrder()).containsExactly(
-                new Object[]{"2015-01-01 01:00:00.000", "source-1"},
-                new Object[]{"2015-01-02 01:00:00.000", null});
+        assertThat(sut.getAllValuesInOrder())
+                .contains(new Object[]{"2015-01-01 01:00:00.000", "source-1"}, Index.atIndex(0))
+                .contains(new Object[]{"2015-01-02 01:00:00.000", null}, Index.atIndex(1));
```
mpfz0r, line 53:
```suggestion
        assertThat(Arrays.stream(sut.getAllValuesInOrder()).toList()).asList()
                .containsExactly(
                        new Object[]{"2015-01-01 01:00:00.000", "source-1"},
                        new Object[]{"2015-01-02 01:00:00.000", null}
                );
```

mpfz0r: how about this?

mpfz0r (changes requested):
two small suggestions. rest LGTM :+1:

graylog2-server/src/test/java/org/graylog/plugins/views/search/export/SimpleMessageChunkTest.java
bernd: @mpfz0r What's the reason for the suggestion? Is it more straightforward for you in that way?

mpfz0r: a little, yeah. at least from the assertj point of view

mpfz0r (approved):

PR #15526, opened 2023-05-17
Add managed dependency for joda-time

mpfz0r:


thll (commented):
Can we remove the management from the `graylog-project-parent` pom then?

thll (approved):

PR #15558, opened 2023-05-22
Add enabled field to EventDefinitionDto

kingzacko1:
## Description

Add `enabled` field to the EventDefinitionDto
/nocl
## Motivation and Context

Some new event definitions, such as System Notification Events, are not able to be scheduled but are instead triggered by some other mechanism within Graylog. Since they can't be scheduled, they need a different field to show them as `enabled`. This field serves that purpose. 

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionDto.java
```diff
@@ -128,6 +129,10 @@ public abstract class EventDefinitionDto extends ScopedEntity implements EventDe
     @Nullable
     public abstract EventDefinitionContextService.SchedulerCtx schedulerCtx();
 
+    @JsonProperty(FIELD_ENABLED)
+    @Nullable
+    public abstract Boolean enabled();
```
danotorrey, line 134:
```suggestion
    @JsonProperty(FIELD_ENABLED)
    public abstract boolean enabled();
```

Small point and definitely not required, but we might be able to make this field to be non-nullable by specifying a `false` default value for it in the builder method. That might make the field slightly easier to use without needing the null checks.

If that causes any issues, we could just keep it as-is with the `@Nullable`.

```
        public static Builder create() {
            return new AutoValue_EventDefinitionDto.Builder()
                    .fieldSpec(ImmutableMap.of())
                    .notifications(ImmutableList.of())
                    .storage(ImmutableList.of())
                    .enabled(false);
        }
```

danotorrey (approved):
@kingzacko1 LGTM and tests successfully! Just left one small question.

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionDto.java
bernd: Yes, please use a default instead of making the field nullable. Thanks!

bernd:
@kingzacko1 @danotorrey I have a few questions.

1. What does it mean for an unscheduled event definition to be "enabled"? And what's supposed to happen if they are disabled?
1. If we add an "enabled" field, we should switch to that field for all event definitions so we don't have to check two states everywhere. Then we can also remove the special handling for system-event notifications. Or do you see an issue with that?

kingzacko1:
> 1. What does it mean for an unscheduled event definition to be "enabled"? And what's supposed to happen if they are disabled?

This has been added for integration of Anomaly Detection into our events system (see Graylog2/graylog-plugin-enterprise#5092 and Graylog2/graylog-plugin-enterprise#5210). The way the AD processor works doesn't lend itself to the events being scheduled. In this case, an enabled event definition would generate a Graylog event when the specified anomaly detector detects an anomaly. A disabled event definition would behave as our AD engine does today: the anomaly event message would be created but no associated event/alert would be generated.

> 2. If we add an "enabled" field, we should switch to that field for all event definitions so we don't have to check two states everywhere. Then we can also remove the special handling for system-event notifications. Or do you see an issue with that?

That would be my personal preference since the existence of a scheduler context has already had to be worked around once with the System Notifications, but I was really only creating this as a draft PR as a potential solution to the problem specific to anomaly detector event definitions not being able to be displayed as `enabled`. I was mostly opening it up to start a discussion, so mission accomplished 😄.  @bernd

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionDto.java
kingzacko1: Updated in latest commits

kingzacko1:
Took a stab at using the `enabled` field only instead of the scheduler context @bernd

danotorrey:
Thanks for the feedback @bernd and additional details @kingzacko1. I agree and like the idea of trying to unify the enabled/disable status tracking. Thanks for the updated code @kingzacko1. I will take a look at that.

I am thinking through the design a bit. If I understand correctly, I think the scheduled event definitions typically perform some query on a schedule, from which matching events are emitted.

Another potential approach: The current design in https://github.com/Graylog2/graylog-plugin-enterprise/pull/5210 seems to be a push-based model where anomaly events are pushed to the AnomalyEventProcessor directly. What if instead, the anomaly processor would query OpenSearch for anomalies (for specific corresponding detectors) and emit those as events (similar to how the `AggregationEventProcessor` works)? That approach would still allow the current existing enabled/disabled to work the same, I think. This also makes me also wonder if we would need to support emitting anomaly messages (as is currently released) and events (together). It would seem to be a bit of duplicated processing if we had subsystems querying for OpenSearch anomalies (if we needed to support both together). That would be one drawback to this idea.

ryan-carroll-graylog (approved):
I know there are some more changes incoming on this PR, but as of now this piece looks solid and tests successfully. Will keep an eye out for updates.

danotorrey (commented):
@kingzacko1 The latest changes here look good nice job in unifying those statuses. I'll take another look/test if you migrate to the status mentioned in https://github.com/Graylog2/graylog-plugin-enterprise/pull/5210#issuecomment-1561761861. Seems like that should fit well within the implementation here.

ryan-carroll-graylog (commented):
@kingzacko1 The latest looks good and seems to be working well with the new state field!

One small thing I'm seeing is that when I create first a new Event Definition it always shows as disabled regardless of UI selection and whether it's actually enabled or not. It looks like a holdover from the DTO object not having the field (we're using an API param in the EventDefinitionResource.create() method). Setting the DTO `state` field based on the `schedule` API param in that method looked like it worked, but not sure if it's better to figure out how to set in the DTO object that comes from the frontend. 

https://github.com/Graylog2/graylog2-server/blob/e3d2bb218c922e1e816d4d255cb4c034aaa8983d/graylog2-server/src/main/java/org/graylog/events/rest/EventDefinitionsResource.java#LL259C12-L259C12

ryan-carroll-graylog:
> One small thing I'm seeing is that when I create first a new Event Definition it always shows as disabled regardless of UI selection and whether it's actually enabled or not.

Doing some more testing, I'm also seeing that Sigma rule Event Definitions are showing `disabled` when created, even though I can see that they have indeed been scheduled. So maybe moving the logic that sets the `EventDefinitionDto.state()` field into the `EventDefinitionHandler` makes more sense since it looks like that's where both of these workflows overlap.

ryan-carroll-graylog (approved):
Looks solid with the latest and standard Event Definition, Sigma Event definition, and Anomaly event definition functionalities all test successfully!

danotorrey (approved):
LGTM and tests successfully! Just noticed one small nit.

graylog2-server/src/main/java/org/graylog/events/rest/EventDefinitionsResource.java
```diff
@@ -357,7 +360,11 @@ public Response bulkSchedule(@ApiParam(name = "Event definitions to enable", req
         final BulkOperationResponse response = bulkScheduleExecutor.executeBulkOperation(bulkOperationRequest,
                 userContext,
                 new AuditParams(EventsAuditEventTypes.EVENT_DEFINITION_UPDATE, "definitionId", EventDefinitionDto.class));
-
+        // Get a list of successfully scheduled event definitions
+        List<String> idsFailedToSchedule = response.failures().stream().map(BulkOperationFailure::entityId).toList();
+        List<String> successfullyScheduledIds = bulkOperationRequest.entityIds().stream()
```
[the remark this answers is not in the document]
kingzacko1: Ah yep, I moved the logic to mark the scheduled event definitions as enabled/disabled and these as well as the ones in bulkUnschedule are no longer necessary. Good find!

HenryTheSir:
Hi @bernd, 
this is a breaking api change aswell. As previously the state of an alert did not get changed on an update via REST API it now gets assumed that it should be disabled. 

best regards

kingzacko1:
Thank you for finding that @HenryTheSir. I've created #17958 to track it and #17959 to fix it.

PR #15615, opened 2023-05-26
Delay opensearch start till configuration is available

todvora:
/nocl
This PR add a `startWithConfig` method that will start the opensearch process. The start may be delayed or triggered anytime, based on configuration change.

## Motivation and Context
Needed for postponing opensearch startup till the TLS configuration is available.

## Types of changes

## Checklist:

### DIFF
--- a/data-node/src/main/java/org/graylog/datanode/management/ManagableProcess.java
+++ b/data-node/src/main/java/org/graylog/datanode/management/ManagableProcess.java
@@ -21,7 +21,9 @@ import org.graylog.datanode.process.ProcessState;
 
 import java.io.IOException;
 
-public interface ManagableProcess {
+public interface ManagableProcess<ConfigurationType> {
+
+    void startWithConfig(ConfigurationType configuration);
 
     void start() throws IOException;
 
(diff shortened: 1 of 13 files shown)


janheise (approved):

data-node/src/main/java/org/graylog/datanode/management/ManagableProcess.java
```diff
@@ -21,7 +21,9 @@
 
 import java.io.IOException;
 
-public interface ManagableProcess {
+public interface ManagableProcess<ConfigurationType> {
+
+    void startWithConfig(ConfigurationType configuration);
 
     void start() throws IOException;
```
luk-kaminski, line 28: Please correct me if I am wrong, but it seems that this method will always fail if executed without executing `startWithConfig` first. Shouldn't we name it differently, i.e. `restart`, so that no one tries to execute it too early?

todvora: good idea, renamed!

data-node/src/main/java/org/graylog/datanode/management/ManagableProcess.java
```diff
@@ -21,7 +21,9 @@
 
 import java.io.IOException;
 
-public interface ManagableProcess {
+public interface ManagableProcess<ConfigurationType> {
```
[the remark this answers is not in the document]
todvora: Ok, done

PR #14041, opened 2022-11-23
Add email attribute to LDAP config

patrickmann:
Fixes #11131 

Currently we use hard-coded attribute names for LDAP email and fail if those are not found.
Provide ability to configure this, defaulting to the current attributes.

PR #15655, opened 2023-05-31
Use correct email attribute for AD users

### LINKED ISSUE #15652: AD authentication backend cannot sync users in 5.1
https://community.graylog.org/t/active-directory-authentication-not-working-in-5-1/28929

> I’m running Graylog with Active Directory Authentication since at least version 4.2. A few days ago I upgraded to version 5.1 (from 5.0). After that I can not login anymore using a user from the AD. Users not associate with the AD can login.
> 
> Using the login form the error message is:
> Error - the server returned: 503 - There was an error fetching a resource: Service Unavailable. Additional information: Authentication service unavailable
> 
> Using the “Test Server connection” in System/Authentication the message is:
> 
> There was an error fetching a resource: Bad Request. Additional information: Unable to map property email_attributes. Known properties include: servers, verify_certificates, user_search_base, transport_security, type, user_name_attribute, user_search_pattern, system_user_dn, system_user_password, user_full_name_attribute
> false

## Possible Solution

5.1 introduced configurable email attribute for LDAP. Since AD shares some of the code with LDAP, this included changes to AD authentication backend. Guessing the problem is this:
https://github.com/Graylog2/graylog2-server/blob/e3b1f04db87ea81347b07fc7f1039038ced5130f/graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java#L153

## Context

Users can no longer log in after upgrading to 5.1

## Your Environment

* Graylog Version: 5.1

patrickmann:
Use `mail` as the hardcoded email attribute for AD user sync. This is the behavior through version 5.0 and was inadvertently broken in 5.1.
See linked issue for details.

Fixed #15652

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
+++ b/graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
@@ -41,7 +41,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import java.security.GeneralSecurityException;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -150,7 +150,7 @@ public class ADAuthServiceBackend implements AuthServiceBackend {
                 .userUniqueIdAttribute(AD_OBJECT_GUID)
                 .userNameAttribute(config.userNameAttribute())
                 .userFullNameAttribute(config.userFullNameAttribute())
-                .emailAttributes(new ArrayList<>())
+                .emailAttributes(Arrays.asList("mail"))
                 .build();
 
         return ldapConnector.searchUserByPrincipal(connection, searchConfig, authCredentials.username());
--- a/changelog/unreleased/issue-15652.toml
+++ b/changelog/unreleased/issue-15652.toml
@@ -1,0 +1,6 @@
+type = "fixed"
+message = "Restore mail attribute for AD authentication."
+
+issues = ["15536"]
+pulls = [""]
+


graylog2-server/src/main/java/org/graylog/security/authservice/backend/ADAuthServiceBackend.java
```diff
@@ -150,7 +150,7 @@ private Optional<LDAPUser> findUser(LDAPConnection connection, AuthServiceCreden
                 .userUniqueIdAttribute(AD_OBJECT_GUID)
                 .userNameAttribute(config.userNameAttribute())
                 .userFullNameAttribute(config.userFullNameAttribute())
-                .emailAttributes(new ArrayList<>())
+                .emailAttributes(Arrays.asList("mail"))
```
mpfz0r, line 153: hmm, isn't that missing the `rfc822Mailbox` attribute? looks like we've had it in the past

mpfz0r: I might be wrong :thinking:

mpfz0r: Referring to this here
https://github.com/Graylog2/graylog2-server/pull/14041/files?diff=unified&w=0#diff-dba3473472c7f22e49e817482ca7db7a3726e55cc78fb011c4c31ad64e9972d4L199-L200

patrickmann: We did - is that an AD or LDAP attribute? I only saw `mail` in the AD docs.

mpfz0r (approved):
Works and LGTM :+1:

PR #15657, opened 2023-05-31
update user session timeout on configuration change

### LINKED ISSUE #15609: Altering user session timeout should invalidate current cookie
When altering a users session timeout policy from "sessions do not time out" to use the global timeout policy, the original cookie is not revoked and thus that user will remain logged in until they decide to logout. Once logged back in the correct timeout is reflected within the cookie.

Is it possible to revoke cookies or alter cookies at the point the setting is changed or will the user always needs logout and back in. If not possible, would it be possible to bounce the user out automatically upon changing the setting?

kodjo-anipah:
## Description

Update user session token lifetime, when the configuration is changed in the user config overview. 
fixes #15609
## Motivation and Context

When altering a users session timeout policy from "sessions do not time out" to use the global timeout policy, the original cookie is not revoked and thus that user will remain logged in until they decide to logout. Once logged back in the correct timeout is reflected within the cookie.

## How Has This Been Tested?

Steps: 
1. Create user and make sure his sessions never time out 
image
2. Log in with that user. 
3. Verify in the database that the user session timeout value is -1
4. Update the timeout for the user and verify that the timeout value changed
5. Perform actions as the user and verify that cached session was also updated accordingly
6. Verify that after timeout passed, user is logged out
## Types of changes

## Checklist:


dennisoelkers (approved):
It works well! Just added a minor suggestion for a code improvement. It is not blocking the merge, so you can decide if you want to add it or not.

PR #15670, opened 2023-06-01
Initialize default outputs and embedd flushing

mpfz0r:
- Call initialize for defaultMessageOutput in OutputSetupService

- Embed `BatchedElasticSearchOutputFlushThread` into `BlockingBatchedESOutput`

/nocl


graylog2-server/src/main/java/org/graylog2/outputs/BlockingBatchedESOutput.java
```diff
@@ -181,6 +187,11 @@ public void stop() {
         super.stop();
     }
 
+    @Override
+    public void initialize() throws Exception {
+        flushService.scheduleAtFixedRate(this::forceFlushIfTimedout, outputFlushInterval, outputFlushInterval, TimeUnit.SECONDS);
```
bernd, line 192: This doesn't implement any error handling and logging. We should implement the same handling and logging we had in the periodical.

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/outputs/BlockingBatchedESOutput.java
mpfz0r: yeah, makes sense :+1:

thll (approved):

mpfz0r:
@bernd all good now?

bernd (approved):

PR #15676, opened 2023-06-02
Process watchdog reimplemented to follow transitions of the process state machine

todvora:
The initial solution used an independent process with busy waiting inside a thread and polling the state of the process. New solution is observing transitions of the state machine. This should be more robust and resilient. It also supports delayed start of the opensearch process, which is required for the preflight configuration.  

/nocl

## How Has This Been Tested?
Existing set of unit and integration tests (including the initial watchdog test)

## Types of changes

## Checklist:


data-node/src/main/java/org/graylog/datanode/process/FailuresCounter.java
```diff
@@ -47,4 +47,8 @@ public synchronized boolean failedTooManyTimes() {
     public synchronized void resetFailuresCounter() {
         this.counter = initialValue;
     }
+
+    public int getCounter() {
+        return counter;
```
luk-kaminski, line 52: Its tiny, but I would name this variable `count`, not `counter`.

data-node/src/test/java/org/graylog/datanode/management/ProcessWatchdogTracerTest.java
```diff
@@ -0,0 +1,61 @@
...
+        Assertions.assertThat(isInStartingState(process)).isTrue();
+
+        // repeat
+        terminateProcess(process);
+        Assertions.assertThat(isInStartingState(process)).isTrue();
+
+        terminateProcess(process);
+        Assertions.assertThat(isInStartingState(process)).isTrue();
+
+        // this is the 4th termination, we give up trying
```
luk-kaminski, line 47: This test makes me feel that the count should be configurable in the watchdog creation, so that we do not have to ask questions like "why 4 times?" in the test.

data-node/src/main/java/org/graylog/datanode/process/FailuresCounter.java
todvora: agree, renamed

data-node/src/test/java/org/graylog/datanode/management/ProcessWatchdogTracerTest.java
todvora: good point, made it explicit in the constructor. Thanks!

luk-kaminski (approved):
I love this change, thank you! :)

PR #15677, opened 2023-06-02
Updating sources dashboard with widgets making use of advanced field types.

dennisoelkers:
## Motivation and Context

This PR is updating the "Sources" dashboard with widgets making use of the new advanced field types to make debugging a local Graylog instance/cluster easier. It adds widgets breaking down the incoming log traffic over time by streams, nodes and inputs. In addition, it adds a second tab showing log volume by size instead of count.

image
image

In order to update the "Sources" dashboard locally, this PR does the following steps:

  - Disabling the previous migration inserting and installing v1 of the Sources Page content pack
  - Adding a new migration inserting v2 of the Sources Page content pack
    - This migration also installs v2 of the Sources page content pack, if one of the following conditions is true:
      - The previous migration never ran (fresh installs)
      - v1 of the Sources Page was installed, but not modified locally
    - v1 is uninstalled, if it was installed before _and_ not modified locally
    - If v1 had been installed before, but the Sources Page was modified, the content pack is inserted but not installed and the previous version is kept. In addition, a notification is raised explaining the rationale and how to install it manually.

A big problem in our content pack updating process right now is that it is not easily possible to update content in place. E.g. it is not possible to update a dashboard keeping its id. Every installation of a content pack results in new ids for entities. This leads to deep links being broken and information about recent usage/favorites is getting lost.

Therefore, the migration is setting the id of the new dashboard to the id of the previous one (if it existed before), in a hacky attempt to retain links/references.

## Types of changes

## Checklist:


todvora (approved):
LGTM, tested locally and everything went as expected.

PR #15703, opened 2023-06-06
fix(5236): Update base docs url

zeeklop:
Updates how DocHelper builds the docs URL. With this update, programmers can specify the version of the docs to open when clicking on the link.

/nocl

## Description

Updates how DocHelper builds the docs URL. With this update, programmers can specify the version of the docs to open when clicking on the link. 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/util/DocsHelper.js
+++ b/graylog2-web-interface/src/util/DocsHelper.js
@@ -66,8 +66,17 @@ class DocsHelper {
 
   DOCS_URL = 'https://docs.graylog.org/docs';
 
-  toString(path) {
-    const baseUrl = this.DOCS_URL;
+  DOCS_URL_BY_VERSION = {
+    '5.0': 'https://go2docs.graylog.org/5-0',
+    // eslint-disable-next-line quote-props
+    '5.1': 'https://go2docs.graylog.org/5-1',
+  };
+
+  toString(path, inVersion = null) {
+    const currentDocsVersion = Object.keys(this.DOCS_URL_BY_VERSION)[Object.keys(this.DOCS_URL_BY_VERSION).length - 1];
+    const version = Object.keys(this.DOCS_URL_BY_VERSION).includes(inVersion) ? inVersion : currentDocsVersion;
+
+    const baseUrl = inVersion ? this.DOCS_URL_BY_VERSION[version] : this.DOCS_URL;
 
     return path === '' ? baseUrl : `${baseUrl}/${path}`;
   }
--- a/graylog2-web-interface/src/components/common/PageHeader.tsx
+++ b/graylog2-web-interface/src/components/common/PageHeader.tsx
@@ -106,7 +106,7 @@ type Props = {
   lifecycle?: 'experimental' | 'legacy',
   lifecycleMessage?: React.ReactNode,
   subpage: boolean,
-  documentationLink?: { title: string, path: string }
+  documentationLink?: { title: string, path: string, appVersion?: string }
 };
 
 /**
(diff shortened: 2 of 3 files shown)


ryan-carroll-graylog (approved):
Tests successfully

graylog2-web-interface/src/util/DocsHelper.js
```diff
@@ -66,8 +66,17 @@ class DocsHelper {
 
   DOCS_URL = 'https://docs.graylog.org/docs';
 
-  toString(path) {
-    const baseUrl = this.DOCS_URL;
+  DOCS_URL_BY_VERSION = {
+    '5.0': 'https://go2docs.graylog.org/5-0',
+    // eslint-disable-next-line quote-props
+    '5.1': 'https://go2docs.graylog.org/5-1',
```
kingzacko1, line 72: If we don't plan on changing the base `https://go2docs.graylog.org/` would it be possible to turn this into a function that doesn't have to be updated before every release?

zeeklop: Thanks Zack. I'm asking the docs team how do they handle the docs version. I'll update this depending on how all that works

kingzacko1: Sounds good! This approach works, it just adds another step someone has to remember to do before an official release goes out the door.

danotorrey: Thanks for checking on that @zeeklop. I am also wondering: would it work to parse the major and minor version numbers from the supplied `appVersion` and programatically concatenate the `major-minor` string from that?

zeeklop: @danotorrey @kingzacko1 Yes, that was my first thought. I've been talking with the docs team trying to figure it out what is the cadence of the docs. Do we get a new URL on each Graylog release, or is there a period of time when the docs are behind?

So far haven't got a straight answer. The Docs team is trying to create a URL with an alias for the latest something like

go2docs.graylog.org/latest | current/

If you look at the latest modification on this file, you'll see that what is doing is getting the mayor and minor release number and matching it with the list of docs URLs using RegEx. If it doesn't find a match, it uses the latest as base.

My concerns are
1. If we use the concatenation approach and the docs are behind, then all links will take the user to a 404-page putting the app in a broken state until the docs catch up.
2. Not all versions of the docs have the same base URL. 3.3 and earlier are in an archive subdomain. Looks like we started using go2docs from 4.3

In conclusion, If the Docs URL gets updated on each release, then we can use concatenation for prod with the caveat that in dev we will have to use the previous version.

Another solution would be to always send the user to the latest version, since the docs page has a dropdown to select after what version you want to see.

Trying to send the user the latest version of the docs could be a little tricky if the URL doesn't get updated during the release of a new minor or mayor version of Graylog. We would have to somehow check first that there is a version of the docs for the app version otherwise, drop a version until we find one that works (no 404)

I thought it would be an easy fix, but looks like the docs team has been at it for quite some time trying to figure it out the best solution for this.

danotorrey: @zeeklop Thanks for the details. Ahh, I see. the older versions use a different scheme. Given the those different schemes, I don't see a better way than the current implementation. I think this change should be helpful, since we probably have a decent portion of users on `5.0` still.

This is only applicable in cases where devs explicitly specify the `appVersion`, right? In that case, then the version -> docs path mapping only needs to be updated when a dev needs to reference a new specific version. The correct mapping could be added at that time, since there is a chance the scheme _may_ change by then.

boosty:
I will close this due to lack of activity. The originating issue has been closed in the meantime: https://github.com/Graylog2/graylog-plugin-enterprise/issues/5236

Feel free to re-open if needed.

PR #15704, opened 2023-06-06
Feat/undo redo action

maxiadlovskii:
## Description

This PR adds 2 buttons to the sidebar with undo/redo actions. These actions change the view state in the proper direction. 
The buffer size is 10. It can be changed if it wouldn't be enough.
## Motivation and Context
#15713 

## Types of changes

## Checklist:


graylog2-web-interface/src/views/components/sidebar/SidebarNavigation.tsx
```diff
@@ -83,14 +83,18 @@ const SidebarNavigation = ({ sections, activeSection, selectSidebarSection, side
...
+
+          return (
+            <NavComponent isSelected={activeSectionKey === key}
+                          icon={icon}
+                          onClick={() => selectSidebarSection(key)}
+                          key={key}
+                          title={title}
+                          sidebarIsPinned={sidebarIsPinned} />
+          );
+        })}
```
linuspahl, line 97: It would be nice to visually distinguish the these actions from the actual sidebar sections.

We could add a separator.
image

linuspahl: Beside the `sections` array, we could also have an `actions` array.

graylog2-web-interface/src/views/components/sidebar/NavItem.tsx
```diff
@@ -21,27 +21,30 @@ import styled, { css } from 'styled-components';
...
-const Container = styled.div<ContainerProps>(({ theme: { colors, fonts }, isSelected, sidebarIsPinned }) => css`
+const Container = styled.div<ContainerProps>(({ theme: { colors, fonts }, isSelected, sidebarIsPinned, disabled }) => css`
   position: relative;
   z-index: 4; /* to render over SidebarNav::before */
   width: 100%;
   height: 40px;
   text-align: center;
-  cursor: pointer;
+  cursor: ${disabled ? 'not-allowed' : 'pointer'};
+  opacity: ${disabled ? 0.65 : 1};
```
linuspahl, line 47: Maybe we can apply the opacity only for the icon and not for the label?
image

linuspahl (changes requested):
I found some problems, but in general it works well!

When I add an aggregation, I need to click undo twice to undo the widget creation (but redo only once afterwards).

undo_twice

When I create a widget, focus it (or open edit mode) and click on undo (which reverts the widget creation), I see an error: `Cannot read properties of undefined (reading 'toArray’)`.
undo creation focus

When I create a dashboard tab and click undo, I get the error `Cannot read properties of undefined (reading 'widgets')`.

graylog2-web-interface/src/views/components/sidebar/SidebarNavigation.tsx
linuspahl: We talked about removing the arrow in the sidebar, because you can open and close the sidebar by clicking in the section icon, but we can implement this in a different PR.

graylog2-web-interface/src/views/logic/slices/undoRedoSlice.ts
```diff
@@ -0,0 +1,100 @@
...
+    setCurrentRevision: (state, action) => ({
+      ...state,
+      currentRevision: action.payload,
+    }),
+  },
+});
+export const undoRedoSliceReducer = undoRedoSlice.reducer;
+export const { setBuffer, setCurrentRevision } = undoRedoSlice.actions;
+
+export const pushIntoBuffer = (bufferItem: BufferItem, setAsLastRevision: boolean = true) => async (dispatch: AppDispatch, getState: () => RootState) => {
```
dennisoelkers, line 60: We can prevent the double dispatch happening in this action by just moving it into the `undoRedoSlice.reducers` object, removing the need for the `setBuffer` action. Instead, we can just turn it into a pure function returning the new state for the slice with the new item being added and the `currentRevision` pointer being updated.

graylog2-web-interface/src/views/logic/slices/undoRedoSlice.ts
```diff
@@ -0,0 +1,100 @@
...
+      buffer: action.payload,
+    }),
+    setCurrentRevision: (state, action) => ({
+      ...state,
+      currentRevision: action.payload,
+    }),
+  },
+});
+export const undoRedoSliceReducer = undoRedoSlice.reducer;
+export const { setBuffer, setCurrentRevision } = undoRedoSlice.actions;
```
dennisoelkers, line 58: These actions do not need to be exported.

graylog2-web-interface/src/views/logic/slices/viewSlice.ts
```diff
@@ -122,7 +123,38 @@ export const loadView = (newView: View, recreateSearch: boolean = false) => asyn
...
+    const updatedViewWithSearch = await _recreateSearch(newView);
+    await dispatch(setView(updatedViewWithSearch, true));
+
+    return dispatch(execute());
+  }
+
+  return dispatch(setView(newView, true));
+};
+
+export const updateViewNoBufferPush = ({ newView, recreateSearch = false, hasToPushRevision = false }: { newView: View, recreateSearch: boolean, hasToPushRevision: boolean }) => async (dispatch: AppDispatch, getState: () => RootState) => {
```
dennisoelkers, line 146: This is essentially the same as `updateView`. Maybe we should add the `pushRevision` argument to `updateView` and delete `updateViewNoBufferPush`?

graylog2-web-interface/src/views/logic/slices/undoRedoSlice.ts
```diff
@@ -0,0 +1,100 @@
...
+import { selectRootUndoRedo, selectUndoRedoAvailability } from 'views/logic/slices/undoRedoSelectors';
+
+const BUFFER_MAX_SIZE = 10;
+
+type BufferItemType = 'view';
+
+export type BufferItem = { type: BufferItemType, state: any }
+
+export type UndoRedoState = {
+  buffer: Array<BufferItem>,
```
dennisoelkers, line 32: `buffer` is a very generic name. Maybe `revisions` would be better?

graylog2-web-interface/src/views/logic/slices/undoRedoSlice.ts
```diff
@@ -0,0 +1,100 @@
...
+  currentRevision: number
+}
+
+const handlers: { [name in BufferItemType]: any} = {
```
dennisoelkers, line 36: With a single entry, we do not need this type-based dispatch. We should remove it, as we can still add it in case we really need it.

graylog2-web-interface/src/views/logic/slices/undoRedoSlice.ts
```diff
@@ -0,0 +1,100 @@
...
+  // if we are in the middle of the buffer, we have to remove all items after current;
+  const cutBuffer = isLast ? buffer : buffer.slice(0, currentRevision);
+  // if we reach max size of the buffer we have to remove first item;
+  const newBuffer: Array<BufferItem> = (cutBuffer.length < BUFFER_MAX_SIZE) ? [...cutBuffer, bufferItem] : [...cutBuffer.slice(1), bufferItem];
+  const newRevision = setAsLastRevision ? newBuffer.length : currentRevision;
+  dispatch(setBuffer(newBuffer));
+  dispatch(setCurrentRevision(newRevision));
+};
+
+export const undo = () => async (dispatch: AppDispatch, getState: () => RootState) => {
```
dennisoelkers, line 72: Right now we have a cyclic dependency between `undoRedoSlice.ts` and `viewSlice.ts`. We can avoid this by moving the `undo` & `redo` action to a separate `undoRedoActions` module.

dennisoelkers (changes requested):
Works fine, thanks! I added some suggestions for improvement of the code structure.

linuspahl (approved):
In some cases I still need two clicks to undo / redo a change, but in general the functionality seems to work well and all other comments have been addressed.

dennisoelkers (approved):

PR #15726, opened 2023-06-08
Don't shutdown the injected daemon scheduler on ES output shutdown

bernd:
Fixes RejectedExecutionExceptions on server shutdown.

- Cancel the task instead of shutting down the executor
- Rename the field from "flushService" to "daemonScheduler"

/nocl (the issue wasn't present in any release)


thll (approved):
Code looks good but I didn't test it yet.

PR #15750, opened 2023-06-13
add time size based rotation to validator

kodjo-anipah:
Small change to limit rotation strategy to TimeBasedSizeOptimizingStrategy. 
This was forgotten when feature was developed and is breaking now. 

Steps to test: 
1. `enabled_index_rotation_strategies = time-size-optimizing`
2. check available rotation strategies are limited to time size optimizing


patrickmann (approved):
Looks good

PR #15772, opened 2023-06-19
Fix wrong attribute in session

janheise:
## Description

A customer had problems logging in and found the following constellation in MongoDB for a session:

`
{
    "attributes" : {
    },
}
`

instead of `"attributes" : BinData(0, "rO0ABXNyABFqYXZhLnV...`

This created problems during login:

```cluster-graylog3-1     | 2023-06-19 11:44:42,787 ERROR: org.graylog2.shared.rest.exceptionmappers.AnyExceptionClassMapper - Unhandled exception in REST resource

cluster-graylog3-1     | java.lang.ClassCastException: class com.mongodb.BasicDBObject cannot be cast to class [B (com.mongodb.BasicDBObject is in unnamed module of loader 'app'; [B is in module java.base of loader 'bootstrap')

cluster-graylog3-1     | 	at org.graylog2.security.MongoDbSession.getAttributes(MongoDbSession.java:69) ~[graylog.jar:?]

cluster-graylog3-1     | 	at org.graylog2.security.MongoDbSession.getUserIdAttribute(MongoDbSession.java:99) ~[graylog.jar:?]
```

Probably a race condition which was emphasized by using a replicaSet on MongoDB brought this up for the first time now. The root problem is, that during session creation, an invalid state is written for `attributes` that is not cleared up in time by the first update on the session before another read. This seems to be no issue if only reading/writing from the main MongoDB node, but the error could be reproduced in this setting using the debugger.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/security/MongoDbSession.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/MongoDbSession.java
@@ -81,22 +81,17 @@ public class MongoDbSession extends PersistedImpl {
         return null;
     }
 
-    public void setAttributes(Map<Object, Object> attributes) {
-        fields.put("attributes", transformAttributes(attributes));
-    }
-
     @SuppressForbidden("Deliberate use of ObjectOutputStream")
-    public static byte[] transformAttributes(Map attributes) {
+    public void setAttributes(Map<Object, Object> attributes) {
         try {
             final ByteArrayOutputStream bos = new ByteArrayOutputStream();
             // FIXME: This could break backward compatibility if different Java versions are being used.
             final ObjectOutputStream oos = new ObjectOutputStream(bos);
             oos.writeObject(attributes);
             oos.close();
-            return bos.toByteArray();
+            fields.put("attributes", bos.toByteArray());
         } catch (IOException e) {
-            LOG.error("too bad :(", e);
-            return null;
+            LOG.error("Error serializing into binary stream for attributes in Mongo: {}", e.getMessage(), e);
         }
     }
 
--- a/graylog2-server/src/main/java/org/graylog2/security/MongoDbSession.java
+++ b/graylog2-server/src/main/java/org/graylog2/security/MongoDbSession.java
@@ -64,19 +64,20 @@ public class MongoDbSession extends PersistedImpl {
 
     @SuppressForbidden("Deliberate use of ObjectInputStream")
     public Map<Object, Object> getAttributes() {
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/java/org/graylog2/security/MongoDbSession.java
```diff
@@ -64,36 +64,39 @@ public Map<String, Validator> getEmbeddedValidations(String key) {
...
             // FIXME: This could break backward compatibility if different Java versions are being used.
             final ObjectInputStream ois = new ObjectInputStream(bis);
             final Object o = ois.readObject();
             return (Map<Object, Object>) o;
-        } catch (IOException e) {
-            LOG.error("little io. wow.", e);
-        } catch (ClassNotFoundException e) {
-            LOG.error("wrong thingy in db", e);
+        } catch (Exception e) {
+            LOG.error("Error deserializing binary stream for attributes from Mongo: " + e.getMessage(), e);
```
bernd, line 79:
```suggestion
            LOG.error("Error deserializing binary stream for attributes from Mongo: {}", e.getMessage(), e);
```

graylog2-server/src/main/java/org/graylog2/security/MongoDbSessionDAO.java
```diff
@@ -60,10 +60,7 @@ protected Serializable doCreate(Session session) {
         fields.put("last_access_time", session.getLastAccessTime());
         fields.put("timeout", session.getTimeout());
         Map<String, Object> attributes = Maps.newHashMap();
-        for (Object key : session.getAttributeKeys()) {
-            attributes.put(key.toString(), session.getAttribute(key));
-        }
-        fields.put("attributes", attributes);
+        fields.put("attributes", MongoDbSession.transformAttributes(attributes));
```
bernd, line 63: Aren't we missing the session attributes now? We are transforming an empty map now. Or are there no session attributes on session creation?

bernd (changes requested):

graylog2-server/src/main/java/org/graylog2/security/MongoDbSessionDAO.java
janheise: This is a mistake, thanks for catching that. But also, when I looked into it, there actually were not session attributes at creation time present.

dennisoelkers (approved):
Looks good, fixes the problem for me!

bernd (approved):

Issue #15524, opened 2023-05-17
Deleted/inaccessible entities should not show up in "Last Opened" of Welcome page


dennisoelkers:
## Expected Behavior

When 

## Current Behavior

When a saved search/dashboard was deleted, it should not show up in "Last Opened" anymore.

![image](https://github.com/Graylog2/graylog2-server/assets/41929/ba49f9fa-28e9-450c-bbbb-27ba88c6b7bd)

## Steps to Reproduce (for bugs)

1. Create Dashboard
2. Open it
3. Delete it
4. Navigate to Welcome page

## Your Environment

* Graylog Version:
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

luk-kaminski:
Hi @dennisoelkers,
I cannot reproduce the problem.

In Jan's code we have the following method:
```
    @Subscribe
    public void removeLastOpenedOnEntityDeletion(final RecentActivityEvent event) {
        // if an entity is deleted, we can no longer see it in the lastOpened collection
        if (event.activityType().equals(ActivityType.DELETE)) {
            final var grn = event.grn().toString();
            final var query = new BasicDBObject(LastOpenedForUserDTO.FIELD_ITEMS + "." + LastOpenedDTO.FIELD_GRN, grn);
            final var modifications = new BasicDBObject("$pull", new BasicDBObject(LastOpenedForUserDTO.FIELD_ITEMS, new BasicDBObject(LastOpenedDTO.FIELD_GRN, grn)));
            db.updateMulti(query, modifications);
        }
    }
```

and it seems to remove entries from `last_opened` collection correctly.

Maybe you have some very old items in `last_opened`, removed before this piece of code was introduced?

luk-kaminski:
The only way to reproduce the problem for me was to remove a dashboard from Mongo DB, manually.
I've implemented a quick solution that addresses this rare scenario.

PR #15779, opened 2023-06-20
Removal of inaccessible entities form Start Page lists

luk-kaminski:
## Description
Removal of inaccesible entities form Start Page lists (favourites, last opened and recent activity).
/nocl
Fixes [#15524](https://github.com/Graylog2/graylog2-server/issues/15524)

## Motivation and Context
See [#15524](https://github.com/Graylog2/graylog2-server/issues/15524)

## How Has This Been Tested?
Manually, with manual removal of entity from Mongo DB.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/plugins/views/startpage/StartPageService.java
+++ b/graylog2-server/src/main/java/org/graylog/plugins/views/startpage/StartPageService.java
@@ -27,7 +27,6 @@ import org.graylog.plugins.views.startpage.lastOpened.LastOpenedDTO;
 import org.graylog.plugins.views.startpage.lastOpened.LastOpenedForUserDTO;
 import org.graylog.plugins.views.startpage.lastOpened.LastOpenedService;
 import org.graylog.plugins.views.startpage.recentActivities.RecentActivity;
-import org.graylog.plugins.views.startpage.recentActivities.RecentActivityDTO;
 import org.graylog.plugins.views.startpage.recentActivities.RecentActivityService;
 import org.graylog2.database.PaginatedList;
 import org.graylog2.lookup.Catalog;
@@ -37,6 +36,7 @@ import org.joda.time.DateTimeZone;
 
 import javax.inject.Inject;
 import java.util.List;
+import java.util.Objects;
 
 public class StartPageService {
     private final Catalog catalog;
@@ -68,27 +68,37 @@ public class StartPageService {
                 .items()
                 .stream()
                 .skip((long) (page - 1) * perPage)
-                .map(i -> new LastOpened(i.grn(), catalog.getTitle(i.grn()), i.timestamp()))
-                .filter(lo -> !lo.title().startsWith(Catalog.UNKNOWN_ENTITY_TEXT))
+                .map(i -> {
+                    final Catalog.Entry entry = catalog.getEntry(i.grn());
+                    if (entry.isUnknown()) {
+                        return null;
+                    }
+                    return new LastOpened(i.grn(), entry.title(), i.timestamp());
+                })
+                .filter(Objects::nonNull)
                 .limit(perPage)
                 .toList();
 
         return PaginatedResponse.create("lastOpened", new PaginatedList<>(items, items.size(), page, perPage));
     }
 
-    private String getTitle(RecentActivityDTO i) {
-        return i.itemTitle() == null ? catalog.getTitle(i.itemGrn()) : i.itemTitle();
-    }
-
     public PaginatedResponse<RecentActivity> findRecentActivityFor(final SearchUser searchUser, int page, int perPage) {
         final var items = recentActivityService.findRecentActivitiesFor(searchUser, page, perPage);
         final var mapped = items.stream()
-                 .map(i -> new RecentActivity(i.id(),
-                        i.activityType(),
-                        i.itemGrn(),
-                        getTitle(i),
-                        i.userName(),
-                        i.timestamp())).toList();
+                .map(i -> {
+                    final Catalog.Entry entry = catalog.getEntry(i.itemGrn());
+                    if (entry.isUnknown()) {
+                        return null;
+                    }
+                    return new RecentActivity(i.id(),
+                            i.activityType(),
+                            i.itemGrn(),
+                            entry.title(),
+                            i.userName(),
+                            i.timestamp());
+                })
+                .filter(Objects::nonNull)
+                .toList();
         return PaginatedResponse.create("recentActivity", new PaginatedList<>(mapped, items.pagination().total(), page, perPage));
     }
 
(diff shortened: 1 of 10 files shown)


graylog2-server/src/main/java/org/graylog2/lookup/Catalog.java
```diff
@@ -54,37 +63,24 @@ public Entry load(String id) {
...
-            var item = cache.get(grn.entity());
-            if(item.title() != null) {
-                return item.title();
-            } else {
-                return "Unknown entity: " + grn;
-            }
+            return cache.get(grn.entity());
         } catch (ExecutionException cex) {
-            return "Unknown entity: " + grn;
+            return Entry.UNKNOWN_ENTITY;
```
dennisoelkers, line 82: I am not sure if we should just catch and ignore the exception here. I think we should let it bubble up to the caller and handle it there.

graylog2-server/src/main/java/org/graylog2/lookup/Catalog.java
```diff
@@ -54,37 +63,24 @@ public Entry load(String id) {
...
                         } else {
-                            return new Entry("Unknown entity: " + id, "Unknown entity: " + id);
+                            return Entry.UNKNOWN_ENTITY;
                         }
                     }
                 });
     }
 
-    public String getTitle(final GRN grn) {
+    public Entry getEntry(final GRN grn) {
```
dennisoelkers, line 78: Maybe we could let this return `Optional<Entry>` and turn the `title` attribute of the `Entry` record into an `Optional<String>` as well? This way we could properly reflect the state of the lookup:

  - `getEntry` returns empty optional: No entry in the catalog
  - `getEntry` returns `Entry` record with `title` being empty optional: Entry is present, but entity has no title
  - `getEntry` returns `Entry` record with `title` being set: Entry is present and has title

dennisoelkers (changes requested):
Looks good so far, added two suggestions inline.

graylog2-server/src/main/java/org/graylog2/lookup/Catalog.java
luk-kaminski: That would complicate ALL places where the catalog is used so far. Every invocation would have to check for those 3 possible states and make decision about what to do with it. 
If you agree, I'd leave it as it is because of that.

dennisoelkers: I am seeing three usages of `Catalog#getEntry` at the moment, so I think it is a good time to make the change before we adopt it in more places.

luk-kaminski: Ok, I will generalize Catalog if you believe it may be used in more places.
I will add a small util class that will encapsulate the logic that is now used in those 3 places where catalog is used.

graylog2-server/src/main/java/org/graylog/plugins/views/startpage/title/StartPageItemTitleRetriever.java
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            final Optional<Catalog.Entry> entry = catalog.getEntry(itemGrn);
+            final Optional<String> title = entry.map(Catalog.Entry::title);
+            if (title.isPresent()) {
+                return title;
+            } else {
+                return entry.map(Catalog.Entry::id);
+            }
+        } catch (ExecutionException ex) {
+            return Optional.empty();
```
dennisoelkers, line 50: What could be a reason for this exception that we do not want to bubble up to the caller and instead we just swallow it?

graylog2-server/src/main/java/org/graylog/plugins/views/startpage/StartPageService.java
```diff
@@ -68,26 +68,32 @@ public PaginatedResponse<LastOpened> findLastOpenedFor(final SearchUser searchUs
...
     public PaginatedResponse<RecentActivity> findRecentActivityFor(final SearchUser searchUser, int page, int perPage) {
         final var items = recentActivityService.findRecentActivitiesFor(searchUser, page, perPage);
         final var mapped = items.stream()
-                 .map(i -> new RecentActivity(i.id(),
-                        i.activityType(),
-                        i.itemGrn(),
-                        getTitle(i),
-                        i.userName(),
-                        i.timestamp())).toList();
+                .map(i -> startPageItemTitleRetriever
```
dennisoelkers, line 85: You can also use `Stream#flatMap` here to get rid of the `orElse`/`filter`-combination below.
```suggestion
                .flatMap(i -> startPageItemTitleRetriever
```

graylog2-server/src/main/java/org/graylog/plugins/views/startpage/StartPageService.java
```diff
@@ -68,26 +68,32 @@ public PaginatedResponse<LastOpened> findLastOpenedFor(final SearchUser searchUs
...
+                        .retrieveTitle(i.itemGrn())
+                        .map(title -> new RecentActivity(i.id(),
+                                i.activityType(),
+                                i.itemGrn(),
+                                title,
+                                i.userName(),
+                                i.timestamp()))
+                        .orElse(null)
```
dennisoelkers, line 93: Related to last comment:
```suggestion
                        .stream()
```

dennisoelkers (changes requested):
Added one question and a minor suggestion for improvement.

graylog2-server/src/main/java/org/graylog/plugins/views/startpage/title/StartPageItemTitleRetriever.java
luk-kaminski: LoadingCache:

  /**
   * Returns the value associated with {@code key} in this cache, first loading that value if
   * necessary. No observable state associated with this cache is modified until loading completes.
   *
   * (...)
   * @throws **ExecutionException if a checked exception was thrown while loading the value.** 
   * (...)
   */
  V get(K key) throws ExecutionException;

Our `load` method implementation in Catalog does not throw any checked exceptions.

dennisoelkers: I see. In this case we should just use `LoadingCache#getUnchecked` in `Catalog#getEntry` to simplify the interface and the exception handling.

graylog2-server/src/main/java/org/graylog/plugins/views/startpage/StartPageService.java
luk-kaminski: Don't you have a feeling that it looks tricky and my explicit approach is more clear?

dennisoelkers: No, not at all. This is exactly what `Optional#stream` is made for and IMO it is clearer than unwrapping an `Optional` with `null` (especially considering that `Optional`s were meant to avoid `null`-handling like this).

dennisoelkers (approved):
Perfect! Thanks for your patience! :pray:

PR #15785, opened 2023-06-20
small fixes, additions to rulebuilder

moesterheld:
## Description
- sort function names
- add implicit to_string fragments
- fix string parameter and primary parameter handling in fragments

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/rulebuilder/db/migrations/V20230613154400_AddImplicitToStringFragments.java
```diff
@@ -0,0 +1,126 @@
...
+                        .build())
+                .fragmentOutputVariable("gl2_fragment_substring_results")
+                .build();
+    }
+
+    private static RuleFragment createDateFragment() {
+        return RuleFragment.builder()
+                .fragment("""
+                        let gl2_fragment_date_results = parse_date(
+                          value: to_string(),
```
kodjo-anipah, line 97: to_string(${value}),

moesterheld: good catch

kodjo-anipah (changes requested):
There seems to be an problem with output variables couldn't test the date fragment. 
Also we should deactivate Rulebuilder for FormatDate.java

moesterheld:
> There seems to be an problem with output variables couldn't test the date fragment.

Problem is fixed

> Also we should deactivate Rulebuilder for FormatDate.java

I would have liked to have this available to use `get_date` to parse the date and then use `format_date` to create a string with another format. But I think we have to talk about how to handle dates in the UI in the next cycle as the date won't be available for setting in the form currently. I will deactivate it for now.

kodjo-anipah (approved):
tested all functions and conditions , LGTM

PR #15790, opened 2023-06-21
Fixing the lifecycle phase for antrun in DataNode so that the test-jar is not pulled in during the wrong phase

janheise:
## Description

Fix problems using the graylog2-tests.jar in the data node.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/data-node/pom.xml
+++ b/data-node/pom.xml
@@ -575,7 +575,7 @@
                 <executions>
                     <execution>
                         <id>fix-opensearch-config-permissions</id>
-                        <phase>process-resources</phase>
+                        <phase>prepare-package</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
--- a/graylog2-server/pom.xml
+++ b/graylog2-server/pom.xml
@@ -871,6 +871,22 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- there is a bug in maven causing it to resolve test-jar types
+                 at compile time rather than test-compile. Move the compilation
+                 of the test classes earlier in the build cycle and also generate the -tests.jar earlier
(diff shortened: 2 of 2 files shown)


bernd (changes requested):
@janheise I found the cause of the problem.

We use the [antrun](https://maven.apache.org/plugins/maven-antrun-plugin/) Maven plugin to run a chmod on some OpenSearch distribution files. That plugin uses a dependency resolution scope of `test`.

See: https://github.com/apache/maven-antrun-plugin/blob/d9667e3990277ea9c3860858713fc25cda2b4823/src/main/java/org/apache/maven/plugins/antrun/AntRunMojo.java#L67

Since we currently execute the antrun plugin in the `process-resources` phase, we pull in the `test` dependency resolution too early and the compile fails.

Moving the antrun plugin execution into a phase after the `compile` phase seems to fix the problem. The `test-compile` phase has run at that time, and the test dependencies are present in the Maven reactor.

```diff
diff --git data-node/pom.xml data-node/pom.xml
index 280e8b1069..67430dfc90 100644
--- data-node/pom.xml
+++ data-node/pom.xml
@@ -572,13 +572,13 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
                 <executions>
                     <execution>
                         <id>fix-opensearch-config-permissions</id>
-                        <phase>process-resources</phase>
+                        <phase>prepare-package</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
                         <configuration>
                             <target>
                                 <!-- Some config files in opensearch-*/config/ use 660 permissions. That's a problem

```

todvora:
Awesome, thank you all !!!

PR #15801, opened 2023-06-22
Ensure className prop is not undefined when using it in template string.

linuspahl:
## Motivation and Context

This PR is fixing a few cases where `className` is used in a template string, but could be undefined.
We ensure it no longer can be undefined. This is a good practice since `${undefined}` result in the following string: `undefined`.

/nocl


grotlue (approved):
Looks good to me!

PR #15804, opened 2023-06-23
add time size based rotation to validator (#15750)

kodjo-anipah:
backport of https://github.com/Graylog2/graylog2-server/pull/15750


patrickmann (approved):
smoke-tested on 5.1: time/size-optimizing is available and I can successfully create an index with it.

PR #15808, opened 2023-06-23
Adding action to cleanup & rebuild all index ranges.

### LINKED ISSUE #11919: Manually run index ranges cleanup through the GUI and api endpoint
## What?

As a Graylog user I would like to manually run indices cleanup via the GUI rather than periodically through a job. Being able to hit an API endpoint would also be helpful.

## Why?

This would be helpful when reindexing things directly into Elasticsearch or when restoring archives for example. Currently when we let the job run, we have to wait by default one hour unless we specify otherwise in the config file.

See HS-575331695 for further context.

dennisoelkers:
## Motivation and Context

This PR is doing two things:

  - Cleans up index ranges when `IndexRangesResource#rebuild` is triggered
  - Adds a maintenance dropdown with a "Cleanup & recalculate all index ranges" button to the general index sets overview page

This helps a lot with getting index ranges synced after Opensearch was wiped/updated, indices were manually deleted etc.

What is missing and should be implemented in a follow-up PR:
  - Also including index range cleanup for the "Recalculate index ranges" action for a specific index set
  - Refactor the cleanup periodical for easier reuse
  - Wrap it in a system job instead of a sync execution

Fixes #11919.

cleanup-all

## Types of changes

## Checklist:


linuspahl (approved):
Frontend looks good to me and the index recalculation seems to work as expected.

mpfz0r (approved):
Works and LGTM. I agree about the follow up PRs. The fact that the cleanup is done asynchronously via the event bus makes a bit nervous

PR #15830, opened 2023-06-28
Limitting retrieval of all entities from DB. Using memoize instead of…

luk-kaminski:
… innefficient cache.

## Description
Previous implementation had a bug - it fetched data about all entities in the DB for each access to a new entity id in a quickly expiring cache (1s).

```                .build(new CacheLoader<>() {
                    @Override
                    public Optional<Entry> load(String id) {
                        var catalog = contentPackService.getEntityExcerpts();
                        (...)
                    }
                });
```
This approach has been modified.
getEntityExcerpts() method is still called, which is wrong, but its results are at least memoized, so that all entity titles are resolved with this one DB call, not many of them, as before.
/nocl

## Motivation and Context
Improve performance before a better approach is agreed on and implemented.

## How Has This Been Tested?
Manually and by running modified unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/full-backend-tests/src/test/java/org/graylog/plugins/views/StartPageLastOpenedIT.java
+++ b/full-backend-tests/src/test/java/org/graylog/plugins/views/StartPageLastOpenedIT.java
@@ -56,10 +56,13 @@ public class StartPageLastOpenedIT {
         api.users().createUser(user);
     }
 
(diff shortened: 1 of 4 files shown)


janheise (approved):
LGTM!

PR #15840, opened 2023-06-29
Let server container fail on guice errors

bernd:
This change lets tests fail early on guice errors instead of waiting for the timeout.

/nocl


janheise (approved):
LGTM!

PR #15842, opened 2023-06-29
feat: Modifies route active check to account for nested routes

zeeklop:
Updates the check for active route to account for nested routes

/nocl

## Description

With this modification we can add the `end` key to a route to signal that it has to match exactly the request route. This helps to account for routes like `/security` and `/security/sigma` where before it will always select the first one. With this modifications we can add `end` to `/security` so it will match that route only when the request route matches exactly. The rest work as it was before.

## Motivation and Context

Consolidate security module with the rest of the application

## Types of changes

## Checklist:


danotorrey (approved):
LGTM and tests successfully.

ryan-carroll-graylog (approved):
Looks good and tests successfully

PR #15870, opened 2023-07-04
Add aws-msk-iam-auth

thll:
Enables authentication with IAM for AWS MSK.
See https://github.com/aws/aws-msk-iam-auth

/nocl For now we only need this internally


kodjo-anipah (approved):
LGTM

PR #15885, opened 2023-07-05
Only display query input suggestions while input has focus.

linuspahl:
**Please note:** this  PR needs a backport for `5.0` and `5.1`.

## Motivation and Context

Before this change we were showing the query input suggestions on every `tokenizerUpdate`. This event will be triggered even when we extend the query by using the "Add to query" field value action.

With this change we also consider if the query input is focussed and only show suggestions in this case.

Fixes https://github.com/Graylog2/graylog2-server/issues/12165

## Types of changes


dennisoelkers (approved):
Perfect fix, works well! :pray:

PR #15891, opened 2023-07-06
Fix join pipeline function parameter name

### LINKED ISSUE #11868: Pipeline Rule Function substring() utilizes Keyword "end"
When creating a pipeline rule, using the substring() function, there is no way to explicitely state the parameter names (i.e. value:, start:, end:) when using an value for end. The "end:" is interpreted in the same way as the "end" finishing the rule (including syntax highlighting), which leads to the rule no longer be saveable (without any error indicator, which led to some confusion on my side).

## Expected Behavior
Pipeline rules utilizing substring() while explicitely stating parameter names should be working.

## Current Behavior
When using substring() with explicit stating of parameter names including "end:", the rule can no longer be saved, without any error indication. There also is no syntax error shown when using the function this way. Only the syntax highlighting gives a hint to the problem.

## Possible Solution
Rename the "end:" parameter to something else which is not a reserved keyword. Alternatively: change the behavior of the pipeline rule editor to not fail due to usage of explicit parameter names.

1. create a new pipeline rule
2. use the substring() function with all parameters explicitely stated
`let result = substring(value:"some text here", start:3, end:5);`
3. try to save the rule

## Context
This problem occured while creating a lengthy text manipulation rule with multiple usages of the substring() function, while still keeping up the readability (by explicitely stating parameter names). Although not using parameter names fixes the problem, this does not help keeping the rule readable and understandable.

## Your Environment
* Graylog Version: 4.2.4+b643d2b

mpfz0r:
We cannot use `end` because it's a reserved keyword in the pipeline language and causes a syntax error.

Instead of adding special handling for the `end` keyword to the pipeline rule language we just rename the parameter to `endIndex`.

This shouldn't break existing rules, because they could've never been saved.

Fixes #11868

The substring function got already fixed by #15470 and should be included in a backport.

### DIFF
--- a/changelog/unreleased/issue-11868.toml
+++ b/changelog/unreleased/issue-11868.toml
@@ -1,0 +1,10 @@
+# PLEASE REMOVE COMMENTS AND OPTIONAL FIELDS! THANKS!
+
+# Entry type according to https://keepachangelog.com/en/1.0.0/
+# One of: a(dded), c(hanged), d(eprecated), r(emoved), f(ixed), s(ecurity)
+type = "fixed"
+message = "Fix join() and substring() pipeline function parameter names."
+
+issues = ["11868"]
+pulls = [""]
+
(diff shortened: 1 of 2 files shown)


kodjo-anipah (approved):
LGTM

PR #15902, opened 2023-07-07
Remove default data_dir path configuration

### LINKED ISSUE #13673: Remove default values for path configurations
## What?

Remove the default values for the `bin_dir`, `data_dir`, and `plugin_dir` configuration options.

- Remove the default values for the options in the `PathConfiguration` class
- Keep the option uncommented in the default `misc/graylog.conf` file, but remove their values
- Document the changes in `UPGRADING.md`
- Update manual setup documentation: https://docs.graylog.org/docs/manual-setup

## Why?

We had a customer that created a custom `graylog.conf` and forgot to include the `*_dir` settings. The Graylog server used the default values instead and created the journal directory in an unexpected directory.

Removing the default values avoids misconfiguration because the server complains on startup when required settings are missing.

Our operating system packages and the container images set these values to working defaults, so removing the default values from the code doesn't affect these common installation methods.
Only users that run Graylog using the tarballs (not recommended anymore) and developers have to adjust their configuration files.

AntonEbel:
## Description

- removed default value for `data_dir` in the code base. (Clarified: The values must still be present in the default `graylog.conf`)
- made `content_packs_dir`, `geo_ip_processor_s3_download_location` and `message_journal_dir` as subdirectories of `data_dir` if not set explicitly

## Motivation and Context

fixes #13673


graylog2-server/src/main/java/org/graylog2/bindings/NamedInjectConfigParametersModule.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static com.github.joschi.jadconfig.ReflectionUtils.getAllFields;
+import static com.github.joschi.jadconfig.ReflectionUtils.getFieldValue;
+import static com.google.inject.name.Names.named;
+import static org.graylog2.shared.utilities.StringUtils.f;
+
+public class NamedInjectConfigParametersModule extends AbstractModule {
```
thll, line 42: What do you think about naming this just `ConfigParametersModule`?

graylog2-server/src/main/java/org/graylog2/bindings/ParameterNamedInjectOverride.java
```diff
@@ -0,0 +1,28 @@
...
+package org.graylog2.bindings;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface ParameterNamedInjectOverride {
```
thll, line 26: What do you think about naming this `NamedBindingOverride` instead?

graylog2-server/src/main/java/org/graylog2/bindings/NamedInjectConfigParametersModule.java
thll, line 42: Can we add some documentation?

graylog2-server/src/main/java/org/graylog2/bindings/ParameterNamedInjectOverride.java
thll, line 26: Can we add some documentation?

thll (commented):
In general I think the approach looks good.

I don't particularly like that we now have `plugin_dir` and `bind_dir` as required parameters, but that of course was the requirement of the original issue.

We will have to adjust some plugin code before merging this change. At least the enterprise integrations plugin.will be affected.

graylog2-server/src/main/java/org/graylog2/bindings/NamedInjectConfigParametersModule.java
```diff
@@ -0,0 +1,121 @@
...
+        Map<String, Method> methodOverrides = ReflectionUtils.getAllMethods(bean.getClass()).stream()
+                .filter(obj -> Objects.nonNull(obj.getAnnotation(ParameterNamedInjectOverride.class)))
+                .collect(Collectors.toMap(obj -> obj.getAnnotation(ParameterNamedInjectOverride.class).value(), Function.identity()));
+
+        for (Field field : fields) {
+            final Parameter parameter = field.getAnnotation(Parameter.class);
+
+            if (parameter != null) {
+                try {
+                    final TypeLiteral typeLiteral = TypeLiteral.get(field.getGenericType());
```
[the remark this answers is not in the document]
mpfz0r: `final TypeLiteral<?>` fixes this

graylog2-server/src/main/java/org/graylog2/bindings/NamedInjectConfigParametersModule.java
mpfz0r: That name is derived from `com.github.joschi.jadconfig.guice.NamedConfigParametersModule`

How about `NamedConfigParametersOverrideModule`? 

However the name, I think some java doc would clarify what it does

mpfz0r:
> I don't particularly like that we now have `plugin_dir` and `bind_dir` as required parameters, but that of course was the requirement of the original issue.

+1 
I think we should discuss that once more. I don't see any harm with these directories, because there is no data being created there.

AntonEbel:
> > I don't particularly like that we now have `plugin_dir` and `bind_dir` as required parameters, but that of course was the requirement of the original issue.
> 
> +1 I think we should discuss that once more. I don't see any harm with these directories, because there is no data being created there.

@bernd What do you think Bernd, should we still remove the default values for `plugin_dir` and `bind_dir`?

bernd:
> > > I don't particularly like that we now have `plugin_dir` and `bind_dir` as required parameters, but that of course was the requirement of the original issue.
> > 
> > 
> > +1 I think we should discuss that once more. I don't see any harm with these directories, because there is no data being created there.
> 
> @bernd What do you think Bernd, should we still remove the default values for `plugin_dir` and `bind_dir`?

@AntonEbel No, let's keep the default values for `plugin_dir` and `bin_dir`. This is also what Sebastian said in the original issue: https://github.com/Graylog2/graylog2-server/issues/13673#issuecomment-1302432427

misc/graylog.conf
```diff
@@ -518,6 +517,7 @@ message_journal_enabled = true
 #   If you create a separate partition for the journal files and use a file system creating directories like 'lost+found'
 #   in the root directory, you need to create a sub directory for your journal.
 #   Otherwise Graylog will log an error message that the journal is corrupt and Graylog will not start.
+# Default: <data_dir>/journal
 message_journal_dir = data/journal
```
mpfz0r, line 521:
```suggestion
#message_journal_dir = data/journal
```

Couldn't we comment that out?

mpfz0r (changes requested):
A few comments

graylog2-server/src/main/java/org/graylog2/bindings/NamedConfigParametersOverrideModule.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static com.github.joschi.jadconfig.ReflectionUtils.getAllFields;
+import static com.github.joschi.jadconfig.ReflectionUtils.getFieldValue;
+import static com.google.inject.name.Names.named;
+import static org.graylog2.shared.utilities.StringUtils.f;
+
+/**
+ * A Guice module which registers all fields of the provided objects annotated with {@link Parameter}
+ * as named bindings and optionally also adds bindings of the object instances themselves.
+ * For more flexibility, getter methods can also be used, which must be annotated with {@link NamedBindingOverride}.
```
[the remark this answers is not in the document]
AntonEbel: Don't think so. I have added this sentence:
`This class is based off {@link com.github.joschi.jadconfig.guice.NamedConfigParametersModule}`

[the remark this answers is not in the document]
mpfz0r: ah, sorry. the diff was underneath that comment, so I missed it.
It's usually good to mark comments as resolved :)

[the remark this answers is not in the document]
AntonEbel: sorry, missed that

mpfz0r (approved):
LGTM

Issue #15678, opened 2023-06-02
Replay link within the default notification template


neddy0graylog:
Include the "Replay Search" link within the default email template, it's easy enough to hardcode in the server address but would be useful to be able to call "http_external_uri" in the email template context so it's more dynamic. The link it's self would be useful when needing to react to the alert and jump straight into Graylog.

`<tr><td>Alert replay</td><td>https://192.168.64.10:9000/alerts/${[event.id](http://event.id/)}/replay-search</td></tr>`
<img width="722" alt="Screenshot 2023-05-31 at 15 20 25" src="https://github.com/Graylog2/graylog2-server/assets/108661699/94ecbe47-1a53-482e-a8b3-bbf46c2d1a1b">

coffee-squirrel:
Case # 1657409922.

PR #15922, opened 2023-07-10
Include link to replay an alert in alert mail notification template.

linuspahl:
## Motivation and Context

With this PR we are including a "Alert Replay" link in the template for alert mail notifications.

Fixes https://github.com/Graylog2/graylog2-server/issues/15678

### DIFF
--- a/changelog/unreleased/issue-15678.toml
+++ b/changelog/unreleased/issue-15678.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Include link to replay an alert in alert mail notification template."
+
+issues = ["15678"]
+pulls = ["15922"]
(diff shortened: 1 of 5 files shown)


luk-kaminski (approved):
LGTM, thx!

HenryTheSir:
hi @linuspahl ,

while building this manuel in current release, i observed that the wrong id field gets used in this PR.

In my opinion the event_definition_id needs to be set in the url and not event.id
And the URL seams to be wrong to? there needs to be an additional definitions after /alerts/ 
best regards

linuspahl:
Hi @HenryTheSir!

just to be sure I understand you correctly. Did you build this functionality manually and are wondering why this PR implements the link differently or did the link this PR implements result in an error for you?

It looks like you are creating a link to replay the search for the event definition of an alert, while this PR implements a link to replay the alert (which also considers e.g. the aggregation time range of an alert). 

Best regards

HenryTheSir:
My bad - sorry this function was completly unaware to me! 
You are right and URL and ID are corret in the url.

best regards

PR #15933, opened 2023-07-11
Only trigger query input autocompletion popover when it is hidden.

### LINKED ISSUE #15718: Field name (or value) suggestion list jumps to first item, even if an item is already selected/highlighted
## Expected Behavior

If a user has "arrowed down" to highlight a field name/value suggestion, and the field name/value suggestion list (or the suggestions up to and including the highlighted field name/value) doesn't change, the highlighted name/value should not reset to the first item in the list.

Alternatively: if the user has highlighted a field name/value suggestion the suggestion list should not be modified/updated.

## Current Behavior

Given I have fields `source_priority`, `source_type`, and `source` (among others), if I type `so` and arrow down to highlight `source` (skipping `source_priority` and `source_type`) I see the highlighted suggestion jump up to the first item in the list (presumably after `/api/search/validate` completes-- ~80ms).  Something similar happens for field value suggestions.

image

I believe this started after we moved from `5.0.x` to `5.1.x`.

1. Have a `5.1.2` environment
3. Have fields with a common prefix
4. Open your browser dev tools and watch XHR requests
5. Start typing that field name prefix, and after the initial suggestions appear arrow down past the first entry and wait
6. Note that after the `/api/search/validate` request completes the first suggestions gets highlighted

## Context

This makes working with field name/value suggestions annoying, since the UI wipes out the suggestion I arrowed down to.

## Your Environment

* Graylog Version: `5.1.2`
* Java Version: `17.0.7`
* Elasticsearch Version: `7.10.2`
* MongoDB Version: `5.0.14`
* Browser version: Firefox 114, Chrome 114

linuspahl:
**Please note:** This PR needs a backport for `5.0` and `5.1`.

## Motivation and Context

As described in #15718 it is possible that the focus for an option in the query input autocompletion gets lost.
This problem occurs because we triggered `editor.execCommand('startAutocomplete');`, even when the autocompletion popover is already visible. Related to https://github.com/Graylog2/graylog2-server/pull/14233.

Fixes #15718

## Types of changes

### DIFF
--- a/changelog/unreleased/issue-15718.toml
+++ b/changelog/unreleased/issue-15718.toml
@@ -1,0 +1,5 @@
+type = "f" # One of: a(dded), c(hanged), d(eprecated), r(emoved), f(ixed), s(ecurity)
+message = "Fixing problem with focus loss in query input autocompletion."
+
+issues = ["15718"]
+pulls = ["15933"]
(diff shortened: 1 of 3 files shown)


dennisoelkers (approved):
Look good, works great!

PR #15948, opened 2023-07-13
Migrating legacy templates to index templates.

### LINKED ISSUE #15782: Migrate usage of legacy templates to index/composable templates.
[Index Templates](https://opensearch.org/docs/latest/im-plugin/index-templates/) have been introduced in Elasticsearch 7.8. The previous "legacy templates" are already deprioritized during index template selection and will be removed at some point, so we should prepare our code to make use of index templates.

dennisoelkers:
## Description

This PR is migrating the usage of [legacy index templates](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-templates-v1.html) to [composable index templates](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/index-templates.html) introduced in ES 7.8.

An important consequence of this is that users who have configured custom legacy templates matching their index patterns need to migrate them to composable index templates as well, as legacy templates have lower priority and could probably not match.

This PR is introducing a feature flag (`composable_index_templates`) which toggles between legacy and composable index templates.

## Motivation and Context

Fixes #15782.

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#5470

## Types of changes

## Checklist:


bernd (commented):
@dennisoelkers Please add an entry to the `UPGRADING.md` file for this change. Thanks!

bernd (commented):
Please check with @Graylog2/seccon about their index template usage. The change probably requires adjustments for the Illuminate product. I don't know how many older Graylog releases new Illuminate releases support, but this might require handling in Illuminate.

danotorrey:
Thanks @bernd. Continuing the Illuminate discussion on the corresponding enterprise PR: https://github.com/Graylog2/graylog-plugin-enterprise/pull/5470#issuecomment-1686978652

graylog2-server/src/main/java/org/graylog2/indexer/indices/Template.java
```diff
@@ -0,0 +1,44 @@
...
+        public Mappings(Map<String, Object> m) {
+            super(m);
+        }
+    }
+
+    public static class Settings extends HashMap<String, Object> {
+        public Settings(Map<String, Object> m) {
+            super(m);
+        }
+    }
```
luk-kaminski, line 35: What about replacing those static classes with records?

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/Elasticsearch7Module.java
```diff
@@ -42,19 +42,25 @@
...
     protected void configure() {
         bindForSupportedVersion(StreamsForFieldRetriever.class).to(StreamsForFieldRetrieverES7.class);
         bindForSupportedVersion(CountsAdapter.class).to(CountsAdapterES7.class);
         bindForSupportedVersion(ClusterAdapter.class).to(ClusterAdapterES7.class);
-        bindForSupportedVersion(IndicesAdapter.class).to(IndicesAdapterES7.class);
+        if (useComposableIndexTemplates) {
+            bindForSupportedVersion(IndicesAdapter.class).to(IndicesAdapterES7.class);
+        } else {
+            bindForSupportedVersion(IndicesAdapter.class).to(LegacyIndicesAdapterES7.class);
+        }
```
luk-kaminski, line 62: There is A LOT of code duplication in `IndicesAdapterES7` and `LegacyIndicesAdapterES7`.
They are different only when it comes to templates.
What about injecting the feature flag directly to the adapter and keeping just single `IndicesAdapterES7` class?
(You could consider moving only template-related code outside of `IndicesAdapterES7` class later on, and spreading it to 2 different classes).

PS
It is the same in OS version.

dennisoelkers: I considered that too, but I think this way it will be easier to just get rid of `LegacyIndicesAdapterES7` when we remove the feature flag again.

luk-kaminski: I'd prefer smth like that:
```
public class IndicesAdapterES7 implements IndicesAdapter {
   
    (...)

    private TemplateHandler templateHandler;

    @Inject
    public IndicesAdapterES7(ElasticsearchClient client,
                             StatsApi statsApi,
                             CatApi catApi,
                             ClusterStateApi clusterStateApi,
                             TemplateHandler templateHandler) {
        this.client = client;
        this.statsApi = statsApi;
        this.catApi = catApi;
        this.clusterStateApi = clusterStateApi;
        this.templateHandler = templateHandler;
    }

    (...)
```
And binding decision for `TemplateHandler`, which, without looking into details of the code, should be as easily removable as current solution, but with less code duplication.

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
```diff
@@ -43,26 +44,23 @@ protected Map<String, Object> analyzerKeyword() {
...
-        final Map<String, Object> settings = Collections.singletonMap(
-                "analysis", Collections.singletonMap("analyzer", analyzerKeyword())
-                );
-        final Map<String, Object> mappings = mapping(analyzer);
+    public Template messageTemplate(final String indexPattern, final String analyzer, final Long order) {
+        var settings = new Template.Settings(Collections.singletonMap(
+                "index", Collections.singletonMap(
+                        "analysis", Collections.singletonMap("analyzer", analyzerKeyword())
+                )
+        ));
```
luk-kaminski, line 52: `Map.of()` can be used instead of `Collections.singletonMap()`

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/Elasticsearch7Module.java
dennisoelkers: Good idea! Did that in [b39f309](https://github.com/Graylog2/graylog2-server/pull/15948/commits/b39f30939e0229d9b9e7e587943b5cad84b77829).

graylog2-server/src/test/java/org/graylog2/indexer/EventsIndexMappingTest.java
```diff
@@ -47,59 +49,61 @@ void createsValidMappingTemplates(final String versionString) throws Exception {
...
+        at.jsonPathAsString("$.properties.timerange_start.format").isEqualTo(DATE_FORMAT);
+        at.jsonPathAsString("$.properties.timerange_end.type").isEqualTo("date");
+        at.jsonPathAsString("$.properties.timerange_end.format").isEqualTo(DATE_FORMAT);
+        at.jsonPathAsString("$.properties.streams.type").isEqualTo("keyword");
+        at.jsonPathAsString("$.properties.source_streams.type").isEqualTo("keyword");
+        at.jsonPathAsString("$.properties.message.type").isEqualTo("text");
+        at.jsonPathAsString("$.properties.message.analyzer").isEqualTo("standard");
+        at.jsonPathAsBoolean("$.properties.message.norms").isFalse();
+        at.jsonPathAsString("$.properties.message.fields.keyword.type").isEqualTo("keyword");
+        at.jsonPathAsString("$.properties.source.type").isEqualTo("keyword");
```
luk-kaminski, line 100: I know it was like that before... and may be a task for different PR, in the cooldown phase... But wouldn't it be nicer to somehow construct the whole expected object and use it in assertion, instead of working with so many elements one by one?

dennisoelkers: I am not sure. Working with the exact, complete object is always prone to make the test brittle when unrelated parts are changed.

graylog-storage-elasticsearch7/src/main/java/org/graylog/storage/elasticsearch7/Elasticsearch7Module.java
luk-kaminski: Thanks!

graylog-storage-opensearch2/src/main/java/org/graylog/storage/opensearch2/IndicesAdapterOS2.java
```diff
@@ -204,19 +212,32 @@ public Map<String, Object> getIndexMetaData(@Nonnull String index) {
...
-        final PutIndexTemplateRequest request = new PutIndexTemplateRequest(templateName)
-                .source(template);
+    public boolean ensureIndexTemplate(String templateName, Template template) {
+        var serializedMapping = serialize(template.mappings());
+        var settings = org.graylog.shaded.opensearch2.org.opensearch.common.settings.Settings.builder().loadFromMap(template.settings()).build();
+        var osTemplate = new org.graylog.shaded.opensearch2.org.opensearch.cluster.metadata.Template(settings, serializedMapping, null);
+        var indexTemplate = new ComposableIndexTemplate(template.indexPatterns(), osTemplate, null, template.order(), null, null);
+        var request = new PutComposableIndexTemplateRequest()
+                .name(templateName)
+                .indexTemplate(indexTemplate);
```
luk-kaminski, line 222: Why don't you use `IndexTemplateAdapter` approach in OS codebase?

graylog2-server/src/main/java/org/graylog2/indexer/indices/Template.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static Template create(List<String> indexPatterns, Mappings mappings, Long order, Settings settings) {
+        return new Template(indexPatterns, mappings, order, settings);
+    }
+
+    public static Template create(String indexPattern, Mappings mappings, Long order, Settings settings) {
+        return Template.create(Collections.singletonList(indexPattern), mappings, order, settings);
+    }
```
luk-kaminski, line 43: Do we believe that records' constructors are worse than static create methods?
If not, I'd remove both create methods, which are a wrappers that don't bring anything new.

dennisoelkers: Static create methods can be refactored more easily, so I would generally favor them over constructors. I would not say we need to use them everywhere though, just in cases where it makes sense (with optional fields, where you do not want to pass  `null` everywhere or wrapping a type like the `String` -> `List<String>` here).

graylog-storage-opensearch2/src/main/java/org/graylog/storage/opensearch2/IndicesAdapterOS2.java
dennisoelkers: I forgot :) Pushed that now.

graylog-storage-opensearch2/src/main/java/org/graylog/storage/opensearch2/IndexTemplateAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.storage.opensearch2;
+
+import org.graylog2.indexer.indices.Template;
+
+public interface IndexTemplateAdapter {
```
luk-kaminski, line 21: This interface has no OS/ES specific code, so it can be moved somewhere else and does not need to be copied in elasticsearch package.

Some parts of duplicated code (i.e. `serialize...` methods) can be moved to this class than, to avoid duplication.

luk-kaminski (approved):
Looks good, thanks!

danotorrey (approved):
LGTM. I did not test support for this comprehensively, but I think this is good to get merged so folks can start testing with it using the feature flag.

paladox:
I'm a little confused with this. Does this mean composable index templates is used by default rather than the legacy one? There's a feature flag but I'm not sure if it automatically changes. I'm also not sure how you set it in graylog server config.

The title just makes it sounds like composable index templates is used by default, but the feature flag suggest it isn't and shouldn't be used in prod?

PR #14222, opened 2022-12-15
Refactor NodeId class and all usages

todvora:
The original NodeId class is doing too much - reading, geneating, persisting and providing node id. It makes sense to separate the persistence logic and the value itself. Additionally, NodeId is now ~~interface~~ abstract class (compatibility reasons), which is easy to use and provides clear communication and mocking options in tests. In the future 6.0 release, we should replace this abstract class with an interface, to fully separate the contract and the implementation.

## Motivation and Context
Code cleanup, separated responsibilities, better interfaces, removed many mocks in tests which increases readability.

## How Has This Been Tested?
Existing set of unit and integration tests, added unit test for the FilePersistedNodeIdProvider

/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#4470,Graylog2/graylog-plugin-integrations#1256,Graylog2/graylog-plugin-enterprise-integrations#957

/nocl

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #15950, opened 2023-07-13
Fix journal show and journal decode command

mpfz0r:
Add missing guice bindings.

Refs #14222 
Refs #14459

/nocl


thll (approved):

thll:
Thanks for the fixes!

Issue #11402, opened 2021-10-01
Timerange summary is not updated for keyword time ranges when search is reexecuted


dennisoelkers:
The time range summary of the search bar controls shows the effective, absolute time range for a keyword time range. The time range boundaries (start/end) are determined once, when the time range is configured. That means that start/end of the time range do not change, even if the search is reexecuted and different effective start/end dates are used:

![timerange-preview-not-updating](https://user-images.githubusercontent.com/41929/135595987-aaf6869b-fa7f-49a7-9171-4b5af687e38a.gif)

Issue #11401, opened 2021-10-01
Timerange summary should show keyword used for keyword time ranges


dennisoelkers:
Currently, the timerange summary (part of the search bar) do show the configured timerange as an absolute time range when a keyword time range is used:

![image](https://user-images.githubusercontent.com/41929/135595466-9ba75238-99d7-44f6-ba49-9b71a9a28f97.png)

While showing the effective time range is useful, we should also show the keyword used (in this example "last five minutes").

PR #15961, opened 2023-07-14
Display actual keyword of keyword time range in search bar.

linuspahl:
This PR is based on https://github.com/Graylog2/graylog2-server/pull/15947 which needs to be merged first.

## Motivation and Context

This PR contains some changes regarding the way we display the search time range. We are:
- displaying the actual keyword time range in the search bar (https://github.com/Graylog2/graylog2-server/issues/11401). 
image
Before we were displaying the "estimated" absolute time range, which could be outdated (https://github.com/Graylog2/graylog2-server/issues/11402), since it depends on the time a user selected the keyword time range.

- displaying the effective search execution time range in the search bar:
image
This will not only be helpful for keyword time ranges, but also relative time ranges. We can consider displaying this information below the search bar. I included it in this section since it is related to the search execution meta data.

- removing the time range preview in the time range picker, since its content is redundant with the content of the tab below.

- displaying a preview of the resulting effective time range for a keyword time range in the time range picker.
image

## How Has This Been Tested?

Fixes https://github.com/Graylog2/graylog2-server/issues/11401
Fixes https://github.com/Graylog2/graylog2-server/issues/11402


coffee-squirrel:
Just FWIW-- as someone that keeps the search sidebar closed most of the time, I'd be interested in having the effective search execution time range info closer to the search bar.  Maybe it could also be in a tooltip set on the value (e.g. `Last 5 minutes`)?

linuspahl:
@coffee-squirrel Thanks for your feedback. The current placement will most likely be just a temporary solution. The tooltip could be a solution. We are currently considering displaying the execution time, effective time range and messages count in a row next to the search bar.

dennisoelkers (approved):
Looks great, works fine!

PR #15974, opened 2023-07-17
Fix tests for OpenSearch and blocked index/cluster after watermark hit

janheise:
OpenSearch 2.6 has a change that after a watermark for the full cluster fails, only resetting the index block is not enough.
cluster.blocks.create_index needs also to get reset.

/nocl


janheise:
solved with https://github.com/Graylog2/graylog2-server/pull/15996

PR #15991, opened 2023-07-19
Return Set instead of Collection from Message#getStreamIds

thll:
Return the more specific collection type. This is a non-breaking change and allows us to use the returned Set without casting or copying.

/nocl


bernd:
@thll @mpfz0r Please create an entry in `UPGRADING.md`. This is a breaking change for 3rd party plugins.

PR #16028, opened 2023-07-24
Template for optional archiver config file

patrickmann:
/nocl unreleased feature
Relates to Graylog2/graylog-plugin-enterprise#5424
We introduce an optional config file for the archiver. A template file helps users discover available configuration settings.


kodjo-anipah (changes requested):
catalog_refresh_interval is a server side configuration

Missing: 
data_dir
kafka_commit_interval,
kafka_custom_properties

Should we make the default config work for local environment?

PR #16035, opened 2023-07-25
add event limit for filter event definition

AntonEbel:
This PR introduces an event limit for the "Filter and Aggregation" event definition. However, only if the user selects "Filter has results". 

How this feature behaves:
- Existing event definitions are not migrated and the event limit 0 is shown in the UI.
- When the event limit is reached, the remaining messages are discarded and a notification is shown that the limit has been reached.
- the validation of the event limit works as follows:
	- the event limit must be greater than 0 (except for existing event definitions).
	- if for an existing event definition the event limit has been set, there is no way back and the user must set a value greater than 0 on the next update.
	- The event limit must be smaller than event_definition_max_event_limit

## Motivation and Context

Graylog2/graylog-plugin-enterprise#5382

## Types of changes


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -194,37 +203,40 @@ public ValidationResult validate() {
         if (!series().isEmpty() && isConditionsEmpty()) {
             validationResult.addError(FIELD_CONDITIONS, "Aggregation with series must also contain conditions");
         }
+        if (eventLimit() <= 0) {
+            validationResult.addError(FIELD_EVENT_LIMIT, "Event limit must be greater than 0.");
```
mpfz0r, line 207: How does this affect existing event definitions?
Do they need to pick a limit before they can be updated?

graylog2-server/src/main/resources/org/graylog2/freemarker/templates/PLAINTEXT/event_limit_reached.ftl
```diff
@@ -0,0 +1,7 @@
+<#if _title>
+    Event limit reached
+</#if>
+
+<#if _description>
+    Event limit »${event_limit}« reached for event definition »${event_definition_title}«. Increase event limit or aggregate events.
```
mpfz0r, line 6: same as above

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -194,37 +203,40 @@ public ValidationResult validate() {
...
+        }
```
moesterheld, line 208: Maybe it would make sense to also set some kind of upper boundary. At least it should be Integer.MAX_VALUE

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
moesterheld, line 1: should the eventLimit also be added to the `AggregationEventProcessorConfigEntity` for content packs?

AntonEbel: good catch. I fixed it.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: It is not possible to check for Integer.MAX_VALUE here, because an exception is thrown beforehand. The error message then looks like this:
Screenshot 2023-08-08 at 17 37 42

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: Yes, that is the current behavior.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: @mpfz0r @EvaZg Should we set a maximum limit value that is smaller than Integer.MAX_VALUE?

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
mpfz0r: It would be nice if this wouldn't throw an error. 
We could migrate them to a new upper limit (see below) I suggest `100_000` events. 
We could check with product if that's a good upper boundary

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: I have added a configurable maximum limit for events

graylog2-server/src/main/java/org/graylog/events/processor/EventDefinitionConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.events.processor;
+
+import com.github.joschi.jadconfig.Parameter;
+import com.github.joschi.jadconfig.validators.PositiveIntegerValidator;
+
+public class EventDefinitionConfiguration {
+
+    @Parameter(value = "event_definition_max_event_limit", validators = PositiveIntegerValidator.class)
```
mpfz0r, line 24: we usually add settings to `misc/graylog.conf` as well

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -236,12 +244,32 @@ private void filterSearch(EventFactory eventFactory, AggregationEventProcessorPa
...
+        } catch (EventLimitReachedException e) {
+            notificationService.publishIfFirst(notificationService.buildNow()
+                    .addType(EVENT_LIMIT_REACHED)
+                    .addKey(eventDefinition.id())
+                    .addDetail("event_definition_title", eventDefinition.title())
+                    .addDetail("event_limit", config.eventLimit())
+                    .addSeverity(Notification.Severity.NORMAL)
+            );
+
+            LOG.debug("Event limit reached at {} for '{}' event definition.", config.eventLimit(), eventDefinition.title());
```
mpfz0r, line 271:
```suggestion
            LOG.debug("Event limit reached at {} for '{}/{}' event definition.", config.eventLimit(), eventDefinition.title(), eventDefinition.id());
```

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
mpfz0r: @moesterheld please re-review, or remove the review fail :)

graylog2-server/src/main/resources/org/graylog2/freemarker/templates/HTML/event_limit_reached.ftl
```diff
@@ -0,0 +1,9 @@
+<#if _title>
+    Event limit reached
+</#if>
+
+<#if _description>
+    <span>
+        Event limit »${event_limit}« reached for event definition »${event_definition_title}«. Try to use a more specific search query or use aggregations. Otherwise try to raise the limit.
```
mpfz0r, line 7: can we also show the eventdefinition id here? e.g. `title (6534de23f213ac)`

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -236,12 +244,32 @@ private void filterSearch(EventFactory eventFactory, AggregationEventProcessorPa
                         .build());
 
                 eventsWithContext.add(EventWithContext.create(event, msg));
+                if (config.eventLimit() != 0) {
+                    if (messageCount.incrementAndGet() >= config.eventLimit()) {
+                        eventLimitReached = true;
```
mpfz0r, line 249: why not just `throw new EventLimitReachedException();;` directly without the boolean?

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -194,37 +206,60 @@ public ValidationResult validate() {
...
+        return validationResult;
+    }
+
+    @Override
+    public ValidationResult validate(@Nullable EventProcessorConfig oldEventProcessorConfig,
+                                     EventDefinitionConfiguration eventDefinitionConfiguration) {
+        final ValidationResult validationResult = new ValidationResult();
+        AggregationEventProcessorConfig oldConfig = (AggregationEventProcessorConfig) oldEventProcessorConfig;
+
+        if (oldConfig == null) {
```
mpfz0r, line 218: I think it might be worth to add a comment here, on why are we doing this.
e.g. `// Only enforce event limit on newly created event definitions`

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -194,37 +206,60 @@ public ValidationResult validate() {
...
+            if (eventLimit() <= 0) {
+                validationResult.addError(FIELD_EVENT_LIMIT, "Event limit must be greater than 0.");
+            }
+        } else {
+            if (oldConfig.eventLimit() != 0 && eventLimit() <= 0) {
```
mpfz0r, line 223: If someone changes his `oldConfig` from aggregation to filter, we can also enforce the new event limit.
I guess we can append `oldConfig.series().isEmpty() &&` to the query to check this

mpfz0r (changes requested):
a few comments, but it's almost over the finish line :)

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
moesterheld: @mpfz0r only added these as comments but will add my approval as they have been addressed.

moesterheld (approved):
Approving, as my comments have been addressed.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: changed it

graylog2-server/src/main/resources/org/graylog2/freemarker/templates/HTML/event_limit_reached.ftl
AntonEbel: fixed

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
AntonEbel: added some comments

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessorConfig.java
```diff
@@ -194,37 +206,60 @@ public ValidationResult validate() {
...
         }
+        return validationResult;
+    }
+
+    @Override
+    public ValidationResult validate(@Nullable EventProcessorConfig oldEventProcessorConfig,
+                                     EventDefinitionConfiguration eventDefinitionConfiguration) {
+        final ValidationResult validationResult = new ValidationResult();
+        AggregationEventProcessorConfig oldConfig = (AggregationEventProcessorConfig) oldEventProcessorConfig;
+
```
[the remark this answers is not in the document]
AntonEbel: applied the commit locally

mpfz0r (approved):
Works and LGTM :+1:

PR #16043, opened 2023-07-25
Fixed & improved tarball installation method

williamtrelawny:
## Description

I made several changes to the `graylogctl` script and related files for the tarball installation method that either fixed issues or made improvements to the user experience when utilizing this installation method:

### [Fixed]
- Missing data directory for Graylog journal to write to.
- Wrong permissions on folders not permitting the `graylog` user to write to.

### [Improved]
- New `.env` file for overriding default values (since the [command line parameters](https://go2docs.graylog.org/5-0/downloading_and_installing_graylog/manual_installation_on_linux.html#CommandLineCLIParameters) specified in our documentation were not actually being applied).
  - Instead of fixing this in `graylogctl`, I think using environment variables is a cleaner and more modern alternative anyway :).
- Added a sample systemd service file (the file provided with the deb/rpm installations does not work with the tarball installation
- Changed folder structure for certain directories for better parity with deb/rpm installations (important for IaC)
  - This includes changing the absolute path for the `node_id` file to a relative path to allow users to keep all Graylog files under the one installation root directory.

Once/if these changes are merged, I will have the docs team update the relevant documentation [here](https://go2docs.graylog.org/5-0/downloading_and_installing_graylog/manual_installation_on_linux.html) as well as add a section to the [default file locations](https://go2docs.graylog.org/5-0/setting_up_graylog/default_file_locations.html) page for manual installs.

## Motivation and Context

During a customer engagement using this installation method, I discovered several issues with this tarball package as well as several UX improvement opportunities that I felt should be addressed.

## How Has This Been Tested?

Though I don't have the ability to test the repo code, I vetted all my changes in a lab environment while committing.

## Types of changes

## Checklist:

### DIFF
--- a/bin/graylogctl
+++ b/bin/graylogctl
@@ -48,9 +48,9 @@ done
 GRAYLOGCTL_DIR=${GRAYLOGCTL_DIR:=$(dirname "$GRAYLOGCTL")}
 GRAYLOG_JVM_DIR="$(dirname "$GRAYLOGCTL_DIR")/jvm"
 GRAYLOG_SERVER_JAR=${GRAYLOG_SERVER_JAR:=graylog.jar}
-GRAYLOG_CONF=${GRAYLOG_CONF:=/etc/graylog/server/server.conf}
+GRAYLOG_CONF=${GRAYLOG_CONF:=etc/graylog/server/server.conf}
 GRAYLOG_PID=${GRAYLOG_PID:=/tmp/graylog.pid}
-LOG_FILE=${LOG_FILE:=log/graylog-server.log}
+GRAYLOG_LOG_FILE=${GRAYLOG_LOG_FILE:=var/log/graylog/server.log}
 LOG4J=${LOG4J:=}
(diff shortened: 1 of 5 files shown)


bernd (changes requested):
We don't want to encourage users to use the tarball installation method.

What's the reason for not using operating system packages or Docker containers in your case?

williamtrelawny:
There is no way to install Graylog into a custom path using the operating system package. This MSP customer wants to install Graylog under `/opt` in their client's environment in this specific case. 

I understand not encouraging the tarball installation method as I'm sure it is probably leftover from the days before the operating system package options were created, but as long as it is listed on our website and docs as an option then we are responsible for supporting it.

bernd:
> I understand not encouraging the tarball installation method as I'm sure it is probably leftover from the days before the operating system package options were created, but as long as it is listed on our website and docs as an option then we are responsible for supporting it.

Yep, the tarball installation method is a leftover and I think we should remove it from the website and our documentation.

PR #16052, opened 2023-07-26
Datanode: Calculating Cert Renewal

janheise:
## Description

Calculating cert renewal

/nocl

## How Has This Been Tested?

unit test

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/security/certutil/CertRenewalService.java
```diff
@@ -0,0 +1,21 @@
...
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+package org.graylog.security.certutil;
+
+public interface CertRenewalService {
+    void checkCertificatesForRenewal();
```
todvora, line 20: I would probably think about separation of the check and handling the notifications. So the CertRenewalService would only return list of certs to renewal and the periodical would trigger notifications?

graylog2-server/src/test/java/org/graylog/security/certutil/CertRenewalServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+        final JobSchedulerClock clock = new JobSchedulerTestClock(now);
+        final var nextRun = clock.nowUTC().plusMinutes(30);
+        final var service = new CertRenewalServiceImpl(clock);
+        // 2 hours is our smalles interval in the FE, so I'm testing only small intervals and expect larger intervals to work accordingly
+        // with a threshold of 10%, a cert should either be invalid if it's no longer valid in 12min (10% of 120min) or if it get's invalid until the next run of the cert checker
+        final var policy = new RenewalPolicy(RenewalPolicy.Mode.MANUAL, "PT2H");
+
+        // renewal check every 30min, so the following should be still valid
+        final var cert1 = generate(now, 0, 35);
+        assertFalse(service.needsRenewal(nextRun, policy, cert1));
```
todvora, line 80: The test suggests that the service interface should return a testable value. Then we wouldn't need to test a package private method and could use a proper return type of the interface.

graylog2-server/src/main/java/org/graylog/security/certutil/CheckForCertRenewalJob.java
```diff
@@ -0,0 +1,100 @@
...
+    @Inject
+    public CheckForCertRenewalJob(final CertRenewalService certRenewalService) {
+        this.certRenewalService = certRenewalService;
+    }
+
+    @Override
+    public JobTriggerUpdate execute(JobExecutionContext ctx) throws JobExecutionException {
+        LOG.debug("Job execute called {}", ctx);
+
+        certRenewalService.checkCertificatesForRenewal();
```
todvora, line 92: here we could have the notification handling

todvora (changes requested):
I think we would benefit from a slight code restructuring. My suggestions are in the comments. Thanks!

todvora (approved):

todvora:
We decided with @janheise that there are some benefits in refactoring the CertRenewalService. But these may be solved in a future refactoring PR to not block work on the datanode now.

Issue #15990, opened 2023-07-19
Secure the DataNode REST interface after successful certificate generation


janheise:
After certificate creation, we also have to secure the REST api with this new certificate. 

also, see https://github.com/Graylog2/graylog2-server/pull/15925#pullrequestreview-1536742113

todvora:
Implemented in https://github.com/Graylog2/graylog2-server/pull/16058. The jersey service now uses the same http certificate that opensearch uses for the rest api.

PR #16058, opened 2023-07-27
Start datanode jersey only after ssl configuration resolved, listen to change events

todvora:
This PR brings HTTPS to the datanode rest API. For that, we are using the very same certificate we are using to encrypt the REST communication from Opensearch. 

Implements https://github.com/Graylog2/graylog2-server/issues/15990

## Description
Changes in startup logic. The JerseyService now waits for a OpensearchConfigurationChangeEvent, which holds all the information about the TLS setup. 

The event is used to communicate with the async jersey service and propagate the exact configuration object. We can't use the OpensearchConfigurationProvider, because during each get call the certificates will be regenerated and encrypted with a different one-time password.  

## Motivation and Context
Secure not just opensearch itself but also the datanode when we already have all the needed pieces. 

## Types of changes

## Checklist:


data-node/src/main/java/org/graylog/datanode/initializers/JerseyService.java
luk-kaminski, line 1: Isn't it a good moment to take more care about this class?
What about removing all the commented lines? Are they needed for smth?
What about some unit tests?

todvora: I am working on a refactoring in https://github.com/Graylog2/graylog2-server/pull/16087 which will change the class anyway. So I don't want to do too much here, so it stays a small and well focused PR. Is that OK?

luk-kaminski (approved):

janheise (approved):

PR #16059, opened 2023-07-27
DataNode: Remove CA from MongoDB on startover

janheise:
## Description

If we press the restart button in the UI, we want to remove the CA from MongoDB.
We do not remove a CA from the filesystem because that was configured manually by the user - so it should remain.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/certutil/CaServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog/security/certutil/CaServiceImpl.java
@@ -126,7 +126,7 @@ public class CaServiceImpl implements CaService {
 
     @Override
     public void startOver() {
-
+        keystoreStorage.removeKeyStore(mongoDbCaLocation);
     }
 
     @Override
--- a/graylog2-server/src/main/java/org/graylog/security/certutil/CaServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog/security/certutil/CaServiceImpl.java
@@ -29,6 +29,7 @@ import org.graylog.security.certutil.keystore.storage.location.KeystoreMongoLoca
 import org.graylog2.Configuration;
 import org.graylog2.bootstrap.preflight.web.resources.model.CA;
 import org.graylog2.bootstrap.preflight.web.resources.model.CAType;
+import org.graylog2.cluster.certificates.CertificatesService;
 import org.graylog2.plugin.system.NodeId;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


graylog2-server/src/main/java/org/graylog/security/certutil/keystore/storage/KeystoreStorage.java
```diff
@@ -32,4 +32,5 @@ void writeKeyStore(final T location,
 
     Optional<KeyStore> readKeyStore(final T location, char[] password) throws KeyStoreStorageException;
 
+    void removeKeyStore(final T location);
```
luk-kaminski, line 35: I'd remove this method from interface. It suggest false assumption - that each storage supports removal.

graylog2-server/src/main/java/org/graylog/security/certutil/CaServiceImpl.java
```diff
@@ -126,7 +126,7 @@ public void upload(String pass, List<FormDataBodyPart> parts) throws CACreationE
 
     @Override
     public void startOver() {
-
+        keystoreStorage.removeKeyStore(mongoDbCaLocation);
```
luk-kaminski, line 129: I'd replace it with:
```
            certificatesService.removeCert(mongoDbCaLocation);

```
And throw away all the remove methods in all the classes.

luk-kaminski: Alternatively, I'd keep the `remove` method in `SmartKeystoreStorage` class only, not making it part of `KeystoreStorage` interface, however.

graylog2-server/src/main/java/org/graylog2/cluster/certificates/CertificatesService.java
```diff
@@ -127,4 +127,14 @@ public Optional<String> readCert(final KeystoreMongoLocation keystoreMongoLocati
         return Optional.empty();
     }
 
+    public void removeCert(final KeystoreMongoLocation keystoreMongoLocation) {
```
luk-kaminski, line 130: This method could potentially return `boolean` to indicate if smth was indeed removed (or `int` with a number of removed documents, as we tend to do in other services).

luk-kaminski (approved):
LGTM, thx!
I've just left a small suggestion about test improvement, I'll leave it to you if you want to do it or not.

Issue #16039, opened 2023-07-25
Search Undo/Redo - inconsistent behavior,


maxiadlovskii:
Working with @dennisoelkers to provide feedback. Its been fun to touch features before release!

[Undo/redo inconsistent](http://videos.graylog.com/watch/86UHfoNngL4RP2Q43eyjcM?)

## Expected Behavior

Undo/redo is consistent every time

## Current Behavior

Widgets may not be removed

## Steps to Reproduce (for bugs)

Inconsistent behavior:

1. start with clean search page
2. Create -> Messasge Count
3. Click Undo
4. notice message count did not disappear, and we can still click undo 1 more time
5. Click undo again
6. message count is removed, no more undo states available, expected state
7. Click redo
8. message count reappears (expected)
9. click undo
10. message count still present (unexpected)

## Your Environment

* Graylog Version: 5.1.3
* Java Version: Bundled
* OpenSearch Version: 2.8
* MongoDB Version: 6.0.7
* Operating System: Ubuntu Server 22.04 LTS
* Browser version: Google Chrome Version 114.0.5735.198 (Official Build) (arm64)

maxiadlovskii:
Note: This bug reproduces only when we create a widget using a sidebar

PR #16061, opened 2023-07-27
Change GenerateNextPosition implementation

maxiadlovskii:
## Description

This PR changes GenerateNextPosition implementation. Now for widgets that don't have defined positions, we create positions starting from row === 1, for the next  + the sum of the height of all prev new widgets. For existing widgets, we change the row by adding the sum of all new widgets' height 

## Motivation and Context
The reason why [#16039](https://github.com/Graylog2/graylog2-server/issues/16039) happens is that during adding a new widget we create it with default positions and for the rest of the widget we just increase the row by 1. This causes situations where the row of the widget may be smaller than the sum of the row and height of the previous widget. In turn, this triggers the `react-grid-layout` functionality that recalculates all the heights and rows for each element. As a result, we run `updateViewState` twice, on creating a new widget and on recalculation. That is why we have 2 revisions in or undo redu buffer after one action and why we need to press undo twice to remove the widget.

fix: [#16039](https://github.com/Graylog2/graylog2-server/issues/16039)

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/views/logic/views/GenerateNextPosition.ts
+++ b/graylog2-web-interface/src/views/logic/views/GenerateNextPosition.ts
@@ -21,14 +21,48 @@ import GetPositionForNewWidget from 'views/logic/views/GetPositionForNewWidget';
 
 import type Widget from '../widgets/Widget';
 
-const incrementRow = (position: WidgetPosition) => position.row + 1;
+export const ConcatPositions = (
+  newPositions: Immutable.Map<string, WidgetPosition>,
+  curPositions: Immutable.Map<string, WidgetPosition>,
+) => {
+  let rowIncrement = 0;
+
+  const newUpdatedPositions: Immutable.Map<string, WidgetPosition> = newPositions.map((initialPosition) => {
+    const defaultHeight = initialPosition.height;
+    const row: number = rowIncrement + initialPosition.row;
+    const widgetPosition = initialPosition.toBuilder().row(row).build();
+    rowIncrement += defaultHeight;
+
+    return widgetPosition;
+  });
+
+  const curUpdatedPositions: Immutable.Map<string, WidgetPosition> = curPositions.map((initialPosition) => {
+    const row: number = rowIncrement + initialPosition.row;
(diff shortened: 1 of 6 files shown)


graylog2-web-interface/src/views/logic/views/GenerateNextPosition.ts
```diff
@@ -21,14 +21,48 @@ import GetPositionForNewWidget from 'views/logic/views/GetPositionForNewWidget';
...
+  let rowIncrement = 0;
+
+  const newUpdatedPositions = newPositions.map((initialPosition) => {
+    const defaultHeight = initialPosition.height;
+    const row: number = rowIncrement + initialPosition.row;
+    const widgetPosition = initialPosition.toBuilder().row(row).build();
+    rowIncrement += defaultHeight;
+
+    return widgetPosition;
+  }) as Immutable.Map<string, WidgetPosition>;
```
linuspahl, line 37: You can also use `}).toMap()` here instead of `Immutable.Map<string, WidgetPosition>`

graylog2-web-interface/src/views/logic/views/GenerateNextPosition.ts
```diff
@@ -21,14 +21,48 @@ import GetPositionForNewWidget from 'views/logic/views/GetPositionForNewWidget';
...
-const incrementRow = (position: WidgetPosition) => position.row + 1;
+export const ConcatPositions = (
+  newPositions: Immutable.Map<string, WidgetPosition>,
+  curPositions: Immutable.Map<string, WidgetPosition>,
+) => {
+  let rowIncrement = 0;
+
+  const newUpdatedPositions = newPositions.map((initialPosition) => {
+    const defaultHeight = initialPosition.height;
+    const row: number = rowIncrement + initialPosition.row;
```
linuspahl, line 32: it looks like `: number` is not necessary here.

linuspahl (approved):
I tested different uses cases and it fixes the described problem. It looks like we are no creating a new row when adding a widget, but it is not a bad behaviour and we plan to attach new widgets at the bottom in the future anyway.

PR #16071, opened 2023-07-31
Add last matched timestamp to EventDefinition

### LINKED ISSUE #15751: Track and display "last triggered" time for event definitions
## What?

Each event definition should have a "last triggered" date value that tracks the last time its criteria matched.  This value should be searchable on the `Event Definitions` page (similar to `created`, `from`, `to` for archives).

## Why?

When there are many event definitions, it can be difficult to determine which are no longer providing value (e.g. existed for years but has never triggered an alert).  Being able to identify these event definitions can help reduce unnecessary load in the environment (removal, frequency adjustment, etc.).

## Your Environment

* Graylog Version: `5.1.2`
* Elasticsearch Version: `7.10.2`
* MongoDB Version: `5.0.14`

patrickmann:
Relates to #15751

Add a new field `matched_at` to `EventDefinition` (similar to `updated_at`) in order to track when the event last fired. 
This field is updated in the `EventProcessor:createEvents` method.


moesterheld:
Hi @patrickmann, 
I think the UI part is still missing for this to really resolve #15751 . Any plans on putting that in this PR?

patrickmann:
> I think the UI part is still missing for this to really resolve #15751 . Any plans on putting that in this PR?

Quite right - thanks. Changed it so we don't inadvertently close the issue. 
I think we can go ahead and get this reviewed and have a separate PR for FE.

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -235,10 +239,14 @@ private void filterSearch(EventFactory eventFactory, AggregationEventProcessorPa
                         .streams(event.getSourceStreams())
                         .build());
 
+                LOG.debug("Creating event {}/{} {}", eventDefinition.title(), eventDefinition.id(), msg.getTimestamp());
                 eventsWithContext.add(EventWithContext.create(event, msg));
             }
 
-            eventsConsumer.accept(eventsWithContext.build());
+            List<EventWithContext> eventsList = eventsWithContext.build();
+            eventDefinitionHandler.updateLastMatched(eventsList);
```
moesterheld, line 247: Couldn't this maybe be moved into `emitEvents` of `EventProcessorEngine`?

This would remove the duplicate code in the different processors.

moesterheld (changes requested):
Works as expected. 

But if possible, I would move the `updateLastMatched` (see comment)

graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
patrickmann: Indeed. I expected event processors would require some specific code based on event type, but that wasn't the case. Much simpler this way.

PR #16072, opened 2023-07-31
fix event definition enable state for filter & aggregation

### LINKED ISSUE #16070: The "Enable" checkbox in the event definition has only enabled status in the UI
When creating a new event definition, the "Enable" checkbox is enabled only in the UI, but the internal state is disabled. 
Only after clicking the checkbox the state is in sync.

Filter & Aggregation:

image
Summery:
image

Event Correlation:
image
Summary:
image

## Your Environment

* Graylog Version: 5.1.1

AntonEbel:
fixes #16070


ousmaneo (approved):
Tested and works

PR #14592, opened 2023-02-02
Disable more weak TLS ciphers by default

mpfz0r:
https://ciphersuite.info/cs/TLS_RSA_WITH_AES_128_GCM_SHA256/
https://ciphersuite.info/cs/TLS_RSA_WITH_AES_256_GCM_SHA384/

This causes no change in the compatibility test when using `testssl.sh`

```
 Running client simulations (HTTP) via sockets 

 Android 6.0                  TLSv1.2 ECDHE-RSA-AES128-GCM-SHA256, 256 bit ECDH (P-256)
 Android 7.0 (native)         TLSv1.2 ECDHE-RSA-AES256-GCM-SHA384, 256 bit ECDH (P-256)
 Android 8.1 (native)         TLSv1.2 ECDHE-RSA-AES256-GCM-SHA384, 253 bit ECDH (X25519)
 Android 9.0 (native)         TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Android 10.0 (native)        TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Android 11 (native)          TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Android 12 (native)          TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Chrome 79 (Win 10)           TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Chrome 101 (Win 10)          TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Firefox 66 (Win 8.1/10)      TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 Firefox 100 (Win 10)         TLSv1.3 TLS_AES_256_GCM_SHA384, 253 bit ECDH (X25519)
 IE 6 XP                      No connection
 IE 8 Win 7                   No connection
 IE 8 XP                      No connection
 IE 11 Win 7                  TLSv1.2 DHE-RSA-AES256-GCM-SHA384, 2048 bit DH  (ffdhe2048)
 IE 11 Win 8.1                TLSv1.2 DHE-RSA-AES256-GCM-SHA384, 2048 bit DH  (ffdhe2048)
 IE 11 Win Phone 8.1          No connection
[...]

PR #16077, opened 2023-07-31
Re-enable weak TLS ciphers for backwards compatibility

### LINKED ISSUE #16076: Restore TLS compatibility with Windows Server 2019
According to 
https://learn.microsoft.com/en-us/windows/win32/secauthn/tls-cipher-suites-in-windows-10-v1809

Windows 10 (Server 2019 is identical in this regard) does not support any cipher we ship Graylog with.

We were removing the last two supported ciphers with
https://github.com/Graylog2/graylog2-server/pull/14592

So for sake of backwards compatibility, we are reverting that change.

mpfz0r:
Revert
 https://github.com/Graylog2/graylog2-server/pull/14592

and allow
https://ciphersuite.info/cs/TLS_RSA_WITH_AES_128_GCM_SHA256/
https://ciphersuite.info/cs/TLS_RSA_WITH_AES_256_GCM_SHA384/

by default again.
This should fix comptibility issues with Windows Server 2019 and older OS.

Fixes #16076


bernd:
@mpfz0r Is there any issue reference? Where is this coming from? Do we have a customer running into issues?

mpfz0r:
> @mpfz0r Is there any issue reference? Where is this coming from? Do we have a customer running into issues?

2 customers and one community member

patrickmann (approved):
Verified we only revert the previous change; and re-enabled ciphers are included in TLS negotiation.

PR #16085, opened 2023-08-01
Referenced search filters are inlined when fetched from Content Packs

### LINKED ISSUE #15841: Unable to edit search filters in dashboard widgets that are imported from content packs
Users are unable to edit search filters used in dashboard widgets that are imported as part of a content pack.

## Expected Behavior

Users should be able to edit the search filters

## Current Behavior

Editing the filter generates an error
filters_error

1. Create a dashboard with widgets that rely on search filters
2. Export this dashboard in a content pack and remove the test dashboard
3. Import the content pack and install 
4. Attempt to edit the filters in the test dashboard widgets

## Context

This would limit the ability of users to customize content after it was imported from the content pack. For Illuminate this would mean the filters could only be edited on the system they were originally generated on, as long as those widgets were never removed.

## Your Environment

* Graylog Version: 5.0.3/5.1.2
* Java Version: openjdk version "17.0.7" 2023-04-18
* Elasticsearch Version: Opensearch 1.3.6
* MongoDB Version:
* Operating System: Ubuntu 20.04.6 LTS
* Browser version: Firefox 114.0.1 (64-bit)

luk-kaminski:
## Description
Fixes #15841.
When creating a new `Content Pack` with an entity that references a `Search Filter` and later on installing this `Content Pack` on a new machine, we will find out that referenced `Search Filter` is missing on a new machine.
It may lead to problems later on, i.e. during `Search Filter` edit attempts, as described in #15841.

While long-term solution is possible, making `Search Filters` part of the `Content Pack`, it may require complex implementation and still lead to many problems (`Search Filters` are hidden under license, permissions considerations etc.).

The simpler solution of the problem is to inline `Search Filters`, so that they are no longer referenced and can be safely edited locally.

The decision was made to keep existing representation of `Search Filters` in the `Content Packs`, with referencing.
`Search Filters` are inlined when `Content Packs` are installed.

The main advantages of that solution:
- The problem can be solved with all the `Content Packs` created so far. They do not have to be re-created, it should be enough to re-install them.
- Proper, original form of the data is kept in `Content Packs`.
- The solution can be changed to more complex one without bigger problems, as we still keep information about original, referenced `Search Filter`.

Alternative would be to inline `Search Filters` when creating `Content Packs`, missing all the advantages described above.

**It is very likely it should be backported to 5.0 and 5.1.**

## Motivation and Context
Problem is described in detail in #15841.

## How Has This Been Tested?
Manually and with new unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/issue-15841.toml
+++ b/changelog/unreleased/issue-15841.toml
@@ -1,0 +1,5 @@
+type = "fixed" # One of: a(dded), c(hanged), d(eprecated), r(emoved), f(ixed), s(ecurity)
+message = "Fixes problems with editing Search Filters of entities installed via Content Packs. Search Filters of those types of entities are not inlined during Content Pack installation and can be edited."
+
+issues = ["15841"]
+pulls = ["16085","Graylog2/graylog-plugin-enterprise/5560"]
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/NativeEntityConverter.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/NativeEntityConverter.java
@@ -17,19 +17,41 @@
 package org.graylog2.contentpacks;
 
 import com.google.common.graph.MutableGraph;
+import org.graylog.plugins.views.search.searchfilters.model.ReferencedSearchFilter;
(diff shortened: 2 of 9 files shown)


todvora (approved):
LGTM!

PR #16088, opened 2023-08-01
Disable users after deactivating auth backend

### LINKED ISSUE #13823: Mark synced users as "disabled" after disabling related authentication service.
## Description
When using an authentication service like LDAP related users will be displayed as "enabled" in the users overview and the user details page. When the authentication service is being disabled, related users are no longer able to login.

In my opinion it makes sense to display the related users as "disabled" after disabling the authentication service.

## Your Environment

* Graylog Version: 5.0

moesterheld:
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#5562

## Description
Makes the `ActiveAuthServiceBackendChangedEvent` aware of previously activated backend to disable users.

## Motivation and Context
fixes #13823 

## How Has This Been Tested?
- Add and activate new LDAP authentication backend
- Login as LDAP user
- Deactivate backend
- User should be disabled
- Add another backend (e.g. Okta)
- Login as user from this backend
- Switch backends by re-enabling LDAP
- Okta user should be disabled
- Login as LDAP user
- LDAP user should be reenabled

## Types of changes

## Checklist:


mpfz0r:
@moesterheld @bernd     Doing this takes away the possibility to switch to a new authentication backend without interrupting currently logged in users. Are we sure this is what we want?

mpfz0r:
Maybe we could do this upon backend deletion instead?

moesterheld:
@bernd @mpfz0r

On deletion won't work since you can only delete a backend with no associated users.

We could only disable non-active users, but that wouldn't make too much sense. We'd have to schedule disabling the active ones at a later point in time. Should we follow this path?

mpfz0r:
> On deletion won't work since you can only delete a backend with no associated users.

Ah right. But couldn't we just delete all the associated users automatically then?
How would that be different?


> We could only disable non-active users, but that wouldn't make too much sense. We'd have to schedule disabling the active ones at a later point in time. Should we follow this path?

Or we could give the user the choice:  `deactivate active users [x]`

moesterheld:
We could also check if there are active users and either prevent deactivation or ask the admin to confirm disabling them. 

The logged in users of the deactivated backend will always run into trouble when their session times out (even now).

mpfz0r:
> We could also check if there are active users and either prevent deactivation or ask the admin to confirm disabling them.

Preventing deactivation would defeat the purpose of a seamless migration.

> The logged in users of the deactivated backend will always run into trouble when their session times out (even now).

If their session times out, they will have to log in again. And then they will login with the new backend.

If we actively kick them out, we interrupt their workflow and they might even lose unsaved changes (e.g while creating a new dashboard)

moesterheld:
I agree. 

However, wouldn't it be confusing to display users with `account_status` enabled as disabled in the UI? How about adding a new flag to the user entity which holds the status of the authentication backend? That way we would move that logic into the backend and the frontend wouldn't have to keep track of the authentication backend states for displaying the users?

bernd:
> I agree.
> 
> However, wouldn't it be confusing to display users with `account_status` enabled as disabled in the UI? How about adding a new flag to the user entity which holds the status of the authentication backend? That way we would move that logic into the backend and the frontend wouldn't have to keep track of the authentication backend states for displaying the users?

@moesterheld I don't think we need to persist the fact that an authentication backend is disabled in the user object. We can determine that based on the auth backend status. We can return that status for each user from the backend, so the frontend doesn't have to compute it. But I would like to avoid touching potentially hundreds of user objects when enabling/disabling auth backends.

We could still use the "Enabled" column and show the reason in the popover.

image

moesterheld:
Agreed. Updating the issue accordingly

PR #16101, opened 2023-08-03
DataNode: Resource for manually initiating certificate Renewal

janheise:
## Description

/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/web/resources/CertificateRenewalResource.java
```diff
@@ -0,0 +1,79 @@
...
+        this.transportAddressSanitizer = transportAddressSanitizer;
+        this.certRenewalService = certRenewalService;
+        this.nodeService = nodeService;
+    }
+
+    @GET
+    // reusing permissions to be the same as for editing the renewal policy, which is below cluster configuration
+    @RequiresPermissions(RestPermissions.CLUSTER_CONFIG_ENTRY_READ)
+    public List<PreflightResource.DataNode> listDataNodesThatNeedCertRenewal() {
+        return nodeService.allActive(Node.Type.DATANODE).values().stream().map(n -> new PreflightResource.DataNode(n.getNodeId(),
```
todvora, line 63: how are the nodes filtered to only return those with the cert renewal need?

janheise: I don't think we should right now. Because even when the cert does not need renewal, you should be able to initiate a renewal manually. 
It's the first draft, I think it needs additional cert info in the reply so the FE can visualise the status with the next steps.

todvora (approved):
LGTM!

PR #16122, opened 2023-08-07
Display indicator when dashboard has unsaved changes.

linuspahl:
## Motivation and Context

With this PR we are unifying the behaviour of the "save" icon on the search and dashboard page. Before this change the dashboard save icon did not indicate unsaved changes.


maxiadlovskii (commented):
It indicates only unsaved dashboards but unsaved changes don't. 
Also when we have an unsaved dashboard indication of the `Save` button doesn't make a lot of sense because the `Save` button is disabled. Maybe make sense to indicate the `save as` button in this case

maxiadlovskii (approved):

PR #16136, opened 2023-08-08
add new endpoints and deprecation notice to upgrade notes

moesterheld:
Add missing new endpoints and deprecation notice for event definition and event notification api to upgrade notes.

/nocl


ousmaneo (approved):
LGTM !

PR #16147, opened 2023-08-09
Fix wrong lookup cache metrics after reconfiguration

### LINKED ISSUE #8633: Lookup table cache size increase not working
When editing (increasing) the lookup table cache size it does not seem to have any effect.

## Expected Behavior

When increasing the cache size of an existing lookup table cache I expect the system to increase the number of entries in the cache (over time) up to the increased cache size limit.

## Current Behavior

When increasing the cache size of an existing lookup table cache (e.g. from 1000 entries to 5000 entries) it does not seem to have an effect.
The `Configured lookup Caches` page shows a stable level of 1000 entries (the old cache size) which does not increase even when manually querying the lookup table with new values that could not have been cached before.

1. Create a MongoDB data adapter
2. Create a corresponding in-memory cache (`Maximum entries: 1000`)
3. Create a lookup table using the new adapter and cache
4. Configure a pipeline to look up values (using `lookup_value()`) using the new lookup table
5. Wait a moment while the `Configured lookup Caches` page shows the `Entries` value for the cache increase until it reaches the configured limit of 1000 entries
6. Edit the existing cache, set `Maximum Entries` to `5000`, make sure to save the changes by clicking `Update Cache`
7. Wait for the pipeline to do more lookups (**Unexpected behavior**: The number of cached entries stays at `1000`)
8. Manually do lookups on the lookup table with values that were never seen before (to ensure that not by pure chance only values that were already cached are looked up). **Expected behavior**: The `Entries` counter should increase as new values were looked up (and the other lookups did not time-out yet)

## Context

Our customer uses a lookup table to check whether some account names are on the list of special accounts (e.g. global admins) that need to be treated differently in some pipelines and dashboard.

## Your Environment

* Graylog Version: Graylog 3.3.2+ade4779 (GL Enterprise customer)
* Elasticsearch Version: 6.8.6
* MongoDB Version: v4.2.8
* Operating System: Centos 7
* Browser version: Google Chrome 83.0.4103.116

## Screenshots
Cache entries always stay at the originally configured limit:
image

Configuration of the cache (showing the increased limit of 5000 entries):
image

mpfz0r:
If a lookup cache gets reconfigured it will not
update the metrics Gauge for `entries`,
but keep using the one of the old instance.

Add a new MetricUtils#replace() method which does exactly that.

Fixes #8633


moesterheld (approved):
works as expected. lgtm

PR #16154, opened 2023-08-10
Fix KafkaContainer class to work with newer bitnami/kafka images

bernd:
The newer images don't use the KAFKA_CFG_ADVERTISED_LISTENERS_FILE
environment anymore, so we have to export the KAFKA_CFG_ADVERTISED_LISTENERS
manually in the command script.

/nocl


todvora (approved):
works locally for me! Thx!

PR #16155, opened 2023-08-10
Add normalize function on changing the  ViewState

maxiadlovskii:
## Description
There are 2 main reasons why this bug happens. 

- we use `loadView` instead of `updateView` when we trying to move the widget. Because of that, we do not put anything into undo redo revisions state
- After moving the widget to another page we are setting not compacted positions (positions with gaps or incorrect row value which has overlaps), because of that, the logic inside React grid make recalculation and update state one more time. 

this PR use updateView instead of loadView and also adds the logic which runs compact function from react grid lib before changing the store.
Also this PR fix the same  problem when user delete some widget

## Motivation and Context
fix: #16149 

## Types of changes

## Checklist:

/nocl


graylog2-web-interface/src/views/logic/views/normalizeViewState.ts
```diff
@@ -0,0 +1,34 @@
...
+  const widgetById = keyBy(widgets.toArray(), 'id');
+  const normalizedWidgetPositions = normalizeWidgetPositions(widgetPositions, widgetById);
+
+  return viewState
+    .toBuilder()
+    .widgetPositions(normalizedWidgetPositions)
+    .build();
+};
+
+export default normalizeViewState;
```
linuspahl, line 34: `normalizeViewState` sounds like we should call this function whenever we update the view state. Maybe name it more widget positions specific, like `normalizeWidgetPositionsInViewState`?

graylog2-web-interface/src/views/logic/widgets/normalizeWidgetPositions.ts
```diff
@@ -0,0 +1,67 @@
...
+    i: id,
+    x: col ? Math.max(col - 1, 0) : 0,
+    y: (row === undefined || row <= 0 ? Infinity : row - 1),
+    h: height || 1,
+    w: width || 1,
+  });
+};
+
+export const positionToLayoutAdapter = (widgetPositions: { [key: string]: WidgetPosition }) => Object.entries(widgetPositions).map(positionToLayoutMapper);
+
```
linuspahl, line 51: I was struggling initially understanding the difference between an adapter and a mapper, especially because they are being used for a single item in one case and for multiple in a different case. Here are some ideas regarding the naming:
`layoutToPositionMapper` -> `layoutItemToPosition`
`layoutToPositionAdapter` -> `layoutToPositions`

`positionToLayoutAdapter` -> `positionToLayoutItem`
`positionToLayoutMapper` -> `positionsToLayout`

linuspahl (approved):
Works well!

graylog2-web-interface/src/views/logic/views/normalizeViewState.ts
maxiadlovskii: I named it like that because `normalizeViewState` returns normalized `ViewState`. For positions, there is another function `normalizeWidgetPositions. So far we normalizing only the positions but in the future `normalizeViewState` might collect all other normalization functions.

Issue #14793, opened 2023-02-27
Field autocomplete makes using the enter/return key to search difficult and cumbersome


drewmiranda-gl:
## What?

When performing search in graylog using the search box, it is very convenient to use the enter/return key to execute the search. However, the same key also inserts suggested autocomplete values for field names and creates a frustrating user experience when you are trying to search for something but your text is replaced with the autocomplete value.

IMO enter/return should not be used to select an autocomplete value (i understand there seems to no true standardization on how autocomplete works with a keyboard, some implementations do use enter/return to select and use the autocomplete value, others use tab) because of the above described issue making it very difficult to quickly perform searches.

https://videos.graylog.com/watch/AefLudH6dvWHvCFjBbocbA?

## Why?

As noted above, trying text in the search box and then pressing enter/return to search causes unexpected behavior because of how field autocomplete works and because the autocomplete value is selected/inserted using the same key as you use to execute search.

## Your Environment

* Graylog Version: 5.0.3
* Operating System: Ubuntu 20.04 LTS
* Browser version: Chrome 110.0.5481.100

Let me know if there are any questions or you wish to discuss further.

dx034:
Are there any plans to fix this in the foreseeable future? Since this behavior started, searching for custom values has become much more cumbersome.

linuspahl:
Thank you for the feedback. I tried to describe the context for this behaviour and implement a fix for it in https://github.com/Graylog2/graylog2-server/pull/16162

The library we use for the query input (the ace editor) has no native support for adjusting the autocomplete keybindings. I had to implement a workaround, it was the only solution I found.

Issue #6909, opened 2019-12-02
Query input autocompletion improvements


isaacegglestone:
I've been informed that the original search bar issue that my team finds a lot more useful than the views search is going to be discontinued.  I would like to ask that it be kept as an optional configuration for views at minimum or to improve the views search to match some of the concerns.

**See my question on why the disparity here for source.**
[](https://community.graylog.org/t/search-bar-not-the-same-in-views-as-in-searches/12977)

Original auto complete functionality in the search feature has the following key benefits:

* Hitting tab includes the colon after the field name.
* Only need to type tab once to get full field name plus colon, vs two tabs to get only field name.
* Original Search includes helpers like exists, NOT exists etc in the autocomplete menu to make things faster.
* The search version of the message area allows you to click the magnifying glass and add fields to your search. We can't see this capability in the views version.

We don't use views at all for the above reasons.

Other complaints about the views search.

There doesn't seem to be any way to collapse the left hand portion of the screen to the left of the messages. This has field names in it and once selected we wish it to be collapsed so that we can see the message in full view.

This issue is also apparent in search view however its not quite as big as the views version. Also the  whole idea that we cannot collapse would be great to have fixed in  a new feature of views.

## Your Environment

* Graylog Version:  Graylog 3.1.3
* Elasticsearch Version: 6.5
* MongoDB Version: 3.6
* Operating System:   ( Linux 4.14.128-112.105.amzn2.x86_64)
* Browser version: Chrome (latest)

Deployed as pods in aws EKS using 
https://github.com/helm/charts/tree/master/stable/graylog
With docker image graylog/graylog:3.1.3

dennisoelkers:
Hey @isaacegglestone, thanks for the input!

Keeping the original search bar is not an option at the moment, as the views are an almost complete rewrite of the current search functionality. We will constantly work on improving its features, especially auto-completion and you can expect it to be far superior to the current search bar in the next versions.

Good news is that the sidebar will be collapsible in 3.2 already. Your suggestions concerning the auto-completion probably make it into one of the next versions after that, as they are really small enough to fix quickly.

We are always thankful for all kinds of feedback, things you are missing or expect differently or stuff that you are happy about!

dennisoelkers:
Most of the issues described in here have already been addressed:

```
* Hitting tab includes the colon after the field name. 
* The search version of the message area allows you to click the magnifying glass and add fields to your search. We can't see this capability in the views version.
```

This one will be included in one of the next versions:

[...]

PR #16162, opened 2023-08-11
Adjust keybindings for query input autocompletion.

linuspahl:
## Motivation and Context

With #6909 we received the feedback that it would be useful to only press `Tab` once to insert the first autocomplete suggestion (instead of pressing it twice). As result we implemented a change to automatically focus the first suggestion. This led to a problem in a different use case. When you search for a custom value like `ssh login` the autocomplete suggested field names for `login` and pressing `Return` did not execute the search but inserted the first suggestion.

We are now no longer focusing the first suggestion and adjusting the behaviour for `Tab` key. When pressing `Tab` while no suggestion is focused we select and insert the first entry.

This way it is possible to press `Return` when searching for a custom value and it still requires only one press to insert the first suggestion.

Fixes https://github.com/Graylog2/graylog2-server/issues/14793

### DIFF
--- a/changelog/unreleased/issue-14793.toml
+++ b/changelog/unreleased/issue-14793.toml
@@ -1,0 +1,14 @@
+type = "fixed"
+message = "Adjust keybindings for query input autocompletion (disable `Return` key)."
+
+issues = ["14793"]
+pulls = ["16162"]
+
+details.user = """
+With #6909 we received the feedback that it would be useful to only press tab once to insert the first autocomplete suggestion (instead of pressing it twice).
+As result we implemented a change to automatically focus the first suggestion. This led to a problem in a different use case.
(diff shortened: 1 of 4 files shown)


maxiadlovskii (commented):
https://github.com/Graylog2/graylog2-server/assets/10741557/03122590-1f00-4d5b-8b15-bdb14b3aa503

When you start typing smth and then press Enter button the search will be executed, but when you continue typing there will be no suggestions. To show them again you need to unfocus  and focus input

linuspahl:
@maxiadlovskii I improved the behaviour. There is still room for improvements because we show different suggestion depending on the context (showing suggestions while starting to write the field name vs. after closing the autocompletion ). But we should improve this in a different PR, I will add it to the description of https://github.com/Graylog2/graylog2-server/issues/16168

maxiadlovskii (approved):

PR #16182, opened 2023-08-15
Add an optional flag to invert remove_field condition

### LINKED ISSUE #8128: wanted list in pipeline
Hi, 

In pipelines the process is a blacklist. What I don't need as a field I have to remove.  I need a rule that delete all fields that are not on the "wanted" list.

patrickmann:
Resolves #8128 

Introduce an optional flag `invert` to `remove_field` function. When `true`, remove all fields except those those that match the given name / regex.


moesterheld (approved):
lgtm

PR #16196, opened 2023-08-17
add error message when accessing a regex result field using `.[a]`

### LINKED ISSUE #16029: Pipeline editor doesn't provide feedback when using incorrect syntax to access the value from a regex result
Pipeline editor doesn't provide feedback when using incorrect syntax to access the value from a regex result.

The following appear to be valid ways to access a value from an array or an object:

`array["key"]`

`array.key`

However, `array.["key"]` is invalid, but no errors are shown in the web interface. A warning is generated in `server.log`

> [PipelineInterpreter] Error evaluating action for rule <test/null> with message: source: 127.0.0.1 | message: ....

## Expected Behavior

Pipeline editor provide feedback that this is invalid syntax and the pipeline rule has become invalid.

## Current Behavior

Pipeline editor doesn't indicate syntax is invalid and pipeline rule will no longer function.

1. Create a new pipeline rule

Example rule to reproduce:

```
rule "test"
when
    true
then
    let rs = regex(
        pattern: "datarows=(.+),\\smetadata",
        value: to_string($message.message),
        group_names: ["a"]
        );
    set_field("debug", rs.["a"]);
end
```

Example sample message:

`$message.message` = 

```
{schema=[{"column_type":"grouping","type":"string","field":"source","name":"grouping: source"},{"column_type":"metric","type":"numeric","function":"count","field":"source","name":"metric: count(source)"},{"column_type":"metric","type":"numeric","function":"sum","field":"gl2_accounted_message_size","name":"metric: sum(gl2_accounted_message_size)"}], datarows=[["joel-surface3",1155,2676106.0],["se-dc1",360,1724082.0],["graylog_api",2,1707.0]], metadata={effective_timerange={from=2023-07-20T19:59:08.263Z, to=2023-07-21T19:59:08.264Z, type=absolute}}}
```

## Your Environment

* Graylog Version: 5.1.3
* Java Version: Bundled
* OpenSearch Version: 2.8.0
* MongoDB Version: 6.0.8
* Operating System: Ubuntu Server 22.04 LTS
* Browser version: Google Chrome Version 115.0.5790.98 (Official Build) (arm64)

moesterheld:
## Description
In pipeline rules, regex result values can be either accessed using `result.key` or `result["key"]`. 
However, when entering it as `result.["key"]`, no syntax error is shown in the rule editor.

This pr changes that. There is no legit other use for the above mentioned notation.

## Motivation and Context
resolves #16029

## How Has This Been Tested?
- create regex results and use the different notations for setting a field, e.g. (message "test")

```
rule "testrule"
when
  true
then
    let rs = regex(
        pattern: "t(.+)t",
        value: to_string($message.message),
        group_names: ["a"]
        );
    set_field("debug", rs["a"]);
end
```

image

## Types of changes


patrickmann (approved):
Looks good, works as expected

Issue #15200, opened 2023-04-12
Ability to have lookup table cache NOT CACHE null/empty values.


drewmiranda-gl:
## What?

This is a follow up of https://github.com/Graylog2/graylog2-server/issues/13579 .

Asking for the ability to have a graylog lookup table NOT CACHE empty/null values. Currently if a cache is used for a lookup table, the result is always cached, regardless of the result being valid or having contents. This creates a problem where unexplainable and ephemeral empty responses from the data adapter get stuck in cache and the lookup table is incorrectly serving an empty result when it could get a working result if it ignored the empty cached value.

## Why?

Using a cache with a lookup table has a lot of practical performance and stability benefits when using lookup tables. However, there are scenarios, for whatever reason and without any clear explanation, where graylog will cache an empty value for a given lookup, despite a valid entry being available from the data adapter. When this empty value is cached, it will continue to be returned until the cache's "expire after" is reached.

When using "Expire after access", the empty value can potentially live on for a long time. For example, if the lookup key is accessed once per minute and the "Expire after access" value is 5 minutes (which is the default i believe), that empty value will continue being served until no log messages reference that lookup key for 5 minutes.

Currently the only workarounds for this are:

* Not using a cache (impractical and IMO will cause a higher liklihood of returning an error or empty value)
* Manually clearing the cache for the offending key or purging cache

## Your Environment

* Graylog Version: Graylog 5.0.6+51f2df8
* ~~Elasticsearch~~ OpenSearch Version: 2.6.0
* MongoDB Version: 6.0.5
* Operating System: Ubuntu Server 22.04 LTS
* Browser version: Google Chrome Version 112.0.5615.49 (Official Build) (arm64)

Please let me know if there are any questions and if there is any testing i can do or help with.

patrickmann:
@drewmiranda-gl Do you envision that this is something you specify in the UI when creating the cache? Or a global configuration under _system/configuration_?

patrickmann:
https://github.com/Graylog2/graylog2-server/issues/16230 introduced a new setting to block caching of null (off by default). 
User should be able to set this from the UI when defining an in-memory cache.

Issue #13579, opened 2022-09-29
Lookup Table Cache caching NULL values, never clears unless Expire After time elapses


drewmiranda-gl:
The behavior of how Lookup Table Cache works can lead to some unexpected (and unwanted) outcomes. For example, if a lookup value does NOT EXIST in cache currently, and the data adapter returns a null value (either because its inaccessible or doesn't have a value to return), the Lookup Table cache saves that value. All subsequent lookups for that value will return null/empty, even if the data adapter can correctly serve the request. Only after one of the "Expire after" conditions is met (such as Expire after access, defaults to 60s, or Expire after write, defaults to Never).

This creates an unwanted condition where a momentary failure of the data adapter causes graylog to be unable to lookup that value, potentially forever. For example, if the cache is set for `Expire after access: 60s` and logs for that lookup value continually are sending requests to the lookup table (and thus its cache), it will always return null/empty.

I observed this behavior with my Graylog Lab and a reverse DNS lookup table, where unexpectedly a lookup table would abruptly stop returning a lookup value for a given IP, despite the DNS server (and thus the data adapter) having a valid value to return.

## Expected Behavior

This is tricky because the caching of NULL/empty is super helpful for performance reasons. My expectation is that as soon as that lookup value is available again, it is populated.

## Current Behavior

A null value is cached and if logs continue to lookup from that lookup table, the cache will never expire, thus always returning a null value.

## Possible Solution

A workaround on the client/customer side is to enable "Expire after write". However, this makes the cache less useful since its emptying itself at a regular interval.

Not caching NULL values will also lead to negative performance impact, especially when looking up large amount of DNS queries where the same IP may not have a result to be found.

IMO it could be helpful (but not sure if possible) for the cache to accommodate a NULL value when the DNS server doesn't have a result, VS caching NULL in the event of a DNS server timeout. It could be helpful to not clear/overwrite an existing cache item if the current result is NULL but there was a valid previous last value.

A more complex (and likely infeasible) approach could be to have a scheduler periodically check the contents of the cache to see if values have changed and invalidate the cache. This gets a bit more complex though than a vanilla data adapter.

## Steps to Reproduce (for bugs)

1. Setup look up table with DNS Lookup (Reverse lookup, PTR), and `Node-local, in-memory cache`
2. Using the edit page for the Lookup table, lookup a value that doesn't exist on the DNS server
[...]

PR #16199, opened 2023-08-18
Setting to not cache null values

patrickmann:
References #13579 and #15200

Data adapters can return non-null values, even when a lookup failed: They typically return an "empty" result with infinite TTL.

>     private static final LookupResult EMPTY_LOOKUP_RESULT = builder()
>             .cacheTTL(NO_TTL)
>             .build();
> 

The lookup table cache does not know to ignore these values. This can cause bad data to end up in the cache where it may reside for a long time until it is evicted.

Introduce a new cache configuration setting to optionally ignore `EMPTY_LOOKUP_RESULT`. This gives users something that is in between the extremes of caching everything or nothing.

Notes: 

- `DEFAULT_ERROR_LOOKUP_RESULT` has a 5s TTL
- Some data adapters like CSV perform a periodic refresh which also purges the cache. These types of adapters may also benefit from ignoring empty results, depending on how the refresh and expiration intervals are configured; and how dynamic the underlying data is.


kodjo-anipah (changes requested):
This would work, but I think it would be better to set the Empty LookUpResult TTL to something different then Long.MAX_VALUE, maybe something more reasonable like 5min. 
In that way even on frequent cache hits for the null value we would still enforce an eviction and refetching after some time.

kodjo-anipah (approved):
LGTM, thanks for your patience

patrickmann:
Reverted change to TTL for empty - we will handle that separately: #16466
Related UI issue: #16230 
Checks failed on unrelated IT test failure - going ahead with merge.

PR #16205, opened 2023-08-18
Introduce bootstrap configuration for plugins

thll:
Bootstrap configuration can be part of a plugin distribution but it will be loaded before the actual plugin is instantiated.

This allows to e.g. specify a plugin-wide switch for a feature which can then be evaluated early in the plugin startup phase. If the feature is disabled, we can skip loading configuration with potentially required parameters and prevent validation errors.

/jpd Graylog2/graylog-plugin-enterprise#5636
/nocl


mpfz0r (approved):
Works, LGTM and I like it :+1:

PR #16241, opened 2023-08-23
Field type management - skeleton

luk-kaminski:
/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeChangeResource.java
```diff
@@ -0,0 +1,147 @@
...
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import java.util.Optional;
+import java.util.Set;
+
+import static org.graylog2.shared.rest.documentation.generator.Generator.CLOUD_VISIBLE;
+
+@RequiresAuthentication
+@Api(value = "System/FieldTypes", tags = {CLOUD_VISIBLE})
+@Path("/system/indices/field_types_change")
```
dennisoelkers, line 51: I think this should be `/system/indices/mappings` instead, REST endpoints usually deal with entities, not actions.

graylog2-server/src/main/resources/org/graylog2/featureflag/feature-flag.config
```diff
@@ -59,6 +59,9 @@
 search_filter=on
 scripting_api_preview=on
 
+# Enabling field types change
+field_types_change=off
```
dennisoelkers, line 63: Let's call this `field_types_management`, similar to the name of the topic.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeChangeResource.java
```diff
@@ -0,0 +1,147 @@
...
+                                    @ApiParam(name = "rotate_immediately", defaultValue = "false")
+                                    @QueryParam("rotate_immediately") boolean rotateImmediately
+    ) {
+        checkPermissions(indexSetsIds);
+
+        //TODO: validation of fields
+
+        CustomFieldMappings.CustomFieldMapping customMapping = new CustomFieldMappings.CustomFieldMapping(fieldName, newType);
+
+        for (String indexSetId : indexSetsIds) {
```
dennisoelkers, line 92: We should extract the actual code into some kind of a service, so the REST resource is not doing too much. This way we can also support reuse as well as supporting potential refactorings (where the actual flow is triggered from somewhere else - e.g. through an approval mechanism).

graylog2-server/src/main/java/org/graylog2/indexer/indexset/CustomFieldMappings.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public record CustomFieldMapping(@JsonProperty("field_name") String fieldName,
+                                     @JsonProperty("type") String physicalType) {
+
+        public Set<CustomFieldMapping> toSet() {
+            return Set.of(this);
+        }
+    }
+
+    public CustomFieldMappings modifiedWith(final Set<CustomFieldMapping> changedMappings) {
```
dennisoelkers, line 43: From the naming it is not really clear what the method is doing. What about `merge` or `mergeWith`?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeChangeResource.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        return Response.ok().build();
+    }
+
+    private IndexSetConfig storeMapping(final CustomFieldMappings.CustomFieldMapping customMapping,
+                                        final IndexSetConfig indexSetConfig) {
+        final IndexSetConfig.Builder builder = indexSetConfig.toBuilder();
+        final CustomFieldMappings previousCustomFieldMappings = indexSetConfig.customFieldMappings();
+        if (previousCustomFieldMappings == null) {
```
dennisoelkers, line 115: If we create an empty instance of `CustomFieldMappings` in the `IndexSetConfig` model when there is none present, we can avoid this check.

dennisoelkers (changes requested):
Adding some first thoughts about naming/structure.

graylog2-server/src/main/java/org/graylog2/indexer/IndexMapping.java
```diff
@@ -104,8 +111,13 @@ protected Map<String, Map<String, Object>> fieldProperties(String analyzer) {
                 .put(Message.FIELD_GL2_MESSAGE_ID, notAnalyzedString())
                 // to support wildcard searches in source we need to lowercase the content (wildcard search lowercases search term)
                 .put("source", analyzedString("analyzer_keyword", true))
-                .put("streams", notAnalyzedString())
-                .build();
+                .put("streams", notAnalyzedString());
+
+        if (customFieldMappings != null) {
+            customFieldMappings.forEach(customMapping -> builder.put(customMapping.fieldName(), type(customMapping.physicalType())));
```
dennisoelkers, line 117: I think we should use our logical types when configuring the mapping and map them to their physical counterparts in here when creating the mapping template.

dennisoelkers (dismissed):

graylog2-server/src/main/java/org/graylog2/indexer/indexset/CustomFieldMappings.java
```diff
@@ -0,0 +1,49 @@
...
+                                     @JsonProperty("type") String physicalType) {
+
+        public Set<CustomFieldMapping> toSet() {
+            return Set.of(this);
+        }
+    }
+
+    public CustomFieldMappings mergeWith(final Set<CustomFieldMapping> changedMappings) {
+        final Set<CustomFieldMapping> modifiedMappings = new HashSet<>(this);
+        modifiedMappings.removeIf(m -> changedMappings.stream().anyMatch(cm -> cm.fieldName().equals(m.fieldName())));
```
todvora, line 45: Is it ok to simply ignore all previous custom mappings? Should we require some kind of confirmation for that?

graylog2-server/src/main/java/org/graylog2/commands/Server.java
```diff
@@ -177,7 +177,7 @@ protected List<Module> getCommandBindings(FeatureFlags featureFlags) {
                 new RetentionStrategyBindings(elasticsearchConfiguration),
                 new PeriodicalBindings(),
                 new ObjectMapperModule(chainingClassLoader),
-                new RestApiBindings(configuration),
+                new RestApiBindings(configuration, featureFlags.isOn("field_types_management")),
```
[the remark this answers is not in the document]
luk-kaminski: I expect that the whole feature flag will be gone. It is temporary, so that if we merge that stuff no client will have this feature prematurely. So I'd not focus too much on discussing how to pass this temporary flag.

graylog2-server/src/main/java/org/graylog2/indexer/indexset/CustomFieldMappings.java
luk-kaminski: Do you suggest UI change (in the table types coming from custom mappings are visualized differently or user is additionally warned before changing them)?

graylog2-server/src/main/java/org/graylog2/indexer/indexset/CustomFieldMappings.java
```diff
@@ -0,0 +1,69 @@
...
+import org.graylog2.indexer.fieldtypes.FieldTypeMapper;
+import org.graylog2.indexer.fieldtypes.FieldTypes;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class CustomFieldMappings extends HashSet<CustomFieldMapping> {
```
todvora, line 29: Is there a reason why we extend the full HashSet class? Do we really want to provide all set methods that are implemented in the hashset?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeMappingsResource.java
```diff
@@ -0,0 +1,101 @@
...
+    @Inject
+    public FieldTypeMappingsResource(final FieldTypeMappingsService fieldTypeMappingsService) {
+        this.fieldTypeMappingsService = fieldTypeMappingsService;
+    }
+
+    @GET
+    @Path("/types")
+    @Timed
+    @ApiOperation(value = "Get list of all types valid inside the indexer")
+    public Map<String, String> getAllFieldTypes() {
```
todvora, line 66: Should we check the `typemappings:read` permission here?

todvora (changes requested):

graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeMappingsResource.java
dennisoelkers: No, this is not returning any configured data, just a list of field types names which are present (`string`, `long`, `ip`, ...).

graylog2-server/src/main/java/org/graylog2/indexer/indexset/CustomFieldMappings.java
dennisoelkers: We do want to be able to treat this as a `Set<CustomFieldMapping>` elsewhere, for iteration, etc. - so it makes sense to use a `HashSet` for the implementation. We do not need to implement any of the methods. There is IMO no smalles interface that we could implement without extending from something.

todvora: ok, makes sense, thanks!

graylog2-server/src/main/java/org/graylog2/rest/resources/system/field_types/FieldTypeMappingsResource.java
todvora: :+1:

todvora (approved):
LGTM

PR #16245, opened 2023-08-23
Adapt DatanodeClusterIT for full cert setup of all nodes

todvora:
This PR extends our existing security integration tests for the Datanode. The cluster formation test now uses the full certificate setup on all nodes. Fixed passwords were removed and are autogenerated now, to ensure that they are all actually used and we don't rely on any internal password knowledge. The truststores are generated from the CA only, not using the HTTP certificate anymore (not needed).

/nocl

## Motivation and Context
Extended tests so we can rely more on the security refactorings we do in the Datanode

## How Has This Been Tested?
By the tests itself.

## Types of changes

## Checklist:

### DIFF
--- a/data-node/src/test/java/org/graylog/datanode/integration/DatanodeClusterIT.java
+++ b/data-node/src/test/java/org/graylog/datanode/integration/DatanodeClusterIT.java
@@ -27,10 +27,6 @@ import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.http.NoHttpResponseException;
 import org.graylog.datanode.configuration.variants.KeystoreInformation;
 import org.graylog.datanode.testinfra.DatanodeContainerizedBackend;
-import org.graylog.security.certutil.CertutilCa;
-import org.graylog.security.certutil.CertutilCert;
-import org.graylog.security.certutil.CertutilHttp;
-import org.graylog.security.certutil.console.TestableConsole;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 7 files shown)


data-node/src/test/java/org/graylog/datanode/integration/DatanodeClusterIT.java
```diff
@@ -16,59 +16,143 @@
...
-
-        secondaryNode = new DatanodeContainerizedBackend(
-                primaryNode.getNetwork(),
-                primaryNode.getMongodbContainer(),
-                "nodeB",
-                datanodeContainer ->
-                        datanodeContainer
-                                .withEnv("GRAYLOG_DATANODE_CLUSTER_INITIAL_MANAGER_NODES", "nodeA")
-                                .withEnv("GRAYLOG_DATANODE_OPENSEARCH_DISCOVERY_SEED_HOSTS", "nodeA:9300")
+    void setUp() throws CertificateException, IOException, KeyStoreException, NoSuchAlgorithmException {
```
luk-kaminski, line 64: You can potentially simplify the `throws` part using `GeneralSecurityException`

data-node/src/test/java/org/graylog/datanode/integration/DatanodeSecurityTestUtils.java
```diff
@@ -0,0 +1,100 @@
...
+        final CertutilCa command = new CertutilCa(certPath.toAbsolutePath().toString(), input);
+        command.run();
+        return new KeystoreInformation(certPath, password);
+    }
+
+    public static KeystoreInformation generateTransportCert(Path dir, KeystoreInformation ca, String... containerHostnames) {
+        final Path transportPath = dir.resolve("transport-" + RandomStringUtils.randomAlphabetic(10) + ".p12");
+        final String password = RandomStringUtils.randomAlphabetic(10);
+        TestableConsole inputCert = TestableConsole.empty().silent()
+                .register("Enter CA password", ca.passwordAsString())
```
luk-kaminski, line 73: Those magic strings like "Enter CA password" appear in many place in the code. I wonder if they do not deserve a constant or another mechanism that will allow to modify them in one place only, whenever they need modification.

luk-kaminski (approved):
Looks good, thanks!
Left just some minor comments, but they are minor ones.

data-node/src/test/java/org/graylog/datanode/integration/DatanodeSecurityTestUtils.java
todvora: Yeah, these user inputs would probably deserve an enum, expressing the intention behind the user-facing prompt. Let me see if I can suggest something nicer.

todvora: Ok, this will deserve its own PR, it will be a slightly bigger refactoring.

PR #16251, opened 2023-08-24
Allow duplicating event definitions with a limit of 0

mpfz0r:
With #16035 we try to enforce a new default event limit, without breaking backwards compatibility.

The validations are making an exemption if an existing event definition is being edited which doesn't have an event limit yet (limit of 0)

Duplicating event definitions was done in the frontend, followed by creating a new one.
This makes it impossible to differentiate between copy and create in the backend.

 - Create a new `/events/definitions/{definitionId}/duplicate` endpoint. Which creates a copy without performing a validation.

 - Also correctly set the state for unscheduled event defintions. This got missed with #15558

Refs #16035

### DIFF
--- a/changelog/unreleased/pr-16251.toml
+++ b/changelog/unreleased/pr-16251.toml
@@ -1,0 +1,6 @@
+type = "a"
+message = "Add API endpoint for event definition duplication."
+
+issues = [""]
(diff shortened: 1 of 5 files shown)


kodjo-anipah (approved):
LGTM
image

PR #16267, opened 2023-08-25
Remove security plugin from matrix tests OS 1.3 instance

todvora:
- Update opensearch to the latest 1.3.12
- Uninstall the security plugin during startup

/nocl

## Motivation and Context
We don't need the security problem and we assume that it causes some problems when only disabled. Removing it completely could help. 

## Types of changes

## Checklist:


todvora:
Not needed, it seems that the 1.3 instance is running reliably after the update to 1.3.12 and we don't need to remove this plugin.

PR #16268, opened 2023-08-25
Update Select.tsx

Srinidhi-Saravanan:
Fixed the binding issue in the Select dropdown 
## Description

Changed the return value of **_formatInputValue** from the Select component. Previously, in _formatInputValue function, if the availableOption  returns empty or undefined, the method is returning the comma separated string of selected values as a key value pair, which is getting selected in the dropdown, which is not an expected behaviour. 
To fix this issue, I have removed the `{ [displayKey]: String(value), [valueKey]: value }` from the return statement.

Screenshot 2023-08-24 at 5 23 47 PM

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/common/Select/Select.tsx
+++ b/graylog2-web-interface/src/components/common/Select/Select.tsx
@@ -510,7 +510,7 @@ class Select<OptionValue> extends React.Component<Props<OptionValue>, State> {
       .map((v) => {
         const availableOption = options.find((option) => option[valueKey || ''] === v);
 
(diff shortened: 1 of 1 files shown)


zeeklop:
Can't reproduce the error on SalesForce wizard. What are the steps to get to this error?

danotorrey:
@Srinidhi-Saravanan It seems like this issue only happens when the list of choices changes since the previously saved state. I am a bit unsure of how this is happening for the Salesforce input. Had the list of choices changed during development?

PR #16319, opened 2023-08-30
add content stream to csp default headers

kodjo-anipah:
/nocl
Add Graylog content stream endpoint to CSP-Header

```
Path parts in the CSP that end in / match any path they are a prefix of. 
For example, example.com/api/ will match URLs like example.com/api/users/new.
```
See section Hosts values https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy


patrickmann (approved):
LGTM

PR #16320, opened 2023-08-30
Check permission before loading content stream tag

ousmaneo:
This PR fixes the permission issue when loading content stream tag.

fixes Graylog2/graylog-plugin-enterprise#5683

## Types of changes

## Checklist:


bernd (commented):
@ousmaneo We now don't use `enterprise-feed` for reader users even if they use Graylog Enterprise, right? Only admins or users with the `licenseinfos:read` permission can see the `enterprise-feed`.

Is that what we want?

ousmaneo:
> @ousmaneo We now don't use `enterprise-feed` for reader users even if they use Graylog Enterprise, right? Only admins or users with the `licenseinfos:read` permission can see the `enterprise-feed`.
> 
> Is that what we want?

Hi @bernd yes you are right, I talked with @patrick about this. Currently, license checks require permission checks. We are checking with Kay.

ousmaneo:
Hi @bernd, After talking with @kay we are going to extend this fix to take care of the case the user does not have permission to check the license.

moesterheld (approved):
backend lgtm

PR #16337, opened 2023-08-31
Introduce IndexingResults class

mpfz0r:
Bulk index requests were only returning errors so far.

Introduce a new value class: `IndexingResults` which
contains bulk indexing `errors()` and `successes()`

This allows bulk index callers to see into which
actual index a message has been ingested.

/nocl

/jpd https://github.com/Graylog2/graylog-plugin-enterprise/pull/5700


dennisoelkers (approved):
Looks great!

PR #16371, opened 2023-09-05
DataNode: removing demo users

janheise:
## Description

Removing the demo users from `internal_users.yml` so they don't end up in the DataNode.

/nocl

## Types of changes

## Checklist:


dennisoelkers (approved):

todvora (commented):
This needs to be merged only after we switch to the jwt auth. 

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Map.put(Object, Object)" because "adminUserConfig" is null
```

Signalizes that we still need at least the admin user in the file for datanode to work properly. We don't add it to the file, we only modify its properties in the current implementation.

dennisoelkers (approved):

PR #16372, opened 2023-09-05
DataNode: Use the correct architecture for the distribution jar in tests

janheise:
## Description

Fix to use the correct jar corresponding to the system architecture in tests.

/nocl

## Types of changes

## Checklist:


todvora (approved):

data-node/src/test/java/org/graylog/datanode/testinfra/DatanodeContainerizedBackend.java
```diff
@@ -113,7 +114,7 @@ private NodeContainerConfig createConfig() {
     }
 
     private static ImageFromDockerfile createDockerImageFile(String opensearchVersion) {
-        final String opensearchTarArchive = "opensearch-" + opensearchVersion + "-linux-x64.tar.gz";
+        final String opensearchTarArchive = "opensearch-" + opensearchVersion + "-linux-" + OpensearchDistribution.archCode(System.getProperty("os.arch")) + ".tar.gz";
```
todvora, line 117: I think we should also add a check that the tar archive exists. This can save us a lot of time if not and provide a valuable feedback even before the container tries to start. Any failure later is far more complicated to detect.

PR #16373, opened 2023-09-05
Fixing up test for MongoDB's preflight freshness check logic.

dennisoelkers:
## Motivation and Context

This PR is fixing issues with the MongoDB preflight check test. Its naming was wrong (so it was never executed) as well as the JSON fixtures did not match the most recent state of the implemented logic.

/nocl

## Types of changes

## Checklist:


bernd:
Thanks for the fix! :pray: Shouldn't we create pre-flight-check tests for MongoDB 6 and 7 instead of merging the test for 5 with the abstract class?

todvora (approved):
Tested locally, works as expected, thanks!

dennisoelkers:
@bernd: I was also thinking about this, but I think that the logic we are testing here is so generic, that the MongoDB-version does not really matter.

PR #16396, opened 2023-09-07
Implement foundation for keyboard shortcuts

maxiadlovskii:
## Description

This PR implements a foundation for keyboard shortcuts in the web interface.

image

When you test this PR please make sure to enable the feature flag `frontend_hotkeys`.

/nocl

### DIFF
--- a/graylog2-web-interface/src/contexts/HotkeysContext.tsx
+++ b/graylog2-web-interface/src/contexts/HotkeysContext.tsx
@@ -21,7 +21,7 @@ import type Immutable from 'immutable';
 
 import { singleton } from 'logic/singleton';
 
-export type ScopeName = 'view' | 'search' | 'dashboard';
+export type ScopeName = 'general' | 'search' | 'dashboard';
 export type ScopeParam = Array<ScopeName> | ScopeName
 export type ActiveHotkey = {
   options?: Options & { scopes: ScopeName },
--- a/graylog2-web-interface/src/contexts/HotkeysContext.tsx
+++ b/graylog2-web-interface/src/contexts/HotkeysContext.tsx
@@ -15,8 +15,6 @@
  * <http://www.mongodb.com/licensing/server-side-public-license>.
  */
 import * as React from 'react';
-import type { Hotkey } from 'react-hotkeys-hook/dist/types';
-import type { Options } from 'react-hotkeys-hook';
 import type Immutable from 'immutable';
 
 import { singleton } from 'logic/singleton';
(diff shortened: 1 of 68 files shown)


graylog2-web-interface/src/contexts/HotkeysProvider.tsx
```diff
@@ -0,0 +1,120 @@
...
+    disableScope: _disableScope,
+    enableScope: _enableScope,
+    setActiveScopes,
+    hotkeys,
+    toggleScope,
+    enabledScopes: enabledScopes as Array<ScopeName>,
+    hotKeysCollection,
+    activeHotkeys,
+    addActiveHotkey,
+    removeActiveHotkey,
```
linuspahl, line 102: It looks like we can simplify the context interface and remove most methods?

graylog2-web-interface/src/hooks/useHotkeys.tsx
```diff
@@ -0,0 +1,87 @@
...
+  document: undefined,
+  ignoreModifiers: false,
+};
+
+const useHotkeys = <T extends HTMLElement>(
+  actionKey: string,
+  callback: HotkeyCallback,
+  options?: Options & { scopes: ScopeName },
+  dependencies?: Array<unknown>,
+) => {
```
linuspahl, line 45: Because we talked about the parameters of this function. I like the following approach:
```
scope: string,
actionKey: string,
callback: HotkeyCallback,
options?: { dependencies: ScopeName }
```
Se the usage would look like this:
`useHotkeys('dashboard', 'SAVE', _onSaveView);` instead of
`useHotkeys('SAVE', _onSaveView, { scopes: 'dashboard' });`

graylog2-web-interface/src/contexts/HotkeysContext.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
+import * as React from 'react';
+import type { Hotkey } from 'react-hotkeys-hook/dist/types';
+import type { Options } from 'react-hotkeys-hook';
+import type Immutable from 'immutable';
+
+import { singleton } from 'logic/singleton';
+
+export type ScopeName = 'view' | 'search' | 'dashboard';
```
linuspahl, line 24: It would be nice to infer the scope names based on the `hotKeysCollection`.

graylog2-web-interface/src/contexts/HotkeysModal.tsx
```diff
@@ -0,0 +1,137 @@
...
+            {
+          Object.entries(hotKeysCollection).map(([scope, collection]: [ScopeName, HotkeyCollection]) => (
+            <HotkeyCollectionSection scope={scope} collection={collection} />
+          ))
+          }
+          </SectionGrid>
+        </Content>
+      </Modal.Body>
+      <Modal.Footer>
+        <LinkButton bsStyle="primary" to="/" target="_blank">Show all hot keys</LinkButton>
```
linuspahl, line 131: Currently this button is in the position where usually the "Save" modal button is. We could also display it nest to the "general hotkeys" section.

graylog2-web-interface/src/views/components/searchbar/saved-search/SearchActionsMenu.tsx
```diff
@@ -160,6 +163,16 @@ const SearchActionsMenu = () => {
     loadAsDashboard(history, view);
   }, [history, view]);
 
+  const isDirty = useIsDirty();
+
+  useHotkeys('SAVE', () => {
+    saveSearch(title);
+  }, { scopes: 'search', enabled: (isDirty && loaded) }, [loaded, isDirty, saveSearch, title]);
```
linuspahl, line 170: IMO we should not disable the save button.
- when there are no changes, we can still save the search and display the user notification, similar to e.g. a Google Drive document.
- when the search has not been saved yet, we can open the dialog to save it (similar to "save as")

In every case we should avoid displaying the browser dialog, to save the website as HTML when clicking this shortcut.

graylog2-web-interface/src/contexts/HotkeysProvider.tsx
```diff
@@ -0,0 +1,120 @@
...
+import HotkeysContext from 'contexts/HotkeysContext';
+import HotkeysModal from 'contexts/HotkeysModal';
+
+export const defaultHotKeysCollection: HotkeyCollections = {
+  view: {
+    title: 'General',
+    description: 'Main hot keys',
+    actions: {
+      UNDO: { keys: 'mod+z', description: 'Undo last action' },
+      REDO: { keys: 'mod+y', description: 'Redo last action' },
```
linuspahl, line 40: I think we should move undo and redo to the search and dashboard scope.

graylog2-web-interface/src/contexts/HotkeysContext.tsx
linuspahl: Having another look it is fine the way it is, because we use this as a type for the hotkeys collection.

graylog2-web-interface/src/contexts/HotkeysModal.tsx
linuspahl: Now I saw that `scopes` is part of the `react-hotkeys-hook` options argument.

graylog2-web-interface/src/hooks/useHotkeys.tsx
linuspahl: Now I saw that `scopes` is part of the `react-hotkeys-hook` options argument.

dennisoelkers (approved):

PR #16294, opened 2023-08-29
Update opensearch to 2.9.0 in matrix integration tests

todvora:
This PR is updating the opensearch version to 2.9. Additionally it enhances the way we are waiting for the cluster formation in tests, adds logging and error handling and adds caching of the datanode docker builds used in the tests.

/nocl
/jenkins-pr-deps Graylog2/graylog-plugin-enterprise#5682

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #16400, opened 2023-09-08
fix initial uname/pwd for datanode cluster tests

todvora:
/nocl

When opensearch nodes join a cluster, only users from the initial manager node will be available. So we need to make sure we use everywhere the same credentials, otherwise we won't be able to communicate with other nodes from the cluster directly.

## Types of changes

## Checklist:


dennisoelkers (approved):

dennisoelkers:
Closing in favor of #16294.

PR #16407, opened 2023-09-08
Extract "no leader node" notifications from NodePingThread

thll:
When indexing into the system events index is blocked by the indexing backend, the NodePingThread would grind to a halt.

We move the notifications related code into a separate periodical so that the node ping thread won't be affected by issues related to notifications.

Also, the new notification periodical tries to handle automatic leader election more graceful by allowing a time window without an active leader node before emitting a notification.

Fixes Graylog2/graylog-plugin-enterprise#5750


graylog2-server/src/main/java/org/graylog2/periodical/LeaderPresenceCheckPeriodical.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public int getInitialDelaySeconds() {
+        return 0;
+    }
+
+    @Override
+    public int getPeriodSeconds() {
+        return 1;
```
bernd, line 160: Do you think we need to run this every second?

thll: No, this is just how it was before, so I kept the interval. Any preference for an alternative value?

bernd: I talked with @thll offline and we decided to use a five-second interval instead of one second. That's still responsive.

bernd (approved):

PR #16411, opened 2023-09-11
Bump actions/checkout and actions/cache to fix cache restore issue

bernd:
Should fix the following issue on GitHub Actions:

```
The value of "length" is out of range. It must be >= 0 && <= 2147483647.
```
Relevant item from the actions/cache changelog:

> Fixed download issue for files > 2GB during restore.

/nocl


dennisoelkers (approved):

bernd:
@dennisoelkers I added an optimization to cancel running builds when a new ref is pushed.

PR #16473, opened 2023-09-14
Fix Message ingestion with duplicate ids

mpfz0r:
If a Message with a preset `_id` is ingested into multiple indices, we fail with a duplicate key exception
when we try to build the IndexingResults.

This regression was introduced with #16337

- Provide a mergeFunction to `toMap()` to fix this

/nocl


thll (approved):
LGTM

PR #15380, opened 2023-04-28
create additional role Sidecar Manager, reviewed existing permissions

moesterheld:
## Description
reviewed permissions of Sidecar System user, made `sidecars/configurations` unavailable to the user since it is only used from the UI, not the sidecar.
added Sidecar Manager role with permissions to manage and read sidecars

## Motivation and Context
closes #12044 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #16477, opened 2023-09-14
fix timestamp of sidecar roles migration

AntonEbel:
## Description

In #15380, the sidecar manager and reader role was introduced, but the duplicate timestamp in the migration sometimes causes the migration to be ignored.

## Types of changes

/nocl


moesterheld (approved):
lgtm

PR #16503, opened 2023-09-18
Remove check for write block for TSO rotation strategy

### LINKED ISSUE #5722: Access message data from a java program
Hello,

I'm currently working on a way to get data from messages sent to Graylog and use them on a java program.
I can't find any documentation or answer to even know if this is possible.

Does anybody have an idea?

patrickmann:
Resolves #5722

Time/size optimizing Rotation strategy (TSO) was filtering for indices that had the write block set when running retention. If it was not present (i.e. setting the write block had failed), then those indices were not being retained. There are legitimate scenarios where we want to delete archives anyway, e.g. if we are in a high watermark situation and indices have `read_only_allow_delete` block set.

None of the other rotation strategies performs this test, so it is consistent to do the same for TSO. This also avoids lots of downstream issues when indices are unexpectedly not being retained and piling up.

### DIFF
--- a/changelog/unreleased/pr-16503.toml
+++ b/changelog/unreleased/pr-16503.toml
@@ -1,5 +1,5 @@
 type = "f"
 message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
 
-issues = ["5722"]
+issues = ["Graylog2/graylog-plugin-enterprise#5722"]
 pulls = ["16481"]
--- a/changelog/unreleased/pr-16503.toml
+++ b/changelog/unreleased/pr-16503.toml
@@ -1,0 +1,5 @@
+type = "f"
+message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
+
+issues = ["5722"]
+pulls = ["16481"]
--- a/changelog/unreleased/pr-16503.toml
+++ b/changelog/unreleased/pr-16503.toml
@@ -2,4 +2,4 @@ type = "f"
 message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
 
 issues = ["Graylog2/graylog-plugin-enterprise#5722"]
-pulls = ["16481"]
+pulls = ["16503"]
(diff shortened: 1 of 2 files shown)


changelog/unreleased/pr-16503.toml
```diff
@@ -0,0 +1,5 @@
+type = "f"
+message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
+
+issues = ["5722"]
```
mpfz0r, line 4:
```suggestion
issues = ["Graylog2/graylog-plugin-enterprise#5722"]
```

changelog/unreleased/pr-16503.toml
```diff
@@ -0,0 +1,5 @@
+type = "f"
+message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
+
+issues = ["Graylog2/graylog-plugin-enterprise#5722"]
+pulls = ["16481"]
```
mpfz0r, line 5:
```suggestion
pulls = ["16503"]
```

and this one is wrong as well

patrickmann: Must be monday :/

graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/AbstractIndexRetentionStrategy.java
```diff
@@ -72,14 +71,12 @@ public void retain(IndexSet indexSet) {
...
         final Map<String, Set<String>> deflectorIndices = indexSet.getAllIndexAliases();
 
-        final IndicesBlockStatus indicesBlocksStatus = indices.getIndicesBlocksStatus(deflectorIndices.keySet().stream().toList());
         // Account for DST and time zones in determining age
         final DateTime now = clock.nowUTC();
         final long cutoffSoft = now.minus(timeBasedConfig.indexLifetimeMin()).getMillis();
         final long cutoffHard = now.minus(timeBasedConfig.indexLifetimeMax()).getMillis();
         final int removeCount = (int)deflectorIndices.keySet()
                 .stream()
-                .filter(indexName -> indexIsReadOnly(indicesBlocksStatus, indexName))
```
mpfz0r, line 82: We have some corresponeding test code for that.


https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/test/java/org/graylog2/indexer/rotation/strategies/TimeBasedSizeOptimizingRotationAndRetentionTest.java#L130

can you check whether that can be deleted or hast to be adopted?

patrickmann: Tests are not affected by the change and passing. There is actually very little testing of retention - the bulk is for rotation.

mpfz0r (approved):

PR #16509, opened 2023-09-18
Remove check for write block for TSO rotation strategy (5.1 backport)

mpfz0r:
Resolves https://github.com/Graylog2/graylog2-server/issues/5722

Time/size optimizing Rotation strategy (TSO) was filtering for indices that had the write block set when running retention. If it was not present (i.e. setting the write block had failed), then those indices were not being retained. There are legitimate scenarios where we want to delete indices anyway, e.g. if we are in a high watermark situation and indices have read_only_allow_delete block set.

None of the other rotation strategies performs this test, so it is consistent to do the same for TSO. This also avoids lots of downstream issues when indices are unexpectedly not being retained and piling up.

(cherry picked from commit 44059ebfd5fa416afbe6ee52dc756bb50242d8cd)

### DIFF
--- a/changelog/unreleased/pr-16503.toml
+++ b/changelog/unreleased/pr-16503.toml
@@ -1,0 +1,5 @@
+type = "f"
+message = "Fixes failure to delete indices with time/size optimizing strategy, when write block is not present."
+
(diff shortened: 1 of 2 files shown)


kodjo-anipah (approved):
LGTM

PR #16515, opened 2023-09-18
Remove Edit Event Definition button from events scoped as immutable

### LINKED ISSUE #15916: Cannot assign, remove, or edit notification in Illuminate created event in 5.1.x
When trying to add a notification to an Illuminate Event in 5.1.x, you are met with a message that says "This particular Event Definition has been marked as immutable when it was created, therefore it cannot be edited," which prevents you from editing anything about the event in question.  That includes adding, removing, or modifying a notification in that event, making it so that illuminate events cannot have notifications attached to them after install of 5.1.x.  This was tested on 5.1.0, 5.1.1, 5.1.2, and 5.1.3

## Expected Behavior
There is an ability to add, remove, and/or modify a notification in an Illuminate event

## Current Behavior
The UI does not allow you to touch notification settings on an Illuminate event

https://graylog.zoom.us/rec/share/SlEUOmRHhnUOY7wmzoi7CxX-6CB8iNh-VYNvUXNkcahc_nl_-zImH1bqV_qJFOrC.UslD2MlK_vRNbWWg?startTime=1688758535000
Passcode: uppk8&hL

kingzacko1:
## Description

* Do not display `Edit Event Definition` button if the event definition is scoped as immutable. The edit option is already hidden from the More Actions menu. 
* Add a `Duplicate Event Definition` button to the ViewEventDefinition page toolbar

## Motivation and Context

closes #15916 

## Types of changes

## Checklist:


graylog2-web-interface/src/pages/ViewEventDefinitionPage.tsx
```diff
@@ -81,11 +83,13 @@ const ViewEventDefinitionPage = () => {
       <PageHeader title={`View "${eventDefinition.title}" Event Definition`}
                   actions={(
                     <ButtonToolbar>
+                      {isMutable && (
```
ryan-carroll-graylog, line 86: @kingzacko1 What do you think about, in the case where the definition is not mutable, adding some text here explaining that they can't edit "System/Illuminate" event definitions and to make a copy if they want to customize?

Just anticipating people not understanding what's happening and thinking this is a bug or something.

Otherwise this looks solid and tests successfully!

kingzacko1: The `SystemNotificationEventEntityScope` actually is mutable so it would solely be Illuminate events definitions without this button right now. Also, it might look a little out of place if we were to try and shove some text where buttons are supposed to be. What do you think instead about adding some boilerplate to the description of each event definition that gets shipped with Illuminate? Something like `Event definitions shipped with Illuminate cannot be directly modified. Please duplicate this event definition and then make modifications to the copy.` at the end of each description. @ryan-carroll-graylog

ryan-carroll-graylog: Ah yeah good point, yeah I think having a note on the Illuminate event definition works if that's if it's not hard to do.
Were you thinking updating the actual bundles or adding it somewhere in the Illuminate install process?

One other thing I was thinking was swapping the "Edit" button for a "Duplicate" button (Maybe with a tooltip explaining why?), but not sure how feasible that is.

In any case, not really sure how deep we want to go since this is just meant to be a temporary guardrail. Maybe @rich-graylog has a preference?

kingzacko1: I did think about potentially adding a `Duplicate` button in the toolbar just generally for event definitions, I think that would make sense as opposed to making the user click back to the list of event definitions to duplicate it.

kingzacko1: @ryan-carroll-graylog latest changes have a `Duplicate Event Definition` button for all event definition view pages (minus `System notification events`). Duplicating from this page opens the newly created event definition for editing to limit a couple clicks.

ryan-carroll-graylog (approved):
Looks good and tests successfully!

Nice call opening up to the edit page of the event definition copy when the duplicate button is clicked, makes the UX very seamless.

danotorrey (approved):
LGTM and tests successfully too!

PR #16577, opened 2023-09-20
Add new GIM Schema fields, update UPGRADING.md (5.1)

danotorrey:
Backport https://github.com/Graylog2/graylog2-server/pull/16169 to the `5.1` branch.

Core changes needed for https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/1093.

/nocl

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -114,7 +114,7 @@ If you still run any of those inputs, please configure the alternative "Kinesis/
 
 ## CrowdStrike input log parsing changes
 
-Several log parsing changes have been made to the CrowdStrike input.
+Several log parsing changes have been made to the CrowdStrike input in Graylog release 5.1.6.
 
 Added fields:
 `event_created`: Contains the `metadata.eventCreationTime` log value.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -112,6 +112,34 @@ The following inputs are no longer available:
 The inputs were marked as deprecated since Graylog version `3.2.0`.
 If you still run any of those inputs, please configure the alternative "Kinesis/CloudWatch" input instead ahead of upgrading.
 
+## CrowdStrike input log parsing changes
+
+Several log parsing changes have been made to the CrowdStrike input.
+
+Added fields:
+`event_created`: Contains the `metadata.eventCreationTime` log value.
+`vendor_subtype`: Contains the `metadata.eventType` log value.
+`vendor_version`: Contains the `metadata.version` log value.
+`event_source_product`: Contains the static value `crowdstrike_falcon`.
+
+Changed fields:
+- `message`: Now contains the JSON content of the log `event` value, effectively the message payload.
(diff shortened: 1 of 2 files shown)


UPGRADING.md
```diff
@@ -112,6 +112,34 @@ The following inputs are no longer available:
...
+- `user_domain`
+- `user_id`
+- `vendor_event_description`
+- `FILE_NAME`
+- `FILE_PATH`
+- `OBJECTIVE`
+- `TECHNIQUE`
+
+Note that additional CrowdStrike message parsing is expected to be released in a future release of Graylog Illuminate.
+
```
kingzacko1, line 142: I don't know if we usually backport changes to the UPGRADING.md. I thought it was supposed to only be for changes when upgrading to the new major/minor version, but in this case 5.1.0-5.1.5 will not have the changes listed here. Not saying it's right or wrong to backport this, I've just not seen it before.

kingzacko1 (approved):

ryan-carroll-graylog (approved):
Looks good pending Zacks UPGRADING.md question. Not sure about it either.

UPGRADING.md
danotorrey: I agree, this scenario is definitely not that common, and this file usually makes the most sense for major versions and the first minor version. But, since we are introducing breaking log parsing changes in a minor upgrade, I think we should give as much heads up to users as possible and include the changes in this file. I agree that users going from 5.1.5 to 5.1.6 are less likely to see it, but there is also the chance of a user going to 5.1.6 from an older major version, and they should see those changes.

danotorrey: Oops, sorry, I forgot that CrowdStrike was introduced in 5.1, so the older minor version upgrade would not apply. But, I think the changes are good to include, just in case a user does review the file.

danotorrey: @kingzacko1 @ryan-carroll-graylog What do you think about making a note in the new section to indicate that the changes were introduced in 5.1.6 to help with clarity?

ryan-carroll-graylog: Seems solid to me and better to err on the side of more info. At the very least it definitely can't hurt.

kingzacko1: I agree. 99% of the time it probably won't make any difference at all, but I think at least mentioning that these changes aren't present until 5.1.6 is worth it.

danotorrey: I agree, probably very slim chance it's used (especially given how new the CrowdStrike input is, and given the bugs that the input had with it - probably not many folks it at all using it). I went ahead and added the version note. Let me know if that looks good to ya'll.

kingzacko1 (approved):

danotorrey:
The build failure is triggered by a missing `Build repository` Github action in 5.1. https://github.com/Graylog2/graylog2-server/pull/16613 it. Just merge in the changes and will merge once the build turns green.

PR #16652, opened 2023-09-25
Update Error Prone

thll:
* Updates [Error Prone](https://errorprone.info/) from version `2.13.1` to version `2.22.0`.
* Removes `plexus-compiler-javac-errorprone` and runs Error Prone as a compiler plugin.
* Fixes error-prone findings

/nocl
/jpd Graylog2/graylog-plugin-enterprise#5857


bernd (approved):
Thank you very much!

PR #16660, opened 2023-09-25
Make time range preset form in time picker an actual form

linuspahl:
## Description

With this PR we are making the time range preset form in the time range picker an actual `form`, mainly to be able to submit the form by pressing the return key.

/nocl


dennisoelkers (approved):
Looks good, works fine!

PR #16661, opened 2023-09-25
DataNode: Fixing discovery process for new nodes

janheise:
## Description

TODO: instead of only adding the hostname, we need the port, too. For that, I guess, we have to add another field like `transport_address` to the node in mongo that is filled by every DataNode once the config is read.

/nocl

## Types of changes

## Checklist:


data-node/src/main/java/org/graylog/datanode/configuration/OpensearchConfigurationProvider.java
```diff
@@ -120,7 +120,10 @@ private ImmutableMap<String, String> commonOpensearchConfig(final Configuration
                 networkHost -> config.put("network.host", networkHost));
         config.put("path.data", Path.of(localConfiguration.getOpensearchDataLocation()).resolve(nodeName).toAbsolutePath().toString());
         config.put("path.logs", Path.of(localConfiguration.getOpensearchLogsLocation()).resolve(nodeName).toAbsolutePath().toString());
-        config.put("cluster.initial_master_nodes", initialManagerNodes);
+
+        if(StringUtils.isNotBlank(localConfiguration.getInitialManagerNodes())) {
+            config.put("cluster.initial_master_nodes", localConfiguration.getInitialManagerNodes());
+        }
```
luk-kaminski, line 126: I am not completely in the topic, but I just wander if we address somewhere this important advice: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery-bootstrap-cluster.html

```
After the cluster has formed, remove the cluster.initial_master_nodes setting from each node’s configuration. It should not be set for master-ineligible nodes, master-eligible nodes joining an existing cluster, or nodes which are restarting.

If you leave cluster.initial_master_nodes in place once the cluster has formed then there is a risk that a future misconfiguration may result in bootstrapping a new cluster alongside your existing cluster. It may not be possible to recover from this situation without losing data.
```

janheise: no we don't. it is already on my internal task list but I'll put it in our board

dennisoelkers (commented):
Still testing, just leaving one comment.

dennisoelkers (approved):

PR #16723, opened 2023-09-27
Skip building assemblies in full backend tests

thll:
Just a test if we can skip building the assemblies when running the full backend tests.


thll:
When running from within maven, we don't build the assemblies again, anyways.

PR #16749, opened 2023-09-28
Use random http auth password during preflight web setup

todvora:
Because preflight web will be probably used even before configuring SSL/HTTPS for graylog, we don't want to risk sending an actual admin password over unencrypted connection. So we generate a random password during each startup, log it in the logs and assume that only admin can reach it there, use it once, configure the preflight and start the real server. 

This PR also adapts our migrations procedure, splitting it into two parts:

- preflight migrations, like the one that creates the random http auth password for preflight
- all other standard migrations that run _after_ the preflight procedure, during regular startup

/nocl

## How Has This Been Tested?
Added unit tests

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/PreflightJerseyService.java
```diff
@@ -69,18 +73,29 @@ public class PreflightJerseyService extends AbstractIdleService {
...
-        this.basicAuthFilter = requireNonNull(basicAuthFilter, "preflightAuthFilter");
+        this.configuration = requireNonNull(httpConfiguration, "configuration");
+        this.basicAuthFilter = createBasicAuthFilter(localConfiguration);
         this.systemRestResources = systemRestResources;
         this.objectMapper = requireNonNull(objectMapper, "objectMapper");
         this.metricRegistry = requireNonNull(metricRegistry, "metricRegistry");
     }
 
+    @NotNull
+    private static BasicAuthFilter createBasicAuthFilter(Configuration localConfiguration) {
```
luk-kaminski, line 89: Does this method need to be static?

luk-kaminski (approved):
Looks good, only one tiny thing to be potentially improved.

graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/PreflightJerseyService.java
todvora: Hm, intellij refactoring selected the static modifier for me. I assume that removes some risks of calling a method from constructor.

janheise (approved):

graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/PreflightConfigServiceImpl.java
```diff
@@ -16,36 +16,54 @@
...
-    public PreflightConfig saveConfiguration() throws ValidationException {
-        final ImmutableMap<String, Object> fields = ImmutableMap.of("result", PreflightConfigResult.FINISHED);
-        final PreflightConfigImpl config = new PreflightConfigImpl(fields);
-        final String id = save(config);
-        return getPersistedConfig().orElseThrow(() -> new IllegalStateException("Failed to obtain configuration that was just stored"));
+    public String getPreflightPassword() {
+        final DBObject doc = getCollection().findOne(new BasicDBObject("type", "preflight_password"));
+        return Optional.ofNullable(doc)
+                .map(d -> (String) d.get("value"))
+                .orElseThrow(() -> new IllegalStateException("Initial password should be present in the DB, inconsistent state"));
```
luk-kaminski, line 67: I just wander what a user can do after receiving this message? Maybe it should tell what kind of document should be inserted manually to MongoDB to make it run?

todvora: Yeah, I am unhappy with the message as well. It should never happen and if that happens it's completely our fault, the user can't do anything about it. So we should make that clear and tell them to contact us, I believe.

janheise: I'm not sure. At this point, isn't it the case that nothing is really configured at this point, so starting over and dropping the mongo db would be sensible?

dennisoelkers (approved):
Looks good, works fine!

PR #16758, opened 2023-09-28
Custom HTTP notification

kingzacko1:
## Description

Adds a customizable HTTP notification that supports different HTTP methods and content types.
Methods:
- POST
- PUT
- GET (any response from the GET is ignored, but the GET option was requested)

Content Types:
- JSON (empty body has the same body of the legacy HTTP notification)
- URL-encode form data (empty body has a similar representation of the entire event but in URL encoded form data)
- plain text

All content types should support the same JMTE substitution. Creating valid JSON strictly using the JMTE templating is a little tedious if trying to loop through the backlog but this template should print the first 3 fields and then combine the message IDs of each message in the backlog if a backlog exists. Similar logic can be followed to generate more complex bodies.
```
{
  "important_key": "ABCDEFGHIJKLMOPQRSTUVWXYZ",
  "importance": 5,
  "title": "${event_definition_title}",
  ${if backlog}"message_ids": [${foreach backlog message}"${message.id}"${if last_message}${else}, ${end}${end}]${end}
}
```
Because this templating can get wonky and difficult to validate, there is no JSON validation. Not sure if there is a great way to do it either, but if anyone has ideas let me know. For now I am putting together some new documentation for this notification and am just documenting that there is no guarantee the generated JSON is going to be valid if you use JMTE.

## Motivation and Context

Existing HTTP notification lacks flexibility only doing preformatted JSON POSTs.

## Types of changes

## Checklist:


zeeklop (approved):
Tested successfully.

graylog2-web-interface/src/components/event-notifications/event-notification-types/HttpNotificationFormV2.tsx
```diff
@@ -0,0 +1,310 @@
...
+            <Input help="HTTP content type used for POST/PUT notifications"
+                   id="notification-content-type"
+                   label="Content Type">
+              <Select id="content-type"
+                      name="content-type"
+                      options={contentTypes}
+                      matchProp="label"
+                      disabled={config.method === 'GET'}
+                      onChange={this.handleContentTypeChange}
+                      value={config.content_type} />
```
ryan-carroll-graylog, line 271: Noticed you can clear the content type for POST and PUT which causes a very cryptic error.
```suggestion
                      clearable={false}
                      value={config.content_type} />
```

danotorrey:
Thanks for all the work getting this up! I will work on getting a review done by tomorrow morning.

graylog2-server/src/main/java/org/graylog/events/notifications/types/HTTPEventNotificationV2.java
```diff
@@ -0,0 +1,247 @@
...
+            case PUT -> builder.put(RequestBody.create(body, getMediaType(config.contentType())));
+            case POST -> builder.post(RequestBody.create(body, getMediaType(config.contentType())));
+        }
+
+        LOG.debug("Requesting HTTP endpoint at <{}> in notification <{}>", config.url(), ctx.notificationId());
+
+        final OkHttpClient httpClient = selectClient(config);
+        try (final Response r = httpClient.newCall(builder.build()).execute()) {
+            if (!r.isSuccessful()) {
+                throw new PermanentEventNotificationException(
```
ryan-carroll-graylog, line 153: @kingzacko1 what do you think adding some more context to this error message, like the response message and/or error body message? Just from my testing it would've been helpful to debug.

Also wondering if we should add a system notification when there's a failure like we do with some other notifications?: https://github.com/Graylog2/graylog2-server/blob/7b43330f0755aab1e9f1fdf520f3502a4cf6f93b/graylog2-server/src/main/java/org/graylog/integrations/notifications/types/SlackEventNotification.java#L110-L117

ryan-carroll-graylog (approved):
This looks awesome and tests successfully! I can see this opening up a ton of possibilities for users, really nice work!

 I left one comment about adding more detailed error reporting that I think would be helpful since there's there's a lot that could go wrong with the open-endedness of this notification.

One other add I think would be helpful, but may be out of scope for this PR: Have the ability to add custom headers to the notification config. As a proof of concept I wanted to POST to the Graylog API, but it was failing with a 400 error. I was able to get the requests to succeed by adding `.header("X-Requested-By", "graylog-api-user")` to the java code, but I can imagine users might want the ability to add headers for other use cases too.

Other than that everything works great from my testing! Thanks for the callout to  https://requestcatcher.com/ that's a super helpful little tool.

kingzacko1:
@ryan-carroll-graylog I agree the option to add HTTP headers would be useful. I will take a stab at adding that since I think the option to call the Graylog API has already been mentioned as a useful outcome of this notification type. We have some time before 6.0 goes out so I'm happy to keep this PR up to date and only merge it in as a finished product.

danotorrey (commented):
@kingzacko1 Great job putting this together! It provides a lot of flexibility and great new functionality. Well done!!

A couple small questions:
Would it be possible to show the Body Template value on the Details page using the code block formatting similar to the email notification? 

image

Example from the email notification:
image

This might be outside the scope of this PR, and not essential, but it might be handy to have a way to get a sample JSON template from within the edit page. One way might be to pre-fill that, similar to the Email Notification, or to be able to click a button/helper link to fill it. There is already that link to the docs there at the bottom of the form, so this would only be a bonus.

kingzacko1:
@danotorrey @ryan-carroll-graylog latest commit should cover both the presentation of the body template in the Details/Summary as well as adding custom Headers. I am working on putting together a pre-populated template based on content type. Shouldn't be too hard so I wanted to get the other stuff that needs a little more testing in first.

graylog2-server/src/main/java/org/graylog/events/notifications/types/HTTPEventNotificationV2.java
kingzacko1: Also going to add this in the next commit!

kingzacko1: @ryan-carroll-graylog so after playing around with this and looking at some error responses, there really isn't ever going to be much of anything in those error response messages. If I'm missing something let me know, but I think this may be about as good as we can do using this Response object.

ryan-carroll-graylog: Ok yeah I think this is solid with the system notification. I was thinking about adding `r.message()` but it's basically just the plain language version of the response code so I don't think it's really necessary.

ryan-carroll-graylog:
Tested out the use of the headers and it looks great! I'm able to successfully hit the Graylog API now.
The body template in the summary pane looks good too. @kingzacko1

kingzacko1:
@ryan-carroll-graylog made one small change that only populates the default body template if the body is either empty or unchanged from the default for the previous content type. Lemme know how that looks!

danotorrey (approved):
Thanks for the updates @kingzacko1! Looks great and tests successfully!

ryan-carroll-graylog (approved):
Looks and works awesome with the latest! Really looking forward to see how people use this!

kingzacko1:
I'm working on the documentation now and will submit it to the docs team later today so it is on their radar, but going to go ahead and merge this. Thank you all for the continued reviews! @ryan-carroll-graylog @danotorrey

stephanrotolante:
@kingzacko1 Do you know when this functionality will be included in a stable release?

PR #16771, opened 2023-09-29
Datanode: Using retryer for connectivity check, hiding exceptions

janheise:
## Description

Using a retryer & multithreading to test connectivity instead of the serial tests before. Also, errors that are expected, are hidden during connectivity check until there seems to be a real problem.

/nocl

## Types of changes

## Checklist:


dennisoelkers (changes requested):
This is not ideal from my point of view:

  - Instead of blocking the periodical, this is now consuming threads from the `daemonScheduler`. With a frequency of 2 seconds, this could lead to an overconsumption of threads when one or more nodes are not reachable
  - The connectivity check per node is unnecessarily delayed by 1 second

todvora:
This seems like a task for a state machine. We are doing something similar internally in the datanode - we know that it started, configure how many failures are allowed and what to do when the counter is depleted or the call succeeds. Maybe we could do to same here.

janheise:
I changed the thread pool to a cached thread pool which should be exactly what we want regarding resource usage here (unless you add hundreds of datanodes during preflight). see https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/Executors.html#newCachedThreadPool()

dennisoelkers (approved):

dennisoelkers (approved):

PR #16778, opened 2023-09-29
Enable requireUpperBoundDeps rule for Maven Enforcer Plugin

thll:
Enables the [requireUpperBoundDeps](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) rule for the Maven Enforcer Plugin.

To satisfy the rule with the current versions of dependencies, we need to start managing a couple of transitive dependencies.

/nocl


thll:
This is outdated and probably of no use anymore. If we want to give it another try we can as well start from scratch.

PR #16835, opened 2023-10-05
DataNode: Starting LeaderElection Thread in Preflight

### LINKED ISSUE #16830: Data Node preflight doesn't get past certificate provisioning in automatic leader election mode
When automatic leader election mode is enabled. The preflight UI gets stuck after provisioning certificates to the data node(s).

A first superficial analysis suggests that not all periodicals which are required to resume setup have been started on the Graylog node(s).

janheise:
## Description

start/stop leaderElection thread during preflight to support `automatic` mode.

fixes #16830

/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/bootstrap/ServerBootstrap.java
```diff
@@ -208,7 +208,10 @@ private void doRunWithPreflightInjector(Injector preflightInjector) {
 
         final ServiceManager serviceManager = preflightInjector.getInstance(ServiceManager.class);
         final LeaderElectionService leaderElectionService = preflightInjector.getInstance(LeaderElectionService.class);
+        final Service manageableLeaderElectionService = preflightInjector.getInstance(Key.get(Service.class, Names.named("LeaderElectionService")));
```
todvora, line 211: Is there a reason why we can't obtain an instance directly by requesting LeaderElectionService.class ?

janheise: That's why I asked @thll and @mpfz0r for a review, too. If you look into `LeaderElectionModule` you can see that the binding is a bit different from what we usually do. That's why I'm looking it up twice because I need both interfaces to work and don't just want to typecast. Maybe there is a better way?

thll: I don't quite remember why we didn't make the `LeaderElection` interface extend `Service`. I'll look into it first thing next week..

thll: Apparently we have introduced the extra binding [here](https://github.com/Graylog2/graylog2-server/pull/11719). But I can't recall why and I guess we can now just omit binding the Service interface and instead have `LeaderElectionService` extend `Service`. I'll create an extra PR to change that,  so that you can merge this one already. Even though it looks a bit ugly now, I don't think there's anything wrong with it.

thll:
@janheise: is entering the preflight mode restricted to initial cluster setup? It looks like that with the `enable_preflight_web` config setting, a node could enter preflight mode in a cluster which is already productive. Theoretically we could end up in a situation where a node in preflight mode claims cluster leadership. If the node doesn't leave preflight mode in a timely fashion, this will be problematic, because the node won't perform the usual tasks of a cluster leader.

janheise:
@thll thx. for the comments. Preflight mode is only for initial setup, I'll think about it a bit.

thll (approved):

Issue #16826, opened 2023-10-05
Starting "neutral" datanode system package without opensearch fails with long stacktrace


moesterheld:
## Expected Behavior
Log should show an error with a note how to resolve it (install opensearch into dist)

## Current Behavior
Long error message with stacktrace
```
2023-10-05T10:06:24.410+02:00 ERROR [CmdLineTool] Startup error:
com.google.inject.ProvisionException: Unable to provision, see the following errors:

1) [Guice/ErrorInjectingConstructor]: RuntimeException: Failed to list content of provided directory /Users/oesterheld/dev/playground/datanode-test/graylog-datanode-5.2.0-beta.1/dist
  at DatanodeConfigurationProvider.<init>(DatanodeConfigurationProvider.java:36)
  at DatanodeConfigurationProvider.class(DatanodeConfigurationProvider.java:30)
  while locating DatanodeConfigurationProvider
  at OpensearchBinPreflightCheck.<init>(OpensearchBinPreflightCheck.java:41)
      \_ for 1st parameter
  while locating OpensearchBinPreflightCheck
  while locating PreflightCheck annotated with @Element(setName=,uniqueId=4, type=MAPBINDER, keyType=String)
  at PreflightCheckService.<init>(PreflightCheckService.java:34)
      \_ for 1st parameter
  while locating PreflightCheckService

Learn more:
  https://github.com/google/guice/wiki/ERROR_INJECTING_CONSTRUCTOR

1 error

======================
Full classname legend:
======================
DatanodeConfigurationProvider: "org.graylog.datanode.configuration.DatanodeConfigurationProvider"
Element:                       "com.google.inject.internal.Element"
OpensearchBinPreflightCheck:   "org.graylog.datanode.bootstrap.preflight.OpensearchBinPreflightCheck"
PreflightCheck:                "org.graylog2.bootstrap.preflight.PreflightCheck"
PreflightCheckService:         "org.graylog2.bootstrap.preflight.PreflightCheckService"
========================
End of classname legend:
========================

	at com.google.inject.internal.InternalProvisionException.toProvisionException(InternalProvisionException.java:251) ~[guice-6.0.0.jar:?]
	at com.google.inject.internal.InjectorImpl$1.get(InjectorImpl.java:1151) ~[guice-6.0.0.jar:?]
	at com.google.inject.internal.InjectorImpl.getInstance(InjectorImpl.java:1186) ~[guice-6.0.0.jar:?]
[...]

PR #16841, opened 2023-10-06
better error message with instructions for problems with opensearch distribution location

todvora:
Fixes https://github.com/Graylog2/graylog2-server/issues/16826

## Description
The new error message looks like this:

/nocl

```
java.lang.IllegalArgumentException: Failed to list content of provided directory. Directory used for Opensearch detection: /tmp/junit10789252857559246705/nonexistent. Please configure opensearch_location to a directory that contains an opensearch distribution for your architecture x64. You can download Opensearch from https://opensearch.org/downloads.html . Please extract the downloaded distribution and point opensearch_location configuration option to that directory.
```

## Motivation and Context
Let's make our error message as helpful as possible

## How Has This Been Tested?
Added unit tests

## Types of changes

## Checklist:

### DIFF
--- a/data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
+++ b/data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
luk-kaminski, line 1: I am not sure if this record does not need refactoring ...
Record suggests it should be a simple, data-containing type.
However, it is full of static methods related to "detection" functionality.
Constructor is public but is not used outside of this record.

Maybe detection should be moved to another class, which will act as factory for the records? Maybe you how another idea how to make it a little bit cleaner?

todvora: Fair point. This started as a simple record and grown and grown. I think a provider and a pure record would be a way to go. Maybe even the validation would be simplified for that. But I don't want to mess with it inside this PR. So if you agree, I'd suggest to merge this improvement in error message and start a new PR with refactoring.

luk-kaminski: Sure, thank you!

luk-kaminski (approved):
Looks good, thx!

janheise (approved):
LGTM!

Issue #13975, opened 2022-11-15
Preserve "Max number of indices" value when changing index retention strategy


coffee-squirrel:
## What?

If a non-[def](https://github.com/Graylog2/graylog2-server/blob/3cb826f2231906ebaf0c967be90291fa0092151e/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/ClosingRetentionStrategyConfig.java#L32)[ault](https://github.com/Graylog2/graylog2-server/blob/dba006b663632f6265882328d3134c7cb03c3456/graylog2-server/src/main/java/org/graylog2/indexer/retention/strategies/DeletionRetentionStrategyConfig.java#L32) "max number of indices" has been set in the index retention config, that value should be preserved when the index retention strategy gets changed.

For example:  if I have an index set's retention strategy set to "Archive" with a "max number of indices" of 90 and change the strategy to "Delete", I would expect the "max number of indices" to remain at 90 (rather than "changing" to the default of 20).

## Why?

If an admin doesn't notice the "max number of indices" gets reset, it could unintentionally result in indices getting archived, deleted, etc.

## Your Environment

* Graylog Version: 4.3.9
* Browser version: Firefox 106

danotorrey:
Verified. Example of the issue:

![Nov-16-2022 13-49-53](https://user-images.githubusercontent.com/3423655/202280053-001c3d08-cc28-4f4c-86ec-5cf820b8aced.gif)

PR #16860, opened 2023-10-09
Preserve "Max number of indices" value when changing index retention strategy

grotlue:
Closes https://github.com/Graylog2/graylog2-server/issues/13975

/jpd Graylog2/graylog-plugin-enterprise#5941

## Description
This PR adds a context and provider to share the `maxNumberOfIndices` across retention strategy configurations.
The context is used in all components with that field.
The archive strategy is part of enterprise. As the context is not shared properly across different modules it is passed through as a prop in this case.

This change applies to the index configuration (`/system/index_sets/$ID/configuration`) as well as the index sets defaults configuration (`system/configurations/Index%20Set%20Defaults`) and should be tested on both.

## Types of changes

## Checklist:


graylog2-web-interface/src/views/types.ts
```diff
@@ -188,12 +188,14 @@ interface ExportFormat {
   fileExtension: string;
 }
 
+export interface SystemConfigurationComponentProps {
+  config: any,
+  updateConfig: (newConfig: any) => any,
```
gally47, line 193: Here too should the type be `StrategyConfig`?

graylog2-web-interface/src/views/types.ts
```diff
@@ -188,12 +188,14 @@ interface ExportFormat {
...
+}
+
 export interface SystemConfiguration {
   configType: string;
-  component: React.ComponentType<{
-    config: any,
-    updateConfig: (newConfig: any) => any,
```
gally47, line 195: same here, `StrategyConfig`?

gally47 (changes requested):
Code looking good, just one type nitpick for `config`.
I will test the UI next.

gally47 (approved):
Tested it, works as expected. Code LGTM.

Issue #16831, opened 2023-10-05
Search loading indicator has transparent background in dark mode


linuspahl:
## Expected Behavior

The search loading indicator has a transparent background in dark mode, which can make it hard to read the text.
![image](https://github.com/Graylog2/graylog2-server/assets/46300478/9d1e51c6-a797-40b8-bcff-930e43a35abb)

This seems to be related to the `Alert` component.

* Graylog Version: 5.2 beta 1

PR #16869, opened 2023-10-09
Remove opacity of search loading indicator.

linuspahl:
## Motivation and Context

This PR is fixing the opacity of the loading indicator for the dark mode.

Before:
image

After
image

I considered changing the background of the alert component, but adding a container for the loading indicator is the more pragmatic approach.

Fixes: https://github.com/Graylog2/graylog2-server/issues/16831
/nocl


dennisoelkers (approved):
Looks good, works fine!

PR #16877, opened 2023-10-10
Ban AWS SDK bundle dependencies via enforcer plugin

bernd:
The bundle dependencies in v1 and v2 would add hundreds of megabytes to our production artifacts because they bundle all AWS SDK libraries and their transitive dependencies.

Refs https://github.com/Graylog2/graylog-plugin-enterprise/issues/5944
Refs https://github.com/Graylog2/forwarder/issues/121

/nocl


bernd:
I kept it in draft status because we need to remove the transitive AWS SDK bundle dependency from the enterprise plugin before this build can turn green.

thll (approved):

PR #16891, opened 2023-10-11
Fix test day findings

### LINKED ISSUE #16823: Pipeline rule field_long doesn't recognize numerical value
I am trying to setup a pipeline rule that verifies if a field value is numerical. For that I am using the 'field_long' check. It doesn't recognize an int/long value, unless I enable the `attemptConversion` flag, then it works as expected. Am I doing something wrong or is it indeed a bug?

image

## Expected Behavior
The field_long function should recognize numerical value of a field.

## Current Behavior
The value is recognized only if `attemptConversion` is enabled.

1. Setup a rule as described in the attached screenshot
2. Let `attemptConversion` unchecked
3. Provide a demo message containing a field that should be tested
4. Observe the _Condition output_ results in bottom right panel.

## Your Environment

* Graylog Version: Graylog Cloud 5.2.0 (406)
* Browser version: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/118.0

### LINKED ISSUE #16861: Rule Simulator fails silently if "message" is no string
## Expected Behavior
Rule simulator should convert message to string in background

## Current Behavior
If a JSON is entered into the rule simulator where the message field is not a string, it fails because message is supposed to be a string in Graylog and no response can be generated (HTTP 400)

## Possible Solution
cast message to string in simulator response

## Your Environment
* Graylog Version: 5.2.0-beta-1.1

moesterheld:
## Description

- explicitly cast "message" field in simulator to string (fixes #16861)
- add conversions directly to type check functions (fixes #16823 )
  - is_bool, is_long, is_double, is_ip work as expected
- add flag that urlconversion failed to URL object 
  - is_url works as expected
- make value in ip/url conversion accept a variable
- adjust some rulebuilder titles
- add new function groups for new asset/illuminate functions

/nocl

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/conversion/IsBoolean.java
```diff
@@ -23,29 +23,43 @@
...
 
     @Override
     public Boolean evaluate(FunctionArgs args, EvaluationContext context) {
         final Object value = valueParam.required(args, context);
-        return value instanceof Boolean;
+        final boolean convert = conversionParam.optional(args, context).orElse(false);
+        if (!convert) {
+            return value instanceof Boolean;
+        }
+        if (value instanceof Boolean) {
```
kodjo-anipah, line 47: would check here with BooleanUtils.toBooleanObject(value) instanceof Boolean and discard the parsing as for false it would other wise return false but that is still a boolean

kodjo-anipah (changes requested):
some small issues left

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/conversion/IsLong.java
```diff
@@ -23,29 +23,44 @@
...
     public IsLong() {
         valueParam = object("value").ruleBuilderVariable().description("Value to check").build();
+        conversionParam = bool("attemptConversion").optional().description("Try to convert value to long from its string representation (default: false)").build();
     }
 
     @Override
     public Boolean evaluate(FunctionArgs args, EvaluationContext context) {
         final Object value = valueParam.required(args, context);
-        return value instanceof Long;
+        final boolean convert = conversionParam.optional(args, context).orElse(false);
```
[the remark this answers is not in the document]
moesterheld: I know. I just didn't want to break backwards compatibility by changing this. 
I just saw that we don't have the `is_number` as a field condition in the rule builder. I will add that to the migration.

graylog2-server/src/main/java/org/graylog/plugins/pipelineprocessor/functions/conversion/IsBoolean.java
moesterheld: thanks for spotting this

kodjo-anipah (approved):
LGTM

PR #16905, opened 2023-10-12
Refactored all usages of directories used by datanode and managed opensearch

todvora:
Refactor all usages and pointers of datanode and managed-opensearch directories. All are now handled on one place only (DatanodeDirectories class), properly resolved, **using nodeId as a subdir** to avoid configuration collisions. Added javadoc for all methods to make clear what are our expectations and usages of these directories.

Additionally we now check the presence and permissions to each of the directories in advance, during preflight checks. This should reveal any possible problems significantly sooner than during runtime later.

/nocl

## How Has This Been Tested?
Added unit tests, existing set of integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
+++ b/data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
@@ -29,7 +29,7 @@ public record OpensearchDistribution(Path directory, String version, @Nullable S
         this(path, version, null, null);
     }
 
-    public Path getOpensearchBinFile() {
+    public Path getOpensearchExecutable() {
         return directory.resolve(Paths.get("bin", "opensearch"));
     }
 }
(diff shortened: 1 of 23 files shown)


data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
```diff
@@ -28,4 +29,7 @@ public OpensearchDistribution(Path path, String version) {
         this(path, version, null, null);
     }
 
+    public Path getOpensearchBinFile() {
```
janheise, line 32: Maybe name this `getOpenSearchBinDir` ? as with most other paths or really `File`?

todvora: This is really the executable bin file. Maybe renaming to getOpensearchExecutable?

janheise (approved):
LGTM!

data-node/src/main/java/org/graylog/datanode/OpensearchDistribution.java
todvora: Done, thx!

PR #16906, opened 2023-10-12
Data Node: Remove unused OpenSearch plugins.

dennisoelkers:
## Motivation and Context

This PR is removing unused OpenSearch plugins from the filesystem, so they do not end up in tarballs/docker images.

No functionality that we are using should be impacted by this change.

/nocl

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.

PR #16936, opened 2023-10-16
Fixing compilation of single module.

dennisoelkers:
## Motivation and Context

In #16906 an implicit dependency on the test jar was introduced, which breaks compilation if only the `graylog2-server` module is compiled.  This PR fixes this.

In addition, it is adding a test case that makes sure that unwanted plugins are not included/loaded in the data node's OpenSearch distribution.

/nocl

## Types of changes

## Checklist:


data-node/src/test/java/org/graylog/datanode/integration/DataNodePluginsIT.java
```diff
@@ -0,0 +1,79 @@
...
+                        Matchers.not(Matchers.containsString("opensearch-notifications")),
+                        Matchers.not(Matchers.containsString("opensearch-notifications-core")),
+                        Matchers.not(Matchers.containsString("opensearch-performance-analyzer")),
+                        Matchers.not(Matchers.containsString("opensearch-reports-scheduler")),
+                        Matchers.not(Matchers.containsString("opensearch-security-analytics")),
+                        Matchers.not(Matchers.containsString("opensearch-sql"))
+                );
+    }
+
+    private void waitForOpensearch(String baseUrl) throws ExecutionException, RetryException {
```
todvora, line 68: You could use DatanodeOpensearchWait utils, they should be more robust and handle all strange states that opensearch is going through when starting.

todvora (changes requested):
Just one minor suggestion

data-node/src/test/java/org/graylog/datanode/integration/DataNodePluginsIT.java
dennisoelkers: Nice! Can we do this in a follow-up PR to get the reviewbot fixed as soon as possible?

todvora (approved):

PR #16941, opened 2023-10-16
Manage dependencies: avro, commons-compress

thll:
Manage avro and commons-compress to avoid pulling in versions which are affected by the following CVEs:

CVE-2023-39410
CVE-2023-42503

Note: these are not direct dependencies of the Graylog server, so only plugins are affected.

/nocl


mpfz0r (approved):
LGTM :+1:

PR #16942, opened 2023-10-16
refactor datanode directories checks, use configuration validators

todvora:
/nocl

## Description

- Use configuration validators to verify existence and permissions of datanode directories
- Remove not needed datanode directories startup check (replaced by validators)
- Simplify security setup configuration

## Motivation and Context
Datanode directories handling can be simplified significantly with configuration validators. The datanode configuration dir is actually optional, so we have to handle it like this.

## How Has This Been Tested?
Existing set of integration tests
:

## Types of changes

## Checklist:


data-node/src/main/java/org/graylog/datanode/Configuration.java
```diff
@@ -246,22 +248,24 @@ public String getOpensearchConfigLocation() {
      * This is a pointer to a directory holding configuration files (and certificates) for the datanode itself.
      * We treat it as read only for the datanode and should never persist anything in it.
      * Use {@link DatanodeDirectories} to obtain a reference to this directory.
+     *
      */
-    public String getDatanodeConfigurationLocation() {
+    @Nullable
```
janheise, line 253: Is @Nullable really working here?

todvora: What do you mean by "working"? I added the annotation just to signalize that the returned value can be null.

janheise (changes requested):

data-node/src/main/java/org/graylog/datanode/Configuration.java
janheise: I've never seen this annotation to be used for return values. but I'm fine with it.

todvora: Ah, now I understand. It's indeed working: 

image

data-node/src/main/java/org/graylog/datanode/configuration/variants/UploadedCertFilesSecureConfiguration.java
```diff
@@ -16,93 +16,116 @@
  */
 package org.graylog.datanode.configuration.variants;
 
+import org.apache.commons.lang.StringUtils;
```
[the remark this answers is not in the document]
todvora: ok, replaced with builtin methods

janheise (approved):

PR #16962, opened 2023-10-17
Fix logging issues

### LINKED ISSUE #10866: Logging-Messages metrics not updated correctly
## Expected Behavior
I expected changes in the metrics/numbers when changing the log-level from info to debug.

## Current Behavior
I can see a number of messages running through the console but the metrics stay the same (see screenshot)
Screenshot from 2021-06-16 11-19-46

1. Go to "System->Logging"
2. Press button "Show log level metrics"
3. Change the log-level for a subsystem from info to debug
4.

## Your Environment

* Graylog Version: 4.10-beta1
* Java Version:
* Elasticsearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

### LINKED ISSUE #16961: Cannot set custom log level using API
In 5.2, it is not possible to set the log level for a logger which hasn't been defined in `log4j2.xml` using the `cluster/system/loggers` endpoint

This is caused by a change introduced when switching to log4j 2.20.0 (introduced in 2.17.2, works fine in 2.17.1).
Here, the root logger is created with `additivity=false`, leading to newly created loggers missing the appenders of the root logger because we create them with the additivity of the root logger.

see https://logging.apache.org/log4j/log4j-2.4/manual/configuration.html#Loggers
"The root logger does not support the additivity attribute since it has no parent."

Thus, we shouldn't rely on the additivity attribute of the root logger

## Your Environment
* Graylog Version: 5.2
* Java Version:
* OpenSearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

moesterheld:
## Description
resolves two logging related issues

## Motivation and Context
fixes #10866 
fixes #16961 

## How Has This Been Tested?
- Go to "System/Logging"
- Change log level for "Subsystem / Authorisation" to debug
- make sure the numbers for debug messages increases
- as this creates a new logger for `org.apache.shiro` which hasn't been defined in `log4j2.xml` yet, #16961 can also be tested

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
@@ -152,7 +152,8 @@ public class LoggersResource extends RestResource {
         if(loggerName.equals(loggerConfig.getName())) {
             loggerConfig.setLevel(level);
         } else {
-            final LoggerConfig newLoggerConfig = new LoggerConfig(loggerName, level, loggerConfig.isAdditive());
+            boolean additive = LogManager.ROOT_LOGGER_NAME.equals(loggerConfig.getName()) || loggerConfig.isAdditive();
+            final LoggerConfig newLoggerConfig = new LoggerConfig(loggerName, level, additive);
             newLoggerConfig.setLevel(level);
             config.addLogger(loggerName, newLoggerConfig);
         }
--- a/graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
+++ b/graylog2-server/src/main/java/org/graylog2/bootstrap/CmdLineTool.java
@@ -320,7 +320,7 @@ public abstract class CmdLineTool implements CliCommand {
             System.exit(1);
         }
 
-        addInstrumentedAppender(metricRegistry, logLevel);
+        addInstrumentedAppender(metricRegistry);
         // Report metrics via JMX.
         final JmxReporter reporter = JmxReporter.forRegistry(metricRegistry).build();
         reporter.start();
@@ -396,13 +396,13 @@ public abstract class CmdLineTool implements CliCommand {
         context.updateLoggers(config);
     }
 
-    private void addInstrumentedAppender(final MetricRegistry metrics, final Level level) {
+    private void addInstrumentedAppender(final MetricRegistry metrics) {
         final InstrumentedAppender appender = new InstrumentedAppender(metrics, null, null, false);
         appender.start();
 
         final LoggerContext context = (LoggerContext) LogManager.getContext(false);
         final org.apache.logging.log4j.core.config.Configuration config = context.getConfiguration();
-        config.getLoggerConfig(LogManager.ROOT_LOGGER_NAME).addAppender(appender, level, null);
+        config.getLoggerConfig(LogManager.ROOT_LOGGER_NAME).addAppender(appender, null, null);
         context.updateLoggers(config);
     }
 


graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -152,7 +152,8 @@ protected void setLoggerLevel(final String loggerName, final Level level) {
         if(loggerName.equals(loggerConfig.getName())) {
             loggerConfig.setLevel(level);
         } else {
-            final LoggerConfig newLoggerConfig = new LoggerConfig(loggerName, level, loggerConfig.isAdditive());
+            boolean additive = LogManager.ROOT_LOGGER_NAME.equals(loggerConfig.getName()) || loggerConfig.isAdditive();
```
thll, line 155:
```suggestion
            final boolean additive = LogManager.ROOT_LOGGER_NAME.equals(loggerConfig.getName()) || loggerConfig.isAdditive();
```

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -152,7 +152,8 @@ protected void setLoggerLevel(final String loggerName, final Level level) {
...
+            final LoggerConfig newLoggerConfig = new LoggerConfig(loggerName, level, additive);
```
thll, line 156: I'm curious. What's the use case for a new logger without additivity? We are not adding any appenders ourselves, or are we? So that would result in a logger without appenders. Is that even useful or could we just always set `additive` to true?

thll (changes requested):
Thanks for spotting and fixing this. Tests successful!

I just have two comments.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
moesterheld: In theory, if a logger is created in `log4j2.xml` for `org.graylog2` with `additivity=false` to use for example different appenders, this would ensure that child loggers for e..g. `org.graylog2.indexer` would inherit that config.

We don't use that and I don't know if anybody does, but I didn't want to break that compatibility.

moesterheld: oh, now I see your point. `org.graylog2.indexer` won't output anything because this way it doesn't have any appenders.
we would either have to add the parent's appenders, but does that make much sense? maybe we should really just set it to `true`

graylog2-server/src/main/java/org/graylog2/rest/resources/system/logs/LoggersResource.java
```diff
@@ -74,7 +74,7 @@ public class LoggersResource extends RestResource {
 
     private static final Map<String, Subsystem> SUBSYSTEMS = ImmutableMap.<String, Subsystem>of(
             "graylog", new Subsystem("Graylog", ImmutableList.of("org.graylog2", "org.graylog"), "All messages from Graylog-owned systems."),
-            "indexer", new Subsystem("Indexer", "org.elasticsearch", "All messages related to indexing and searching."),
+            "indexer", new Subsystem("Indexer", "org.opensearch", "All messages related to indexing and searching."),
```
thll, line 77: Did that end up in here on purpose?

thll (approved):
Code looks good. I didn't test this again but I'm approving so that this PR is not blocked. 

One comment about a change that looks like it made it in accidentally, please have a look at it.

PR #16963, opened 2023-10-17
Add execution duration to jobs overview

mpfz0r:
The duration info will be displayed as a tool-tip on the status badge. Knowing how long it took a job (e.g. archive restore) to complete can be helpful.

Furthermore, this info is persisted with the trigger in the DB, so we could also build something like: show me the (current) slowest running event definition.

/nocl

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
+++ b/graylog2-server/src/main/java/org/graylog/scheduler/DBJobTriggerService.java
@@ -30,6 +30,7 @@ import org.graylog2.database.MongoConnection;
 import org.graylog2.plugin.system.NodeId;
 import org.graylog2.shared.utilities.MongoQueryUtils;
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.mongojack.DBCursor;
 import org.mongojack.DBQuery;
 import org.mongojack.DBQuery.Query;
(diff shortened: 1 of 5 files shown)


kodjo-anipah (changes requested):
LGTM for regular job overview but instant archiving is using a different API. I think it will be easier for Ousmane to include that information if we add it to the DTO InstantArchivingJobProgress org.graylog.plugins.storage.server.InstantArchivingService#getInstantArchivingJobProgress

mpfz0r:
> LGTM for regular job overview but instant archiving is using a different API. I think it will be easier for Ousmane to include that information if we add it to the DTO InstantArchivingJobProgress org.graylog.plugins.storage.server.InstantArchivingService#getInstantArchivingJobProgress

Yes, but I was wondering if this isn't just redundant code. Ideally we should be using the generic endpoint everywhere and delete that one.
@ousmaneo is that a lot of effort to move that over?

We can also just leave it as it is. It's more of a debugging feature and having one place in the UI to look at it is probably ok.

kodjo-anipah (approved):

ousmaneo:
> > LGTM for regular job overview but instant archiving is using a different API. I think it will be easier for Ousmane to include that information if we add it to the DTO InstantArchivingJobProgress org.graylog.plugins.storage.server.InstantArchivingService#getInstantArchivingJobProgress
> 
> Yes, but I was wondering if this isn't just redundant code. Ideally we should be using the generic endpoint everywhere and delete that one. @ousmaneo is that a lot of effort to move that over?
> 
> We can also just leave it as it is. It's more of a debugging feature and having one place in the UI to look at it is probably ok.

Hi @mpfz0r, @kodjo-anipah it's not a lot of effort to unify it I think it would be good to do it, I will start it and open a PR.

Issue #16096, opened 2023-08-02
Absolute Datetime picker jump to previous day when we click on beginning of the day


ousmaneo:
When selecting Absolute date for the search, the selected day jumps to the previous day when we click the "`Beginning of the day`" Hourglass icon. All subsequent day selections afterward also jump to the previous day. (See below)

![Absolute-time-issue](https://github.com/Graylog2/graylog2-server/assets/1381455/5e9e1294-236a-4af5-a55c-7131fbb23728)
 

## Steps to Reproduce (for bugs)

1. Go to search
2. Select absolute datetime
3. Click on the hourglass icon for `from` or `to`
4.

## Your Environment

* Graylog Version:
* Java Version:
* OpenSearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

linuspahl:
It looks like this error also occurs for the date time picker in the sigma rules overview.

![image](https://github.com/Graylog2/graylog2-server/assets/46300478/99ac92a8-65b9-4576-b8ce-04f3b29ec332)

PR #16973, opened 2023-10-18
Correctly highlight selected day in date picker.

linuspahl:
**Please note**, this PR needs a backport for `5.0` and `5.1`

## Motivation and Context

As described in https://github.com/Graylog2/graylog2-server/issues/16096 the highlight for the selected date in the date picker can be wrong for a date time like `2023-10-17 00:00:00`. In this case it displayed the previous day as active.

The reason is that the date time does not contain a time zone information and `moment()` still uses the estimated user time zone internally. For example when your browser timezone is `Europe/Paris` the UTC time for `2023-10-17 00:00:00` is `2023-10-16 22:00:00`.

We are now:
- using `moment.utc()` instead of `moment()` to make it's behaviour predictable for date times without a time zone.
This only effects few cases, because usually timestamps in the frontend have a timezone.
- considering the user time zone when handling dates in the date picker.

There are more improvements we can make regarding the date time utils, but I will do them in a follow-up PR, so this PR will be easy to backport.

Fixes: https://github.com/Graylog2/graylog2-server/issues/16096
Fixes: https://github.com/Graylog2/graylog-plugin-enterprise/issues/5998

### DIFF
--- a/changelog/unreleased/issue-16096.toml
+++ b/changelog/unreleased/issue-16096.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Fix timezone issue with date picker, which resulted in higlighting the wrong selected day."
+
+issues = ["16096"]
+pulls = ["16973"]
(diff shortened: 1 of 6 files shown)


ousmaneo (approved):
Tested and works. Thank you for this fix.

PR #16999, opened 2023-10-19
Fixing extraction order of auth infos in proxied resources.

### LINKED ISSUE #16985: Data Node: Cached basic auth credentials cause problems after completed preflight mode
## Expected Behavior

After completing datanode setup in the preflight mode, the Graylog UI should be fully functional without requiring to close and re-open the browser.

## Current Behavior

After completing datanode setup in preflight mode, the browser automatically presents the Graylog login. After logging in as an admin user, the UI appears to work but has subtle failures: the input counter doesn't work and trying to open any of the node details pages on `system/nodes` fails with an error.

## Possible Solution

I noticed that along with the auth cookie, the browser sends an Authorization header containing the basic auth credentials from preflight mode with every request to the backend. It looks like as if the backend treats the cookie with precedence when it authenticates the request, so we are logged in as the correct user. However, when a request is a "proxied request", i.e. it triggers secondary calls by the receiving node to other nodes in the cluster, the Authorization header is consulted.

We should make sure that we use the same source of authentication for both first-hand and proxied calls. Also, if it would be possible to clear the basic auth credentials from the browser cache, that would also be beneficial. However to my understanding this might be tricky.

This piece of code might be worth looking at: https://github.com/Graylog2/graylog2-server/blob/36ea21727f34aa19c9a9f9de4983d5a5d64e122d/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java#L111

1. Start a new cluster and complete preflight mode.
2. Log into Graylog without closing the browser.
3. Observe above mentioned irregularities.

## Your Environment

* Graylog Version: 5.2.0-beta.3 (Docker)
* Java Version:
* OpenSearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

dennisoelkers:
## Motivation and Context

The `ProxiedResource` superclass used by REST resources which forward requests to other Graylog nodes in the cluster is providing logic to extract the authentication info of the current user in order to add it to the request sent to another Graylog node.

Prior to this PR, the extraction logic is operating in this order, returning early when the first one succeeds:

  - Extracting the `Authorization` header if present
  - Extracting a session cookie if present

For our general authentication logic, this order is reversed. This is a problem when a request has both, the `Authorization`-header as well as a session cookie.

This PR is now aligning the logic of the `ProxiedResource` with the general authentication logic, assigning precedence to the session cookie over the `Authorization` header.

Fixes #16985.

## Types of changes

## Checklist:

### DIFF
--- a/changelog/unreleased/pr-16999.toml
+++ b/changelog/unreleased/pr-16999.toml
@@ -1,0 +1,5 @@
+type = "f"
+message = "Aligning order of precedence of auth info in proxied resources with general authentication logic."
+
+issues = ["16985"]
+pulls = ["16999"]
--- a/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/shared/rest/resources/ProxiedResource.java
@@ -109,17 +109,17 @@ public abstract class ProxiedResource extends RestResource {
     }
 
     public static String authenticationToken(HttpHeaders httpHeaders) {
-        final List<String> authorizationHeader = httpHeaders.getRequestHeader("Authorization");
-        if (authorizationHeader != null && !authorizationHeader.isEmpty()) {
-            return authorizationHeader.get(0);
-        }
-
         final Cookie authenticationCookie = httpHeaders.getCookies().get("authentication");
         if (authenticationCookie != null) {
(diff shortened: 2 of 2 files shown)


todvora (approved):
LGTM!

PR #17000, opened 2023-10-19
Propagating error during field type change to frontend.

### LINKED ISSUE #16978: Error in field type mapping doesn't propagate to frontend
If a change in field mapping fails, we are just logging the error but never propagate it to the frontend. So the user will never know that there was a problem and the change is has not applied properly.

image
 

## Expected Behavior
Log the error **and** propagate it to the frontend (re-throw at least).

## Current Behavior
Only logging the error (swallowing it)

## Possible Solution
Log the error **and** propagate it to the frontend (re-throw at least).

## Your Environment

* Graylog Version: 5.2 beta

dennisoelkers:
## Motivation and Context

This PR is making sure that any error during a field type change is propagated to the frontend.

Fixes #16978.

/nocl

## Types of changes

## Checklist:


todvora (approved):
Easy fix, thanks!

PR #17057, opened 2023-10-24
Index set field types list

luk-kaminski:
## Description
Index set field types list for per-index-set field type management page.
/nocl

## Motivation and Context
Provide a list of field types for UI table of field types for certain index set.

## How Has This Been Tested?
Manually and with unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
@@ -170,6 +170,25 @@ public interface IndexSet extends Comparable<IndexSet> {
      */
     IndexSetConfig getConfig();
 
+    default String getNthIndexBeforeActiveIndexSet(final int n) {
+        final String activeWriteIndex = getActiveWriteIndex();
+        if (activeWriteIndex != null) {
+            final Optional<Integer> deflectorNumber = extractIndexNumber(activeWriteIndex);
+            final String indexPrefix = getIndexPrefix();
+            return deflectorNumber
+                    .map(num -> {
+                        final int indexNumber = num - n;
+                        if (indexNumber >= 0) {
+                            return indexPrefix + MongoIndexSet.SEPARATOR + indexNumber;
+                        }
+                        return null;
+                    })
+                    .orElse(null);
+        }
+
+        return null;
+    }
+
     class IndexNameComparator implements Comparator<String> {
         private final IndexSet indexSet;
 
--- a/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/utils/FieldTypeDTOsMerger.java
+++ b/graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/utils/FieldTypeDTOsMerger.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the Server Side Public License, version 1,
+ * as published by MongoDB, Inc.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Server Side Public License for more details.
+ *
+ * You should have received a copy of the Server Side Public License
+ * along with this program. If not, see
+ * <http://www.mongodb.com/licensing/server-side-public-license>.
+ */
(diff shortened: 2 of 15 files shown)


graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
```diff
@@ -170,6 +170,25 @@ public interface IndexSet extends Comparable<IndexSet> {
      */
     IndexSetConfig getConfig();
 
+    default String getNthIndexBeforeActiveIndexSet(final int n) {
```
todvora, line 173: Could we return optional instead of nulls here?

graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/utils/FieldTypeDTOsMerger.java
```diff
@@ -0,0 +1,49 @@
...
+package org.graylog2.indexer.fieldtypes.utils;
+
+import org.graylog2.indexer.fieldtypes.FieldTypeDTO;
+import org.graylog2.indexer.indexset.CustomFieldMappings;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class FieldTypeDTOsMerger {
```
todvora, line 26: I'd suggest to simplify that class to a utility (=static method call). You will still keep the separated logic, simple unit test and get rid of all the dependency injection, pulling it through constructors...

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/responses/IndexSetFieldType.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Comparator;
+import java.util.List;
+
+public record IndexSetFieldType(@JsonProperty(FIELD_NAME) String fieldName,
+                                @JsonProperty(TYPE) String type,
+                                @JsonProperty(IS_CUSTOM) boolean isCustom,
+                                @JsonProperty(IS_RESERVED) boolean isReserved) {
+
+    public static final String FIELD_NAME = "field_name";
```
todvora, line 32: What about making these constants private?

graylog2-server/src/test/java/org/graylog2/indexer/fieldtypes/IndexFieldTypesListServiceTest.java
```diff
@@ -0,0 +1,150 @@
...
+    IndexFieldTypesListService toTest;
+
+    @Mock
+    IndexFieldTypesService indexFieldTypesService;
+    @Mock
+    IndexSetService indexSetService;
+    @Mock
+    MongoIndexSet.Factory indexSetFactory;
+    @Mock
+    FieldTypeDTOsMerger fieldTypeDTOsMerger;
```
todvora, line 57: I wouldn't mock this field here if it's not really used and usage of the proper class doesn't cause any issue.

todvora (approved):
I have some minor suggestions but don't want to block your work. Feel free to merge.

graylog2-server/src/main/java/org/graylog2/rest/resources/system/indexer/responses/IndexSetFieldType.java
luk-kaminski: I will make them package-protected, because of the usage in tests. It is also possible that in the future they may need to go back to public, but we will see...

graylog2-server/src/main/java/org/graylog2/indexer/IndexSet.java
luk-kaminski: All index-getting methods in IndexSet class return plain String, not Optional<String>. Do you believe that it is not inconsistent to use different approaches? If we believe Optionals would be better, shouldn't we have a separate task to change it everywhere in this class, to be consistent?

graylog2-server/src/main/java/org/graylog2/indexer/fieldtypes/utils/FieldTypeDTOsMerger.java
luk-kaminski: I do not generally like using static methods. 
I am also aware that merging is something that can be changed (for now it was decided to use 2 latest indices and custom mappings), so I am pretty happy that this is clearly visible as a separate dependency, not just a call to a static method.

luk-kaminski:
> I have some minor suggestions but don't want to block your work. Feel free to merge.

Thanks, Tomas! I have made changes according to your 2 suggestions and commented on 2 remaining ones. If you believe that they deserve further discussion outside of this PR, please let me know.

PR #17087, opened 2023-10-26
Exclude non-message streams from search in aggregation events

patrickmann:
Relates to Graylog2/graylog-plugin-enterprise#6042

When no source stream is given for an event definition we use all streams in the ES/OS search query. We now exclude all non-message streams. 

## Motivation and Context

Users cannot select non-message streams as a source stream in the event definition. But these were still being included in the search when no source stream had been explicitly defined. This is unnecessary, inconsistent, and hurts performance. It may also push the search URI over the length limit, when there is a large number of streams.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
+++ b/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
@@ -43,6 +43,7 @@ import org.graylog.events.search.MoreSearch;
 import org.graylog.plugins.views.search.elasticsearch.ElasticsearchQueryString;
 import org.graylog.plugins.views.search.errors.ParameterExpansionError;
 import org.graylog.plugins.views.search.errors.SearchException;
+import org.graylog.plugins.views.search.rest.PermittedStreams;
 import org.graylog.plugins.views.search.searchtypes.pivot.HasField;
 import org.graylog.plugins.views.search.searchtypes.pivot.SeriesSpec;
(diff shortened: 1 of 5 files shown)


graylog2-server/src/main/java/org/graylog/events/processor/aggregation/AggregationEventProcessor.java
```diff
@@ -219,7 +220,7 @@ private Set<String> getStreams(AggregationEventProcessorParameters parameters) {
 
     private void filterSearch(EventFactory eventFactory, AggregationEventProcessorParameters parameters,
                               EventConsumer<List<EventWithContext>> eventsConsumer) throws EventProcessorException {
-        final Set<String> streams = getStreams(parameters);
+        final Set<String> streams = eventStreamService.buildEventSourceStreams(getStreams(parameters), Collections.emptySet());
```
thll, line 223: I think this is the wrong method, because, as the name suggests, it is supposed to operate on the results of a search. I think we should use the same method that we use for aggregations [here](https://github.com/Graylog2/graylog2-server/blob/bf91ef97a613a20df8d37aba7879445d8ff5ba45/graylog2-server/src/main/java/org/graylog/events/processor/aggregation/PivotAggregationSearch.java#L372).

It appears as if that one will also take care of filtering out the investigation streams, if the enterprise plugin is present. Otherwise we'll end up using different stream filters on the search queries for aggregation and filter events.

thll (changes requested):
The current implementation produces different search query streams filters for aggregation events and for filter events. I think those two should be consistent.

thll (approved):

Issue #16836, opened 2023-10-05
Change field type dialog doesn't accurately describe the needed steps


Joel-Duffield-Graylog:
The current version of the change field type dialog doesn't directly call out that this change will only apply in the future and that the index must be rotated, and some other missing information.

## Current Behavior
The dialog currently has this wording
![image](https://github.com/Graylog2/graylog2-server/assets/108024302/9c5c7b7a-dc39-46d1-9aac-71cb6b0346f7)

## Possible Solution
* The dialog should list the current field type somewhere near the new field type and call out the name of the field again (it feels hidden in the title of the dialog, sort of like a before and after on the left and right kind of thing.
* The dialog should maybe explicitly lay out the steps, 1. make this change, 2. rotate the indexes, 3. all new messages fields will now use this type, 4. any widgets that cross the time before and after this change not work due to mixed types. etc
* Maybe even have rotate indexes checked by default.

## Context

This is a great much needed feature, but it could cause a lot of pain for people that don't fully understand it.

## Your Environment

* Graylog Version: 5.2 Beta 1

PR #17119, opened 2023-10-30
Refactoring field type management. Add more info details.

maxiadlovskii:
## Description
This PR adds more information to clarify which changes should be made and why 
1. Adds more field name references
2. Always shows index selection table
3. Change titles
4. Select rotation by default
5. add rotation info hint

## Motivation and Context
fix: https://github.com/Graylog2/graylog2-server/issues/16836

## Types of changes

## Checklist:

/nocl


graylog2-web-interface/src/views/logic/fieldactions/ChangeFieldType/IndexSetsTable.tsx
```diff
@@ -110,56 +108,43 @@ const IndexSetsTable = ({ field, setIndexSetSelection, fieldTypes, initialSelect
...
-                                                   pageSize={layoutConfig.pageSize}
-                                                   visibleColumns={layoutConfig.displayedAttributes} />
-                )}
-              </PaginatedList>
-            </Container>
-          )
-        }
-    </>
+    <Container>
+      <ControlLabel>Select targeted index sets</ControlLabel>
```
linuspahl, line 117: Sometimes the titles in this modal are capitalized, sometimes they are not. It would be nice to unify their styling by capitalizing all headlines.

graylog2-web-interface/src/views/logic/fieldactions/ChangeFieldType/ChangeFieldTypeModal.tsx
```diff
@@ -132,9 +132,14 @@ const ChangeFieldTypeModal = ({ show, onClose, field }: Props) => {
                         required />
         </Input>
         <Alert bsStyle="info">
-          By default the type will be changed in all index sets of the current message/search. By expanding the next section, you can select for which index sets you would like to make the change.
+          By default the <b>{newFieldType}</b> as a field type for <b>{field}</b> will be changed in all index sets of the current message/search. You can select for which index sets you would like to make the change.
```
linuspahl, line 135: When there is no field type selected it reads like this "By default the as a field type for **action** will be changed in all index sets"
When a field type got selected, the sentence it still reads strange: "By default the **boolean** as a field type for **action** will be changed in all index sets "

graylog2-web-interface/src/views/logic/fieldactions/ChangeFieldType/ChangeFieldTypeModal.tsx
```diff
@@ -132,9 +132,14 @@ const ChangeFieldTypeModal = ({ show, onClose, field }: Props) => {
...
         </Alert>
         <IndexSetsTable field={field} setIndexSetSelection={setIndexSetSelection} fieldTypes={fieldTypes} initialSelection={initialSelection} />
+        <ControlLabel>Select rotation strategy
+          <HoverForHelp displayLeftMargin>
+            To see and use the <b>{newFieldType}</b> as a field type for <b>{field}</b>, you have to rotate indices. You can automatically rotate affected indices after submitting this form or do that manually later.
+          </HoverForHelp>
+        </ControlLabel>
```
linuspahl, line 142: I was wondering why you are using the `ControlLabel` component here and in other cases, instead of a headline? Probably to make the  "Select Field Type For field" a label, which is connected with the input?

Since the label / headline is not connected to an input in all other cases, I was wondering if headlines make more sense.

I was also wondering why you used an alert for the "Select Field Type For controller" description and a popover for the "Select rotation strategy" description, don't they serve the same purpose?

I tried out using actual headlines (`h5`, because the modal headline seems to be `h4`, with a bottom margin of 5px) and paragraphs instead of an alert / popover and imo it looks a bit cleaner.

Before:
image


After:
image

linuspahl: I just saw that the "select new field type" description should be the description for the "select targeted index sets", but I think the point I am trying to make is still valid.

linuspahl: We could still add a description for the "Select New Field Type" section like "Select a new field type for the **controller** field.", instead of mentioning the field type in the section headline ("Select Field Type For controller")

linuspahl (approved):

PR #17136, opened 2023-10-31
set state to SKIPPED if "resume startup" is called with no data nodes

### LINKED ISSUE #17117: Preflight mode without datanode ends up with IllegalStateException
If the user opens the preflight configuration for the first time and decides to "resume startup" without configuring any datanodes, the app, due to IndexerDiscoveryProvider, triggers IllegalStateException.

image

## Expected Behavior
Before datanode and preflight, the default behaviour was to use whatever listens on `http://127.0.0.1:9200` if the user does not configure anything else. 

## Current Behavior
We don't discover any datanodes but we see that the user finished the preflight (`PreflightConfigResult.FINISHED`), so we asume the datanode path and fail immediately.  

## Possible Solution
* Persist the PreflightConfigResult.SKIPPED if user decides to resume startup without any datanodes
* Warn user that resuming startup will require either a datanode instance or explicitly configured indexer in the config file

1. Start opensearch on localhost:9200
2. Start fresh GL server
3. Open the preflight interface
4. Don't change anything, click resume startup
5.

## Your Environment

* Graylog Version:
* Java Version:
* OpenSearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

moesterheld:
## Description
In the preflight UI, the preflight configuration state is now set to `SKIPPED` if "resume startup" was clicked and there were no data nodes available.

/nocl

## Motivation and Context
In the `SKIPPED` state, Graylog will fall back to using localhost:9200 in the IndexerDiscoveryProvider.
Before, it was always set to `FINISHED`, whether there were data nodes available or not, causing the server to fail at startup with an `IllegalStateException`

fixes #17117 

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/web/resources/PreflightStatusResource.java
+++ b/graylog2-server/src/main/java/org/graylog2/bootstrap/preflight/web/resources/PreflightStatusResource.java
@@ -23,7 +23,6 @@ import org.graylog2.bootstrap.preflight.PreflightConfigResult;
 import org.graylog2.bootstrap.preflight.PreflightConfigService;
 import org.graylog2.bootstrap.preflight.PreflightConstants;
 import org.graylog2.plugin.Version;
-import org.graylog2.plugin.database.ValidationException;
 
 import javax.inject.Inject;
 import javax.ws.rs.GET;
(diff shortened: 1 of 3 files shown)


todvora (approved):
LGTM, thanks!

PR #17182, opened 2023-11-06
Converts the time format without adjusting to the user timezone

simonychuang:
Currently for a user with a timezone of UTC(-X) when using the date picker the displayed selected date is one day before the actual clicked date.

please see https://github.com/Graylog2/graylog-plugin-enterprise/pull/6095

/nocl

## How Has This Been Tested?

Tested manually on local env

## Types of changes

## Checklist:


ryan-carroll-graylog (approved):
Looks good and tests successfully!

danotorrey (approved):
LGTM and tests successfully with users in timezones that span different days.

ryan-carroll-graylog:
@simonychuang looks like some front end tests are failing if you're not already tracking.

simonychuang:
> @simonychuang looks like some front end tests are failing if you're not already tracking.

@ryan-carroll-graylog Yeah going to check with Linus because it looks like it was intentional to convert the date into the user timezone so just going to make sure we don't break anything by making this change

ryan-carroll-graylog:
> > @simonychuang looks like some front end tests are failing if you're not already tracking.
> 
> @ryan-carroll-graylog Yeah going to check with Linus because it looks like it was intentional to convert the date into the user timezone so just going to make sure we don't break anything by making this change

Ah ok gotcha, good call.

linuspahl:
I can have a detailed look, once I am back next week.

In every case we intentionally convert the date into the user timezone. Here is the related issue:
https://github.com/Graylog2/graylog2-server/issues/16096
and the related PR: https://github.com/Graylog2/graylog-plugin-enterprise/pull/5555

I can imagine the described problem will occur again with these changes.

My guess is we need to adjust the format outside of this component, before we make the request.

linuspahl (changes requested):
I gave it a try and when I am testing it with a user that has e.g. the timezone Europe/Berlin, I can reproduce the bug described in https://github.com/Graylog2/graylog2-server/issues/16096

linuspahl:
I can also see a benefit in always using the internal format (ISO 8601) for the new selected date time, provided by the date time picker `onChange` method.

We could just implement

```
const _onChange = useCallback((newDate: Date, modifiers: DayModifiers, event: React.MouseEvent<HTMLDivElement>) => (
    onChange(adjustFormat(newDate, 'internal'), modifiers, event)
  ), [onChange]);```

and pass it to the day picker. But we should still check if this causes problems in other cases where to implement the date picker.

zeeklop (approved):
Looks good to me. Nice work!

Tested with `Europe/Berlin` `America/Argentina/Buenos_Aires`  and `America/New_York` all 3 time zones worked as expected.

linuspahl:
Closing this in favour of https://github.com/Graylog2/graylog2-server/pull/17677

PR #17200, opened 2023-11-07
Request a dispatched PR build

mpfz0r:
Dispatch a PR build to graylog-project-internal.
Provide all the necessary informations so that the dispatched build can update the commit status on the calling PR.

/prd https://github.com/Graylog2/graylog-plugin-enterprise/pull/6111

/nocl

### DIFF
--- a/.github/workflows/dispatch-pr-build.yml
+++ b/.github/workflows/dispatch-pr-build.yml
@@ -30,7 +30,7 @@ jobs:
         run: >
           gh workflow run -R Graylog2/graylog-project-internal pr-build.yml
           -f caller_repo=${{ github.repository }}
-          -f caller_pr_nr=${{ steps.pr_string.outputs.pr_nr }}
+          -f caller_pr_nr=${{ github.event.number }}
           -f caller_base_branch=${{ github.base_ref || github.ref_name }}
           -f caller_head_branch=${{ github.head_ref }}
           -f head_sha=${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/dispatch-pr-build.yml
+++ b/.github/workflows/dispatch-pr-build.yml
@@ -22,7 +22,7 @@ jobs:
 
           # Add any PR dependencies to the list of pulls
           pulls+=($(grep -P '^/(jenkins-pr-deps|jpd|prd)' <<< "${{ github.event.pull_request.body }}" | \
-            grep -ioP '(Graylog2/\S+?#|https?://github.com/Graylog2/\S+?/pull/)[0-9]+') || true)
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-pr-build.yml
```diff
@@ -0,0 +1,60 @@
...
+          if [ "${{ github.repository }}" = Graylog2/graylog2-server ]; then
+              core_pr=$pull_pr
+              enterprise_pr=${{ steps.retrieve_enterprise_pr.outputs.pr }}
+          else
+              core_pr="${{ steps.retrieve_core_pr.outputs.pr }}"
+              enterprise_pr=$pull_pr
+          fi
+          echo pr_nr="$pull_nr" >> $GITHUB_OUTPUT
+          PRS=( $core_pr $enterprise_pr )
+          echo pr_string=$(echo ${PRS[@]} | tr ' ' ',') >> $GITHUB_OUTPUT
```
bernd, line 42: I think we can replace the collection of dependencies and building the pull request string with the following.

- Since we only trigger the build on pull requests, we don't need to check if the event is a pull request
- We don't need to dispatch on the repository name and can just take whatever is defined after `/jpd` (that's what Jenkins is currently doing as well)

What do you think?

```suggestion
      - name: Build pull requests string
        id: pr_string
        shell: bash
        run: |
          # Add the PR that triggered this workflow
          pulls=("${{ github.repository }}#${{ github.event.number }}")

          # Add any PR dependencies to the list of pulls
          pulls+=($(grep -P '^/(jenkins-pr-deps|jpd|prd)' <<< "${{ github.event.pull_request.body }}" | \
            grep -ioP '(Graylog2/\S+?#|https?://github.com/Graylog2/\S+?/pull/)[0-9]+'))

          echo "pr_string=$(tr ' ' ',' <<< "${pulls[@]}")" >> $GITHUB_OUTPUT
```

bernd (changes requested):

.github/workflows/dispatch-pr-build.yml
```diff
@@ -0,0 +1,60 @@
+name: Request dispatched PR Build
+
+on:
+  - pull_request
```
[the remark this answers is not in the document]
mpfz0r: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-multiple-events

`on` accepts single values, arrays, or a map.

.github/workflows/dispatch-pr-build.yml
mpfz0r: sure. let's keep it simple :+1:

mpfz0r: needs a small adjustment because the next step used the `pr_nr` output. will fix that with a separate commit.

bernd (approved):

PR #17202, opened 2023-11-07
Add changelog check workflow

mpfz0r:
/nocl

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,19 @@
+name: "Changelog Check"
+
+on:
+  pull_request:
+    types:
+      - "opened"
+      - "synchronize"
+      - "reopened"
+      - "edited"
+    issue_comment:
+
+
+jobs:
+  test:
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,19 @@
+name: "Changelog Check"
+
+on:
+  pull_request:
+    types:
+      - "opened"
+      - "synchronize"
+      - "reopened"
+      - "edited"
+    issue_comment:
```
bernd, line 10: Does this trigger the workflow on every issue comment in this repository?

bernd (changes requested):

.github/workflows/changelog.yml
mpfz0r: it does. actually the `edited` type of the pull_request should be enough. 
Although it will also trigger it if the title is edited. But I think we can live with that for now

bernd (approved):

PR #17246, opened 2023-11-10
Add CRC to sidecar default template

### LINKED ISSUE #17040: Update Sidecar default templates to newest version
With https://github.com/Graylog2/graylog2-server/pull/15571
we introduced better sidecar default configurations.

Parts of these improved default configs are only rolled out to newly installed Graylog servers.
This is because the migration rather simplistically only creates the defaults once.
Existing ones will never be updated.

The reason behind that, is that we cannot easily change configurations, which might
already be in productive use without causing a behavior change.

What we could do however, is to update the "default collector template configs". 
But only if the user hasn't changed them already.
For this, we could maintain a list of checksums, created from our previously shipped default templates,
to detect if the user hasn't modified them already.
In this case it would be safe to overwrite the default collectors templates.

(This is issue is the longer version of this comment
https://github.com/Graylog2/graylog2-server/issues/14228#issuecomment-1668064026 )

patrickmann:
Resolves #17040

Add a CRC for field `default_template` in collection `sidecar_collectors`. Disallow updates to the template if it 
has been changed, i.e. the CRC no longer matches.

## How Has This Been Tested?

Tested locally:
- CRC field is added if not present
- if CRC is null, compare to hardcoded CRCs
- templates are not updated if CRC mismatch
- templates updated if CRC match
- template not updated if update is identical to current version

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog/plugins/sidecar/rest/models/Collector.java
```diff
@@ -66,6 +101,31 @@ public abstract class Collector {
     @Nullable
     public abstract String defaultTemplate();
 
+    @JsonProperty(FIELD_DEFAULT_TEMPLATE_CRC)
+    @Nullable
+    public abstract Long defaultTemplateCRC();
+
+    @JsonIgnore
+    public boolean defaultTemplateUpdated() {
+        long crc = checksum(defaultTemplate().getBytes(StandardCharsets.UTF_8));
```
AntonEbel, line 110: `defaultTemplate()` is `@Nullable` so `getBytes` may produce `NullPointerException`. Don't know if this really can happen but I think we should handle this to be on the safe side.

AntonEbel (changes requested):

AntonEbel:
I also realized that when you update a collector via the UI. The default_template_crc field is deleted from the DB because the UI does not send this value to the backend. Maybe we need a change to the frontend here.

graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/CollectorService.java
```diff
@@ -101,7 +102,8 @@ public Collector fromRequest(Collector request) {
                 request.executablePath(),
                 request.executeParameters(),
                 request.validationParameters(),
-                request.defaultTemplate());
+                request.defaultTemplate(),
+                request.defaultTemplate() == null ? null : Collector.checksum(request.defaultTemplate().getBytes(StandardCharsets.UTF_8)));
```
AntonEbel, line 106: I think this won't work if there are changes in the default template, because now you would save the CRC from the changed template. If you want to do it without frontend changes, you have to load the existing collector and take the CRC from it.

AntonEbel (changes requested):

graylog2-server/src/main/java/org/graylog/plugins/sidecar/services/CollectorService.java
patrickmann: Yeah - quick and dirty is not a good idea. I actually encountered this myself earlier. Will make a more considered chante.

patrickmann: The FE shouldn't need to know anything about CRC. I added the required logic to the REST APIs.

AntonEbel: I think we can always set defaultTemplateCRC to null here. Otherwise it's a bit confusing why we always calculate a new CRC.

AntonEbel (approved):
LGTM

Issue #17247, opened 2023-11-10
Data Node: Publish node status


moesterheld:
## What?
Data Nodes should publish their status so it can be visualized and used for validating actions in Graylog.

## Why?
necessary for data node life cycle management in Graylog

janheise:
done

Issue #16820, opened 2023-10-05
DataNode: remove a node from the cluster


janheise:
## What?

Adding a new DataNode to an existing cluster is very easy - I think we should also automate the removal of nodes, too.

Also, seems not too complicated to automate it. The links describe the process for ES but I guess it's similar for OS.

https://www.elastic.co/guide/en/elasticsearch/reference/current/add-elasticsearch-nodes.html
https://opster.com/guides/elasticsearch/operations/elasticsearch-remove-node/

## Your Environment

* Graylog Version:
* OpenSearch Version:
* MongoDB Version:
* Operating System:
* Browser version:

janheise:
done

PR #17248, opened 2023-11-10
Remove a data node from the cluster

moesterheld:
## Description
Adds the REMOVING/REMOVED states and transitions to the datanode
image

Removal and Reset can be triggered from the Graylog REST API.
Graylog Server publishes events to the cluster event bus which are then consumed by the data nodes.

Next step is #17247 which is also needed to delete removed nodes from `unicast_hosts` upon restart.

## Motivation and Context
backend functionality for #16820 

## How Has This Been Tested?
- create data node cluster
- call DELETE on <graylog>/api/datanode/<nodeid> 
- when finished, POST to <graylog>/api/datanode/<nodeid>/reset

## Types of changes

## Checklist:


janheise:
@moesterheld if it makes sense, please add an IT test

data-node/src/main/java/org/graylog/datanode/rest/StatusController.java
```diff
@@ -48,4 +49,9 @@ public DataNodeStatus status() {
                 new StatusResponse(datanodeConfiguration.opensearchDistributionProvider().get().version(), openSearch.processInfo())
         );
     }
+
+    @DELETE
+    public void remove() {
```
todvora, line 54: I'd suggest moving this to a different controller (management?), to keep the operations separated. 

The management controller probably shouldn't be accessible if the datanode is started in an insecure mode and there is no auth that would guard these, potentially very dangerous, actions.

todvora (commented):
I think we need to secure the delete action or disable it if we are running in the insecure mode of the datanode.

data-node/src/main/java/org/graylog/datanode/rest/StatusController.java
janheise: I agree

moesterheld: Good point. I only added the endpoint to be able to trigger removal from the integration test. 
I'll move it to a separate controller and secure it with an annotation

todvora (approved):
LGTM!

PR #17255, opened 2023-11-13
Allow configuring exact datanode directories without automatic nodeid subdirs

todvora:
## Description
This PR changes handling of datanode directories. Originally we used directories configured in opensearch_config_location, opensearch_data_location and opensearch_logs_location as root directories, creating a Node-ID named subdirectory for each. Now we use the provided directories directly, without the subdir logic. 

To prevent using the same directories by several datanodes, we are now writing `datanode.lock` files in each of the three directories. 

TODO: this should use proper locking mechanism for the lock file.

## Motivation and Context
The Node-ID subdirectories would prevent us from doing a rolling upgrade of the embedded opensearch, as we are unable to point the configuration to existing opensearch data directory.

**This change is backwards compatible**. We'll still use the old node-id named subdirs if we detect them. 

## How Has This Been Tested?
Added unit tests, existing set of integration tests, manual testing

## Types of changes

## Checklist:


data-node/src/main/java/org/graylog/datanode/Configuration.java
```diff
@@ -93,7 +93,7 @@ public class Configuration extends BaseConfiguration {
     @Parameter(value = "opensearch_config_location", required = true, validators = DirectoryWritableValidator.class)
     private Path opensearchConfigLocation = Path.of("datanode/config");
 
-    @Parameter(value = "config_location", validators = DirectoryPathReadableValidator.class)
```
moesterheld, line 96: unused import can be removed

data-node/src/main/java/org/graylog/datanode/bindings/PreflightChecksBindings.java
```diff
@@ -29,6 +30,7 @@ public class PreflightChecksBindings extends AbstractModule {
     protected void configure() {
         addPreflightCheck(OpensearchConfigSync.class);
         addPreflightCheck(OpensearchBinPreflightCheck.class);
+        addPreflightCheck(DatanodeDirectoriesLockfileCheck.class);
```
moesterheld, line 33: should this test run before `OpensearchConfigSync`?

moesterheld (approved):
two minor comments, otherwise works and lgtm.

moesterheld (changes requested):
waiting for the changes to the locking mechanism

data-node/src/main/java/org/graylog/datanode/bindings/PreflightChecksBindings.java
todvora: This just adds the binding to a map, so I think we can't really influence the order of the preflight checks. This would need some deeper refactoring.

data-node/src/main/java/org/graylog/datanode/Configuration.java
todvora: Done, thanks!

data-node/src/main/java/org/graylog/datanode/bindings/PreflightChecksBindings.java
moesterheld: from https://google.github.io/guice/api-docs/4.1/javadoc/com/google/inject/multibindings/MapBinder.html
"The map's iteration order is consistent with the binding order" , at least if defined in the same module.

moesterheld (approved):
LGTM

as mentioned, I would change the preflight checks order as this should work (see my new comment). but that's not a must.

PR #17260, opened 2023-11-13
Persist currently selected perspective in user preferences.

dennisoelkers:
## Motivation and Context

This PR is adding functionality to persist the current perspective, when it is changed. The way it works is the same as for the current theme:

  - when the current user is read-only, the setting is persisted in local storage
  - when the current user is read-writable, the setting is persisted in the user preferences

As this is very similar to theming, an abstract `usePersistedSetting` hook is introduced, but inside the `PerspectiveProvider` module. A follow-up PR is already prepared which extracts and reuses it from within `usePreferredColorScheme`.

Fixes Graylog2/graylog-plugin-enterprise#6103.

/nocl

## Types of changes

## Checklist:


zeeklop (approved):
Perspective option persistence tested successfully.
Loaded the application, changed the perspective, closed the browser and when I opened it again I landed on the perspective previously selected.

LGTM.

linuspahl (approved):
There is one linter hint, but otherwise it looks good!

PR #17318, opened 2023-11-15
Remove no longer needed CSS prefixes.

linuspahl:
## Description

With this PR we are removing no longer needed CSS prefixes. We can remove these because all related styles are well supported, by the browsers we support.

https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix

/nocl


dennisoelkers (approved):
Thanks! :pray:

PR #17319, opened 2023-11-15
Move pull request string extraction to gpi and add autodetection for pr-deps comment change

mpfz0r:
/nocl

The edit event can detect whether the pr-deps comment has changed and will retrigger the build automatically:

image

### DIFF
--- a/graylog2-server/src/main/resources/org/graylog2/featureflag/feature-flag.config
+++ b/graylog2-server/src/main/resources/org/graylog2/featureflag/feature-flag.config
@@ -68,6 +68,9 @@ cloud_inputs=on
 # Preflight web for cluster bootstrapping
 preflight_web=on
 
+# Instant archiving is disabled by default
+instant_archiving=off
+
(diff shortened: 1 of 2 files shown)


graylog2-server/src/main/resources/org/graylog2/featureflag/feature-flag.config
```diff
@@ -68,9 +68,6 @@ cloud_inputs=on
 # Preflight web for cluster bootstrapping
 preflight_web=on
 
-# Instant archiving is disabled by default
```
bernd, line 71: Unrelated change.

bernd (changes requested):

graylog2-server/src/main/resources/org/graylog2/featureflag/feature-flag.config
mpfz0r: ah. dang

bernd (approved):

PR #17347, opened 2023-11-16
Update stream throughput metric in route_to_stream pipeline function

patrickmann:
Resolves Graylog2/graylog-plugin-enterprise#6138

Update stream metrics in  `route_to_stream` pipeline function.

## Description

Stream throughput did not reflect messages that were routed via `route_to_stream` pipeline function, since this code path does not involve the `StreamRouterEngine`.

### DIFF
--- a/changelog/unreleased/pr-1111.toml
+++ b/changelog/unreleased/pr-1111.toml
@@ -1,0 +1,5 @@
+type = "fixed"
+message = "Show correct throughput when messages are routed via a pipeline rule."
+
+issues = ["Graylog2/graylog-plugin-enterprise#6138"]
+pulls = ["1111"]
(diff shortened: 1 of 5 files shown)


mpfz0r (changes requested):
While this works,

the fix is a bit too simplistic.
With pipeline rules you can also remove streams from messages,
our add them twice.

Here is an alternative idea how to fix that: 
https://github.com/Graylog2/graylog2-server/pull/17456
But my fix feels a bit messy, because the error parts of these metrics are still in the streamrouter..
Maybe @bernd or @thll  have a better idea?

thll:
> Here is an alternative idea how to fix that: #17456 But my fix feels a bit messy, because the error parts of these metrics are still in the streamrouter.. Maybe @bernd or @thll have a better idea?

Good idea! I think it makes totally sense to delay recording of the metrics to a point in time where we know that the streams won't be changed anymore.

patrickmann:
Closing in favor of #17456

PR #17370, opened 2023-11-17
Resolve stream references to a view's state value widgets in addition to searches

ryan-carroll-graylog:
Fixes issue with content pack installations where the stream reference of the `search` entity has been removed but not the stream reference in the associated `state` `widget`.

## Description

Seams to be a bug in conjunction with the code here: https://github.com/Graylog2/graylog2-server/blob/38c3bc6ad085f007f4d4b29ed2729d06f06cf9bf/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackPersistenceService.java#L130-L168

Added in https://github.com/Graylog2/graylog2-server/pull/10972

This issue occurs when you have a content pack with a dashboard and stream where the dashboard references the stream.
If the referenced stream doesn't exist when the content pack is uploaded, the code linked above removes the stream reference from the `search` entity but not the `widget` entities in the `state`.

When this happens the content pack installation will sometimes fail (sometimes not) depending of the random order of the set containing the entites to install, because the resolution code changed in this PR is not handling the `widgets` in the search to ensure the proper order of entities: 
https://github.com/Graylog2/graylog2-server/blob/38c3bc6ad085f007f4d4b29ed2729d06f06cf9bf/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java#L119
https://github.com/Graylog2/graylog2-server/blob/38c3bc6ad085f007f4d4b29ed2729d06f06cf9bf/graylog2-server/src/main/java/org/graylog2/contentpacks/ContentPackService.java#L414

Not sure if the best solution is this fix, handling the `widget` entities in `ContentPackPersistenceService.filterMissingResourcesAndInsert()`, or both.

## How Has This Been Tested?

I was able to reproduce this reliably with these two identical content packs (minus identifying info), but because of the non-deterministic nature of this can't guarantee the same results when testing elsewhere.

Upload both on a system where the `Illuminate:Cisco Device Messages` does NOT exist.
Installing this pack fails:
[content-pack-ebe297d5-469c-4c5c-a978-0a1ca7d558ce-1.json](https://github.com/Graylog2/graylog2-server/files/13397959/content-pack-ebe297d5-469c-4c5c-a978-0a1ca7d558ce-1.json)

But this
[content-pack-2a26feda-c426-4c78-989d-4a05e68ace8a-1.json](https://github.com/Graylog2/graylog2-server/files/13397960/content-pack-2a26feda-c426-4c78-989d-4a05e68ace8a-1.json)
 one succeeds:

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/ViewFacade.java
+++ b/graylog2-server/src/main/java/org/graylog2/contentpacks/facades/ViewFacade.java
@@ -42,6 +42,7 @@ import org.graylog2.contentpacks.model.entities.NativeEntityDescriptor;
 import org.graylog2.contentpacks.model.entities.SearchEntity;
 import org.graylog2.contentpacks.model.entities.ViewEntity;
 import org.graylog2.contentpacks.model.entities.ViewStateEntity;
+import org.graylog2.contentpacks.model.entities.WidgetEntity;
 import org.graylog2.contentpacks.model.entities.references.ValueReference;
 import org.graylog2.plugin.database.users.User;
 import org.graylog2.shared.users.UserService;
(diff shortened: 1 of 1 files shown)


ryan-carroll-graylog:
@danotorrey I think I've nailed down the source of the issue you were seeing in https://github.com/Graylog2/graylog2-server/pull/17089#pullrequestreview-1729665681

Need to do a little more testing/research to nail down exactly how to handle this but after some testing and recreating this solely on master I'm pretty confident the issue is not caused by the changes in https://github.com/Graylog2/graylog2-server/pull/17089
so thinking it's better to tackle in this separate PR.

Definitely would love your feedback on whether this makes sense and tracks with what you were seeing.

danotorrey:
Thanks @ryan-carroll-graylog! Working on testing that now...

ryan-carroll-graylog:
Closing in favor of adding this change as a part of https://github.com/Graylog2/graylog2-server/pull/17089, to handle the stream title functionality correctly with this.

PR #17378, opened 2023-11-20
Restore finding user menu in tests.

dennisoelkers:
## Motivation and Context

In #17344, the navigation dropdown component was replaced.  Unfortunately, this broke finding the user menu programmatically.

This PR is now adding a title prop to make the user menu findable again.

/nocl

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-web-interface/src/components/navigation/UserMenu.tsx
+++ b/graylog2-web-interface/src/components/navigation/UserMenu.tsx
@@ -58,9 +58,7 @@ const UserMenu = ({ fullName, readOnly = true, userId }: Props) => {
   };
 
   return (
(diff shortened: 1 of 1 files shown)


ousmaneo (approved):
LGTM!

linuspahl (commented):
The core problem seems to be that we are no longer passing the `aria-label` to the dropdown button inside `NavDropdown`. Since we are also defining an aria label when using the `NavDropdown ` in `AdaptableQueryTabs`, we should fix this prop. 

I can imagine replacing the `NavDropdown` `aria-label` prop with e.g. a `hoverTitle` prop and using it as a title for `DropdownTrigger`. This way:
- we show the title on hover
- we define the title for the actual button and not just the icon

Issue #15194, opened 2023-04-12
Autoregister nodes in NodeService#markAsAlive


todvora:
## What?
We are always using the markAsAlive call with a try-catch wrapper that registers the node if the node is currently unknown:

https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/java/org/graylog2/periodical/NodePingThread.java#L68

## Why?
We could do the same operation already on the service level instead of forcing all usages to go through this try-catch logic. This would simplify our code and make it more resilient.

todvora:
Implemented in https://github.com/Graylog2/graylog2-server/pull/17382

PR #17382, opened 2023-11-20
Publish and show data node state information

moesterheld:
## Description
Data nodes now publish the state of the state machine in the database.
This information is also included in the node in Graylog. An endpoint specifically for data nodes has been added which provides this information in the paginated list needed for the new data table.

## Motivation and Context
The data node page should include status information.
Also needed for using the status information to validated actions performed against data nodes in the future.

## Types of changes

## Checklist:

### DIFF
--- a/graylog2-server/src/main/java/org/graylog/security/certutil/CertRenewalServiceImpl.java
+++ b/graylog2-server/src/main/java/org/graylog/security/certutil/CertRenewalServiceImpl.java
@@ -49,6 +49,7 @@ import java.security.cert.X509Certificate;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -203,13 +204,19 @@ public class CertRenewalServiceImpl implements CertRenewalService {
     @Override
     public List<DataNode> findNodes() {
         final Map<String, Node> activeDataNodes = nodeService.allActive(Node.Type.DATANODE);
-        return activeDataNodes.values().stream().map(node -> {
+        return addProvisioningInformation(activeDataNodes.values());
+    }
+
+    @Override
+    public List<DataNode> addProvisioningInformation(Collection<Node> nodes) {
+        return nodes.stream().map(node -> {
             final var keystore = loadKeyStoreForNode(node);
             final var certificate = keystore.flatMap(this::getCertificateForNode);
             final var certValidUntil = certificate.map(cert -> cert.getNotAfter().toInstant().atZone(ZoneId.systemDefault()).toLocalDateTime());
             final var config = getDataNodeProvisioningConfig(node).orElseThrow(() -> new IllegalStateException("No config found for data node " + node.getNodeId()));
             return new DataNode(node.getNodeId(),
                     node.getType(),
+                    node.getDataNodeStatus(),
                     node.getTransportAddress(),
                     config.state(),
                     config.errorMsg(),
--- a/graylog2-server/src/main/java/org/graylog2/cluster/PaginatedNodeService.java
+++ b/graylog2-server/src/main/java/org/graylog2/cluster/PaginatedNodeService.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2020 Graylog, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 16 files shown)


graylog2-server/src/main/java/org/graylog2/cluster/PaginatedNodeService.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public PaginatedList<Node> searchPaginated(SearchQuery query,
+                                               String sortByField, String sortOrder, int page, int perPage) {
+        final PaginatedList<NodeImpl> results = findPaginatedWithQueryFilterAndSort(query.toDBQuery(), node -> true,
+                getSortBuilder(sortOrder, sortByField), page, perPage);
+        return new PaginatedList<>(results.delegate().stream().map(n -> (Node) n).toList(),
+                results.pagination().total(), results.pagination().page(), results.pagination().perPage());
+    }
+
+    public PaginatedList<Node> searchPaginatedDatanodes(SearchQuery query,
```
todvora, line 46: Maybe generalize this service and let user define the node type as a param?

graylog2-server/src/main/java/org/graylog2/rest/models/system/cluster/responses/NodeSummary.java
```diff
@@ -56,6 +59,10 @@ public boolean isMaster() {
     @JsonProperty
     public abstract String hostname();
 
+    @JsonProperty
+    @Nullable
+    public abstract DataNodeStatus datanodeStatus();
```
todvora, line 64: What about creating a DataNodeSummary class that would provide all these datanode-specific methods and the nodeservice would return the correct implementation based on the type / database entry info?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterResource.java
```diff
@@ -79,6 +118,33 @@ public NodeSummaryList nodes() {
         return NodeSummaryList.create(nodeList);
     }
 
+    @GET
+    @Timed
+    @Path("/datanodes")
+    @ApiOperation(value = "Get a paginated list of all datanodes in this cluster")
+    public PageListResponse<CertRenewalService.DataNode> dataNodes(@ApiParam(name = "page") @QueryParam("page") @DefaultValue("1") int page,
+                                                                   @ApiParam(name = "per_page") @QueryParam("per_page") @DefaultValue("50") int perPage,
+                                                                   @ApiParam(name = "query") @QueryParam("query") @DefaultValue("") String query,
```
todvora, line 127: What's the purpose of the query here?

graylog2-server/src/main/java/org/graylog2/rest/resources/system/ClusterResource.java
```diff
@@ -53,14 +67,39 @@
 @Produces(MediaType.APPLICATION_JSON)
 public class ClusterResource extends RestResource {
     private final NodeService nodeService;
+    private final PaginatedNodeService paginatedNodeService;
```
todvora, line 70: When I see the dependencies here, I'd be thinking about moving it to an own resource.

graylog2-server/src/main/java/org/graylog/security/certutil/CertRenewalServiceImpl.java
```diff
@@ -203,13 +204,19 @@ public void initiateRenewalForNode(final String nodeId) {
     @Override
     public List<DataNode> findNodes() {
         final Map<String, Node> activeDataNodes = nodeService.allActive(Node.Type.DATANODE);
-        return activeDataNodes.values().stream().map(node -> {
+        return addProvisioningInformation(activeDataNodes.values());
+    }
+
+    @Override
+    public List<DataNode> addProvisioningInformation(Collection<Node> nodes) {
```
todvora, line 211: Could we rename the method to signalize that it extends the nodes collection and is not adding anything to the database?

todvora (changes requested):
I think we would benefit from a cleaner separation of the datanode functionality. I'd suggest splitting node services and controllers instead of extending existing classes.

todvora (approved):
LGTM! I really like the separation of server nodes and data nodes in the services and DTOs. Additionally the ping method makes everything easier. It implements a feature requested in https://github.com/Graylog2/graylog2-server/issues/15194

PR #17402, opened 2023-11-21
add validations for data node lifecycle functions

### LINKED ISSUE #17383: Add validations for data node lifecycle actions
## What?
Operations which control the data node lifecycle should perform validations.

## Why?
When performing certain actions, it should be checked that the operations are valid for the current state of the data node cluster, i.e.
- only allow one node removal at a time
- do not allow removal of the last node
- do a rolling restart of the data node cluster

moesterheld:
## Motivation and Context
closes #17383 

## Types of changes

## Checklist:


graylog2-server/src/main/java/org/graylog2/migrations/V20231107164300_CreateDataNodeManagerRole.java
```diff
@@ -38,6 +38,6 @@ public ZonedDateTime createdAt() {
     @Override
     public void upgrade() {
         helpers.ensureBuiltinRole("Data Node Manager", "Grants control to manage the data node cluster (built-in)",
-                Set.of(RestPermissions.DATANODE_REMOVE, RestPermissions.DATANODE_RESET));
+                Set.of(RestPermissions.DATANODE_REMOVE, RestPermissions.DATANODE_RESET, RestPermissions.DATANODE_STOP, RestPermissions.DATANODE_START));
```
todvora, line 41: Is it safe to modify an existing migration? Between releases only developers will be possibly influenced by this change, but maybe it would be safer to create a completely new migration?

todvora (approved):
LGTM, my only concern is the changed existing migration

graylog2-server/src/main/java/org/graylog2/migrations/V20231107164300_CreateDataNodeManagerRole.java
moesterheld: AFAIK the migrations will always run unless you implement a mechanism, so this should be fine in between releases.
But if you think it would be better, I can separate that into a new migration.

todvora: You are right, it will run every time. Then it's all good!

PR #17407, opened 2023-11-21
Allow displayName in systemConfiguration plugin export

### LINKED ISSUE #15939: Allow plugins to set their display name for the configuration page
Plugins can not configure their display name on the new configuration page. Instead the full class identifier is used, which also has an overflow issue.

## Expected Behavior

My expectation would be that this is configurable via the PluginManifest component.

## Current Behavior

image

## Context

With the refactoring of the configuration page in 5.x the plugins were moved in a sub-menu that uses "display names" for the plugins.
These are hard coded on the server-side. Since 5.1.3 (67839e96e97302870a4cefb573e179fc8e22acc0) there is a fallback for unknown plugins to default the full class string. 

I filled this a bug, because I consider this to be a regression. On the previous version of the configuration page the plugin names were properly displayed via the respective plugins config component. 

## Your Environment

* Graylog Version: 5.1.3

ousmaneo:
These changes refactor the Configuration menu to get the Plugin's `displayName` from the plugin configuration instead of hardcoded values. This will allow custom-made Plugins to set their display name.

fix #15939 

## Types of changes

## Checklist:


grotlue (approved):
Looks good! Is there any point in plugin related documentation to put this so custom plugins devs are aware?

PR #14502, opened 2023-01-24
Feature/14285 migrate event definitions page to new bulk friendly design

moesterheld:
Migrate Event Definitions and Event Notifications to new Entity List

closes #14285 

## Description
- added `paginated` endpoints for event definitions and notifications, deprecated old endpoints
- changed entity list in ui to new component

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (non-breaking change)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #17435, opened 2023-11-22
Fix saving of system notification event definition

mpfz0r:
On save, the UI is passing back the "scheduled=false" param from the context.

There's several ways we could fix this, but it's probably easiest to not throw an exception, but just ignore the unschedule call.

This probably got broken with #14502


moesterheld (approved):
LGTM

PR #17436, opened 2023-11-22
Check vm.max_map_count in datanode preflight

### LINKED ISSUE #17430: create and include check for vm.max_map_count in DataNode startup/preflight UI before cert provisioning
OpenSearch requires `vm.max_map_count` to be a minimum value which is larger than the default on many systems. 
Ensure acceptable value in datanode preflight.

patrickmann:
Resolves #17430 

Check vm.max_map_count in datanode preflight


janheise (changes requested):
on macOS, `sysctl` is `/usr/sbin/sysctl` - we use `org.apache.commons.exec.OS.isFamilyMac()` in other places to distinguish. Can you please add this?

janheise:
Sorry that I was wrong and you had to revert. Linux and Docker on macOS need the check. On macOS natively (OpenSearch is not supported) the warning is fine as it works without the setting.

janheise (approved):
LGTM

